PR #1023, opened 2018-08-10
Add support for xids in the cluster manager

pnarayanan:
Xids allows for selectively ignoring certain InstanceConfig changes.
This is aimed to help during deployments.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
@@ -152,6 +152,13 @@ public class ClusterMapConfig {
   @Default(MAX_REPLICAS_ALL_DATACENTERS)
   public final String clusterMapDefaultPartitionClass;
 
+  /**
+   * The current xid for this cluster manager. Any changes beyond this xid will be ignored by the cluster manager.
+   */
+  @Config("clustermap.current.xid")
+  @Default("Long.MAX")
+  public final Long clustermapCurrentXid;
+
   public ClusterMapConfig(VerifiableProperties verifiableProperties) {
     clusterMapFixedTimeoutDatanodeErrorThreshold =
         verifiableProperties.getIntInRange("clustermap.fixedtimeout.datanode.error.threshold", 3, 1, 100);
@@ -181,5 +188,6 @@ public class ClusterMapConfig {
     clusterMapResolveHostnames = verifiableProperties.getBoolean("clustermap.resolve.hostnames", true);
     clusterMapDefaultPartitionClass =
         verifiableProperties.getString("clustermap.default.partition.class", MAX_REPLICAS_ALL_DATACENTERS);
+    clustermapCurrentXid = verifiableProperties.getLong("clustermap.current.xid", Long.MAX_VALUE);
   }
 }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
@@ -52,6 +52,7 @@ public class ClusterMapUtils {
   static final String DATACENTER_STR = "datacenter";
   static final String DATACENTER_ID_STR = "id";
   static final String SCHEMA_VERSION_STR = "schemaVersion";
+  static final String XID_STR = "xid";
   static final int MIN_PORT = 1025;
   static final int MAX_PORT = 65535;
   static final long MIN_REPLICA_CAPACITY_IN_BYTES = 1024 * 1024 * 1024L;
@@ -178,6 +179,17 @@ public class ClusterMapUtils {
   static Integer getSslPortStr(InstanceConfig instanceConfig) {
     String sslPortStr = instanceConfig.getRecord().getSimpleField(SSLPORT_STR);
     return sslPortStr == null ? null : Integer.valueOf(sslPortStr);
+  }
+
+  /**
+   * Get the xid associated with this instance. The xid is like a timestamp or a change number, so if it is absent,
+   * a value representing the earliest point in time is returned.
+   * @param instanceConfig the {@link InstanceConfig} associated with the interested instance.
+   * @return the xid associated with the given instance.
+   */
+  static Long getXid(InstanceConfig instanceConfig) {
+    String xid = instanceConfig.getRecord().getSimpleField(XID_STR);
+    return xid == null ? Long.MIN_VALUE : Long.valueOf(xid);
   }
 
   /**
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
@@ -187,7 +187,7 @@ public class ClusterMapUtils {
    * @param instanceConfig the {@link InstanceConfig} associated with the interested instance.
    * @return the xid associated with the given instance.
    */
-  static Long getXid(InstanceConfig instanceConfig) {
+  static long getXid(InstanceConfig instanceConfig) {
     String xid = instanceConfig.getRecord().getSimpleField(XID_STR);
     return xid == null ? Long.MIN_VALUE : Long.valueOf(xid);
   }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
@@ -74,6 +74,7 @@ class HelixClusterManager implements ClusterMap {
   private final AtomicLong sealedStateChangeCounter = new AtomicLong(0);
   final HelixClusterManagerMetrics helixClusterManagerMetrics;
   private final PartitionSelectionHelper partitionSelectionHelper;
+  private final long currentXid;
 
   /**
    * Instantiate a HelixClusterManager.
@@ -85,6 +86,7 @@ class HelixClusterManager implements ClusterMap {
   HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceName, HelixFactory helixFactory,
       MetricRegistry metricRegistry) throws IOException {
     this.clusterMapConfig = clusterMapConfig;
+    currentXid = clusterMapConfig.clustermapCurrentXid;
     this.metricRegistry = metricRegistry;
     clusterName = clusterMapConfig.clusterMapClusterName;
     helixClusterManagerCallback = new HelixClusterManagerCallback();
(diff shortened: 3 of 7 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -359,21 +372,31 @@ private void initializeInstances(List<InstanceConfig> instanceConfigs) throws Ex
...
+                    instanceConfig);
+              } else {
+                HashSet<String> sealedReplicas = new HashSet<>(getSealedReplicas(instanceConfig));
+                HashSet<String> stoppedReplicas = new HashSet<>(getStoppedReplicas(instanceConfig));
+                for (AmbryReplica replica : ambryDataNodeToAmbryReplicas.get(node)) {
+                  replica.setSealedState(sealedReplicas.contains(replica.getPartitionId().toPathString()));
+                  replica.setStoppedState(stoppedReplicas.contains(replica.getPartitionId().toPathString()));
+                }
+              }
+              break;
```
jsjtzyy: Shouldn't we place `break` right before `default`?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -359,21 +372,31 @@ private void initializeInstances(List<InstanceConfig> instanceConfigs) throws Ex
...
+            String instanceName = instanceConfig.getInstanceName();
+            long instanceXid = getXid(instanceConfig);
+            AmbryDataNode node = instanceNameToAmbryDataNode.get(instanceName);
+            if (instanceXid <= currentXid) {
+              if (node == null) {
+                logger.info("Dynamic addition of new nodes is not yet supported, ignoring InstanceConfig {}",
+                    instanceConfig);
+              } else {
+                HashSet<String> sealedReplicas = new HashSet<>(getSealedReplicas(instanceConfig));
+                HashSet<String> stoppedReplicas = new HashSet<>(getStoppedReplicas(instanceConfig));
```
jsjtzyy: Can you change `HashSet<String> sealedReplicas` to `Set<String> sealedReplicas`? same for `stoppedReplicas `

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -335,17 +337,28 @@ private void initializeInstances(List<InstanceConfig> instanceConfigs) throws Ex
...
-          initializeDisksAndReplicasOnNode(datanode, instanceConfig);
-          instanceNameToAmbryDataNode.put(instanceName, datanode);
-          allInstances.add(instanceName);
-        } else {
-          logger.error("Unknown InstanceConfig schema version: {}, ignoring.", schemaVersion);
+        switch (schemaVersion) {
+          case 0:
+            String instanceName = instanceConfig.getInstanceName();
+            long instanceXid = getXid(instanceConfig);
+            if (instanceXid <= currentXid) {
```
jsjtzyy: Are we able to test this method ?  Same for `updateStateOfReplicas ` method.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/MockHelixCluster.java
```diff
@@ -206,6 +206,14 @@ InstanceConfig getInstanceConfig(String instanceName) {
     return instanceConfig;
   }
 
+  List<InstanceConfig> getAllInstanceConfigs() {
```
jsjtzyy: This method is used in next PR, right?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -359,21 +372,31 @@ private void initializeInstances(List<InstanceConfig> instanceConfigs) throws Ex
     private void updateStateOfReplicas(List<InstanceConfig> instanceConfigs) {
```
jsjtzyy: Can you update the java doc of both `updateStateOfReplicas` and `initializeInstances` method?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: Yes, missed it here. Thanks for catching.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: done.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: Yes, unit tests are being added. Will update.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/MockHelixCluster.java
pnarayanan: Yes, this is needed for the unit tests, which is in progress. I just put this up, so the review can start, since we seemed to need this quickly.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: Done.

pnarayanan:
@jsjtzyy Added unit tests for the new changes. Please take a look.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: Added tests.

ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
```diff
@@ -152,6 +152,13 @@
   @Default(MAX_REPLICAS_ALL_DATACENTERS)
   public final String clusterMapDefaultPartitionClass;
 
+  /**
+   * The current xid for this cluster manager. Any changes beyond this xid will be ignored by the cluster manager.
+   */
+  @Config("clustermap.current.xid")
+  @Default("Long.MAX")
```
jsjtzyy: Correct me if I am wrong, the default value of current xid is better to be `Long.MIN` otherwise, if not specified, both frontend and server will honor any instanceConfig changes.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -482,6 +486,60 @@ public void dynamicNodeAdditionsTest() throws Exception {
...
+
+    AmbryReplica ignoreInstanceReplica = null;
+    for (AmbryDataNode dataNode : clusterManager.getDataNodeIds()) {
+      String instanceName = ClusterMapUtils.getInstanceName(dataNode.getHostname(), dataNode.getPort());
+      if (instanceName.equals(ignoreInstanceName)) {
+        ignoreInstanceReplica = clusterManager.getReplicaIds(dataNode).get(0);
+        ignoreInstanceConfig.getRecord()
+            .setListField(STOPPED_REPLICAS_STR,
+                Collections.singletonList(ignoreInstanceReplica.getPartitionId().toPathString()));
+      }
```
jsjtzyy: Maybe here we can break the loop.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -482,6 +486,60 @@ public void dynamicNodeAdditionsTest() throws Exception {
...
+    }
+    helixCluster.triggerInstanceConfigChangeNotification();
+    // Because the XID was higher, the change reflecting this replica being stopped will not be absorbed.
+    assertFalse(ignoreInstanceReplica.isDown());
```
jsjtzyy: (Optional) I think it would be better to verify that clustermap indeed received the instanceConfig change and `ignoreInstanceReplica` in this instanceConfig was indeed in stopped state. Do you think it's feasible?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -677,5 +701,20 @@ long getAllocatedUsableCapacity() {
       return clusterWideAllocatedUsableCapacityBytes;
     }
   }
+
+  /**
+   * @return the current xid;
+   */
+  protected long getCurrentXid() {
```
jsjtzyy: This method is for future use, right? (looks like is never invoked in this PR)

jsjtzyy:
One thing we need to consider is when xid and override are both brought into code, 4 cases should be discussed.   (Following is my tentative analysis, feel free to comment)

        Xid                |        Override         |   Source of Truth
---------------------------------------------------------
     honor               |      enabled            |    Override
     ignore              |      enabled            |    Override
     honor               |      disabled           |     Xid
     ignore              |      disabled           |     Xid
----------------------------------------------------------

ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
pnarayanan: `Long.MIN` should not be the default, as that would cause all changes to be ignored. Setting default as `Long.MAX` preserves existing behavior.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
pnarayanan: true.

pnarayanan:
Addressed comments.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: This is a getter and setter for the current xid config of the Helix cluster manager (not the InstanceConfigs), so the bootstrap tool does not affect this. This is primarily for helping test the code, but in the future, if there is a way to dynamically set the configs then these can potentially be useful.

The other option is to reinitialize the `HelixClusterManager`.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
```diff
@@ -180,6 +181,17 @@ static Integer getSslPortStr(InstanceConfig instanceConfig) {
     return sslPortStr == null ? null : Integer.valueOf(sslPortStr);
   }
 
+  /**
+   * Get the xid associated with this instance. The xid is like a timestamp or a change number, so if it is absent,
+   * a value representing the earliest point in time is returned.
+   * @param instanceConfig the {@link InstanceConfig} associated with the interested instance.
+   * @return the xid associated with the given instance.
+   */
+  static Long getXid(InstanceConfig instanceConfig) {
```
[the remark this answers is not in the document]
pnarayanan: done.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -328,52 +330,74 @@ public void onInstanceConfigChange(List<InstanceConfig> configs, NotificationCon
...
+              } else {
+                Set<String> sealedReplicas = new HashSet<>(getSealedReplicas(instanceConfig));
+                Set<String> stoppedReplicas = new HashSet<>(getStoppedReplicas(instanceConfig));
+                for (AmbryReplica replica : ambryDataNodeToAmbryReplicas.get(node)) {
+                  replica.setSealedState(sealedReplicas.contains(replica.getPartitionId().toPathString()));
+                  replica.setStoppedState(stoppedReplicas.contains(replica.getPartitionId().toPathString()));
+                }
+              }
+            } else {
+              logger.info(
```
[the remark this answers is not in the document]
pnarayanan: This seemed useful to have as an info log as the update events themselves is expected to be low for the nodes whose updates are being ignored. I will make this a trace and add the metric instead.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: Removed these methods. Testing those flows by simulating the advancement of the current xid by moving back the xid in the `InstanceConfig`s.

pnarayanan:
Addressed the recent comments.

PR #1028, opened 2018-08-16
Fix the random NPE in xidTest

jsjtzyy:
This PR ensures that two different instanceConfigs are randomly selected in xidTest.


pnarayanan:
Thanks for identifying and fixing this. I was under the false impression that subsequent `nextInt()`s are guaranteed to be different.

PR #1047, opened 2018-09-10
[Move Replica] Make DiskSpaceAllocator thread-safe for runtime initialization

jsjtzyy:
Add two methods in ReserveFileMap for adding additional segments and deleting extra segments. With these changes, DiskSpaceAllocator would be thread safe when intializing again after adding/removing stores.


jsjtzyy:
Based on offline discussion, this PR cannot address all concurrent cases. However, for store addition/deletion, we can skip re-initialization step. That means don't call `initializePool` after adding/deleting stores without bringing any potential risks.  
So, let's hold off on committing this PR and move forward first. I created the issue #1061 to keep track of improvement on DiskSpaceAllocator. We can continue on this when we get a chance in the future.

cgtz:
Closing in favor of approaches mentioned in #1061

PR #1053, opened 2018-09-14
Ignore dynamic fields when comparing InstanceConfigs

pnarayanan:
When comparing InstanceConfig generated from json files with what
exists in Helix in order to determine if an update is necessary,
ignore the dynamic fields such as sealed lists and stopped lists,
since those are from now on Helix-only fields.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
@@ -505,6 +505,9 @@ class HelixBootstrapUpgradeUtil {
           ArrayList<String> newInstances = new ArrayList<>(instanceSetInStatic);
           Collections.shuffle(newInstances);
           resourceIs.setPreferenceList(partitionName, newInstances);
+          // Existing resources may not have ANY_LIVEINSTANCE set as the numReplicas (which allows for different replication
+          // for different partitions under the same resource). So set it here.
+          resourceIs.setReplicas(ResourceConfig.ResourceConfigConstants.ANY_LIVEINSTANCE.toString());
           resourceModified = true;
         }
       }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
@@ -508,7 +508,7 @@ class HelixBootstrapUpgradeUtil {
           resourceModified = true;
         }
       }
-      if (resourceModified) {
+      if (!dryRun && resourceModified) {
         if (resourceIs.getPartitionSet().isEmpty()) {
           dcAdmin.dropResource(clusterName, resourceName);
           resourcesDropped++;
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
@@ -419,13 +419,7 @@ class HelixBootstrapUpgradeUtil {
       InstanceConfig instanceConfigInHelix = dcAdmin.getInstanceConfig(clusterName, instanceName);
       InstanceConfig instanceConfigFromStatic =
           createInstanceConfigFromStaticInfo(dcToInstanceNameToDataNodeId.get(dcName).get(instanceName),
-              partitionsToInstancesInDc);
-      // Note about the following check: This compares ZNode data for exact equality, which suffices for now.
-      // However, in the future, this needs to discount comparisons for certain things that are dynamically set by
-      // instances (and hence will have outdated information in the static map). Such as
-      // 1. RO/RW status of replicas.
-      // 2. Replica availability (for which support has not yet been added).
-      // 3. xid.
+              partitionsToInstancesInDc, instanceToDiskReplicasMap, instanceConfigInHelix);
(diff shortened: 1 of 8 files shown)


ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
```diff
@@ -40,6 +43,7 @@
 import static com.github.ambry.clustermap.HelixBootstrapUpgradeTool.*;
 import static com.github.ambry.clustermap.TestUtils.*;
 import static com.github.ambry.utils.TestUtils.*;
+import static org.hamcrest.CoreMatchers.*;
```
jsjtzyy: Do we have to import this?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -570,10 +564,16 @@ public int compare(ReplicaId a, ReplicaId b) {
...
+   *                                These are the SEALED state and STOPPED_REPLICAS configurations. If this field is null,
+   *                                then these fields are derived from the json files. This can happen if this is a newly
+   *                                added node.
    * @return the constructed {@link InstanceConfig}
    */
-  private InstanceConfig createInstanceConfigFromStaticInfo(DataNodeId node,
-      Map<String, Set<String>> partitionToInstances) {
+  static InstanceConfig createInstanceConfigFromStaticInfo(DataNodeId node,
+      Map<String, Set<String>> partitionToInstances,
+      Map<String, Map<DiskId, SortedSet<Replica>>> instanceToDiskReplicasMap, InstanceConfig referenceInstanceConfig) {
```
jsjtzyy: Why pass `instanceToDiskReplicasMap` as a parameter into this method? It seems `instanceToDiskReplicasMap` is global variable and doesn't get changed in this method.

pnarayanan: Wanted to make this method static so that it can be easily tested.

ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
pnarayanan: It helps with one of the asserts. `hamcrest` comes with `junit`, so we are not bringing in anything new.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixParticipantTest.java
```diff
@@ -100,15 +100,15 @@ public void testGetAndSetReplicaSealedState() throws IOException {
...
     HelixManager helixManager = helixManagerFactory.getZKHelixManager(null, null, null, null);
     HelixAdmin helixAdmin = helixManager.getClusterManagmentTool();
     InstanceConfig instanceConfig = new InstanceConfig("someInstanceId");
     helixAdmin.setInstanceConfig(clusterName, instanceName, instanceConfig);
 
-    //Make sure the current sealedReplicas list is null
+    //Make sure the current sealedReplicas list is empty
     List<String> sealedReplicas = helixParticipant.getSealedReplicas();
-    assertNull("sealedReplicas is not null", sealedReplicas);
+    assertEquals("sealedReplicas should be empty", Collections.emptyList(), sealedReplicas);
```
jsjtzyy: Can you also change line204~line206 for `StoppedReplicas`?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -511,10 +505,13 @@ private void addUpdateResources(String dcName, Map<String, Set<String>> partitio
...
           Collections.shuffle(newInstances);
           resourceIs.setPreferenceList(partitionName, newInstances);
+          // Existing resources may not have ANY_LIVEINSTANCE set as the numReplicas (which allows for different replication
+          // for different partitions under the same resource). So set it here.
+          resourceIs.setReplicas(ResourceConfig.ResourceConfigConstants.ANY_LIVEINSTANCE.toString());
           resourceModified = true;
         }
       }
-      if (resourceModified) {
+      if (!dryRun && resourceModified) {
```
jsjtzyy: Correct me if I am wrong, if `!dryRun && resourceModified == true` and if `resourceIs.getPartitionSet().isEmpty() == false`, the extra partition has been removed from current resource, should we update the num of partition in current resource. Something like: `resourceIs.setNumPartitions( # of partition)`?

pnarayanan: It does not affect us really as Helix is fine with it and because we do not use a resource in any meaningful way. That said, I agree that this is a good idea, so have made the change.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/TestUtils.java
```diff
@@ -481,6 +482,8 @@ static void checkReturnedPartitions(List<? extends PartitionId> partitionIds,
 
     private HardwareLayout hardwareLayout;
 
+    ClusterMapConfig clusterMapConfig;
```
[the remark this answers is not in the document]
pnarayanan: Done.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixParticipantTest.java
pnarayanan: Done.

PR #1079, opened 2018-10-02
Save local backup when using AccountUpdateTool

cgtz:
Allow the user to set up a backup directory path so that the state
before performing an account metadata update is saved.

### DIFF
--- a/ambry-tools/src/main/java/com.github.ambry/account/AccountUpdateTool.java
+++ b/ambry-tools/src/main/java/com.github.ambry/account/AccountUpdateTool.java
@@ -159,6 +159,13 @@ public class AccountUpdateTool {
         .ofType(Short.class)
         .defaultsTo(Container.getCurrentJsonVersion());
 
+    ArgumentAcceptingOptionSpec<String> backupDirOpt = parser.accepts("containerSchemaVersion",
+        "Optional local backup directory path. Defaults to /tmp/account-update-tool-backups.")
+        .withRequiredArg()
+        .describedAs("backup_dir")
+        .ofType(String.class)
+        .defaultsTo("/tmp/account-update-tool-backups");
+
     parser.accepts("help", "print this help message.");
 
     parser.accepts("h", "print this help message.");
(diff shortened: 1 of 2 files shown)


ambry-tools/src/main/java/com.github.ambry/account/AccountUpdateTool.java
```diff
@@ -159,6 +159,13 @@ public static void main(String args[]) throws Exception {
         .ofType(Short.class)
         .defaultsTo(Container.getCurrentJsonVersion());
 
+    ArgumentAcceptingOptionSpec<String> backupDirOpt = parser.accepts("containerSchemaVersion",
+        "Optional local backup directory path. Defaults to /tmp/account-update-tool-backups.")
+        .withRequiredArg()
+        .describedAs("backup_dir")
+        .ofType(String.class)
+        .defaultsTo("/tmp/account-update-tool-backups");
```
jsjtzyy: please use constants for this default directory path.

ambry-tools/src/main/java/com.github.ambry/account/AccountUpdateTool.java
```diff
@@ -235,6 +245,7 @@ private static AccountService getHelixAccountService(String zkServer, String sto
         String.valueOf(zkSessionTimeoutMs));
     helixConfigProps.setProperty(HelixAccountServiceConfig.ZK_CLIENT_CONNECT_STRING_KEY, zkServer);
     helixConfigProps.setProperty(HelixPropertyStoreConfig.HELIX_PROPERTY_STORE_PREFIX + "root.path", storePath);
+    helixConfigProps.setProperty(HelixAccountServiceConfig.BACKUP_DIRECTORY_KEY, backupDir);
```
jsjtzyy: Do we need to specify default `backDir` in `HelixAccountServiceConfig`? Currently, the default value is empty string.

cgtz: This feature is disabled by default. That is why the default is empty string

jsjtzyy:
lets merge this tool first.

PR #1084, opened 2018-10-09
Make cross-colo listening optional within Helix ClusterManager

pnarayanan:
Introduce listenCrossColo config in ClusterMapConfig that defaults
to true. It can be set to false when making sensitive changes to
the cluster information in ZK so that a bad change does not get
propagated across colos. This helps in safely and selectively
introducing new changes.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
@@ -33,6 +33,7 @@ import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 import org.I0Itec.zkclient.IZkDataListener;
 import org.apache.helix.AccessOption;
 import org.apache.helix.HelixManager;
@@ -364,6 +365,15 @@ class HelixClusterManager implements ClusterMap {
       }
     }
     return null;
+  }
+
+  /**
+   * @return a map of datacenter names to {@link HelixManager}
+   */
+  Map<String, HelixManager> getHelixManagerMap() {
+    return dcToDcZkInfo.entrySet()
+        .stream()
+        .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().helixManager));
   }
 
   /**
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
@@ -116,41 +116,42 @@ class HelixClusterManager implements ClusterMap {
         String dcName = entry.getKey();
         String zkConnectStr = entry.getValue().getZkConnectStr();
         ClusterChangeHandler clusterChangeHandler = new ClusterChangeHandler(dcName);
-        // Initialize from every datacenter in a separate thread to speed things up.
-        Utils.newThread(new Runnable() {
-          @Override
-          public void run() {
-            try {
-              HelixManager manager;
-              if (dcName.equals(clusterMapConfig.clusterMapDatacenterName)) {
-                manager = localManager;
-              } else {
-                manager =
-                    helixFactory.getZKHelixManager(clusterName, selfInstanceName, InstanceType.SPECTATOR, zkConnectStr);
-                logger.info("Connecting to Helix manager at {}", zkConnectStr);
-                manager.connect();
-                logger.info("Established connection to Helix manager at {}", zkConnectStr);
-              }
-              DcInfo dcInfo = new DcInfo(dcName, entry.getValue(), manager, clusterChangeHandler);
-              dcToDcZkInfo.put(dcName, dcInfo);
-              dcIdToDcName.put(dcInfo.dcZkInfo.getDcId(), dcName);
-
-              // The initial instance config change notification is required to populate the static cluster
-              // information, and only after that is complete do we want the live instance change notification to
-              // come in. We do not need to do anything extra to ensure this, however, since Helix provides the initial
-              // notification for a change from within the same thread that adds the listener, in the context of the add
-              // call. Therefore, when the call to add a listener returns, the initial notification will have been
-              // received and handled.
-              manager.addInstanceConfigChangeListener(clusterChangeHandler);
-              logger.info("Registered instance config change listeners for Helix manager at {}", zkConnectStr);
-              // Now register listeners to get notified on live instance change in every datacenter.
-              manager.addLiveInstanceChangeListener(clusterChangeHandler);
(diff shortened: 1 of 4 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -116,41 +116,42 @@
...
+            // notification for a change from within the same thread that adds the listener, in the context of the add
+            // call. Therefore, when the call to add a listener returns, the initial notification will have been
+            // received and handled.
+            manager.addInstanceConfigChangeListener(clusterChangeHandler);
+            logger.info("Registered instance config change listeners for Helix manager at {}", zkConnectStr);
+            // Now register listeners to get notified on live instance change in every datacenter.
+            manager.addLiveInstanceChangeListener(clusterChangeHandler);
+            logger.info("Registered live instance change listeners for Helix manager at {}", zkConnectStr);
+            if (!clusterMapConfig.clustermapListenCrossColo && manager != localManager) {
+              manager.disconnect();
```
lightningrob: Wouldn't it be more efficient to move the property check up and skip the connection if we don't need it?

pnarayanan: We need to connect and get the cluster data from cross-colo ZK servers for initialization. We then close the connection, to stop listening for changes.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -699,6 +708,17 @@ public void xidTest() throws Exception {
...
+  public void listenCrossColoTest() throws Exception {
+    assumeTrue(!useComposite && listenCrossColo);
+    Counter instanceTriggerCounter =
+        ((HelixClusterManager) clusterManager).helixClusterManagerMetrics.instanceConfigChangeTriggerCount;
+    long instanceConfigChangeTriggerCount = instanceTriggerCounter.getCount();
+    helixCluster.triggerInstanceConfigChangeNotification();
+    assertEquals("Number of trigger count should be in accordance to listenCrossColo value",
+        instanceConfigChangeTriggerCount + (listenCrossColo ? dcs.length : 1), instanceTriggerCounter.getCount());
+  }
+
```
jsjtzyy: Also I think we may need to verify more things in this test. For example, when `listenCrossColo = false`, we need to verify that initialization still succeeded and the entire clustermap gets fully populated. After instantiation of `HelixClusterManager`, the connection in other remote DCs should be disconnected. And then, we can mock instanceConfig in remote DC to verify that local DC won't be affected (nothing happened locally)

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -699,6 +708,17 @@ public void xidTest() throws Exception {
     assertTrue(ignoreInstanceReplica.isDown());
   }
 
+  @Test
```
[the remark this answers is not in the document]
pnarayanan: Done.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -699,6 +708,17 @@ public void xidTest() throws Exception {
...
+  public void listenCrossColoTest() throws Exception {
+    assumeTrue(!useComposite && listenCrossColo);
```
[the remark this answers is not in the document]
pnarayanan: Thanks for catching this. This was tested, but this was added mistakenly.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
pnarayanan: Valid points, will make some changes.

pnarayanan: Made changes to verify that the Helix manager is connected to remote if and only if `listenCrossColo` is true. Also ensuring that triggering InstanceConfig change to a remote only affects the helix Cluster Manager if `listenCrossColo` is true.

pnarayanan:
@lightningrob @jsjtzyy addressed comments.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -699,6 +711,53 @@ public void xidTest() throws Exception {
...
+    Counter instanceTriggerCounter = helixClusterManager.helixClusterManagerMetrics.instanceConfigChangeTriggerCount;
+    Map<String, HelixManager> helixManagerMap = helixClusterManager.getHelixManagerMap();
+    for (Map.Entry<String, HelixManager> entry : helixManagerMap.entrySet()) {
+      if (entry.getKey().equals(localDc)) {
+        assertTrue("Helix cluster manager should always be connected to the local Helix manager",
+            entry.getValue().isConnected());
+      } else {
+        assertEquals(
+            "Helix cluster manager should be connected to the remote Helix managers if and only if listenCrossColo is"
+                + "set to false", listenCrossColo, entry.getValue().isConnected());
```
jsjtzyy: minor: should be `if and only if listenCrossColo is set to true`

pnarayanan: Addressed.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -699,6 +711,53 @@ public void xidTest() throws Exception {
...
+        .stream()
+        .filter(e -> writablePartitions.contains(e.getPartitionId()))
+        .findAny()
+        .get()
+        .getPartitionId();
+    remoteInstanceConfig.getRecord()
+        .setListField(SEALED_STR, Collections.singletonList(partitionIdToSealInRemote.toPathString()));
+    helixCluster.triggerInstanceConfigChangeNotification();
+    assertEquals("If replica in remote is sealed, partition should be sealed if and only if listenCrossColo is true "
+            + "and override is disabled", !listenCrossColo || overrideEnabled,
```
jsjtzyy: If this test is only to verify listen cross colos, maybe we don't need to include `overrideEnabled` in this test, because we don't know the override states. In this test, if `overrideEnabled == true` and the override state is `ReadOnly`, the assert still may fail.  This is my two cents. Will approve.

pnarayanan: What the test does is mark as `ReadOnly` a partition that is guaranteed to be `ReadWrite` even if `overrideEnabled == true` - by choosing a partition on the remote that is part of `getWritablePartitions()`.

PR #1085, opened 2018-10-10
Adding tests to verify replication with different versions of BlobId behaves correctly

jsjtzyy:
This PR added test to mock a scenario where replication with different versions of BlobId on local and remote hosts. The intention is to ensure that different combinations of BlobIds as well as messages are transformed and replicated as expected.

No code changes in production code, only for testing.

### DIFF
--- a/ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
+++ b/ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
@@ -117,6 +117,8 @@ public class ReplicationTest {
   private static int CONSTANT_TIME_MS = 100000;
   private static long EXPIRY_TIME_MS = SystemTime.getInstance().milliseconds() + TimeUnit.DAYS.toMillis(7);
   private static long UPDATED_EXPIRY_TIME_MS = SystemTime.getInstance().milliseconds() + TimeUnit.DAYS.toMillis(14);
+  private static final short VERSION_2 = 2;
+  private static final short VERSION_5 = 5;
   private final MockTime time = new MockTime();
   private ReplicationConfig config;
 
@@ -680,19 +682,13 @@ public class ReplicationTest {
    * fixMissingStoreKeys() should succeed without exceptions. The blob should not be put locally.
    */
   @Test
-  public void testDeletionAfterMetadataExchange() throws Exception {
-    MockClusterMap clusterMap = new MockClusterMap();
-    Pair<Host, Host> localAndRemoteHosts = getLocalAndRemoteHosts(clusterMap);
-    Host localHost = localAndRemoteHosts.getFirst();
-    Host remoteHost = localAndRemoteHosts.getSecond();
-    MockStoreKeyConverterFactory storeKeyConverterFactory = new MockStoreKeyConverterFactory(null, null);
-    storeKeyConverterFactory.setConversionMap(new HashMap<>());
-    storeKeyConverterFactory.setReturnInputIfAbsent(true);
-    MockStoreKeyConverterFactory.MockStoreKeyConverter storeKeyConverter =
-        storeKeyConverterFactory.getStoreKeyConverter();
-
+  public void deletionAfterMetadataExchangeTest() throws Exception {
+    int batchSize = 400;
+    ReplicationTestSetup testSetup = new ReplicationTestSetup(batchSize);
     short blobIdVersion = CommonTestUtils.getCurrentBlobIdVersion();
-    List<PartitionId> partitionIds = clusterMap.getWritablePartitionIds(null);
+    List<PartitionId> partitionIds = testSetup.partitionIds;
+    Host remoteHost = testSetup.remoteHost;
+    Host localHost = testSetup.localHost;
     Map<PartitionId, Set<StoreKey>> idsToExpectByPartition = new HashMap<>();
     for (int i = 0; i < partitionIds.size(); i++) {
       PartitionId partitionId = partitionIds.get(i);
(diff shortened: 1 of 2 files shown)


ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
```diff
@@ -885,6 +885,141 @@ public void testExpiryAfterMetadataExchange() throws Exception {
             .collect(Collectors.toSet()));
   }
 
+  /**
+   * Test the case where 1.remote host has different versions of blobIds and local host is empty; 2. remote and local hosts
```
dharju: put 1. and 2. on different lines?

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
```diff
@@ -2039,12 +2195,14 @@ public long sizeInBytes() {
     private List<ByteBuffer> buffersToReturn;
     private Map<PartitionId, List<MessageInfo>> infosToReturn;
     private Map<PartitionId, List<MessageMetadata>> messageMetadatasToReturn;
+    private Map<StoreKey, StoreKey> conversionMap;
     private ReplicaMetadataRequest metadataRequest;
     private GetRequest getRequest;
 
-    MockConnection(Host host, int maxSizeToReturn) {
+    MockConnection(Host host, int maxSizeToReturn, Map<StoreKey, StoreKey> conversionMap) {
```
dharju: Instead of having one ctor whee most of the time you fill the last argument with `null`, create another ctor `MockConnection(Host host, int maxSizeToReturn)` that calls `this(host, maxSizeToReturn, null)`, and then replace all the calls for `new MockConnection(host, int, null)` with `new MockConnection(host, int)`

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
```diff
@@ -885,6 +885,141 @@ public void testExpiryAfterMetadataExchange() throws Exception {
...
+    MockStoreKeyConverterFactory storeKeyConverterFactory = new MockStoreKeyConverterFactory(null, null);
+    storeKeyConverterFactory.setConversionMap(new HashMap<>());
+    storeKeyConverterFactory.setReturnInputIfAbsent(true);
+    MockStoreKeyConverterFactory.MockStoreKeyConverter storeKeyConverter =
+        storeKeyConverterFactory.getStoreKeyConverter();
+
+    PartitionId partitionId = clusterMap.getWritablePartitionIds(null).get(0);
+    // Only Version_2 and Version_5 blobIds are used in this tests.
+    // For simplicity, we use A1 to denote Version_2 blobId, A2 to denote Version_5 blobId.
+    // For example, A1P means Version_2 PUT message, A2D means Version_5 DELETE message, etc.
```
dharju: Instead of A1 and A2 use O (for old) and N (for new)

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
```diff
@@ -885,6 +885,141 @@ public void testExpiryAfterMetadataExchange() throws Exception {
...
+      // Verify
+      String resultStr = caseAndResult.getValue();
+      String[] expectedResults = resultStr.equals("") ? new String[0] : resultStr.split(" ");
+      int size = localHost.infosByPartition.get(partitionId).size();
+      assertEquals("Mismatch in number of messages on local host after replication", expectedResults.length, size);
+      for (int i = 0; i < size; ++i) {
+        switch (expectedResults[i]) {
+          case "A1P":
+            assertEquals("Mismatch in blodId on local host after replication", blobIdV2.toString(),
+                localHost.infosByPartition.get(partitionId).get(i).getStoreKey().toString());
```
dharju: `localHost.infosByPartition.get(partitionId).get(i).getStoreKey().toString()` and `localHost.infosByPartition.get(partitionId).get(i).isDeleted()` are used through out these cases, can you assign them to variables at the beginning of the `for` loop?

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
```diff
@@ -885,6 +885,141 @@ public void testExpiryAfterMetadataExchange() throws Exception {
...
+            assertEquals("Mismatch in blodId on local host after replication", blobIdV5.toString(),
+                localHost.infosByPartition.get(partitionId).get(i).getStoreKey().toString());
+            assertTrue("Mismatch in message type on local host after replication",
+                localHost.infosByPartition.get(partitionId).get(i).isDeleted());
+            break;
+        }
+      }
+
+      // Clean up partitions on both local and remote
+      localHost.clearAll();
```
dharju: Instead of having to clean up after every test case, might be cleaner to structure this as multiple tests with a `setup(...); testCase = ...; testAndVerify(testCase);` sort of set up, with each test having the same three line setup but with a different testCase.  This is also preferable as we can see if multiple test cases break (as opposed to failing out if one case fails) and it's easier to see which test case failed.

Generally having lots of small tests is preferable to having a few big tests that each test a lot of things.

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
jsjtzyy: done

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
jsjtzyy: Good point. Fixed.

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
jsjtzyy: Right, O(old) and N(new) may be more appropriate for open source code. Fixed.

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
jsjtzyy: Changed. Thanks for suggestion.

jsjtzyy:
@dharju reminder to review. Addressed all the comments and refactored the test methods.

dharju:
Cool, will look at when I can pull out my laptop (in plane right now)

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

PR #1090, opened 2018-10-18
Granular (colo-level) update support in the Helix tool

pnarayanan:
Add a "dcs" argument that takes a comma-separated list of dcs that
need to be affected in a given run. This provides an option to update
selectively and limit the effect of a bad change (when used along with
the listenCrossColo config within the HelixClusterManager).

### DIFF
--- a/ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
+++ b/ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
@@ -163,6 +163,7 @@ public class HelixBootstrapUpgradeToolTest {
    */
   @Test
   public void testIncompleteZKHostInfo() throws Exception {
+    assumeTrue(dcStr.equalsIgnoreCase("all"));
     if (testHardwareLayout.getDatacenterCount() > 1) {
       JSONObject partialZkJson =
           constructZkLayoutJSON(Collections.singleton(dcsToZkInfo.entrySet().iterator().next().getValue()));
--- a/ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
+++ b/ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
@@ -37,6 +37,7 @@ import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.store.HelixPropertyStore;
 import org.json.JSONException;
 import org.json.JSONObject;
+import org.junit.After;
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.BeforeClass;
@@ -56,7 +57,6 @@ import static org.junit.Assume.*;
 public class HelixBootstrapUpgradeToolTest {
   private static String tempDirPath;
   private static final Map<String, ZkInfo> dcsToZkInfo = new HashMap<>();
-  private static final Map<String, HelixPropertyStore<ZNRecord>> dcsToPropertyStore = new HashMap<>();
   private static final String dcs[] = new String[]{"DC0", "DC1"};
   private static final byte ids[] = new byte[]{(byte) 0, (byte) 1};
   private final String hardwareLayoutPath;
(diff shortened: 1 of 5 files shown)


pnarayanan:
@vgkholla @jsjtzyy addressed comments.

ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
```diff
@@ -87,24 +94,68 @@ public static void initialize() throws IOException {
     }
   }
 
+  @Parameterized.Parameters
+  public static List<Object[]> data() {
+    return Arrays.asList(new Object[][]{{"DC1"}, {"DC1, DC0"}, {"all"}});
```
jsjtzyy: I feel like test state has been persisted in cluster. That is, if we change the order, like move `{"DC1"}` to the end. `testEverything` will fail. Can you fix this?

ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
```diff
@@ -293,25 +345,29 @@ private void uploadClusterConfigsAndVerify() throws Exception {
...
       String getPath = ClusterMapUtils.PROPERTYSTORE_ZNODE_PATH;
       ZNRecord zNRecord = propertyStore.get(getPath, null, AccessOption.PERSISTENT);
-      assertNotNull(zNRecord);
-      Map<String, Map<String, String>> overridePartition = zNRecord.getMapFields();
-      Set<String> writableInDC = new HashSet<>();
-      for (Map.Entry<String, Map<String, String>> entry : overridePartition.entrySet()) {
-        if (entry.getValue().get(ClusterMapUtils.PARTITION_STATE).equals(ClusterMapUtils.READ_WRITE_STR)) {
-          writableInDC.add(entry.getKey());
+      if (!activeDcSet.contains(zkInfo.getDcName())) {
+        assertNull(zNRecord);
```
jsjtzyy: In my humble opinion, the testing here need to be improved, it's better to make record have previous state and compare zNRecord from not specified DC to this previous state. An alternative way is to deploy to all DCs at the end of this method and get the previous record at the very beginning of the method.

pnarayanan: You are right in that there are many ways this can be improved. 
Here we are checking that the node never gets created if that dc is not part of the active set. I guess your suggestion is to see that if it is already created, then it does not get updated. Though, I do feel that it is an overkill, since it does not exercise any different code - basically, the constructor only initializes connections to active dcs. The same code is getting exercised in either case. Note that we are not interested in checking if other dcs get affected if admin connections are not made to them.

Let me know what you think, we can have an offline conversation if you wish.

ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
```diff
@@ -87,24 +94,76 @@ public static void initialize() throws IOException {
...
+        HelixBootstrapUpgradeUtil.parseAndUpdateDcInfoFromArg("DC0, DC1", zkLayoutPath).keySet());
+    Assert.assertEquals(expected, HelixBootstrapUpgradeUtil.parseAndUpdateDcInfoFromArg("all", zkLayoutPath).keySet());
   }
 
   /**
    * Test the case where the zkHosts JSON does not have an entry for every Datacenter in the static clustermap.
    */
   @Test
   public void testIncompleteZKHostInfo() throws Exception {
+    assumeTrue(dcStr.equalsIgnoreCase("all") || dcStr.replaceAll("\\p{Space}", "").split(",").length == dcs.length);
```
jsjtzyy: I think `assumeTrue` is unnecessary. Such test is supposed to be performed when all DCs are specified.

pnarayanan:
@jsjtzyy addressed comments.

ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
pnarayanan: The test is being performed when ALL (all dcs) is specified. `assumeTrue()` basically ensures that it is only run when this condition is true.
Basically, the test does not completely make sense when the input is not ALL. The static clustermap can contain more datacenters than what the user intends to update when input is not ALL, so this test is being skipped in those cases.

PR #1095, opened 2018-10-26
Add config for selecting PRNG algorithm in JdkSslFactory

cgtz:
- Add a config to allow `JdkSslFactory` to use non-default SecureRandom algorithms.
- Clean up the constructor code/unneeded members in `JdkSslFactory`

This is inspired by this [kafka PR](https://github.com/apache/kafka/pull/1747). This allows ambry users to choose from the `SecureRandom` implementations described [here](https://docs.oracle.com/javase/8/docs/technotes/guides/security/StandardNames.html#SecureRandom). Some of these algorithms can deliver much better performance than the `/dev/random` based default, which has a habit of running out of entropy.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/SSLConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/SSLConfig.java
@@ -17,7 +17,10 @@ package com.github.ambry.config;
  * The configs for SSL
  */
 public class SSLConfig {
-  /**
+  private static final String PREFIX = "ssl.";
+  public static final String SECURE_RANDOM_ALGORITHM_KEY = PREFIX + "secure.random.algorithm";
+
+ /**
    * The SSL protocol for SSLContext
    */
   @Config("ssl.context.protocol")
@@ -44,6 +47,14 @@ public class SSLConfig {
   @Config("ssl.endpoint.identification.algorithm")
   @Default("")
   public final String sslEndpointIdentificationAlgorithm;
+
+  /**
+   * The {@code SecureRandom} PRNG algorithm to use for SSL cryptography operations. This is only honored by
+   * {@code JdkSslFactory}. {@code NettySslFactory} uses a native SSL impl and does not rely on {@code SecureRandom}.
+   */
+  @Config(SECURE_RANDOM_ALGORITHM_KEY)
+  @Default("")
+  public final String sslSecureRandomAlgorithm;
 
   /**
    * The SSL client authentication config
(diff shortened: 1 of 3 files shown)


ambry-api/src/main/java/com.github.ambry/config/SSLConfig.java
```diff
@@ -45,6 +48,14 @@
   @Default("")
   public final String sslEndpointIdentificationAlgorithm;
 
+  /**
+   * The {@code SecureRandom} PRNG algorithm to use for SSL cryptography operations. This is only honored by
+   * {@code JdkSslFactory}. {@code NettySslFactory} uses a native SSL impl and does not rely on {@code SecureRandom}.
+   */
+  @Config(SECURE_RANDOM_ALGORITHM_KEY)
```
zzmao: How about just use "ssl.secure.random.algorithm" to line with others? I feel  "ssl.secure.random.algorithm" has better readability.

ambry-commons/src/test/java/com.github.ambry.commons/JdkSslFactoryTest.java
```diff
@@ -13,8 +13,18 @@
...
 package com.github.ambry.commons;
 
+import com.github.ambry.config.SSLConfig;
+import com.github.ambry.config.VerifiableProperties;
+import com.github.ambry.utils.TestUtils;
+import java.io.File;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.Properties;
+import org.junit.Assert;
```
zzmao: Not used.

ambry-api/src/main/java/com.github.ambry/config/SSLConfig.java
cgtz: I can just use the raw string in this case for consistency. In general, I prefer using constants since they make it easier to prevent typos in the constructor and test code, but I can try do that as a future PR for all of the configs in this class.

ambry-commons/src/test/java/com.github.ambry.commons/JdkSslFactoryTest.java
cgtz: removed

ambry-api/src/main/java/com.github.ambry/config/SSLConfig.java
zzmao: I see. You need to use it in test. LGTM.

ambry-commons/src/test/java/com.github.ambry.commons/JdkSslFactoryTest.java
```diff
@@ -28,5 +38,27 @@
   @Test
   public void testSSLFactory() throws Exception {
     TestSSLUtils.testSSLFactoryImpl(JdkSslFactory.class.getName());
+
+    // test features specific to JDK impls, like the PRNG algorithm config
+    File trustStoreFile = File.createTempFile("truststore", ".jks");
+    Properties props = new Properties();
+    TestSSLUtils.addSSLProperties(props, "DC1,DC2,DC3", SSLFactory.Mode.CLIENT, trustStoreFile, "client");
+    for (String prngAlgorithm : new String[]{"NativePRNGNonBlocking", "SHA1PRNG", "Windows-PRNG", "badbadinvalid"}) {
```
zzmao: Also test `""`?

cgtz: added test

Issue #1069, opened 2018-09-25
[Helix cluster manager] Nodes should initialize even if there are ZK Connectivity problems


pnarayanan:
For the scope of this issue, we need to ensure that nodes (frontends and servers) initialize just fine as long as there is connectivity to the local ZooKeeper cluster. This is particularly important during deployments.

Local ZK connectivity should be sufficient to support most operations (slight degradation with cross-colo should be okay).

PR #1103, opened 2018-11-14
Make cluster manager initialization more resilient

pnarayanan:
Ensure that Helix cluster manager initializes and runs (albeit in a
degraded state) even if there are connectivity issues with remote
ZK clusters. Local colo connectivity is necessary for basic operations.

The degradation involves the following:
- No cross-colo requests from frontends and cross-colo replication lag at
  servers, leading to absence of read-after-write guarantees.
- Absence of discovering sealed state of partitions if something changes
  in remote colo (this can happen even when connectivity is lost after
  initialization today).

However, this is a small price to pay given that nodes can continue to start
up, intialize and be available for requests, and process most requests (all
PUT requests, all GET requests for older blobs and all GET requests for recent
blobs in the local colo.

Fixes linkedin/ambry#1069


ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -155,87 +166,94 @@
...
-      close();
-      throw new IOException("Encountered exception while parsing json, connecting to Helix or initializing",
-          initializationException.get());
     }
     localDatacenterId = dcToDcZkInfo.get(clusterMapConfig.clusterMapDatacenterName).dcZkInfo.getDcId();
     partitionSelectionHelper =
         new PartitionSelectionHelper(partitionMap.values(), clusterMapConfig.clusterMapDatacenterName);
   }
 
+  private void setException(boolean isLocalDc, Exception e) {
```
jsjtzyy: doc this method please.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -155,87 +166,94 @@
...
+    if (isLocalDc) {
+      initializationException.compareAndSet(null, e);
+    } else {
+      remoteInitializationException.compareAndSet(null, e);
+      remoteInitializationFailedCount.incrementAndGet();
```
jsjtzyy: I am thinking in a very rare scenario, for a single remote datacenter, the initialization failure could be counted twice. The first time in `ClusterChangeHandler`, the second time may be in the initialization thread after `ClusterChangeHandler` is added as listener.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -155,87 +166,94 @@
...
+    ZkHelixPropertyStore<ZNRecord> helixPropertyStore;
+    manager = helixFactory.getZKHelixManager(clusterName, instanceName, InstanceType.SPECTATOR, zkConnectStr);
+    logger.info("Connecting to Helix manager in local zookeeper at {}", zkConnectStr);
+    manager.connect();
+    logger.info("Established connection to Helix manager in local zookeeper at {}", zkConnectStr);
+    helixPropertyStore = manager.getHelixPropertyStore();
+    logger.info("HelixPropertyStore from local datacenter {} is: {}", dcZkInfo.getDcName(), helixPropertyStore);
+    IZkDataListener dataListener = new IZkDataListener() {
+      @Override
+      public void handleDataChange(String dataPath, Object data) {
```
jsjtzyy: I think we'd better keep the original method signature. That is keep the `throws Exception`  here. I agree with other code changes in this method.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -155,87 +166,94 @@
...
-      ZNRecord zNRecord = helixPropertyStore.get(ClusterMapUtils.ZNODE_PATH, null, AccessOption.PERSISTENT);
-      if (clusterMapConfig.clusterMapEnablePartitionOverride) {
-        if (zNRecord != null) {
-          partitionOverrideInfoMap.putAll(zNRecord.getMapFields());
-          logger.info("partitionOverrideInfoMap is initialized!");
-        } else {
-          logger.warn("ZNRecord from HelixPropertyStore is NULL, the partitionOverrideInfoMap is empty.");
-        }
+      @Override
+      public void handleDataDeleted(String dataPath) {
```
jsjtzyy: same here.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
pnarayanan: True. Probably not a big deal, but I have changed this to use a hash map to make it more accurate.

pnarayanan:
@jsjtzyy Addressed comments.

pnarayanan:
@jsjtzyy Addressed your comment.

PR #1104, opened 2018-11-21
Merge recent kafka changes into SSLTransmission

cgtz:
SSLTransmission was originally based on Kafka's SslTransportLayer class. However,
it is missing a couple years worth of fixes and performance improvements. This
PR brings a couple of fixes:
- Read until the socket has no more data instead of requiring an additional
  selector loop (KAFKA-6258)
- Fixes for buffer overflow handling (KAFKA-2534)

For reference, here is the class that I diff-ed against: https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/network/SslTransportLayer.java

For more information about each patch, see:
- [KAFKA-6258](https://issues.apache.org/jira/browse/KAFKA-6258)
- [KAFKA-2534](https://issues.apache.org/jira/browse/KAFKA-2534)


ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -399,40 +411,44 @@ public boolean read() throws IOException {
...
+      netReadBuffer = Utils.ensureCapacity(netReadBuffer, netReadBufferSize());
       if (netReadBuffer.remaining() > 0) {
-        int netread = socketChannel.read(netReadBuffer);
-        if (netread == 0) {
-          return read;
-        } else if (netread < 0) {
-          throw new EOFException("EOF during read");
-        }
+        netread = readFromSocketChannel();
+        // TODO look at this
```
zzmao: Looks safe to remove.

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -444,7 +460,7 @@ public int read(ByteBuffer dst) throws IOException {
         }
```
zzmao: Move the log and metric outside of the while loop? `while (dst.remaining() > 0) `

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -201,16 +211,19 @@ private void handshake() throws IOException {
...
           handshakeResult = handshakeWrap(write);
-          if (handshakeResult.getStatus() == SSLEngineResult.Status.BUFFER_OVERFLOW) {
-            int currentPacketBufferSize = packetBufferSize();
-            netWriteBuffer = Utils.ensureCapacity(netWriteBuffer, currentPacketBufferSize);
-            if (netWriteBuffer.position() >= currentPacketBufferSize) {
-              throw new IllegalStateException("Buffer overflow when available data size (" + netWriteBuffer.position()
-                  + ") >= network buffer size (" + currentPacketBufferSize + ")");
+          if (handshakeResult.getStatus() == Status.BUFFER_OVERFLOW) {
+            int currentNetWriteBufferSize = netWriteBufferSize();
+            netWriteBuffer.compact();
```
zzmao: Not sure how this work. Will talk offline.

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
cgtz: The metric in `read()` already covers this case.

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
cgtz: I think the bug before happened when some bytes were written/read from the socket but not all.

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -31,6 +31,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static javax.net.ssl.SSLEngineResult.*;
+
 
 /**
  * Handles all the SSL related interactions. It is mainly responsible for establishing the handshake completely,
```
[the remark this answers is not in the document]
cgtz: Yep, our test coverage is lacking. I will work on getting some of the new unit test cases over from kafka

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -219,30 +232,32 @@ private void handshake() throws IOException {
...
-            netReadBuffer = Utils.ensureCapacity(netReadBuffer, currentPacketBufferSize);
-            if (netReadBuffer.position() >= currentPacketBufferSize) {
-              throw new IllegalStateException("Buffer underflow when there is available data");
+          do {
+            handshakeResult = handshakeUnwrap(read);
+            if (handshakeResult.getStatus() == Status.BUFFER_OVERFLOW) {
+              int currentAppBufferSize = applicationBufferSize();
+              appReadBuffer = Utils.ensureCapacity(appReadBuffer, currentAppBufferSize);
+              if (appReadBuffer.position() > currentAppBufferSize) {
+                throw new IllegalStateException("Buffer underflow when available data size (" + appReadBuffer.position()
```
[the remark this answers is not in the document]
cgtz: Fixed the log message and unified some of the duplicate code.

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -201,16 +211,19 @@ private void handshake() throws IOException {
...
-            netWriteBuffer = Utils.ensureCapacity(netWriteBuffer, currentPacketBufferSize);
-            if (netWriteBuffer.position() >= currentPacketBufferSize) {
-              throw new IllegalStateException("Buffer overflow when available data size (" + netWriteBuffer.position()
-                  + ") >= network buffer size (" + currentPacketBufferSize + ")");
+          if (handshakeResult.getStatus() == Status.BUFFER_OVERFLOW) {
+            int currentNetWriteBufferSize = netWriteBufferSize();
+            netWriteBuffer.compact();
+            netWriteBuffer = Utils.ensureCapacity(netWriteBuffer, currentNetWriteBufferSize);
+            netWriteBuffer.flip();
+            if (netWriteBuffer.limit() >= currentNetWriteBufferSize) {
```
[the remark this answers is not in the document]
cgtz: I actually don't know the purpose of these checks (our code and kafka code had the same checks before (https://github.com/apache/kafka/pull/205/files#diff-fce430ae21a0c98d82da6d4aa551f824)). These paths also aren't exercised in tests, so they seem some like some sort of assertion for a case that should never happen. Let me think about whether we should have these at all.

cgtz:
@dharju @zzmao, sorry for the long delay. Comments are addressed now. To get some of the code paths covered, I had to make some methods package private as was done in https://github.com/apache/kafka/pull/205

ambry-network/src/test/java/com.github.ambry.network/SSLSelectorTest.java
```diff
@@ -206,33 +218,160 @@ public void testSSLConnect() throws IOException {
...
+
+  /**
+   * Tests handling of BUFFER_OVERFLOW during wrap when network write buffer is smaller than SSL session packet buffer
+   * size.
+   */
+  @Test
+  public void testNetWriteBufferResize() throws Exception {
+    useCustomBufferSizeSelector(null, 10, null);
+    String connectionId = blockingSSLConnect(DEFAULT_SOCKET_BUF_SIZE);
+    String message = SelectorTest.randomString(10, new Random());
```
zzmao: I though `BUFFER_OEVERFLOW` is in `SSLTransmission::write(ByteBuffer)`, but it turns out no. In step debugging, I found `BUFFER_OVERFLOW` happened in `SSLTransmission::handshake()` only and `netWriteBuffer` is expanded to `16KB` after handshake. 

I also changed message size to 4*16KB to make it > `netWriteBuffer `, but `write(ByteBuffer)->warp()` still returns OK.  

No change needed, just FYI.

ambry-api/src/main/java/com.github.ambry/network/BoundedByteBufferReceive.java
```diff
@@ -55,7 +55,6 @@ public long readFrom(ReadableByteChannel channel) throws IOException {
         sizeBuffer.flip();
         sizeToRead = sizeBuffer.getLong();
         sizeRead += 8;
-        bytesRead += 8;
```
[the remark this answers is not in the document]
cgtz: This was actually a bug (although one that only affected metrics). Line 50 already accounts for reading the size field, so the removed line actually made the returned size 8 bytes larger than it should have been.

ambry-network/src/test/java/com.github.ambry.network/SSLSelectorTest.java
```diff
@@ -206,33 +218,160 @@ public void testSSLConnect() throws IOException {
...
+            return getAndUpdateSize(super.netWriteBufferSize(), netWriteBufSizeOverride, true);
+          }
+
+          @Override
+          protected int appReadBufferSize() {
+            return getAndUpdateSize(super.appReadBufferSize(), appReadBufSizeOverride, true);
+          }
+
+          /**
+           * Helper method for incrementing the
```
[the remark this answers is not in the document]
cgtz: Completed

ambry-network/src/test/java/com.github.ambry.network/SSLSelectorTest.java
cgtz: Yep, unfortunately this test only tests the buffer expansion flow within the handshake code. If I think of a good way to test it within the write call, I'll publish a follow up PR.

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -201,16 +211,11 @@ private void handshake() throws IOException {
...
-          if (handshakeResult.getStatus() == SSLEngineResult.Status.BUFFER_OVERFLOW) {
-            int currentPacketBufferSize = packetBufferSize();
-            netWriteBuffer = Utils.ensureCapacity(netWriteBuffer, currentPacketBufferSize);
-            if (netWriteBuffer.position() >= currentPacketBufferSize) {
-              throw new IllegalStateException("Buffer overflow when available data size (" + netWriteBuffer.position()
-                  + ") >= network buffer size (" + currentPacketBufferSize + ")");
-            }
-          } else if (handshakeResult.getStatus() == SSLEngineResult.Status.BUFFER_UNDERFLOW) {
+          if (handshakeResult.getStatus() == Status.BUFFER_OVERFLOW) {
+            handleWrapOverflow();
```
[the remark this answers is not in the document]
cgtz: Yes, this seems to work and probably eliminates an unneeded loop to call handshakeUnwrap again.

ambry-api/src/main/java/com.github.ambry/network/BoundedByteBufferReceive.java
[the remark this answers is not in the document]
zzmao: good catch!

ambry-api/src/main/java/com.github.ambry/network/BoundedByteBufferReceive.java
```diff
@@ -55,7 +55,6 @@ public long readFrom(ReadableByteChannel channel) throws IOException {
         sizeBuffer.flip();
```
zzmao: Minor: Should we throw some exception if `sizeBuffer.position() != sizeBuffer.capacity()`?

cgtz: Discussed this offline. The way this code is written is actually correct since it accounts for the case where the network channel does not have 8 bytes of data ready yet (but might in a subsequent call to `readFrom`)

PR #1105, opened 2018-11-22
Warp data as many as possible in SSLTranmission#write(ByteBuffer)

zzmao:
Write to the socket until no more space instead of requiring an additional selector loop. 
They can give us performance gain when PUT blobs.


cgtz:
minor typo with commit message: warp instead of wrap.

We can merge this after #1104 is merged

PR #1111, opened 2018-12-07
Address review comments on #1109

lightningrob:
Increase zookeeper timeout to make AccountUpdateToolTest pass


lightningrob:
Ran "gradle build test" and it passed with flying colors.

PR #1116, opened 2018-12-17
Make Bootstrap tool use Helix Cluster Manager for validation

pnarayanan:
Instantiate a Helix cluster manager within the bootstrap tool to listen
on the changes that it makes itself. This cluster manager's state will
be checked during validation. If there are other errors that can cause
errors such as Out of Memory in the Helix client, then those can be
immediately detected because the tool itself will be affected.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
@@ -155,7 +155,7 @@ class HelixBootstrapUpgradeUtil {
    * @param dryRun if true, perform a dry run; do not update anything in Helix.
    * @param forceRemove if true, removes any hosts from Helix not present in the json files.
    * @param helixAdminFactory the {@link HelixAdminFactory} to use to instantiate {@link HelixAdmin}
-   * @param startValidatingClusterManager whether validation should include staring up a {@link HelixClusterManager}
+   * @param startValidatingClusterManager whether validation should include starting up a {@link HelixClusterManager}
    * @throws IOException if there is an error reading a file.
    * @throws JSONException if there is an error parsing the JSON content in any of the files.
    */
@@ -698,7 +698,7 @@ class HelixBootstrapUpgradeUtil {
   /**
    * Initialize a map of dataCenter to HelixAdmin based on the given zk Connect Strings.
    */
-  private void maybeAddCluster() throws IOException {
+  private void maybeAddCluster() {
     for (Map.Entry<String, HelixAdmin> entry : adminForDc.entrySet()) {
       // Add a cluster entry in every DC
       String dcName = entry.getKey();
@@ -711,6 +711,10 @@ class HelixBootstrapUpgradeUtil {
     }
   }
 
+  /**
+   * Start the Helix Cluster Manager to be used for validation.
+   * @throws IOException if there was an error instantiating the cluster manager.
+   */
   private void startClusterManager() throws IOException {
     Properties props = new Properties();
     props.setProperty("clustermap.host.name", "localhost");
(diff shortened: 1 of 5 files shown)


pnarayanan:
@jsjtzyy fixed test failures and rebased with master. Good to review when you get a chance. Thanks.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -702,6 +711,17 @@ private void maybeAddCluster() {
...
   }
 
+  private void startClusterManager() throws IOException {
+    Properties props = new Properties();
+    props.setProperty("clustermap.host.name", "localhost");
+    props.setProperty("clustermap.cluster.name", clusterName);
+    props.setProperty("clustermap.datacenter.name", adminForDc.keySet().iterator().next());
+    props.setProperty("clustermap.dcs.zk.connect.strings", Utils.readStringFromFile(zkLayoutPath));
+    ClusterMapConfig clusterMapConfig = new ClusterMapConfig(new VerifiableProperties(props));
+    HelixClusterAgentsFactory helixClusterAgentsFactory = new HelixClusterAgentsFactory(clusterMapConfig, null, null);
```
jsjtzyy: just to confirm, the validating HelixClusterManager is going to talk to one zookeeper in one of the specified datacenters? Here "specified" means the selected DCs which will be updated/affected by Helix tool. And do you think it's worth setting up a second validating HelixClusterManager to simulate those in untouched DCs?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -719,7 +739,14 @@ private void validateAndClose() throws Exception {
     try {
```
jsjtzyy: no Exception is thrown in this method

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -433,21 +441,26 @@ AmbryReplica getReplicaForPartitionOnNode(String hostname, int port, String part
...
-        } else {
-          updateStateOfReplicas(configs);
+          sealedStateChangeCounter.incrementAndGet();
+          helixClusterManagerMetrics.instanceConfigChangeTriggerCount.inc();
         }
-        sealedStateChangeCounter.incrementAndGet();
-        helixClusterManagerMetrics.instanceConfigChangeTriggerCount.inc();
+      } catch (Throwable t) {
+        errorCount.incrementAndGet();
+        throw t;
```
jsjtzyy: In actual node (not the validating one), how the `Throwable` would be handled?

ambry-tools/src/integration-test/java/com.github.ambry/clustermap/HelixBootstrapUpgradeToolTest.java
```diff
@@ -172,7 +177,8 @@ public void testIncompleteZKHostInfo() throws Exception {
       Utils.writeJsonObjectToFile(testPartitionLayout.getPartitionLayout().toJSONObject(), partitionLayoutPath);
       try {
         HelixBootstrapUpgradeUtil.bootstrapOrUpgrade(hardwareLayoutPath, partitionLayoutPath, zkLayoutPath,
-            CLUSTER_NAME_PREFIX, dcStr, DEFAULT_MAX_PARTITIONS_PER_RESOURCE, false, false, new HelixAdminFactory());
+            CLUSTER_NAME_PREFIX, dcStr, DEFAULT_MAX_PARTITIONS_PER_RESOURCE, false, false, new HelixAdminFactory(),
+            true);
```
jsjtzyy: I think, maybe, we should also check the error count from validating HelixClusterManager. From current tests, I don't see the verification on functionality of validating HelixClusterManager.

pnarayanan: `HelixBootstrapUpgradeUtil#validateAndClose()` checks for it, are you suggesting something else?

jsjtzyy: I was suggesting getting the error count from `validatingHelixClusterManager` after running the tool. Looks like the validating piece of code in `validateAndClose()` (see follow) is not covered by this test. 
```
if (validatingHelixClusterManager != null) {
        ensureOrThrow(validatingHelixClusterManager.getErrorCount() == 0,
            "Helix cluster manager should not have encountered any errors");
      }
```

PR #1117, opened 2018-12-20
Allow NettyServer to use its own ssl impl

cgtz:
Previously, the Netty code was forced to use the same SSL impl as the
router. This adds a config to let NettyServer use a different SSLFactory
implementation. This can be used to leverage the performance benefits of
NettySslFactory within the HTTP server, without affecting the router,
which has some compatability issues with Netty's OpenSSL-based
SSLEngine.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
@@ -121,6 +121,15 @@ public class NettyConfig {
   @Default("20 * 1024 * 1024")
   public final long nettyMultipartPostMaxSizeBytes;
 
+  /**
+   * If set, use this implementation of {@link com.github.ambry.commons.SSLFactory} to use for the netty HTTP server.
+   * Otherwise, share the factory instance with the router.
+   */
+  public static final String SSL_FACTORY_KEY = "netty.server.ssl.factory";
+  @Config(SSL_FACTORY_KEY)
+  @Default("")
+  public final String nettyServerSslFactory;
+
   public NettyConfig(VerifiableProperties verifiableProperties) {
     nettyServerBossThreadCount = verifiableProperties.getInt("netty.server.boss.thread.count", 1);
     nettyServerIdleTimeSeconds = verifiableProperties.getInt("netty.server.idle.time.seconds", 60);
@@ -139,5 +148,6 @@ public class NettyConfig {
         Arrays.asList(verifiableProperties.getString("netty.server.blacklisted.query.params", "").split(",")));
     nettyMultipartPostMaxSizeBytes =
         verifiableProperties.getLongInRange("netty.multipart.post.max.size.bytes", 20 * 1024 * 1024, 0, Long.MAX_VALUE);
+    nettyServerSslFactory = verifiableProperties.getString(SSL_FACTORY_KEY, "");
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
@@ -22,6 +22,8 @@ import java.util.Set;
  * Configuration parameters required by Netty.
  */
 public class NettyConfig {
+  public static final String SSL_FACTORY_KEY = "netty.server.ssl.factory";
+
   /**
    * Number of netty boss threads.
    */
@@ -125,7 +127,6 @@ public class NettyConfig {
    * If set, use this implementation of {@link com.github.ambry.commons.SSLFactory} to use for the netty HTTP server.
    * Otherwise, share the factory instance with the router.
    */
-  public static final String SSL_FACTORY_KEY = "netty.server.ssl.factory";
   @Config(SSL_FACTORY_KEY)
   @Default("")
   public final String nettyServerSslFactory;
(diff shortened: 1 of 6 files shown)


ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
```diff
@@ -121,6 +121,15 @@
   @Default("20 * 1024 * 1024")
   public final long nettyMultipartPostMaxSizeBytes;
 
+  /**
+   * If set, use this implementation of {@link com.github.ambry.commons.SSLFactory} to use for the netty HTTP server.
+   * Otherwise, share the factory instance with the router.
+   */
+  public static final String SSL_FACTORY_KEY = "netty.server.ssl.factory";
```
jsjtzyy: move this string to the beginning of this class?

ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
```diff
@@ -121,6 +121,15 @@
   @Default("20 * 1024 * 1024")
   public final long nettyMultipartPostMaxSizeBytes;
 
+  /**
+   * If set, use this implementation of {@link com.github.ambry.commons.SSLFactory} to use for the netty HTTP server.
```
jsjtzyy: Is this Netty's own ssl implementation?

ambry-rest/src/main/java/com.github.ambry.rest/NettyServerFactory.java
```diff
@@ -47,17 +52,24 @@
    * @param sslFactory the {@link SSLFactory} used to construct the {@link javax.net.ssl.SSLEngine} used for handling
    *                   SSL requests.
    * @throws IllegalArgumentException if any of the arguments are null.
+   * @throws ReflectiveOperationException if a netty-specific {@link SSLFactory} cannot be instantiated via reflection.
    */
   public NettyServerFactory(VerifiableProperties verifiableProperties, MetricRegistry metricRegistry,
       final RestRequestHandler requestHandler, final PublicAccessLogger publicAccessLogger,
-      final RestServerState restServerState, SSLFactory sslFactory) {
+      final RestServerState restServerState, SSLFactory sslFactory) throws ReflectiveOperationException {
```
lightningrob: Since we're declaring that here, why not also change Utils.getObj() to throw ReflectiveOperationException instead of the five it currently declares?  I wasn't aware until now of this base class that was apparently added in JDK 1.7.

ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
lightningrob: +1.  It would be good to have property constants for the others too, but that can be done later.

ambry-rest/src/main/java/com.github.ambry.rest/NettyServerFactory.java
```diff
@@ -47,17 +52,24 @@
...
     nettyConfig = new NettyConfig(verifiableProperties);
-    if (sslFactory == null && nettyConfig.nettyServerEnableSSL) {
-      throw new IllegalArgumentException("NettyServer requires SSL, but sslFactory is null");
+    if (nettyConfig.nettyServerEnableSSL) {
+      if (!nettyConfig.nettyServerSslFactory.isEmpty()) {
+        LOGGER.info("Using " + nettyConfig.nettyServerSslFactory + " for Netty SSL instead of the shared instance.");
+        sslFactory = Utils.getObj(nettyConfig.nettyServerSslFactory, new SSLConfig(verifiableProperties));
+      }
+      if (sslFactory == null) {
+        throw new IllegalArgumentException("NettyServer requires SSL, but sslFactory is null");
```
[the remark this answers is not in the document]
cgtz: Sorry, perhaps I should have created a new variable instead of overwriting the constructor argument. The provided constructor argument is the "default" sslFactory that is shared with the router and is constructed at a higher level. If the "netty.server.ssl.factory" option is set, then line 68 will ensure that netty constructs and uses "its own" factory.

ambry-rest/src/main/java/com.github.ambry.rest/NettyServerFactory.java
cgtz: Changed the various `getObj` methods to throw this exception.

ambry-api/src/main/java/com.github.ambry/config/NettyConfig.java
cgtz: For now, I'll keep the others as is, but I'll look into constant-ifying this and other config classes in the future. Maybe there's a simple way to do it en-masse using intellij refactoring tools.

ambry-rest/src/main/java/com.github.ambry.rest/NettyServerFactory.java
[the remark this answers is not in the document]
lightningrob: ok thanks for explaining.  Maybe could be documented a bit clearer.

ambry-rest/src/main/java/com.github.ambry.rest/NettyServerFactory.java
lightningrob: Did you check that in?  Or will that be in a followup?

cgtz: Checked in now. Sorry, got interrupted in between writing these comments and git pushing

PR #1119, opened 2019-01-02
Introducing secure path validation in Container

jsjtzyy:
1. Introduced secure path validation as a new property in Container
2. Added secure path validation in security service if required by Container
3. Added method to support getting secure path from URI
4. Secure path is configurable in FrontendConfig

### DIFF
--- a/ambry-frontend/src/main/java/com.github.ambry.frontend/AmbrySecurityService.java
+++ b/ambry-frontend/src/main/java/com.github.ambry.frontend/AmbrySecurityService.java
@@ -379,6 +379,7 @@ class AmbrySecurityService implements SecurityService {
     if (targetContainer.isSecurePathRequired()) {
       String securePath = prefixToCheck.startsWith("/") ? prefixToCheck.substring(1) : prefixToCheck;
       if (!securePath.equals(expectSecurePath)) {
+        frontendMetrics.securePathValidationFailedCount.inc();
         throw new RestServiceException("Secure path in restRequest doesn't match the expected one",
             RestServiceErrorCode.AccessDenied);
       }
--- a/ambry-frontend/src/main/java/com.github.ambry.frontend/FrontendMetrics.java
+++ b/ambry-frontend/src/main/java/com.github.ambry.frontend/FrontendMetrics.java
@@ -148,6 +148,7 @@ public class FrontendMetrics {
   public final Histogram securityServiceProcessRequestTimeInMs;
   public final Histogram securityServicePostProcessRequestTimeInMs;
   public final Histogram securityServiceProcessResponseTimeInMs;
+  public final Counter securePathValidationFailedCount;
   // AmbryIdConverter
   public final Histogram idConverterProcessingTimeInMs;
   // GetPeersHandler
@@ -383,6 +384,8 @@ public class FrontendMetrics {
         metricRegistry.histogram(MetricRegistry.name(AmbrySecurityService.class, "RequestPostProcessingTimeInMs"));
     securityServiceProcessResponseTimeInMs =
         metricRegistry.histogram(MetricRegistry.name(AmbrySecurityService.class, "ResponseProcessingTimeInMs"));
+    securePathValidationFailedCount =
+        metricRegistry.counter(MetricRegistry.name(AmbrySecurityService.class, "SecurePathValidationFailedCount"));
     // AmbryIdConverter
     idConverterProcessingTimeInMs =
         metricRegistry.histogram(MetricRegistry.name(AmbryIdConverterFactory.class, "ProcessingTimeInMs"));


ambry-api/src/main/java/com.github.ambry/rest/RestUtils.java
```diff
@@ -562,6 +562,17 @@ public static String getOperationOrBlobIdFromUri(RestRequest restRequest, RestUt
       endIndex = endIndex - subResource.name().length() - 1;
     }
     String operationOrBlobId = path.substring(startIndex, endIndex);
+    // remove signed path if present and if it's not a signed id
```
jsjtzyy: This part is a little complicated, or maybe there is a better implementation. The code was written based following considerations:
1. Current URIs (i.e. get blob/ get BlobInfo/ signed id) should still work correctly without any errors;
2. When `Signed Path` is embedded in URI, this method can still extract BlobId correctly;
3. The logic in this method is, after removing query string/prefix/subResource, if there are more than 2 segments ( or 3 if leading slash exists), the first non-empty segment is signed path. 
4. This is a little controversial because it requires `getOperation` like GET_PEERS etc only occupies one segment. The URI like `/peers/random/extra` is not allowed any more, otherwise, the first segment will be considered signed path and be removed. 
5. For 4, I currently don't have a better way to address this. Let me know if you have good approach.

ambry-api/src/main/java/com.github.ambry/rest/RestUtils.java
```diff
@@ -589,6 +600,36 @@ public static String getOperationOrBlobIdFromUri(RestRequest restRequest, RestUt
...
+    }
+    if (path.startsWith("/")) {
+      // remove leading slash
+      path = path.substring(1);
+    }
+    String[] segments = path.split("/");
+    // extract signed path
+    // 1. if there are more than 1 segment   (i.e. "signed-path/blobId")
+    // 2. if there is 1 segment and uri ends with "/" (i.e. "signed-path/")
+    if (segments.length > 1 || (segments.length == 1 && path.endsWith("/"))) {
```
jsjtzyy: I don't know if it's safe to extract signed path when there is 1 segment and URI ends with "/", i.e. `signed-path/`. Please be strict on this if condition and correct me if any potential counter examples making the code vulnerable.

ambry-api/src/test/java/com.github.ambry/rest/RestUtilsTest.java
```diff
@@ -549,12 +554,21 @@ public void getOperationOrBlobIdFromUriTest() throws JSONException, UnsupportedE
...
           if (prefixesToRemove.contains(prefixToTestOn) || !expectedOutput.equals(blobId)) {
-            expectedOutput =
-                prefixesToRemove.contains(prefixToTestOn) ? expectedOutput : prefixToTestOn + expectedOutput;
+            if (!prefixesToRemove.contains(prefixToTestOn)) {
+              // remove query part if present
+              int endIndex = testPath.indexOf("?");
+              String processedTestPath = endIndex == -1 ? testPath : testPath.substring(0, endIndex);
+              // there are 2 cases to consider when prefix is not removed:
+              // 1. "/orNotToRemove/baseId" and expected output is "/baseId", because the first segment is considered as signed-path
+              // 2. "/orNotToRemove/" and expected output is "/orNotToRemove/", because there are only 2 segments (split by "/") and first segment won't be removed
```
jsjtzyy: The URIs themself are incorrect, we need to reach an agreement on how to handle such URIs. 
There is one risky case we should pay attention to if using my currently implementation:

If the URI is `/orNotToRemove/baseId` for container without signed path validation, then it should throw `BadRequest` exception. However in current implementation, the `/orNotToRemove` is removed in `getOperationOrBlobIdFromUri`, we should add extra check in `postProcessRequest` to ensure this URI with invalid prefix is rejected.

ambry-api/src/main/java/com.github.ambry/rest/RestUtils.java
```diff
@@ -832,6 +873,29 @@ public static void accountAndContainerNamePreconditionCheck(RestRequest restRequ
...
+    String signedPathInUri = getSignedPath(restRequest);
+    if (targetContainer.isSignedPathValidationRequired()) {
+      if (!signedPathInUri.equals(expectSignedPath)) {
+        throw new RestServiceException("Signed path in restRequest doesn't match the expected one",
+            RestServiceErrorCode.AccessDenied);
+      }
+    } else {
+      String signedIdPrefixWithNoSlash = SIGNED_ID_PREFIX.substring(0, SIGNED_ID_PREFIX.indexOf("/"));
+      if (!signedPathInUri.isEmpty() && !signedPathInUri.equals(signedIdPrefixWithNoSlash)) {
+        throw new RestServiceException("Invalid path is found in URI", RestServiceErrorCode.BadRequest);
```
jsjtzyy: This `if condition` checks invalid path in URI when container doesn't require signed path validation and it is not a signed id.  Thus, `/orNotToRemove/blobId` will be rejected here.

ambry-frontend/src/main/java/com.github.ambry.frontend/AmbrySecurityService.java
```diff
@@ -341,4 +354,22 @@ private void setAccountAndContainerHeaders(RestRequest restRequest, RestResponse
...
+   * @param restRequest the {@link RestRequest} that may contain signed path.
+   * @param expectCdnSignedPath the expected cnd signed path specified in {@link com.github.ambry.config.FrontendConfig}.
+   * @throws RestServiceException
+   */
+  private void validateCdnSignedPathIfRequired(RestRequest restRequest, String expectCdnSignedPath)
+      throws RestServiceException {
+    Container targetContainer = getContainerFromArgs(restRequest.getArgs());
+    if (targetContainer.isCdnSignedPathValidationRequired()) {
+      String signedPathInUri = getCdnSignedPath(restRequest);
+      if (!signedPathInUri.equals(expectCdnSignedPath)) {
```
jsjtzyy: After reconsideration, I think we should be strict on signed path validation in open source code. That is, if URI contains pure signed id or blob id (no path at the beginning), we still reject it because the empty path doesn't match expected signed path. Rejecting all mismatched URIs is more aligned with initial intention of introducing signed path validation for certain container. We can make extra changes in closed source code to differentiate external and internal URIs if needed.

jsjtzyy:
./gradlew build && ./gradlew test succeeded. Ready for review.

jsjtzyy:
Rebased and fixed the conflicts. Ready for review.

ambry-api/src/main/java/com.github.ambry/rest/RestUtils.java
```diff
@@ -589,6 +589,39 @@ public static String getOperationOrBlobIdFromUri(RestRequest restRequest, RestUt
...
+      // remove sub resource from path
+      endIndex = endIndex - subResource.name().length() - 1;
+      path = path.substring(0, endIndex);
+    }
+    if (path.startsWith("/")) {
+      // remove leading slash
+      path = path.substring(1);
+    }
+    // if this is a signed id, return empty secure path
+    if (!path.startsWith(RestUtils.SIGNED_ID_PREFIX)) {
```
cgtz: As we discussed offline, there is no reason that a signed ID cannot have a secure path prefix ahead of it.

For example, I think that calling this method on `/secure-path/signedId/ABCDEF` should return `secure-path`. 

I think that we can actually combine most of the logic of `getOperationOrBlobId` and `getSecurePath`. Let's see if we can make a helper method that returns a `Pair<String,String>` that includes the path prefix and operationOrBlobId. This new method could be called directly in `SecurityService` since you need both the path prefix and operationOrBlobId and we can remove the `getSecurePath` method.

```java
  public static Pair<String, String> getPrefixAndResourceFromUri(RestRequest restRequest, SubResource subResource,
      List<String> prefixesToRemove) {
    String path = restRequest.getPath();
    int startIndex = 0;
    // remove query string.
    int endIndex = path.indexOf("?");
    if (endIndex == -1) {
      endIndex = path.length();
    }
    // remove prefix.
    String prefixFound = "";
    if (prefixesToRemove != null) {
      for (String prefix : prefixesToRemove) {
        if (path.startsWith(prefix)) {
          prefixFound = prefix;
          startIndex = prefix.length();
          break;
        }
      }
    }
    // remove subresource if present.
    if (subResource != null) {
      // "- 1" removes the "slash" that precedes the sub-resource.
      endIndex = endIndex - subResource.name().length() - 1;
    }
    String operationOrBlobId = path.substring(startIndex, endIndex);
    if ((operationOrBlobId.isEmpty() || operationOrBlobId.equals("/")) && restRequest.getArgs()
        .containsKey(Headers.BLOB_ID)) {
      operationOrBlobId = restRequest.getArgs().get(Headers.BLOB_ID).toString();
    }
    return new Pair<>(prefixFound, operationOrBlobId);
  }

  public static String getOperationOrBlobIdFromUri(RestRequest restRequest, SubResource subResource,
      List<String> prefixesToRemove) {
    return getPrefixAndResourceFromUri(restRequest, subResource, prefixesToRemove).getSecond();
  }
```

ambry-api/src/main/java/com.github.ambry/frontend/Operations.java
```diff
@@ -22,4 +27,10 @@
   public static final String UPDATE_TTL = "updateTtl";
   public static final String STITCH = "stitch";
   public static final String GET_CLUSTER_MAP_SNAPSHOT = "getClusterMapSnapshot";
+  public static final Set<String> OperationsSet;
+
+  static {
+    OperationsSet = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
```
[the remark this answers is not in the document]
cgtz: I also remembered that we had some code in AmbryLI that did this same thing using reflection (so that we don't have to update the set every time we add a new operation):
```java
  static {
    INTERNAL_KEYS =
        Collections.unmodifiableList(getStaticFieldValuesAsStrings(InternalKeys.class).collect(Collectors.toList()));
    OPERATIONS =
        Collections.unmodifiableSet(getStaticFieldValuesAsStrings(Operations.class).collect(Collectors.toSet()));
  }

  /**
   * Get a stream of the accessible {@link String} values of the static fields in the provided {@link Class}.
   * @param type the {@link Class} to get static field values from.
   * @return a {@link Stream} of the static field value strings.
   */
  private static Stream<String> getStaticFieldValuesAsStrings(Class type) {
    return Arrays.stream(type.getFields()).filter(field -> Modifier.isStatic(field.getModifiers())).map(field -> {
      try {
        return field.get(null).toString();
      } catch (IllegalAccessException e) {
        throw new IllegalStateException("Could not get value of a static field, " + field + ", in " + type, e);
      }
    });
  }
```

Consider moving this over to open source.

ambry-frontend/src/main/java/com.github.ambry.frontend/AmbrySecurityService.java
```diff
@@ -113,6 +113,19 @@ public void postProcessRequest(RestRequest restRequest, Callback<Void> callback)
       } catch (Exception e) {
         exception = e;
       }
+    } else if (restRequest.getRestMethod() == RestMethod.GET) {
+      SubResource subresource = getBlobSubResource(restRequest);
```
cgtz: Let's rewrite this to get the prefix and secure path in one step:
```
Pair<String> prefixAndResource = getPrefixAndResourceFromUri(restRequest, subResource, frontendConfig.pathPrefixesToRemove);
if (needs checking) {
  validateSecurePathIfRequired(prefixAndResource.getFirst(), frontendConfig.securePathPrefix);
}
```

ambry-api/src/test/java/com.github.ambry/rest/RestUtilsTest.java
```diff
@@ -604,6 +605,54 @@ public void getBlobSubResourceTest() throws JSONException, UnsupportedEncodingEx
...
+  /**
+   * Tests {@link RestUtils#getSecurePath(RestRequest)}.
+   * @throws Exception
+   */
+  @Test
+  public void getSecurePathTest() throws Exception {
+    String blobId = UtilsTest.getRandomString(10);
+    String operation = "validOp";
+    String queryString = "?queryParam1=queryValue1&queryParam2=queryParam2=queryValue2";
+    String[] validIdUris = {"/", "/" + blobId, "", blobId, RestUtils.SIGNED_ID_PREFIX + "/" + blobId};
```
cgtz: let's account for the securePathPrefix + signedId case in these tests.

ambry-api/src/main/java/com.github.ambry/account/Container.java
```diff
@@ -311,6 +313,7 @@
   private final boolean mediaScanDisabled;
   private final String replicationPolicy;
   private final boolean ttlRequired;
+  private final boolean securePathValidationRequired;
```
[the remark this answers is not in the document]
jsjtzyy: Renamed. Thanks for the suggestion.

ambry-api/src/main/java/com.github.ambry/config/FrontendConfig.java
```diff
@@ -205,8 +212,10 @@ public FrontendConfig(VerifiableProperties verifiableProperties) {
         "com.github.ambry.frontend.AmbryUrlSigningServiceFactory");
     idSigningServiceFactory = verifiableProperties.getString(ID_SIGNING_SERVICE_FACTORY_KEY,
         "com.github.ambry.frontend.AmbryIdSigningServiceFactory");
-    pathPrefixesToRemove =
-        Arrays.asList(verifiableProperties.getString("frontend.path.prefixes.to.remove", "").split(","));
+    securePathToValidate = verifiableProperties.getString("frontend.secure.path.to.validate", "");
```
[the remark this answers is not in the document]
jsjtzyy: Renamed.

ambry-api/src/main/java/com.github.ambry/frontend/Operations.java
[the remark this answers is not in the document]
jsjtzyy: Good idea. This helper method is moved to `Utils.java` and invoked in `AmbrySecurityService`.

jsjtzyy:
@cgtz addressed your comments. Code is ready for review.

PR #1120, opened 2019-01-03
Pass IdSigningService into SecurityServiceFactory

cgtz:
Allow SecurityService implementations to leverage IdSigningService, if
they need to.


ambry-frontend/src/main/java/com.github.ambry.frontend/AmbrySecurityServiceFactory.java
```diff
@@ -31,7 +31,7 @@
   private final UrlSigningService urlSigningService;
 
   public AmbrySecurityServiceFactory(VerifiableProperties verifiableProperties, ClusterMap clusterMap,
-      AccountService accountService, UrlSigningService urlSigningService,
+      AccountService accountService, UrlSigningService urlSigningService, IdSigningService idSigningService,
```
jsjtzyy: is it worth keeping `idSigningService` as a variable inside class for future use?

cgtz: I think I prefer leaving it out, since the variable is easy to add later on and I don't want to add additional unused code.

PR #1123, opened 2019-01-09
Avoid returning 416 when range end past blob

cgtz:
The original implementation of ByteRange.toResolvedByteRange was
stricter than the RFC called for and required that the end of the range
be less than the blob size. This change allows ranges where the end
offset or suffix length is greater than the blob size to be served.

Per rfc7233:
   A client can limit the number of bytes requested without knowing the
   size of the selected representation.  If the last-byte-pos value is
   absent, or if the value is greater than or equal to the current
   length of the representation data, the byte range is interpreted as
   the remainder of the representation
and
   If the selected representation is shorter than the specified
   suffix-length, the entire representation is used.

This change also refactors the implementation of ByteRange to use
inheritance instead of switch statements.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
+++ b/ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
@@ -31,7 +31,7 @@ public abstract class ByteRange {
       throw new IllegalArgumentException(
           "Invalid range offsets provided for ByteRange; startOffset=" + startOffset + ", endOffset=" + endOffset);
     }
-    return new OffsetRange(startOffset, endOffset);
+    return new ClosedRange(startOffset, endOffset);
   }
 
   /**
@@ -44,7 +44,7 @@ public abstract class ByteRange {
     if (startOffset < 0) {
       throw new IllegalArgumentException("Invalid range offsets provided for ByteRange; startOffset=" + startOffset);
     }
-    return new FromStartOffset(startOffset);
+    return new OpenRange(startOffset);
   }
 
   /**
@@ -57,7 +57,13 @@ public abstract class ByteRange {
     if (lastNBytes < 0) {
       throw new IllegalArgumentException("Invalid range offsets provided for ByteRange; lastNBytes=" + lastNBytes);
     }
-    return new LastNBytes(lastNBytes);
+    return new SuffixRange(lastNBytes);
+  }
+
+  /**
+   * All implementations should be defined in this class.
+   */
+  private ByteRange() {
   }
 
   // Implement the following methods if they are supported for the range type.
@@ -142,14 +148,14 @@ public abstract class ByteRange {
   /**
    * A range from a start offset to an end offset.
    */
-  private static class OffsetRange extends ByteRange {
+  private static class ClosedRange extends ByteRange {
     private long startOffset;
     private long endOffset;
 
     /**
      * @see ByteRange#fromOffsetRange(long, long)
      */
-    private OffsetRange(long startOffset, long endOffset) {
+    private ClosedRange(long startOffset, long endOffset) {
       this.startOffset = startOffset;
       this.endOffset = endOffset;
     }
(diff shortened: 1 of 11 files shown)


ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
```diff
@@ -17,208 +17,110 @@
...
-   * @throws IllegalArgumentException
+   * @throws IllegalArgumentException if the start offset is less than 0, or the end offset is less than the start.
    */
   public static ByteRange fromOffsetRange(long startOffset, long endOffset) {
     if (startOffset < 0 || endOffset < startOffset) {
       throw new IllegalArgumentException(
           "Invalid range offsets provided for ByteRange; startOffset=" + startOffset + ", endOffset=" + endOffset);
     }
-    return new ByteRange(startOffset, endOffset, ByteRangeType.OFFSET_RANGE);
+    return new OffsetRange(startOffset, endOffset);
```
lightningrob: An abstract class shouldn't create instances of its subclasses.  Could test static constructors be moved to a factory class eg. ByteRangeFactory?

ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
```diff
@@ -236,4 +144,183 @@ public int hashCode() {
...
+    @Override
+    public int hashCode() {
+      return 31 * Long.hashCode(startOffset) + Long.hashCode(endOffset);
+    }
+  }
+
+  /**
+   * A range from a start offset to the end of an object.
+   */
+  private static class OpenRange extends ByteRange {
```
pnarayanan: why is `getRangeSize()` not defined for this?

ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
pnarayanan: abstract class instantiating subclasses does feel a bit odd. I do like the idea of avoiding the switch logic though. Is there a better way here?

cgtz: Would deprecation be an acceptable compromise? I could move the impls and static methods into a `ByteRanges` util class:

```java
public abstract class ByteRange {
  // use ByteRanges.fromStartOffset instead
  @Deprecated
  static ByteRange fromStartOffset(long start, long end) {
    return ByteRanges.fromStartOffset(start, end);
  }
  ...
  abstract ByteRange toResolvedByteRange(long totalSize);
  ...
  long getStartOffset() {
    throw new UnsupportedOperationException();
  }
```

```java
public class ByteRanges {
  public ByteRange fromStartOffset(long start, long end) {
    return new ClosedRange(start, end);
  }
  ...
  private static class ClosedRanged extends ByteRange {
    ...
  }
}
```

ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
cgtz: Since only the start offset is known for this range type, so you don't know how long the range is without knowing the total object size.

ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
lightningrob: LGTM

cgtz:
@pnarayanan @lightningrob Updated PR with the util class change

ambry-api/src/main/java/com.github.ambry/router/ByteRange.java
```diff
@@ -17,208 +17,112 @@
...
+   * @throws IllegalArgumentException if the start offset is less than 0, or the end offset is less than the start.
    */
+  @Deprecated
   public static ByteRange fromOffsetRange(long startOffset, long endOffset) {
-    if (startOffset < 0 || endOffset < startOffset) {
-      throw new IllegalArgumentException(
-          "Invalid range offsets provided for ByteRange; startOffset=" + startOffset + ", endOffset=" + endOffset);
-    }
-    return new ByteRange(startOffset, endOffset, ByteRangeType.OFFSET_RANGE);
+    return ByteRange.fromOffsetRange(startOffset, endOffset);
```
lightningrob: This should be ByteRanges.  Calling this will result in a stack overflow.
Even deprecated methods should have test cases.  It looks like all the test cases were changed to use the utility class.

cgtz: Yikes, thanks for catching this. I have added tests for all of the deprecated methods

PR #1124, opened 2019-01-09
Added isBackedUp option for Container objects

dharju:


ambry-api/src/main/java/com.github.ambry/account/ContainerBuilder.java
```diff
@@ -162,6 +164,16 @@ public ContainerBuilder setCacheable(boolean cacheable) {
     return this;
   }
 
+  /**
+   * Sets the backup setting of the {@link Container} to build
+   * @param isBackedUp The backup setting to set.
+   * @return This builder.
+   */
+  public ContainerBuilder setIsBackedUp(boolean isBackedUp) {
```
cgtz: If you decide to change the field name, make sure to update the methods/variables in the builder too.

ambry-api/src/main/java/com.github.ambry/account/Container.java
```diff
@@ -380,7 +388,8 @@ private Container(JSONObject metadata, short parentAccountId) throws JSONExcepti
    */
   Container(short id, String name, ContainerStatus status, String description, boolean encrypted,
       boolean previouslyEncrypted, boolean cacheable, boolean mediaScanDisabled, String replicationPolicy,
-      boolean ttlRequired, Set<String> contentTypeWhitelistForFilenamesOnDownload, short parentAccountId) {
+      boolean ttlRequired, Set<String> contentTypeWhitelistForFilenamesOnDownload, short parentAccountId,
```
cgtz: another minor thing: Can you keep `parentAccountId` as the last constructor arg, since it's the only field that doesn't get serialized into json.

ambry-api/src/main/java/com.github.ambry/account/Container.java
```diff
@@ -380,7 +388,8 @@ private Container(JSONObject metadata, short parentAccountId) throws JSONExcepti
    */
```
cgtz: add a doc for the new arg.

ambry-api/src/main/java/com.github.ambry/account/Container.java
```diff
@@ -63,6 +63,7 @@
   static final String STATUS_KEY = "status";
   static final String DESCRIPTION_KEY = "description";
   static final String IS_PRIVATE_KEY = "isPrivate";
+  static final String IS_BACKED_UP_KEY = "isBackedUp";
```
jsjtzyy: Does this field mean that blobs in this container are required to be backed up in cloud? If so, maybe we could rename it to `backUpRequired` or `cloudBackUpRequired`?

jsjtzyy:
Looks like you need to update the containerBuilder in `InMemAccountService.java` as well.

ambry-api/src/main/java/com.github.ambry/account/Container.java
lightningrob: I vote for `backupEnabled`.

dharju: will change to backupEnabled

ambry-api/src/main/java/com.github.ambry/account/Container.java
dharju: sure

ambry-api/src/main/java/com.github.ambry/account/Container.java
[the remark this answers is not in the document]
dharju: changing to "BACKUP_ENABLED"

ambry-api/src/main/java/com.github.ambry/account/ContainerBuilder.java
dharju: 👍

dharju:
> Looks like you need to update the containerBuilder in `InMemAccountService.java` as well.

added

ambry-api/src/main/java/com.github.ambry/account/Container.java
dharju: added

PR #1130, opened 2019-01-28
Support to pass externalAssetTag from HTTP header to NotificationSystem.

zzmao:
Add support to pass externalAssetTag from HTTP header to NotificationSystem. 
The String is pass through `BlobProperties`.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
+++ b/ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
@@ -143,6 +143,9 @@ public class BlobProperties {
     return isEncrypted;
   }
 
+  /**
+   * @return the ExternalAssetTag of a uploaded blob. Can be null.
+   */
   public String getExternalAssetTag() {
     return externalAssetTag;
   }
--- a/ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
+++ b/ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
@@ -32,6 +32,8 @@ public class BlobProperties {
   private final boolean isEncrypted;
   private long blobSize;
   private long timeToLiveInSeconds;
+  // Non persistent blob properties.
+  private final String externalAssetTag;
 
   /**
    * @param blobSize The size of the blob in bytes
@@ -42,7 +44,7 @@ public class BlobProperties {
    */
   public BlobProperties(long blobSize, String serviceId, short accountId, short containerId, boolean isEncrypted) {
     this(blobSize, serviceId, null, null, false, Utils.Infinite_Time, SystemTime.getInstance().milliseconds(),
-        accountId, containerId, isEncrypted);
+        accountId, containerId, isEncrypted, null);
   }
 
   /**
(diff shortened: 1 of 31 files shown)


zzmao:
@lightningrob , after a discussion with @cgtz and Matt. We think we will never store this tag to storage node since it may contain PII. 
To avoid confusion in `BlobProperties` where all members are persisted, we moved it to `PutBlobOptions`.  
How do you think of this?

ambry-api/src/main/java/com.github.ambry/rest/RestUtils.java
```diff
@@ -177,6 +177,10 @@
      * The maximum size of the blob that can be uploaded using the URL.
      */
     public static final String MAX_UPLOAD_SIZE = "x-ambry-max-upload-size";
+    /**
+     * An externalAssetTag for this blob.
+     */
+    public static final String EXTERNAL_ASSET_TAG = "x-ambry-external-asset-tag";
```
lightningrob: Ambry client also defines this.  We could move them both to ambryclient-utils module.

lightningrob:
> @lightningrob , after a discussion with @cgtz and Matt. We think we will never store this tag to storage node since it may contain PII.
> To avoid confusion in `BlobProperties` where all members are persisted, we moved it to `PutBlobOptions`.
> How do you think of this?

I guess there's no perfect place to put this tag.  PutBlobOptions doesn't seem like the best fit either, but I don't have a better suggestion.  If we later add any additional parameters, we can combine them into a data structure.

ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
```diff
@@ -34,9 +34,10 @@
    * @param account The {@link Account} for the blob
    * @param container The {@link Container} for the blob
    * @param notificationBlobType The {@link NotificationBlobType} of this blob.
+   * @param externalAssetTag The external asset tag of this blob.
    */
   void onBlobCreated(String blobId, BlobProperties blobProperties, Account account, Container container,
```
cgtz: at some point (maybe not this PR), we might want to combine the account, container, notificationBlobType and externalAssetTag arguments into a data structure that gets passed into this method

zzmao: @lightningrob , After a discussion with @cgtz , we can actually create a new class to include `Account`, `Container` and `externalAssetTag`. All non-persistent blob properties can be included in this class in the future. Tentatively name it as `NonPersistentBlobProperties`.

zzmao: Looks like what I proposed above is not a good way. 
Introducing `NonPersistentBlobProperties`/`BlobNotifications` all requires change to ```Future<String> putBlob(BlobProperties blobProperties, byte[] userMetadata, ReadableStreamChannel channel,
      PutBlobOptions options, Callback<String> callback)```. 
I prefer not to change this elegant interface. 
Will add `externalAssetTag` to `BlobProperties` and comment it as non-persistent property.  @cgtz  @lightningrob

zzmao:
Introduced `externalAssetTag` to `BlobProperties`. Refactor changed a lot files, but main changes are from `BlobProperties.java`, `PutOperation.java`, `PutManagerTest.java`.

ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
```diff
@@ -137,6 +143,10 @@ public boolean isEncrypted() {
     return isEncrypted;
   }
 
+  public String getExternalAssetTag() {
```
lightningrob: Javadoc (even though existing methods don't all have them)

PR #1144, opened 2019-03-31
Added encryption to VCR

dharju:
Added CloudBlobCryptoService interface, implementation, and tests.  100% coverage of AzureCloudBlobCryptoService


ambry-api/src/main/java/com.github.ambry/config/CloudConfig.java
```diff
@@ -24,6 +24,16 @@
   public static final String CLOUD_DESTINATION_FACTORY_CLASS = "cloud.destination.factory.class";
   public static final String DEFAULT_CLOUD_DESTINATION_FACTORY_CLASS =
       "com.github.ambry.cloud.azure.AzureCloudDestinationFactory";
+  public static final String CRYPTO_SERVICE_FACTORY_CLASS = "vcr.crypto.service.factory.class";
```
lightningrob: Are crypto and kms service factories already defined in another config which we can pass to the CloudBackupManager instead of duplicating them?  Will we ever have different factories used for router and vcr?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -14,6 +14,7 @@
 package com.github.ambry.cloud;
 
 import com.codahale.metrics.MetricRegistry;
+import com.github.ambry.cloud.azure.AzureCloudBlobCryptoService;
```
lightningrob: Shouldn't be here.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -52,7 +54,8 @@ public CloudBlobMetadata() {
    * @param expirationTime The blob expiration time.
    * @param size The blob size.
    */
-  public CloudBlobMetadata(BlobId blobId, long creationTime, long expirationTime, long size) {
+  public CloudBlobMetadata(BlobId blobId, long creationTime, long expirationTime, long size,
+      boolean usesCloudEncryption) {
```
lightningrob: Don't you need to say: this. usesCloudEncryption = usesCloudEncryption ?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -213,6 +216,23 @@ public CloudBlobMetadata setContainerId(int containerId) {
...
 
+  /**
+   * @return whether the blob was cloud encrypted.
+   */
+  public boolean getUsesCloudEncryption() {
+    return usesCloudEncryption;
+  }
+
+  /**
+   * Set the encryption context.
```
lightningrob: Minor: "context" seems weird here since it's not the name of method or argument.  Maybe: set the cloud encryption flag?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -96,14 +101,21 @@ public void put(MessageWriteSet messageSetToWrite) throws StoreException {
...
-  private void putBlob(MessageInfo messageInfo, ByteBuffer messageBuf, long size) throws CloudStorageException {
+  private void putBlob(MessageInfo messageInfo, ByteBuffer messageBuf, long size)
+      throws CloudStorageException, IOException, GeneralSecurityException {
     boolean performUpload = messageInfo.getExpirationTimeInMs() == Utils.Infinite_Time && !messageInfo.isDeleted();
-    // May need to encrypt buffer before upload
     if (performUpload) {
       BlobId blobId = (BlobId) messageInfo.getStoreKey();
+      // TODO: would be more efficient to call blobId.isEncrypted()
+      boolean usesCloudEncryption = false;
+      if (cryptoService != null && (blobId.getVersion() < 4 || !BlobId.isEncrypted(blobId.getID()))) {
```
lightningrob: Can use constant for "4".
More importantly: if the blob is not already encrypted, the config flag "requireEncryption" is true, and cryptoService is null, upload should be skipped.  (You'll see some of this logic when you merge with master.)

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudBlobCryptoService.java
```diff
@@ -0,0 +1,158 @@
...
+      }
+      byte[] encryptedKey = new byte[encryptedKeySize];
+      byte[] encryptedData = new byte[encryptedDataSize];
+      buffer.get(encryptedKey);
+      buffer.get(encryptedData);
+      crc.update(buffer.array(), startOffset, INITIAL_MESSAGE_LENGTH + encryptedKeySize + encryptedDataSize);
+      long expectedCrc = buffer.getLong();
+      long actualCrc = crc.getValue();
+      if (actualCrc != expectedCrc) {
+        throw new GeneralSecurityException("Encrypted blob is corrupt");
```
lightningrob: Include expected and actual crc in message?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudBlobCryptoService.java
```diff
@@ -0,0 +1,158 @@
...
+ * Implementation of CloudBlobCryptoService, encrypts byte buffers by
+ * 1. generating a random encryption key
+ * 2. encrypting that random key with a key from KeyManagementService
+ *    that is associated with the given context string
+ * 3. encrypting a data byte buffer with the generated random key
+ * 4. returning a byte buffer containing the serialized encrypted key
+ *    and encrypted data (see {@link EncryptedKeyAndEncryptedData} below
+ *    for serialization description
+ */
+public class AzureCloudBlobCryptoService implements CloudBlobCryptoService {
```
lightningrob: You can leave Azure out of the name (and package) since it has nothing Azure specific.
I'm concerned that we're adding another message format (EncryptedKeyAndEncryptedData).  How much work would it be to use the same one used by the router?
By not deserializing the blob properties, we are missing some opportunities to avoid encryption on blobs that are already encrypted but the blobId bit was not set.  (I think this only applies to V3 blobs so maybe not a large set.)

ambry-api/src/main/java/com.github.ambry/config/CloudConfig.java
```diff
@@ -24,6 +24,16 @@
...
+  public static final String DEFAULT_CRYPTO_SERVICE_FACTORY_CLASS = "com.github.ambry.router.GCMCryptoServiceFactory";
+  public static final String KMS_SERVICE_FACTORY_CLASS = "vcr.kms.service.factory.class";
+  public static final String DEFAULT_KMS_SERVICE_FACTORY_CLASS =
+      "com.github.ambry.router.SingleKeyManagementServiceFactory";
+  public static final String KMS_SERVICE_KEY_CONTEXT = "vcr.kms.service.key.context";
+  public static final String DEFAULT_KMS_SERVICE_KEY_CONTEXT = "vcr.kms.service.key.context";
```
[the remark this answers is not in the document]
dharju: Can you elaborate?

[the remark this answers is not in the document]
lightningrob: The DEFAULT value just repeats the name of the key.  Shouldn't it be a value like "default-backup"?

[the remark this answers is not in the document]
dharju: Ah of course, you're right.  I'll change that

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudBlobCryptoService.java
dharju: Yeah I can name it something different.

My concern with using the current router encryption scheme is that to do that, I'd have to
1. Parse the message buffer into its component headers and fields
2. Adjust all the offsets in the header message as the blob encryption field is among the first fields, and adjust any related calculated CRCs
3. Live with weird inconsistent data states like the blob encryption bit in the blob ID being out-of-sync with the actual encryption status beyond Version 3.

The encryption tool would also be limited to use on specific versions of blobs, and would have to change or be tested whenever new versions of blobs (or new versions of components of blobs like message headers) are introduced.  By decoupling cloud backup encryption from whatever the blob content is, it becomes a significantly less complex and more modular, independent, and cleaner component (you could use it to encrypt-decrypt any ByteBuffer, not just Ambry blobs)

I thought about that when writing it (we'd be potentially encrypting V3s that actually were already encrypted but weren't indicated as such according to their blob ID encryption bit) but I felt that 
1. we haven't been making any more new V3s for a while, so this won't be a recurring issue
2. the scale of the affected "double" encrypted blobs would be limited

Talking to other engineers on the team, it sounds like it may not even be a problem as the issue was contained before it was deployed so that's something I'll confirm

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
dharju: removed

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudBlobCryptoService.java
zzmao: Looked at the code, looks like it's possible to re-use existing code for un-encrypted blobs and use same format like router.

For deserialization: 
We have `MessageFormatRecord::deserializeBlobAll`, it's pretty neat to use. 

For encryption:
just encrypt it like `encryptJob`.  

For actual write:
After encryption, you can construct `PutMessageFormatInputStream`(to serialize) and write it to CloudStore. (`handlePutRequest` is an example).

dharju: @zzmao it is indeed possible.  It is also indeed more error prone and more likely to involve unindented consequences by mutating the data.  I am very much aware of `MessageFormatRecord` and `PutMessageFormatInputStream`; I wrote `BlobIdTransformer` which uses these classes to parse and mutate blob records for the rewrite.  It is through that experience that I know that bit fiddling with existing data can be very problematic and risky.

This current encryption scheme is independent of the data that it actually encrypts, and thus is independent of any version changes, future or past, that may happen with any components of `MessageFormatRecord` or `PutMessageFormatInputStream`.  I talked with @cgtz and @pnarayanan too about this and they agreed with me that having an additional header that isolates this encryption process from other data manipulation processes is the way to go.

zzmao: I see. Mutating data is indeed error prone.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -28,6 +28,7 @@
   public static final String FIELD_EXPIRATION_TIME = "expirationTime";
   public static final String FIELD_ACCOUNT_ID = "accountId";
   public static final String FIELD_CONTAINER_ID = "containerId";
+  public static final String FIELD_USES_CLOUD_ENCRYPTION = "usesCloudEncryption";
```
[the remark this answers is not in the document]
dharju: Does "DB" refer to `CloudBlobMetadata`?  If it does what does "UserMetadata" refer to?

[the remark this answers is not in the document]
zzmao: DB is cosmodb. 
UserMetadata means Azure blob user metadata. (for DB rebuild)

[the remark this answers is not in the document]
dharju: So what classes do I need to change to accommodate this beyond `CloudBlobMetadata`?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
dharju: changed

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
dharju: Yeah I had originally had it as the encryption context string but changed it later.  Changed the comment

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
dharju: added

ambry-api/src/main/java/com.github.ambry/config/CloudConfig.java
```diff
@@ -24,6 +24,16 @@
...
+  public static final String CLOUD_BLOB_CRYPTO_SERVICE_FACTORY_CLASS = "vcr.cloud.blob.crypto.service.factory.class";
```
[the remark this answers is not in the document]
dharju: had mistaken that as a convention. removed

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudBlobCryptoService.java
dharju: sure, added

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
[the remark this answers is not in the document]
lightningrob: AzureCloudDestination

ambry-api/src/main/java/com.github.ambry/config/CloudConfig.java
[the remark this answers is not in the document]
dharju: Yeah they can be found in `RouterConfig`.  It's unlikely, and if you don't mind the extra dependency I can change it to use the router config.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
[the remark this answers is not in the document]
dharju: By `AzureCloudDestination` you're just talking about changing the `getMetadataMap` method right?

lightningrob:
Correct.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudBlobCryptoService.java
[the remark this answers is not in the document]
dharju: I think taking out the factory puts this in conflict with your earlier suggestion about storing the encryption method in the metadata.  If we have a factory, the factory can take the encryption method in as an argument and deliver the correct CloudBlobCryptoAgent implementation to decrypt the data

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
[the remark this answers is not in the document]
dharju: Changing to this

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudBlobCryptoService.java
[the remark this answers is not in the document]
zzmao: OK to use this factory, but I don't think it's a conflict. Will comment in another place.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -96,14 +101,26 @@ public void put(MessageWriteSet messageSetToWrite) throws StoreException {
...
+  private void putBlob(MessageInfo messageInfo, ByteBuffer messageBuf, long size)
+      throws CloudStorageException, IOException, GeneralSecurityException {
     boolean performUpload = messageInfo.getExpirationTimeInMs() == Utils.Infinite_Time && !messageInfo.isDeleted();
-    // May need to encrypt buffer before upload
     if (performUpload) {
       BlobId blobId = (BlobId) messageInfo.getStoreKey();
+      // TODO: would be more efficient to call blobId.isEncrypted()
+      String kmsContext = cryptoService.getEncryptionContext();
+      CloudBlobMetadata.EncryptionType encryptionType;
+      if (cryptoService != null && (blobId.getVersion() < BlobId.BLOB_ID_V4 || !BlobId.isEncrypted(blobId.getID()))) {
```
zzmao: rename it to cryptoAgent.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -96,14 +101,26 @@ public void put(MessageWriteSet messageSetToWrite) throws StoreException {
...
+        // Need to encrypt the buffer before upload
+        messageBuf = cryptoService.encrypt(messageBuf);
+        encryptionType = CloudBlobMetadata.EncryptionType.VCR;
```
zzmao: you can add an interface to `CryptoAgent`: `getVcrEncryptionFormat()`,  call it here and pass it to CloudBlobMetadata.

zzmao: vcrEncryptionFormat could be the class name of Encryption Format .

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -38,6 +41,19 @@
   private int containerId;
   private long expirationTime;
   private long deletionTime;
+  private String kmsContext;
+  private EncryptionType encryptionType;
```
zzmao: Does `encryptionInitiator` sounds better?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -38,6 +41,19 @@
...
+  private VcrEncryptionFormat vcrEncryptionFormat;
+
```
zzmao: Looks like we also need to store `CryptoService(Factory)` and `KMS(Factory)` in metadata. 

Because they are all varying between ambry and ambryLI. 

To encrypt a blob, we need `cryptoService`, `kmsService`, `kmsContext`, and `encryptionFormat`. Same requirements in decryption. So it's nice to store all of them in metadata. 

Storing class name is my preference, but also open to other suggestions.

zzmao: To keep integrity, `kmsScope` is also suggested to store in metadata.

zzmao: With this in mind, I think it's better to construct `cryptoAgent` in `CloudBlobStore` constructor instead of passing it in. 
Main reasons are:
1. This looks cleaner for CloudBackupManger` and `CloudBlobStore` constructor. 
2. CryptoAgent can be generated in fly by looking at metadata. 

It's OK to pass `vProps` to `CloudBackupManger` and then to `CloudBlobStore`.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobCryptoAgentImpl.java
```diff
@@ -0,0 +1,163 @@
...
+   *                    supports int sized sizes
+   *
+   *  encrypted key   - The encrypted key
+   *
+   *  encrypted data  - The encrypted data
+   *
+   *  crc             - The crc of the message
+   *
+   */
+  static class EncryptedKeyAndEncryptedData {
```
zzmao: `CloudEncryptionFormatV1`?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobCryptoAgentImpl.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    static EncryptedKeyAndEncryptedData deserialize(ByteBuffer buffer) throws GeneralSecurityException {
+      int startOffset = buffer.position();
+      Crc32 crc = new Crc32();
+      short version = buffer.getShort();
+      if (version != 1) {
+        throw new GeneralSecurityException("Unrecognized version");
+      }
+      int encryptedKeySize = buffer.getInt();
```
zzmao: Should we add specified (de)serialization based on version?

if version == 1{

} else {
....
}

as a placeholder.

dharju: Original intent was to add this when we actually had new versions

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
dharju: seems like adding all this information to cloudBlobMetadata may create a security risk (which is what encryption is attempting to protect against)

ambry-router/src/main/java/com.github.ambry.router/SingleKeyManagementService.java
```diff
@@ -67,6 +72,11 @@ public void register(short accountId, short containerId) {
    */
   @Override
   public SecretKeySpec getKey(short accountId, short containerId) throws GeneralSecurityException {
```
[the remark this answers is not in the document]
lightningrob: I prefer to keep the interface backwards compatible.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
lightningrob: I don't want to crowd the metadata with all that stuff.  How does the router do it?  Can't these parameters go into the blob block header so whoever consumes it later knows how to decrypt it?

zzmao: Yes. They can. @lightningrob

lightningrob:
Closing obsolete PR

PR #1142, opened 2019-03-28
Add VCRServer, StaticVCRCluster and tests

lightningrob:

PR #1148, opened 2019-04-03
Add VirtualReplicatorClusterFactory and refactor endToEndCloudBackupTest SSL config.

zzmao:
Add VirtualReplicatorClusterFactory interface and StaticVcrClusterFactory implementation.
Refactor endToEndCloudBackupTest SSL config.
Follow up PR of #1142 .


ambry-cloud/src/main/java/com.github.ambry.cloud/StaticVcrClusterFactory.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  public VirtualReplicatorCluster getVirtualReplicatorCluster() {
+    if (virtualReplicatorCluster == null) {
+      virtualReplicatorCluster = new StaticVcrCluster(cloudConfig, clusterMapConfig, clusterMap);
+    }
+    return virtualReplicatorCluster;
+  }
+
+  @Override
```
lightningrob: Minor: don't need this override if it's a noop.

ambry-cloud/src/main/java/com.github.ambry.cloud/StaticVcrClusterFactory.java
```diff
@@ -0,0 +1,50 @@
...
+  private VirtualReplicatorCluster virtualReplicatorCluster;
+
+  public StaticVcrClusterFactory(CloudConfig cloudConfig, ClusterMapConfig clusterMapConfig, ClusterMap clusterMap) {
+    this.cloudConfig = cloudConfig;
+    this.clusterMapConfig = clusterMapConfig;
+    this.clusterMap = clusterMap;
+  }
+
+  @Override
+  public VirtualReplicatorCluster getVirtualReplicatorCluster() {
```
dharju: Singleton isn't threadsafe.  We're not worried about multiple threads calling this method?

zzmao: Good point.
So far, we don't have multiple thread use case here. It initialized once in VCRServer(). 
Also, StaticVcrClusterFactory is a temp solution for experiment. We will move to HelixVcrClusterFactory soon.

dharju: Will `HelixVcrClusterFactory` access this in a single thread context?  

This page has a good design suggestion for how to write this in an efficient and thread-safe manner:

https://www.journaldev.com/171/thread-safety-in-java-singleton-classes-with-example-code
image

dharju: not sure if the `mutex` is necessary though as I think `synchronized` in a static context locks on the class object anyway

zzmao: `synchronized` added since this is a one time call. Will improve it if complex usages come. 
Thanks for the article.

lightningrob: Another approach is to create the instance upfront in the constructor (eager vs lazy).  Then there is no need for synchronization.  Lazy construction is preferred if the object is heavyweight and you don't know ahead of time whether it will be needed.  But this getter will not be called many times.

PR #1150, opened 2019-04-05
Introduce (un)satisfied request metrics to better evaluate availability

jsjtzyy:
Introduced satisfied/unsatisfied metrics to evaluate availability for each
type request in terms of latency and transmission rate. Set up separate
configurable thresholds to determine whether request is satisfied or not.


ambry-api/src/main/java/com.github.ambry/config/PerformanceConfig.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Configuration parameters required to help evaluate performance of Ambry, i.e latency, health.
+ */
+public class PerformanceConfig {
+
+  /**
+   * The time to first byte (millisecond) threshold for success GET request(2xx status code). The request is considered
+   * "Unsatisfied" if the time exceeds this threshold.
+   */
+  @Config("performance.success.get.time.to.first.byte.threshold")
```
cgtz: Take a look at `FrontendConfig::restRequestQuota`. For that config, Ze used a json object so that we didn't have to make a separate config for each rest method.

jsjtzyy: good point, I have removed unnecessary configs.

ambry-api/src/main/java/com.github.ambry/rest/RestRequestMetrics.java
```diff
@@ -90,5 +95,8 @@ public RestRequestMetrics(Class ownerClass, String requestType, MetricRegistry m
 
     operationRate = metricRegistry.meter(MetricRegistry.name(ownerClass, requestType + OPERATION_RATE_SUFFIX));
     operationError = metricRegistry.counter(MetricRegistry.name(ownerClass, requestType + OPERATION_ERROR_SUFFIX));
+
+    unSatisfiedRequestCount = metricRegistry.counter(MetricRegistry.name(ownerClass, requestType + UNSATISFIED_REQUEST_COUNT_SUFFIX));
```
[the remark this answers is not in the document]
jsjtzyy: good catch, fixed.

ambry-api/src/main/java/com.github.ambry/rest/RestRequestMetricsTracker.java
```diff
@@ -237,6 +237,25 @@ public void recordMetrics() {
...
 
+  public long getRoundTripTimeInMs() {
+    return nioMetricsTracker.roundTripTimeInMs;
+  }
+
+  public long getTimeToFirstByteInMs(){
+    return nioMetricsTracker.timeToFirstByteInMs;
+  }
+
+  public void updateUnSatisfiedRequestCounter(){
```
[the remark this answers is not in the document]
jsjtzyy: I noticed that the metrics would be injected later and `recordMetrics` is called when closing the request. Hence, the `updateUnsatisfiedRequestCounter` is called only after request is closed. I see your point here, maybe I should try to integrate it with `recordMetrics` in case anyone invoke this in other places.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -305,6 +314,70 @@ void close(Exception exception) {
...
+  private void evaluateHealthAndUpdateMetrics() {
+    RestRequestMetricsTracker restRequestMetricsTracker = request.getRestRequestMetricsTracker();
+    long roundTripTimeInMs = restRequestMetricsTracker.getRoundTripTimeInMs();
+    long timeToFirstBytesInMs = restRequestMetricsTracker.getTimeToFirstByteInMs();
+    long totalOutboundBytes = totalBytesReceived.get();
+    long totalInboundBytes = request.getBytesReceived();
+    boolean isSatisfied = false;
+    switch (request.getRestMethod()) {
+      case GET:
+        if (successResponseSet.contains(responseStatus) && (
```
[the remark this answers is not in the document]
jsjtzyy: good idea. Will make changes as suggested.

jsjtzyy:
addressed @cgtz 's comments. PR is ready for review.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -305,9 +321,58 @@ void close(Exception exception) {
...
 
+  private void evaluatePerformanceAndUpdateMetrics() {
+    RestRequestMetricsTracker restRequestMetricsTracker = request.getMetricsTracker();
+    long roundTripTimeInMs = restRequestMetricsTracker.getRoundTripTimeInMs();
+    long timeToFirstBytesInMs = restRequestMetricsTracker.getTimeToFirstByteInMs();
+    long totalOutboundBytes = totalBytesReceived.get();
+    long totalInboundBytes = request.getBytesReceived();
+    RestMethod method = request.getRestMethod();
+    Map<PerformanceIndex, Long> requestPerfToCheck = new HashMap<>();
+    responseStatus = errorResponseStatus == null ? responseStatus : errorResponseStatus;
```
jsjtzyy: this statement is indispensable because in some error cases, the request is closed before `responseStatus` gets populated.

jsjtzyy:
@cgtz gentle reminder to review

ambry-api/src/main/java/com.github.ambry/commons/Threshold.java
```diff
@@ -0,0 +1,66 @@
...
+
+import com.github.ambry.utils.Pair;
+import java.util.Map;
+
+
+/**
+ * A data structure holding various thresholds based on different criteria.
+ */
+public class Threshold {
+  private final Map<PerformanceIndex, Pair<Long, Boolean>> perfIndexAndCriteria;
```
lightningrob: I don't find the Pair construct very intuitive until I dive into it.  Suggest adding a simple class with fields thresholdValue and isUpperBound (or lower) to make it a bit clearer.  or a json map like Casey suggested.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -305,9 +321,58 @@ void close(Exception exception) {
   private void closeRequest() {
     if (request != null && request.isOpen()) {
       request.close();
+      evaluatePerformanceAndUpdateMetrics();
     }
   }
 
+  private void evaluatePerformanceAndUpdateMetrics() {
```
lightningrob: javadoc

ambry-api/src/main/java/com.github.ambry/commons/Threshold.java
cgtz: +1 Since this is passed in by callers to the constructor

ambry-api/src/main/java/com.github.ambry/commons/Threshold.java
```diff
@@ -0,0 +1,66 @@
...
+package com.github.ambry.commons;
+
+import com.github.ambry.utils.Pair;
+import java.util.Map;
+
+
+/**
+ * A data structure holding various thresholds based on different criteria.
+ */
+public class Threshold {
```
cgtz: I think calling this class Thresholds instead of Threshold will be good to reflect that there are multiple checks done.

ambry-api/src/main/java/com.github.ambry/config/PerformanceConfig.java
```diff
@@ -0,0 +1,121 @@
...
+  private static final String DEFAULT_REST_REQUEST_TOTAL_TIME_THRESHOLD_STR =
+      "{\"PUT\": \"Long.MAX_VALUE\",\"GET\": \"Long.MAX_VALUE\",\"POST\": \"Long.MAX_VALUE\",\"HEAD\": \"Long.MAX_VALUE\",\"DELETE\": \"Long.MAX_VALUE\"}";
+
+  /**
+   * The round trip time (millisecond) threshold for each type of rest request with non-success response(3xx or 4xx status code).
+   * The request is considered "Unsatisfied" if total time exceeds this threshold.
+   */
+  @Config(NON_SUCCESS_REST_REQUEST_TOTAL_TIME_THRESHOLD_STR)
+  @Default(DEFAULT_REST_REQUEST_TOTAL_TIME_THRESHOLD_STR)
+  private final String nonSuccessRestRequestTotalTimeThresholds;
```
cgtz: will we have separate thresholds for encrypted blob operations in the future?

What about things like getBlobInfo vs getBlob vs getSignedUrl. Or putBlob vs stitchBlob. I suppose we can start by setting such limits conservatively and then adding additional options in the future.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -91,6 +100,11 @@
   private final Queue<Chunk> chunksToWrite = new ConcurrentLinkedQueue<Chunk>();
   private final Queue<Chunk> chunksAwaitingCallback = new ConcurrentLinkedQueue<Chunk>();
   private final AtomicLong chunksToWriteCount = new AtomicLong(0);
+  // update endpoints in following two response sets when ResponseStatus is updated.
+  private final Set<ResponseStatus> successResponseSet =
```
cgtz: these should be static. Also, this is a little hard to maintain if we add additional response codes. Perhaps a better way to do it would be to add a status code field to each `ResponseStatus` enum value.

ex:
```java
enum ResponseStatus {
  Ok(200),
  ...
  ServiceUnavailable(503),
  InsufficientCapacity(507);

  private final int statusCode;

  ResponseStatus(int statusCode) {
    this.statusCode = statusCode;
  }

  public int getStatusCode() {
    return statusCode;
  }

  boolean isSuccess() {
    return statusCode >= 200 && statusCode < 300;
  }

  boolean isRedirection() {
    return statusCode >= 300 && statusCode < 400;
  }

  boolean isClientError() {
    return statusCode >= 400 && statusCode < 500;
  }

  boolean isServerError() {
    return statusCode >= 500;
  }
}
```

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -305,9 +321,58 @@ void close(Exception exception) {
...
+    long totalInboundBytes = request.getBytesReceived();
+    RestMethod method = request.getRestMethod();
+    Map<PerformanceIndex, Long> requestPerfToCheck = new HashMap<>();
+    responseStatus = errorResponseStatus == null ? responseStatus : errorResponseStatus;
+    boolean shouldSkipCheck = false;
+    switch (method) {
+      case GET:
+        if (successResponseSet.contains(responseStatus)) {
+          requestPerfToCheck.put(PerformanceIndex.TimeToFirstByte, timeToFirstBytesInMs);
+          requestPerfToCheck.put(PerformanceIndex.AverageBandwidth, totalOutboundBytes * 1000L / roundTripTimeInMs);
```
cgtz: this could potentially result in an exception if roundTripTimeInMs is zero since you are using integer division

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -305,9 +321,58 @@ void close(Exception exception) {
...
+      case PUT:
+      case HEAD:
+      case DELETE:
+        if (successResponseSet.contains(responseStatus) || nonSuccessResponseSet.contains(responseStatus)) {
+          requestPerfToCheck.put(PerformanceIndex.RoundTripTime, roundTripTimeInMs);
+        }
+        break;
+      default:
+        shouldSkipCheck = true;
+        logger.warn("Temporarily no evaluation criteria for " + method + " request. Mark as satisfied directly");
```
cgtz: this will log every time an OPTIONS call is made; let's make this debug level.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -305,9 +321,58 @@ void close(Exception exception) {
...
+    }
+    EnumMap<RestMethod, Threshold> thresholdsByMethod =
+        successResponseSet.contains(responseStatus) ? perfConfig.successRequestThresholds
+            : perfConfig.nonSuccessRequestThresholds;
+    if (!shouldSkipCheck && (requestPerfToCheck.isEmpty() || !thresholdsByMethod.get(method)
```
cgtz: So the request is unsatisfied if `requestPerfToCheck` is empty but shouldSkipCheck is false? Is this meant to account for the 5xx error case?

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
jsjtzyy: Fixed.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
jsjtzyy: Slightly changed to `roundTripTimeInMs == 0L ? Long.MAX_VALUE : totalOutboundBytes * 1000L / roundTripTimeInMs`

ambry-api/src/main/java/com.github.ambry/commons/Threshold.java
jsjtzyy: Thanks for both suggestions. Will use a separate `Criteria` class holding threshold value and `isUpperBound`.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
jsjtzyy: added

ambry-api/src/main/java/com.github.ambry/commons/PerformanceIndex.java
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.commons;
+
+public enum PerformanceIndex {
```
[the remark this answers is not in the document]
jsjtzyy: added

ambry-api/src/main/java/com.github.ambry/commons/Threshold.java
```diff
@@ -0,0 +1,66 @@
...
+   * Compare the measured value with predefined criteria.
+   * @param valueToCheck the measured value to check.
+   * @param criteria the predefined threshold and rule.
+   * @return {@code true} if measured value meets criteria. {@code false} otherwise.
+   */
+  private boolean compareWithCriteria(Long valueToCheck, Pair<Long, Boolean> criteria) {
+    if (valueToCheck == null) {
+      throw new IllegalArgumentException("Input value is empty which cannot be applied to threshold check.");
+    }
+    boolean shouldBelowThreshold = criteria.getSecond();
```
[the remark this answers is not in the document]
jsjtzyy: Use enum instead.

ambry-api/src/main/java/com.github.ambry/commons/Threshold.java
jsjtzyy: good point, refactored the class name in multiple files.

ambry-api/src/main/java/com.github.ambry/config/PerformanceConfig.java
jsjtzyy: This PR doesn't distinguish encrypted blob from non-encrypted one. The suggestion is good and I will raise a ticket to keep track of this. Maybe we can evaluate different types of requests in a more accurate way.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
jsjtzyy: Great idea. Changes made in `ResponseStatus`.

ambry-api/src/main/java/com.github.ambry/config/PerformanceConfig.java
```diff
@@ -0,0 +1,121 @@
...
+ * Configuration parameters required to help evaluate performance of Ambry(i.e latency, health).
+ */
+public class PerformanceConfig {
+
+  public static final String NON_SUCCESS_REST_REQUEST_TOTAL_TIME_THRESHOLD_STR =
+      "performance.non.success.rest.request.total.time.threshold";
+  public static final String SUCCESS_REST_REQUEST_TOTAL_TIME_THRESHOLD_STR =
+      "performance.success.rest.request.total.time.threshold";
+  private static final String DEFAULT_REST_REQUEST_TOTAL_TIME_THRESHOLD_STR =
+      "{\"PUT\": \"Long.MAX_VALUE\",\"GET\": \"Long.MAX_VALUE\",\"POST\": \"Long.MAX_VALUE\",\"HEAD\": \"Long.MAX_VALUE\",\"DELETE\": \"Long.MAX_VALUE\"}";
```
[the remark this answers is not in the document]
jsjtzyy: Fixed.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
jsjtzyy: Choose to use `responseStatus.isServerError()` to make it clearer.

ambry-api/src/main/java/com.github.ambry/commons/Criteria.java
```diff
@@ -0,0 +1,59 @@
...
+
+  /**
+   * Check if measured value meets the predefined requirement(threshold).
+   * @param value the measured value to check.
+   * @return {@code true} if measured value meets requirement. {@code false} otherwise.
+   * @throws NullPointerException when the input value is null.
+   */
+  boolean meetRequirement(Long value) throws NullPointerException {
+    Objects.requireNonNull(value, "Input value is empty which cannot be applied to threshold check.");
+    return boundType == BoundType.UpperBound ? value < threshold : value > threshold;
```
lightningrob: I would make it <= (allow equal case to meet requirement). Is there a specific reason to be strict about it?

jsjtzyy: hmm, maybe I should use `<=` and `>=` here. It doesn't need to be strict about it.

PR #1151, opened 2019-04-05
Add support for cluster name in request URI

cgtz:
To support multicluster deployments of ambry, this PR:
- adds support for including a path segment for the cluster name in the
  request URI.
- centralizes the request path parsing in
  AmbryBlobStorageService.preProcessAndRouteRequest. Instead of requiring
  string parsing in multiple places/classes, a RequestPath object as an
  internal request argument that can be referred to by components that
  need it.
- makes the checking of prefixes and operations in the path more robust
  to variations in path segments and separators.
- Also includes some fixes to make parsing comma separated list configs
  more convenient.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/FrontendConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/FrontendConfig.java
@@ -254,8 +254,6 @@ public class FrontendConfig {
    * @return the string with the leading and trailing slash remove.
    */
   private String stripLeadingAndTrailingSlash(String string) {
-    int startIndex = string.startsWith("/") ? 1 : 0;
-    int endIndex = Math.max(string.length() - (string.endsWith("/") ? 1 : 0), startIndex);
-    return string.substring(startIndex, endIndex);
+    return string.replaceAll("^/|/$", "");
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/FrontendConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/FrontendConfig.java
@@ -15,9 +15,11 @@ package com.github.ambry.config;
 
 import com.github.ambry.protocol.GetOption;
 import com.github.ambry.router.GetBlobOptions;
-import java.util.Arrays;
+import com.github.ambry.utils.Utils;
+import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 
 /**
@@ -101,6 +103,13 @@ public class FrontendConfig {
   @Config("frontend.path.prefixes.to.remove")
   @Default("")
   public final List<String> pathPrefixesToRemove;
+
+  /**
+   * The secure path to validate if required for certain container.
+   */
+  @Config("frontend.secure.path.prefix")
+  @Default("")
+  public final String securePathPrefix;
 
   /**
    * Specifies the blob size in bytes beyond which chunked response will be sent for a getBlob() call
@@ -192,13 +201,6 @@ public class FrontendConfig {
   @Default("20 * 1024 * 1024")
   public final long maxJsonRequestSizeBytes;
 
-  /**
-   * The secure path to validate if required for certain container.
-   */
-  @Config("frontend.secure.path.prefix")
-  @Default("")
-  public final String securePathPrefix;
-
   public FrontendConfig(VerifiableProperties verifiableProperties) {
     cacheValiditySeconds = verifiableProperties.getLong("frontend.cache.validity.seconds", 365 * 24 * 60 * 60);
     optionsValiditySeconds = verifiableProperties.getLong("frontend.options.validity.seconds", 24 * 60 * 60);
@@ -213,9 +215,13 @@ public class FrontendConfig {
     idSigningServiceFactory = verifiableProperties.getString(ID_SIGNING_SERVICE_FACTORY_KEY,
         "com.github.ambry.frontend.AmbryIdSigningServiceFactory");
     securePathPrefix = verifiableProperties.getString("frontend.secure.path.prefix", "");
-    pathPrefixesToRemove = Arrays.asList(
-        ((securePathPrefix.isEmpty() ? "" : "/" + securePathPrefix + ",") + verifiableProperties.getString(
-            "frontend.path.prefixes.to.remove", "")).split(","));
+    List<String> pathPrefixesFromConfig =
+        Utils.splitString(verifiableProperties.getString("frontend.path.prefixes.to.remove", ""), ",");
+    if (!securePathPrefix.isEmpty()) {
+      pathPrefixesFromConfig.add(securePathPrefix);
+    }
+    pathPrefixesToRemove = Collections.unmodifiableList(
+        pathPrefixesFromConfig.stream().map(this::stripLeadingAndTrailingSlash).collect(Collectors.toList()));
     chunkedGetResponseThresholdInBytes =
         verifiableProperties.getInt("frontend.chunked.get.response.threshold.in.bytes", 8192);
     allowServiceIdBasedPostRequest =
@@ -240,5 +246,16 @@ public class FrontendConfig {
         (int) TimeUnit.DAYS.toSeconds(30), 0, Integer.MAX_VALUE);
     maxJsonRequestSizeBytes =
         verifiableProperties.getLongInRange(MAX_JSON_REQUEST_SIZE_BYTES_KEY, 20 * 1024 * 1024, 0, Long.MAX_VALUE);
+  }
+
+  /**
+   * If the string starts or ends with a slash, remove them.
+   * @param string the string to strip.
+   * @return the string with the leading and trailing slash remove.
+   */
+  private String stripLeadingAndTrailingSlash(String string) {
+    int startIndex = string.startsWith("/") ? 1 : 0;
+    int endIndex = Math.max(string.length() - (string.endsWith("/") ? 1 : 0), startIndex);
+    return string.substring(startIndex, endIndex);
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
+++ b/ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
@@ -1,0 +1,192 @@
+/*
+ * Copyright 2019 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.rest;
+
+import java.util.List;
+
+
+public class RequestPath {
+  private final String prefix;
+  private final String clusterName;
+  private final String operationOrBlobId;
+  private final RestUtils.SubResource subResource;
+  private String operationOrBlobIdWithoutLeadingSlash = null;
+
+  /**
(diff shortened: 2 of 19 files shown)


ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
```diff
@@ -0,0 +1,192 @@
...
+   * For example, something like {@code /prefixOne/clusterA/parts/of/blobId/BlobInfo}, {@code /clusterA/operation},
+   * or {@code /blobId}, among others.
+   * @param restRequest {@link RestRequest} containing metadata about the request.
+   * @param prefixesToRemove the list of prefixes that could precede the other parts of the URL. Removal of
+   *                         prefixes earlier in the list will be preferred to removal of the ones later in the list.
+   * @param clusterName the cluster name to recognize and handle when parsing the URL. Case is ignored when matching
+   *                    this path segment.
+   * @return a {@link RequestPath} object.
+   */
+  public static RequestPath parse(RestRequest restRequest, List<String> prefixesToRemove, String clusterName) {
```
jsjtzyy: correct me if I am wrong,  I didn't see the query parameters (if there are any in the `restRequest`) are parsed from the uri, are they extracted somewhere else?

cgtz: The removal of the query parameters in this method wasn't actually needed since the `RestRequest` interface requires that `getPath()` will exclude any query parameters:
```java
  /**
   * Return the path (the parts of the URI after the domain excluding query parameters).
   * @return path String that represents part of the uri excluding domain and query parameters.
   */
  String getPath();
```
The `RestRequest` implementation should parse the query parameters during construction and include them in the map returned by `getArgs()`

ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
```diff
@@ -0,0 +1,192 @@
...
+
+    // if there are at least 2 path segments (*/*) after the current position,
+    // test if the last segment is a sub-resource
+    RestUtils.SubResource subResource = null;
+    int lastSlashOffset = path.lastIndexOf('/');
+    if (lastSlashOffset > offset) {
+      try {
+        subResource = RestUtils.SubResource.valueOf(path.substring(lastSlashOffset + 1));
+      } catch (IllegalArgumentException e) {
+        // nothing to do.
```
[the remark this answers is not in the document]
cgtz: This was taken from the `RestUtils::getBlobSubResource` method. We're actually using the exception for control flow here (which is not the best practice). If the last segment is something else (i.e. just a part of a request path but not a blob subresource), `valueOf` will throw `IllegalArgumentException`.

I will look for a cleaner way to do this. For example, I could loop through the values of the SubResource enum and check for equality.

ambry-frontend/src/main/java/com.github.ambry.frontend/AmbrySecurityService.java
```diff
@@ -178,9 +171,9 @@ public void processResponse(RestRequest restRequest, RestResponseChannel respons
             setHeadResponseHeaders(blobInfo, options, restRequest, responseChannel);
             break;
           case GET:
-            if (!operationOrBlobId.equals(Operations.GET_SIGNED_URL)) {
+            if (!requestPath.matchesOperation(Operations.GET_SIGNED_URL)) {
               responseChannel.setStatus(ResponseStatus.Ok);
-              RestUtils.SubResource subResource = RestUtils.getBlobSubResource(restRequest);
+              RestUtils.SubResource subResource = RestUtils.getRequestPath(restRequest).getSubResource();
```
jsjtzyy: `getRequestPath` is invoked in 3 places in this file. Two specified the package `RestUtils`, one not. Please make this consistent.

ambry-frontend/src/test/java/com.github.ambry.frontend/AmbryBlobStorageServiceTest.java
```diff
@@ -2807,7 +2805,12 @@ public void close() throws IOException {
   /**
    * Enumerates the different operation types in the router.
    */
-  enum OpType {DeleteBlob, GetBlob, PutBlob, StitchBlob, UpdateBlobTtl}
+  enum OpType {
+    DeleteBlob,
+    GetBlob,
+    PutBlob,
+    StitchBlob,
+    UpdateBlobTtl}
```
jsjtzyy: minor: format

ambry-frontend/src/test/java/com.github.ambry.frontend/PostBlobHandlerTest.java
```diff
@@ -502,4 +501,13 @@ private void stitchBlobAndVerify(byte[] requestBody, List<ChunkInfo> expectedSti
           errorChecker);
     }
   }
+
+  private RestRequest getRestRequest(JSONObject headers, String path, byte[] requestBody)
```
jsjtzyy: java doc, please

ambry-rest/src/main/java/com.github.ambry.rest/NettySslFactory.java
```diff
@@ -166,19 +166,17 @@ private static TrustManagerFactory getTrustManagerFactory(SSLConfig config)
...
   /**
    * @param config the {@link SSLConfig}.
    * @return the list of supported cipher suites, or {@code null} if the configs did not specify any
    */
   private static String[] getEnabledProtocols(SSLConfig config) {
-    String[] enabledProtocols = config.sslEnabledProtocols.split(",");
-    return enabledProtocols.length > 0 && !(enabledProtocols.length == 1 && enabledProtocols[0].isEmpty())
-        ? enabledProtocols : null;
+    List<String> enabledProtocols = Utils.splitString(config.sslEnabledProtocols, ",");
+    return !enabledProtocols.isEmpty() ? enabledProtocols.toArray(new String[0]) : null;
```
jsjtzyy: original code returns `enabledProtocols`, why here changed to `enabledProtocols.toArray(new String[0])`?

ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
```diff
@@ -0,0 +1,192 @@
...
+          offset = nextSegmentOffset;
+          break;
+        }
+      }
+    }
+
+    // check if the next path segment matches the cluster name.
+    String clusterNameFound = "";
+    if (clusterName != null) {
+      int nextSegmentOffset = matchPathSegments(path, offset, clusterName, true);
```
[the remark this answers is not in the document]
cgtz: confirmed

ambry-utils/src/main/java/com.github.ambry.utils/Utils.java
```diff
@@ -734,15 +734,15 @@ public static long addSecondsToEpochTime(long epochTimeInMs, long deltaTimeInSec
...
+   * @param data the string to split. If this string is empty, a zero element list will be returned.
+   * @param delimiter the delimiter for splitting.
+   * @return a mutable list of items.
    */
   public static ArrayList<String> splitString(String data, String delimiter) {
     if (data == null) {
       throw new IllegalArgumentException("Passed in string is null ");
     }
-    return new ArrayList<>(Arrays.asList(data.split(delimiter)));
+    return data.isEmpty() ? new ArrayList<>() : new ArrayList<>(Arrays.asList(data.split(delimiter)));
```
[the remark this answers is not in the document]
cgtz: Upon offline discussion, we have decided to filter out any empty segments that may come from splitting the string.

ambry-api/src/main/java/com.github.ambry/rest/RestUtils.java
```diff
@@ -307,8 +313,7 @@
      * "replicas" here means the string representation of all the replicas (i.e. host:port/path) where the blob might
      * reside.
      */
-    Replicas
-  }
+    Replicas}
```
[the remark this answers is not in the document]
cgtz: sorry, this is fixed after I upgraded to intellij 2019.1

ambry-commons/src/main/java/com.github.ambry.commons/JdkSslFactory.java
```diff
@@ -54,18 +54,10 @@ public JdkSslFactory(SSLConfig sslConfig) throws GeneralSecurityException, IOExc
     sslContext = createSSLContext(sslConfig);
 
     ArrayList<String> cipherSuitesList = Utils.splitString(sslConfig.sslCipherSuites, ",");
-    if (cipherSuitesList.size() > 0 && !(cipherSuitesList.size() == 1 && cipherSuitesList.get(0).isEmpty())) {
-      cipherSuites = cipherSuitesList.toArray(new String[cipherSuitesList.size()]);
-    } else {
-      cipherSuites = null;
-    }
+    cipherSuites = !cipherSuitesList.isEmpty() ? cipherSuitesList.toArray(new String[0]) : null;
```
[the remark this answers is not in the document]
cgtz: It was actually an intellij suggestion to change this. If you provide a zero length array, the method will use it to infer the type of array to create and set the size appropriately.

The intellij explanation was that in old versions of java this was a slow operation, but the jvm now has an optimization that makes it at least as fast as the alternative: https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size

ambry-frontend/src/main/java/com.github.ambry.frontend/AmbrySecurityService.java
cgtz: removed all `RestUtils` qualifiers from this class since we use a static import

ambry-rest/src/main/java/com.github.ambry.rest/NettySslFactory.java
cgtz: I was using `Array::split` before which meant that the content was already in an array. I switched to using the `Utils::splitString` convenience method which requires converting back to an array.

ambry-commons/src/main/java/com.github.ambry.commons/JdkSslFactory.java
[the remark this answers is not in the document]
jsjtzyy: Suggest adding comment here in case anyone rewrites this in the future due to misunderstanding.

ambry-api/src/main/java/com.github.ambry/config/FrontendConfig.java
```diff
@@ -241,4 +247,15 @@ public FrontendConfig(VerifiableProperties verifiableProperties) {
     maxJsonRequestSizeBytes =
         verifiableProperties.getLongInRange(MAX_JSON_REQUEST_SIZE_BYTES_KEY, 20 * 1024 * 1024, 0, Long.MAX_VALUE);
   }
+
+  /**
+   * If the string starts or ends with a slash, remove them.
+   * @param string the string to strip.
+   * @return the string with the leading and trailing slash remove.
+   */
+  private String stripLeadingAndTrailingSlash(String string) {
```
pnarayanan: Is this not exactly `string.replaceAll("^/|/$", "");`?

ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
```diff
@@ -39,7 +41,24 @@
...
+   * part of a {@link RestRequest}.
+   *
+   * @param path the path
+   * @param args a map containing any request arguments. This might be used to look for blob IDs supplied via query
+   *             parameters or headers.
+   * @param prefixesToRemove the list of prefixes that could precede the other parts of the URL. Removal of
+   *                         prefixes earlier in the list will be preferred to removal of the ones later in the list.
+   * @param clusterName the cluster name to recognize and handle when parsing the URL. Case is ignored when matching
+   *                    this path segment.
+   * @return
```
jsjtzyy: complete this comment?

ambry-api/src/main/java/com.github.ambry/config/FrontendConfig.java
cgtz: Yep, that's cleaner. Replaced it with the regex

cgtz:
@jsjtzyy @pnarayanan addressed comments

PR #1154, opened 2019-04-15
Combined cloud blob crypto service

lightningrob:
Combined changes from @dharju and @lightningrob.


ambry-api/src/main/java/com.github.ambry/config/CloudConfig.java
```diff
@@ -86,6 +96,34 @@
   @Default("false")
   public final boolean vcrRequireEncryption;
 
+  /**
+   * The kms service factory class name.
+   */
+  @Config(KMS_SERVICE_KEY_CONTEXT)
+  @Default(DEFAULT_KMS_SERVICE_KEY_CONTEXT)
+  public final String kmsServiceKeyContext;
```
zzmao: Prefix with `vcr`?

VCR_KMS_SERVICE_KEY_CONTEXT
DEFAULT_VCR_KMS_SERVICE_KEY_CONTEXT
vcrKmsServiceKeyContext
?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -100,20 +111,50 @@ public void put(MessageWriteSet messageSetToWrite) throws StoreException {
...
+            return;
+          }
+        }
+      } else {
+        // encryption not required, upload as is
       }
       CloudBlobMetadata blobMetadata =
           new CloudBlobMetadata(blobId, messageInfo.getOperationTimeMs(), messageInfo.getExpirationTimeInMs(),
-              messageInfo.getSize());
+              messageInfo.getSize(), encryptionOrigin, kmsContext);
```
zzmao: Is there a reason why not add encryption algorithm or cryptoAgent? 

If we are required to change encryption algorithm, there should be a way to know blobs already uploaded were with different algorithm. So I think either encryption algorithm or cryptoAgent should be added as metadata for backwards compatibility. It can be stored in payload or DB. 

I prefer to store agentFactory in payload, so that when a blob is being get, `CloudBlobStore` can create `deCryptoAgent` while parsing payload.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -43,30 +45,39 @@
...
-  CloudBlobStore(PartitionId partitionId, CloudConfig cloudConfig, CloudDestination cloudDestination) {
+  CloudBlobStore(PartitionId partitionId, CloudConfig cloudConfig, CloudDestination cloudDestination,
+      CloudBlobCryptoAgent cryptoAgent, VcrMetrics vcrMetrics) {
+    this.cloudDestination = Objects.requireNonNull(cloudDestination, "cloudDestination is required");
     this.partitionId = Objects.requireNonNull(partitionId, "partitionId is required");
     this.cloudConfig = Objects.requireNonNull(cloudConfig, "cloudConfig is required");
-    this.cloudDestination = Objects.requireNonNull(cloudDestination, "cloudDestination is required");
+    this.vcrMetrics = Objects.requireNonNull(vcrMetrics, "vcrMetrics is required");
     minTtlMillis = TimeUnit.DAYS.toMillis(cloudConfig.vcrMinTtlDays);
+    this.cryptoAgent = cryptoAgent;
```
zzmao: I feel it would be cleaner if we can generate cyrptoAgent based on config instead of from arguments. 

Looks like not straight forward but doable.

ambry-api/src/main/java/com.github.ambry/config/CloudConfig.java
lightningrob: ok

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
lightningrob: The problem is, we don't necessarily even know the encryption algorithm being used.  In some cases the implementation details are hidden.
Storing the agentFactory name in the payload is problematic since it's a variable length string (another size we need to add to header).  We could add it to metadata, but I don't see a great advantage.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
lightningrob: I don't follow you.  The VcrServer instantiates CloudBlobCryptoAgentFactory from config properties and passes it to CloudBlobManager/CloudBlobStore.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
lightningrob: I added cryptoAgentFactory to metadata.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
lightningrob: I see.  Yes, changed it to pass the agent factory to CloudBlobStore.

zzmao: Even factory, we can generate it inside `CloudBlobStore`, via `Utils.getObj()`.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
zzmao: cool.

ambry-api/src/main/java/com.github.ambry/config/CloudConfig.java
```diff
@@ -34,6 +37,13 @@
       "com.github.ambry.cloud.StaticVcrClusterFactory";
   private static final String DEFAULT_CLOUD_DESTINATION_FACTORY_CLASS =
       "com.github.ambry.cloud.azure.AzureCloudDestinationFactory";
+  public static final String KMS_SERVICE_KEY_CONTEXT = "kms.service.key.context";
+  public static final String DEFAULT_KMS_SERVICE_KEY_CONTEXT = "default-backup";
+  public static final String CLOUD_BLOB_CRYPTO_AGENT_FACTORY_CLASS = "cloud.blob.crypto.agent.factory.class";
+  public static final String DEFAULT_CLOUD_BLOB_CRYPTO_AGENT_FACTORY_CLASS =
+      "com.github.ambry.cloud.CloudBlobCryptoAgentFactoryImpl";
+  public static final String VCR_ENCRYPTION_FORMAT = "vcr.encryption.format";
```
zzmao: Looks like not used. Can we remove them?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -28,6 +28,9 @@
   public static final String FIELD_EXPIRATION_TIME = "expirationTime";
   public static final String FIELD_ACCOUNT_ID = "accountId";
   public static final String FIELD_CONTAINER_ID = "containerId";
+  public static final String FIELD_ENCRYPTION_ORIGIN = "encryptionOrigin";
```
zzmao: Is `initiator` better than `origin` in this context?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -383,6 +383,9 @@ private void updateErrorMetrics(Exception e) {
     map.put(CloudBlobMetadata.FIELD_EXPIRATION_TIME, String.valueOf(cloudBlobMetadata.getExpirationTime()));
     map.put(CloudBlobMetadata.FIELD_ACCOUNT_ID, String.valueOf(cloudBlobMetadata.getAccountId()));
     map.put(CloudBlobMetadata.FIELD_CONTAINER_ID, String.valueOf(cloudBlobMetadata.getContainerId()));
+    map.put(CloudBlobMetadata.FIELD_ENCRYPTION_ORIGIN, cloudBlobMetadata.getEncryptionOrigin().name());
+    map.put(CloudBlobMetadata.FIELD_VCR_KMS_CONTEXT, String.valueOf(cloudBlobMetadata.getVcrKmsContext()));
+    map.put(CloudBlobMetadata.FIELD_CRYPTO_AGENT_FACTORY, String.valueOf(cloudBlobMetadata.getCryptoAgentFactory()));
```
zzmao: Is it a problem if database corrupted?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -33,40 +34,53 @@
...
    * @param partitionId partition associated with BlobStore.
+   * @param cloudDestination the {@link CloudDestination}.
    * @param cloudConfig the {@link CloudConfig} to use.
    * @param cloudDestination the {@link CloudDestination} to use.
+   * @param cryptoAgentFactory the {@link CloudBlobCryptoAgentFactory} to use for encryption.
+   * @param vcrMetrics the {@link VcrMetrics} to use.
    */
-  CloudBlobStore(PartitionId partitionId, CloudConfig cloudConfig, CloudDestination cloudDestination) {
+  CloudBlobStore(PartitionId partitionId, CloudConfig cloudConfig, CloudDestination cloudDestination,
+      CloudBlobCryptoAgentFactory cryptoAgentFactory, VcrMetrics vcrMetrics) {
```
zzmao: Try not to pass in `cryptoAgentFactory` . Let's talk offline.

zzmao: Reasons:
1. CloudBackupManager and CloudBlobStore interfaces looks cleaner.
2. For CloudBlobStore.get(), we need to dynamically load `CloudBlobCryptoAgentFactory` anyway. 

So I think we can pass in `property` and `registry` instead of a `CloudBlobCryptoAgentFactory`.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
lightningrob: Either seems okay.  I prefer origin.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: Definitely, but then we won't be able to recover or do much else without reconstructing the DB from the blob metadata.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobCryptoAgentImpl.java
```diff
@@ -0,0 +1,189 @@
...
+  @Override
+  public ByteBuffer encrypt(ByteBuffer buffer) throws GeneralSecurityException {
+    Object key = kms.getRandomKey();
+    Object contextKey = kms.getKey(context);
+    ByteBuffer encryptedKey = cryptoService.encryptKey(key, contextKey);
+    ByteBuffer encryptedDataBuffer = cryptoService.encrypt(buffer, key);
+    ByteBuffer output = ByteBuffer.allocate(
+        EncryptedDataPayload.OVERHEAD_LENGTH + encryptedKey.array().length + encryptedDataBuffer.array().length);
+    EncryptedDataPayload.serialize(output, new EncryptedDataPayload(encryptedKey, encryptedDataBuffer));
+    return output;
```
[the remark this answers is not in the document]
lightningrob: I added it.  Need to test once the download part is implemented.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobCryptoAgentImpl.java
```diff
@@ -0,0 +1,189 @@
...
+      }
+      byte[] encryptedKey = new byte[encryptedKeySize];
+      byte[] encryptedData = new byte[encryptedDataSize];
+      try {
+        inputBuffer.get(encryptedKey);
+      } catch (Exception e) {
+        throw new GeneralSecurityException("Reading encrypted key from buffer", e);
+      }
+      try {
+        inputBuffer.get(encryptedData);
```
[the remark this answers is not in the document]
lightningrob: Wrapped all.

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrServer.java
```diff
@@ -146,11 +146,13 @@ public void startup() throws InstantiationException {
 
       StoreKeyConverterFactory storeKeyConverterFactory =
           Utils.getObj(serverConfig.serverStoreKeyConverterFactory, properties, registry);
-
+      CloudBlobCryptoAgentFactory cloudBlobCryptoAgentFactory =
```
zzmao: No longer used.

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrServer.java
```diff
@@ -146,11 +146,13 @@ public void startup() throws InstantiationException {
...
+          Utils.getObj(cloudConfig.cloudBlobCryptoAgentFactoryClass, properties, clusterMapConfig.clusterMapClusterName,
+              clusterMap.getMetricRegistry());
       cloudBackupManager =
-          new CloudBackupManager(cloudConfig, replicationConfig, clusterMapConfig, storeConfig, storeKeyFactory,
-              clusterMap, virtualReplicatorCluster, cloudDestinationFactory, scheduler, connectionPool, registry,
-              notificationSystem, storeKeyConverterFactory, serverConfig.serverMessageTransformer);
+          new CloudBackupManager(properties, cloudConfig, replicationConfig, clusterMapConfig, storeConfig,
+              storeKeyFactory, clusterMap, virtualReplicatorCluster, cloudDestinationFactory, scheduler, connectionPool,
+              registry, notificationSystem, storeKeyConverterFactory, serverConfig.serverMessageTransformer, metrics);
```
zzmao: Looks like metrics is not necessary here. Only  metrics.vcrStartTimeInMs used in this function, but it's a one time display. 

How about just remove vcrMetrics from this function? CloudBlobStore can create vcrMetrics based on given registry.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -38,6 +41,25 @@
...
+  public enum EncryptionOrigin {
+
+    /** No encryption */
+    NONE,
+    /** Encrypted by Router */
+    ROUTER,
+    /** Encrypted by VCR */
+    VCR,
+    /** Encrypted by Cloud Data Store */
+    CLOUD
```
zzmao: In prod, if we enable cloud encryption. How to notify a live VCR?

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
```diff
@@ -93,7 +96,8 @@ public void testNormalFlow() throws Exception {
         BlobDataType.DATACHUNK);
     InputStream inputStream = getBlobInputStream(blobSize);
     CloudBlobMetadata cloudBlobMetadata =
-        new CloudBlobMetadata(blobId, System.currentTimeMillis(), Utils.Infinite_Time, blobSize);
+        new CloudBlobMetadata(blobId, System.currentTimeMillis(), Utils.Infinite_Time, blobSize,
+            CloudBlobMetadata.EncryptionOrigin.NONE, vcrKmsContext, cryptoAgentFactory);
```
zzmao: Use '.VCR' ?

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
```diff
@@ -126,7 +130,9 @@ public void testBatchQuery() throws Exception {
               BlobDataType.DATACHUNK);
       blobIdList.add(blobId);
       InputStream inputStream = getBlobInputStream(blobSize);
-      CloudBlobMetadata cloudBlobMetadata = new CloudBlobMetadata(blobId, creationTime, Utils.Infinite_Time, blobSize);
+      CloudBlobMetadata cloudBlobMetadata =
+          new CloudBlobMetadata(blobId, creationTime, Utils.Infinite_Time, blobSize,
+              CloudBlobMetadata.EncryptionOrigin.NONE, vcrKmsContext, cryptoAgentFactory);
```
zzmao: Use '.VCR' ?

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrServer.java
lightningrob: Removed the two metrics.  Also creating metrics object in cloud backup manager to share across CloudBlobStores.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
lightningrob: CLOUD is here mainly for completeness, I'm not even sure when we would use it.

zzmao: Yes. I understand CLOUD is for completeness. 
I am thinking that if our VCRs are working in PROD and we are going to enable azure encryption. In this case, we need to shutdown all VCR and set `EncryptionOrigin` from NONE to CLOUD, and then enable azure encryption.

lightningrob: Per discussion, removed CLOUD for now and made clear that encryptionOrigin is only from Ambry's perspective.

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrMetrics.java
```diff
@@ -13,18 +13,34 @@
...
+  public final Timer blobEncryptionTime;
+  public final Timer blobDecryptionTime;
 
   public VcrMetrics(MetricRegistry registry) {
-    // TODO: add metrics to registry
-    vcrStartTimeInMs = registry.histogram(MetricRegistry.name(VcrServer.class, "VcrStartTimeInMs"));
-    vcrShutdownTimeInMs = registry.histogram(MetricRegistry.name(VcrServer.class, "VcrShutdownTimeInMs"));
+    this.registry = registry;
+    blobEncryptionCount = registry.counter(MetricRegistry.name(CloudBlobStore.class, "BlobEncryptionCount"));
+    blobDecryptionCount = registry.counter(MetricRegistry.name(CloudBlobStore.class, "BlobDet4acryptionCount"));
```
dharju: Typo? "BlobDet4acryptionCount"

PR #1157, opened 2019-04-16
Misc minor fixes in Helix tool, ReplicationEngine and Composite Manager

jsjtzyy:
1. Add null check in getDatacenterName() in CompositeClusterManager.
2. Make instance config change ignore count more accurate.
3. Ensure Helix bootstrap/upgrade tool can exit in the end.
4. Break the loop if remoteReplicaInfo is found in getRemoteReplicaInfo().

### DIFF
--- a/ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
+++ b/ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
@@ -1840,6 +1840,7 @@ public class ReplicationTest {
                 new RemoteReplicaInfo(peerReplicaId, replicaId, store, new MockFindToken(0, 0), Long.MAX_VALUE,
                     SystemTime.getInstance(), new Port(peerReplicaId.getDataNodeId().getPort(), PortType.PLAINTEXT));
             remoteReplicaInfos.add(remoteReplicaInfo);
+            break;
           }
         }
       }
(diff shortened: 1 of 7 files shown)


jsjtzyy:

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
```diff
@@ -1840,6 +1840,7 @@ void addMessage(PartitionId id, MessageInfo messageInfo, ByteBuffer buffer) {
                 new RemoteReplicaInfo(peerReplicaId, replicaId, store, new MockFindToken(0, 0), Long.MAX_VALUE,
                     SystemTime.getInstance(), new Port(peerReplicaId.getDataNodeId().getPort(), PortType.PLAINTEXT));
             remoteReplicaInfos.add(remoteReplicaInfo);
+            break;
```
zzmao: It's possible to have two replicas(belong to same partition) on same hosts. We can't assume there is only one, right?

jsjtzyy: I made the change based on the assumption that no two replicas belonging to same partition would reside on same node. Since this is only a test and we may allow two replicas on single node in the future, so I decide to remove the break here.

PR #1161, opened 2019-04-26
Introduce container metrics

cgtz:
Introduce request metrics scoped to individual containers for a subset
of the operations. These will be created lazily when a request comes for
a container to limit the number of metrics created for rarely used
containers.
Refactor RestRequestMetricsGroup to make it usable for metrics without
encrypted variants.
Remove completely unused legacy metrics interfaces.


zzmao:
Do we have the plan to have container metric whitelist ?

cgtz:
I am trying to avoid having a whitelist, but might have to revisit if there are significant performance issues. In closed-source code, I will build a configurable off switch for reporting to ingraphs, but try starting with the feature enabled first. I don't think the cost of reporting to JMX should be high, so I didn't put an offswitch in the metrics class.

ambry-api/src/main/java/com.github.ambry/frontend/ContainerMetrics.java
```diff
@@ -0,0 +1,121 @@
...
+        metricRegistry.histogram(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "RoundTripTimeInMs"));
+
+    successCount = metricRegistry.counter(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "SuccessCount"));
+    redirectionCount =
+        metricRegistry.counter(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "RedirectionCount"));
+    clientErrorCount =
+        metricRegistry.counter(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "ClientErrorCount"));
+    serverErrorCount =
+        metricRegistry.counter(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "ServerErrorCount"));
+
```
lightningrob: empty line but not category divider

ambry-api/src/main/java/com.github.ambry/frontend/ContainerMetrics.java
```diff
@@ -0,0 +1,121 @@
...
+        case Forbidden:
+          forbiddenCount.inc();
+          break;
+        case NotFound:
+          notFoundCount.inc();
+          break;
+        case Gone:
+          goneCount.inc();
+          break;
+      }
```
lightningrob: Default in case new statuses get added?

ambry-frontend/src/main/java/com.github.ambry.frontend/RestRequestMetricsGroup.java
```diff
@@ -0,0 +1,97 @@
...
+  private static final String SSL_SUFFIX = "Ssl";
+  private static final String ENCRYPTED_SUFFIX = "Encrypted";
+
+  private final ConcurrentMap<Pair<String, String>, ContainerMetrics> instantiatedContainerMetrics =
+      new ConcurrentHashMap<>();
+  private final String requestType;
+  private final boolean encryptedMetricsEnabled;
+  private final boolean containerMetricsEnabled;
+  private final MetricRegistry metricRegistry;
+  private final RestRequestMetrics blobMetrics;
```
lightningrob: Could call this blobPlainMetrics or blobPlaintextMetrics to make clear that it's not all blob requests.

ambry-frontend/src/main/java/com.github.ambry.frontend/RestRequestMetricsGroup.java
```diff
@@ -0,0 +1,97 @@
...
+    blobEncryptedMetrics =
+        encryptedMetricsEnabled ? new RestRequestMetrics(ownerClass, requestType + ENCRYPTED_SUFFIX, metricRegistry)
+            : null;
+    blobSslEncryptedMetrics =
+        encryptedMetricsEnabled ? new RestRequestMetrics(ownerClass, requestType + SSL_SUFFIX + ENCRYPTED_SUFFIX,
+            metricRegistry) : null;
+  }
+
+  /**
+   * Fetches the appropriate {@link RestRequestMetrics} based on the params
```
lightningrob: End comment lines with period.

ambry-api/src/main/java/com.github.ambry/frontend/ContainerMetrics.java
cgtz: This switch isn't inclusive of all error codes, only the ones that customers expressed interest in monitoring. So, I'm not sure if there would be anything that would go in a default block. The count of all the other 4xx error code can be computed with `clientErrorCount - badRequestCount - unauthorizedCount - forbiddenCount - notFoundCount - goneCount`

cgtz: Or do you think it's better if we get rid of the overall `clientErrorCount` and make a count for the default case, `otherClientErrorCount`

lightningrob: It seems fine as it is.  If nothing needs to be done in default block and it's not an error case, then it can be skipped.

ambry-api/src/main/java/com.github.ambry/frontend/ContainerMetrics.java
```diff
@@ -0,0 +1,121 @@
...
+  // 403
+  private final Counter forbiddenCount;
+  // 404
+  private final Counter notFoundCount;
+  // 410
+  private final Counter goneCount;
+
+  /**
+   * Metric names will be in the following format:
+   * {@code com.github.ambry.frontend.ContainerMetrics.{accountName}__{containerName}__{operationType}{metric}}
```
zzmao: Minor: only two _ here.

ambry-frontend/src/main/java/com.github.ambry.frontend/FrontendUtils.java
```diff
@@ -19,6 +19,7 @@
 import com.github.ambry.commons.CopyingAsyncWritableChannel;
 import com.github.ambry.rest.RestServiceErrorCode;
 import com.github.ambry.rest.RestServiceException;
+import com.github.ambry.rest.RestUtils;
```
zzmao: Not used.

gradle/buildscript.gradle
```diff
@@ -17,5 +17,5 @@ repositories {
 }
 
 dependencies {
-    classpath "gradle.plugin.nl.javadude.gradle.plugins:license-gradle-plugin:0.12.1"
```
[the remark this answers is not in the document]
cgtz: This upgrades the gradle plugin for checking that there is a license header for each java file. They changed the classpath of the plugin, so that's why this line changed.

PR #1162, opened 2019-04-29
Fixing the intermittent Helix tool test failure

jsjtzyy:
The replica size is changed in testEverything() method. If new replica
from initial partition is added to new node disk and is not reset when
calling writeBootstrapOrUpgrade(), then the second time it is invoked,
validating cluster manager in Helix tool would see different size of
replica of same partition and fails the test.
This patch ensures that if new replica is from initial partition, the
replica is only allowed to be added to original nodes in HardwareLayout,
which would be reset by writeBootstrapOrUpgrade() and avoid inconsistent
replica size error in validating cluster manager.

### DIFF
--- a/ambry-clustermap/src/test/java/com.github.ambry.clustermap/TestUtils.java
+++ b/ambry-clustermap/src/test/java/com.github.ambry.clustermap/TestUtils.java
@@ -679,7 +679,7 @@ public class TestUtils {
       for (Datacenter dcObj : hardwareLayout.getDatacenters()) {
         dataNodes.addAll(dcObj.getDataNodes());
       }
-      return dataNodes;
+      return Collections.unmodifiableList(dataNodes);
     }
 
     public Datacenter getRandomDatacenter() {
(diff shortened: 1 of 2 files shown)


ambry-clustermap/src/test/java/com.github.ambry.clustermap/TestUtils.java
```diff
@@ -671,6 +671,17 @@ public boolean isRackAware() {
...
 
+    /**
+     * @return all existing nodes in hardware layout when this method is being invoked.
+     */
+    public List<DataNode> getAllExistingDataNodes() {
+      List<DataNode> dataNodes = new ArrayList<>();
+      for (Datacenter dcObj : hardwareLayout.getDatacenters()) {
+        dataNodes.addAll(dcObj.getDataNodes());
+      }
+      return dataNodes;
```
pnarayanan: `Collections.unmodifiable`?

PR #1163, opened 2019-05-01
Avoid unnecessary round trips to Azure when uploading blobs

lightningrob:
Cache context key in crypto agent


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -205,6 +216,7 @@ public void delete(MessageWriteSet messageSetToDelete) throws StoreException {
   public void updateTtl(MessageWriteSet messageSetToUpdate) throws StoreException {
     checkStarted();
     // Note: we skipped uploading the blob on PUT record if the TTL was below threshold.
+    //
```
dharju: remove

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -219,6 +213,44 @@ public boolean uploadBlob(BlobId blobId, long blobSize, CloudBlobMetadata cloudB
...
+      logger.debug("Uploaded blob {} to Azure container {}.", blobId, azureContainer.getName());
+      return true;
+    } catch (StorageException sex) {
+      if (sex.getHttpStatusCode() == HttpURLConnection.HTTP_CONFLICT) {
+        // The blob already exists
+        logger.debug("Skipped upload of existing blob {}.", blobId);
+        azureMetrics.blobUploadSkippedCount.inc();
+        return false;
+      } else {
+        throw sex;
```
dharju: lol

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -343,10 +382,22 @@ private CloudBlobContainer getContainer(BlobId blobId, boolean autoCreate)
...
+  }
+
+  /**
+   * Utility method to ensure that the requested container exists in the storage account.
+   * @param azureContainer the container that must exist.
+   * @throws StorageException if the operation fails.
+   */
+  private void ensureCreated(CloudBlobContainer azureContainer) throws StorageException {
+    if (!knownContainers.contains(azureContainer.getName())) {
       azureContainer.createIfNotExists(BlobContainerPublicAccessType.CONTAINER, new BlobRequestOptions(),
```
dharju: So we're not worried about containers getting deleted at this point?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -69,6 +73,8 @@
   private final RequestOptions defaultRequestOptions = new RequestOptions();
   private final OperationContext blobOpContext = new OperationContext();
   private final AzureMetrics azureMetrics;
+  // Containers known to exist in the storage account
+  private final Set<String> knownContainers = Collections.synchronizedSet(new HashSet<>());
```
dharju: Create thread safe hash set from ConcurrentHashMap for better concurrency?

ConcurrentHashMap<String, Integer> map = new ConcurrentHashMap<>(); 
knownContainers = map.newKeySet();

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -219,6 +213,44 @@ public boolean uploadBlob(BlobId blobId, long blobSize, CloudBlobMetadata cloudB
...
+      CloudBlockBlob azureBlob = azureContainer.getBlockBlobReference(azureBlobName);
+      azureBlob.setMetadata(getMetadataMap(cloudBlobMetadata));
+      azureBlob.upload(blobInputStream, inputLength, condition, options, blobOpContext);
+      logger.debug("Uploaded blob {} to Azure container {}.", blobId, azureContainer.getName());
+      return true;
+    } catch (StorageException sex) {
+      if (sex.getHttpStatusCode() == HttpURLConnection.HTTP_CONFLICT) {
+        // The blob already exists
+        logger.debug("Skipped upload of existing blob {}.", blobId);
+        azureMetrics.blobUploadSkippedCount.inc();
```
zzmao: Let's rename it to `blobUploadConflictCount` because there is a same name metric in vcrMetrics.blobUploadSkippedCount. They indicate different things.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: TBH I hadn't considered that, but thinking about it now, if someone deletes the container it will likely be part of wiping out the backup (at least for that partition) and starting over, in which case the VCR should be restarted.  And if the container is deleted accidentally or maliciously, IMO it will be better for the VCR to start throwing exceptions and up the error count rather than silently recreating the container and marching on.  So I prefer to keep the code as it is.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: Can you actually modify a key set that is backed by a map?

lightningrob: Never mind, you said newKeySet(), not keySet().

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: Done

zzmao:
LGTM. Restarted build&test.

PR #1164, opened 2019-05-01
Make CompositeClusterManager more robust to NPEs and other errors

cgtz:
- Handle equality checks correctly for methods that can return null.
- Catch checked exceptions that HelixClusterManager may throw. This
  ensures that StaticClusterManager acts as the source of truth and
  failures in HelixClusterManager will not affect
  CompositeClusterManager's operation.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/CompositeClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/CompositeClusterManager.java
@@ -45,7 +45,8 @@ class CompositeClusterManager implements ClusterMap {
    * @param helixClusterManager the {@link HelixClusterManager} instance to use for comparison of views.
    */
   CompositeClusterManager(StaticClusterManager staticClusterManager, HelixClusterManager helixClusterManager) {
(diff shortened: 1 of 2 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/CompositeClusterManager.java
```diff
@@ -31,6 +34,7 @@
  * and reports inconsistencies in the views from the two underlying cluster managers.
  */
 class CompositeClusterManager implements ClusterMap {
+  private static final Logger LOGGER = LoggerFactory.getLogger(CompositeClusterManager.class);
   final StaticClusterManager staticClusterManager;
   final HelixClusterManager helixClusterManager;
   final HelixClusterManagerMetrics helixClusterManagerMetrics;
```
jsjtzyy: Let's add strict check for both `helixClusterManager` and `helixClusterManagerMetrics`.  That is, check they are `null` in all places where dereference is invoked.

cgtz: After fixing the constructor, all places where these are used are surrounded by a null check.

Issue #1167, opened 2019-05-05
NullPointerException when error message of captured exception is null


jsjtzyy:
#1025 is merged with aim to catch disk I/O error. However, comparing error message with `IO_ERROR_STR` defined in `StoreException` is not a good way because error message could be null for IOException. Following is two examples from server log:
```
2019/05/03 00:00:00.130 ERROR [ReplicaThread] [Replica Thread-Inter0**-***] [ambry-server] [] Remote node: DataNode[***-****.***.linkedin.com:****] Thread name: Replica Thread-Inter0**-*** Remote replica: Replica[***-****.***.linkedin.com:****:/mnt/u010/ambrydata/43]
java.lang.NullPointerException
```
```
2019/05/04 00:11:34.476 ERROR [DiskManager] [store-startup-Partition[165]] [ambry-server] [] Exception while starting store for the partitionPartition[165]
com.github.ambry.store.StoreException: Error while starting store for dir /mnt/u010/ambrydata/165
        at com.github.ambry.store.BlobStore.start(BlobStore.java:233)
        at com.github.ambry.store.DiskManager.lambda$start$17(DiskManager.java:125)
        at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.NullPointerException
        at com.github.ambry.store.LogSegment.<init>(LogSegment.java:139)
        at com.github.ambry.store.Log.getSegmentsToLoad(Log.java:342)
        at com.github.ambry.store.Log.<init>(Log.java:78)
        at com.github.ambry.store.BlobStore.start(BlobStore.java:209)
        ... 2 more
```
We should first address NPE problem in this case and then figure out a good way to determine if IOException is related to Disk issue.

jsjtzyy:
addressed in #1168

PR #1168, opened 2019-05-06
Fix NullPointerException and 'no more capacity' issues in blob store

jsjtzyy:
1. NPE occurs when error message of captured IOException is null. In
this case, we should use Objects.equals to compare error string.
2. 'no more capacity' issue occurs when new log segment creation failed,
however, the code didn't restore the counter of remainingUnallocatedSegments.
This patch will catch such failure and restore counter as well as clean
up the allocated file.

This pr will address #1167


zzmao:
Thanks for fixing it!

PR #1170, opened 2019-05-09
Minor fixes in IOException catch and metrics registry

jsjtzyy:
1. When log segment instantiation encounters IOException, no matter
freeing log segment succeeds or not, ensure that store exception is
correctly thrown in ensureCapacity method.
2. Register the custom percentile metrics during RouterMetrics
initialization rather than register them every time adaptive tracker is
created. This will avoid unnecessary overhead.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/store/StoreException.java
+++ b/ambry-api/src/main/java/com.github.ambry/store/StoreException.java
@@ -13,6 +13,9 @@
  */
 package com.github.ambry.store;
 
+import java.util.Objects;
+
+
 public class StoreException extends Exception {
   public static final String INTERNAL_ERROR_STR =
       "a fault occurred in a recent unsafe memory access operation in compiled Java code";
@@ -37,5 +40,15 @@ public class StoreException extends Exception {
 
   public StoreErrorCodes getErrorCode() {
     return this.error;
+  }
+
+  /**
+   * Resolve the error code from {@link Throwable}. This is to determine if exception is related to real disk I/O issue.
+   * @param t the {@link Throwable} to check
+   * @return the {@link StoreErrorCodes} based on the error message in exception/error.
+   */
+  public static StoreErrorCodes resolveErrorCode(Throwable t) {
+    return Objects.equals(t.getMessage(), StoreException.IO_ERROR_STR) || Objects.equals(t.getMessage(),
+        StoreException.INTERNAL_ERROR_STR) ? StoreErrorCodes.IOError : StoreErrorCodes.Unknown_Error;
   }
 }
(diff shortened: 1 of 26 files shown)


ambry-router/src/test/java/com.github.ambry.router/CryptoJobHandlerTest.java
```diff
@@ -69,7 +69,7 @@ public CryptoJobHandlerTest() throws IOException, GeneralSecurityException {
     cryptoService = new GCMCryptoServiceFactory(verifiableProperties, REGISTRY).getCryptoService();
     cryptoJobHandler = new CryptoJobHandler(DEFAULT_THREAD_COUNT);
     referenceClusterMap = new MockClusterMap();
-    routerMetrics = new NonBlockingRouterMetrics(referenceClusterMap);
+    routerMetrics = new NonBlockingRouterMetrics(referenceClusterMap, null);
```
jsjtzyy: I purposely set  `RouterConfig`  = `null` when creating `NonBlockingRouterMetrics` in several tests. This is to verify that even if router config is mistakenly set to null in the ctor of `NonBlockingRouterMetrics`, there would be no failure during instantiation.

ambry-store/src/main/java/com.github.ambry.store/Log.java
```diff
@@ -470,17 +466,15 @@ private void ensureCapacity(long writeSize) throws StoreException {
           new LogSegment(segmentNameAndFilename.getFirst(), newSegmentFile, segmentCapacity, metrics, true);
       segmentsByName.put(segmentNameAndFilename.getFirst(), newSegment);
     } catch (StoreException e) {
-      logger.error("Failed to create new log segment {} with store exception: ", segmentNameAndFilename.getFirst(), e);
       try {
```
cgtz: a test for this block will be good, but its ok if its too cumbersome.

ambry-api/src/main/java/com.github.ambry/store/StoreException.java
```diff
@@ -13,6 +13,9 @@
  */
 package com.github.ambry.store;
 
+import java.util.Objects;
+
+
 public class StoreException extends Exception {
   public static final String INTERNAL_ERROR_STR =
```
cgtz: Can these be private or package-private

ambry-store/src/main/java/com.github.ambry.store/Log.java
jsjtzyy: test added

ambry-api/src/main/java/com.github.ambry/store/StoreException.java
jsjtzyy: changed it to package-private

jsjtzyy:
./gradlew build && ./gradlew test succeeded. @cgtz @zzmao  reminder to review.

PR #1171, opened 2019-05-10
Introduce HelixVcrCluster to CloudBackupManager and refactor ReplicationEngine and ReplicaThread.

zzmao:
Introduce HelixVcrCluster to CloudBackupManager.
Refactor ReplicationEngine and ReplicaThread.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/VirtualReplicatorCluster.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/VirtualReplicatorCluster.java
@@ -14,6 +14,7 @@
 package com.github.ambry.clustermap;
 
 import java.util.List;
+import org.apache.helix.InstanceType;
 
 
 /**
@@ -33,6 +34,12 @@ public interface VirtualReplicatorCluster extends AutoCloseable {
    * @return DataNodeId.
    */
   DataNodeId getCurrentDataNodeId();
+
+  /**
+   * Join the cluster with a role.
+   * @param role the role
+   */
+  void participate(InstanceType role) throws Exception;
 
   /**
    * Gets all {@link PartitionId}s assigned to current node.
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/VirtualReplicatorCluster.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/VirtualReplicatorCluster.java
@@ -36,8 +36,8 @@ public interface VirtualReplicatorCluster extends AutoCloseable {
   DataNodeId getCurrentDataNodeId();
 
   /**
-   * Join the cluster with a role.
-   * @param role the role
+   * Join the cluster as {@link InstanceType#PARTICIPANT} or {@link InstanceType#SPECTATOR} .
+   * @param role the {@link InstanceType}.
    */
   void participate(InstanceType role) throws Exception;
 
--- a/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
@@ -107,7 +107,11 @@ public class ReplicationConfig {
   public final boolean replicationIncludeAll;
 
   /**
-   * If true, replication token will be persisted explicitly.
+   * If true, replication token will be persisted when shutdown or replica remote.
+   * For CloudBackupManger and ReplicationManager, token is persisted when shutdown if this config is true.
+   * For CloudBackupManger, where replica may be removed, token is also persisted if this config is true.
+   * ReplicationManager doesn't support replica remove now.
+   * This is used for test only as of now.
    */
   @Config("replication.persist.token.on.shutdown.or.replica.remove")
   @Default("true")
--- a/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
@@ -106,6 +106,13 @@ public class ReplicationConfig {
   @Default("true")
   public final boolean replicationIncludeAll;
 
+  /**
+   * If true, replication token will be persisted explicitly.
+   */
+  @Config("replication.persist.token.on.shutdown.or.replica.remove")
+  @Default("true")
+  public final boolean replicationPersistTokenOnShutdownOrReplicaRemove;
+
   public ReplicationConfig(VerifiableProperties verifiableProperties) {
 
     replicationTokenFactory =
@@ -133,5 +140,7 @@ public class ReplicationConfig {
     replicationFetchSizeInBytes =
         verifiableProperties.getLongInRange("replication.fetch.size.in.bytes", 1048576, 1, Long.MAX_VALUE);
     replicationIncludeAll = verifiableProperties.getBoolean("replication.include.all", true);
+    replicationPersistTokenOnShutdownOrReplicaRemove =
+        verifiableProperties.getBoolean("replication.persist.token.on.shutdown.or.replica.remove", true);
   }
 }
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
@@ -70,38 +70,9 @@ public class CloudBackupManager extends ReplicationEngine {
     this.virtualReplicatorCluster = virtualReplicatorCluster;
     this.vcrMetrics = new VcrMetrics(metricRegistry);
     this.cloudDestination = cloudDestinationFactory.getCloudDestination();
-    persistor =
+    this.persistor =
         new CloudTokenPersistor(replicaTokenFileName, mountPathToPartitionInfoList, replicationMetrics, clusterMap,
             factory, cloudDestination);
-    try {
-      virtualReplicatorCluster.participate(InstanceType.PARTICIPANT);
-    } catch (Exception e) {
-      throw new ReplicationException("Cluster participate failed.", e);
-    }
-    // Some sleep time can reduce listener message.
-    List<? extends PartitionId> partitionIds = virtualReplicatorCluster.getAssignedPartitionIds();
-    for (PartitionId partitionId : partitionIds) {
-      addPartition(partitionId);
-    }
-
-    // Add listener for new coming assigned partition
-    virtualReplicatorCluster.addListener(new VirtualReplicatorClusterListener() {
-      @Override
-      public void onPartitionAdded(PartitionId partitionId) {
-        try {
-          addPartition(partitionId);
-        } catch (ReplicationException e) {
-          logger.error("Exception on adding partition: ", e);
-        }
-      }
-
-      @Override
-      public void onPartitionRemoved(PartitionId partitionId) {
-        removeRemoteReplicaInfoFromReplicaThread(partitionToPartitionInfo.get(partitionId).getRemoteReplicaInfos());
-      }
-    });
-
-    replicationMetrics.populatePerColoMetrics(replicaThreadPoolByDc.keySet());
   }
 
   /**
@@ -174,11 +145,32 @@ public class CloudBackupManager extends ReplicationEngine {
 
   @Override
   public void start() throws ReplicationException {
+    // Add listener for new coming assigned partition
+    virtualReplicatorCluster.addListener(new VirtualReplicatorClusterListener() {
+      @Override
+      public void onPartitionAdded(PartitionId partitionId) {
+        try {
+          addPartition(partitionId);
+        } catch (ReplicationException e) {
+          logger.error("Exception on adding partition: ", e);
+        }
+      }
+
+      @Override
+      public void onPartitionRemoved(PartitionId partitionId) {
+        removeRemoteReplicaInfoFromReplicaThread(partitionToPartitionInfo.get(partitionId).getRemoteReplicaInfos());
+      }
+    });
+
+    try {
+      virtualReplicatorCluster.participate(InstanceType.PARTICIPANT);
+    } catch (Exception e) {
+      throw new ReplicationException("Cluster participate failed.", e);
+    }
+
     // start background persistent thread
     // start scheduler thread to persist index in the background
-    if (persistor != null) {
-      this.scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
-          replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
-    }
+    this.scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
+        replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
   }
 }
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
@@ -36,6 +36,7 @@ import com.github.ambry.store.StoreException;
 import com.github.ambry.store.StoreKeyConverterFactory;
 import com.github.ambry.store.StoreKeyFactory;
 import com.github.ambry.utils.SystemTime;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -73,6 +74,43 @@ public class CloudBackupManager extends ReplicationEngine {
     this.persistor =
         new CloudTokenPersistor(replicaTokenFileName, mountPathToPartitionInfoList, replicationMetrics, clusterMap,
             factory, cloudDestination);
+  }
+
+  @Override
+  public void start() throws ReplicationException {
+    // Add listener for new coming assigned partition
+    virtualReplicatorCluster.addListener(new VirtualReplicatorClusterListener() {
+      @Override
+      public void onPartitionAdded(PartitionId partitionId) {
+        try {
+          addPartition(partitionId);
+        } catch (ReplicationException e) {
+          logger.error("Exception on adding partition{}: ", partitionId, e);
+        }
+      }
+
+      @Override
+      public void onPartitionRemoved(PartitionId partitionId) {
+        PartitionInfo partitionInfo = partitionToPartitionInfo.get(partitionId);
+        removeRemoteReplicaInfoFromReplicaThread(partitionInfo.getRemoteReplicaInfos());
+        try {
+          persistor.write(partitionInfo.getLocalReplicaId().getMountPath(), false);
+        } catch (IOException | ReplicationException e) {
+          logger.error("Exception on token write when remove partition{}: ", partitionId, e);
+        }
+      }
+    });
+
+    try {
+      virtualReplicatorCluster.participate(InstanceType.PARTICIPANT);
+    } catch (Exception e) {
+      throw new ReplicationException("Cluster participate failed.", e);
+    }
+
+    // start background persistent thread
+    // start scheduler thread to persist index in the background
+    this.scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
+        replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
   }
 
   /**
@@ -121,56 +159,34 @@ public class CloudBackupManager extends ReplicationEngine {
     } else {
       throw new ReplicationException("Failed to add partition " + partitionId + ", because no peer replicas found.");
     }
-    // read replication token
-    for (RemoteReplicaInfo remoteReplicaInfo : remoteReplicaInfos) {
-      List<RemoteReplicaInfo.ReplicaTokenInfo> tokenInfos = persistor.retrieve(cloudReplica.getMountPath());
-      for (RemoteReplicaInfo.ReplicaTokenInfo tokenInfo : tokenInfos) {
-        DataNodeId dataNodeId = remoteReplicaInfo.getReplicaId().getDataNodeId();
-        if (dataNodeId.getHostname().equalsIgnoreCase(tokenInfo.getHostname())
-            && dataNodeId.getPort() == tokenInfo.getPort() && remoteReplicaInfo.getReplicaId()
-            .getReplicaPath()
-            .equals(tokenInfo.getReplicaPath())) {
-          logger.info("Read token for partition {} remote host {} port {} token {}", partitionId,
-              tokenInfo.getHostname(), tokenInfo.getPort(), tokenInfo.getReplicaToken());
-          remoteReplicaInfo.initializeTokens(tokenInfo.getReplicaToken());
-          remoteReplicaInfo.setTotalBytesReadFromLocalStore(tokenInfo.getTotalBytesReadFromLocalStore());
-          break;
+    // Reload replication token if exist.
+    List<RemoteReplicaInfo.ReplicaTokenInfo> tokenInfos = persistor.retrieve(cloudReplica.getMountPath());
+    if (tokenInfos.size() != 0) {
+      for (RemoteReplicaInfo remoteReplicaInfo : remoteReplicaInfos) {
+        boolean tokenReloaded = false;
+        for (RemoteReplicaInfo.ReplicaTokenInfo tokenInfo : tokenInfos) {
+          DataNodeId dataNodeId = remoteReplicaInfo.getReplicaId().getDataNodeId();
+          if (dataNodeId.getHostname().equalsIgnoreCase(tokenInfo.getHostname())
+              && dataNodeId.getPort() == tokenInfo.getPort() && remoteReplicaInfo.getReplicaId()
+              .getReplicaPath()
+              .equals(tokenInfo.getReplicaPath())) {
+            logger.info("Read token for partition {} remote host {} port {} token {}", partitionId,
+                tokenInfo.getHostname(), tokenInfo.getPort(), tokenInfo.getReplicaToken());
+            tokenReloaded = true;
+            remoteReplicaInfo.initializeTokens(tokenInfo.getReplicaToken());
+            remoteReplicaInfo.setTotalBytesReadFromLocalStore(tokenInfo.getTotalBytesReadFromLocalStore());
+            break;
+          }
+        }
+        if (tokenReloaded == false) {
+          logger.warn("Token reload failed. remoteReplicaInfo: {} tokenInfos: {}", remoteReplicaInfo, tokenInfos);
         }
       }
     }
-    // create thread if necessary.
+    // Add remoteReplicaInfos to {@link ReplicaThread}.
     addRemoteReplicaInfoToReplicaThread(remoteReplicaInfos, true);
     return true;
   }
-
-  @Override
-  public void start() throws ReplicationException {
-    // Add listener for new coming assigned partition
-    virtualReplicatorCluster.addListener(new VirtualReplicatorClusterListener() {
-      @Override
-      public void onPartitionAdded(PartitionId partitionId) {
-        try {
-          addPartition(partitionId);
-        } catch (ReplicationException e) {
-          logger.error("Exception on adding partition: ", e);
-        }
-      }
-
-      @Override
-      public void onPartitionRemoved(PartitionId partitionId) {
-        removeRemoteReplicaInfoFromReplicaThread(partitionToPartitionInfo.get(partitionId).getRemoteReplicaInfos());
-      }
-    });
-
-    try {
-      virtualReplicatorCluster.participate(InstanceType.PARTICIPANT);
-    } catch (Exception e) {
-      throw new ReplicationException("Cluster participate failed.", e);
-    }
-
-    // start background persistent thread
-    // start scheduler thread to persist index in the background
-    this.scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
-        replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
-  }
 }
+
+
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
@@ -71,16 +71,15 @@ public class CloudBackupManager extends ReplicationEngine {
     this.vcrMetrics = new VcrMetrics(metricRegistry);
     this.cloudDestination = cloudDestinationFactory.getCloudDestination();
     persistor =
-        new CloudTokenPersistor(replicaTokenFileName, partitionGroupedByMountPath, replicationMetrics, clusterMap,
+        new CloudTokenPersistor(replicaTokenFileName, mountPathToPartitionInfoList, replicationMetrics, clusterMap,
             factory, cloudDestination);
     try {
       virtualReplicatorCluster.participate(InstanceType.PARTICIPANT);
     } catch (Exception e) {
       throw new ReplicationException("Cluster participate failed.", e);
     }
-    // Some sleep time
+    // Some sleep time can reduce listener message.
     List<? extends PartitionId> partitionIds = virtualReplicatorCluster.getAssignedPartitionIds();
-    System.out.println("pre add partition" + partitionIds);
     for (PartitionId partitionId : partitionIds) {
       addPartition(partitionId);
     }
@@ -92,21 +91,25 @@ public class CloudBackupManager extends ReplicationEngine {
         try {
           addPartition(partitionId);
         } catch (ReplicationException e) {
-          System.out.println("Exception on add: " + e);
+          logger.error("Exception on adding partition: ", e);
         }
       }
 
       @Override
       public void onPartitionRemoved(PartitionId partitionId) {
-
+        removeRemoteReplicaInfoFromReplicaThread(partitionToPartitionInfo.get(partitionId).getRemoteReplicaInfos());
       }
     });
 
     replicationMetrics.populatePerColoMetrics(replicaThreadPoolByDc.keySet());
   }
 
-  synchronized boolean addPartition(PartitionId partitionId) throws ReplicationException {
-    if (partitionsToReplicate.get(partitionId) != null) {
+  /**
+   * Add given {@link PartitionId} and its {@link RemoteReplicaInfo}s to backup list.
+   * @param partitionId the {@link PartitionId} to add.
+   */
+  boolean addPartition(PartitionId partitionId) throws ReplicationException {
+    if (partitionToPartitionInfo.get(partitionId) != null) {
       return false;
     }
     ReplicaId cloudReplica =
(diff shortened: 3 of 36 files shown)


zzmao:
@pnarayanan @cgtz  If there isn't enough time for you to review, you can focus on ReplicationEngine and ReplicaThread refactor.

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
```diff
@@ -1083,21 +1164,21 @@ public void replicaThreadTest() throws Exception {
...
     missingKeys = new int[replicasToReplicate.get(remoteHost.dataNodeId).size()];
     assertMissingKeys(missingKeys, batchSize, replicaThread, remoteHost, replicasToReplicate);
 
     // 4. add more put messages and verify that replication continues and is throttled appropriately.
     addPutMessagesToReplicasOfPartition(partitionId, Arrays.asList(localHost, remoteHost), 3);
     currentTimeMs = time.milliseconds();
-    replicaThread.replicate(new ArrayList<>(replicasToReplicate.values()));
+    replicaThread.replicate();
     assertMissingKeys(missingKeys, batchSize, replicaThread, remoteHost, replicasToReplicate);
     assertEquals("Replica thread should sleep exactly " + expectedThrottleDurationMs + " since remote has new token",
```
lightningrob: It's generally not a good idea to rely on times being exact, especially to the ms.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +64,121 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+    // create thread if necessary.
+    addRemoteReplicaInfoToReplicaThread(remoteReplicaInfos, true);
+    return true;
+  }
+
+  @Override
+  public void start() throws ReplicationException {
+    // start background persistent thread
+    // start scheduler thread to persist index in the background
+    if (persistor != null) {
```
lightningrob: It's never null any more.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +64,121 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+    this.properties = properties;
+    this.cloudConfig = cloudConfig;
+    this.storeConfig = storeConfig;
+    this.virtualReplicatorCluster = virtualReplicatorCluster;
+    this.vcrMetrics = new VcrMetrics(metricRegistry);
+    this.cloudDestination = cloudDestinationFactory.getCloudDestination();
+    persistor =
+        new CloudTokenPersistor(replicaTokenFileName, mountPathToPartitionInfoList, replicationMetrics, clusterMap,
+            factory, cloudDestination);
+    try {
```
lightningrob: It feels like most of this logic should be in start() rather than constructor.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -229,20 +219,51 @@ public void run() {
...
   /**
-   * Replicas from the given replicas
-   * @param replicasToReplicate list of {@link RemoteReplicaInfo} by data node
+   * Do a lazy removal of {@link RemoteReplicaInfo} from current {@link ReplicaThread}.
+   * @param remoteReplicaInfo {@link RemoteReplicaInfo} to remove.
+   */
+  void removeRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInfo) {
+    lock.lock();
+    replicaLazyRemoveSet.add(remoteReplicaInfo);
+    lock.unlock();
```
lightningrob: Every call to unlock needs to be in a finally block.
Even better, if it's a simple mutex lock, can switch to using synchronized(someLockObject) which handles unlocking automatically.

ambry-protocol/src/main/java/com.github.ambry.protocol/GetRequest.java
```diff
@@ -40,6 +40,7 @@
   private static final int Partition_Request_Info_List_Size = 4;
   private static final short Get_Request_Version_V2 = 2;
   public static final String Replication_Client_Id_Prefix = "replication-fetch-";
+  public static final String Cloud_Replica_Keyword = "vcr";
```
lightningrob: Does this need to be in this class?  Seems like it's only used in cloud package.

ambry-replication/src/test/java/com.github.ambry.replication/ReplicationTest.java
zzmao: `MockTime` is used.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
zzmao: Fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
zzmao: addressed.

ambry-protocol/src/main/java/com.github.ambry.protocol/GetRequest.java
zzmao: It's also used in `getRemoteReplicaInfo`, otherwise it reports error: ```ReplicaMetaDataRequest from unknown Replica {}```.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
zzmao: There is a critical section in `replicate`. It's not suitable to use `synchronized ` for its case.

`finally` is used.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +64,113 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+        retList = (value == null ? new ArrayList<>() : value);
+        retList.add(partitionInfo);
+        return retList;
+      });
+    } else {
+      throw new ReplicationException("Failed to add partition " + partitionId + ", because no peer replicas found.");
+    }
+    // read replication token
+    for (RemoteReplicaInfo remoteReplicaInfo : remoteReplicaInfos) {
+      List<RemoteReplicaInfo.ReplicaTokenInfo> tokenInfos = persistor.retrieve(cloudReplica.getMountPath());
```
lightningrob: Persistor method should be called once per partition.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +64,113 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
-    persistor =
-        new CloudTokenPersistor(replicaTokenFileName, partitionGroupedByMountPath, replicationMetrics, clusterMap,
-            factory, cloudDestination);
+    // create thread if necessary.
+    addRemoteReplicaInfoToReplicaThread(remoteReplicaInfos, true);
+    return true;
+  }
+
+  @Override
+  public void start() throws ReplicationException {
```
lightningrob: Minor: move this method above addPartition()

ambry-cloud/src/main/java/com.github.ambry.cloud/HelixVcrCluster.java
```diff
@@ -142,6 +136,17 @@ public DataNodeId getCurrentDataNodeId() {
     return currentDataNode;
   }
 
+  @Override
+  public void participate(InstanceType role) throws Exception {
+    manager = HelixManagerFactory.getZKHelixManager(vcrClusterName, vcrInstanceName, InstanceType.PARTICIPANT,
```
lightningrob: use role

ambry-cloud/src/main/java/com.github.ambry.cloud/HelixVcrCluster.java
```diff
@@ -99,7 +91,8 @@ public void addPartition(String partitionIdStr) {
         for (VirtualReplicatorClusterListener listener : listeners) {
           listener.onPartitionAdded(partitionId);
         }
-        logger.info("Partition {} is added to current VCR: {}", partitionIdStr, vcrInstanceName);
+        logger.info("Partition {} is added to current VCR: {}. Assigned partitions: {}", partitionIdStr,
+            vcrInstanceName, assignedPartitionIds);
```
lightningrob: assignedPartitionIds can get large.  We may want to log the whole set only in debug mode, in case addPartition gets called many times on startup.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -229,20 +219,58 @@ public void run() {
...
   /**
-   * Replicas from the given replicas
-   * @param replicasToReplicate list of {@link RemoteReplicaInfo} by data node
+   * Do a lazy removal of {@link RemoteReplicaInfo} from current {@link ReplicaThread}.
+   * @param remoteReplicaInfo {@link RemoteReplicaInfo} to remove.
+   */
+  void removeRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInfo) {
+    lock.lock();
+    try {
+      replicaLazyRemoveSet.add(remoteReplicaInfo);
```
lightningrob: As discussed, it's not clear that the lazy remove set is needed.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationEngine.java
```diff
@@ -262,105 +221,106 @@ public void shutdown() throws ReplicationException {
...
-        replicaThreadCount = dataNodesCount;
-      }
-
-      ResponseHandler responseHandler = new ResponseHandler(clusterMap);
+  protected void addRemoteReplicaInfoToReplicaThread(List<RemoteReplicaInfo> remoteReplicaInfos, boolean startThread) {
+    for (RemoteReplicaInfo remoteReplicaInfo : remoteReplicaInfos) {
+      DataNodeId dataNodeIdToReplicate = remoteReplicaInfo.getReplicaId().getDataNodeId();
+      String datacenter = dataNodeIdToReplicate.getDatacenterName();
+      createThreadPoolIfNecessary(datacenter, startThread);
+      ReplicaThread replicaThread = dataNodeIdToReplicaThread.computeIfAbsent(dataNodeIdToReplicate,
```
lightningrob: Minor: throw IllegalStateException if threadpool was not created for datacenter, to avoid NPE.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationEngine.java
```diff
@@ -262,105 +221,106 @@ public void shutdown() throws ReplicationException {
...
+  /**
+   * Create thread pool for a datacenter.
+   * @param datacenter The datacenter String.
+   * @param numberOfThreads number of threads to create for the thread pool.
+   * @param startThread If thread needs to be started when create.
+   */
+  private List<ReplicaThread> createThreadPool(String datacenter, int numberOfThreads, boolean startThread) {
+    nextReplicaThreadIndex.put(datacenter, new AtomicInteger(0));
+    List<ReplicaThread> replicaThreads = new ArrayList<>();
+    if (numberOfThreads <= 0) {
```
lightningrob: numberOfThreads < 0 should throw IllegalArgumentException

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +64,113 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+    // Add listener for new coming assigned partition
+    virtualReplicatorCluster.addListener(new VirtualReplicatorClusterListener() {
+      @Override
+      public void onPartitionAdded(PartitionId partitionId) {
+        try {
+          addPartition(partitionId);
+        } catch (ReplicationException e) {
+          logger.error("Exception on adding partition: ", e);
```
lightningrob: Add metric (maybe gauge) to track partitions not added.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -66,24 +69,55 @@ public ReplicationManager(ReplicationConfig replicationConfig, ClusterMapConfig
...
+      }
+
+      // start background persistent thread
+      // start scheduler thread to persist replica token in the background
+      if (persistor != null) {
+        this.scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
+            replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
+      }
+    } catch (IOException e) {
+      logger.error("IO error while starting replication", e);
```
lightningrob: I know this is copied code, but why are we not throwing ReplicationException in this case?

lightningrob:
Review completed.  Will approve after comments addressed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +64,113 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+        }
+      }
+
+      @Override
+      public void onPartitionRemoved(PartitionId partitionId) {
+        removeRemoteReplicaInfoFromReplicaThread(partitionToPartitionInfo.get(partitionId).getRemoteReplicaInfos());
```
lightningrob: When a partition is removed, we should flush its tokens so the new node hosting the partitions can get the latest offset.  I think we can call persistor.write(mountPath, true) here.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
zzmao: We discussed this before in a meeting. We want the code continue to run even if `ReplicationManger` is not started.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationEngine.java
zzmao: This is a bug for ReplicationManger. ReplicationManager should be able to start even if number of threads in remote DCs is 0. But if NPE or other exception is throwed, ReplicationManager constructor failed.

So a if condition is added here, please take a look at.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +65,130 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+    this.scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
+        replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
+  }
+
+  /**
+   * Add given {@link PartitionId} and its {@link RemoteReplicaInfo}s to backup list.
+   * @param partitionId the {@link PartitionId} to add.
+   */
+  boolean addPartition(PartitionId partitionId) throws ReplicationException {
+    if (partitionToPartitionInfo.get(partitionId) != null) {
```
lightningrob: partitionToPartitionInfo.hasKey(partitionId) ?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +65,130 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+        replicationMetrics.addRemoteReplicaToLagMetrics(remoteReplicaInfo);
+        replicationMetrics.createRemoteReplicaErrorMetrics(remoteReplicaInfo);
+        remoteReplicaInfos.add(remoteReplicaInfo);
+      }
+      PartitionInfo partitionInfo = new PartitionInfo(remoteReplicaInfos, partitionId, cloudStore, cloudReplica);
+      partitionToPartitionInfo.put(partitionId, partitionInfo);
+      mountPathToPartitionInfoList.compute(cloudReplica.getMountPath(), (key, value) -> {
+        // For CloudBackUpManger, at most one PartitionInfo in the list.
+        List<PartitionInfo> retList;
+        retList = (value == null ? new ArrayList<>() : value);
```
lightningrob: This can be on one line.  Also parenthesis should be around (value == null).

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -229,20 +218,63 @@ public void run() {
...
   /**
-   * Replicas from the given replicas
-   * @param replicasToReplicate list of {@link RemoteReplicaInfo} by data node
+   * Remove {@link RemoteReplicaInfo} from current {@link ReplicaThread}.
+   * @param remoteReplicaInfo {@link RemoteReplicaInfo} to remove.
+   */
+  void removeRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInfo) {
+    lock.lock();
+    try {
+      replicasToReplicateGroupedByNode.get(remoteReplicaInfo.getReplicaId().getDataNodeId()).remove(remoteReplicaInfo);
```
lightningrob: I would handle the case of get() returning null, even if unlikely.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +65,130 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+  @Override
+  public void start() throws ReplicationException {
+    // Add listener for new coming assigned partition
+    virtualReplicatorCluster.addListener(new VirtualReplicatorClusterListener() {
+      @Override
+      public void onPartitionAdded(PartitionId partitionId) {
+        try {
+          addPartition(partitionId);
+        } catch (ReplicationException e) {
+          logger.error("Exception on adding partition{}: ", partitionId, e);
```
lightningrob: Missing space after partition.

ambry-replication/src/main/java/com.github.ambry.replication/RemoteReplicaInfo.java
```diff
@@ -170,7 +179,20 @@ synchronized void onTokenPersisted() {
...
+  public int compareTo(RemoteReplicaInfo other) {
+    return this.getReplicaId()
+        .getPartitionId()
+        .toPathString()
+        .compareTo(other.getReplicaId().getPartitionId().toPathString());
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    return toString().equals(obj.toString());
```
jsjtzyy: I wonder why we didn't check the class type of `obj` here. Feel like we are supposed to check as follows:
```
    if (this == obj) {
      return true;
    }
    if (obj == null || getClass() != obj.getClass()) {
      return false;
    }
    RemoteReplicaInfo info = (RemoteReplicaInfo) obj;
    return compareTo(info) == 0;
```

ambry-replication/src/main/java/com.github.ambry.replication/RemoteReplicaInfo.java
```diff
@@ -170,7 +179,20 @@ synchronized void onTokenPersisted() {
...
-    return replicaId.toString();
+    return replicaId.getPartitionId() + ":" + replicaId.getDataNodeId() + ":" + replicaId.getReplicaPath();
+  }
+
+  @Override
+  public int compareTo(RemoteReplicaInfo other) {
+    return this.getReplicaId()
+        .getPartitionId()
+        .toPathString()
+        .compareTo(other.getReplicaId().getPartitionId().toPathString());
```
jsjtzyy: I think this can be simplified as:
```
replicaId.getPartitionId()
.compareTo(other.getReplicaId().getPartitionId());
```

ambry-replication/src/main/java/com.github.ambry.replication/RemoteReplicaInfo.java
zzmao: That's some stale code for test. Will address it.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +65,128 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
+            new RemoteReplicaInfo(peerReplica, cloudReplica, cloudStore, factory.getNewFindToken(),
+                storeConfig.storeDataFlushIntervalSeconds * SystemTime.MsPerSec * Replication_Delay_Multiplier,
+                SystemTime.getInstance(), peerReplica.getDataNodeId().getPortToConnectTo());
+        replicationMetrics.addRemoteReplicaToLagMetrics(remoteReplicaInfo);
+        replicationMetrics.createRemoteReplicaErrorMetrics(remoteReplicaInfo);
+        remoteReplicaInfos.add(remoteReplicaInfo);
+      }
+      PartitionInfo partitionInfo = new PartitionInfo(remoteReplicaInfos, partitionId, cloudStore, cloudReplica);
+      partitionToPartitionInfo.put(partitionId, partitionInfo);
+      mountPathToPartitionInfoList.compute(cloudReplica.getMountPath(), (key, value) -> {
```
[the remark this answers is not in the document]
zzmao: `mountPathToPartitionInfoList` is a concurrency map, but it's value `ArrayList` is not. `compute` is used to make sure `ArrayList` update thread-safe.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -229,20 +218,78 @@ public void run() {
...
+    return remoteReplicaInfoList;
+  }
+
+  /**
+   * Do replication for replicas grouped by {@link DataNodeId}
+   */
+  public void replicate() {
     boolean allCaughtUp = true;
-    for (List<RemoteReplicaInfo> replicasToReplicatePerNode : replicasToReplicate) {
+    List<DataNodeId> remoteDataNodes = new ArrayList<>(replicasToReplicateGroupedByNode.keySet());
```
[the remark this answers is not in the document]
zzmao: We need a lock. Thanks!

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +65,128 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
         }
-        PartitionInfo partitionInfo = new PartitionInfo(remoteReplicas, partitionId, cloudStore, cloudReplica);
-        partitionsToReplicate.put(partitionId, partitionInfo);
-        partitionGroupedByMountPath.computeIfAbsent(cloudReplica.getMountPath(), key -> new ArrayList<>())
-            .add(partitionInfo);
-      } else {
-        throw new ReplicationException(
-            "Failed to back up partition " + partitionId + " because no peer replicas found.");
+        if (tokenReloaded == false) {
+          logger.warn("Token reload failed. remoteReplicaInfo: {} tokenInfos: {}", remoteReplicaInfo, tokenInfos);
```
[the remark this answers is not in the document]
zzmao: Thanks! Also updated comment.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +65,128 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
-          updateReplicasToReplicate(peerReplica.getDataNodeId().getDatacenterName(), remoteReplicaInfo);
         }
-        PartitionInfo partitionInfo = new PartitionInfo(remoteReplicas, partitionId, cloudStore, cloudReplica);
-        partitionsToReplicate.put(partitionId, partitionInfo);
-        partitionGroupedByMountPath.computeIfAbsent(cloudReplica.getMountPath(), key -> new ArrayList<>())
-            .add(partitionInfo);
-      } else {
-        throw new ReplicationException(
-            "Failed to back up partition " + partitionId + " because no peer replicas found.");
+        if (tokenReloaded == false) {
```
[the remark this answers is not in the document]
zzmao: fixed.

ambry-replication/src/main/java/com.github.ambry.replication/RemoteReplicaInfo.java
```diff
@@ -170,7 +179,20 @@ synchronized void onTokenPersisted() {
 
   @Override
   public String toString() {
-    return replicaId.toString();
+    return replicaId.getPartitionId() + ":" + replicaId.getDataNodeId() + ":" + replicaId.getReplicaPath();
+  }
+
+  @Override
```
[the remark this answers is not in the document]
zzmao: `compare` is only used in test. So compareTo is removed here. Added comparator in `ReplicaThreadTest.java`.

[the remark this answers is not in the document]
zzmao: @jsjtzyy

ambry-cloud/src/main/java/com.github.ambry.cloud/StaticVcrCluster.java
```diff
@@ -77,6 +78,10 @@ public DataNodeId getCurrentDataNodeId() {
     return currentDataNode;
   }
 
+  @Override
+  public void participate(InstanceType role) throws Exception {
+  }
+
   @Override
   public List<? extends PartitionId> getAssignedPartitionIds() {
     return assignedPartitionIds;
```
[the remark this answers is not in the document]
zzmao: Good catch. Added. 

Will add some tests in follow up PRs.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +64,113 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
-      ReplicaId cloudReplica =
-          new CloudReplica(cloudConfig, partitionId, virtualReplicatorCluster.getCurrentDataNodeId());
-      Store cloudStore = new CloudBlobStore(properties, partitionId, cloudDestination, vcrMetrics);
-      try {
-        cloudStore.start();
-      } catch (StoreException e) {
-        throw new ReplicationException("Can't start CloudStore " + cloudStore, e);
+  /**
+   * Add given {@link PartitionId} and its {@link RemoteReplicaInfo}s to backup list.
+   * @param partitionId the {@link PartitionId} to add.
```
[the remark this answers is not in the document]
zzmao: no return value for this method. Added doc for @throw.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationMetrics.java
```diff
@@ -228,138 +230,142 @@ public ReplicationMetrics(MetricRegistry registry, List<? extends ReplicaId> rep
...
   /**
-   * Updates per colo metrics for each thread pool
-   * @param datacenters List of datacenters to replicate from
+   * Updates given colo metrics.
+   * @param datacenter The datacenter to replicate from.
    */
-  public void populatePerColoMetrics(Set<String> datacenters) {
+  public void populateSingleColoMetrics(String datacenter) {
+    Meter interColoReplicationBytesRatePerDC =
+        registry.meter(MetricRegistry.name(ReplicaThread.class, "Inter-" + datacenter + "-ReplicationBytesRate"));
```
cgtz: we can do this in future prs, but I think it would simplify things a lot if we created a `DatacenterMetrics` class to group all of these metrics together. This would allow avoid the need to create separate maps for every single metric in favor of a map from datacenter name to `DatacenterMetrics` object.

zzmao: Will do it in follow up PRs. Other comments are addressed.

zzmao:
Fix a bug: partitionInfo should be removed.
Add try/catch in helix context to avoid running into error state.
Add tests.

ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
```diff
@@ -106,6 +106,13 @@
   @Default("true")
   public final boolean replicationIncludeAll;
 
+  /**
+   * If true, replication token will be persisted explicitly.
+   */
+  @Config("replication.persist.token.on.shutdown.or.replica.remove")
```
jsjtzyy: Could you make the comment clearer? I am trying to understand difference of `shutdown` and `replica move`. Does "token will be persisted explicitly" mean clean shutdown scenario?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
```diff
@@ -54,51 +65,155 @@ public CloudBackupManager(VerifiableProperties properties, CloudConfig cloudConf
...
-        throw new ReplicationException("Can't start CloudStore " + cloudStore, e);
+      @Override
+      public void onPartitionRemoved(PartitionId partitionId) {
+        try {
+          PartitionInfo partitionInfo = partitionToPartitionInfo.remove(partitionId);
+          if (partitionInfo == null) {
+            logger.error("PartitionId {} not exist when remove from {}: ", partitionId, dataNodeId);
+            vcrMetrics.removePartitionErrorCount.inc();
+          }
+          removeRemoteReplicaInfoFromReplicaThread(partitionInfo.getRemoteReplicaInfos());
```
jsjtzyy: `partitionInfo` can be null here.

ambry-utils/src/test/java/com.github.ambry.utils/TestUtils.java
```diff
@@ -356,5 +358,28 @@ public void shutdown() {
...
+
+  /**
+   * Periodically check expectedValue and actualValue until timeout.
+   * @param expectedValue the expected value.
+   * @param expressionToCheck the expression to check.
+   * @param timeoutInMs the time out in millisecond.
+   * @return true if value match.
+   */
+  public static <T> boolean checkAndSleep(T expectedValue, Supplier expressionToCheck, int timeoutInMs) {
+    int checkIntervalInMs = 100;
```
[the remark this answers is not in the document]
zzmao: sure.

ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
zzmao: For CloudBackupManger and ReplicationManager, token is persisted when shutdown.
For CloudBackupManger, replica may be removed, token is also persisted for this case.
ReplicationManager doesn't support replica remove now.
This is used for test only as of now. 

I can add above into code comment.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBackupManager.java
zzmao: Will fix. Thanks!

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -229,20 +219,79 @@ public void run() {
...
+    logger.trace("RemoteReplicaInfo {} is removed from ReplicaThread {}.", remoteReplicaInfo, threadName);
+  }
+
+  /**
+   * Add a {@link RemoteReplicaInfo} to current {@link ReplicaThread}.
+   * @param remoteReplicaInfo {@link RemoteReplicaInfo} to add.
+   */
+  public void addRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInfo) {
+    lock.lock();
+    try {
```
jsjtzyy: I don't see the necessity to add `try` block here.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -229,20 +219,79 @@ public void run() {
...
 
   /**
-   * Replicas from the given replicas
-   * @param replicasToReplicate list of {@link RemoteReplicaInfo} by data node
+   * Remove {@link RemoteReplicaInfo} from current {@link ReplicaThread}.
+   * @param remoteReplicaInfo {@link RemoteReplicaInfo} to remove.
+   */
+  void removeRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInfo) {
+    lock.lock();
+    try {
```
jsjtzyy: maybe we can remove `try` block here

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -229,20 +219,79 @@ public void run() {
...
+      DataNodeId dataNodeId = remoteReplicaInfo.getReplicaId().getDataNodeId();
+      Set<RemoteReplicaInfo> remoteReplicaInfos = replicasToReplicateGroupedByNode.get(dataNodeId);
+      if (remoteReplicaInfos != null) {
+        if (!remoteReplicaInfos.remove(remoteReplicaInfo)) {
```
jsjtzyy: In this remove replica method, do we need to update `allReplicatedPartitions` as well? If yes, we have to convert `allReplicatedPartitions` to a Map<PartitionId, List<ReplicaId>>.

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrMetrics.java
```diff
@@ -42,6 +45,10 @@ public VcrMetrics(MetricRegistry registry) {
     blobDecryptionTime = registry.timer(MetricRegistry.name(CloudBlobStore.class, "BlobDecryptionTime"));
     blobUploadSkippedCount = registry.counter(MetricRegistry.name(CloudBlobStore.class, "BlobUploadSkippedCount"));
     updateTtlNotSetError = registry.counter(MetricRegistry.name(CloudBlobStore.class, "UpdateTtlNotSetError"));
+    addPartitionErrorCount = registry.counter(MetricRegistry.name(CloudBlobStore.class, "AddPartitionErrorCount"));
+    removePartitionErrorCount =
+        registry.counter(MetricRegistry.name(CloudBlobStore.class, "RemovePartitionErrorCount"));
+    tokenReloadWarnCount = registry.counter(MetricRegistry.name(CloudBlobStore.class, "TokenReloadWarnCount"));
```
jsjtzyy: Instead of `CloudBlobStore.class`, I think here should be `CloudBackupManager.class`.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
zzmao: That's a safe guard. @lightningrob suggests add try{} finally{} for all lock operation. (In case some runtime exception: NPE)

cgtz: +1, if possible, unlock calls should go in the finally block so that the lock is guaranteed to be released at some point

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
zzmao: Thanks for pointing out this. 
there is some overlap between disable partition and remove replica, but there is also conflict: one by partition, the other by replica. 
It's not a small change to solve this problem. So I would like do it in follow up PRs. 
But the code is good because we:
1. Don't do dynamically add/remote replica to ReplicationManager.
2. Don't disable/enable partition to CloudBackUpManager.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
jsjtzyy: I see. Good to know. Thanks

ambry-api/src/main/java/com.github.ambry/clustermap/VirtualReplicatorCluster.java
```diff
@@ -34,6 +35,12 @@
    */
   DataNodeId getCurrentDataNodeId();
 
+  /**
+   * Join the cluster as {@link InstanceType#PARTICIPANT} or {@link InstanceType#SPECTATOR} .
+   * @param role the {@link InstanceType}.
+   */
+  void participate(InstanceType role) throws Exception;
```
lightningrob: Not good to expose Helix artifacts in an interface that is supposed to be generic.  Prefer to split this into two methods, participate() and spectate().
Also, it doesn't look like this is ever called with the SPECTATOR role.  Will it be?

zzmao: `SPECTATOR ` will be called for recoveryManager.
HelixClustrerMap somehow used the same method, so I followed it. Open to split it.

lightningrob: If you mean ClusterParticipant, that participate method doesn't have the Helix enum argument and doesn't seem to have the spectator option.  If you want to keep it as one method, then I suggest adding our own enum that maps to Helix's.  But I prefer to split it.

PR #1172, opened 2019-05-15
Add logs in CompositeClusterManager to debug writable partition mismatch

jsjtzyy:
Add more logs in CompositeClusterManager to debug the mismatch on
writable partitions. Also it records which replica is down for certain
partition.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
@@ -192,7 +192,7 @@ public class ClusterMapConfig {
         verifiableProperties.getIntInRange("clustermap.fixedtimeout.disk.retry.backoff.ms", 10 * 60 * 1000, 1,
             30 * 60 * 1000);
     clusterMapFixedTimeoutReplicaErrorThreshold =
-        verifiableProperties.getIntInRange("clustermap.fixedtimeout.replica.error.threshold", 1, 1, 100);
+        verifiableProperties.getIntInRange("clustermap.fixedtimeout.replica.error.threshold", 1, 1, Integer.MAX_VALUE);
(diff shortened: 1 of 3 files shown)


ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
```diff
@@ -192,7 +192,7 @@ public ClusterMapConfig(VerifiableProperties verifiableProperties) {
         verifiableProperties.getIntInRange("clustermap.fixedtimeout.disk.retry.backoff.ms", 10 * 60 * 1000, 1,
             30 * 60 * 1000);
     clusterMapFixedTimeoutReplicaErrorThreshold =
-        verifiableProperties.getIntInRange("clustermap.fixedtimeout.replica.error.threshold", 1, 1, 100);
+        verifiableProperties.getIntInRange("clustermap.fixedtimeout.replica.error.threshold", 1, 1, Integer.MAX_VALUE);
```
jsjtzyy: Change this to `Integer.MAX_VALUE` with intention to disable replica state policy if needed.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CompositeClusterManager.java
```diff
@@ -289,8 +303,7 @@ private boolean areEqual(List<PartitionId> partitionListOne, List<PartitionId> p
  */
 class DuplicatingInputStream extends FilterInputStream {
   private enum Mode {
-    SAVE_READS,
-    SERVE_DUPLICATES
+    SAVE_READS, SERVE_DUPLICATES
```
pnarayanan: does this have to be formatted this way?

jsjtzyy: right, this is due to file format.

PR #1173, opened 2019-05-16
Exclude timed out response from adaptive tracker's histogram

jsjtzyy:
1. Introduce router error code into operation tracker's onResponse().
2. Update histogram when error code doesn't equal OperationTimedOut.
This will help adaptive tracker keep valid latency data points in
histogram and will issue 2nd request in time if server crashed or is
down for deployment.

### DIFF
--- a/ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
@@ -77,15 +77,15 @@ class AdaptiveOperationTracker extends SimpleOperationTracker {
   }
 
   @Override
-  public void onResponse(ReplicaId replicaId, RouterRequestFinalState routerRequestFinalState) {
-    super.onResponse(replicaId, routerRequestFinalState);
+  public void onResponse(ReplicaId replicaId, ServerRequestFinalState serverRequestFinalState) {
+    super.onResponse(replicaId, serverRequestFinalState);
     long elapsedTime;
     if (unexpiredRequestSendTimes.containsKey(replicaId)) {
       elapsedTime = time.milliseconds() - unexpiredRequestSendTimes.remove(replicaId).getSecond();
     } else {
       elapsedTime = time.milliseconds() - expiredRequestSendTimes.remove(replicaId);
     }
-    if (routerRequestFinalState != RouterRequestFinalState.TIMED_OUT) {
+    if (serverRequestFinalState != ServerRequestFinalState.TIMED_OUT) {
       getLatencyHistogram(replicaId).update(elapsedTime);
     }
   }
--- a/ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
@@ -77,15 +77,17 @@ class AdaptiveOperationTracker extends SimpleOperationTracker {
   }
 
   @Override
-  public void onResponse(ReplicaId replicaId, boolean isSuccessFul) {
-    super.onResponse(replicaId, isSuccessFul);
+  public void onResponse(ReplicaId replicaId, boolean isSuccessFul, RouterErrorCode routerErrorCode) {
+    super.onResponse(replicaId, isSuccessFul, routerErrorCode);
     long elapsedTime;
     if (unexpiredRequestSendTimes.containsKey(replicaId)) {
       elapsedTime = time.milliseconds() - unexpiredRequestSendTimes.remove(replicaId).getSecond();
     } else {
       elapsedTime = time.milliseconds() - expiredRequestSendTimes.remove(replicaId);
     }
-    getLatencyHistogram(replicaId).update(elapsedTime);
+    if (routerErrorCode != RouterErrorCode.OperationTimedOut) {
+      getLatencyHistogram(replicaId).update(elapsedTime);
+    }
   }
 
   @Override
@@ -101,7 +103,7 @@ class AdaptiveOperationTracker extends SimpleOperationTracker {
    * @return the {@link Histogram} that tracks requests to the class of replicas (intra or inter DC) that
    * {@code replicaId} belongs to.
    */
-  private Histogram getLatencyHistogram(ReplicaId replicaId) {
+  Histogram getLatencyHistogram(ReplicaId replicaId) {
     if (replicaId.getDataNodeId().getDatacenterName().equals(datacenterName)) {
       return localColoTracker;
     }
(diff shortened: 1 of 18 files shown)


ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
```diff
@@ -77,15 +77,17 @@
...
+  public void onResponse(ReplicaId replicaId, boolean isSuccessFul, RouterErrorCode routerErrorCode) {
+    super.onResponse(replicaId, isSuccessFul, routerErrorCode);
     long elapsedTime;
     if (unexpiredRequestSendTimes.containsKey(replicaId)) {
       elapsedTime = time.milliseconds() - unexpiredRequestSendTimes.remove(replicaId).getSecond();
     } else {
       elapsedTime = time.milliseconds() - expiredRequestSendTimes.remove(replicaId);
     }
-    getLatencyHistogram(replicaId).update(elapsedTime);
+    if (routerErrorCode != RouterErrorCode.OperationTimedOut) {
```
jsjtzyy: If any concern about `routerErrorCode == null` here, please let me know.

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
```diff
@@ -77,15 +77,17 @@
   }
 
   @Override
-  public void onResponse(ReplicaId replicaId, boolean isSuccessFul) {
-    super.onResponse(replicaId, isSuccessFul);
+  public void onResponse(ReplicaId replicaId, boolean isSuccessFul, RouterErrorCode routerErrorCode) {
```
cgtz: how about just passing `boolean updateLatencyHistogram` instead of the error code? This will keep the tracker agnostic to the meaning of different error codes and will match what was done with the `isSuccessful` variable.

jsjtzyy: Thanks for the suggestion. After reconsideration, it might be reasonable to keep router error code in method signature. Two humble thoughts: 1. I changed the `onResponse` signature in `OperationTracker` interface which is implemented by `SimpleOperationTracker` and `AdaptiveOperationTracker`. It seems `updateLatencyHistogram` doesn't apply to `SimpleOperationTracker`.  2. Passing router error code might give us some benefits in the future if we want to change behavior of both operation trackers based on the router error code.

jsjtzyy:
./gradlew build && ./gradlew test succeeded. @pnarayanan @zzmao reminder to review. Thanks!

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
pnarayanan: How about getting rid of the `boolean` in that case and determine success or failure purely based on the error code?

jsjtzyy: It is viable but a little aggressive in this PR.  I see your points here, instead of relying on the caller, the operation tracker should resolve the `isSuccessful` status by itself. Personally, I am a little conservative on this, maybe we can remove that in the future PR. What do you think?

pnarayanan: On second thought, you may not be able to determine success or failure based on the error code alone, as the interpretation might be different for different operation types (for example, `Blob_Deleted` is an error for a Get, but perhaps not for a Delete operation). It would still be nice if the signature can be made more elegant, but I dont' have a good suggestion at this point.

cgtz: Perhaps we could create another enum to represent how the operation tracker should handle the response, but abstracts out the actual error code.

e.g.: `enum OperationTracker.Status {SUCCESS, FAILURE, TIMED_OUT }`

jsjtzyy: sounds like a good idea. Let me make the changes as suggested.

zzmao: `isSuccessFul` is not able to provided enough information, so you introduce `RouterErrorCode`. 
But with both `isSuccessFul` and `RouterErrorCode`, it's a little redundant.

So I think there are two options,
1. Just pass in RouterErrorCode, internally distinguish `isSuccess` and `shoudUpdate`. 
2. Instead of passing in boolean, we can pass in a int(enum), different type do different things. 

Looks like 1 is better.

zzmao: ha, you are discussing in the thread below.

jsjtzyy: took @cgtz 's suggestion and changed the method signature.

jsjtzyy: Instead using of `OperationTracker.Status`, I decide to use the  `RequestResult` here. It is more reasonable in context of `onResponse` method. Also `RequestResult` is more general, it includes the case where request is not successfully sent out due to connection checkout timeout etc.

ambry-router/src/test/java/com.github.ambry.router/GetBlobInfoOperationTest.java
```diff
@@ -269,19 +277,20 @@ public void testPollAndResponseHandling() throws Exception {
...
+  public void testRouterRequestTimeoutAllFailure() {
     NonBlockingRouter.currentOperationsCount.incrementAndGet();
     GetBlobInfoOperation op =
         new GetBlobInfoOperation(routerConfig, routerMetrics, mockClusterMap, responseHandler, blobId, options, null,
             routerCallback, kms, cryptoService, cryptoJobHandler, time, false);
     requestRegistrationCallback.requestListToFill = new ArrayList<>();
     op.poll(requestRegistrationCallback);
+    int count = 0;
     while (!op.isOperationComplete()) {
       time.sleep(routerConfig.routerRequestTimeoutMs + 1);
```
zzmao: Reduce the default timeout.

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
zzmao: `RequestResult ` is a very common name. A more descriptive name would be good. 

Also it would be great if add some comments about the case @pnarayanan mentioned, to remind us in the future.

jsjtzyy: changed the enum name to `RouterRequestFinalState`

PR #1175, opened 2019-05-23
Stitched blobs support different sized component blobs

dharju:
Changed stitched blobs to work with blobs with different data content sizes.

Note: Still need to add serializeMetadataContentV3 into BlobIdTransformer, but regular stitched blob operation changes are ready to be looked at

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -194,6 +194,13 @@ public class RouterConfig {
   public final short routerBlobidCurrentVersion;
 
   /**
+   * The version to use for new metadata blobs.
+   */
+  @Config("router.metadata.content.version")
+  @Default("2")
+  public final short routerMetadataContentVersion;
+
+  /**
    * The KeyManagementServiceFactory that will be used to fetch {@link com.github.ambry.router.KeyManagementService}
    */
   @Config("router.key.management.service.factory")
@@ -292,6 +299,8 @@ public class RouterConfig {
     routerBlobidCurrentVersion =
         verifiableProperties.getShortFromAllowedValues("router.blobid.current.version", (short) 6,
             new Short[]{1, 2, 3, 4, 5, 6});
+    routerMetadataContentVersion =
+        verifiableProperties.getShortFromAllowedValues("router.metadata.content.version", (short) 2, new Short[]{2, 3});
     routerKeyManagementServiceFactory =
         verifiableProperties.getString("router.key.management.service.factory", DEFAULT_KMS_FACTORY);
     routerCryptoServiceFactory =
--- a/ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
+++ b/ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
@@ -27,39 +27,12 @@ import java.util.List;
  */
 public class CompositeBlobInfo {
 
-  /**
-   * POJO class for holding a store key, the size of the data content it refers to,
-   * and the offset of the data in the larger composite blob
-   */
-  public static class StoreKeyAndSizeAndOffset {
-    private final StoreKey storeKey;
-    private final long offset;
-    private final long size;
-
-    public StoreKeyAndSizeAndOffset(StoreKey storeKey, long offset, long size) {
-      this.storeKey = storeKey;
-      this.offset = offset;
-      this.size = size;
-    }
-
-    public StoreKey getStoreKey() {
-      return storeKey;
-    }
-
-    public long getOffset() {
-      return offset;
-    }
-
-    public long getSize() {
-      return size;
-    }
-  }
-
   private final int chunkSize;
   private final long totalSize;
   private final List<StoreKey> keys;
   private final List<Long> offsets = new ArrayList<>();
-  private final List<StoreKeyAndSizeAndOffset> keysAndSizesAndOffsets = new ArrayList<>();
+  private final List<ChunkMetadata> chunkMetadataList = new ArrayList<>();
+  private final short metadataContentVersion;
 
   /**
    * Construct a {@link CompositeBlobInfo} object.
@@ -71,27 +44,29 @@ public class CompositeBlobInfo {
     this.chunkSize = chunkSize;
     this.totalSize = totalSize;
     this.keys = keys;
+    metadataContentVersion = MessageFormatRecord.Metadata_Content_Version_V2;
     long last = 0;
     for (StoreKey key : keys) {
       offsets.add(last);
-      keysAndSizesAndOffsets.add(new StoreKeyAndSizeAndOffset(key, last, Math.min(chunkSize, totalSize - last)));
+      chunkMetadataList.add(new ChunkMetadata(key, last, Math.min(chunkSize, totalSize - last)));
       last += chunkSize;
     }
   }
 
   /**
    * Construct a {@link CompositeBlobInfo} object.
-   * @param keysAndContentSizes
+   * @param keysAndContentSizes list of store keys and the size of the data content they reference
    */
   public CompositeBlobInfo(List<Pair<StoreKey, Long>> keysAndContentSizes) {
     this.chunkSize = -1;
     this.keys = new ArrayList<>();
     long last = 0;
+    metadataContentVersion = MessageFormatRecord.Metadata_Content_Version_V3;
     for (Pair<StoreKey, Long> keyAndContentSize : keysAndContentSizes) {
       StoreKey key = keyAndContentSize.getFirst();
       keys.add(key);
       offsets.add(last);
-      keysAndSizesAndOffsets.add(new StoreKeyAndSizeAndOffset(key, last, keyAndContentSize.getSecond()));
+      chunkMetadataList.add(new ChunkMetadata(key, last, keyAndContentSize.getSecond()));
       last += keyAndContentSize.getSecond();
     }
     this.totalSize = last;
@@ -103,26 +78,29 @@ public class CompositeBlobInfo {
    * blob
    * @param start inclusive starting byte index
    * @param end inclusive ending byte index
-   * @return
+   * @return the list of {@link ChunkMetadata} that lie upon the inclusive range
+   * between 'start' and 'end'
    */
-  public List<StoreKeyAndSizeAndOffset> getStoreKeysInByteRange(long start, long end) {
+  public List<ChunkMetadata> getStoreKeysInByteRange(long start, long end) {
     if (end < start || start < 0L || end >= totalSize) {
       throw new IllegalArgumentException(
           "Bad input parameters, start=" + start + " end=" + end + " totalSize=" + totalSize);
     }
-    int idx = Collections.binarySearch(offsets, start);
+    int startIdx = Collections.binarySearch(offsets, start);
     //binarySearch returns -(insertion point) - 1 if not an exact match, which points to the index of
-    //the first element greater than the key, or list.size() if all elements in the list are
-    // less than the specified key.
-    if (idx < 0) {
-      idx = -idx - 2; //points to element with offset closest to but less than 'start'
+    //the first element greater than 'start', or list.size() if all elements in the list are
+    // less than 'start'.
+    if (startIdx < 0) {
+      startIdx = -startIdx - 2; //points to element with offset closest to but less than 'start'
     }
-    List<StoreKeyAndSizeAndOffset> ans = new ArrayList<>();
-    while (idx < keys.size() && offsets.get(idx) <= end) {
-      ans.add(keysAndSizesAndOffsets.get(idx));
-      idx++;
+
+    int endIdx = Collections.binarySearch(offsets, end);
+    if (endIdx < 0) {
+      endIdx = -endIdx - 2; //points to element with offset closest to but less than 'end'
     }
-    return ans;
+    endIdx++;//List.subList expects an exclusive index
+
+    return chunkMetadataList.subList(startIdx, endIdx);
   }
 
   /**
@@ -152,9 +130,45 @@ public class CompositeBlobInfo {
   /**
    * Get the list of keys for the composite blob's data chunks, along with the
    * key's data content size and offset relative to the total composite blob
-   * @return A list of {@link StoreKeyAndSizeAndOffset}
+   * @return A list of {@link ChunkMetadata}
    */
-  public List<StoreKeyAndSizeAndOffset> getKeysAndSizesAndOffsets() {
-    return keysAndSizesAndOffsets;
+  public List<ChunkMetadata> getChunkMetadataList() {
+    return chunkMetadataList;
+  }
+
+  /**
+   * Return the version of the metadata content record that this object was deserialized from
+   * @return
+   */
+  public short getMetadataContentVersion() {
+    return metadataContentVersion;
+  }
+
+  /**
+   * POJO class for holding a store key, the size of the data content it refers to,
+   * and the offset of the data in the larger composite blob
+   */
+  public static class ChunkMetadata {
+    private final StoreKey storeKey;
+    private final long offset;
+    private final long size;
+
+    public ChunkMetadata(StoreKey storeKey, long offset, long size) {
+      this.storeKey = storeKey;
+      this.offset = offset;
+      this.size = size;
+    }
+
+    public StoreKey getStoreKey() {
+      return storeKey;
+    }
+
+    public long getOffset() {
+      return offset;
+    }
+
+    public long getSize() {
+      return size;
+    }
   }
 }
--- a/ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
+++ b/ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
@@ -15,6 +15,9 @@
 package com.github.ambry.messageformat;
 
 import com.github.ambry.store.StoreKey;
+import com.github.ambry.utils.Pair;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 
 
@@ -23,9 +26,41 @@ import java.util.List;
  * total composite blob size, and a list of keys for the blob's data chunks.
  */
 public class CompositeBlobInfo {
+
+  /**
+   * POJO class for holding a store key, the size of the data content it refers to,
+   * and the offset of the data in the larger composite blob
+   */
+  public static class StoreKeyAndSizeAndOffset {
+    private final StoreKey storeKey;
+    private final long offset;
+    private final long size;
+
+    public StoreKeyAndSizeAndOffset(StoreKey storeKey, long offset, long size) {
+      this.storeKey = storeKey;
+      this.offset = offset;
+      this.size = size;
+    }
+
+    public StoreKey getStoreKey() {
+      return storeKey;
+    }
+
+    public long getOffset() {
+      return offset;
+    }
+
+    public long getSize() {
+      return size;
+    }
+  }
+
+
   private final int chunkSize;
   private final long totalSize;
   private final List<StoreKey> keys;
+  private final List<Long> offsets = new ArrayList<>();
+  private final List<StoreKeyAndSizeAndOffset> keysAndSizesAndOffsets = new ArrayList<>();
 
   /**
    * Construct a {@link CompositeBlobInfo} object.
@@ -37,6 +72,49 @@ public class CompositeBlobInfo {
     this.chunkSize = chunkSize;
     this.totalSize = totalSize;
     this.keys = keys;
+    long last = 0;
+    for (StoreKey key : keys) {
+      offsets.add(last);
+      keysAndSizesAndOffsets.add(new StoreKeyAndSizeAndOffset(key, last, Math.min(chunkSize, totalSize - last)));
+      last += chunkSize;
+    }
+  }
+
+  /**
+   * Construct a {@link CompositeBlobInfo} object.
+   * @param keysAndContentSizes
+   */
+  public CompositeBlobInfo(List<Pair<StoreKey,Long>> keysAndContentSizes) {
+    this.chunkSize = -1;
+    this.keys = new ArrayList<>();
+    long last = 0;
+    for (Pair<StoreKey,Long> keyAndContentSize : keysAndContentSizes) {
+      StoreKey key = keyAndContentSize.getFirst();
+      keys.add(key);
+      offsets.add(last);
+      keysAndSizesAndOffsets.add(new StoreKeyAndSizeAndOffset(key, last, keyAndContentSize.getSecond()));
+      last += keyAndContentSize.getSecond();
+    }
+    this.totalSize = last;
+  }
+
+  public List<StoreKeyAndSizeAndOffset> getStoreKeysInByteRange(long start, long end) {
+    if (end < start || start < 0L || end >= totalSize) {
+      throw new IllegalArgumentException("Bad input parameters, start="+start+" end="+end+" totalSize="+totalSize);
+    }
+    int idx = Collections.binarySearch(offsets, start);
+    //binarySearch returns -(insertion point) - 1 if not an exact match, which points to the index of
+    //the first element greater than the key, or list.size() if all elements in the list are
+    // less than the specified key.
+    if (idx < 0) {
+      idx = -idx-2; //points to element with offset closest to but less than 'start'
+    }
+    List<StoreKeyAndSizeAndOffset> ans = new ArrayList<>();
+    while(idx < keys.size() && offsets.get(idx) <= end) {
+      ans.add(keysAndSizesAndOffsets.get(idx));
+      idx++;
+    }
+    return ans;
   }
 
   /**
@@ -61,5 +139,9 @@ public class CompositeBlobInfo {
    */
   public List<StoreKey> getKeys() {
     return keys;
+  }
+
+  public List<StoreKeyAndSizeAndOffset> getKeysAndSizesAndOffsets() {
+    return keysAndSizesAndOffsets;
   }
 }
--- a/ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
+++ b/ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
@@ -55,7 +55,6 @@ public class CompositeBlobInfo {
     }
   }
 
-
   private final int chunkSize;
   private final long totalSize;
   private final List<StoreKey> keys;
@@ -84,11 +83,11 @@ public class CompositeBlobInfo {
    * Construct a {@link CompositeBlobInfo} object.
    * @param keysAndContentSizes
    */
-  public CompositeBlobInfo(List<Pair<StoreKey,Long>> keysAndContentSizes) {
+  public CompositeBlobInfo(List<Pair<StoreKey, Long>> keysAndContentSizes) {
     this.chunkSize = -1;
     this.keys = new ArrayList<>();
     long last = 0;
-    for (Pair<StoreKey,Long> keyAndContentSize : keysAndContentSizes) {
+    for (Pair<StoreKey, Long> keyAndContentSize : keysAndContentSizes) {
       StoreKey key = keyAndContentSize.getFirst();
       keys.add(key);
       offsets.add(last);
@@ -108,17 +107,18 @@ public class CompositeBlobInfo {
    */
   public List<StoreKeyAndSizeAndOffset> getStoreKeysInByteRange(long start, long end) {
     if (end < start || start < 0L || end >= totalSize) {
-      throw new IllegalArgumentException("Bad input parameters, start="+start+" end="+end+" totalSize="+totalSize);
+      throw new IllegalArgumentException(
+          "Bad input parameters, start=" + start + " end=" + end + " totalSize=" + totalSize);
     }
     int idx = Collections.binarySearch(offsets, start);
     //binarySearch returns -(insertion point) - 1 if not an exact match, which points to the index of
     //the first element greater than the key, or list.size() if all elements in the list are
     // less than the specified key.
     if (idx < 0) {
-      idx = -idx-2; //points to element with offset closest to but less than 'start'
+      idx = -idx - 2; //points to element with offset closest to but less than 'start'
     }
     List<StoreKeyAndSizeAndOffset> ans = new ArrayList<>();
-    while(idx < keys.size() && offsets.get(idx) <= end) {
+    while (idx < keys.size() && offsets.get(idx) <= end) {
       ans.add(keysAndSizesAndOffsets.get(idx));
       idx++;
     }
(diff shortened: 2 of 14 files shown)


ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1469,6 +1471,87 @@ public static CompositeBlobInfo deserializeMetadataContentRecord(DataInputStream
...
+     * Deserialize a metadata content record from a stream.
+     * @param stream The stream to read the serialized record from.
+     * @param storeKeyFactory The factory to use for parsing keys in the serialized metadata content record.
+     * @return A {@link CompositeBlobInfo} object with the chunk size and list of keys from the record.
+     * @throws IOException
+     */
+    public static CompositeBlobInfo deserializeMetadataContentRecord(DataInputStream stream,
+        StoreKeyFactory storeKeyFactory) throws IOException {
+      List<Pair<StoreKey, Long>> keysAndContentSizes = new ArrayList<>();
+      stream.readLong(); //total size
```
jsjtzyy: I suggest passing the total size into `CompositeBlobInfo` constructor and compare it with the calculated total size (which is `last` in your code). If they don't match, throw an exception.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -937,9 +943,9 @@ protected ByteBuffer filterChunkToRange(ByteBuffer buf) {
         buf.position(0);
         buf.limit(0);
       } else {
-        long startOffsetInThisChunk = chunkIndex == 0 ? resolvedByteRange.getStartOffset() % chunkSize : 0;
```
jsjtzyy: I wonder previously if `chunkIndex > 0`, why the `startOffsetInThisChunk` is 0 here?

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -937,9 +943,9 @@ protected ByteBuffer filterChunkToRange(ByteBuffer buf) {
...
+        long startOffsetInThisChunk = chunkIndex == 0 ? resolvedByteRange.getStartOffset() - offset : 0;
         long endOffsetInThisChunk =
-            chunkIndex == (numChunksTotal - 1) ? (resolvedByteRange.getEndOffset() % chunkSize) + 1 : chunkSize;
```
jsjtzyy: One more minor question: is `endOffsetInThisChunk` exclusive or inclusive?
Looks like it is exclusive, right?

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -984,14 +990,14 @@ boolean isComplete() {
...
-    private List<StoreKey> keys;
+    private List<CompositeBlobInfo.StoreKeyAndSizeAndOffset> keysAndSizesAndOffsets;
     private BlobProperties serverBlobProperties;
 
     /**
      * Construct a FirstGetChunk and initialize it with the {@link BlobId} of the overall operation.
      */
     FirstGetChunk() {
-      super(-1, blobId);
+      super(-1, new CompositeBlobInfo.StoreKeyAndSizeAndOffset(blobId, 0L, -1L));
```
jsjtzyy: It seems the `chunkSize` is initialized to `-1L` here and I don't see it gets populated again because you removed several codes regarding setting `chunkSize`.  Correct me if I am wrong.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -1051,7 +1057,6 @@ protected void maybeProcessCallbacks() {
                   decryptCallbackResultInfo.result.getDecryptedUserMetadata().array());
             }
             totalSize = decryptCallbackResultInfo.result.getDecryptedBlobContent().remaining();
-            chunkSize = totalSize;
```
jsjtzyy: Similar to previous comment. Where the `chunkSize` is populated? I notice that `getKeysAndSizesAndOffsets` contains chunk size info, however, note that `filterChunkToRange` method directly uses variable `chunkSize`, which might not be initialized. Please double check code there.

jsjtzyy: Also, we can discuss offline about the chunk size of a simple blob.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1604,12 +1606,14 @@ private void finalizeMetadataChunk() {
               passedInBlobProperties.isEncrypted(), passedInBlobProperties.getExternalAssetTag());
       if (isStitchOperation() || getNumDataChunks() > 1) {
         // values returned are in the right order as TreeMap returns them in key-order.
-        List<StoreKey> orderedChunkIdList = new ArrayList<>(indexToChunkIds.values());
-        buf = MetadataContentSerDe.serializeMetadataContent(intermediateChunkSize, getBlobSize(), orderedChunkIdList);
+        List<Pair<StoreKey, Long>> orderedChunkIdList = indexToChunkIds.values().stream().collect(
+            Collectors.toList());
+
+        buf = MetadataContentSerDe.serializeMetadataContentV3(getBlobSize(), orderedChunkIdList);
```
cgtz: We don't want to start writing in v3 before all frontends have been deployed with read support for metadata content v3. 

Perhaps we should add a router config for the metadata content version like we have for blob id version and then pass the version number into the `MetadataContentSerDe.serializeMetadataContent()` method

dharju:
Will be adding more unit tests (specifically unit test for MessageFormatRecord) and incorporation into BlobIdTransformer

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -23,9 +26,41 @@
...
+      return offset;
+    }
+
+    public long getSize() {
+      return size;
+    }
+  }
+
+
   private final int chunkSize;
```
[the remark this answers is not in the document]
dharju: Chunk size would still be used for deserializing V2 metadata content and keys is used so that getKeys() call doesn't have to parse the StoreKeyAndSizeAndOffset object every time it is called

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1469,6 +1471,87 @@ public static CompositeBlobInfo deserializeMetadataContentRecord(DataInputStream
...
+
+    /**
+     * Get the total size of the metadata content record.
+     * @param keySize The size of each key in bytes.
+     * @param numberOfKeys The total number of keys.
+     * @return The total size in bytes.
+     */
+    public static int getMetadataContentSize(int keySize, int numberOfKeys) {
+      return Version_Field_Size_In_Bytes + TOTAL_SIZE_FIELD_SIZE_IN_BYTES + NUM_OF_KEYS_FIELD_SIZE_IN_BYTES + (
+          numberOfKeys * keySize * SIZE_OF_BLOB_FIELD_IN_BYTES);
```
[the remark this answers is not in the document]
dharju: You're right, adding unit test to catch this

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +72,57 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+    for (StoreKey key : keys) {
+      offsets.add(last);
+      keysAndSizesAndOffsets.add(new StoreKeyAndSizeAndOffset(key, last, Math.min(chunkSize, totalSize - last)));
+      last += chunkSize;
+    }
+  }
+
+  /**
+   * Construct a {@link CompositeBlobInfo} object.
+   * @param keysAndContentSizes
```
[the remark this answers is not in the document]
dharju: added

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +72,57 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+    this.totalSize = last;
+  }
+
+  /**
+   * Returns the keys (along with their data content size and offset relative to the total
+   * composite blob) of the chunks related to the given byte range of the entire composite
+   * blob
+   * @param start inclusive starting byte index
+   * @param end inclusive ending byte index
+   * @return
```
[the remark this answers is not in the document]
dharju: added

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +72,57 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+  public List<StoreKeyAndSizeAndOffset> getStoreKeysInByteRange(long start, long end) {
+    if (end < start || start < 0L || end >= totalSize) {
+      throw new IllegalArgumentException("Bad input parameters, start="+start+" end="+end+" totalSize="+totalSize);
+    }
+    int idx = Collections.binarySearch(offsets, start);
+    //binarySearch returns -(insertion point) - 1 if not an exact match, which points to the index of
+    //the first element greater than the key, or list.size() if all elements in the list are
+    // less than the specified key.
+    if (idx < 0) {
+      idx = -idx-2; //points to element with offset closest to but less than 'start'
```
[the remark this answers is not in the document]
dharju: The first offset will always be '0', so idx will never be -1 (it would only be -1 if 'start' was less than the first offset)

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +72,57 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+   * @return
+   */
+  public List<StoreKeyAndSizeAndOffset> getStoreKeysInByteRange(long start, long end) {
+    if (end < start || start < 0L || end >= totalSize) {
+      throw new IllegalArgumentException("Bad input parameters, start="+start+" end="+end+" totalSize="+totalSize);
+    }
+    int idx = Collections.binarySearch(offsets, start);
+    //binarySearch returns -(insertion point) - 1 if not an exact match, which points to the index of
+    //the first element greater than the key, or list.size() if all elements in the list are
+    // less than the specified key.
```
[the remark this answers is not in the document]
dharju: sure, changed to 'start'

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +72,57 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+   * Returns the keys (along with their data content size and offset relative to the total
+   * composite blob) of the chunks related to the given byte range of the entire composite
+   * blob
+   * @param start inclusive starting byte index
+   * @param end inclusive ending byte index
+   * @return
+   */
+  public List<StoreKeyAndSizeAndOffset> getStoreKeysInByteRange(long start, long end) {
+    if (end < start || start < 0L || end >= totalSize) {
+      throw new IllegalArgumentException("Bad input parameters, start="+start+" end="+end+" totalSize="+totalSize);
```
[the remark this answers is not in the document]
dharju: done

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
dharju: I'm not for adding the total size as a param to the `CompositeBlobInfo` ctor to just perform a check, but I will add the check here, and in `serializeMetadataContentRecord`

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1604,12 +1606,14 @@ private void finalizeMetadataChunk() {
               passedInBlobProperties.isEncrypted(), passedInBlobProperties.getExternalAssetTag());
       if (isStitchOperation() || getNumDataChunks() > 1) {
         // values returned are in the right order as TreeMap returns them in key-order.
-        List<StoreKey> orderedChunkIdList = new ArrayList<>(indexToChunkIds.values());
-        buf = MetadataContentSerDe.serializeMetadataContent(intermediateChunkSize, getBlobSize(), orderedChunkIdList);
+        List<Pair<StoreKey, Long>> orderedChunkIdList = indexToChunkIds.values().stream().collect(
+            Collectors.toList());
```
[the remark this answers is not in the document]
dharju: I don't think there's a specific reason, replaced

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1619,7 +1623,8 @@ private void finalizeMetadataChunk() {
      * @return a list of all of the successfully put chunk ids associated with this blob
      */
     List<StoreKey> getSuccessfullyPutChunkIds() {
-      return new ArrayList<>(indexToChunkIds.values());
+      return indexToChunkIds.values().stream().map(b -> b.getFirst()).collect(
+          Collectors.toList());
```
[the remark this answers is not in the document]
dharju: sure

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
dharju: exclusive since it's used in `Buffer.limit()`.  I can rename it `endOffsetInThisChunkExclusive`

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
dharju: because it's relative to the chunk itself. The chunkIndexes only refer to chunks that are within a byte range.  If the chunkIndex refers to a chunk beyond the first chunk, it will have to start at 0.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
dharju: That's right, but that's ok since chunkSize is only used when calculating the end of data chunk for a byte range operation if that data chunk is not the last data chunk.  For simple data blobs, the FirstChunk will always be the last data chunk so it's unused, and for metadata blobs the size of the FirstChunk is irrelevant.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
dharju: It's ok since chunkSize is only used when calculating the end of data chunk for a byte range operation if that data chunk is not the last data chunk. For simple data blobs, the FirstChunk will always be the last data chunk so it's unused, and for metadata blobs the size of the FirstChunk is irrelevant.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -23,9 +26,41 @@
  * total composite blob size, and a list of keys for the blob's data chunks.
  */
 public class CompositeBlobInfo {
+
+  /**
+   * POJO class for holding a store key, the size of the data content it refers to,
+   * and the offset of the data in the larger composite blob
+   */
+  public static class StoreKeyAndSizeAndOffset {
```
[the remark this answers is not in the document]
dharju: sure

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +72,57 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+      throw new IllegalArgumentException("Bad input parameters, start="+start+" end="+end+" totalSize="+totalSize);
+    }
+    int idx = Collections.binarySearch(offsets, start);
+    //binarySearch returns -(insertion point) - 1 if not an exact match, which points to the index of
+    //the first element greater than the key, or list.size() if all elements in the list are
+    // less than the specified key.
+    if (idx < 0) {
+      idx = -idx-2; //points to element with offset closest to but less than 'start'
+    }
+    List<StoreKeyAndSizeAndOffset> ans = new ArrayList<>();
```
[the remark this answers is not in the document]
dharju: We can, and another cool thing about using `subList()` is that I can do another binary search for the end index replacing the linearly scaling while loop and make this method a O(log(n)) operation (since `subList()` itself is a constant operation for `ArrayList`).  Great suggestion!

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
dharju: added config

dharju:
Fixing failing check, also changing PutManagerTest, GetManagerTest, and NonBlockingRouterTest to test against both old and new versions of metadata content

dharju:
Looking at GetBlobOperationTest failure, appears to be triggered by random chunk size, indicates that a kind of corner case is missing from the test, will fix

dharju:
Source of error was test with a very low probability of generating an incorrect test input.  Fixing

ambry-replication/src/test/java/com.github.ambry.replication/BlobIdTransformerTest.java
```diff
@@ -155,15 +157,23 @@ public void testBasicOperation() throws Exception {
...
-    InputAndExpected inputAndExpected =
-        new InputAndExpected(BLOB_ID_VERSION_1_METADATA_CONVERTED, VALID_MESSAGE_FORMAT_INPUT_STREAM_IMPLS[0], false,
-            new String[]{BLOB_ID_VERSION_1_DATACHUNK_0_CONVERTED.getFirst(),
-                BLOB_ID_VERSION_1_DATACHUNK_1_CONVERTED.getFirst()},
-            new String[]{BLOB_ID_VERSION_1_DATACHUNK_0_CONVERTED.getSecond(),
-                BLOB_ID_VERSION_1_DATACHUNK_1_CONVERTED.getSecond()});
-    TransformationOutput output = transformer.transform(inputAndExpected.getInput());
-    assertNull("output exception should be null", output.getException());
-    verifyOutput(output.getMsg(), inputAndExpected.getExpected());
+    for (int i = 0; i < 2; i++) {
```
jsjtzyy: Move this piece of code to `BlobIdTransformerTest` ctor. Use parameters to test both V2 and V3 metadataContent in this class.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -62,4 +126,49 @@ public int getChunkSize() {
   public List<StoreKey> getKeys() {
```
cgtz: An alternative to materializing a list of `StoreKey`s for every `CompositeBlobInfo` is to return a "view" of `chunkMetadataList`:
```java
  public List<StoreKey> getKeys() {
    return new AbstractList<StoreKey>() {
      @Override
      public StoreKey get(int index) {
        return chunkMetadataList.get(index).getStoreKey();
      }

      @Override
      public int size() {
        return chunkMetadataList.size();
      }
    };
  }
```

It seems like this would suffice for the two use cases that need the `getKeys()` method, the `getChunkIdsOnly` option in the router and `BlobIdTransformer`, as both of them are just interested in iterating through all of the keys and not mutating the list.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1469,6 +1471,97 @@ public static CompositeBlobInfo deserializeMetadataContentRecord(DataInputStream
       return new CompositeBlobInfo(chunkSize, totalSize, keys);
     }
   }
+
+  /**
+   *  - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+   * |         |              |            |            |          |            |            |          |
+   * | version |  total size  | # of keys  |  size of   |   key1   |  size of   |    key2    |  ......  |
+   * |(2 bytes)|  (8 bytes)   | (4 bytes)  | key1 blob  |          | key2 blob  |            |  ......  |
```
cgtz: Is the totalSize field meant for sanity checking that `sum(chunksizes) == totalsize` or are there other places where it is used?

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -552,11 +555,14 @@ void reset() {
     /**
      * Assign a chunk of the overall blob to this GetChunk.
      * @param index the index of the chunk of the overall blob that needs to be fetched through this GetChunk.
-     * @param id the id of the chunk of the overall blob that needs to be fetched through this GetChunk.
+     * @param keyAndSizeAndOffset the id, data content size, and offset (relative to the total size)
+     *                           of the chunk of the overall blob that needs to be fetched through this GetChunk.
      */
-    void initialize(int index, BlobId id) {
+    void initialize(int index, CompositeBlobInfo.ChunkMetadata keyAndSizeAndOffset) {
```
cgtz: change variable names in this class (and others) from `keyAndSizeAndOffset` to `chunkMetadata`

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
cgtz: Sorry about the confusion around `chunkIndex`. Since we now have the `offset` variable, you could consider changing this to `long startOffsetInThisChunk = Math.max(resolvedByteRange.getStartOffset() - offset, 0)` if you prefer.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
cgtz: If you feel that it is cleaner, you could also express this as `long endOffsetInThisChunkExclusive = Math.min(resolvedByteRange.getEndOffset() - offset + 1, chunkSize)`

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +44,63 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+  /**
+   * Returns the keys (along with their data content size and offset relative to the total
+   * composite blob) of the chunks related to the given byte range of the entire composite
+   * blob
+   * @param start inclusive starting byte index
+   * @param end inclusive ending byte index
+   * @return the list of {@link ChunkMetadata} that lie upon the inclusive range
+   * between 'start' and 'end'
+   */
+  public List<ChunkMetadata> getStoreKeysInByteRange(long start, long end) {
```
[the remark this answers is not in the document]
dharju: agreed, actually writing that now.  `CompositeBlobInfo` didn't have a unit test before because it was a POJO class but with this method it should be tested.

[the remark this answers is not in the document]
dharju: Can you explain "0 byte chunks in the middle of the blob"?

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -37,6 +44,63 @@ public CompositeBlobInfo(int chunkSize, long totalSize, List<StoreKey> keys) {
...
+    // less than 'start'.
+    if (startIdx < 0) {
+      startIdx = -startIdx - 2; //points to element with offset closest to but less than 'start'
+    }
+
+    int endIdx = Collections.binarySearch(offsets, end);
+    if (endIdx < 0) {
+      endIdx = -endIdx - 2; //points to element with offset closest to but less than 'end'
+    }
+    endIdx++;//List.subList expects an exclusive index
```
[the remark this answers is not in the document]
dharju: done

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -193,6 +193,13 @@
   @Default("5")
   public final short routerBlobidCurrentVersion;
 
+  /**
+   * The version to use for new metadata blobs.
+   */
+  @Config("router.metadata.content.version")
```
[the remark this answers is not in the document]
dharju: I like `router.metadata.content.version` because it matches the terminology used in the `Metadata_Content_Format_*` classes

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
```diff
@@ -62,4 +126,49 @@ public int getChunkSize() {
...
+   * key's data content size and offset relative to the total composite blob
+   * @return A list of {@link ChunkMetadata}
+   */
+  public List<ChunkMetadata> getChunkMetadataList() {
+    return chunkMetadataList;
+  }
+
+  /**
+   * Return the version of the metadata content record that this object was deserialized from
+   * @return
```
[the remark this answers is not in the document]
dharju: done

ambry-messageformat/src/test/java/com.github.ambry.messageformat/MessageFormatRecordTest.java
```diff
@@ -507,6 +553,27 @@ private CompositeBlobInfo deserializeMetadataContentV2(ByteBuffer metadataConten
...
+        keysAndContentSizes.get(0).getFirst().sizeInBytes(), keysAndContentSizes.size());
+    ByteBuffer metadataContent = ByteBuffer.allocate(size);
+    MessageFormatRecord.Metadata_Content_Format_V3.serializeMetadataContentRecord(metadataContent, totalSize,
+        keysAndContentSizes);
+    metadataContent.flip();
+    return metadataContent;
+  }
+
+  private CompositeBlobInfo deserializeMetadataContentV3(ByteBuffer metadataContent, StoreKeyFactory storeKeyFactory)
+      throws MessageFormatException, IOException {
```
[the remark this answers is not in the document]
dharju: removed

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
dharju: done

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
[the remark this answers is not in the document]
cgtz: Sure. So I was thinking that since we allow 0 byte simple blob uploads, they are valid candidates for stitching. So, someone could potentially create a blob with chunk sizes like: 25, 30, 0, 0, 70, 0, 5

If we call this method with a `start` offset of 30, would the range start from chunk 1 or 2?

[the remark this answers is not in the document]
dharju: As discussed in person, going to change the code so that stitched blobs will not allow blobs of 0 length to be stitched (will throw an exception if that happens)

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
dharju: Yeah it's only used for checking

ambry-messageformat/src/main/java/com.github.ambry.messageformat/CompositeBlobInfo.java
dharju: Cool idea, I'll use it

dharju:
made the max, min changes suggested above

ambry-replication/src/test/java/com.github.ambry.replication/BlobIdTransformerTest.java
dharju: agreed, changed

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -943,8 +943,9 @@ protected ByteBuffer filterChunkToRange(ByteBuffer buf) {
         buf.position(0);
         buf.limit(0);
       } else {
-        long startOffsetInThisChunk = Math.max(resolvedByteRange.getStartOffset() - offset, 0);
-        long endOffsetInThisChunkExclusive = Math.min(resolvedByteRange.getEndOffset() - offset + 1, chunkSize);
+        long startOffsetInThisChunk = chunkIndex == 0 ? resolvedByteRange.getStartOffset() - offset : 0;
```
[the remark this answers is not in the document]
dharju: It broke a GetBlobOperationTest test so I changed it back

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -1241,11 +1243,13 @@ private void initializeDataChunks() {
         numChunksTotal = 0;
         dataChunks = null;
       } else {
-        chunkIdIterator = keys.listIterator();
-        numChunksTotal = keys.size();
-        dataChunks = new GetChunk[Math.min(keys.size(), NonBlockingRouter.MAX_IN_MEM_CHUNKS)];
+        chunkIdIterator = keysAndSizesAndOffsets.listIterator();
```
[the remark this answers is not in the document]
dharju: thought I caught them all... done

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1604,12 +1612,21 @@ private void finalizeMetadataChunk() {
               passedInBlobProperties.isEncrypted(), passedInBlobProperties.getExternalAssetTag());
       if (isStitchOperation() || getNumDataChunks() > 1) {
         // values returned are in the right order as TreeMap returns them in key-order.
-        List<StoreKey> orderedChunkIdList = new ArrayList<>(indexToChunkIds.values());
-        buf = MetadataContentSerDe.serializeMetadataContent(intermediateChunkSize, getBlobSize(), orderedChunkIdList);
+        List<Pair<StoreKey, Long>> orderedChunkIdList = new ArrayList<>(indexToChunkIdsAndChunkSizes.values());
```
cgtz: Move this to inside the else if (line 1620) so that we don't have to construct it for v2

dharju: done

PR #1178, opened 2019-06-03
Log put counts in CloudBlobStore for debugging.

lightningrob:
Better debug messages for token persistence.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -64,6 +64,10 @@ class CloudBlobStore implements Store {
   private final long minTtlMillis;
   private final boolean requireEncryption;
   private boolean started;
+  // Internal counts for put requests to this partition store since store startup (includes all replicas)
+  private long numPutsRequested = 0, numPutsDone = 0, numPutsSkipped = 0;
+  private long lastTimeLoggedCounts = System.currentTimeMillis();
+  private long logCountsInterval = 60000;
 
   /**
    * Constructor for CloudBlobStore
@@ -115,6 +119,8 @@ class CloudBlobStore implements Store {
     // Write the blobs in the message set
     CloudWriteChannel cloudWriter = new CloudWriteChannel(this, messageSetToWrite.getMessageSetInfo());
     messageSetToWrite.writeTo(cloudWriter);
+
+    logPutCountsPeriodically();
   }
 
   /**
@@ -126,6 +132,7 @@ class CloudBlobStore implements Store {
    */
   private void putBlob(MessageInfo messageInfo, ByteBuffer messageBuf, long size)
       throws CloudStorageException, IOException, GeneralSecurityException {
+    numPutsRequested++;
     if (shouldUpload(messageInfo)) {
       BlobId blobId = (BlobId) messageInfo.getStoreKey();
       boolean isRouterEncrypted = isRouterEncrypted(blobId);
@@ -144,6 +151,7 @@ class CloudBlobStore implements Store {
             bufferChanged = true;
           } catch (GeneralSecurityException ex) {
             vcrMetrics.blobEncryptionErrorCount.inc();
+            throw ex;
           } finally {
             encryptionTimer.stop();
           }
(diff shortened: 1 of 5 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -160,10 +168,26 @@ private void putBlob(MessageInfo messageInfo, ByteBuffer messageBuf, long size)
...
       logger.trace("Blob is skipped: {}", messageInfo);
       vcrMetrics.blobUploadSkippedCount.inc();
+      numPutsSkipped++;
+    }
+  }
+
+  /**
+   * Log the collected put counts on the prescribed interval.
+   */
+  private void logPutCountsPeriodically() {
```
zzmao: Why not use a Guage?  Guage can dispaly in-memory variables. 

Actually, we have all these vars in azureMetrics
```
  public final Counter blobUploadRequestCount;
  public final Counter blobUploadSuccessCount;
  public final Counter blobUploadConflictCount;
```
We can have a Guage version of them.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaTokenPersistor.java
```diff
@@ -41,7 +41,7 @@
  */
 public abstract class ReplicaTokenPersistor implements Runnable {
 
-  private static final Logger logger = LoggerFactory.getLogger(DiskTokenPersistor.class);
+  private static final Logger logger = LoggerFactory.getLogger(ReplicaTokenPersistor.class);
```
jsjtzyy: Thanks for catching this.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -64,6 +64,10 @@
   private final long minTtlMillis;
   private final boolean requireEncryption;
   private boolean started;
+  // Internal counts for put requests to this partition store since store startup (includes all replicas)
+  private long numPutsRequested = 0, numPutsDone = 0, numPutsSkipped = 0;
+  private long lastTimeLoggedCounts = System.currentTimeMillis();
+  private long logCountsInterval = 60000;
```
jsjtzyy: make this configurable?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -115,6 +119,8 @@ public void put(MessageWriteSet messageSetToWrite) throws StoreException {
     // Write the blobs in the message set
     CloudWriteChannel cloudWriter = new CloudWriteChannel(this, messageSetToWrite.getMessageSetInfo());
     messageSetToWrite.writeTo(cloudWriter);
+
+    logPutCountsPeriodically();
```
jsjtzyy: I feel like `logPutCountsPeriodically()` depends on `put` method. That is, if `put` method, for some reason, is not invoked then `logPutCountsPeriodically()` won't be called. I am thinking if a background thread to periodically log put counts would be an option. What do you think?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
lightningrob: These are logged per partition, whereas the sensors are per host.  Adding sensors for each partition will become unwieldy.  This is strictly for debugging.

zzmao: ok.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
zzmao: That's a good point. While put rate is very small, we may not be able to see the print out, but a background thread is a little bit heavy.

jsjtzyy: Another way (maybe long-term approach) is to implement VCR's own stats manager to collect all statistics we are interested in and report to Helix or back up locally.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
lightningrob: I may not check in these logging changes.  I'm testing them on one EI host.

cgtz:
@lightningrob, should this PR be closed?

PR #1181, opened 2019-06-05
Use customized blobId string as Azure blob name.

zzmao:
Use customized blobId string as Azure blob name.
Add blobUploadSuccessByteRate and blobUploadSuccessLatency to AzureMetrics.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
@@ -31,6 +31,7 @@ public class CloudBlobMetadata {
   public static final String FIELD_ENCRYPTION_ORIGIN = "encryptionOrigin";
   public static final String FIELD_VCR_KMS_CONTEXT = "vcrKmsContext";
   public static final String FIELD_CRYPTO_AGENT_FACTORY = "cryptoAgentFactory";
+  public static final String FIELD_CLOUD_BLOB_NAME = "cloudBlobName";
 
   private String id;
   private String partitionId;
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
@@ -44,6 +44,7 @@ public class CloudBlobMetadata {
   private EncryptionOrigin encryptionOrigin;
   private String vcrKmsContext;
   private String cryptoAgentFactory;
+  private String cloudBlobName;
 
   /**
    * Possible values of encryption origin for cloud stored blobs.
@@ -57,7 +58,6 @@ public class CloudBlobMetadata {
     ROUTER,
     /** Encrypted by VCR */
     VCR
-
   }
 
   /**
@@ -91,6 +91,7 @@ public class CloudBlobMetadata {
     this.encryptionOrigin = encryptionOrigin;
     this.vcrKmsContext = vcrKmsContext;
     this.cryptoAgentFactory = cryptoAgentFactory;
+    this.cloudBlobName = blobId.getID();
   }
 
   /**
@@ -273,6 +274,23 @@ public class CloudBlobMetadata {
    */
   public CloudBlobMetadata setVcrKmsContext(String vcrKmsContext) {
     this.vcrKmsContext = vcrKmsContext;
+    return this;
+  }
+
+  /**
+   * @return the blob's name in cloud.
+   */
+  public String getCloudBlobName() {
+    return cloudBlobName;
+  }
+
+  /**
+   * Sets blob's name in cloud.
+   * @param cloudBlobName the blob's name in cloud.
+   * @return this instance.
+   */
+  public CloudBlobMetadata setCloudBlobName(String cloudBlobName) {
+    this.cloudBlobName = cloudBlobName;
     return this;
   }
 
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
@@ -14,7 +14,6 @@
 package com.github.ambry.cloud.azure;
 
 import com.codahale.metrics.Counter;
-import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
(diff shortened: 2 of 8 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
```diff
@@ -41,7 +42,8 @@
   public static final String BLOB_DELETION_TIME = "BlobDeletionTime";
   public static final String BLOB_DELETE_ERROR_COUNT = "BlobDeleteErrorCount";
   public static final String CONFIG_ERROR_COUNT = "ConfigErrorCount";
-  public static final String BLOB_UPLOAD_RATE = "BlobUploadRate";
+  public static final String BLOB_UPLOAD_TOTAL_BYTE_RATE = "BlobUploadTotalByteRate";
+  public static final String BLOB_UPLOAD_TOTAL_LATENCY = "BlobUploadTotalLatency";
```
zzmao: Latency can be translated to two graph one is meter and the other is timer.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
```diff
@@ -63,7 +65,8 @@
   public final Timer blobDeletionTime;
   public final Counter blobDeleteErrorCount;
   public final Counter configErrorCount;
-  public final Meter blobUploadRate;
+  public final Meter blobUploadSuccessByteRate;
+  public final Histogram blobUploadSuccessLatency;
```
lightningrob: Why histogram and not timer?  Should the others also be histograms?  We are able to get mean/95/99 values of the timer metrics.

zzmao: Histogram is a combination of Meter and Timer. On ingraph, Histogram will show mean/95/99 and rate. 

Timer is OK. Timer can show mean/95/99 .

zzmao: Thanks for @cgtz  pointing out. It's timer that has histogram and meter.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -276,6 +277,23 @@ public CloudBlobMetadata setVcrKmsContext(String vcrKmsContext) {
...
+  }
+
+  /**
+   * Sets blob's name in cloud.
+   * @param cloudBlobName the blob's name in cloud.
+   * @return this instance.
+   */
+  public CloudBlobMetadata setCloudBlobName(String cloudBlobName) {
+    this.cloudBlobName = cloudBlobName;
+    return this;
```
jsjtzyy: Minor and optional: I see multiple set methods in this class return `this`. I feel like we may use `CloudBlobMetadataBuilder` to instantiate `CloudBlobMetadata`. Currently, most set methods are never called.

zzmao: According to @lightningrob setMethod is used by cosmoDB client.

Let me think about builder.

zzmao: Will not make change this time.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
zzmao: Timer used.
Renamed some metric for readability on ingrpah. 
Fixed a bug.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
```diff
@@ -41,13 +40,14 @@
   public static final String BLOB_DELETION_TIME = "BlobDeletionTime";
   public static final String BLOB_DELETE_ERROR_COUNT = "BlobDeleteErrorCount";
   public static final String CONFIG_ERROR_COUNT = "ConfigErrorCount";
-  public static final String BLOB_UPLOAD_RATE = "BlobUploadRate";
+  public static final String BACKUP_SUCCESS_LATENCY = "backupSuccessLatency";
+  public static final String BACKUP_SUCCESS_BYTE_RATE = "backupSuccessByteRate";
+  public static final String BACKUP_ERROR_COUNT = "backupErrorCount";
```
lightningrob: I don't like this change.  BlobUploadErrorCount was more consistent with the other upload counts.
Also, your metric names are camel case inconsistent with existing ones.

ambry-tools/src/main/java/com.github.ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -124,12 +124,12 @@ static void createCluster(String destZkString, String destClusterName) {
         DedicatedZkClientFactory.getInstance().buildZkClient(new HelixZkClient.ZkConnectionConfig(destZkString));
     destZkClient.setZkSerializer(new ZNRecordSerializer());
     HelixAdmin destAdmin = new ZKHelixAdmin(destZkClient);
-    if (destAdmin.getClusters().contains(destClusterName)) {
-      System.err.println("Failed to create cluster becuase " + destClusterName + " already exist.");
-      return;
-    }
-    ClusterSetup clusterSetup = new ClusterSetup(destZkString);
-    clusterSetup.addCluster(destClusterName, true);
+//    if (destAdmin.getClusters().contains(destClusterName)) {
```
lightningrob: Not good to have commented code.  Is it still needed?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
zzmao: The original BlobUploadErrorCount includes both azure blob upload failure and document upsert failure. The name is confusing. 
Since It's  blob upload error + document error, I rename it to BackupErrorCount to be clear. 

Also, the change moved blob upload related metric inside `uploadIfNotExists`. (blob upload error count can be derived from `blobUploadRequestCount` - `blobUploadSuccessCount`)


Will change the constant string style.

ambry-tools/src/main/java/com.github.ambry/clustermap/HelixVcrPopulateTool.java
zzmao: Removed `//` .

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureMetrics.java
zzmao: Besides the change, Looks like `updateErrorMetrics` is over reused. It was called two place, create and update. We can't know what caused error directly.

lightningrob: I think updateErrorMetrics is okay, it distinguishes between Cosmos and storage errors.
I'm fine with the other change.

PR #1184, opened 2019-06-10
Introducing node-level/disk-level histogram in adaptive tracker

jsjtzyy:
This PR introduces two more operation tracker scope: DataNode and Disk
to allow adaptive tracker to use node-level/disk-level histogram. This
provides different granularity when determining if a certain request is
past due or not. Since all operation tracker scopes are configurable, we
can tune the adaptive tracker and identify the bottleneck of serving
requests on server side.


jsjtzyy:
./gradlew build && ./gradlew test succeeded

PR #1186, opened 2019-06-10
Maintain connection low watermark in NetworkClient

cgtz:
If prewarming is enabled, maintain a minimum number of connections to
each data node. The goal of this change is to allow the frontend to keep
connection pools adequately sized even when servers are restarted or
other disconnections occur.

To track the connection pools that are below the watermark, some of the
connection establishment logic was moved from NetworkClient to
ConnectionTracker.

Upgrade metrics library from codahale to dropwizard as the library was
renamed and dropwizard has the newest updates.


ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
```diff
@@ -56,21 +61,85 @@ boolean mayCreateNewConnection(String host, Port port) {
...
 
   /**
-   * Start tracking a new connection id associated with the given host and port. Note that this connection will not
-   * be made available for checking out until a {@link #checkInConnection(String)} is called on it.
+   * Configure the connection tracker to keep a specified percentage of connections to this (host, port) ready for use.
+   * @param host the hostname.
+   * @param port the port number.
+   * @param warmUpPercentage percentage of max connections to this (host, port) that should be kept ready for use.
+   */
+  void enableConnectionWarmUp(String host, Port port, int warmUpPercentage) {
```
zzmao: Feel more comfortable to call it `setLowWatermark `, but open to it.

ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
```diff
@@ -134,54 +206,51 @@ int getAvailableConnectionsCount() {
...
-      poolCount = 0;
-      maxConnectionsToHostPort = poolLimit;
-      availableConnections = new LinkedList<String>();
+    HostPortPoolManager(String host, Port port, int poolLimit) {
+      this.host = host;
+      this.port = port;
+      this.poolLimit = poolLimit;
+    }
+
+    boolean hasReachedLowWatermark() {
```
zzmao: `isAboveLowWatermark`?

ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
```diff
@@ -25,28 +31,27 @@
  */
 
 class ConnectionTracker {
-  private final HashMap<String, HostPortPoolManager> hostPortToPoolManager;
-  private final HashMap<String, HostPortPoolManager> connectionIdToPoolManager;
+  private static final Logger LOGGER = LoggerFactory.getLogger(ConnectionTracker.class);
+  private final HashMap<Pair<String, Port>, HostPortPoolManager> hostPortToPoolManager = new HashMap<>();
+  private final HashMap<String, HostPortPoolManager> connectionIdToPoolManager = new HashMap<>();
+  private final HashSet<HostPortPoolManager> poolManagersBelowWatermark = new HashSet<>();
```
dharju: `watermark` is sort of odd terminology to me, that come from something?  I would think something like `poolManagersBelowMinimumActiveConnections` would make more sense

ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -114,18 +114,8 @@ private String generateConnectionId(SocketChannel channel) {
     int localPort = socket.getLocalPort();
     String remoteHost = socket.getInetAddress().getHostAddress();
     int remotePort = socket.getPort();
-    long connectionIdSuffix = IdGenerator.getAndIncrement();
```
dharju: Isn't `StringBuilder` more efficient than concatenating strings in Java?  Was this just a code readability vs. performance efficiency tradeoff?

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -138,7 +137,7 @@ public void close() {
       socketChannel.close();
     } catch (IOException ie) {
       metrics.selectorCloseSocketErrorCount.inc();
-      logger.warn("Failed to send SSL close message ", ie);
```
dharju: ?

ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
```diff
@@ -56,21 +61,85 @@ boolean mayCreateNewConnection(String host, Port port) {
...
+    return connectAndTrack(connectionFactory, host, port, true);
+  }
+
+  /**
+   * @see #connectAndTrack(ConnectionFactory, String, Port)
+   * @param connectionFactory the {@link ConnectionFactory} for interfacing with the networking layer.
    * @param host the host to which this connection belongs.
    * @param port the port on the host to which this connection belongs.
-   * @param connId the connection id of the connection.
+   * @param editPoolManagersBelowWatermark true to allow this method to edit {@link #poolManagersBelowWatermark}.
```
dharju: Do we ever add to `poolManagersBelowWatermark` when this is true or is it always removing?  If the latter perhaps we should call this `shouldRemovePoolManagersBelowWatermark`, if former `shouldEditPoolManagersBelowWatermark` (this is actually kind of a weird parameter in general since it's exposing the management of an internal detail (the `poolManagersBelowWatermark` map) to the user.  Can you explain its necessity in the context of its use?)

ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
```diff
@@ -56,21 +61,85 @@ boolean mayCreateNewConnection(String host, Port port) {
   }
 
   /**
-   * Start tracking a new connection id associated with the given host and port. Note that this connection will not
-   * be made available for checking out until a {@link #checkInConnection(String)} is called on it.
+   * Configure the connection tracker to keep a specified percentage of connections to this (host, port) ready for use.
+   * @param host the hostname.
+   * @param port the port number.
+   * @param warmUpPercentage percentage of max connections to this (host, port) that should be kept ready for use.
```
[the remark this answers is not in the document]
cgtz: rounded down. I will document that

ambry-network/src/main/java/com.github.ambry.network/Selector.java
cgtz: That is true that `StringBuilder` is more efficient than multiple string concatenations. However, for simple usages of string `+` (single line, no loops, no method call barriers), java will automatically convert this into a `StringBuilder` usage. Intellij suggests replacing such usages of `StringBuilder` with `+`. Inspecting the bytecode for this line,  `StringBuilder` is in fact used:
```
    LINENUMBER 118 L6
    NEW java/lang/StringBuilder
    DUP
    INVOKESPECIAL java/lang/StringBuilder.<init> ()V
    ALOAD 3
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    LDC ":"
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    ILOAD 4
    INVOKEVIRTUAL java/lang/StringBuilder.append (I)Ljava/lang/StringBuilder;
    LDC "-"
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    ALOAD 5
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    LDC ":"
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    ILOAD 6
    INVOKEVIRTUAL java/lang/StringBuilder.append (I)Ljava/lang/StringBuilder;
    LDC "_"
    INVOKEVIRTUAL java/lang/StringBuilder.append (Ljava/lang/String;)Ljava/lang/StringBuilder;
    LLOAD 7
    INVOKEVIRTUAL java/lang/StringBuilder.append (J)Ljava/lang/StringBuilder;
    INVOKEVIRTUAL java/lang/StringBuilder.toString ()Ljava/lang/String;
    ARETURN
```

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
cgtz: This log message ends up being pretty noisy in practice. Often SSLTransmission.close() is called when a machine is down and the connection failed, which means that there is no way to actually send the ssl close message.

ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
cgtz: You are right, this method only removes from that set. 

This is a private helper method meant to allow `connectAndTrack` and `replenishConnections` to share code. `Set::remove()` cannot be called in `replenishConnections` because it uses an `Iterator` so I added that flag to allow the two methods to both work correctly.

As an alternative, I can just duplicate the small bit of shared code and get rid of this method to reduce the confusion.

ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
cgtz: I think the netty code uses watermark in a few places, but mostly when dealing with buffers, not connections (https://netty.io/4.1/api/io/netty/channel/WriteBufferWaterMark.html). I'm not too tied to the terminology though.

ambry-network/src/main/java/com.github.ambry.network/ConnectionTracker.java
cgtz: Changed the codebase to use `minActiveConnections` instead of `lowWatermark`

cgtz:
@zzmao @dharju addressed review comments

ambry-network/src/test/java/com.github.ambry.network/ConnectionTrackerTest.java
```diff
@@ -205,6 +203,76 @@ public void testConnectionTracker() {
...
+
+    // check out connections
+    String conn1 = connectionTracker.checkOutConnection("host1", plainTextPort, dataNodeId1);
+    Assert.assertNotNull(conn1);
+    String conn2 = connectionTracker.checkOutConnection("host2", sslPort, dataNodeId2);
+    Assert.assertNotNull(conn2);
+    availableCount -= 2;
+    assertCounts(totalConnectionsCount, availableCount);
+
+    // destroy one and return the otheer and then replenish
```
zzmao: typo: other

ambry-network/src/main/java/com.github.ambry.network/Selector.java
dharju: oh wow cool, didn't know that

ambry-network/src/test/java/com.github.ambry.network/NetworkClientTest.java
```diff
@@ -313,6 +316,51 @@ public void testConnectionInitializationFailures() {
...
+    AtomicInteger nextCorrelationId = new AtomicInteger(1);
+    Function<Integer, List<RequestInfo>> requestGen = numRequests -> IntStream.range(0, numRequests)
+        .mapToObj(
+            i -> new RequestInfo(sslHost, sslPort, new MockSend(nextCorrelationId.getAndIncrement()), replicaOnSslNode))
+        .collect(Collectors.toList());
+    // 1 host x 1 port x 3 connections x 100%
+    AtomicInteger expectedConnectCalls = new AtomicInteger(3);
+    Runnable checkConnectCalls = () -> Assert.assertEquals(expectedConnectCalls.get(), selector.connectCallCount());
+
+    networkClient.warmUpConnections(Collections.singletonList(replicaOnSslNode.getDataNodeId()), 100, 2000,
```
dharju: replace '100' and '2000' with static final ints?

ambry-network/src/test/java/com.github.ambry.network/ConnectionTrackerTest.java
```diff
@@ -205,6 +203,76 @@ public void testConnectionTracker() {
...
+    availableCount += minActiveConnectionsCount;
+    assertCounts(totalConnectionsCount, availableCount);
+    Assert.assertTrue(connectionTracker.mayCreateNewConnection("host1", plainTextPort, dataNodeId1));
+    Assert.assertFalse(connectionTracker.mayCreateNewConnection("host2", sslPort, dataNodeId2));
+
+    // remove 2 connections
+    newConnections.stream().limit(2).forEach(connectionTracker::removeConnection);
+    totalConnectionsCount -= 2;
+    availableCount -= 2;
+    assertCounts(totalConnectionsCount, availableCount);
```
dharju: it doesn't seem like this is really testing anything after line 239 (I don't see where line 244 affects `totalConnectionsCount` or `availableCount`)

ambry-network/src/test/java/com.github.ambry.network/ConnectionTrackerTest.java
```diff
@@ -205,6 +203,76 @@ public void testConnectionTracker() {
...
+
+    // replenish connections again
+    connectionTracker.replenishConnections(this::mockNewConnection);
+    newConnections = getNewlyEstablishedConnections();
+    newConnections.forEach(connectionTracker::checkInConnection);
+    totalConnectionsCount += 2;
+    availableCount += 2;
+    assertCounts(totalConnectionsCount, availableCount);
```
dharju: same comment as above

ambry-network/src/test/java/com.github.ambry.network/NetworkClientTest.java
cgtz: replaced the poll timeout and warmup timeout with a constant.

The warmup % has a different meaning than poll timeout, so I made it a local variable

ambry-network/src/test/java/com.github.ambry.network/ConnectionTrackerTest.java
cgtz: 244 takes 2 connections and removes them from the connection pool, thus decreasing the connection count and connections available to check out by 2. This is mostly preparation for the replenishConnections call on line 250, but I wanted to ensure that the connection count was as expected before that call is made.

ambry-network/src/test/java/com.github.ambry.network/ConnectionTrackerTest.java
cgtz: A connection in connectionTracker must be checked in before it is available for use. Prior to being checked in, the connection is counted as initiated, so it still counts towards the max connection limit, but is not available to be used for making a request yet.

This tests that the connections that replenishConnections can be used as usual

ambry-network/src/test/java/com.github.ambry.network/ConnectionTrackerTest.java
dharju: oh I see, `assertCounts` looks at newConnections.  Sorry, I had misread it and thought it was just comparing the two integers

ambry-network/src/test/java/com.github.ambry.network/ConnectionTrackerTest.java
dharju: I see what's going on now as mentioned above

PR #1189, opened 2019-06-13
Add support for querying and purging dead blobs in Azure

lightningrob:
Also added optimization to skip updating blob metadata if field value has not changed.
Note: integration into blob store will happen in follow up PR

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
@@ -38,7 +38,7 @@ public interface CloudDestination {
       throws CloudStorageException;
 
   /**
-   * Delete blob in the cloud destination.
+   * Mark a blob as deleted in the cloud destination.
    * @param blobId id of the Ambry blob
    * @param deletionTime time of blob deletion
    * @return flag indicating whether the blob was deleted
@@ -64,6 +64,31 @@ public interface CloudDestination {
   Map<String, CloudBlobMetadata> getBlobMetadata(List<BlobId> blobIds) throws CloudStorageException;
 
   /**
+   * Get the list of blobs in the specified partition that have been deleted or expired for at least the
+   * configured retention period.
+   * @param partitionPath the partition to query.
+   * @return a List of {@link CloudBlobMetadata} referencing the dead blobs found.
+   * @throws CloudStorageException
+   */
+  List<CloudBlobMetadata> getDeadBlobs(String partitionPath) throws CloudStorageException;
+
+  /**
+   * Permanently delete the specified blob in the cloud destination.
+   * @param blobMetadata the {@link CloudBlobMetadata} referencing the blob to purge.
+   * @return flag indicating whether the blob was successfully purged.
+   * @throws CloudStorageException if the purge operation fails.
+   */
+  boolean purgeBlob(CloudBlobMetadata blobMetadata) throws CloudStorageException;
+
+  /**
+   * Permanently delete the specified blobs in the cloud destination.
+   * @param blobMetadataList the list of {@link CloudBlobMetadata} referencing the blobs to purge.
+   * @return the number of blobs successfully purged.
+   * @throws CloudStorageException if the purge operation fails for any blob.
+   */
+  int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageException;
+
+  /**
    * Checks whether the blob exists in the cloud destination.
    * @param blobId id of the Ambry blob to check.
    * @return {@code true} if the blob exists, otherwise {@code false}.
@@ -78,8 +103,7 @@ public interface CloudDestination {
    * @param inputStream the InputStream containing the replica tokens.
    * @throws CloudStorageException if the upload encounters an error.
    */
-  void persistTokens(String partitionPath, String tokenFileName, InputStream inputStream)
-      throws CloudStorageException;
+  void persistTokens(String partitionPath, String tokenFileName, InputStream inputStream) throws CloudStorageException;
 
   /**
    * Retrieve the persisted replica tokens, if any, for the specified Ambry partition.
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
@@ -424,13 +424,22 @@ class AzureCloudDestination implements CloudDestination {
       // delete blob from storage
       CloudBlobContainer azureContainer = azureBlobClient.getContainerReference(containerName);
       CloudBlockBlob azureBlob = azureContainer.getBlockBlobReference(blobFileName);
-      azureBlob.deleteIfExists();
+      boolean deletedSomething = azureBlob.deleteIfExists();
 
       // Delete the document too
       String docLink = cosmosCollectionLink + "/docs/" + blobId;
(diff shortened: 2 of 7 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
```diff
@@ -63,6 +63,31 @@ boolean uploadBlob(BlobId blobId, long inputLength, CloudBlobMetadata cloudBlobM
...
+   */
+  List<CloudBlobMetadata> getDeadBlobs(String partitionPath) throws CloudStorageException;
+
+  /**
+   * Permanently delete the specified blob in the cloud destination.
+   * @param blobMetadata the {@link CloudBlobMetadata} referencing the blob to purge.
+   * @return flag indicating whether the blob was successfully purged.
+   * @throws CloudStorageException if the purge operation fails.
+   */
+  boolean purgeBlob(CloudBlobMetadata blobMetadata) throws CloudStorageException;
```
cgtz: when will this method return false but not throw CloudStorageException?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -367,6 +410,48 @@ private boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value
...
+      deleteTimer.stop();
+    }
+  }
+
+  @Override
+  public int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageException {
+    int numPurged = 0;
+    for (CloudBlobMetadata blobMetadata : blobMetadataList) {
+      purgeBlob(blobMetadata);
+      numPurged++;
```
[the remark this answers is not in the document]
cgtz: Since purgeBlob was changed to return true/false, should numPurged only be incremented when a deletion actually happened?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -279,22 +287,50 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
     azureMetrics.documentQueryCount.inc();
-    Timer.Context queryTimer = azureMetrics.documentQueryTime.time();
-    String quotedBlobIds =
-        String.join(",", blobIds.stream().map(s -> '"' + s.getID() + '"').collect(Collectors.toList()));
-    String querySpec = String.format(BATCH_ID_QUERY_TEMPLATE, quotedBlobIds);
     FeedOptions feedOptions = new FeedOptions();
-    feedOptions.setPartitionKey(new PartitionKey(blobIds.get(0).getPartition().toPathString()));
-    FeedResponse<Document> response = documentClient.queryDocuments(cosmosCollectionLink, querySpec, feedOptions);
+    feedOptions.setPartitionKey(new PartitionKey(partitionPath));
+    FeedResponse<Document> response = documentClient.queryDocuments(cosmosCollectionLink, query, feedOptions);
```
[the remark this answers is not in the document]
lightningrob: Yes.  Changed to use it.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -367,6 +410,48 @@ private boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value
...
+      // delete blob from storage
+      CloudBlobContainer azureContainer = azureBlobClient.getContainerReference(containerName);
+      CloudBlockBlob azureBlob = azureContainer.getBlockBlobReference(blobFileName);
+      azureBlob.deleteIfExists();
+
+      // Delete the document too
+      String docLink = cosmosCollectionLink + "/docs/" + blobId;
+      documentClient.deleteDocument(docLink, options);
+      azureMetrics.blobDeletedCount.inc();
+      return true;
```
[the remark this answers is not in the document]
lightningrob: Fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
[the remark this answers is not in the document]
lightningrob: Fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
zzmao: Should we log some key information when blob got purged? or send to tracking event?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -367,6 +416,58 @@ private boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value
...
+    String containerName = getAzureContainerName(partitionPath);
+    RequestOptions options = new RequestOptions();
+    options.setPartitionKey(new PartitionKey(partitionPath));
+    azureMetrics.blobDeleteRequestCount.inc();
+    Timer.Context deleteTimer = azureMetrics.blobDeletionTime.time();
+    try {
+      // delete blob from storage
+      CloudBlobContainer azureContainer = azureBlobClient.getContainerReference(containerName);
+      CloudBlockBlob azureBlob = azureContainer.getBlockBlobReference(blobFileName);
+      boolean deletedSomething = azureBlob.deleteIfExists();
```
zzmao: Name it `deleteResult`?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -279,22 +291,52 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
+      List<CloudBlobMetadata> metadataList =
+          runMetadataQuery(blobIds.get(0).getPartition().toPathString(), new SqlQuerySpec(query));
+      return metadataList.stream().collect(Collectors.toMap(m -> m.getId(), Function.identity()));
+    } finally {
+      queryTimer.stop();
+    }
+  }
+
+  @Override
+  public List<CloudBlobMetadata> getDeadBlobs(String partitionPath) throws CloudStorageException {
```
zzmao: Good to have a configurable `limit`.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
lightningrob: I will add a debug log message, don't want to make it info due to the high volume.
A tracking event is worth considering.  The existing compactor currently doesn't do that, so the VCR behavior is consistent.  Prefer to punt on that for now.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: I think deletedSomething is more specific in this case.  I can rename to deletionDone.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: Will add that in a follow up PR.

PR #1191, opened 2019-06-14
Aggregate router exceptions from all replicas and derive it from majority

justinlin-linkedin:
GetChunkOperation would save all the router exceptions from each replica and aggregate them at the last to determine which error code to return to client.


justinlin-linkedin:
Duplicate to #1196

PR #1193, opened 2019-06-17
Segmented get metrics

dharju:


ambry-frontend/src/main/java/com.github.ambry.frontend/AmbryBlobStorageService.java
```diff
@@ -413,6 +413,9 @@ private static RestRequestMetricsGroup getMetricsGroupForGet(FrontendMetrics fro
         case Replicas:
           group = frontendMetrics.getReplicasMetricsGroup;
           break;
+        case Segment:
+          group = frontendMetrics.getSegmentMetricsGroup;
```
cgtz: discussed offline. probably better to use getBlobMetricsGroup here to keep metrics grouping consistent with range requests.

PR #1195, opened 2019-06-17
Add per partition replication lag metrics.

zzmao:
Allow non-clusterMap node to get replication lag from metadata exchange.
Add per partition replication lag for VCR backup.

### DIFF
--- a/ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
@@ -879,6 +879,8 @@ public class ReplicaThread implements Runnable {
               totalBlobsFixed += messageInfoList.size();
               remoteReplicaInfo.setToken(exchangeMetadataResponse.remoteToken);
               remoteReplicaInfo.setLocalLagFromRemoteInBytes(exchangeMetadataResponse.localLagFromRemoteInBytes);
+              replicationMetrics.updateLagMetricForRemoteReplica(remoteReplicaInfo,
+                  exchangeMetadataResponse.localLagFromRemoteInBytes);
               logger.trace("Remote node: {} Thread name: {} Remote replica: {} Token after speaking to remote node: {}",
                   remoteNode, threadName, remoteReplicaInfo.getReplicaId(), exchangeMetadataResponse.remoteToken);
             } catch (StoreException e) {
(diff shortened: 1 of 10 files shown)


jsjtzyy:
Can this PR be extended to normal data nodes?  It seems very useful if the number of metrics won't be a  problem. I feel like it could be applied to monitor replication progress during migration.

zzmao:
I feel it would be some stress for ingraph, let me check if it can be configurable.

zzmao:
Config added so the metric can be enabled per host for normal node, which is usually what we want to observe.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationMetrics.java
```diff
@@ -406,6 +410,15 @@ public void incrementInvalidMessageError(PartitionId partitionId) {
     }
   }
 
+  public void addMetricsForPartition(PartitionId partitionId) {
+    if (partitionLags.put(partitionId, new HashMap<>()) == null) {
```
lightningrob: This overwrite doesn't look correct.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationMetrics.java
```diff
@@ -406,6 +410,15 @@ public void incrementInvalidMessageError(PartitionId partitionId) {
     }
   }
 
+  public void addMetricsForPartition(PartitionId partitionId) {
```
lightningrob: Need javadoc

zzmao:
@lightningrob  Thought about your suggestion last night. Store lag per partition in Helix is possible and it is friendly to our metric system.  @jsjtzyy 
But this needs some effort(figure out how persist value in helix), so I would like to have this PR first and get an overview about the backup progress.
Since the metrics are configurable, we can turn it off any time. Meanwhile, we can check how big impact is it to our metric system.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -879,6 +879,8 @@ private void writeMessagesToLocalStoreAndAdvanceTokens(List<ExchangeMetadataResp
               totalBlobsFixed += messageInfoList.size();
               remoteReplicaInfo.setToken(exchangeMetadataResponse.remoteToken);
               remoteReplicaInfo.setLocalLagFromRemoteInBytes(exchangeMetadataResponse.localLagFromRemoteInBytes);
+              replicationMetrics.updateLagMetricForRemoteReplica(remoteReplicaInfo,
```
lightningrob: Can we update the replication test to verify that the new metrics are getting populated?

zzmao: Test added.

PR #1196, opened 2019-06-18
Improve OperationTracker to terminate when there are two NOT_FOUND from originating DC

justinlin-linkedin:
NOT_FOUND will be treated as a special error code for OperationTracker.

When there are two NOT_FOUND responses from originating dc, operation tracker would terminate the operation (including GetBlob, GetBlobInfo, DeleteBlob and TtlUpdate). The reason we choose to terminate at two, is because all the mutation operations would require two success responses and we have three replicas. So N + M > R.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -303,9 +303,9 @@ public class RouterConfig {
    * NOT_FOUND responses returned from originating dc. Notice that some of the blob ids don't have the datacenter id, it
    * will have no effect on those blobs.
    */
-  @Config("router.operation.tracker.originating.dc.notfound.enabled")
+  @Config("router.operation.tracker.terminate.on.not.found.enabled")
   @Default("false")
-  public final boolean routerOperationTrackerOriginatingDcNotFoundEnabled;
+  public final boolean routerOperationTrackerTerminateOnNotFoundEnabled;
 
   /**
    * The maximum number of inflight requests that allowed for adaptive tracker. If current number of inflight requests
@@ -427,7 +427,7 @@ public class RouterConfig {
       throw new IllegalArgumentException(
           "Operation tracker parallelism is larger than operation tracker max inflight number");
     }
-    routerOperationTrackerOriginatingDcNotFoundEnabled =
-        verifiableProperties.getBoolean("router.operation.tracker.originating.dc.notfound.enabled", false);
+    routerOperationTrackerTerminateOnNotFoundEnabled =
+        verifiableProperties.getBoolean("router.operation.tracker.terminate.on.not.found.enabled", false);
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -303,8 +303,8 @@ public class RouterConfig {
    * NOT_FOUND responses returned from originating dc. Notice that some of the blob ids don't have the datacenter id, it
    * will have no effect on those blobs.
    */
-  @Config("router.operation.tracker.originatingdc.notfound.enabled")
-  @Default("true")
+  @Config("router.operation.tracker.originating.dc.notfound.enabled")
+  @Default("false")
   public final boolean routerOperationTrackerOriginatingDcNotFoundEnabled;
 
   /**
@@ -428,6 +428,6 @@ public class RouterConfig {
           "Operation tracker parallelism is larger than operation tracker max inflight number");
     }
     routerOperationTrackerOriginatingDcNotFoundEnabled =
-        verifiableProperties.getBoolean("router.operation.tracker.originatingdc.notfound.enabled", true);
+        verifiableProperties.getBoolean("router.operation.tracker.originating.dc.notfound.enabled", false);
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -298,6 +298,11 @@ public class RouterConfig {
   @Default("1000")
   public final long routerOperationTrackerMinDataPointsRequired;
 
+  /**
+   * If this config is set to {@code true} the operation tracker would terminate operations when there are more than 2
+   * NOT_FOUND responses returned from originating dc. Notice that some of the blob ids don't have the datacenter id, it
+   * will have no effect on those blobs.
+   */
   @Config("router.operation.tracker.originatingdc.notfound.enabled")
   @Default("true")
   public final boolean routerOperationTrackerOriginatingDcNotFoundEnabled;
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -298,6 +298,10 @@ public class RouterConfig {
   @Default("1000")
   public final long routerOperationTrackerMinDataPointsRequired;
 
+  @Config("router.operation.tracker.originatingdc.notfound.enabled")
+  @Default("true")
+  public final boolean routerOperationTrackerOriginatingDcNotFoundEnabled;
+
   /**
    * The maximum number of inflight requests that allowed for adaptive tracker. If current number of inflight requests
    * is larger than or equal to this threshold, tracker shouldn't send out any request even though the oldest is past due.
@@ -418,5 +422,7 @@ public class RouterConfig {
       throw new IllegalArgumentException(
           "Operation tracker parallelism is larger than operation tracker max inflight number");
     }
+    routerOperationTrackerOriginatingDcNotFoundEnabled =
+        verifiableProperties.getBoolean("router.operation.tracker.originatingdc.notfound.enabled", true);
   }
 }
--- a/ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
@@ -287,8 +287,7 @@ class DeleteOperation {
    * @param replicaId the {@link ReplicaId} associated with the failed response.
    * @param exception the {@link RouterException} associated with the failed response.
    */
-  void onErrorResponse(ReplicaId replicaId, RouterException exception) {
-    logger.info("Get exception: " + exception);
+  private void onErrorResponse(ReplicaId replicaId, RouterException exception) {
     operationTracker.onResponse(replicaId,
         TrackedRequestFinalState.fromRouterErrorCodeToFinalState(exception.getErrorCode()));
     setOperationException(exception);
@@ -308,7 +307,7 @@ class DeleteOperation {
       if (operationTracker.hasSucceeded()) {
         operationException.set(null);
       } else if (operationTracker.hasFailedOnNotFound()) {
-        operationException.set(new RouterException("", RouterErrorCode.BlobDoesNotExist));
+        operationException.set(new RouterException("Operation failed on BlobNotFound", RouterErrorCode.BlobDoesNotExist));
       }
       operationCompleted = true;
     }
--- a/ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
@@ -65,8 +65,6 @@ class DeleteOperation {
   // the cause for failure of this operation. This will be set if and when the operation encounters an irrecoverable
   // failure.
   private final AtomicReference<Exception> operationException = new AtomicReference<Exception>();
-  // RouterErrorCode that is resolved from all the received ServerErrorCode for this operation.
-  private RouterErrorCode resolvedRouterErrorCode;
   // Denotes whether the operation is complete.
   private boolean operationCompleted = false;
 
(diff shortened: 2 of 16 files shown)


justinlin-linkedin:
This PR is working on the same problem as #1191

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
```diff
@@ -237,80 +251,52 @@ private void cleanupExpiredInflightRequests() {
...
-      if (getPrecedenceLevel(error) < getPrecedenceLevel(resolvedRouterErrorCode)) {
-        resolvedRouterErrorCode = error;
-      }
-    }
-    operationTracker.onResponse(replica,
-        resolvedRouterErrorCode == RouterErrorCode.OperationTimedOut ? TrackedRequestFinalState.TIMED_OUT
-            : TrackedRequestFinalState.FAILURE);
-    if (error != RouterErrorCode.BlobDeleted && error != RouterErrorCode.BlobExpired) {
+  void onErrorResponse(ReplicaId replicaId, RouterException exception) {
+    logger.info("Get exception: " + exception);
```
jsjtzyy: I think we don't need log here because in most places where `onErrorResponse` is called, there are TRACE logs.

jsjtzyy:
@justinlin-linkedin  could you rebase this PR on top of latest master branch first? (Ensure it uses same version libraries)

ambry-router/src/test/java/com.github.ambry.router/GetBlobInfoOperationTest.java
```diff
@@ -353,6 +353,104 @@ public void testTimeoutAndBlobNotFound() throws Exception {
         tracker.getLatencyHistogram(remoteReplica).getCount());
   }
 
+  /**
+   * Test the case where originating replicas return Blob_Not_found and the rest returns IO_Error.
+   * @throws Exception
+   */
+  @Test
+  public void testIOErrorAndBlobNotFoundOriginatingNotFound() throws Exception {
```
jsjtzyy: can we make the test name more concise? `testIOErrorAndBlobNotFoundInOriginDc`?

ambry-router/src/test/java/com.github.ambry.router/GetBlobInfoOperationTest.java
```diff
@@ -353,6 +353,104 @@ public void testTimeoutAndBlobNotFound() throws Exception {
...
+    routerConfig = new RouterConfig(new VerifiableProperties(props));
+  }
+
+
+  /**
+   * Test the case where originating replicas return Blob_Not_found and the rest times out.
+   * @throws Exception
+   */
+  @Test
+  public void testTimeoutAndBlobNotFoundOriginatingNotFound() throws Exception {
```
jsjtzyy: same here for the name.

ambry-router/src/test/java/com.github.ambry.router/GetBlobOperationTest.java
```diff
@@ -531,6 +531,54 @@ public void testRequestTimeoutAndBlobNotFound() throws Exception {
         crossColoTracker.getCount());
   }
 
+  /**
+   * Test the case where originating replicas return Blob_Not_found and the rest times out.
+   * @throws Exception
+   */
+  @Test
+  public void testTimeoutAndBlobNotFoundOriginatingNotFound() throws Exception {
```
jsjtzyy: same here

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
```diff
@@ -431,6 +431,79 @@ public void replicasOrderTestOriginatingDcOnly() {
     assertTrue("Operation should be done", ot.isDone());
   }
 
+  /**
+   * Test the case when NotFound Error should be disabled since the cross colo is disabled.
+   */
+  @Test
+  public void originatingDcNotFoundColoDisableTest() {
```
jsjtzyy: `blobNotFoundInOriginDcAndCrossColoDisabledTest`?

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
```diff
@@ -192,15 +192,29 @@ void handleResponse(ResponseInfo responseInfo, DeleteResponse deleteResponse) {
...
+          if (getError == ServerErrorCode.No_Error || getError == ServerErrorCode.Blob_Deleted) {
+            operationTracker.onResponse(replica, TrackedRequestFinalState.SUCCESS);
+            if (RouterUtils.isRemoteReplica(routerConfig, replica)) {
+              logger.trace("Cross colo request successful for remote replica {} in {} ", replica.getDataNodeId(),
+                  replica.getDataNodeId().getDatacenterName());
+              routerMetrics.crossColoSuccessCount.inc();
+            }
+          } else {
+            logger.trace("Replica {} returned an error {} for a delete request with response correlationId : {} ",
+                replica.getDataNodeId(), getError, deleteRequest.getCorrelationId());
```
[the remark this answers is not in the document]
justinlin-linkedin: are you saying I should just use replica, instead of replica.getDataNodeID()?

ambry-router/src/main/java/com.github.ambry.router/GetOperation.java
```diff
@@ -184,7 +184,7 @@ void setOperationException(Exception exception) {
    * @param routerErrorCode The {@link RouterErrorCode} for which to get its precedence level.
    * @return The precedence level of the {@link RouterErrorCode}.
    */
-  private int getPrecedenceLevel(RouterErrorCode routerErrorCode) {
+  protected int getPrecedenceLevel(RouterErrorCode routerErrorCode) {
```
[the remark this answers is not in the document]
justinlin-linkedin: this function should be visible to it's subclasses, GetBlobInfoOperation and GetBlobOperation. Protected should be a better modifier here.

ambry-router/src/main/java/com.github.ambry.router/TtlUpdateOperation.java
```diff
@@ -313,14 +302,32 @@ private void updateOperationState(ReplicaId replica, RouterErrorCode error) {
...
 
+  /**
+   * Set the exception associated with this operation.
+   * First, if current operationException is null, directly set operationException as exception;
+   * Second, if operationException exists, compare ErrorCodes of exception and existing operation Exception depending
+   * on precedence level. An ErrorCode with a smaller precedence level overrides an ErrorCode with a larger precedence
+   * level. Update the operationException if necessary.
+   * @param exception the {@link RouterException} to possibly set.
+   */
+  void setOperationException(Exception exception) {
```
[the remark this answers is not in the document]
justinlin-linkedin: It's not necessarily a `RouterException`, at least in GetOperation. When reading bytes from tcp channel can result in non-RouterException.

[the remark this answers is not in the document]
jsjtzyy: True, but the method in DeleteOperation and TtlUpdateOperation is always taking RouterException. I feel like `operationException.compareAndSet(null, exception);` in `else` block never gets invoked.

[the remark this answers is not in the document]
justinlin-linkedin: right, will update

ambry-router/src/main/java/com.github.ambry.router/TtlUpdateOperation.java
```diff
@@ -51,7 +52,7 @@
   // The operation tracker that tracks the state of this operation.
   private final OperationTracker operationTracker;
   // A map used to find inflight requests using a correlation id.
-  private final Map<Integer, TtlUpdateRequestInfo> ttlUpdateRequestInfos = new HashMap<>();
+  private final Map<Integer, TtlUpdateRequestInfo> ttlUpdateRequestInfos = new TreeMap<>();
```
dharju: why this change?

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
```diff
@@ -96,7 +95,7 @@
     this.callback = callback;
     this.time = time;
     this.deletionTimeMs = time.milliseconds();
-    this.deleteRequestInfos = new HashMap<Integer, DeleteRequestInfo>();
+    this.deleteRequestInfos = new TreeMap<Integer, DeleteRequestInfo>();
```
dharju: Why use treemap here?

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
```diff
@@ -319,14 +301,26 @@ private void updateOperationState(ReplicaId replica, RouterErrorCode error) {
...
 
+  /**
+   * Set the exception associated with this operation.
+   * If operationException exists, compare ErrorCodes of exception and existing operation Exception depending
+   * on precedence level. An ErrorCode with a smaller precedence level overrides an ErrorCode with a larger precedence
+   * level. Update the operationException if necessary.
+   * @param exception the {@link RouterException} to possibly set.
+   */
+  void setOperationException(RouterException exception) {
+    RouterUtils.replaceOperationException(operationException, (RouterException) exception, this::getPrecedenceLevel);
```
dharju: You don't need to cast here since you're accepting a `RouterException`

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
justinlin-linkedin: In fetchRequests function, this map will be iterated based on the order of it's key. So I change it to a sorted tree.

ambry-router/src/main/java/com.github.ambry.router/TtlUpdateOperation.java
justinlin-linkedin: In fetchRequests function, this map will be iterated based on the order of it's key. So I change it to a sorted tree.

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
justinlin-linkedin: updated.

ambry-router/src/main/java/com.github.ambry.router/TtlUpdateOperation.java
```diff
@@ -313,14 +302,27 @@ private void updateOperationState(ReplicaId replica, RouterErrorCode error) {
...
 
+  /**
+   * Set the exception associated with this operation.
+   * If operationException exists, compare ErrorCodes of exception and existing operation Exception depending
+   * on precedence level. An ErrorCode with a smaller precedence level overrides an ErrorCode with a larger precedence
+   * level. Update the operationException if necessary.
+   * @param exception the {@link RouterException} to possibly set.
+   */
+  void setOperationException(Exception exception) {
+    RouterUtils.replaceOperationException(operationException, (RouterException) exception, this::getPrecedenceLevel);
```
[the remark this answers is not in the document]
justinlin-linkedin: you are right, updated.

ambry-router/src/test/java/com.github.ambry.router/DeleteManagerTest.java
```diff
@@ -255,6 +262,13 @@ public void routerErrorCodeResolutionTest() throws Exception {
     codesToSetAndTest.put(ServerErrorCode.Replica_Unavailable, RouterErrorCode.AmbryUnavailable);
     codesToSetAndTest.put(ServerErrorCode.Partition_Unknown, RouterErrorCode.UnexpectedInternalError);
     doRouterErrorCodeResolutionTest(codesToSetAndTest);
+
```
dharju: You need this code block here?  Looks like it was accidentally copied here from above

ambry-router/src/main/java/com.github.ambry.router/TtlUpdateOperation.java
```diff
@@ -313,14 +302,27 @@ private void updateOperationState(ReplicaId replica, RouterErrorCode error) {
   private void checkAndMaybeComplete() {
     // operationCompleted is true if Blob_Authorization_Failure was received.
     if (operationTracker.isDone() || operationCompleted) {
-      if (!operationTracker.hasSucceeded()) {
-        setOperationException(
-            new RouterException("The TtlUpdateOperation could not be completed.", resolvedRouterErrorCode));
+      if (operationTracker.hasSucceeded()) {
+        operationException.set(null);
```
dharju: It doesn't default to null?  It's possible to not be null at this point?

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
dharju: how does that affect functionality?

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -298,6 +298,15 @@
   @Default("1000")
   public final long routerOperationTrackerMinDataPointsRequired;
 
+  /**
```
dharju: Did you run the code formatter on all the files?

ambry-router/src/test/java/com.github.ambry.router/DeleteManagerTest.java
justinlin-linkedin: Just realize this is an old change. updated.

ambry-router/src/main/java/com.github.ambry.router/TtlUpdateOperation.java
justinlin-linkedin: it's defaulted to be null, but it's also possible to be not null at this point. we will send three requests to ambry-server and every error response for those requests will set the operationException to something not null.

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
justinlin-linkedin: updated.

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
justinlin-linkedin: Since DeleteRequestInfo has a creationtTime, and the bigger the key is in the map, the latter the creationTime is in the request. We can rely on this correlation while trying to find all the timed out requests. As long as we hit a request that is not timed out, we can be sure that all the requests following that is not timed out.

PR #1198, opened 2019-06-20
Add configurable warm up for remote datacenter connections

cgtz:
Previously, only local datacenter connections could be prewarmed in the
router. This adds a new config to allow prewarming remote datacenter
connections. This could be useful in deployments where cross-DC requests
are likely.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -66,13 +66,23 @@ public class RouterConfig {
 
   /**
    * The percentage of {@link RouterConfig#routerScalingUnitMaxConnectionsPerPortSsl} or
-   * {@link RouterConfig#routerScalingUnitMaxConnectionsPerPortPlainText} to warm up in the startup.
+   * {@link RouterConfig#routerScalingUnitMaxConnectionsPerPortPlainText} to warm up for data nodes in the local
+   * datacenter during startup.
    * {@link RouterConfig#routerConnectionsWarmUpTimeoutMs} may need to be adjusted.
    */
-  @Config("router.connections.warm.up.percentage.per.port")
+  @Config("router.connections.local.dc.warm.up.percentage")
   @Default("25")
-  public final int routerConnectionsWarmUpPercentagePerPort;
+  public final int routerConnectionsLocalDcWarmUpPercentage;
 
+  /**
+   * The percentage of {@link RouterConfig#routerScalingUnitMaxConnectionsPerPortSsl} or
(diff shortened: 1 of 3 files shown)


ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -66,13 +66,23 @@
...
    */
-  @Config("router.connections.warm.up.percentage.per.port")
+  @Config("router.connections.local.dc.warm.up.percentage")
   @Default("25")
-  public final int routerConnectionsWarmUpPercentagePerPort;
+  public final int routerConnectionsLocalDcWarmUpPercentage;
 
+  /**
+   * The percentage of {@link RouterConfig#routerScalingUnitMaxConnectionsPerPortSsl} or
+   * {@link RouterConfig#routerScalingUnitMaxConnectionsPerPortPlainText} to warm up for data nodes in the local
```
ankagrawal: The comments for routerConnectionsRemoteDcWarmUpPercentage need to be changed perhaps!

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -300,8 +310,10 @@ public RouterConfig(VerifiableProperties verifiableProperties) {
         verifiableProperties.getIntInRange("router.scaling.unit.max.connections.per.port.plain.text", 5, 1, 100);
     routerScalingUnitMaxConnectionsPerPortSsl =
         verifiableProperties.getIntInRange("router.scaling.unit.max.connections.per.port.ssl", 2, 1, 100);
-    routerConnectionsWarmUpPercentagePerPort =
-        verifiableProperties.getIntInRange("router.connections.warm.up.percentage.per.port", 25, 0, 100);
+    routerConnectionsLocalDcWarmUpPercentage =
+        verifiableProperties.getIntInRange("router.connections.local.dc.warm.up.percentage", 25, 0, 100);
+    routerConnectionsRemoteDcWarmUpPercentage =
+        verifiableProperties.getIntInRange("router.connections.remote.dc.warm.up.percentage", 0, 0, 100);
```
ankagrawal: Should the default value be 25 here? We are setting default in the declaration decorator to 25.

cgtz: ah thanks for catching that. The default should be 0 since I feel that warming up cross DC connections only applies to specific ambry use cases, but I will fix the decorator.

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
cgtz: fixed!

PR #1203, opened 2019-06-21
Add off switch for replenishConnections and honor datanode health

cgtz:
Make connection replenishment configurable and off by default.
Avoid trying to initiate connections with nodes detected as down.

### DIFF
--- a/ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
+++ b/ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
@@ -106,8 +106,10 @@ public class NetworkClient implements Closeable {
         pendingRequests.add(new RequestMetadata(time.milliseconds(), requestInfo, clientNetworkRequestMetrics));
       }
       List<NetworkSend> sends = prepareSends(responseInfoList);
-      int connectionsInitiated = connectionTracker.replenishConnections(this::connect);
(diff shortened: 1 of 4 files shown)


ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
```diff
@@ -106,8 +106,10 @@ public NetworkClient(Selector selector, NetworkConfig networkConfig, NetworkMetr
         pendingRequests.add(new RequestMetadata(time.milliseconds(), requestInfo, clientNetworkRequestMetrics));
       }
       List<NetworkSend> sends = prepareSends(responseInfoList);
-      int connectionsInitiated = connectionTracker.replenishConnections(this::connect);
-      networkMetrics.connectionReplenished.inc(connectionsInitiated);
+      if (networkConfig.networkClientEnableConnectionReplenishment) {
+        int connectionsInitiated = connectionTracker.replenishConnections(this::connect);
+        networkMetrics.connectionReplenished.inc(connectionsInitiated);
```
jsjtzyy: minor: do you need to update `networkMetrics.connectionReplenished` in method `warmUpConnections` as well?

cgtz: I think it is better to have this metric only apply to connections replenished in sendAndPoll

PR #1206, opened 2019-06-26
Shutdown request handler before network server in ambry server

justinlin-linkedin:


ambry-server/src/main/java/com.github.ambry.server/AmbryServer.java
```diff
@@ -219,12 +219,12 @@ public void shutdown() {
       if (statsManager != null) {
         statsManager.shutdown();
       }
-      if (networkServer != null) {
-        networkServer.shutdown();
```
cgtz: I wonder what will happen for the requests that are currently being handled by requestHandlerPool if networkServer is shutdown? I think from the client side, they should see the disconnect quickly, but we have to double check that the RequestHandlers won't error out if the network components are shut down (my guess is no since they communicate via queues, but still worth checking)

justinlin-linkedin: Some experiments on the server side are definitely needed.

All the communication between network server and the request handler is through a concurrent blocking queue, so you are right about that request handlers won't error out. After this change, what should be happening is that, when we shutdown the request handler, it will handle all the pending requests in the blocking queue and enqueue corresponding responses to yet another blocking queue. But all the requests received during that will not be handled at all (request handlers are shut down). And there is no guarantee that the responses will be sent out either. My thought on this change is that, it's probably going to gracefully handle some pending requests but not all.

Even w/o this change, the client should be notified fairly soon since the the server might just RST the connection. Either way, clients need to be aware of the sudden termination of the connections.

PR #1212, opened 2019-07-09
Allow store to skip disk flush during shutdown if disk I/O error occurs

jsjtzyy:
Ambry now is able to catch disk I/O error and dynamically shut down
stores on bad disk. However, shutdown process always attampts persist
data onto bad disk which causes shutdown failure. This PR allows store
to skip disk flush if disk I/O error occurs. This ensures shutdown
process completes and marks store stopped.

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
@@ -689,10 +689,10 @@ class BlobStore implements Store {
 
   /**
    * Shuts down the store.
-   * @param shouldSkipDiskFlush {@code true} should skip any disk flush operations during shutdown. {@code false} otherwise.
+   * @param skipDiskFlush {@code true} should skip any disk flush operations during shutdown. {@code false} otherwise.
    * @throws StoreException
    */
-  private void shutdown(boolean shouldSkipDiskFlush) throws StoreException {
+  private void shutdown(boolean skipDiskFlush) throws StoreException {
     long startTimeInMs = time.milliseconds();
     synchronized (storeWriteLock) {
       checkStarted();
@@ -700,8 +700,8 @@ class BlobStore implements Store {
         logger.info("Store : " + dataDir + " shutting down");
         blobStoreStats.close();
         compactor.close(30);
-        index.close(shouldSkipDiskFlush);
-        log.close(shouldSkipDiskFlush);
+        index.close(skipDiskFlush);
+        log.close(skipDiskFlush);
         metrics.deregisterMetrics(storeId);
         started = false;
       } catch (Exception e) {
--- a/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
@@ -684,6 +684,15 @@ class BlobStore implements Store {
 
   @Override
   public void shutdown() throws StoreException {
+    shutdown(false);
+  }
+
+  /**
+   * Shuts down the store.
+   * @param shouldSkipDiskFlush {@code true} should skip any disk flush operations during shutdown. {@code false} otherwise.
+   * @throws StoreException
+   */
+  private void shutdown(boolean shouldSkipDiskFlush) throws StoreException {
     long startTimeInMs = time.milliseconds();
     synchronized (storeWriteLock) {
       checkStarted();
@@ -691,8 +700,8 @@ class BlobStore implements Store {
         logger.info("Store : " + dataDir + " shutting down");
         blobStoreStats.close();
         compactor.close(30);
-        index.close();
-        log.close();
+        index.close(shouldSkipDiskFlush);
+        log.close(shouldSkipDiskFlush);
         metrics.deregisterMetrics(storeId);
         started = false;
       } catch (Exception e) {
@@ -715,7 +724,7 @@ class BlobStore implements Store {
     int count = errorCount.incrementAndGet();
     if (count == config.storeIoErrorCountToTriggerShutdown) {
       logger.error("Shutting down BlobStore {} because IO error count exceeds threshold", storeId);
-      shutdown();
+      shutdown(true);
       metrics.storeIoErrorTriggeredShutdownCount.inc();
     }
   }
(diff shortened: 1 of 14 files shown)


jsjtzyy:
Most of code changes come from `shutdown()` and `close()` refactoring. Unit tests are added and ./gradlew build, ./gradlew test succeeded

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -684,15 +684,24 @@ CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy) throws
...
   public void shutdown() throws StoreException {
+    shutdown(false);
+  }
+
+  /**
+   * Shuts down the store.
+   * @param shouldSkipDiskFlush {@code true} should skip any disk flush operations during shutdown. {@code false} otherwise.
+   * @throws StoreException
+   */
+  private void shutdown(boolean shouldSkipDiskFlush) throws StoreException {
```
lightningrob: Minor: skipDiskFlush (without should)

ambry-store/src/main/java/com.github.ambry.store/LogSegment.java
```diff
@@ -55,6 +57,8 @@
   private final AtomicLong endOffset;
   private final AtomicLong refCount = new AtomicLong(0);
   private final AtomicBoolean open = new AtomicBoolean(true);
+  private final Logger logger = LoggerFactory.getLogger(getClass());
```
lightningrob: We use static logger in most places - more efficient since one per class rather than instance.

ambry-store/src/main/java/com.github.ambry.store/LogSegment.java
```diff
@@ -463,11 +489,24 @@ void flush() throws IOException {
...
+   * @param shouldSkipDiskFlush whether to skip any disk flush operations.
+   * @throws IOException if there is an I/O error while closing the log segment.
    */
-  void close() throws IOException {
+  void close(boolean shouldSkipDiskFlush) throws IOException {
     if (open.compareAndSet(true, false)) {
-      flush();
-      fileChannel.close();
+      if (!shouldSkipDiskFlush) {
+        flush();
```
lightningrob: Why skip this completely, as opposed to catching and ignoring the exception like you do when closing the channel?

ambry-store/src/test/java/com.github.ambry.store/LogSegmentTest.java
```diff
@@ -69,6 +71,7 @@ public void cleanup() {
       }
     }
     assertTrue("The directory [" + tempDir.getAbsolutePath() + "] could not be deleted", tempDir.delete());
+    validateMockitoUsage();
```
lightningrob: Why is this here?

jsjtzyy: Quote from: https://static.javadoc.io/org.mockito/mockito-core/2.9.0/org/mockito/Mockito.html#validateMockitoUsage()
```
Sometimes though, you might want to validate the framework usage explicitly. For example, one of the users wanted to put validateMockitoUsage() in his @After method so that he knows immediately when he misused Mockito. Without it, he would have known about it not sooner than next time he used the framework. One more benefit of having validateMockitoUsage() in @After is that jUnit runner and rule will always fail in the test method with defect whereas ordinary 'next-time' validation might fail the next test method.
```

ambry-store/src/main/java/com.github.ambry.store/LogSegment.java
jsjtzyy: good point, will fix this.

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
jsjtzyy: will change as suggested

ambry-store/src/main/java/com.github.ambry.store/LogSegment.java
jsjtzyy: `flush()` is always attempting to write into disk, which will definitely throw IOException if the disk is bad. The code skips it to avoid unnecessary disk flush operations (note that this method is in `LogSegment`, each store may attempt to write to disk multiple times). However, I am not sure `fileChannel.close()` will invoke IOException if there is disk I/O error. My intention is to give it a try to close the file descriptor of this segment.

jsjtzyy:
@lightningrob partially addressed your comments. Gentle reminder to review. Thanks!

ambry-store/src/test/java/com.github.ambry.store/LogSegmentTest.java
```diff
@@ -422,6 +425,72 @@ public void writeFromTest() throws IOException, StoreException {
...
+  /**
+   * Test that closing log segment (skip disk flush) and exception occurs. The shutdown process should continue.
+   */
+  @Test
+  public void closeLogSegmentWithExceptionTest() throws Exception {
+    Pair<LogSegment, FileChannel> segmentAndFileChannel = getSegmentAndFileChannel("log_current3");
+    LogSegment segment = segmentAndFileChannel.getFirst();
+    FileChannel mockFileChannel = segmentAndFileChannel.getSecond();
+
+    // test that log segment is being closed due to disk I/O error (shouldSkipDiskFlush = true) and exception occurs
```
lightningrob: Minor: adjust comment

ambry-store/src/test/java/com.github.ambry.store/LogSegmentTest.java
```diff
@@ -422,6 +425,72 @@ public void writeFromTest() throws IOException, StoreException {
...
+    doThrow(new IOException("close channel failure")).when(mockFileChannel).close();
+    segment.close(true);
+    verify(mockFileChannel, times(0)).force(true);
+    assertTrue("File couldn't be deleted.", (new File(tempDir, segment.getName()).delete()));
+
+    segmentAndFileChannel = getSegmentAndFileChannel("log_current4");
+    segment = segmentAndFileChannel.getFirst();
+    mockFileChannel = segmentAndFileChannel.getSecond();
+
+    // test that log segment is being closed during normal shutdown (shouldSkipDiskFlush = false) and exception occurs
```
lightningrob: Minor: here too

PR #1215, opened 2019-07-13
Free PutOperation references earlier

cgtz:
This contains a couple of fixes to free references to PutOperations
earlier and to allow GC to take place:
- Since NettyMessageProcessor keeps a reference to the last NettyRequest
object sent over an HTTP connection, there can be a link to a callback
that prevents a previous PutOperation's buffer from being GCed. The
easiest way to address this is to nullify the callback reference after
it is used.
- correlationIdToPutOperation map would previously retain a reference to
a PutOperation even if the operation was fully complete. This means that
large buffers would be retained until all responses are received, even
if the server response is no longer relevant to the operation.

### DIFF
--- a/ambry-rest/src/main/java/com.github.ambry.rest/NettyRequest.java
+++ b/ambry-rest/src/main/java/com.github.ambry.rest/NettyRequest.java
@@ -45,6 +45,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.net.ssl.SSLSession;
 import javax.servlet.http.Cookie;
@@ -642,7 +643,7 @@ class NettyRequest implements RestRequest {
      */
     public final FutureResult<Long> futureResult = new FutureResult<Long>();
 
-    private final Callback<Long> callback;
+    private Callback<Long> callback;
     private final AtomicLong totalBytesRead = new AtomicLong(0);
     private final AtomicBoolean callbackInvoked = new AtomicBoolean(false);
 
@@ -673,6 +674,9 @@ class NettyRequest implements RestRequest {
         futureResult.done(totalBytesRead.get(), exception);
         if (callback != null) {
           callback.onCompletion(totalBytesRead.get(), exception);
+          // the callback may hold a reference to a large buffer. Setting the callback to null allows for memory to be
+          // freed before the next request comes over this connection.
+          callback = null;
         }
       }
     }
--- a/ambry-rest/src/main/java/com.github.ambry.rest/NettyRequest.java
+++ b/ambry-rest/src/main/java/com.github.ambry.rest/NettyRequest.java
@@ -45,7 +45,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReentrantLock;
 import javax.net.ssl.SSLSession;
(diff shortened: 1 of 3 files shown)


zzmao:
Changes LGTM. Make sure all tests can pass.

ambry-rest/src/main/java/com.github.ambry.rest/NettyRequest.java
```diff
@@ -45,6 +45,7 @@
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
```
jsjtzyy: can be removed

ambry-router/src/main/java/com.github.ambry.router/PutManager.java
```diff
@@ -282,6 +283,11 @@ void onComplete(PutOperation op) {
       (op.isEncryptionEnabled() ? routerMetrics.putEncryptedBlobOperationLatencyMs
           : routerMetrics.putBlobOperationLatencyMs).update(operationLatencyMs);
     }
+    // Preemptively clean up of correlation IDs to avoid retaining a reference to the PutOperation for longer than
+    // required. We do not care about any other
+    for (Integer correlationId : op.getInFlightCorrelationIds()) {
+      correlationIdToPutOperation.remove(correlationId);
+    }
```
jsjtzyy: Thanks for making this change. Can it be applied to GET Operation?

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -798,6 +799,15 @@ String getServiceId() {
     return passedInBlobProperties.getServiceId();
   }
 
+  /**
+   * This will return a view of the correlation IDs of requests that may still be in flight. This is not threadsafe
+   * and should only be called from the main event loop.
+   * @return the set of correlation IDs of requests that are still in flight.
+   */
+  Set<Integer> getInFlightCorrelationIds() {
+    return correlationIdToPutChunk.keySet();
```
dharju: Are all reads/writes to `correlationIdToPutChunk` done by the main loop?  What other threads touch `PutOperation`?

jsjtzyy: if this is a view, can we use UnmodifiableSet?

ambry-rest/src/main/java/com.github.ambry.rest/NettyRequest.java
```diff
@@ -673,6 +674,9 @@ public void invokeCallback(Exception exception) {
         futureResult.done(totalBytesRead.get(), exception);
         if (callback != null) {
           callback.onCompletion(totalBytesRead.get(), exception);
+          // the callback may hold a reference to a large buffer. Setting the callback to null allows for memory to be
+          // freed before the next request comes over this connection.
+          callback = null;
```
dharju: I wonder if we can add junit tests for this by using a `WeakReference` to see if an object has been `enqueued` or not....

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
cgtz: Yep this is a thread unsafe map to begin with and all operations are done from the poll and handleResponse methods, which are called in the main loop.

ambry-router/src/main/java/com.github.ambry.router/PutManager.java
cgtz: I will look at this in the future for Get/Delete/TtlUpdate. The memory issues I saw in the heap dump did not seem to be as apparent for GetOperation since chunk buffer references are removed as soon as they are written out.

ambry-rest/src/main/java/com.github.ambry.rest/NettyRequest.java
cgtz: I think it might be hard to make such a test deterministic, but I will keep thinking about if there is a way to better test this. There is always reflection, but that is a little intrusive into the inner structure of the class being tested.

PR #1216, opened 2019-07-15
Implement VCR findEntriesSince method

lightningrob:
Also, add batching to Cosmos missing keys query to prevent overflow

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -249,10 +249,7 @@ class CloudBlobStore implements Store {
         return new FindInfo(Collections.emptyList(), inputToken);
       } else {
         List<MessageInfo> messageEntries = new ArrayList<>();
-        long bytesReadThisQuery = 0;
-        String latestBlobIdString = null;
         for (CloudBlobMetadata metadata : results) {
-          latestBlobIdString = metadata.getId();
           BlobId blobId = new BlobId(metadata.getId(), clusterMap);
           long operationTime = (metadata.getDeletionTime() > 0) ? metadata.getDeletionTime()
               : (metadata.getCreationTime() > 0) ? metadata.getCreationTime() : metadata.getUploadTime();
@@ -262,11 +259,10 @@ class CloudBlobStore implements Store {
               new MessageInfo(blobId, metadata.getSize(), isDeleted, isTtlUpdated, metadata.getExpirationTime(),
                   (short) metadata.getAccountId(), (short) metadata.getContainerId(), operationTime);
           messageEntries.add(messageInfo);
-          bytesReadThisQuery += metadata.getSize();
         }
-        CloudFindToken outputToken =
-            new CloudFindToken(results.get(results.size() - 1).getUploadTime(), latestBlobIdString,
-                inputToken.getBytesRead() + bytesReadThisQuery);
+
+        // Build the new find token from the original one and the query results
+        CloudFindToken outputToken = CloudFindToken.getUpdatedToken(inputToken, results);
         return new FindInfo(messageEntries, outputToken);
       }
     } catch (CloudStorageException | IOException ex) {
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -14,6 +14,7 @@
 package com.github.ambry.cloud;
 
 import com.codahale.metrics.Timer;
+import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.commons.BlobId;
 import com.github.ambry.config.CloudConfig;
@@ -37,6 +38,8 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.ReadableByteChannel;
 import java.security.GeneralSecurityException;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.List;
@@ -58,6 +61,7 @@ class CloudBlobStore implements Store {
   private static final Logger logger = LoggerFactory.getLogger(CloudBlobStore.class);
   private final PartitionId partitionId;
   private final CloudDestination cloudDestination;
+  private final ClusterMap clusterMap;
   private final CloudBlobCryptoAgentFactory cryptoAgentFactory;
   private final CloudBlobCryptoAgent cryptoAgent;
   private final VcrMetrics vcrMetrics;
@@ -75,9 +79,10 @@ class CloudBlobStore implements Store {
    */
   CloudBlobStore(VerifiableProperties properties, PartitionId partitionId, CloudDestination cloudDestination,
       VcrMetrics vcrMetrics) throws IllegalStateException {
-
+    // TODO: need clusterMap too
     CloudConfig cloudConfig = new CloudConfig(properties);
     ClusterMapConfig clusterMapConfig = new ClusterMapConfig(properties);
+    clusterMap = null;
     this.cloudDestination = Objects.requireNonNull(cloudDestination, "cloudDestination is required");
     this.partitionId = Objects.requireNonNull(partitionId, "partitionId is required");
     this.vcrMetrics = Objects.requireNonNull(vcrMetrics, "vcrMetrics is required");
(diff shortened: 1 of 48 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -245,6 +256,24 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
+    try {
+      String containerName = getAzureContainerName(partitionPath);
+      CloudBlobContainer azureContainer = azureBlobClient.getContainerReference(containerName);
+      CloudBlockBlob azureBlob = azureContainer.getBlockBlobReference(blobFileName);
+      if (!azureBlob.exists(null, null, blobOpContext)) {
+        return null;
+      }
+      azureBlob.download(outputStream, null, null, blobOpContext);
+      azureBlob.getMetadata(); // convert to CBM
+      return null;
```
dharju: So this method always returns null? What's its purpose?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -269,14 +298,54 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
     } catch (DocumentClientException dex) {
       throw new CloudStorageException("Failed to query dead blobs for partition " + partitionPath, dex);
     }
   }
 
+  @Override
+  public List<CloudBlobMetadata> findEntriesSince(String partitionPath, CloudFindToken findToken,
+      long maxTotalSizeOfEntries) throws CloudStorageException {
+    SqlQuerySpec entriesSinceQuery = new SqlQuerySpec(ENTRIES_SINCE_QUERY_TEMPLATE,
+        new SqlParameterCollection(new SqlParameter(LIMIT_PARAM, 1000),
```
dharju: Make the `1000` a static final variable?

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
```diff
@@ -249,6 +250,56 @@ public void testPurgeDeadBlobs() throws Exception {
     cleanup();
   }
 
+  /**
```
dharju: Test coverage %?

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureCloudDestinationTest.java
```diff
@@ -253,6 +258,51 @@ public void testQueryMetadata() throws Exception {
     assertEquals(1, azureMetrics.deadBlobsQueryTime.getCount());
   }
 
+  /** Test findEntriesSince. */
```
dharju: Test coverage %?

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -218,6 +223,56 @@ public void testFindMissingKeys() throws Exception {
     assertEquals("Wrong number of missing keys", count, missingKeys.size());
   }
 
+  /** Test the CloudBlobStore findEntriesSince method. */
```
dharju: Test coverage %?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: Sorry, this was temporary unfinished code that was not meant for this PR.  I'll remove it.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: sure

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
lightningrob: Not following.  Are you asking what the test coverage is or saying it should be mentioned in comments?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -236,7 +241,33 @@ public void updateTtl(MessageWriteSet messageSetToUpdate) throws StoreException
...
+      if (results.isEmpty()) {
+        return new FindInfo(Collections.emptyList(), inputToken);
+      } else {
+        List<MessageInfo> messageEntries = new ArrayList<>();
+        for (CloudBlobMetadata metadata : results) {
+          BlobId blobId = new BlobId(metadata.getId(), clusterMap);
+          long operationTime = (metadata.getDeletionTime() > 0) ? metadata.getDeletionTime()
+              : (metadata.getCreationTime() > 0) ? metadata.getCreationTime() : metadata.getUploadTime();
+          boolean isDeleted = metadata.getDeletionTime() > 0;
+          boolean isTtlUpdated = false;  // No way to know
```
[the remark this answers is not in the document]
lightningrob: I'm not sure.  We don't track the ttl update history in blob metadata, so an Azure blob with expirationTime == -1 can be either ttl updated or never had one to begin with.  When the blob gets recovered to Ambry storage, will it matter?

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
dharju: Sorry, just asking what it is for the intended tested object, or how much of the newly added code is tested, can be put in the comments.

lightningrob: For both AzureCloudDestination and CloudBlobStore, the new method has 85-90% coverage, only  exception throw is missing.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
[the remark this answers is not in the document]
dharju: Looking through the code at where isTtlUpdated is read, I agree, it shouldn't be a problem if the expirationTime is accurate.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudFindToken.java
```diff
@@ -0,0 +1,130 @@
...
+  public String getLatestBlobId() {
+    return latestBlobId;
+  }
+
+  public long getLatestUploadTime() {
+    return latestUploadTime;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
dharju: +1 This is a good example of a good implementation of `hashCode` and `equals` since it makes good use of `Objects.equals` (which is good for checking equality with objects that may be null) and `Objects.hash` (which outsources the hard parts of developing a good hash with multiple fields) and I like the use of `getClass()` instead of `instanceof` since it demands an exact class match instead of subtypes being ok

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -68,11 +70,18 @@
...
+  private static final String LIMIT_PARAM = "@limit";
+  private static final String TIME_SINCE_PARAM = "@timesince";
   private static final String BATCH_ID_QUERY_TEMPLATE = "SELECT * FROM c WHERE c.id IN (%s)";
   static final String DEAD_BLOBS_QUERY_TEMPLATE =
-      "SELECT * FROM c WHERE (c." + CloudBlobMetadata.FIELD_DELETION_TIME + " BETWEEN 1 AND " + THRESHOLD_PARAM + ")"
-          + " OR (c." + CloudBlobMetadata.FIELD_EXPIRATION_TIME + " BETWEEN 1 AND " + THRESHOLD_PARAM + ")";
+      "SELECT TOP " + LIMIT_PARAM + " * FROM c WHERE (c." + CloudBlobMetadata.FIELD_DELETION_TIME + " BETWEEN 1 AND " + THRESHOLD_PARAM + ")"
+          + " OR (c." + CloudBlobMetadata.FIELD_EXPIRATION_TIME + " BETWEEN 1 AND " + THRESHOLD_PARAM + ")"
+          + " ORDER BY c." + CloudBlobMetadata.FIELD_UPLOAD_TIME + " ASC";
+  static final String ENTRIES_SINCE_QUERY_TEMPLATE =
```
zzmao: If multiple blobs have the same time stamp and we do the same query multiple times, do blobs' order always same ?

`findEntriesSince()` skips blobs before `latestBlob`, if `latestBlob`'s position is non-deterministic, some blobs will be skipped incorrectly. 

Can we order by both timestamp and id?

zzmao: Another conner case is, in a very short time window but lots of blobs with same timestamp, after `findEntriesSince()`first query, new blob with same timestamp comes, if it has smaller alphabeta order than `latestBlob`, it will be skipped. (assumed we ordered by both time and id)

This looks like a race condition if we do bi-directional data synchronization.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudFindToken.java
lightningrob: I used the Intellij code generator so can't claim too much credit :)

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: I tried adding id to the order by clause but unfortunately CosmosDB doesn't support that:
`The order by query does not have a corresponding composite index that it can be served from.`
However, I changed the logic to only skip the lastBlobId in the token while iterating, not all the blobs *until* that one.  So nothing should be skipped now.
For what it's worth, the chance of an uploadTime collision for the same partition is extremely unlikely since the upload latency is over 100ms.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -306,9 +332,8 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
       List<CloudBlobMetadata> cappedResults = new ArrayList<>();
       boolean foundLastBlob = (findToken.getLatestBlobId() == null);
```
zzmao: Looks like no longer needed.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
zzmao: Cool. 
But let's add some comment for this unlikely scenario, in case we want to scale it, for example, multiple VCRs upload same partition.

zzmao:
Chunk part looks good to me.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: Added.

PR #1217, opened 2019-07-16
Allow NettyPerfClient to connect multiple frontends

jsjtzyy:
1. log response based on response code range rather than single response
code value.
2. allow perfClient to connect multiple frontends.

### DIFF
--- a/ambry-tools/src/main/java/com.github.ambry/tools/perf/rest/NettyPerfClient.java
+++ b/ambry-tools/src/main/java/com.github.ambry/tools/perf/rest/NettyPerfClient.java
@@ -30,6 +30,7 @@ import io.netty.bootstrap.Bootstrap;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInitializer;
@@ -47,7 +48,6 @@ import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpObject;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpResponse;
-import io.netty.handler.codec.http.HttpResponseStatus;
 import io.netty.handler.codec.http.HttpUtil;
 import io.netty.handler.codec.http.HttpVersion;
 import io.netty.handler.codec.http.LastHttpContent;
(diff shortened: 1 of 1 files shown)


ambry-tools/src/main/java/com.github.ambry/tools/perf/rest/NettyPerfClient.java
```diff
@@ -446,8 +462,15 @@ public void channelRead0(ChannelHandlerContext ctx, HttpObject in) {
...
-          logger.error("Got Response code {} and headers were {}", response.status().code(), response.headers());
+        if (response.status().code() >= 200 && response.status().code() < 300) {
+          logger.trace("Request succeeded");
+          if (response.headers().contains("Location")) {
+            logger.info(response.headers().get("Location"));
+          }
+        } else if (response.status().code() >= 300 && response.status().code() < 400) {
+          logger.warn("Redirection code {} and headers were {}", response.status().code(), response.headers());
+        } else {
+          logger.error("Response error code {} and headers were {}", response.status().code(), response.headers());
```
justinlin-linkedin: any reason to log these messages?

ambry-tools/src/main/java/com.github.ambry/tools/perf/rest/NettyPerfClient.java
```diff
@@ -355,20 +360,30 @@ protected void start() {
     b.group(group).channel(NioSocketChannel.class).handler(new ChannelInitializer<SocketChannel>() {
       @Override
       public void initChannel(SocketChannel ch) throws Exception {
+        channelConnectLatch.await();
         if (sslFactory != null) {
+          String host = channelToHost.get(ch);
+          if (host == null) {
+            throw new Exception("Unrecognized host!");
+          }
+          logger.info("Initializing the channel to {}", host);
```
justinlin-linkedin: this if statement and logging could be moved to the chnnalConnectListener so we can get rid of the latch wait here.

ambry-tools/src/main/java/com.github.ambry/tools/perf/rest/NettyPerfClient.java
jsjtzyy: Recording 50x error code might be useful to remind us there is something wrong on server side. 40x can drive us to double check if server behaves as expected.

ambry-tools/src/main/java/com.github.ambry/tools/perf/rest/NettyPerfClient.java
justinlin-linkedin: resolved by offline discussion

ambry-tools/src/main/java/com.github.ambry/tools/perf/rest/NettyPerfClient.java
```diff
@@ -136,11 +140,12 @@
      * Parses the arguments provided and extracts them into variables that can be retrieved.
      * @param args the command line argument list.
      */
-    protected ClientArgs(String args[]) {
+    ClientArgs(String[] args) {
```
zzmao: Minor: Use 'protected' to be more strict?

zzmao: Please ignore this. Looks good.

PR #1218, opened 2019-07-16
Misc changes for adaptive operation tracker

jsjtzyy:
1. add log info to record which type of adaptive tracker is used
2. enforce max number of inlight requests for adaptive tracker
3. make excluding timedout request configurable
4. support periodically dumping resource-level histogram to log file

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -407,7 +407,7 @@ public class RouterConfig {
     routerOperationTrackerMinDataPointsRequired =
         verifiableProperties.getLong("router.operation.tracker.min.data.points.required", 1000L);
     routerOperationTrackerMaxInflightRequests =
-        verifiableProperties.getIntInRange("router.adaptive.tracker.max.inflight.requests", 2, 1, Integer.MAX_VALUE);
+        verifiableProperties.getIntInRange("router.operation.tracker.max.inflight.requests", 2, 1, Integer.MAX_VALUE);
     routerOperationTrackerExcludeTimeoutEnabled =
         verifiableProperties.getBoolean("router.operation.tracker.exclude.timeout.enabled", false);
     routerOperationTrackerHistogramDumpEnabled =
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -301,10 +301,16 @@ public class RouterConfig {
   /**
    * The maximum number of inflight requests that allowed for adaptive tracker. If current number of inflight requests
    * is larger than or equal to this threshold, tracker shouldn't send out any request even though the oldest is past due.
+   * {@link RouterConfig#routerGetRequestParallelism} is a suggestive number that operation tracker uses to determine how
+   * many requests can be outstanding in parallel (assuming request gets response in time). Adaptive tracker is allowed
+   * to issue more requests (total inflight requests may exceed #routerGetRequestParallelism) if old request is past due.
+   * {@link RouterConfig#routerOperationTrackerMaxInflightRequests} is the strict upper bound that at any point of time,
+   * number of inflight requests issued by adaptive tracker should not exceed this number. Hence, for adaptive tracker,
+   * inflight requests number should always be within [0, #routerOperationTrackerMaxInflightRequests]
    */
-  @Config("router.adaptive.tracker.max.inflight.requests")
+  @Config("router.operation.tracker.max.inflight.requests")
   @Default("2")
-  public final int routerAdaptiveTrackerMaxInflightRequests;
+  public final int routerOperationTrackerMaxInflightRequests;
 
   /**
    * Indicates whether to enable excluding timed out requests in Histogram reservoir.
@@ -400,7 +406,7 @@ public class RouterConfig {
         verifiableProperties.getDouble("router.operation.tracker.reservoir.decay.factor", 0.015);
     routerOperationTrackerMinDataPointsRequired =
         verifiableProperties.getLong("router.operation.tracker.min.data.points.required", 1000L);
-    routerAdaptiveTrackerMaxInflightRequests =
+    routerOperationTrackerMaxInflightRequests =
         verifiableProperties.getIntInRange("router.adaptive.tracker.max.inflight.requests", 2, 1, Integer.MAX_VALUE);
     routerOperationTrackerExcludeTimeoutEnabled =
         verifiableProperties.getBoolean("router.operation.tracker.exclude.timeout.enabled", false);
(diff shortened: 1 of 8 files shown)


jsjtzyy:
will keep adding java docs

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
```diff
@@ -84,6 +84,10 @@
       localDcResourceToHistogram = getResourceToLatencyMap(routerOperation, true);
       crossDcResourceToHistogram = getResourceToLatencyMap(routerOperation, false);
     }
+    if (parallelism > routerConfig.routerAdaptiveTrackerMaxInflightRequests) {
```
cgtz: I think it would be good to also enforce this in `RouterConfig` constructor so that startup will fail if it is set improperly instead of failing continuously at runtime.

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -298,6 +298,35 @@
   @Default("1000")
   public final long routerOperationTrackerMinDataPointsRequired;
 
+  /**
+   * The maximum number of inflight requests that allowed for adaptive tracker. If current number of inflight requests
+   * is larger than or equal to this threshold, tracker shouldn't send out any request even though the oldest is past due.
```
cgtz: minor: Seems like all the other adaptive configs start with router.operation.tracker, even if they are only relevant to `AdaptiveOperationTracker`. Could we keep this the same?

cgtz: Could you also document the difference between parallelism and max.inflight.requests?

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
```diff
@@ -219,7 +224,8 @@ private Counter getWholeDcPastDueCounter(RouterOperation routerOperation) {
 
     @Override
     public boolean hasNext() {
-      return replicaIterator.hasNext() && (inflightCount < parallelism || isOldestRequestPastDue());
+      return replicaIterator.hasNext() && (inflightCount < parallelism || (
```
cgtz: It might be more readable to break this into multiple if statements:
```java
      if (replicaIterator.hasNext()) {
        if (inflightCount < parallelism) {
          return true;
        }
        if (inflightCount < routerConfig.routerAdaptiveTrackerMaxInflightRequests && isOldestRequestPastDue()) {
          return true;
        }
      }
      return false;
```

Up to your personal preference, though

ambry-router/src/main/java/com.github.ambry.router/NonBlockingRouterMetrics.java
```diff
@@ -892,6 +915,50 @@ void trackAgeAtAccess(long creationTimeMs) {
...
+
+  /**
+   * A thread that helps periodically dump resource-level histogram (with given percentile) into log file.
+   */
+  private class HistogramDumper implements Runnable {
+
+    @Override
+    public void run() {
+      double quantile = routerConfig.routerLatencyToleranceQuantile;
+      for (Map.Entry<Resource, Histogram> resourceToHistogram : getBlobLocalDcResourceToLatency.entrySet()) {
```
cgtz: I think all of these maps can be null.

jsjtzyy: Good catch! Will add check here.

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
jsjtzyy: sure, will do

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
jsjtzyy: Good point. Will make the change.

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
jsjtzyy: True, your code is more readable. I will take your advice.

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -400,13 +406,17 @@ public RouterConfig(VerifiableProperties verifiableProperties) {
         verifiableProperties.getDouble("router.operation.tracker.reservoir.decay.factor", 0.015);
     routerOperationTrackerMinDataPointsRequired =
         verifiableProperties.getLong("router.operation.tracker.min.data.points.required", 1000L);
-    routerAdaptiveTrackerMaxInflightRequests =
+    routerOperationTrackerMaxInflightRequests =
         verifiableProperties.getIntInRange("router.adaptive.tracker.max.inflight.requests", 2, 1, Integer.MAX_VALUE);
```
cgtz: change adaptive to operation here

ambry-router/src/test/java/com.github.ambry.router/AdaptiveOperationTrackerTest.java
```diff
@@ -652,6 +701,8 @@ private RouterConfig createRouterConfig(boolean crossColoEnabled, int successTar
     props.setProperty("router.get.replicas.required", Integer.toString(Integer.MAX_VALUE));
     props.setProperty("router.latency.tolerance.quantile", Double.toString(QUANTILE));
     props.setProperty("router.operation.tracker.metric.scope", trackerScope.toString());
+    props.setProperty("router.adaptive.tracker.max.inflight.requests", Integer.toString(maxInflightNum));
```
cgtz: also change here after changing config key

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
```diff
@@ -555,6 +555,7 @@ private OperationTracker getOperationTracker(boolean crossColoEnabled, int succe
         Boolean.toString(includeNonOriginatingDcReplicas));
     props.setProperty("router.get.replicas.required", Integer.toString(replicasRequired));
     props.setProperty("router.latency.tolerance.quantile", Double.toString(QUANTILE));
+    props.setProperty("router.adaptive.tracker.max.inflight.requests", Integer.toString(parallelism));
```
cgtz: and here

jsjtzyy: sorry, will fix soon

ambry-router/src/main/java/com.github.ambry.router/NonBlockingRouterMetrics.java
```diff
@@ -892,6 +911,50 @@ void trackAgeAtAccess(long creationTimeMs) {
...
+            quantile * 100, histogram.getSnapshot().getValue(quantile));
+      }
+      for (Map.Entry<Resource, Histogram> resourceToHistogram : getBlobCrossDcResourceToLatency.entrySet()) {
+        Resource resource = resourceToHistogram.getKey();
+        Histogram histogram = resourceToHistogram.getValue();
+        logger.info("{} GetBlob cross DC latency histogram {}th percentile in ms: {}", resource.toString(),
+            quantile * 100, histogram.getSnapshot().getValue(quantile));
+      }
+      for (Map.Entry<Resource, Histogram> resourceToHistogram : getBlobInfoLocalDcResourceToLatency.entrySet()) {
+        Resource resource = resourceToHistogram.getKey();
```
zzmao: Create a function, method or class to make a general histgram dumper?

jsjtzyy: Sure but I need to think about this. Will make the change in future PR.

PR #1222, opened 2019-07-23
Passing IfModifiedSince to amybry server

justinlin-linkedin:
Passing IfModifiedSince http header to ambry-server so when the content hasn't been updated, server doesn't have to return the content back, rather just return a blob_not_modified server code.
This is the implementation on the server side. After all servers are upgraded, we need to change the implementation on the client side to utilize this new logic.

### DIFF
--- a/ambry-server/src/main/java/com.github.ambry.server/AmbryRequests.java
+++ b/ambry-server/src/main/java/com.github.ambry.server/AmbryRequests.java
@@ -381,9 +381,27 @@ public class AmbryRequests implements RequestAPI {
           }
         }
       }
-      CompositeSend compositeSend = new CompositeSend(messagesToSendList);
-      response = new GetResponse(getRequest.getCorrelationId(), getRequest.getClientId(), partitionResponseInfoList,
-          compositeSend, ServerErrorCode.No_Error);
+      if (getRequest.getIfModifiedSince() != GetRequest.EMPTY_IF_MODIFED_SINCE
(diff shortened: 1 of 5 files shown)


lightningrob:
amybry?

ambry-server/src/main/java/com.github.ambry.server/AmbryRequests.java
```diff
@@ -381,9 +381,27 @@ public void handleGetRequest(Request request) throws IOException, InterruptedExc
...
+          // 2. GetRequest includes deleted blob and this blob is deleted. Then the operationTimeMs would still be the
+          // creationTimeMs since this is guaranteed by the the implementation of BlobStore.
+          MessageInfo info = partitionResponseInfoList.get(0).getMessageInfoList().get(0);
+          long creationTime = info.getOperationTimeMs();
+          long creationTimeInSecond = creationTime - creationTime % 1000;
+          if (creationTimeInSecond <= getRequest.getIfModifiedSince()) {
+            response = new GetResponse(getRequest.getCorrelationId(), getRequest.getClientId(), ServerErrorCode.Blob_Not_Modified);
+          }
+      }
+      if (response == null) {
```
lightningrob: This can be an else condition to previous if-clause.

justinlin-linkedin:
Offline discussion with Casey, the benefits of this PR is debatable at best. Will close this one for now.

PR #1223, opened 2019-07-24
Set permissions for ambry data and operation files

jsjtzyy:
For security concern, data files and operation files are required to be
set with specified permissions. For example, user data related files
should be set "rw-rw----" and operation files (usually generated by
Ambry itself and don't contain user data/metadata) are allowed to set
"rw-rw-r--". The permissions are configurable and could be changed in
StoreConfig.


ambry-utils/src/main/java/com.github.ambry.utils/Utils.java
```diff
@@ -890,6 +893,18 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc) throws IOEx
     }
   }
 
+  /**
+   * Set permissions for given files.
+   * @param files a list of files that set specified permissions
+   * @param permissions a set of {@link PosixFilePermission} associated with given files
+   * @throws IOException
+   */
+  public static void setFilePermission(List<File> files, Set<PosixFilePermission> permissions) throws IOException {
```
lightningrob: Since most of this method's callers pass only one file, I would add a version that accepts a single file, to avoid wrapping in a collection each time.

jsjtzyy: Right. For single file case, looks like there is no need to write another version, I can directly use `Files.setPosixFilePermissions` method.

jsjtzyy:
Verified on Red Hat 7 machine. @zzmao reminder to review. Thanks!

ambry-utils/src/main/java/com.github.ambry.utils/Utils.java
zzmao: `setFilesPermission`?

PR #1224, opened 2019-07-25
Ensure blobId with prefix in request header is correctly handled

jsjtzyy:
In original code, GetSignedUrlHandler and TtlUpdateHandler didn't
remove cluster name prefix (if exists) in blob id that specified in
RestRequest Header. The whole blob id string (including prefix) is
passed into IdConverter which may cause blob id cannot be identified.
This PR fixes such issue by parsing valid blob id from blob id string
in request header.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
+++ b/ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
@@ -57,7 +57,7 @@ public class RequestPath {
       // to ensure IdConverter and IdSigningService can receive a valid blob id and correctly identify it.
       String blobIdStr =
           parse(blobIdHeader, Collections.emptyMap(), prefixesToRemove, clusterName).getOperationOrBlobId(false);
-      restRequest.setArg(RestUtils.Headers.BLOB_ID, blobIdStr);
+      restRequest.setArg(Headers.BLOB_ID, blobIdStr);
     }
     return parse(restRequest.getPath(), restRequest.getArgs(), prefixesToRemove, clusterName);
   }
--- a/ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
+++ b/ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
@@ -16,8 +16,7 @@ package com.github.ambry.rest;
 import java.util.List;
 import java.util.Map;
 
-import static com.github.ambry.rest.RestUtils.SubResource;
-
+import static com.github.ambry.rest.RestUtils.*;
 import static com.github.ambry.router.GetBlobOptions.*;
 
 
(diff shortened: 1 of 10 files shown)


ambry-frontend/src/main/java/com.github.ambry.frontend/AmbryBlobStorageService.java
```diff
@@ -146,13 +146,13 @@ public void start() throws InstantiationException {
     getPeersHandler = new GetPeersHandler(clusterMap, securityService, frontendMetrics);
     getSignedUrlHandler =
         new GetSignedUrlHandler(urlSigningService, securityService, idConverter, accountAndContainerInjector,
-            frontendMetrics, clusterMap);
+            frontendMetrics, clusterMap, clusterName, frontendConfig);
     postBlobHandler =
         new PostBlobHandler(securityService, idConverter, idSigningService, router, accountAndContainerInjector,
```
cgtz: Unfortunately, I think there is also an issue with stitchBlob that needs to be fixed in PostBlobHandler.
```
      for (String signedChunkId : signedChunkIds) {
        signedChunkId = RequestPath.parse(signedChunkId, Collections.emptyMap(), frontendConfig.pathPrefixesToRemove, clusterName).getOperationOrBlobId(false);
```

ambry-api/src/main/java/com.github.ambry/rest/RequestPath.java
```diff
@@ -50,6 +49,15 @@
    */
   public static RequestPath parse(RestRequest restRequest, List<String> prefixesToRemove, String clusterName)
       throws RestServiceException {
+    String blobIdHeader = RestUtils.getHeader(restRequest.getArgs(), Headers.BLOB_ID, false);
+    if (blobIdHeader != null) {
+      // If blob id is specified in the rest request header and it has cluster name prefix, we remove its prefix and
+      // use this id (with no prefix now) to update "x-ambry-blob-id" header. The purpose is to ensure IdConverter and
+      // IdSigningService can receive a valid blob id and correctly identify it.
+      int nextOffset = matchPathSegments(blobIdHeader, 0, clusterName, true);
+      String blobIdStr = nextOffset >= 0 ? blobIdHeader.substring(nextOffset) : blobIdHeader;
```
cgtz: Could you add a test for this to RequestPathTest?

jsjtzyy: will do

cgtz: Let's do it this way so that it is consistent with signedChunkId: 
```
    String blobIdHeader = RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.BLOB_ID, false);
    if (blobIdHeader != null) {
      String blobIdStr =
          parse(blobIdHeader, Collections.emptyMap(), prefixesToRemove, clusterName).getOperationOrBlobId(false);
      restRequest.setArg(RestUtils.Headers.BLOB_ID, blobIdStr);
    }
```
This will also ensure sub resources are stripped off, which our code can't handle

ambry-frontend/src/test/java/com.github.ambry.frontend/FrontendIntegrationTest.java
```diff
@@ -561,7 +565,7 @@ private static VerifiableProperties buildFrontendVProps(File trustStoreFile)
     TestSSLUtils.addSSLProperties(properties, "", SSLFactory.Mode.SERVER, trustStoreFile, "frontend");
```
cgtz: Could you also change uploadDataChunksAndVerify to add the prefixes to signedChunkIds

PR #1225, opened 2019-07-25
Add cache of recently seen blobs in CloudBlobStore

lightningrob:
To reduce unnecessary Azure round trips when multiple replicas send same data.


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -88,6 +98,7 @@
     this.vcrMetrics = Objects.requireNonNull(vcrMetrics, "vcrMetrics is required");
     minTtlMillis = TimeUnit.DAYS.toMillis(cloudConfig.vcrMinTtlDays);
     requireEncryption = cloudConfig.vcrRequireEncryption;
+    recentBlobCache = Collections.synchronizedMap(new RecentBlobCache(1000, cloudConfig.recentBlobCacheLimit));
```
dharju: Make `1000` a static final int

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -383,4 +428,21 @@ public void appendFrom(ReadableByteChannel channel, long size) throws StoreExcep
...
     }
   }
+
+  /**
+   * A local LRU cache of recent blobs processed by this store.
+   */
+  private class RecentBlobCache extends LinkedHashMap<String, BlobState> {
+    private final int maxEntries;
+
+    public RecentBlobCache(int initialCapacity, int maxEntries) {
```
dharju: The intent is to use the insertion ordered LinkedHashMap not the accessed-order LinkedHashMap, right?

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -221,6 +235,15 @@ public void testFindMissingKeys() throws Exception {
...
     assertEquals("Wrong number of missing keys", count, missingKeys.size());
+
+    // Add keys to cache and rerun (should be cached)
+    for (StoreKey storeKey : keys) {
+      store.addToCache(storeKey.getID(), CloudBlobStore.BlobState.CREATED);
+    }
+    missingKeys = store.findMissingKeys(keys);
+    assertTrue("Expected no missing keys", missingKeys.isEmpty());
+    // getBlobMetadata should not have been called a second time.
+    verify(dest).getBlobMetadata(anyList());
```
dharju: Perhaps a unit test to confirm that the LRU nature of the cache works as expected (set the cache limit artificially low, add PUTs over that number, confirm that redoing the PUTs yields (number of PUTs - limit) unskipped cloud PUTs)?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
lightningrob: Changed to access order and added test.

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
lightningrob: Done (method a bit different).

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -59,12 +61,20 @@
...
+  private static final float cacheLoadFactor = 0.75f;
   private final PartitionId partitionId;
   private final CloudDestination cloudDestination;
   private final ClusterMap clusterMap;
   private final CloudBlobCryptoAgentFactory cryptoAgentFactory;
   private final CloudBlobCryptoAgent cryptoAgent;
   private final VcrMetrics vcrMetrics;
+
+  /** The lifecycle state of a recently seen blob. */
+  enum BlobState {CREATED, TTL_UPDATED, DELETED}
```
zzmao: Minior: move `BlobState` to `RecentBlobCache`?

lightningrob: Can't because of dependency, but moved enum to bottom of class.

PR #1227, opened 2019-07-30
Create executor pool to take over network read/write tasks.

zzmao:
SSL transmission latency was high in recent performance test, especially in PUT. 
The reason is selector.poll() needs to iterate all events and do SSL encryption + network IO for each of them. This increased round trip time dramatically in the queuing system.
The change introduced an executor pool to take over event handling and selector.poll() blocks until events done.  
In the performance test, 7MB blobs were sent at 20QPS, PUT latency P95 was decreased to 380ms from 1.4kms with this change.

### DIFF
--- a/ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
+++ b/ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
@@ -160,7 +160,9 @@ public class SSLTransmission extends Transmission implements ReadableByteChannel
   private boolean flush(ByteBuffer buf) throws IOException {
     int remaining = buf.remaining();
     if (remaining > 0) {
+      long start = SystemTime.getInstance().nanoseconds();
       int written = socketChannel.write(buf);
+      logger.trace("Flushed {} bytes in {} ns", written, SystemTime.getInstance().nanoseconds() - start);
       return written >= remaining;
     }
     return true;
@@ -531,13 +533,12 @@ public class SSLTransmission extends Transmission implements ReadableByteChannel
     int written = 0;
     while (src.remaining() != 0) {
       netWriteBuffer.clear();
-      long startTimeMs = SystemTime.getInstance().milliseconds();
+      long startTimeNs = SystemTime.getInstance().nanoseconds();
       SSLEngineResult wrapResult = sslEngine.wrap(src, netWriteBuffer);
-      long encryptionTimeMs = SystemTime.getInstance().milliseconds() - startTimeMs;
-      logger.trace("SSL encryption time: {} ms for {} bytes", encryptionTimeMs, wrapResult.bytesConsumed());
+      long encryptionTimeNs = SystemTime.getInstance().nanoseconds() - startTimeNs;
+      logger.trace("SSL encryption time: {} ns for {} bytes", encryptionTimeNs, wrapResult.bytesConsumed());
       if (wrapResult.bytesConsumed() > 0) {
-        metrics.sslEncryptionTimeInUsPerKB.update(
-            TimeUnit.MILLISECONDS.toMicros(encryptionTimeMs) * 1024 / wrapResult.bytesConsumed());
+        metrics.sslEncryptionTimeInUsPerKB.update(encryptionTimeNs / wrapResult.bytesConsumed());
       }
       netWriteBuffer.flip();
       //handle ssl renegotiation
--- a/ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
+++ b/ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
@@ -160,9 +160,9 @@ public class SSLTransmission extends Transmission implements ReadableByteChannel
   private boolean flush(ByteBuffer buf) throws IOException {
     int remaining = buf.remaining();
     if (remaining > 0) {
-      long start = SystemTime.getInstance().nanoseconds();
+      long startNs = SystemTime.getInstance().nanoseconds();
       int written = socketChannel.write(buf);
-      logger.trace("Flushed {} bytes in {} ns", written, SystemTime.getInstance().nanoseconds() - start);
+      logger.trace("Flushed {} bytes in {} ns", written, SystemTime.getInstance().nanoseconds() - startNs);
       return written >= remaining;
     }
     return true;
--- a/ambry-network/src/main/java/com.github.ambry.network/Selector.java
+++ b/ambry-network/src/main/java/com.github.ambry.network/Selector.java
@@ -32,6 +32,10 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicLong;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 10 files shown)


ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -556,16 +590,20 @@ private void read(SelectionKey key, Transmission transmission) throws IOExceptio
         transmission.onReceiveComplete();
         transmission.clearReceive();
       }
+    } catch (IOException e) {
+      handleReadWriteIOException(e, key);
```
cgtz: Let's handle this from within the poll method since `close` modifies thread unsafe data structs

ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -556,16 +590,20 @@ private void read(SelectionKey key, Transmission transmission) throws IOExceptio
         transmission.onReceiveComplete();
```
cgtz: let's have this method return the NetworkReceive if `readComplete` and then use the future result to modify completedReceives from the main thread.

ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -291,13 +301,17 @@ public void poll(long timeoutMs) throws IOException {
...
+   * @param sends The list of new sends to initiate.
    *
    * @throws IOException If a send is given for which we have no existing connection or for which there is
    *         already an in-progress send
    */
   @Override
   public void poll(long timeoutMs, List<NetworkSend> sends) throws IOException {
+    if (executorPool != null) {
+      pollWithExecutorPool(timeoutMs, sends);
+      return;
```
cgtz: minor: move the code below line 314 into another method like `pollOnMainThread()` and use an if-else to switch between them.

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
```diff
@@ -531,13 +533,12 @@ public int write(ByteBuffer src) throws IOException {
...
+      long startTimeNs = SystemTime.getInstance().nanoseconds();
       SSLEngineResult wrapResult = sslEngine.wrap(src, netWriteBuffer);
-      long encryptionTimeMs = SystemTime.getInstance().milliseconds() - startTimeMs;
-      logger.trace("SSL encryption time: {} ms for {} bytes", encryptionTimeMs, wrapResult.bytesConsumed());
+      long encryptionTimeNs = SystemTime.getInstance().nanoseconds() - startTimeNs;
+      logger.trace("SSL encryption time: {} ns for {} bytes", encryptionTimeNs, wrapResult.bytesConsumed());
       if (wrapResult.bytesConsumed() > 0) {
-        metrics.sslEncryptionTimeInUsPerKB.update(
-            TimeUnit.MILLISECONDS.toMicros(encryptionTimeMs) * 1024 / wrapResult.bytesConsumed());
+        metrics.sslEncryptionTimeInUsPerKB.update(encryptionTimeNs / wrapResult.bytesConsumed());
```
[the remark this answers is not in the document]
zzmao: sslEncryptionTimeInUsPerKB = Us/KB = (ns/1000) / (bytes/1024) = ns / bytes. Looks good, no?

ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -544,18 +677,38 @@ private SelectionKey keyForId(String id) {
...
+   * Handle read/write IOException
+   */
+  private void handleReadWriteIOException(IOException e, SelectionKey key) {
+    String socketDescription = socketDescription(channel(key));
+    if (e instanceof EOFException || e instanceof ConnectException) {
+      metrics.selectorDisconnectedErrorCount.inc();
+      logger.error("Connection {} disconnected", socketDescription, e);
+    } else {
+      metrics.selectorIOErrorCount.inc();
+      logger.warn("Error in I/O with connection to {}", socketDescription, e);
```
jsjtzyy: why here we use `warn` instead of `error` level?

zzmao: It was `warn` before. It may cause by network issue and is not critical to us.

ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -291,13 +301,24 @@ public void poll(long timeoutMs) throws IOException {
...
+      pollWithExecutorPool(timeoutMs, sends);
+    } else {
+      pollOnMainThread(timeoutMs, sends);
+    }
+  }
+
+  /**
+   * Process read/write events on current thread.
+   */
+  public void pollOnMainThread(long timeoutMs, List<NetworkSend> sends) throws IOException {
```
jsjtzyy: can be `private` here

ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -544,18 +677,38 @@ private SelectionKey keyForId(String id) {
...
+      logger.warn("Error in I/O with connection to {}", socketDescription, e);
+    }
+  }
+
   /**
    * Process reads from ready sockets
+   * @return the {@link NetworkReceive} if no IOException during read().
    */
-  private void read(SelectionKey key, Transmission transmission) throws IOException {
+  private NetworkReceive read(SelectionKey key, Transmission transmission) throws IOException {
```
jsjtzyy: both `read` and `write` methods can remove `throws IOException` since there is no exception thrown from these two methods.

ambry-network/src/main/java/com.github.ambry.network/Selector.java
```diff
@@ -370,6 +396,114 @@ public void poll(long timeoutMs, List<NetworkSend> sends) throws IOException {
...
+        try {
+          NetworkReceive networkReceive = future.get();
+          if (networkReceive != null) {
+            readWriteKeySet.remove(keyForId(networkReceive.getConnectionId()));
+            if (networkReceive.getReceivedBytes().isReadComplete()) {
+              this.completedReceives.add(networkReceive);
+            }
+          }
+        } catch (InterruptedException | ExecutionException e) {
+          if (!(e.getCause() instanceof IOException)) {
```
jsjtzyy: Why not catch `IOException` and `Exception` separately?

ambry-network/src/test/java/com.github.ambry.network/SSLSelectorTest.java
```diff
@@ -307,6 +318,7 @@ private String blockingSSLConnect(int socketBufSize) throws IOException {
     String connectionId =
         selector.connect(new InetSocketAddress("localhost", server.port), socketBufSize, socketBufSize, PortType.SSL);
     while (!selector.connected().contains(connectionId)) {
+      System.out.println("here");
```
jsjtzyy: remove?

ambry-network/src/main/java/com.github.ambry.network/SSLTransmission.java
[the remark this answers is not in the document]
cgtz: I think it's only roughly equivalent if you use KiB, since its actually 1 Us/KiB = 1000/1024 ns/bytes.

If you use power of 10 KB, then they would be equivalent

ambry-network/src/test/java/com.github.ambry.network/SSLSelectorTest.java
```diff
@@ -327,7 +339,7 @@ private void useCustomBufferSizeSelector(final Integer netReadBufSizeStart, fina
     selector.close();
     NetworkMetrics metrics = new NetworkMetrics(new MetricRegistry());
     Time time = SystemTime.getInstance();
-    selector = new Selector(metrics, time, clientSSLFactory) {
+    selector = new Selector(metrics, time, clientSSLFactory, 0) {
```
jsjtzyy: You can pass in different `poolSize` to this method.

ambry-network/src/main/java/com.github.ambry.network/Selector.java
zzmao: fixed.

ambry-network/src/main/java/com.github.ambry.network/Selector.java
zzmao: fixed.

ambry-network/src/main/java/com.github.ambry.network/Selector.java
cgtz: copy logic from write future get()

zzmao: fixed.

PR #1228, opened 2019-08-05
Introducing Ambry partition state model and update Helix tool

jsjtzyy:
1. Introduce new state model that incorporates BOOTSTRAP and INACTIVE
states.
2. Add clustermap config that allows participant to choose which state
model to use. Previous LeadStandby Model is default option.
3. Update Helix bootstrap tool to supporting specifying state model when
creating new cluster or adding new state model to existing cluster.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 LinkedIn Corp. All rights reserved.
+ * Copyright 2019 LinkedIn Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryStateModelFactory.java
```diff
@@ -13,25 +13,38 @@
...
    */
   @Override
   public StateModel createNewStateModel(String resourceName, String partitionName) {
-    return new AmbryStateModel();
+    StateModel stateModelToReturn;
+    switch (ambryStateModelDef) {
+      case AmbryStateModelDefinition.AMBRY_LEADER_STANDBY_MODEL:
+        stateModelToReturn = new AmbryPartitionStateModel();
+        break;
+      default:
```
jiajunwang: Better to explicitly check DEFAULT_STATE_MODEL_DEF here. For the other cases, shall throw an exception the earlier to better.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.helix.participant.statemachine.StateModel;
+import org.apache.helix.participant.statemachine.StateModelInfo;
+import org.apache.helix.participant.statemachine.StateModelParser;
+import org.apache.helix.participant.statemachine.Transition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+@StateModelInfo(initialState = "OFFLINE", states = {"BOOTSTRAP", "LEADER", "STANDBY", "INACTIVE"})
+public class AmbryPartitionStateModel extends StateModel {
```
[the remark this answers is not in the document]
jsjtzyy: Right, will add that method. Thanks for reminding me of this.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryStateModelFactory.java
jsjtzyy: Sure, I will also add a test case for the change.

PR #1229, opened 2019-08-05
Implement get functionality for Azure Vcr

ankagrawal:
This review is the implementation of get functionality for Azure Vcr. I am still planning to add a few test cases.


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -252,6 +255,31 @@ private boolean uploadIfNotExists(BlobId blobId, long inputLength, CloudBlobMeta
...
+  @Override
+  public BlobReadInfo downloadBlob(BlobId blobId) throws CloudStorageException {
+    azureMetrics.blobDownloadRequestCount.inc();
+    Timer.Context storageTimer = azureMetrics.blobDownloadTime.time();
+    BlobReadInfo blobReadInfo = null;
+    try {
+      List<BlobId> blobList = new ArrayList<>();
+      blobList.add(blobId);
+      CloudBlobMetadata blobMetadata = getBlobMetadata(blobList).values().iterator().next();
+      ByteBuffer outputBuffer = ByteBuffer.allocate((int)blobMetadata.getSize());
```
ankagrawal: I am assuming here that blob sizes are 4-8MB, so this type conversion from long to int will not lead to any loss of data. If we support very large blobs such that their size is greater than Integer.MAX_VALUE, then this logic needs a rethink. But I am assuming that in such case, a lot of other things might need reconsideration.
@lightningrob  Please let me know if you agree with this assumption.

ankagrawal: Also in order to be safe, I would recommend changing the size attribute to be an integer instead of long. Will make everything cleaner. I can create a separate ticket for that.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
```diff
@@ -37,6 +37,14 @@
   boolean uploadBlob(BlobId blobId, long inputLength, CloudBlobMetadata cloudBlobMetadata, InputStream blobInputStream)
       throws CloudStorageException;
 
+  /**
+   * Download blob from the cloud destination.
+   * @param blobId id of the Ambry blob to be downloaded
+   * @param outputStream outputstream to populate the downloaded data with
+   * @throws CloudStorageException if the download encounters an error.
+   */
+  void downloadBlob(BlobId blobId, OutputStream outputStream) throws CloudStorageException;
```
lightningrob: Was expecting this to return CloudBlobMetadata.  Does that require a separate call?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -0,0 +1,135 @@
...
+import com.github.ambry.utils.ByteBufferOutputStream;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.WritableByteChannel;
+import java.util.List;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+class BlobReadInfo {
```
lightningrob: Can this be static inner class?
Also javadoc

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -0,0 +1,135 @@
...
+  private final CloudBlobMetadata blobMetadata;
+  private final BlobId blobId;
+  private ByteBuffer prefetchedBuffer;
+  private boolean isPrefetched;
+
+  public BlobReadInfo(CloudBlobMetadata blobMetadata, BlobId blobId) {
```
lightningrob: javadoc all methods

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -0,0 +1,135 @@
...
+
+  public void prefetchBlob(CloudDestination cloudDestination) throws CloudStorageException {
+    prefetchedBuffer = ByteBuffer.allocate((int) blobMetadata.getSize());
+    ByteBufferOutputStream outputStream = new ByteBufferOutputStream(prefetchedBuffer);
+    cloudDestination.downloadBlob(blobId, outputStream);
+    isPrefetched = true;
+  }
+
+  public ByteBuffer downloadBlob(CloudDestination cloudDestination) throws CloudStorageException {
+    ByteBuffer blobByteBuffer = ByteBuffer.allocate((int) blobMetadata.getSize());
```
lightningrob: Duplicate code in these two methods.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -0,0 +1,135 @@
...
+    validateIndex(index);
+    long written = 0;
+    BlobReadInfo blobReadInfo = blobReadInfoList.get(index);
+
+    ByteBuffer outputBuffer;
+    try {
+      if (blobReadInfo.isPrefetched()) {
+        outputBuffer = blobReadInfo.getPrefetchedBuffer();
+      } else {
+        outputBuffer = blobReadInfo.downloadBlob(cloudDestination);
```
lightningrob: In the non-prefetch case, we shouldn't need to write the outputsteam to an in memory buffer that will be passed to channel.  It can be streamed instead.

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +615,120 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+    blobIds.add(blobIdList.get(0));
+    StoreInfo storeInfo = cloudBlobStore.get(blobIds, null);
+    assert (storeInfo.getMessageReadSetInfo().get(0).getStoreKey().equals(blobIdList.get(0)));
+
+    //try get for a list of blobs that exist
+    blobIds = new ArrayList<>(2);
+    blobIds.add(blobIdList.get(0));
+    blobIds.add(blobIdList.get(1));
+    storeInfo = cloudBlobStore.get(blobIds, null);
+    assert (storeInfo.getMessageReadSetInfo().get(0).getStoreKey().equals(blobIdList.get(0)));
```
zzmao: We usually use `org.junit.Assert.*`. You can use `assertTrue` here.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -252,6 +252,23 @@ private boolean uploadIfNotExists(BlobId blobId, long inputLength, CloudBlobMeta
...
+    azureMetrics.blobDownloadRequestCount.inc();
+    Timer.Context storageTimer = azureMetrics.blobDownloadTime.time();
+    try {
+      CloudBlobContainer azureContainer = getContainer(blobId, false);
+      String azureBlobName = getAzureBlobName(blobId);
+      CloudBlockBlob azureBlob = azureContainer.getBlockBlobReference(azureBlobName);
+      azureBlob.download(outputStream);
+      azureMetrics.blobUploadSuccessCount.inc();
+    } catch (URISyntaxException | StorageException e) {
+      throw new CloudStorageException("Error donwloading blob " + blobId, e);
```
zzmao: typo . donwloading--> downloading

ambry-cloud/src/test/java/com.github.ambry.cloud/LatchBasedInMemoryCloudDestination.java
```diff
@@ -75,6 +75,22 @@ synchronized public boolean uploadBlob(BlobId blobId, long blobSize, CloudBlobMe
     return true;
   }
 
+  @Override
+  public void downloadBlob(BlobId blobId, OutputStream outputStream) throws CloudStorageException {
+    BlobReadInfo blobReadInfo = null;
```
zzmao: Not used.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -114,7 +114,34 @@ public void start() throws StoreException {
...
+    checkStarted();
+    if (ids.size() > 1 && ids.size() != new HashSet<>(ids).size()) {
+      throw new IllegalArgumentException("The list of IDs provided contains duplicates");
+    }
+    List<BlobReadInfo> blobReadInfos = new ArrayList<>(ids.size());
+    List<MessageInfo> messageInfos = new ArrayList<>(ids.size());
+    try {
+      List<BlobId> blobIdList = ids.stream().map(key -> (BlobId) key).collect(Collectors.toList());
+      Map<String, CloudBlobMetadata> cloudBlobMetadataListMap = cloudDestination.getBlobMetadata(blobIdList);
+      if (cloudBlobMetadataListMap.size() < ids.size()) {
```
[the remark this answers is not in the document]
ankagrawal: I think that BlobStore's get method does return an exception if one of the keys is not found. Basically BlobStore::get calls PersistentIndex::getBlobReadInfo for each id, which in turn tries to find the id in the index. If the id is not found in the index, it seems to be throwing StoreException.

In our case it would be more efficient to get all the metadata for the list of ids in one cosmos call. If we do it in the CloudBlobStore::get() method, then we will know if the blobs are present, and be able to throw exception in case they aren't. This will be in line with the behavior of BlobStore::get. 

The actual CloudDestination::downloadBlob happens either during prefetch, or when the response if being returned to frontend.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
ankagrawal: This doesn't return a CloudBlobMetadata because we already obtained it during BlobStore::get. This is the actual download of blob from blob store, which happens either during prefetch or when the response is being returned to frontend.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: sure.

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
ankagrawal: sure.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
ankagrawal: sure

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: will fix this.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: sure

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: sure

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -114,7 +114,34 @@ public void start() throws StoreException {
...
+        Set<BlobId> missingBlobs = blobIdList.stream()
+            .filter(blobId -> !cloudBlobMetadataListMap.containsKey(blobId))
+            .collect(Collectors.toSet());
+        throw new StoreException("Some of the keys were missing in the cloud metadata store: " + missingBlobs,
+            StoreErrorCodes.IOError);
+      }
+      for (BlobId blobId : blobIdList) {
+        CloudBlobMetadata blobMetadata = cloudBlobMetadataListMap.get(blobId.getID());
+        MessageInfo messageInfo = new MessageInfo(blobId, blobMetadata.getSize(), blobMetadata.getExpirationTime(),
+            (short) blobMetadata.getAccountId(), (short) blobMetadata.getContainerId(), 0);
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -114,7 +114,34 @@ public void start() throws StoreException {
 
   @Override
   public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> storeGetOptions) throws StoreException {
-    throw new UnsupportedOperationException("Method not supported");
+    checkStarted();
+    if (ids.size() > 1 && ids.size() != new HashSet<>(ids).size()) {
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -368,20 +446,52 @@ private void checkStarted() throws StoreException {
...
       }
     }
   }
 
+  /**
+   * Get duplicates in a list of {@code StoreKey}
+   * @param keys keys list of {@link StoreKey} to find duplicates in
+   * @return {@code Set} of duplicate {@code StoreKey}
+   */
+  private Set<StoreKey> getDuplicates(List<? extends StoreKey> keys) {
```
lightningrob: Do we need this method?  I think we can move the logic to checkDuplicates(List<StoreKey>) and call that from checkDuplicates(MessageWriteSet)

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -114,7 +114,85 @@ public void start() throws StoreException {
...
+  }
+
+  /**
+   * Check if the blob is expired
+   * @param metadata to check for expiration
+   * @return true if expired. false otherwise
+   */
+  static boolean isBlobExpired(CloudBlobMetadata metadata) {
+    return metadata.getExpirationTime() != Utils.Infinite_Time
+        && metadata.getExpirationTime() < System.currentTimeMillis();
```
lightningrob: Not good to get system time in method that is called in a loop.  Please pass system time to method and call once at top of loop.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -114,7 +114,85 @@ public void start() throws StoreException {
...
+   * @param metadata blob metadata from which to derive operation time.
+   * @return operation time.
+   */
+  private long getOperationTime(CloudBlobMetadata metadata) {
+    if (isBlobDeleted(metadata)) {
+      return metadata.getDeletionTime();
+    } else if (isBlobExpired(metadata)) {
+      return metadata.getExpirationTime();
+    }
+    return metadata.getCreationTime();
```
lightningrob: Unfortunately, many blobs uploaded to Azure have creation time = -1.  (No way you could have known that.) . Please check for this and use upload time as backup.
I'm not sure how the operation time is being used downstream, but what we have in CloudBlobMetadata may not be that meaningful.  For example, deletion time isn't the same as operation time (time deletion was issued), and expiration time is set separately in the MessageInfo.  Need to think more about this.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
lightningrob: What you're saying makes sense.  But I'd rather not see the interface contract be designed based on the particulars of how the callers are using it.  This is okay for the recovery flow.  But when we are live serving blobs from Azure, a GET request will want to make a single call to get blob data and metadata (no initial call to getBlobMetadata().

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -0,0 +1,174 @@
...
+        written = channel.write(outputBuffer);
+      } else {
+        blobReadInfo.downloadBlob(cloudDestination, Channels.newOutputStream(channel));
+        written = blobReadInfo.getBlobSize();
+      }
+    } catch (CloudStorageException ex) {
+      throw new IOException("Download of cloud blob " + blobReadInfoList.get(index).getBlobId().getID() + " failed");
+    }
+    logger.trace("Downloaded {} bytes to the write channel from the cloud blob : {}", written,
+        blobReadInfoList.get(index).getBlobId().getID());
```
lightningrob: Minor: can use blobReadInfo

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -0,0 +1,174 @@
...
+      this.isPrefetched = false;
+    }
+
+    /**
+     * Prefetch the {@code blob} from {@code CloudDestination} and put it in {@code prefetchedBuffer}
+     * @param cloudDestination {@code CloudDestination} implementation representing the cloud from which download will happen.
+     * @throws CloudStorageException if blob cloud not be downloaded
+     */
+    public void prefetchBlob(CloudDestination cloudDestination) throws CloudStorageException {
+      prefetchedBuffer = ByteBuffer.allocate((int) blobMetadata.getSize());
```
lightningrob: Minor: add comment explaining that the casting is okay because the blobs are chunked in the cloud, but that if we ever decide to upload entire blobs (and let Azure do the chunking for us), then the prefetching logic will have to change.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -252,6 +252,23 @@ private boolean uploadIfNotExists(BlobId blobId, long inputLength, CloudBlobMeta
...
+  @Override
+  public void downloadBlob(BlobId blobId, OutputStream outputStream) throws CloudStorageException {
+    azureMetrics.blobDownloadRequestCount.inc();
+    Timer.Context storageTimer = azureMetrics.blobDownloadTime.time();
+    try {
+      CloudBlobContainer azureContainer = getContainer(blobId, false);
+      String azureBlobName = getAzureBlobName(blobId);
+      CloudBlockBlob azureBlob = azureContainer.getBlockBlobReference(azureBlobName);
+      azureBlob.download(outputStream);
+      azureMetrics.blobUploadSuccessCount.inc();
```
lightningrob: downloadSuccess

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -252,6 +252,23 @@ private boolean uploadIfNotExists(BlobId blobId, long inputLength, CloudBlobMeta
...
+    } catch (URISyntaxException | StorageException e) {
+      throw new CloudStorageException("Error downloading blob " + blobId, e);
```
lightningrob: downloadErrorCount here?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
ankagrawal: Is there any reason why we have blobs uploaded to Azure with a different creation time than the creation time reported from the originating blobstore partition?
Also is there a reason we don't have "operationTime" in CloudBlobMetadata? If we have an additional field called "operationTime" in the CloudBlobMetadata (similar to one we have in the index metadata on stores), then we can use the following rules to populate operationTime on CloudBlobMetadata:

1. If the blob is replicated or backed-up to a replica (irrespective of the destination being cloud or not), the operation time should just be copied from what is present in the original data.
2. Once we are serving both get and put from azure, we can then use the actual time of operation as the "operationTime". This when replicated to other Azure colos or to linkedin replicas, will follow the rule 1 above.

Unless ofcourse I am missing something here.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudDestination.java
ankagrawal: Sure.. I will keep that in mind. Also I will try to see if we can revisit this logic for the recovery too in future iteration.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
ankagrawal: fixed.

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +617,121 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+    blobIds.add(blobIdList.get(0));
+    blobIds.add(blobIdList.get(1));
+    storeInfo = cloudBlobStore.get(blobIds, getOptions);
+    assertTrue(storeInfo.getMessageReadSetInfo().get(0).getStoreKey().equals(blobIdList.get(0)));
+    assertTrue(storeInfo.getMessageReadSetInfo().get(1).getStoreKey().equals(blobIdList.get(1)));
+
+    //try get for a blob that doesnt exist
+    blobIds = new ArrayList<>(1);
+    id = blobIdList.get(2);
+    blobIds.add(id);
```
lightningrob: blobIds = Collections.singletonList(blobIdList.get(2));

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +617,121 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+    try {
+      storeInfo = cloudBlobStore.get(blobIds, getOptions);
+      fail("A get for non existent blob should have thrown an exception");
+    } catch (StoreException ex) {
```
lightningrob: assert error code is ID_Not_Found

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +617,121 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+    }
+
+    //try get for a list of blob such that some of them dont exist
+    blobIds = new ArrayList<>(2);
+    blobIds.add(blobIdList.get(0));
+    blobIds.add(blobIdList.get(2));
+    try {
+      storeInfo = cloudBlobStore.get(blobIds, getOptions);
+      fail("A get for non existent blob should have thrown an exception");
+    } catch (StoreException ex) {
```
lightningrob: same assertion here

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +617,121 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+      if (replica.getDataNodeId() == remoteHost.dataNodeId) {
+        RemoteReplicaInfo remoteReplicaInfo =
+            new RemoteReplicaInfo(replica, cloudReplica, cloudBlobStore, new MockFindToken(0, 0), Long.MAX_VALUE,
+                SystemTime.getInstance(), new Port(remoteHost.dataNodeId.getPort(), PortType.PLAINTEXT));
+        replicaThread.addRemoteReplicaInfo(remoteReplicaInfo);
+        break;
+      }
+    }
+
+    EnumSet<StoreGetOptions> getOptions = EnumSet.noneOf(StoreGetOptions.class);
```
lightningrob: Please also test the other options and verify you get the correct error code.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
lightningrob: Including constructors.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -114,7 +114,86 @@ public void start() throws StoreException {
...
+    }
+  }
+
+  /**
+   * Gets the operation time for a blob from blob metadata based on the blob's current state and timestamp recorded for that state.
+   * @param metadata blob metadata from which to derive operation time.
+   * @return operation time.
+   */
+  private long getOperationTime(CloudBlobMetadata metadata) {
+    long currentTimeStamp = System.currentTimeMillis();
```
lightningrob: We have the same issue in this method with System.currentTimeMillis() being called in a loop.  Please call it once at the beginning of the get method and pass to all utility methods that need it.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -252,6 +252,24 @@ private boolean uploadIfNotExists(BlobId blobId, long inputLength, CloudBlobMeta
     }
   }
 
+  @Override
+  public void downloadBlob(BlobId blobId, OutputStream outputStream) throws CloudStorageException {
```
lightningrob: Please add test cases in both AzureCloudDestinationTest and AzureIntegrationTest.  The latter in particular can serve as a real-world correctness test for the complete upload/download cycle.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
ankagrawal: sure.

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
ankagrawal: fixed.

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
```diff
@@ -330,6 +349,19 @@ public void testTokens() throws Exception {
         azureDest.retrieveTokens("unknown-path", tokenFileName, outputStream));
   }
 
+  /**
+   * test download matches the blob previously uploaded.
+   * @param blobId id of the blob to download
+   * @param uploadedData data uploaded to the blob
+   * @throws CloudStorageException
+   */
+  private void testDownload(BlobId blobId, byte[] uploadedData) throws CloudStorageException, IOException {
```
lightningrob: Minor: rename this eg. verifyDownloadMatches

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +621,196 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+        "Number of records returned by get should be " + blobIds.size() + " found " + storeInfo.getMessageReadSetInfo()
+            .size(), storeInfo.getMessageReadSetInfo().size() == blobIds.size());
+    for (int i = 0; i < storeInfo.getMessageReadSetInfo().size(); i++) {
+      if (blobIdToUploadedDataMap.containsKey(storeInfo.getMessageReadSetInfo().get(i).getStoreKey())) {
+        ByteBuffer uploadedData = blobIdToUploadedDataMap.get(storeInfo.getMessageReadSetInfo().get(i).getStoreKey());
+        ByteBuffer downloadedData = ByteBuffer.allocate((int) storeInfo.getMessageReadSetInfo().get(i).getSize());
+        WritableByteChannel writableByteChannel = Channels.newChannel(new ByteBufferOutputStream(downloadedData));
+        storeInfo.getMessageReadSet().writeTo(i, writableByteChannel, -1, -1);
+        downloadedData.flip();
+        assertTrue(uploadedData.equals(downloadedData));
```
lightningrob: All the places where you call assertTrue(expected.equals(actual)), better to use assertEquals("error message", expected, actual)

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +621,196 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+  }
+
+  /**
+   * Test cloud store get with a list of blobs such that atleast one of the blobs is not present in the store
+   * @param blobIds list of blobids to get
+   */
+  private void testGetWithAtleastOneNonExistentBlob(List<BlobId> blobIds) {
+    try {
+      store.get(blobIds, EnumSet.noneOf(StoreGetOptions.class));
+      assertTrue("get with any non existent id should fail", false);
```
lightningrob: change to: fail(message)

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +621,196 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+  }
+
+  /**
+   * Test cloud store get with a list of blobs such that atleast one of them is deleted, but {@code StoreGetOptions.Store_Include_Deleted} is not set in getoptions
+   * @param blobIds list of blobids to get
+   */
+  private void testGetForDeletedBlobWithoutIncludeDeleteOption(List<BlobId> blobIds) {
+    try {
+      store.get(blobIds, EnumSet.noneOf(StoreGetOptions.class));
+      assertTrue(
```
lightningrob: fail

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +621,196 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+          "get should fail for a deleted blob, if StoreGetOptions.Store_Include_Deleted is not set in get options",
+          false);
+    } catch (StoreException e) {
+      assertTrue(
```
lightningrob: assertEquals

ambry-cloud/src/test/java/com.github.ambry.cloud/CloudBlobStoreTest.java
```diff
@@ -614,4 +621,196 @@ private BlobId getUniqueId(short accountId, short containerId, boolean encrypted
...
+   * @param blobIds list of blobids to get
+   */
+  private void testGetForDeletedBlobWithoutIncludeExpiredOption(List<BlobId> blobIds) {
+    try {
+      store.get(blobIds, EnumSet.noneOf(StoreGetOptions.class));
+      assertTrue(
+          "get should fail for a expired blob, if StoreGetOptions.Store_Include_Expired is not set in get options",
+          false);
+    } catch (StoreException e) {
+      assertTrue(
```
lightningrob: assertEquals and remove the expected value from message string

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -114,7 +114,86 @@ public void start() throws StoreException {
...
+   * @throws StoreException if the {@code CloudBlobMetadata} isnt valid
+   */
+  private void validateCloudMetadata(Map<String, CloudBlobMetadata> cloudBlobMetadataListMap,
+      EnumSet<StoreGetOptions> storeGetOptions) throws StoreException {
+    for (String key : cloudBlobMetadataListMap.keySet()) {
+      if (isBlobDeleted(cloudBlobMetadataListMap.get(key)) && !storeGetOptions.contains(
+          StoreGetOptions.Store_Include_Deleted)) {
+        throw new StoreException("Id " + key + " has been deleted on the cloud", StoreErrorCodes.ID_Deleted);
+      }
+      long currentTimestamp = System.currentTimeMillis();
```
lightningrob: This is still inside the for loop.  Please move it out.

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
```diff
@@ -119,9 +126,17 @@ public void testNormalFlow() throws Exception {
...
+
     azureDest.purgeBlob(metadata);
     assertTrue("Expected empty set after purge",
         azureDest.getBlobMetadata(Collections.singletonList(blobId)).isEmpty());
+
+    // Get blob should fail after purge
+    try {
+      verifyDownloadMatches(blobId, uploadData);
+    } catch (CloudStorageException csex) {
+      assertTrue(csex.getMessage().contains("Error downloading blob"));
```
lightningrob: Minor: not good practice for tests to check for exact error text, makes tests brittle.

PR #1226, opened 2019-07-29
Support dynamically adding store to storage manager

jsjtzyy:
1.allow server to dynamically add store via StorageManager
2.code changes made in StorageManager, DiskManager, CompactionManager
3.support adding store onto a new disk, which is empty initially.

This PR is dependent on #1210. Once this PR is merged, we can close #994 .

PR #1232, opened 2019-08-08
Support dynamically removing store from storage manager

jsjtzyy:
1. Given partition id, remove the corresponding store from storage
manager, disk manager and compaction manager.
2. Support deleting allocated files of store and returning swap
segments to reserver pool if necessary.

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
@@ -22,8 +22,11 @@ import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.io.File;
 import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -682,6 +685,36 @@ class BlobStore implements Store {
         LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats);
   }
 
+  /**
+   * Delete files of this store.
+   * This is the last step to remove store from this node. Return swap segments (if any) to reserve pool and delete all
+   * files/dirs associated with this store. This method is invoked by transition in AmbryStateModel (OFFLINE -> DROPPED)
+   */
+  public void deleteStoreFiles() throws StoreException, IOException {
+    // step0: ensure the store has been shut down
+    if (started) {
+      throw new IllegalStateException("Store is still started. Deleting store files is not allowed.");
+    }
+    // step1: return occupied swap segments (if any) to reserve pool
+    String[] swapSegmentsInUse = compactor.getSwapSegmentsInUse();
+    if (swapSegmentsInUse.length > 0) {
+      for (String fileName : swapSegmentsInUse) {
+        logger.trace("Returning swap segment {} to reserve pool", fileName);
+        File swapSegmentTempFile = new File(dataDir, fileName);
+        diskSpaceAllocator.free(swapSegmentTempFile, config.storeSegmentSizeInBytes, storeId, true);
+      }
+    }
+    // step2: delete all files
+    logger.info("Deleting store {} directory", storeId);
+    File storeDir = new File(dataDir);
+    try {
+      Files.walk(storeDir.toPath()).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
+    } catch (Exception e) {
+      throw new IOException("Couldn't delete store directory " + dataDir);
+    }
+    logger.info("All files of store {} deleted", storeId);
+  }
+
   @Override
   public void shutdown() throws StoreException {
     shutdown(false);
@@ -751,9 +784,10 @@ class BlobStore implements Store {
    */
   DiskSpaceRequirements getDiskSpaceRequirements() throws StoreException {
     checkStarted();
-    DiskSpaceRequirements requirements = log.isLogSegmented() ? new DiskSpaceRequirements(
-        replicaId.getPartitionId().toPathString(), log.getSegmentCapacity(),
-        log.getRemainingUnallocatedSegments(), compactor.getSwapSegmentsInUse()) : null;
+    DiskSpaceRequirements requirements =
+        log.isLogSegmented() ? new DiskSpaceRequirements(replicaId.getPartitionId().toPathString(),
+            log.getSegmentCapacity(), log.getRemainingUnallocatedSegments(), compactor.getSwapSegmentsInUse().length)
+            : null;
     logger.info("Store {} has disk space requirements: {}", storeId, requirements);
     return requirements;
   }
--- a/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
@@ -705,7 +705,7 @@ class BlobStore implements Store {
     try {
       Utils.deleteFileOrDirectory(storeDir);
     } catch (Exception e) {
-      throw new IOException("Couldn't delete store directory " + dataDir);
+      throw new IOException("Couldn't delete store directory " + dataDir, e);
     }
     logger.info("All files of store {} deleted", storeId);
   }
--- a/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
@@ -22,11 +22,8 @@ import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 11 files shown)


ambry-store/src/main/java/com.github.ambry.store/CompactionManager.java
```diff
@@ -153,6 +153,30 @@ boolean controlCompactionForBlobStore(BlobStore store, boolean enable) {
...
 
+  /**
+   * Remove store from compaction manager.
+   * @param store the {@link BlobStore} to remove
+   * @return {@code true} if store is removed successfully. {@code false} if not.
+   */
+  boolean removeBlobStore(BlobStore store) {
+    boolean result;
+    if (compactionExecutor == null) {
+      stores.remove(store);
```
jsjtzyy: Once #1226 is merged, `stores` will become concurrent set which should be ok here.

ambry-store/src/test/java/com.github.ambry.store/BlobStoreTest.java
```diff
@@ -1180,6 +1181,66 @@ public void storeIoErrorCountTest() throws StoreException, IOException {
     reloadStore();
   }
 
+  /**
+   * Test both success and failure cases when deleting store files.
+   * @throws Exception
+   */
+  @Test
+  public void deleteStoreFilesTest() throws Exception {
+    assumeTrue(isLogSegmented);
```
cgtz: Could this code be tested with non segmented logs? Perhaps just put the test of swap segments within `if (isLogSegmented)`

ambry-store/src/test/java/com.github.ambry.store/BlobStoreTest.java
```diff
@@ -1180,6 +1181,66 @@ public void storeIoErrorCountTest() throws StoreException, IOException {
...
+    store.shutdown();
+    // create test store directory
+    File storeDir = StoreTestUtils.createTempDirectory("store-" + storeId);
+    File reserveDir = StoreTestUtils.createTempDirectory("reserve-pool");
+    DiskSpaceAllocator diskAllocator =
+        new DiskSpaceAllocator(true, reserveDir, 0, new StorageManagerMetrics(new MetricRegistry()));
+    StoreConfig config = new StoreConfig(new VerifiableProperties(properties));
+    MetricRegistry registry = new MetricRegistry();
+    StoreMetrics metrics = new StoreMetrics(registry);
+    //createBlobStore(getMockReplicaId(storeDir.getAbsolutePath()));
```
cgtz: remove commented code

ambry-store/src/main/java/com.github.ambry.store/StoreMetrics.java
```diff
@@ -219,6 +219,8 @@ private void deregisterIndexGauges(String storeId) {
     registry.remove(MetricRegistry.name(Log.class, prefix + "CurrentCapacityUsed"));
     registry.remove(MetricRegistry.name(Log.class, prefix + "PercentageUsedCapacity"));
     registry.remove(MetricRegistry.name(Log.class, prefix + "CurrentSegmentCount"));
+    registry.remove(MetricRegistry.name(Log.class, "ByteBufferForAppendTotalCount"));
+    registry.remove(MetricRegistry.name(Log.class, "UnderCompaction" + SEPERATOR + "ByteBufferForAppendTotalCount"));
```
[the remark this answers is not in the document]
jsjtzyy: fixed

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -682,6 +685,36 @@ CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy) throws
...
+        logger.trace("Returning swap segment {} to reserve pool", fileName);
+        File swapSegmentTempFile = new File(dataDir, fileName);
+        diskSpaceAllocator.free(swapSegmentTempFile, config.storeSegmentSizeInBytes, storeId, true);
+      }
+    }
+    // step2: delete all files
+    logger.info("Deleting store {} directory", storeId);
+    File storeDir = new File(dataDir);
+    try {
+      Files.walk(storeDir.toPath()).sorted(Comparator.reverseOrder()).map(Path::toFile).forEach(File::delete);
```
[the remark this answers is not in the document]
jsjtzyy: ah, I didn't notice that. Will change this piece of code.

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
```diff
@@ -290,6 +289,26 @@ public boolean shutdownBlobStore(PartitionId id) {
     return diskManager != null && diskManager.shutdownBlobStore(id);
   }
 
+  /**
+   * Remove store from storage manager.
+   * @param id the {@link PartitionId} associated with store
+   * @return {@code true} if removal succeeds. {@code false} otherwise.
+   */
+  public boolean removeBlobStore(PartitionId id) {
```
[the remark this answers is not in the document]
jsjtzyy: I am slightly leaning towards current way to make it sort of consistent with `scheduleNextForCompaction`, `controlCompactionForBlobStore` etc. Returning boolean can be explicitly verified in unit tests.

ambry-store/src/test/java/com.github.ambry.store/BlobStoreTest.java
jsjtzyy: Good point. I added some codes to ensure both segmented and non-segmented cases are tests.

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -682,6 +682,34 @@ CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy) throws
...
+   * Delete files of this store.
+   * This is the last step to remove store from this node. Return swap segments (if any) to reserve pool and delete all
+   * files/dirs associated with this store. This method is invoked by transition in AmbryStateModel (OFFLINE -> DROPPED)
+   */
+  public void deleteStoreFiles() throws StoreException, IOException {
+    // step0: ensure the store has been shut down
+    if (started) {
+      throw new IllegalStateException("Store is still started. Deleting store files is not allowed.");
+    }
+    // step1: return occupied swap segments (if any) to reserve pool
```
lightningrob: Nitpick: say "Step 1", "Step 2", ...

ambry-store/src/main/java/com.github.ambry.store/CompactionManager.java
```diff
@@ -154,6 +154,28 @@ void controlCompactionForBlobStore(BlobStore store, boolean enable) {
     }
   }
 
+  /**
+   * Remove store from compaction manager.
+   * @param store the {@link BlobStore} to remove
+   * @return {@code true} if store is removed successfully. {@code false} if not.
+   */
+  boolean removeBlobStore(BlobStore store) {
+    boolean result;
```
lightningrob: It's not necessary to keep the result in a local variable unless it is intermediate state.  In all cases here, you can simply return the result.

jsjtzyy:
Addressed @cgtz 's comments. @lightningrob gentle reminder to review.

ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
```diff
@@ -359,6 +359,32 @@ boolean shutdownBlobStore(PartitionId id) {
...
+   * @param id the {@link PartitionId} of the {@link BlobStore} which should be removed.
+   * @return {@code true} if store removal was successful. {@code false} if not.
+   */
+  boolean removeBlobStore(PartitionId id) {
+    BlobStore store = stores.get(id);
+    if (store == null) {
+      logger.info("Store {} is not found in disk manager", id);
+      return true;
+    }
+    if (!running || store.isStarted()) {
```
lightningrob: Why is (running == false) an error condition here?

ambry-store/src/main/java/com.github.ambry.store/CompactionManager.java
```diff
@@ -154,6 +154,28 @@ void controlCompactionForBlobStore(BlobStore store, boolean enable) {
...
+    if (compactionExecutor == null) {
+      stores.remove(store);
+      result = true;
+    } else if (!compactionExecutor.getStoresDisabledCompaction().contains(store)) {
+      logger.error("Fail to remove store ({}) from compaction manager because compaction of it is still enabled",
```
lightningrob: This seems like it should be an IllegalStateException.  Is there a valid scenario where compaction is still enabled when removeStore is called?  If we return false and compaction is turned off later, will the store ever be removed?

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
jsjtzyy: Fixed

ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
jsjtzyy: `running` here indicates if the `DiskManager` is running. If `DiskManager` is closed, we are not supposed to remove store because current in-memory status in `DiskManager` is no longer up-to-date.

ambry-store/src/main/java/com.github.ambry.store/CompactionManager.java
jsjtzyy: Fixed

ambry-store/src/main/java/com.github.ambry.store/CompactionManager.java
jsjtzyy: I agree with your point, however, I am trying to use same way in `scheduleNextForCompaction` method (line 360). Returning false may happen when disabling compaction on the store hasn't completed or hasn't been executed. The Helix state model will retry entire workflow (disable compaction, remove store) to guarantee the store is correctly removed.

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -682,6 +682,34 @@ CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy) throws
...
+    for (String fileName : swapSegmentsInUse) {
+      logger.info("Returning swap segment {} to reserve pool", fileName);
+      File swapSegmentTempFile = new File(dataDir, fileName);
+      diskSpaceAllocator.free(swapSegmentTempFile, config.storeSegmentSizeInBytes, storeId, true);
+    }
+    // Step 2: delete all files
+    logger.info("Deleting store {} directory", storeId);
+    File storeDir = new File(dataDir);
+    try {
+      Utils.deleteFileOrDirectory(storeDir);
```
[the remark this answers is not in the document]
jsjtzyy: Thanks for reminding me of this. I prefer to keep this catch in case there is any non-IOException. Also, I would take your advice to set the exception `e`

PR #1234, opened 2019-08-12
Implement helixAccountService with version history and blobs

justinlin-linkedin:
This is the first step of implementation for this design https://docs.google.com/document/d/1Lr30Uph1KL0AMHznTGEdC_JG3W6_d_NKNTa9FbBTTjI/edit

In this implementation, HelixAccountService has two separated implementation of AccountMetadataStore, which fetches account metadata and update it.

Old approach is still the default implementation and new approach would be enabled after.

The next step is to 
1) implement a function to copy the account metadata from the old path to ambry server and create that blob id list.
2) limit the size of list by removing the blobs when the size of list exceeds the count in the config.

Notice that this implementation might end up with some blob leak.


ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -55,10 +69,20 @@
...
+ *   new way, it internally stores the list of blob ids that point to different versions of {@link Account} metadata.
+ *   In both way, the latest full {@link Account} metadata will be cached locally, so serving an {@link Account} query
+ *   will not incur any calls to remote {@code ZooKeeper} or {@code AmbryServer}.
+ * </p>
+ * <p>
+ *   After transition, this implementation would not only save the latest set of {@link Account} metadata, but also save
+ *   several previous versions. It keeps a list of blob ids in {@link HelixPropertyStore} and remove the earliest one
+ *   when it reaches the limit of the version numbers. There are several benefits from this approach.
+ *   <ul>
+ *     <li>Revolting changes would be much easier.</li>
```
lightningrob: revoking?  reverting?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -423,119 +435,253 @@ private void checkOpen() {
...
+   */
+  Collection<Account> getAccounts() {
+    return Collections.unmodifiableCollection(idToAccountMap.values());
+  }
+}
+
+/**
+ * An AccountMetadataStore is a storage to keep and retrieve {@link Account} metadata.
+ */
+interface AccountMetadataStore {
```
lightningrob: Please refactor this class as it has gotten way too complex.  You have an interface and multiple implementations of it all in this class.

ambry-account/src/test/java/com/github/ambry/account/HelixAccountServiceTest.java
```diff
@@ -115,11 +128,13 @@ public HelixAccountServiceTest() throws Exception {
     helixConfigProps.setProperty(HelixPropertyStoreConfig.HELIX_PROPERTY_STORE_PREFIX + "root.path", STORE_ROOT_PATH);
     accountBackupDir = Paths.get(TestUtils.getTempDir("account-backup")).toAbsolutePath();
     helixConfigProps.setProperty(HelixAccountServiceConfig.BACKUP_DIRECTORY_KEY, accountBackupDir.toString());
+    helixConfigProps.setProperty(HelixAccountServiceConfig.USE_NEW_ZNODE_PATH, String.valueOf(useNewZNodePath));
     vHelixConfigProps = new VerifiableProperties(helixConfigProps);
     storeConfig = new HelixPropertyStoreConfig(vHelixConfigProps);
     notifier = new MockNotifier<>();
     mockHelixAccountServiceFactory =
         new MockHelixAccountServiceFactory(vHelixConfigProps, new MetricRegistry(), notifier, null);
```
lightningrob: You can pass useNewZNodePath and mockRouter to the factory constructor.  Then all the tests won't need the if/setupRouter code, the factory can do that.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
justinlin-linkedin: updated.

ambry-account/src/test/java/com/github/ambry/account/HelixAccountServiceTest.java
justinlin-linkedin: good idea, updated.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
justinlin-linkedin: It's actually not in the class, but in the same file. I separated the interface and two implementations out of the class but didn't create a new file for each of them, thinking that the legacy one would be removed later any way.
Should I just create separated files?

cgtz: Yeah, probably better to just create separate files in the same package for each.

justinlin-linkedin: updated

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -0,0 +1,331 @@
...
+      object.put(entry.getKey(), entry.getValue());
+    }
+
+    ByteBufferReadableStreamChannel channel =
+        new ByteBufferReadableStreamChannel(ByteBuffer.wrap(object.toString().getBytes(Charsets.UTF_8)));
+    BlobProperties properties = new BlobProperties(channel.getSize(), SERVICE_ID, ACCOUNT_ID, CONTAINER_ID, false);
+    return router.putBlob(properties, null, channel, PutBlobOptions.DEFAULT).get();
+  }
+
+  private class ZKUpdater implements DataUpdater<ZNRecord> {
```
lightningrob: javadoc

ambry-account/src/main/java/com/github/ambry/account/LegacyMetadataStore.java
```diff
@@ -0,0 +1,171 @@
...
+  LegacyMetadataStore(HelixAccountService accountService, HelixPropertyStore<ZNRecord> helixStore,
+      HelixAccountServiceConfig config) throws IOException {
+    this.helixStore = helixStore;
+    this.accountService = accountService;
+    this.config = config;
+  }
+
+  @Override
+  public void fetchAndUpdateCache(boolean isCalledFromListener) {
+    lock.lock();
```
lightningrob: Since the lock is basically a mutex, could just synchronize this method.

ambry-account/src/main/java/com/github/ambry/account/AccountMetadataStore.java
```diff
@@ -0,0 +1,38 @@
...
+
+
+/**
+ * An AccountMetadataStore is a storage to keep and retrieve {@link Account} metadata.
+ */
+interface AccountMetadataStore {
+
+  /**
+   * fetchAndUpdateCache would fetch the latest full set of {@link Account} metadata and update the cache
+   * in the {@link HelixAccountService}.
```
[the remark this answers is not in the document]
lightningrob: I recommend changing this from interface to abstract class with common constructor.

ambry-account/src/main/java/com/github/ambry/account/LegacyMetadataStore.java
```diff
@@ -0,0 +1,171 @@
...
+import org.I0Itec.zkclient.DataUpdater;
+import org.apache.helix.AccessOption;
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.HelixPropertyStore;
+import org.json.JSONException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+class LegacyMetadataStore implements AccountMetadataStore {
```
lightningrob: javadoc

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -147,9 +148,32 @@
...
+      initialFetchAndSchedule();
+    }
+  }
+
+  /**
+   * set the router to the given one.
+   * @param router The router to set.
+   * @throws IllegalStateException when the router already set up.
+   */
+  public void setupRouter(final Router router) throws IllegalStateException {
```
lightningrob: final not necessary

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -423,267 +419,176 @@ private void checkOpen() {
...
+  }
 
-    /**
-     * Gets all the {@link Account}s in this {@code AccountInfoMap} in a {@link Collection}.
-     * @return A {@link Collection} of all the {@link Account}s in this map.
-     */
-    private Collection<Account> getAccounts() {
-      return Collections.unmodifiableCollection(idToAccountMap.values());
+  /**
+   * If a backup file has not yet been created, reserve a new one with the following file name format:
```
lightningrob: Minor: not quite accurate since you are reserving a new backup file regardless of whether one already exists.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -0,0 +1,331 @@
...
+          newVersion = blobIDAndVersion.version + 1;
+
+          // Start Step 2:
+          accountMap = readAccountMetadataFromBlobID(blobIDAndVersion.blobID);
+          // make this list mutable
+          accountBlobIDs = new ArrayList<>(accountBlobIDs);
+        } catch (JSONException e) {
+          logger.error("Exception occurred when parsing the blob id list from {}", accountBlobIDs);
+          accountService.accountServiceMetrics.remoteDataCorruptionErrorCount.inc();
+          throw new IllegalStateException("Exception occurred when parsing blob id list", e);
```
lightningrob: why not use the same error message that you logged, and skip the logging?  If the caller will eventually log the exception, we don't want to log it multiple times on stack.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -0,0 +1,331 @@
...
+      lock.unlock();
+    }
+  }
+
+  private Map<String, String> readAccountMetadataFromBlobID(String blobID) {
+    long startTimeMs = System.currentTimeMillis();
+    Future<GetBlobResult> resultF = accountService.router.get().getBlob(blobID, new GetBlobOptionsBuilder().build());
+    try {
+      GetBlobResult result = resultF.get();
+      accountService.accountServiceMetrics.accountFetchFromAmbryTimeInMs.update(
```
lightningrob: All the places where you update metrics, please use a saved instance of accountServiceMetrics and don't get it from accountService.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -0,0 +1,331 @@
...
+      int blobSize = (int) result.getBlobInfo().getBlobProperties().getBlobSize();
+      InputStream input = new ReadableStreamChannelInputStream(result.getBlobDataChannel());
+      byte[] bytes = Utils.readBytesFromStream(input, blobSize);
+
+      JSONObject object = new JSONObject(new String(bytes, Charsets.UTF_8));
+      Map<String, String> map = new HashMap<>();
+      object.keySet().stream().forEach(key -> map.put(key, object.getString(key)));
+      return map;
+    } catch (Exception e) {
+      logger.debug("Failed to read account metadata from blob id={}", blobID, e);
```
lightningrob: Don't ever use debug to log an error.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -0,0 +1,331 @@
...
+    private String newBlobID = null;
+
+    ZKUpdater(Collection<Account> accounts) {
+      this.accounts = accounts;
+      Pair<String, Path> backupPrefixAndPath = null;
+      if (accountService.backupDirPath != null) {
+        try {
+          backupPrefixAndPath = accountService.reserveBackupFile();
+        } catch (IOException e) {
+          logger.error("Error reserving backup file", e);
```
lightningrob: What happens later if the backup file wasn't created?  We skip the backups?

ambry-account/src/main/java/com/github/ambry/account/AccountMetadataStore.java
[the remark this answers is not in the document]
justinlin-linkedin: good idea. Updated.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -0,0 +1,331 @@
...
+      // be caught by Helix and helixStore#update will return false.
+      if (accountService.hasConflictingAccount(this.accounts, localAccountInfoMap)) {
+        // Throw exception, so that helixStore can capture and terminate the update operation
+        throw new IllegalArgumentException(
+            "Updating accounts failed because one account to update conflicts with existing accounts");
+      }
+      for (Account account : this.accounts) {
+        try {
+          accountMap.put(String.valueOf(account.getId()), account.toJson(true).toString());
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
justinlin-linkedin: toJson might throw JSONException.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
justinlin-linkedin: updated

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
justinlin-linkedin: Helix would actually catch the exception without logging it out. The function is basically a callback function and throwing any Exception would terminate the execution of HelixPropertyStore.update function and force it to return false.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
justinlin-linkedin: With the current logic, yes.
The current backup file logic is subpar at best. It doesn't guarantee the atomicity of writes and doesn't clean up old files. But I am using the same set of tests for the new implementation so I don't want to change the backup logic. Will make it better at later PR.

ambry-account/src/main/java/com/github/ambry/account/AccountMetadataStore.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.HelixPropertyStore;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * An AccountMetadataStore is a storage to keep and retrieve {@link Account} metadata. This is the base
+ */
+abstract class AccountMetadataStore {
+  private static final Logger logger = LoggerFactory.getLogger(LegacyMetadataStore.class);
```
lightningrob: Wrong class.

ambry-account/src/main/java/com/github/ambry/account/AccountMetadataStore.java
```diff
@@ -0,0 +1,103 @@
...
+
+  protected final AccountServiceMetrics accountServiceMetrics;
+  protected final LocalBackup backup;
+  protected final String znRecordPath;
+  private final HelixPropertyStore<ZNRecord> helixStore;
+  private final ReentrantLock lock = new ReentrantLock();
+
+  AccountMetadataStore(AccountServiceMetrics accountServiceMetrics, LocalBackup backup,
```
lightningrob: javadoc

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -415,247 +512,42 @@ private boolean hasConflictingAccount(Collection<Account> accountsToSet, Account
...
-    }
+  public LocalBackup(AccountServiceMetrics accountServiceMetrics, HelixAccountServiceConfig config) throws IOException {
+    this.accountServiceMetrics = accountServiceMetrics;
+    backupDirPath = config.backupDir.isEmpty() ? null : Files.createDirectories(Paths.get(config.backupDir));
   }
 
   /**
-   * A {@link DataUpdater} to be used for updating {@link #FULL_ACCOUNT_METADATA_PATH} inside of
-   * {@link #updateAccounts(Collection)}
+   * Reserve a new one with the following file name format: {@code {yyyyMMdd}T{HHmmss}.{unique long}.old}.
```
lightningrob: "a new backup file"

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
lightningrob: Fine.  Then create the string once with String.format("... %s") and use in both places.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
lightningrob: Okay.  I don't like the code but I can see you didn't change it only moved it around.

ambry-api/src/main/java/com.github.ambry/config/HelixAccountServiceConfig.java
```diff
@@ -54,12 +57,28 @@
...
+  @Config(USE_NEW_ZNODE_PATH)
+  @Default("false")
+  public final boolean useNewZNodePath;
+
+  /**
+   * The number of latest versions of account metadata to save in ambry.
+   */
+  @Config(VERSION_COUNT)
+  @Default("10")
+  public final int versionCount;
```
jsjtzyy: better to rename this, current name doesn't quite match the comment.  How about `maxNumOfVersionsInMem` or `maxNumOfCachedVersions`?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -138,18 +131,63 @@
...
-    backupDirPath = config.backupDir.isEmpty() ? null : Files.createDirectories(Paths.get(config.backupDir));
     if (notifier != null) {
       notifier.subscribe(ACCOUNT_METADATA_CHANGE_TOPIC, changeTopicListener);
     } else {
       logger.warn("Notifier is null. Account updates cannot be notified to other entities. Local account cache may not "
           + "be in sync with remote account data.");
       accountServiceMetrics.nullNotifierCount.inc();
     }
+    if (config.useNewZNodePath) {
+      accountMetadataStore = new RouterStore(this.accountServiceMetrics, backup, helixStore, router);
```
jsjtzyy: I understand this is first PR to improve account metadata and I believe `setupRouter` will be invoked in future PR, however, here `router` might not be initialized, can we do some proactive check in `RouterStore`?
Eventually, after transition is done, we should explicitly call `setupRouter` in `RestServer.java`

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -138,18 +131,63 @@
...
+      initialFetchAndSchedule();
+    }
+  }
+
+  /**
+   * A synchronized function to fetch account metadata from {@link AccountMetadataStore} and update the in memory cache.
+   * @param isCalledFromListener True is this function is invoked in the {@link TopicListener}.
+   */
+  private synchronized void fetchAndUpdateCache(boolean isCalledFromListener) {
+    Map<String, String> accountMap = accountMetadataStore.fetchAccountMetadata();
```
jsjtzyy: Although this is `synchronized` method, there might be race condition when `initialFetchAndSchedule` is called before `accountMetadataStore` is instantiated. My suggestion is, if possible, move `notifier.subscribe()` to the end of constructor.

ambry-api/src/main/java/com.github.ambry/config/HelixAccountServiceConfig.java
justinlin-linkedin: I will remove this option for this PR since it's logic is not yet implemented. Will revisit it at next PR.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
justinlin-linkedin: good idea, updated

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
justinlin-linkedin: Good Idea, I move the notifier.subscribe into the initialFetchAndSchedule function so there will be no race condition here.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -0,0 +1,345 @@
...
+          // make this list mutable
+          accountBlobIDs = new ArrayList<>(accountBlobIDs);
+        } catch (JSONException e) {
+          accountServiceMetrics.remoteDataCorruptionErrorCount.inc();
+          errorMessage = String.format("Exception occurred when parsing the blob id list from {}", accountBlobIDs);
+          logger.error(errorMessage);
+          throw new IllegalStateException(errorMessage, e);
+        } catch (Exception e) {
+          errorMessage =
+              String.format("Unexpected exception occurred when parsing the blob id list from {}", accountBlobIDs);
```
jsjtzyy: if you choose to String.format, can you change it to following style (basically replace `{}` with `%s`):
```
String.format("Unexpected exception occurred when parsing the blob id list from %s", accountBlobIDs);
```
Make change to all occurrences. This can pass intellij inspection.

justinlin-linkedin: figured that string concatenation would just do the work.

ambry-account/src/main/java/com/github/ambry/account/AccountMetadataStore.java
```diff
@@ -0,0 +1,107 @@
...
+   * when there is no {@link Account} created.
+   * @return {@link Account} metadata in a map.
+   */
+  Map<String, String> fetchAccountMetadata() {
+    long startTimeMs = System.currentTimeMillis();
+    logger.trace("Start reading ZNRecord from path={}", znRecordPath);
+    ZNRecord znRecord = helixStore.get(znRecordPath, null, AccessOption.PERSISTENT);
+    logger.trace("Fetched ZNRecord from path={}, took time={} ms", znRecordPath, startTimeMs);
+    if (znRecord == null) {
+      logger.debug("The ZNRecord to read does not exist on path={}", znRecordPath);
```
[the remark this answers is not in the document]
justinlin-linkedin: but it's not a warning message. ZNRecord can be null here. It's one of the legit state.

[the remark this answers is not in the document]
jsjtzyy: If not `warn`, I would suggest `info` because such info is supposed to be printed out (`debug` level is turned off by default).  I feel like ZNRecord should exist in almost all cases, if not, it's actually a warning that either the ZNode is deleted or znRecordPath is incorrect.

PR #1235, opened 2019-08-13
Upgrade gradle, fix test logging and int test CPU usage

cgtz:
- Upgrade to gradle 5.2.1. As part of this, I have separated the
  integration tests into their own task so that they can be run
  independently. The "allTest" target will do what the "test" target
  used to do.
- Suppress some especially noisy logs in integration tests. These log
  messages were firing continuously when the node-down scenarios were
  tested.
- Configure MockCluster replication threads to do a 100 ms sleep per
  iteration. This greatly reduces the CPU requirements for running the
  integration tests and allows us to run them on travis-ci.
- Set up travis to run integration tests too.


jsjtzyy:
We can close #1031 once current PR is merged.

ambry-store/src/test/java/com.github.ambry.store/IndexSegmentTest.java
```diff
@@ -793,9 +796,6 @@ private void getEntriesSinceTest(NavigableMap<MockId, NavigableSet<IndexValue>>
     highestIdIncluded = referenceIndex.lastKey().equals(idToCheck) ? null : referenceIndex.lastKey();
     // check the case where maxSize is more than the number of entries in the segment
     doGetEntriesSinceTest(referenceIndex, segment, idToCheck, maxSize + 1, 0, highestIdIncluded);
```
jsjtzyy: One thing we can improve in the future is `doGetEntriesSinceTest` is invoked multiple times and even in the loop. I find that `doGetEntriesSinceTest` test both `getEntriesSince()` and `getIndexEntriesSince()` methods. Note that  `getIndexEntriesSince()` with `oneEntryPerKey = true` is actually tested in `getIndexEntriesSince()`. It seems we don't have to explicitly test `oneEntryPerKey = true` for `getIndexEntriesSince()` again. (My two cents.)

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: java
 dist: trusty
 script:
-  - ./gradlew build && ./gradlew codeCoverageReport
+  - ./gradlew build && ./gradlew intTest && ./gradlew codeCoverageReport
```
lightningrob: Doesn't codeCoverageReport force all tests to be run?  I would think your gradle changes make the intTest here unnecessary.

ambry-store/src/test/java/com.github.ambry.store/IndexSegmentTest.java
cgtz: For now I will leave the test since whether or not it is called is dependent on the implementation of `getEntriesSince` and `getIndexEntriesSince` is still exposed to the package. I'll think about how to further reduce the test length in future PRs.

PR #1237, opened 2019-08-14
Misc fixes in cluster manager, replication and compaction manager

jsjtzyy:
1. Set DEBUG and TRACE log level in Histogram dumper
2. Fixed race condition in compaction manager test
3. Printed out remote dc exception during helix cluster manager
initialization (previously it was swallowed and hard to debug)
4. Moved partition lag update operatio to exchange metadata method in
replication manager
5. Paused replication on store that is not started. Replication can
resume if store is restarted.

### DIFF
--- a/ambry-server/src/test/java/com.github.ambry.server/AmbryRequestsTest.java
+++ b/ambry-server/src/test/java/com.github.ambry.server/AmbryRequestsTest.java
@@ -1245,10 +1245,12 @@ public class AmbryRequestsTest {
      * An empty {@link Store} implementation.
      */
     private Store store = new Store() {
+      boolean started;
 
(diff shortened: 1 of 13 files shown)


ambry-server/src/test/java/com.github.ambry.server/AmbryRequestsTest.java
```diff
@@ -1245,10 +1245,12 @@ public void sendResponse(Send payloadToSend, Request originalRequest, ServerNetw
      * An empty {@link Store} implementation.
      */
     private Store store = new Store() {
+      boolean started;
```
justinlin-linkedin: nit: volatile

PR #1239, opened 2019-08-16
Enable shipkit release publishing

cgtz:
- Set up the shipkit (https://github.com/mockito/shipkit/blob/master/docs/getting-started.md) gradle plugin for publishing releases to bintray.
- Configure travis to automaticly generate github releases and maven artifacts when PRs are merged to the master branch.

Tested locally using `./gradlew testRelease`
I have set up travis-ci environment variables to use my bintray/github account/api keys.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,5 @@
 language: java
 dist: trusty
-script:
-  - ./gradlew build codeCoverageReport
 sudo: required
 jdk:
   - oraclejdk8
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -17,5 +15,15 @@ addons:
...
       - oracle-java8-unlimited-jce-policy
+jobs:
+  include:
+    - stage: test
+      script: "./gradlew -s --scan build codeCoverageReport"
+    - stage: deploy
+      script: "./gradlew -s build && ./gradlew ciPerformRelease"
+stages:
+  - test
+  - name: deploy
```
lightningrob: curious why you have name: before deploy but not test.  Are they equivalent?

build.gradle
```diff
@@ -105,6 +109,12 @@ subprojects {
     artifacts {
         archives testJar
     }
+
+    javadoc {
+        // TODO audit and fix our javadocs so that we don't need this setting
```
lightningrob: Can add comment that this was added because shipkit auto-builds javadoc which we were skipping before.

.travis.yml
cgtz: Yep, they're equivalent, but I added the name: for consistency

mockitoguy:
This looks good. Thank you!

PR #1240, opened 2019-08-16
Implement decryption for cloud blobstore download blob 

ankagrawal:
for cases where the blob was encrypted by cloud blobstore before uploading

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
@@ -46,6 +46,7 @@ public class CloudBlobMetadata {
   private String vcrKmsContext;
   private String cryptoAgentFactory;
   private String cloudBlobName;
+  private long encryptedSize;
 
   /**
    * Possible values of encryption origin for cloud stored blobs.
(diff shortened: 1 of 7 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -146,10 +147,17 @@ public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> stor
     return new StoreInfo(messageReadSet, messageInfos);
   }
 
-  public void downloadBlob(BlobId blobId, OutputStream outputStream) throws StoreException {
+  public void downloadBlob(CloudBlobMetadata cloudBlobMetadata, BlobId blobId, OutputStream outputStream) throws StoreException {
```
lightningrob: Make this package private.  Unless this will be called outside the cloud package, in which case the caller probably shouldn't know about CloudBlobMetadata.
Also javadoc.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -80,6 +81,24 @@ public CloudBlobMetadata() {
    */
   public CloudBlobMetadata(BlobId blobId, long creationTime, long expirationTime, long size,
       EncryptionOrigin encryptionOrigin, String vcrKmsContext, String cryptoAgentFactory) {
+    this(blobId, creationTime, expirationTime, size, encryptionOrigin, vcrKmsContext, cryptoAgentFactory, -1);
+  }
+
+  /**
```
lightningrob: If we're going to have two constructors, let's make the first one be without encryption (defaults/nulls for those params) and the other with all the params.

ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -146,10 +147,17 @@ public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> stor
     return new StoreInfo(messageReadSet, messageInfos);
   }
 
-  public void downloadBlob(BlobId blobId, OutputStream outputStream) throws StoreException {
+  void downloadBlob(CloudBlobMetadata cloudBlobMetadata, BlobId blobId, OutputStream outputStream) throws StoreException {
```
lightningrob: Please add javadoc (I realize it was missing before).

PR #1241, opened 2019-08-17
Undelete base records

dharju:
Added undelete subrecord, new delete subrecord version, and recordVersion to blob properties for support of undelete feature

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
+++ b/ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
@@ -32,6 +32,7 @@ public class BlobProperties {
   private final boolean isEncrypted;
   private long blobSize;
   private long timeToLiveInSeconds;
+  private short recordVersion;
   // Non persistent blob properties.
   private final String externalAssetTag;
 
@@ -44,7 +45,7 @@ public class BlobProperties {
    */
   public BlobProperties(long blobSize, String serviceId, short accountId, short containerId, boolean isEncrypted) {
     this(blobSize, serviceId, null, null, false, Utils.Infinite_Time, SystemTime.getInstance().milliseconds(),
-        accountId, containerId, isEncrypted, null);
+        accountId, containerId, isEncrypted, (short) 0, null);
   }
 
   /**
@@ -62,7 +63,7 @@ public class BlobProperties {
   public BlobProperties(long blobSize, String serviceId, String ownerId, String contentType, boolean isPrivate,
       long timeToLiveInSeconds, short accountId, short containerId, boolean isEncrypted, String externalAssetTag) {
     this(blobSize, serviceId, ownerId, contentType, isPrivate, timeToLiveInSeconds,
-        SystemTime.getInstance().milliseconds(), accountId, containerId, isEncrypted, externalAssetTag);
+        SystemTime.getInstance().milliseconds(), accountId, containerId, isEncrypted, (short) 0, externalAssetTag);
   }
 
   /**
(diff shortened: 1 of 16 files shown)


ambry-messageformat/src/main/java/com.github.ambry.messageformat/SubRecord.java
```diff
@@ -0,0 +1,8 @@
+package com.github.ambry.messageformat;
```
jsjtzyy: Please add license in this file. Travis build failed due to following issue:
```
* What went wrong:
Execution failed for task ':ambry-messageformat:licenseMain'.
> License violations were found: /home/travis/build/linkedin/ambry/ambry-messageformat/src/main/java/com.github.ambry.messageformat/SubRecord.java}
```

ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
```diff
@@ -32,6 +32,7 @@
   private final boolean isEncrypted;
   private long blobSize;
   private long timeToLiveInSeconds;
+  private short recordVersion;
```
zzmao: Since both Put record and Update record need `version`, can we add `version` in a common place? `MessageHeader_Format` looks a good place for me.

Can we name it to a `special` name to avoid confusion with record version or format version? `updateVersion` or `evolutionVersion`?

ambry-messageformat/src/main/java/com.github.ambry.messageformat/SubRecord.java
```diff
@@ -0,0 +1,8 @@
...
+
+public abstract class SubRecord {
```
zzmao: License and java doc.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/SubRecord.java
dharju: oops, thought I caught these.  Pushing change

ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
dharju: I agree that `recordVersion` may not be the best name.  `updateVersion` I think may be clearer so I'm open to changing to that.

I understand your motivation for wanting to put it in the header due to its being shared by both PUT blobs and UPDATE blobs, but I don't think it's the right place as the purpose of the header is simply to tell the receiving parser the payload size and field demarcations of the payload.  Plugging in `updateVersion` into the message header feels like subverting the purpose of the message header, and adding identification logic in a place that it doesn't belong.  Also TTLUpdate doesn't use `updateVersion`

ambry-messageformat/src/main/java/com.github.ambry.messageformat/SubRecord.java
dharju: will add

ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
zzmao: BlobProperties is for user, we shouldn't include `updateVersion` to confuse them.

The message header is a metadata container. `updateVersion` is also a kind of metadata to all messages. Plus, It's shared by all message.

Adding version identification logic based on message header is what we want. For example, in `BlobStoreRecover.recover()`, we need `updateVersion` to recover `Index` which should also include a `updateVersion`. With version in message header, all message can share same code and recover with easy logic. Particularly, you don't need to parse BlobProperties for PUT. 
 
I understand ttlUpdate doesn't need a version, but we should add version to it. Reasons are:
1. logic easy, just look one step back and increase version number, for all kind of message.
2. open the door of multiple ttlUpdate.

jsjtzyy:
@dharju could your rebase your branch. I feel like, for some reason, current PR incorporates code changes from other PRs.  More than 3600 lines change and some codes are for AccountService.

dharju:
@jsjtzyy force-pushed rebase

ambry-api/src/main/java/com.github.ambry/messageformat/BlobProperties.java
dharju: Does that mean you're ok with the current code or do you still think having it in the message header is preferable?  Did you happen to look at how indexing, compaction, and/or replication are done at a deeper level and see if one way would make implementing those other needed functions easier?

zzmao: Prefer message header. 

Looking at index, compaction, replication. Will send out a doc by EOW.

jsjtzyy: Have we reached an agreement on 1. where the version should be placed. 2. explicit rules of updating the versions ? According to original Priyesh's design, it seems we use convention something like P0, D0, U0, D1, U1 ...

jsjtzyy:
@dharju Can we close this PR? I think you have another PR including same changes that is already merged.

jsjtzyy:
Close this PR, the same changes are already merged.

PR #1242, opened 2019-08-19
Use Timer instead of Histgram on some metrics

zzmao:
For metrics below, we want to know both latency histogram and rate. 
Timer, as a combination of Histogram and Meter, is the best choice.

networkClientSendAndPollTime
networkClientRequestQueueTime
networkClientRoundTripTime
networkClientTotalTime
putChunkOperationLatencyMs


justinlin-linkedin:
Can you mention the reason why you want to use Timer, in the description?

PR #1243, opened 2019-08-19
Add config option to connect to Cosmos with DirectHttps mode

lightningrob:


ambry-cloud/src/test/resources/azure-test.properties
```diff
@@ -13,3 +13,4 @@ azure.storage.connection.string=<storage-account-credentials>
 cosmos.endpoint=<cosmos-url>
 cosmos.collection.link=/dbs/ambry-metadata/colls/blob-metadata
 cosmos.key=<cosmos-key>
+cosmos.direct.https=true
```
ankagrawal: minor: do we need to specify this in the sample properties file?

PR #1252, opened 2019-08-29
HelixAccountService: limit the number of versions in the list

justinlin-linkedin:
Enforcing a upper bound on how many versions of account metadata to save in the ambry-server.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
@@ -317,7 +317,7 @@ class RouterStore extends AccountMetadataStore {
           // Block this execution? or maybe wait for a while then get out?
           router.get().deleteBlob(newBlobID, SERVICE_ID).get();
         } catch (Exception e) {
-          logger.error("Failed to delete blob={} because of {}", newBlobID, e);
+          logger.error("Failed to delete blob=" + newBlobID, e);
           accountServiceMetrics.accountDeletesToAmbryServerErrorCount.inc();
         }
       }
@@ -330,7 +330,7 @@ class RouterStore extends AccountMetadataStore {
             // Block this execution? or maybe wait for a while then get out?
             router.get().deleteBlob(blobID, SERVICE_ID).get();
           } catch (Exception e) {
-            logger.error("Failed to delete blob={} because of {}", blobID, e);
+            logger.error("Failed to delete blob=" + blobID, e);
             accountServiceMetrics.accountDeletesToAmbryServerErrorCount.inc();
           }
         }
--- a/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
@@ -26,7 +26,9 @@ import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -61,8 +63,8 @@ class RouterStore extends AccountMetadataStore {
   private static final Short CONTAINER_ID = Container.HELIX_ACCOUNT_SERVICE_CONTAINER_ID;
   private static final String SERVICE_ID = "helixAccountService";
 
+  private final int maxNumberOfVersionsToSave;
   private final AtomicReference<Router> router;
-
   // If forBackFill is true, then when updating the account metadata, we don't create backup files and we don't merge
   // accounts from ambry-server with the provided accounts set.
   private final boolean forBackFill;
@@ -76,10 +78,11 @@ class RouterStore extends AccountMetadataStore {
    * @param forBackFill True if this {@link RouterStore} is created for backfill accounts to new zookeeper node.
    */
   RouterStore(AccountServiceMetrics accountServiceMetrics, BackupFileManager backupFileManager,
-      HelixPropertyStore<ZNRecord> helixStore, AtomicReference<Router> router, boolean forBackFill) {
+      HelixPropertyStore<ZNRecord> helixStore, AtomicReference<Router> router, boolean forBackFill, int maxNumberOfVersionsToSave) {
     super(accountServiceMetrics, backupFileManager, helixStore, ACCOUNT_METADATA_BLOB_IDS_PATH);
     this.router = router;
     this.forBackFill = forBackFill;
+    this.maxNumberOfVersionsToSave = maxNumberOfVersionsToSave;
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -163,6 +168,16 @@ static String writeAccountMapToRouter(Map<String, String> accountMap, Router rou
     return router.putBlob(properties, null, channel, PutBlobOptions.DEFAULT).get();
   }
 
+  /**
+   * Helper function to log the error message out and throw an {@link IllegalStateException}.
+   * @param errorMessage The error message.
+   * @param cause The cause exception.
+   */
+  private void logAndThrowIllegalStateException(String errorMessage, Exception cause) {
+    logger.error(errorMessage, cause);
```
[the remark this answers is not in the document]
justinlin-linkedin: This is actually a callback function, where the caller would ignore all the exceptions thrown by this callback function. That's why we need to log some error message out.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -197,76 +215,68 @@ public ZNRecord update(ZNRecord znRecord) {
...
         try {
-          BlobIDAndVersion blobIDAndVersion = null;
-          for (String accountBlobIDInJson : accountBlobIDs) {
-            BlobIDAndVersion current = BlobIDAndVersion.fromJson(accountBlobIDInJson);
-            if (blobIDAndVersion == null || blobIDAndVersion.version < current.version) {
-              blobIDAndVersion = current;
-            }
-          }
+          // Parse the json string list and get the BlobIDAndVersion with the latest version number.
+          blobIDAndVersionsJson.stream()
```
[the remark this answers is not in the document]
justinlin-linkedin: good catch.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -197,76 +215,68 @@ public ZNRecord update(ZNRecord znRecord) {
...
+              .forEach(accountBlobIDInJson -> blobIDAndVersions.add(BlobIDAndVersion.fromJson(accountBlobIDInJson)));
+          Collections.sort(blobIDAndVersions, Comparator.comparing(BlobIDAndVersion::getVersion));
+          BlobIDAndVersion blobIDAndVersion = blobIDAndVersions.get(blobIDAndVersions.size() - 1);
```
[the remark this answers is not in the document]
justinlin-linkedin: the difference is that one need to sort the list and the other only has to get the latest blobIDAndVersion. I figure it's too small to create a function for.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -277,15 +284,33 @@ public ZNRecord update(ZNRecord znRecord) {
...
-      accountBlobIDs.add(new BlobIDAndVersion(this.newBlobID, newVersion).toJson());
-      recordToUpdate.setListField(ACCOUNT_METADATA_BLOB_IDS_LIST_KEY, accountBlobIDs);
+      if (blobIDAndVersions.size() + 1 > totalNumberOfVersionToKeep) {
+        Iterator<BlobIDAndVersion> iter = blobIDAndVersions.iterator();
+        while (blobIDAndVersions.size() + 1 > totalNumberOfVersionToKeep) {
+          BlobIDAndVersion blobIDAndVersion = iter.next();
+          iter.remove();
+          logger.info("Removing blob " + blobIDAndVersion.getBlobID() + " at version " + blobIDAndVersion.getVersion());
+          try {
+            router.get().deleteBlob(blobIDAndVersion.getBlobID(), SERVICE_ID).get();
```
cgtz: Are there any edge cases where we would not want to delete the oldest blob from within the update logic. 
The main case to consider is probably where the update to zookeeper does not succeed. In this case we would have deleted a blob in ambry that could still exist in the current znode.

However, this may not be a problem if `totalNumberOfVersionsToKeep` is sufficiently large that the chance of someone else still interested in reading that blob is very small.

ambry-api/src/main/java/com.github.ambry/config/HelixAccountServiceConfig.java
```diff
@@ -98,6 +101,15 @@
...
   public final boolean enableServeFromBackup;
 
+  /**
+   * Total number of previous versions of account metadata to keep in the system. Every update account http request would
+   * generate a new version. And when the number of versions surpasses this number, HelixAccountService will purge the
+   * oldest one to make room for the new one.
+   */
+  @Config(TOTAL_NUMBER_OF_VERSION_TO_KEEP)
+  @Default("100")
+  public final int totalNumberOfVersionToKeep;
```
cgtz: Ah, probably if this is set to 100, the deleteBlob is likely to not be an issue.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -197,76 +215,65 @@ public ZNRecord update(ZNRecord znRecord) {
...
       }
-
-      String errorMessage = null;
-      List<String> accountBlobIDs = recordToUpdate.getListField(ACCOUNT_METADATA_BLOB_IDS_LIST_KEY);
+      // This is the version number for the new blob id.
       int newVersion = 1;
-      Map<String, String> accountMap = null;
-      if (accountBlobIDs != null && accountBlobIDs.size() != 0) {
-        // parse the json string list and get the blob id with the latest version
+      List<BlobIDAndVersion> blobIDAndVersions = new ArrayList<>();
```
cgtz: Maybe consider using a LinkedList if you are going to be removing from the start of the list, since that can be O(n) in an ArrayList.
Alternatively, you can sort in reverse order and remove from the end of the ArrayList

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -277,15 +284,33 @@ public ZNRecord update(ZNRecord znRecord) {
...
+          logger.info("Removing blob " + blobIDAndVersion.getBlobID() + " at version " + blobIDAndVersion.getVersion());
+          try {
+            router.get().deleteBlob(blobIDAndVersion.getBlobID(), SERVICE_ID).get();
+          } catch (Exception e) {
+            logger.error("Failed to delete blob={} from older version because of {}", blobIDAndVersion.getBlobID(), e);
+            accountServiceMetrics.accountDeletesToAmbryServerErrorCount.inc();
+          }
+        }
+        // Clear json string list since it's not sorted.
+        blobIDAndVersionsJson.clear();
```
cgtz: Could you just do `blobIDAndVersionsJson = blobIDAndVersions.stream().map(BlobIDAndVersion::toJson).collect(Collectors.toList())`?

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
justinlin-linkedin: delete the blobs after successfully updates zookeeper node.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
justinlin-linkedin: good catch, I will use linkedlist instead.

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -319,13 +313,28 @@ public void afterUpdate(boolean isUpdateSucceeded) {
...
+      // Notice this logic might end up with the dangling blobs, when the process crashes before the for loop.
+      // But since the frequency to update account metadata is pretty rare, it won't be a big problem.
+      if (isUpdateSucceeded && oldBlobIDsToDelete != null) {
+        for (String blobID : oldBlobIDsToDelete) {
+          try {
+            logger.info("Removing blob " + blobID);
+            // Block this execution? or maybe wait for a while then get out?
+            router.get().deleteBlob(blobID, SERVICE_ID).get();
+          } catch (Exception e) {
+            logger.error("Failed to delete blob={} because of {}", blobID, e);
```
cgtz: minor: take out the last `{}`. Otherwise, only the exception message will appear in the log instead of the stack trace

PR #1258, opened 2019-09-04
Add chunk ID to get exceptions, add max in-mem chunks config

cgtz:
Add the chunk ID to the exception messages for any exceptions that
occur when fetching a data chunk in GetBlobOperation.
    
Add config for the max number of data chunks to keep in memory for
a single PutOperations or GetBlobOperation. This also determines
the max number of chunk requests that may be in flight to a server
node in a single operation.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
@@ -343,11 +343,18 @@ public class RouterConfig {
   public final long routerOperationTrackerHistogramDumpPeriod;
 
   /**
-   * The max number of chunks per GetBlobOperation or PutOperation that may be buffered in memory.
+   * The max number of chunks per PutOperation that may be buffered in memory.
    */
-  @Config("router.max.in.mem.chunks")
+  @Config("router.max.in.mem.put.chunks")
   @Default("4")
-  public final int routerMaxInMemChunks;
+  public final int routerMaxInMemPutChunks;
+
+  /**
+   * The max number of chunks per GetBlobOperation that may be buffered in memory.
+   */
+  @Config("router.max.in.mem.get.chunks")
+  @Default("4")
+  public final int routerMaxInMemGetChunks;
 
   /**
    * Create a RouterConfig instance.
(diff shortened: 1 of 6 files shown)


ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -342,6 +342,13 @@
   @Default("600")
   public final long routerOperationTrackerHistogramDumpPeriod;
 
+  /**
+   * The max number of chunks per GetBlobOperation or PutOperation that may be buffered in memory.
+   */
+  @Config("router.max.in.mem.chunks")
+  @Default("4")
```
ankagrawal: How did we arrive at this default value?

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
```diff
@@ -429,5 +436,6 @@ public RouterConfig(VerifiableProperties verifiableProperties) {
     }
     routerOperationTrackerTerminateOnNotFoundEnabled =
         verifiableProperties.getBoolean("router.operation.tracker.terminate.on.not.found.enabled", false);
+    routerMaxInMemChunks = verifiableProperties.getIntInRange("router.max.in.mem.chunks", 4, 1, Integer.MAX_VALUE);
```
ankagrawal: Also Integer.MAX_VALUE might not be a valid end for the range. We cannot buffer so much data in memory. Maybe Integer.MAX_VALUE/CHUNK_SIZE(8MB) is a better end?

ambry-router/src/test/java/com.github.ambry.router/PutOperationTest.java
```diff
@@ -103,7 +103,7 @@ public void testSendIncomplete() throws Exception {
         mockNetworkClient.getAndClearWokenUpStatus());
     // A poll should therefore return requestParallelism number of requests from each chunk
     op.poll(requestRegistrationCallback);
-    Assert.assertEquals(NonBlockingRouter.MAX_IN_MEM_CHUNKS * requestParallelism, requestInfos.size());
+    Assert.assertEquals(routerConfig.routerMaxInMemChunks * requestParallelism, requestInfos.size());
 
     // There are MAX_IN_MEM_CHUNKS + 1 data chunks for this blob (and a metadata chunk).
```
ankagrawal: We need to update this comment as it references MAX_IN_MEM_CHUNKS.

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
cgtz: Good suggestion!

ambry-api/src/main/java/com.github.ambry/config/RouterConfig.java
cgtz: This came from the constant that was previously used, `NonBlockingRouter.MAX_IN_MEM_CHUNKS`

PR #1259, opened 2019-09-09
Support adding/removing store via AdminRequest

jsjtzyy:
1. Refactored store control admin request to support adding/removing store.
2. Changes made in replication manager to support adding/removing replica
3. Added method in clustermap that allows caller to get detailed info of new replica from Helix PropertyStore. 
4. Changed removeBlobStore in StorageManager to return the store if it is successfully removed.
5. Moved MockStorageManager to a separate file which can be re-used by other tests.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
@@ -123,6 +123,14 @@ public interface ClusterMap extends AutoCloseable {
   JSONObject getSnapshot();
 
   /**
+   * Attempt to get new replica of certain partition that resides on given data node.
+   * @param partitionIdStr the partition id string
+   * @param dataNodeId the {@link DataNodeId} on which new replica is placed
+   * @return {@link ReplicaId} if there is a new replica satisfying given partition and data node. {@code null} otherwise.
+   */
+  ReplicaId getNewReplica(String partitionIdStr, DataNodeId dataNodeId);
+
+  /**
    * Close the cluster map. Any cleanups should be done in this call.
    */
   @Override
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
@@ -123,12 +123,14 @@ public interface ClusterMap extends AutoCloseable {
   JSONObject getSnapshot();
 
   /**
-   * Attempt to get new replica of certain partition that resides on given data node.
+   * Attempt to get a bootstrap replica of certain partition that is supposed to be added onto specified data node.
+   * This method is designed to fetch detailed infos about bootstrap replica and create an instance of this replica. The
+   * purpose is to support dynamically adding new replica to specified data node.
    * @param partitionIdStr the partition id string
-   * @param dataNodeId the {@link DataNodeId} on which new replica is placed
+   * @param dataNodeId the {@link DataNodeId} on which bootstrap replica is placed
    * @return {@link ReplicaId} if there is a new replica satisfying given partition and data node. {@code null} otherwise.
    */
-  ReplicaId getNewReplica(String partitionIdStr, DataNodeId dataNodeId);
+  ReplicaId getBootstrapReplica(String partitionIdStr, DataNodeId dataNodeId);
 
   /**
    * Close the cluster map. Any cleanups should be done in this call.
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
@@ -29,6 +29,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
@@ -87,6 +88,7 @@ class HelixClusterManager implements ClusterMap {
   final HelixClusterManagerMetrics helixClusterManagerMetrics;
   private final PartitionSelectionHelper partitionSelectionHelper;
   private final Map<String, Map<String, String>> partitionOverrideInfoMap = new HashMap<>();
+  private ZkHelixPropertyStore<ZNRecord> helixPropertyStoreInLocalDc = null;
   // The current xid currently does not change after instantiation. This can change in the future, allowing the cluster
   // manager to dynamically incorporate newer changes in the cluster. This variable is atomic so that the gauge metric
   // reflects the current value.
@@ -216,13 +218,13 @@ class HelixClusterManager implements ClusterMap {
     DcZkInfo dcZkInfo = dataCenterToZkAddress.get(clusterMapConfig.clusterMapDatacenterName);
     String zkConnectStr = dcZkInfo.getZkConnectStr();
     HelixManager manager;
-    ZkHelixPropertyStore<ZNRecord> helixPropertyStore;
     manager = helixFactory.getZKHelixManager(clusterName, instanceName, InstanceType.SPECTATOR, zkConnectStr);
     logger.info("Connecting to Helix manager in local zookeeper at {}", zkConnectStr);
     manager.connect();
     logger.info("Established connection to Helix manager in local zookeeper at {}", zkConnectStr);
-    helixPropertyStore = manager.getHelixPropertyStore();
-    logger.info("HelixPropertyStore from local datacenter {} is: {}", dcZkInfo.getDcName(), helixPropertyStore);
+    helixPropertyStoreInLocalDc = manager.getHelixPropertyStore();
+    logger.info("HelixPropertyStore from local datacenter {} is: {}", dcZkInfo.getDcName(),
+        helixPropertyStoreInLocalDc);
     IZkDataListener dataListener = new IZkDataListener() {
       @Override
       public void handleDataChange(String dataPath, Object data) {
@@ -235,10 +237,9 @@ class HelixClusterManager implements ClusterMap {
       }
     };
     logger.info("Subscribing data listener to HelixPropertyStore.");
-    helixPropertyStore.subscribeDataChanges(ClusterMapUtils.PARTITION_OVERRIDE_ZNODE_PATH, dataListener);
-    logger.info("Getting ZNRecord from HelixPropertyStore");
-    ZNRecord zNRecord =
-        helixPropertyStore.get(ClusterMapUtils.PARTITION_OVERRIDE_ZNODE_PATH, null, AccessOption.PERSISTENT);
+    helixPropertyStoreInLocalDc.subscribeDataChanges(PARTITION_OVERRIDE_ZNODE_PATH, dataListener);
+    logger.info("Getting PartitionOverride ZNRecord from HelixPropertyStore");
+    ZNRecord zNRecord = helixPropertyStoreInLocalDc.get(PARTITION_OVERRIDE_ZNODE_PATH, null, AccessOption.PERSISTENT);
     if (clusterMapConfig.clusterMapEnablePartitionOverride) {
       if (zNRecord != null) {
         partitionOverrideInfoMap.putAll(zNRecord.getMapFields());
@@ -379,6 +380,55 @@ class HelixClusterManager implements ClusterMap {
   @Override
   public List<PartitionId> getAllPartitionIds(String partitionClass) {
     return partitionSelectionHelper.getPartitions(partitionClass);
+  }
+
+  @Override
+  public ReplicaId getNewReplica(String partitionIdStr, DataNodeId dataNodeId) {
+    ReplicaId newReplica = null;
+    logger.info("Getting ReplicaAddition ZNRecord from HelixPropertyStore in local DC.");
+    ZNRecord zNRecord = helixPropertyStoreInLocalDc.get(REPLICA_ADDITION_ZNODE_PATH, null, AccessOption.PERSISTENT);
+    if (zNRecord != null) {
+      String instanceName = getInstanceName(dataNodeId.getHostname(), dataNodeId.getPort());
+      Map<String, Map<String, String>> partitionToReplicas = zNRecord.getMapFields();
+      Map<String, String> replicaInfos = partitionToReplicas.get(partitionIdStr);
+      if (replicaInfos != null && replicaInfos.containsKey(instanceName)) {
+        long replicaCapacity = Long.valueOf(replicaInfos.get(REPLICAS_CAPACITY_STR));
+        String partitionClass = replicaInfos.get(PARTITION_CLASS_STR);
+        AmbryPartition mappedPartition = partitionNameToAmbryPartition.get(partitionIdStr);
+        if (mappedPartition == null) {
+          logger.info("Partition {} is currently not present in cluster map, creating a new partition.",
+              partitionIdStr);
+          mappedPartition =
+              new AmbryPartition(Long.valueOf(partitionIdStr), partitionClass, helixClusterManagerCallback);
+        }
+        // Check if data or disk is in current cluster map, if not, set newReplica to null.
+        AmbryDataNode dataNode = instanceNameToAmbryDataNode.get(instanceName);
+        String mountPathFromHelix = replicaInfos.get(instanceName);
+        Set<AmbryDisk> disks = dataNode != null ? ambryDataNodeToAmbryDisks.get(dataNode) : null;
+        Optional<AmbryDisk> potentialDisk =
+            disks != null ? disks.stream().filter(d -> d.getMountPath().equals(mountPathFromHelix)).findAny()
+                : Optional.empty();
+        if (dataNode != null && potentialDisk.isPresent()) {
+          try {
+            newReplica =
+                new AmbryReplica(clusterMapConfig, mappedPartition, potentialDisk.get(), true, replicaCapacity, false);
+          } catch (Exception e) {
+            logger.error("Failed to create new replica for partition {} on {} due to exception: ", partitionIdStr,
+                instanceName, e);
+            newReplica = null;
+          }
+        } else {
+          logger.error(
+              "Either datanode or disk that associated with new replica is not found in cluster map. Cannot create new replica.");
+        }
+      } else {
+        logger.warn("Partition {} or replica on host {} is not found in replica info map", partitionIdStr,
+            instanceName);
+      }
+    } else {
+      logger.warn("ZNRecord from HelixPropertyStore is NULL, partition to replicaInfo map doesn't exist.");
+    }
+    return newReplica;
   }
 
   /**
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
@@ -382,51 +382,72 @@ class HelixClusterManager implements ClusterMap {
     return partitionSelectionHelper.getPartitions(partitionClass);
   }
 
+  /**
+   * {@inheritDoc}
+   * To create bootstrap replica, {@link HelixClusterManager} needs to fetch replica info (i.e. capacity, mount path)
+   * from Helix PropertyStore. This method looks up the ZNode in local datacenter and does some validation. Right now,
+   * {@link HelixClusterManager} supports getting bootstrap replica of new partition but it doesn't support getting replica
+   * residing on hosts that are not present in clustermap.
+   * The ZNRecord of REPLICA_ADDITION_ZNODE has following format in mapFields.
+   * <p/>
+   * "mapFields": {
+   *     "1": {
+   *         "replicaCapacityInBytes": 107374182400,
+   *         "partitionClass": "max-replicas-all-datacenters",
+   *         "localhost1_17088": "/tmp/c/1",
+   *         "localhost2_17088": "/tmp/d/1"
+   *     },
+   *     "2": {
+   *         "replicaCapacityInBytes": 107374182400,
+   *         "partitionClass": "max-replicas-all-datacenters",
+   *         "localhost3_17088": "/tmp/e/1"
+   *     }
+   * }
+   * In above example, two bootstrap replicas of partition[1] will be added to localhost1 and localhost2 respectively.
+   * The host name is followed by mount path on which the bootstrap replica should be placed.
+   */
   @Override
-  public ReplicaId getNewReplica(String partitionIdStr, DataNodeId dataNodeId) {
+  public ReplicaId getBootstrapReplica(String partitionIdStr, DataNodeId dataNodeId) {
     ReplicaId newReplica = null;
     logger.info("Getting ReplicaAddition ZNRecord from HelixPropertyStore in local DC.");
     ZNRecord zNRecord = helixPropertyStoreInLocalDc.get(REPLICA_ADDITION_ZNODE_PATH, null, AccessOption.PERSISTENT);
-    if (zNRecord != null) {
-      String instanceName = getInstanceName(dataNodeId.getHostname(), dataNodeId.getPort());
-      Map<String, Map<String, String>> partitionToReplicas = zNRecord.getMapFields();
-      Map<String, String> replicaInfos = partitionToReplicas.get(partitionIdStr);
-      if (replicaInfos != null && replicaInfos.containsKey(instanceName)) {
-        long replicaCapacity = Long.valueOf(replicaInfos.get(REPLICAS_CAPACITY_STR));
-        String partitionClass = replicaInfos.get(PARTITION_CLASS_STR);
-        AmbryPartition mappedPartition = partitionNameToAmbryPartition.get(partitionIdStr);
-        if (mappedPartition == null) {
-          logger.info("Partition {} is currently not present in cluster map, creating a new partition.",
-              partitionIdStr);
-          mappedPartition =
-              new AmbryPartition(Long.valueOf(partitionIdStr), partitionClass, helixClusterManagerCallback);
-        }
-        // Check if data node or disk is in current cluster map, if not, set newReplica to null.
-        AmbryDataNode dataNode = instanceNameToAmbryDataNode.get(instanceName);
-        String mountPathFromHelix = replicaInfos.get(instanceName);
-        Set<AmbryDisk> disks = dataNode != null ? ambryDataNodeToAmbryDisks.get(dataNode) : null;
-        Optional<AmbryDisk> potentialDisk =
-            disks != null ? disks.stream().filter(d -> d.getMountPath().equals(mountPathFromHelix)).findAny()
-                : Optional.empty();
-        if (dataNode != null && potentialDisk.isPresent()) {
-          try {
-            newReplica =
-                new AmbryReplica(clusterMapConfig, mappedPartition, potentialDisk.get(), true, replicaCapacity, false);
-          } catch (Exception e) {
-            logger.error("Failed to create new replica for partition {} on {} due to exception: ", partitionIdStr,
-                instanceName, e);
-            newReplica = null;
-          }
-        } else {
-          logger.error(
-              "Either datanode or disk that associated with new replica is not found in cluster map. Cannot create new replica.");
-        }
-      } else {
-        logger.warn("Partition {} or replica on host {} is not found in replica info map", partitionIdStr,
-            instanceName);
+    if (zNRecord == null) {
+      logger.warn("ZNRecord from HelixPropertyStore is NULL, partition to replicaInfo map doesn't exist.");
+      return null;
+    }
+    String instanceName = getInstanceName(dataNodeId.getHostname(), dataNodeId.getPort());
+    Map<String, Map<String, String>> partitionToReplicas = zNRecord.getMapFields();
+    Map<String, String> replicaInfos = partitionToReplicas.get(partitionIdStr);
+    if (replicaInfos == null || !replicaInfos.containsKey(instanceName)) {
+      logger.warn("Partition {} or replica on host {} is not found in replica info map", partitionIdStr, instanceName);
+      return null;
+    }
+    long replicaCapacity = Long.parseLong(replicaInfos.get(REPLICAS_CAPACITY_STR));
+    String partitionClass = replicaInfos.get(PARTITION_CLASS_STR);
+    AmbryPartition mappedPartition = partitionNameToAmbryPartition.get(partitionIdStr);
+    if (mappedPartition == null) {
+      logger.info("Partition {} is currently not present in cluster map, creating a new partition.", partitionIdStr);
+      mappedPartition = new AmbryPartition(Long.parseLong(partitionIdStr), partitionClass, helixClusterManagerCallback);
+    }
+    // Check if data node or disk is in current cluster map, if not, set newReplica to null.
+    AmbryDataNode dataNode = instanceNameToAmbryDataNode.get(instanceName);
+    String mountPathFromHelix = replicaInfos.get(instanceName);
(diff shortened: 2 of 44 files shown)


jsjtzyy:
Still working on unit tests and integration tests. Also this PR partially depends on #1249

jsjtzyy:
This PR contains lots of changes and refactoring in testing. Reviewers can start with production code first.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +107,83 @@ public void start() throws ReplicationException {
...
+    logger.info("{} is successfully added into replication manager", replicaId.getPartitionId());
+    return true;
+  }
+
+  /**
+   * Remove replica from replication manager
+   * @param replicaId the replica to remove
+   * @return {@code true} if replica is successfully removed. {@code false} otherwise
+   */
+  public boolean removeReplica(ReplicaId replicaId) {
```
jsjtzyy: Here same method signature is intentionally adopted for this PR and #1262, which allows us to put an abstract method ReplicationEngine. Both regular ReplicationManager and VcrReplicationManager.

zzmao:
Have some difficulties to understand `populateRemoteReplicaAndPartitionInfos` and `getNewReplica`. Need to talk offline.

jsjtzyy:
#643

ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
```diff
@@ -122,6 +122,15 @@
    */
   JSONObject getSnapshot();
 
+  /**
+   * Get new replica of certain partition that resides on given host.
+   * @param partitionIdStr the partition id string
+   * @param hostname the host on which replica should reside.
```
zzmao: `currentNode.hostname` and `currentNode.port` are passed in, why not use `DataNode` as argument?

ambry-protocol/src/main/java/com.github.ambry.protocol/BlobStoreControlAdminRequest.java
```diff
@@ -35,21 +36,31 @@
...
+    short versionId = stream.readShort();
+    short numReplicasCaughtUpPerPartition;
+    BlobStoreControlRequestType controlRequestType;
+    switch (versionId) {
+      case VERSION_V1:
+        numReplicasCaughtUpPerPartition = stream.readShort();
+        controlRequestType =
+            stream.readByte() == 1 ? BlobStoreControlRequestType.StartStore : BlobStoreControlRequestType.StopStore;
+        break;
+      case VERSION_V2:
```
zzmao: Is `VERSION_V2` needed? Look like `BlobStoreControlRequestType.values()[stream.readByte()]` can fully cover `VERSION_V1`.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
```diff
@@ -533,6 +533,12 @@ public JSONObject getSnapshot() {
     return snapshot;
   }
 
+  @Override
+  public ReplicaId getNewReplica(String partitionIdStr, String hostname, Integer port) {
+    logger.warn("Adding new replica is currently not supported in static cluster manager. Return null here");
```
zzmao: throw `UnsupportedException`?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +107,83 @@ public void start() throws ReplicationException {
...
+  }
+
+  /**
+   * Given peer replicas of certain local replica, populate {@link RemoteReplicaInfo}(s) and {@link PartitionInfo} that
+   * associates with this local replica. Also this method updates partitionToPartitionInfo map and mountPathToPartitionInfos map.
+   * @param peerReplicas the list peer replicas of certain local replica
+   * @param replicaId the local replica
+   * @return list of {@link RemoteReplicaInfo} associated with local replica.
+   */
+  private List<RemoteReplicaInfo> populateRemoteReplicaAndPartitionInfos(List<? extends ReplicaId> peerReplicas,
```
zzmao: It's good to split this function to two: `createRemoteReplicaInfos()` and `updatePartitionInfoMaps`

ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
```diff
@@ -231,12 +231,13 @@ void shutdown() throws InterruptedException {
 
   /**
    * @param id the {@link PartitionId} to find the store for.
+   * @param skipStateCheck whether to skip checking state of store.
```
zzmao: Introduce `skipStateCheck` makes the interface complex and hard to understand. Can we alway return `store` and check `isStarted()` in the caller?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +107,83 @@ public void start() throws ReplicationException {
...
+      logger.error("{} already exists in replication manager, rejecting adding replica request.",
+          replicaId.getPartitionId());
+      return false;
+    }
+    List<? extends ReplicaId> peerReplicas = replicaId.getPeerReplicaIds();
+    List<RemoteReplicaInfo> remoteReplicaInfos = new ArrayList<>();
+    if (peerReplicas != null) {
+      remoteReplicaInfos = populateRemoteReplicaAndPartitionInfos(peerReplicas, replicaId);
+    }
+    logger.info("Assigning thread for {}", replicaId.getPartitionId());
```
zzmao: Move the logic below under `if (peerReplicas != null) {`?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +107,83 @@ public void start() throws ReplicationException {
...
+    if (!partitionToPartitionInfo.containsKey(replicaId.getPartitionId())) {
+      logger.error("{} doesn't exist in replication manager, skipping removing replica request.",
+          replicaId.getPartitionId());
+      return false;
+    }
+    PartitionInfo partitionInfo = partitionToPartitionInfo.get(replicaId.getPartitionId());
+    List<RemoteReplicaInfo> remoteReplicaInfos = partitionInfo.getRemoteReplicaInfos();
+    logger.info("Removing remote replicas of {} from replica threads", replicaId.getPartitionId());
+    removeRemoteReplicaInfoFromReplicaThread(remoteReplicaInfos);
+    mountPathToPartitionInfos.get(replicaId.getMountPath()).remove(partitionInfo);
```
zzmao: `computeIfPresent` is atomic for concurrent map. This can be used to avoid heavy `copyOnWriteArrayList`

ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
```diff
@@ -122,6 +122,15 @@
    */
   JSONObject getSnapshot();
 
+  /**
+   * Get new replica of certain partition that resides on given host.
```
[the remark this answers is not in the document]
jsjtzyy: It's possible that one host has several new replicas but they should come from different partitions. In this method, neither argument `partitionIdStr` nor return value `ReplicaId` is collection, because the method is supposed to be called by state model where each partition get its new replica separately.  I will add some comments for this method.

ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
jsjtzyy: good idea. I will figure out a way to pass in `DataNodeId`

jsjtzyy: sure, thanks for letting me know.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
jsjtzyy: fixed

ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
```diff
@@ -231,12 +231,13 @@ void shutdown() throws InterruptedException {
...
    * @return the associated {@link Store}, or {@code null} if the partition is not on this disk, or the store is not
    *         started.
    */
-  Store getStore(PartitionId id) {
```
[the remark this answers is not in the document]
jsjtzyy: right, I made the change as you suggested.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/MockClusterMap.java
```diff
@@ -293,11 +307,11 @@ public DataNodeId getDataNodeId(String hostname, int port) {
 
   @Override
   public List<ReplicaId> getReplicaIds(DataNodeId dataNodeId) {
-    ArrayList<ReplicaId> replicaIdsToReturn = new ArrayList<ReplicaId>();
+    ArrayList<ReplicaId> replicaIdsToReturn = new ArrayList<>();
     for (PartitionId partitionId : partitions.values()) {
       List<? extends ReplicaId> replicaIds = partitionId.getReplicaIds();
       for (ReplicaId replicaId : replicaIds) {
-        if (replicaId.getDataNodeId().getHostname().compareTo(dataNodeId.getHostname()) == 0
+        if (replicaId.getDataNodeId().getHostname().equals(dataNodeId.getHostname())
```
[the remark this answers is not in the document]
jsjtzyy: cool, will take your advice.

ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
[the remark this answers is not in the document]
zzmao: Is `getStore(PartitionId, false)` really needed?  Can we add extra logic to check after `getStore(PartitionId)`? 

Less interface is better than more.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: done

ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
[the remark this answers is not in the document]
jsjtzyy: I see your point here. I temporarily don't want to change current logic in most cases where `getStore(partitionId)` is called. However, I do need to get the store that is already shutdown when removing replica. Note that, remove replica always happens after stop replica. Hence, this method allows us to get a "stopped" store.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: good point. Let me remove `copyOnWriteArrayList`

ambry-clustermap/src/test/java/com.github.ambry.clustermap/MockClusterMap.java
```diff
@@ -436,6 +450,23 @@ public JSONObject getSnapshot() {
     return snapshot;
   }
 
+  @Override
+  public ReplicaId getNewReplica(String partitionIdStr, String hostname, Integer port) {
+    ReplicaId newReplica = null;
+    for (Map.Entry<Long, PartitionId> entry : partitions.entrySet()) {
```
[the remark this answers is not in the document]
jsjtzyy: correct, changes are made as you suggested

ambry-protocol/src/main/java/com.github.ambry.protocol/BlobStoreControlAdminRequest.java
jsjtzyy: Thanks @zzmao @ankagrawal for your inputs here. You are right, no need to keep version here. I will simplify this piece of code.

ambry-protocol/src/main/java/com.github.ambry.protocol/BlobStoreControlAdminRequest.java
```diff
@@ -35,21 +36,31 @@
...
     }
-    short numReplicasCaughtUpPerPartition = stream.readShort();
-    boolean enable = stream.readByte() == 1;
-    return new BlobStoreControlAdminRequest(numReplicasCaughtUpPerPartition, enable, adminRequest);
+    return new BlobStoreControlAdminRequest(numReplicasCaughtUpPerPartition, controlRequestType, adminRequest);
   }
 
-  public BlobStoreControlAdminRequest(short numReplicasCaughtUpPerPartition, boolean enable,
-      AdminRequest adminRequest) {
+  public BlobStoreControlAdminRequest(short numReplicasCaughtUpPerPartition,
```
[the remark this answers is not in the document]
jsjtzyy: sure

jsjtzyy:
Gentle reminder to review. Thanks in advance.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +107,97 @@ public void start() throws ReplicationException {
...
+  /**
+   * Remove replica from replication manager
+   * @param replicaId the replica to remove
+   * @return {@code true} if replica is successfully removed. {@code false} otherwise
+   */
+  public boolean removeReplica(ReplicaId replicaId) {
+    if (!partitionToPartitionInfo.containsKey(replicaId.getPartitionId())) {
+      logger.error("{} doesn't exist in replication manager, skipping removing replica request.",
+          replicaId.getPartitionId());
+      return false;
```
ankagrawal: Can the add/remove replica code cause thread safety concerns ? Let's discuss! There are some subtle differences between ConcurrentMap and ConcurrentHashMap and we should try to explicitly use ConcurrentHashMap is we want certain guarantees.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -381,6 +382,55 @@ public PartitionId getRandomWritablePartition(String partitionClass, List<Partit
...
+        String partitionClass = replicaInfos.get(PARTITION_CLASS_STR);
+        AmbryPartition mappedPartition = partitionNameToAmbryPartition.get(partitionIdStr);
+        if (mappedPartition == null) {
+          logger.info("Partition {} is currently not present in cluster map, creating a new partition.",
+              partitionIdStr);
+          mappedPartition =
+              new AmbryPartition(Long.valueOf(partitionIdStr), partitionClass, helixClusterManagerCallback);
+        }
+        // Check if data node or disk is in current cluster map, if not, set newReplica to null.
+        AmbryDataNode dataNode = instanceNameToAmbryDataNode.get(instanceName);
```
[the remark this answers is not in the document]
jsjtzyy: Discussed this offline. Currently `AmbryRequests` calls this method. Eventually, state model of each partition will call this method.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -381,6 +382,55 @@ public PartitionId getRandomWritablePartition(String partitionClass, List<Partit
...
+        String mountPathFromHelix = replicaInfos.get(instanceName);
+        Set<AmbryDisk> disks = dataNode != null ? ambryDataNodeToAmbryDisks.get(dataNode) : null;
+        Optional<AmbryDisk> potentialDisk =
+            disks != null ? disks.stream().filter(d -> d.getMountPath().equals(mountPathFromHelix)).findAny()
```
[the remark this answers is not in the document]
jsjtzyy: SRE is supposed to guarantee new added replica has valid mount path otherwise the creation should fail.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +107,97 @@ public void start() throws ReplicationException {
...
+    }
+    PartitionInfo partitionInfo = partitionToPartitionInfo.get(replicaId.getPartitionId());
+    List<RemoteReplicaInfo> remoteReplicaInfos = partitionInfo.getRemoteReplicaInfos();
+    logger.info("Removing remote replicas of {} from replica threads", replicaId.getPartitionId());
+    removeRemoteReplicaInfoFromReplicaThread(remoteReplicaInfos);
+    mountPathToPartitionInfos.computeIfPresent(replicaId.getMountPath(), (k, v) -> {
+      v.remove(partitionInfo);
+      return v;
+    });
+    //mountPathToPartitionInfos.get(replicaId.getMountPath()).remove(partitionInfo);
```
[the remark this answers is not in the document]
jsjtzyy: done

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: Good point. I decided to add `ReadWriteLock` in replication manager and disk manager.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -381,6 +382,55 @@ public PartitionId getRandomWritablePartition(String partitionClass, List<Partit
...
+                : Optional.empty();
+        if (dataNode != null && potentialDisk.isPresent()) {
```
cgtz: Can you remove the dataNode != null check since you check that on line 407?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
```diff
@@ -533,6 +533,12 @@ public JSONObject getSnapshot() {
     return snapshot;
   }
 
+  @Override
+  public ReplicaId getNewReplica(String partitionIdStr, DataNodeId dataNodeId) {
+    throw new UnsupportedOperationException(
+        "Adding new replica is currently not supported in static cluster manager. Return null here");
```
cgtz: Should this return null like the exception message says?

ambry-protocol/src/main/java/com.github.ambry.protocol/BlobStoreControlAdminRequest.java
```diff
@@ -35,21 +35,25 @@
    */
   public static BlobStoreControlAdminRequest readFrom(DataInputStream stream, AdminRequest adminRequest)
       throws IOException {
-    Short versionId = stream.readShort();
-    if (!versionId.equals(VERSION_V1)) {
-      throw new IllegalStateException("Unrecognized version for BlobStoreControlAdminRequest: " + versionId);
-    }
+    // read the version
+    stream.readShort();
```
cgtz: So, are we just erasing all remnants of V1 of this request? I guess thats ok since this is an admin operation that isn't called by the router in a live environment. However, you should probably add some notes that say why v1 is not present?

Do you still want to have the check that the version number matches the expected value (2) ?

ambry-protocol/src/main/java/com.github.ambry.protocol/BlobStoreControlRequestType.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.protocol;
+
+/**
+ * Enum of types of BlobStore control request.
+ * The order of these enums should not be changed since their relative position goes into the serialized form of
+ * requests.
+ */
+public enum BlobStoreControlRequestType {
```
cgtz: `BlobStoreControlAction`?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +110,111 @@ public void start() throws ReplicationException {
...
+  }
+
+  /**
+   * Remove replica from replication manager
+   * @param replicaId the replica to remove
+   * @return {@code true} if replica is successfully removed. {@code false} otherwise
+   */
+  public boolean removeReplica(ReplicaId replicaId) {
+    boolean succeed = false;
+    rwLock.readLock().lock();
```
cgtz: why do all of these methods that mutate the data structures take the read lock? shouldn't they take the write lock since they mutate the data structure?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +110,111 @@ public void start() throws ReplicationException {
...
+    PartitionId partition = replicaId.getPartitionId();
+    Store store = storageManager.getStore(partition);
+    for (ReplicaId remoteReplica : peerReplicas) {
+      // We need to ensure that a replica token gets persisted only after the corresponding data in the
+      // store gets flushed to disk. We use the store flush interval multiplied by a constant factor
+      // to determine the token flush interval
+      FindToken findToken =
+          this.tokenHelper.getFindTokenFactoryFromReplicaType(remoteReplica.getReplicaType()).getNewFindToken();
+      RemoteReplicaInfo remoteReplicaInfo = new RemoteReplicaInfo(remoteReplica, replicaId, store, findToken,
+          storeConfig.storeDataFlushIntervalSeconds * SystemTime.MsPerSec * Replication_Delay_Multiplier,
```
cgtz: `TimeUnit.SECONDS.toMillis(storeConfig.storeDataFlushIntervalSeconds)`

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
cgtz: I don't see any usages of the write lock in this class. Typically a readwritelock is used to allow for concurrent readers during normal scenarios (by using the read lock), but lock the entire datastructure (ensure that all readers finish there reads and make future readers wait by using the write lock) when editing.

ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
```diff
@@ -288,29 +292,30 @@ boolean addBlobStore(ReplicaId replica) {
       logger.error("Failed to add {} because disk manager is not running", replica.getPartitionId());
       return false;
     }
-    boolean succeed = true;
+    boolean succeed = false;
     BlobStore store =
         new BlobStore(replica, storeConfig, scheduler, longLivedTaskScheduler, diskIOScheduler, diskSpaceAllocator,
             storeMainMetrics, storeUnderCompactionMetrics, keyFactory, recovery, hardDelete, replicaStatusDelegate,
             time);
+    rwLock.readLock().lock();
```
cgtz: Same comment for this class.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -381,6 +382,55 @@ public PartitionId getRandomWritablePartition(String partitionClass, List<Partit
...
+      Map<String, String> replicaInfos = partitionToReplicas.get(partitionIdStr);
+      if (replicaInfos != null && replicaInfos.containsKey(instanceName)) {
+        long replicaCapacity = Long.valueOf(replicaInfos.get(REPLICAS_CAPACITY_STR));
+        String partitionClass = replicaInfos.get(PARTITION_CLASS_STR);
+        AmbryPartition mappedPartition = partitionNameToAmbryPartition.get(partitionIdStr);
+        if (mappedPartition == null) {
+          logger.info("Partition {} is currently not present in cluster map, creating a new partition.",
+              partitionIdStr);
+          mappedPartition =
+              new AmbryPartition(Long.valueOf(partitionIdStr), partitionClass, helixClusterManagerCallback);
```
cgtz: intellij suggests using `Long.parseLong` here to avoid unnecessary boxing

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -381,6 +382,55 @@ public PartitionId getRandomWritablePartition(String partitionClass, List<Partit
...
+  public ReplicaId getNewReplica(String partitionIdStr, DataNodeId dataNodeId) {
+    ReplicaId newReplica = null;
+    logger.info("Getting ReplicaAddition ZNRecord from HelixPropertyStore in local DC.");
+    ZNRecord zNRecord = helixPropertyStoreInLocalDc.get(REPLICA_ADDITION_ZNODE_PATH, null, AccessOption.PERSISTENT);
+    if (zNRecord != null) {
+      String instanceName = getInstanceName(dataNodeId.getHostname(), dataNodeId.getPort());
+      Map<String, Map<String, String>> partitionToReplicas = zNRecord.getMapFields();
+      Map<String, String> replicaInfos = partitionToReplicas.get(partitionIdStr);
+      if (replicaInfos != null && replicaInfos.containsKey(instanceName)) {
+        long replicaCapacity = Long.valueOf(replicaInfos.get(REPLICAS_CAPACITY_STR));
```
cgtz: also here

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -381,6 +382,55 @@ public PartitionId getRandomWritablePartition(String partitionClass, List<Partit
...
+        if (mappedPartition == null) {
+          logger.info("Partition {} is currently not present in cluster map, creating a new partition.",
+              partitionIdStr);
+          mappedPartition =
+              new AmbryPartition(Long.valueOf(partitionIdStr), partitionClass, helixClusterManagerCallback);
+        }
+        // Check if data node or disk is in current cluster map, if not, set newReplica to null.
+        AmbryDataNode dataNode = instanceNameToAmbryDataNode.get(instanceName);
+        String mountPathFromHelix = replicaInfos.get(instanceName);
+        Set<AmbryDisk> disks = dataNode != null ? ambryDataNodeToAmbryDisks.get(dataNode) : null;
```
cgtz: Could also do
```java
        Optional<AmbryDisk> potentialDisk =
            Optional.ofNullable(dataNode != null ? ambryDataNodeToAmbryDisks.get(dataNode) : null)
                .flatMap(disks -> disks.stream().filter(d -> d.getMountPath().equals(mountPathFromHelix)).findAny());
```

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +110,111 @@ public void start() throws ReplicationException {
...
+      if (!partitionToPartitionInfo.containsKey(replicaId.getPartitionId())) {
+        logger.error("{} doesn't exist in replication manager, skipping removing replica request.",
+            replicaId.getPartitionId());
+      } else {
+        PartitionInfo partitionInfo = partitionToPartitionInfo.get(replicaId.getPartitionId());
+        List<RemoteReplicaInfo> remoteReplicaInfos = partitionInfo.getRemoteReplicaInfos();
+        logger.info("Removing remote replicas of {} from replica threads", replicaId.getPartitionId());
+        removeRemoteReplicaInfoFromReplicaThread(remoteReplicaInfos);
+        mountPathToPartitionInfos.computeIfPresent(replicaId.getMountPath(), (k, v) -> {
+          v.remove(partitionInfo);
```
cgtz: since you are removing arbitrary elements from `v`, maybe it should be a `HashSet` instead of `ArrayList`

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -121,4 +110,111 @@ public void start() throws ReplicationException {
...
+    boolean succeed = false;
+    rwLock.readLock().lock();
+    try {
+      if (partitionToPartitionInfo.containsKey(replicaId.getPartitionId())) {
+        logger.error("{} already exists in replication manager, rejecting adding replica request.",
+            replicaId.getPartitionId());
+      } else {
+        List<? extends ReplicaId> peerReplicas = replicaId.getPeerReplicaIds();
+        List<RemoteReplicaInfo> remoteReplicaInfos = new ArrayList<>();
+        if (peerReplicas != null) {
```
cgtz: what is the case where `peerReplicas` is null?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: right, will change this to simple `ReentrantLock`.

ambry-api/src/main/java/com.github.ambry/clustermap/ClusterMap.java
```diff
@@ -122,6 +122,14 @@
    */
   JSONObject getSnapshot();
 
+  /**
+   * Attempt to get new replica of certain partition that resides on given data node.
```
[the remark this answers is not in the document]
jsjtzyy: This is actually "new replica" that will be added to specified datanode. Currently, it is used in HelixClusterManager only because target node needs replica detailed info to create it. The replica info is stored in Helix PropertyStore. Before moving replica, we use Helix Bootstrap tool to upload new replica infos onto Helix PropertyStore, thus target node can fetch this info to create new replica. Let me know if you come up with a better name to disambiguate in this context. Thanks

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
jsjtzyy: sure, fair point.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
jsjtzyy: updated the exception message. Feel like `return null` is unneeded.

ambry-protocol/src/main/java/com.github.ambry.protocol/BlobStoreControlRequestType.java
jsjtzyy: makes more sense, will change as you suggested.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
jsjtzyy: fixed

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
jsjtzyy: fixed

ambry-protocol/src/main/java/com.github.ambry.protocol/BlobStoreControlAdminRequest.java
jsjtzyy: I feel like maybe we can still use V1 for this request. Current change doesn't really modify serialize/deserialize logic of `BlobStoreControlAdminRequest`, it interprets `stream.readByte()` as a new introduced enum type instead of boolean value. I think keeping using V1 makes sense here (also, we can keep checking version number in this method).

cgtz: Yep, I agree that in this circumstance it makes sense to keep using v1

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
cgtz: With just a simple mutex, wouldn't you have to synchronize all reads/writes to these data structures? Would it be better to take a read lock when reading from the shared data structures and taking a write lock when editing them? We can discuss in person.

PR #1261, opened 2019-09-13
AccountService's RouterStore wait for cross colo replication before reading blob from server.

justinlin-linkedin:
When there is an update to account metadata, a new blob will be created and the blob id will be broadcasted to all the frontends across all the colos in the same fabric group. Since blob id is broadcasted through zookeeper and zookeeper is much faster than ambry server's replication, most of the time, ambry-frontend would read the new blob from remote colo, which results in lots of cross-colo read.

This PR is trying to mitigate this problem. It will wait for a while before reading the new blob in ambry-frontend. The wait duration should equal to 95 percentile of cross colo replication. In this way, most of the time, ambry-frontend would read the new blob from local colo.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
@@ -13,8 +13,10 @@
  */
 package com.github.ambry.account;
 
+import com.github.ambry.commons.BlobId;
 import com.github.ambry.commons.ByteBufferReadableStreamChannel;
 import com.github.ambry.commons.ReadableStreamChannelInputStream;
+import com.github.ambry.config.HelixAccountServiceConfig;
 import com.github.ambry.messageformat.BlobProperties;
 import com.github.ambry.router.GetBlobOptionsBuilder;
 import com.github.ambry.router.GetBlobResult;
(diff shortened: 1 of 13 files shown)


ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -109,6 +114,19 @@
           .max(Comparator.comparing(BlobIDAndVersion::getVersion))
           .get();
 
+      // We don't want to wait for cross colo replication when this function is not called from listener
+      // since the blob data is probably already replicated to the local colo.
+      if (isCalledFromListener && waitTimeForCrossColoReplicationMs > 0 && router.get().getClusterMap() != null) {
+        try {
+          BlobId blobID = new BlobId(blobIDAndVersion.getBlobID(), router.get().getClusterMap());
+          if (blobID.getDatacenterId() != router.get().getClusterMap().getLocalDatacenterId()) {
```
jsjtzyy: Looks like there is no test that covers this piece of code.

jsjtzyy:
Let's hold off (or close) this PR temporarily since we need to address multi-cluster issue first.

PR #1264, opened 2019-09-23
Ignore liveness update of current node if it is actually alive

jsjtzyy:
When starting ambry-server, HelixClusterManager is instantiated before
participating into cluster. During this short window, Helix may send
notification that makes this server mark itself is down. This brings
Disk_Unavailable errors when handling replication requests and checking
the state of disk(disk state depends on node state). Such error is
misleading because disk is actually good and the server is able to serve
replication request. Hence, this PR makes node ignore liveness update of
itself if it is actually alive. (It should be safe, because other
frontends and servers will mark it down and no subsequent requests will
be routed to it)

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
@@ -61,7 +61,7 @@ import static com.github.ambry.clustermap.ClusterMapUtils.*;
  * @see <a href="http://helix.apache.org">http://helix.apache.org</a>
  */
 class HelixClusterManager implements ClusterMap {
-  private final Logger logger = LoggerFactory.getLogger(getClass());
+  private static final Logger logger = LoggerFactory.getLogger(HelixClusterManager.class);
   private final String clusterName;
   private final String selfInstanceName;
   private final MetricRegistry metricRegistry;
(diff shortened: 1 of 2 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -590,7 +590,13 @@ private void updateInstanceLiveness(List<LiveInstance> liveInstances) {
           liveInstancesSet.add(liveInstance.getInstanceName());
         }
         for (String instanceName : allInstances) {
-          if (liveInstancesSet.contains(instanceName)) {
+          // when starting ambry-server, HelixClusterManager is instantiated before participating into cluster. During
+          // this short window, Helix may send notification that makes this server mark itself is down. This brings
+          // Disk_Unavailable errors when handling replication requests and checking the state of disk (disk state depends
+          // on node state). The Disk_Unavailable is misleading, because disk is actually good and the server is able to
+          // serve replication request. Hence, if instance name equals self instance name, cluster manager of this node
+          // ignores its own liveness notification from Helix to avoid incorrect Disk_Unavailable error.
```
lightningrob: I don't think we need this much explanation of what we saw without the fix.  It's enough to say that the list of live instances doesn't include this node since it hasn't joined yet.

PR #1266, opened 2019-09-25
Add a new version in the ReplicaTokenSerde class

ankagrawal:
Add a new version in the ReplicaTokenSerde to account for persisting replicatype.
Write tests for DiskTokenPersistor.
Note that this fixes a bug in commit 9487e8be3e7db9da2d994aa70f273b739665235f related to deserializing existing replica token files with the new replca token serialization format.


ambry-server/src/integration-test/java/com.github.ambry.server/ServerTestUtil.java
```diff
@@ -2003,7 +2003,7 @@ private static void checkReplicaTokens(MockClusterMap clusterMap, DataNodeId dat
           DataInputStream dataInputStream = new DataInputStream(crcStream);
           try {
             short version = dataInputStream.readShort();
-            assertEquals(0, version);
+            assertEquals(1, version);
 
             System.out.println("setToCheck" + setToCheck.size());
```
jsjtzyy: remove this?

ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
```diff
@@ -27,7 +27,7 @@
   /**
    * The factory class the replicatio uses to create cloud token
```
jsjtzyy: minor: typo `replication`

ambry-server/src/integration-test/java/com.github.ambry.server/ServerTestUtil.java
ankagrawal: this was from before, but fixed nonetheless!

ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
ankagrawal: fixed!

PR #1267, opened 2019-09-26
Fix a bug to make failure on reading blob fails account service update

justinlin-linkedin:
When reading blob fails, the account service update should also fail. This pr would fix this bug.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
@@ -237,6 +237,10 @@ class RouterStore extends AccountMetadataStore {
           // If this is not for backfill, then just read account metadata from blob, otherwise, initialize it with
           // an empty map and fill it up with the accountMap passed to constructor.
           accountMap = (!forBackFill) ? readAccountMetadataFromBlobID(blobIDAndVersion.blobID) : new HashMap<>();
+          if (accountMap == null) {
+            logAndThrowIllegalStateException("Fail to read account metadata from blob id " + blobIDAndVersion.blobID,
+                null);
+          }
         } catch (JSONException e) {
           accountServiceMetrics.remoteDataCorruptionErrorCount.inc();
           logAndThrowIllegalStateException(
--- a/ambry-account/src/test/java/com/github/ambry/account/RouterStoreTest.java
+++ b/ambry-account/src/test/java/com/github/ambry/account/RouterStoreTest.java
@@ -161,8 +161,7 @@ public class RouterStoreTest {
     for (Map.Entry<Short, Account> entry : anotherIdToRefAccountMap.entrySet()) {
       idToRefAccountMap.put(entry.getKey(), entry.getValue());
     }
-    boolean succeeded = store.updateAccounts(anotherIdToRefAccountMap.values());
-    assertFalse("Update account should fail on read error", succeeded);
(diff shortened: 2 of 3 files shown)


ambry-account/src/test/java/com/github/ambry/account/RouterStoreTest.java
```diff
@@ -137,6 +135,36 @@ public void testUpdateAndFetch() throws Exception {
...
+    // Now clear the router
+    router.close();
+
+    // generate another new account and test update and fetch on this account
+    Map<Short, Account> anotherIdToRefAccountMap = new HashMap<>();
+    AccountTestUtils.generateRefAccounts(anotherIdToRefAccountMap, idtoRefContainerMap, accountIDSet, 1, 1);
+    for (Map.Entry<Short, Account> entry : anotherIdToRefAccountMap.entrySet()) {
+      idToRefAccountMap.put(entry.getKey(), entry.getValue());
+    }
+    boolean succeeded = store.updateAccounts(anotherIdToRefAccountMap.values());
```
jsjtzyy: It seems no need to use `succeeded` boolean variable here. Merge line164 and 165

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
```diff
@@ -237,6 +237,10 @@ public ZNRecord update(ZNRecord znRecord) {
           // If this is not for backfill, then just read account metadata from blob, otherwise, initialize it with
           // an empty map and fill it up with the accountMap passed to constructor.
           accountMap = (!forBackFill) ? readAccountMetadataFromBlobID(blobIDAndVersion.blobID) : new HashMap<>();
+          if (accountMap == null) {
+            logAndThrowIllegalStateException("Fail to read account metadata from blob id " + blobIDAndVersion.blobID,
```
jsjtzyy: why not do this in `readAccountMetadataFromBlobID()` ?

ambry-account/src/test/java/com/github/ambry/account/RouterStoreTest.java
```diff
@@ -137,6 +135,35 @@ public void testUpdateAndFetch() throws Exception {
...
+        new RouterStore(accountServiceMetrics, backup, helixStore, new AtomicReference<>(router), forBackfill,
+            TOTAL_NUMBER_OF_VERSION_TO_KEEP);
+    Map<Short, Account> idToRefAccountMap = new HashMap<>();
+    Map<Short, Map<Short, Container>> idtoRefContainerMap = new HashMap<>();
+    Set<Short> accountIDSet = new HashSet<>();
+    // generate an new account and test update and fetch on this account
+    AccountTestUtils.generateRefAccounts(idToRefAccountMap, idtoRefContainerMap, accountIDSet, 1, 1);
+    assertUpdateAndFetch(store, idToRefAccountMap, idToRefAccountMap, 1, 1);
+
+    // Now clear the router
```
jsjtzyy: minor: Now close the router to trigger get blob failure

ambry-account/src/main/java/com/github/ambry/account/RouterStore.java
justinlin-linkedin: I suppose we could, the readAccountMetadataFromBlobID function is also called by the fetchAccountMetadataFromZNRecord function and that function should not throw any exception. So in the readAccountMetadataFromBlobID, I catch the exception and return null.

PR #1268, opened 2019-09-26
Add a cli tool to perform some account metadata related operations

justinlin-linkedin:


ambry-tools/src/main/java/com.github.ambry/account/AccountTool.java
```diff
@@ -0,0 +1,526 @@
...
+   * @return The {@link Router}.
+   * @throws Exception Any unexpected exception.
+   */
+  private Router getRouter() throws Exception {
+    // Create a HelixAccountService for the router.
+    AccountServiceFactory accountServiceFactory =
+        Utils.getObj("com.github.ambry.account.HelixAccountServiceFactory", verifiableProperties, registry);
+    accountService = accountServiceFactory.getAccountService();
+
+    notifier = ((HelixAccountService) accountService).getNotifier();
```
cgtz: Instead of duplicating the notification logic in this tool, could we us HelixAccountService.updateAccounts to implement the updateAccounts method?

ambry-tools/src/main/java/com.github.ambry/account/AccountTool.java
```diff
@@ -0,0 +1,526 @@
...
+import org.apache.helix.ZNRecord;
+import org.apache.helix.store.HelixPropertyStore;
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+
+/**
+ * <p>
+ *   This is a command-line tool can be used to perform a number of {@link Account} related operations.
```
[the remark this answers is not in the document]
justinlin-linkedin: My plan to push the new account metadata management to prod and then just remove the AccoutUpdateTool.

ambry-tools/src/main/java/com.github.ambry/account/AccountTool.java
justinlin-linkedin: The problem is that when using HelixAccountService.upateAccounts, it would merge the accounts in the argument with the previous accounts. But I want to totally overwrite the accounts. I should probably verify that there is no conflict even if I want to overwrite the accounts. I will update accordingly.

ambry-tools/src/main/java/com.github.ambry/account/AccountTool.java
```diff
@@ -0,0 +1,542 @@
...
+  private static final int ZK_CLIENT_CONNECTION_TIMEOUT_MS = 5000;
+  private static final int ZK_CLIENT_SESSION_TIMEOUT_MS = 20000;
+  private static final String DEFAULT_HOSTNAME = "localhost";
+
+  private String clusterName = null;
+  private String hostname = null;
+  private String dcName = null;
+  private String clusterMapDcsZkConnectString = null;
+
+  private static final List<String> ACTION_CHOICES = Arrays.asList("list", "view", "rollback", "update");
```
jsjtzyy: Can we change this to Enum? Refer to `DumpIndexTool.java` and i think enum `Operation` there is an example.

ambry-tools/src/main/java/com.github.ambry/account/AccountTool.java
```diff
@@ -0,0 +1,542 @@
...
+   */
+  private Router getRouter() throws Exception {
+    // Create a HelixAccountService for the router.
+    AccountServiceFactory accountServiceFactory =
+        Utils.getObj("com.github.ambry.account.HelixAccountServiceFactory", verifiableProperties, registry);
+    accountService = (HelixAccountService) accountServiceFactory.getAccountService();
+
+    ClusterMapConfig clusterMapConfig = new ClusterMapConfig(verifiableProperties);
+    ClusterAgentsFactory clusterAgentsFactory =
+        Utils.getObj(clusterMapConfig.clusterMapClusterAgentsFactory, clusterMapConfig, "", "");
```
jsjtzyy: I suggest passing PartitionLayout and HardwareLayout as well. I know it is not needed for HelixClusterManager but it's better to support update account when we are using StaticClusterManager.

justinlin-linkedin: Added hardware layout and partition layout as optional parameter.

PR #1273, opened 2019-10-04
Revert "Shutdown request handler before network server in ambry server (#1206)"

cgtz:
This reverts commit 063ca5ee6a91536a57352ad65f4484c7f19cb7e0.
We have some suspicions that the frontend may not be able to deal with
this nicely if it is in the middle of writing to a socket.


zzmao:
Will merge after travis test.

PR #1274, opened 2019-10-04
Fix TransmissionSendAllTime and TransmissionReceiveAllTime names error.

zzmao:
TransmissionSendAllTime and TransmissionReceiveAllTime were registered incorrectly.

### DIFF
--- a/ambry-network/src/main/java/com.github.ambry.network/NetworkMetrics.java
+++ b/ambry-network/src/main/java/com.github.ambry.network/NetworkMetrics.java
@@ -127,9 +127,9 @@ public class NetworkMetrics {
 
     transmissionSendPendingTime =
         registry.histogram(MetricRegistry.name(Selector.class, "TransmissionSendPendingTime"));
-    transmissionSendAllTime = registry.histogram(MetricRegistry.name(Selector.class, "TransmissionSendTime"));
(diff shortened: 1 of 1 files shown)


ambry-network/src/main/java/com.github.ambry.network/NetworkMetrics.java
```diff
@@ -127,9 +127,9 @@ public NetworkMetrics(MetricRegistry registry) {
 
     transmissionSendPendingTime =
         registry.histogram(MetricRegistry.name(Selector.class, "TransmissionSendPendingTime"));
-    transmissionSendAllTime = registry.histogram(MetricRegistry.name(Selector.class, "TransmissionSendTime"));
+    transmissionSendAllTime = registry.histogram(MetricRegistry.name(Selector.class, "TransmissionSendAllTime"));
```
ankagrawal: With this change metrics transmissionsendtime and transmissionsendalltime both seem to have same name now. transmissionsendtime one is set on every write and transmissionsendalltime is set on completion of send. Why do you want to make this change?

ankagrawal: Never mind.. I see the issue now. I interpreted it backwards.

zzmao: TransmissionSendAllTime should be used. And the change is to use `TransmissionSendAllTime `.

PR #1275, opened 2019-10-07
Vcrrequest init error fix

ankagrawal:
This PR fixes a bug that was encountered while testing recovery. Vcr was trying to get local replicas map using the "getReplicaIds" API in HelixClusterManager. But this API has a check for being called only in a AmbryDataNode. 
As a fix this PR overrides the get local replicas functionality in VcrRequests. Also Vcr nodes should be able to handle requests for all partitions. The overridden get local replicas functionality exhibits this behavior too.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageManager.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageManager.java
@@ -50,9 +50,8 @@ public class CloudStorageManager implements StoreManager {
 
   @Override
   public boolean addBlobStore(ReplicaId replica) {
-    partitionToStore.computeIfAbsent(replica.getPartitionId(),
-        store -> new CloudBlobStore(properties, replica.getPartitionId(), cloudDestination, clusterMap, vcrMetrics));
-    return startBlobStore(replica.getPartitionId());
+    createAndStartBlobStoreIfAbsent(replica.getPartitionId());
+    return partitionToStore.get(replica.getPartitionId()) != null;
   }
 
   @Override
@@ -67,8 +66,8 @@ public class CloudStorageManager implements StoreManager {
 
   @Override
   public Store getStore(PartitionId id) {
-    Store store = partitionToStore.computeIfAbsent(id, store -> new CloudBlobStore(properties, id, cloudDestination, clusterMap, vcrMetrics));
-    return (store != null && store.isStarted()) ? store : null;
+    createAndStartBlobStoreIfAbsent(id);
+    return partitionToStore.get(id);
   }
 
   @Override
@@ -113,5 +112,18 @@ public class CloudStorageManager implements StoreManager {
      * the local partition status for a vcr node should always be available.
      */
     return ServerErrorCode.No_Error;
+  }
+
+  private void createAndStartBlobStoreIfAbsent(PartitionId partitionId) {
+    partitionToStore.computeIfAbsent(partitionId, value -> {
+      CloudBlobStore store = null;
+      try {
+        store = new CloudBlobStore(properties, partitionId, cloudDestination, clusterMap, vcrMetrics);
+        store.start();
+      } catch (StoreException ex) {
+        logger.error("Can't start CloudStore for partition {}" + partitionId, ex);
+      }
+      return store;
+    });
   }
 }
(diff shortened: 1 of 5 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/VcrRequests.java
```diff
@@ -74,14 +78,22 @@ protected ServerErrorCode validateRequest(PartitionId partition, RequestOrRespon
...
+  @Override
+  protected Map<PartitionId, ReplicaId> createLocalPartitionToReplicaMap() {
+    //Vcr should be able to handle requests for all partitions
+    List<? extends PartitionId> partitionIds = clusterMap.getAllPartitionIds(null);
+    Map<PartitionId, ReplicaId> partitionToReplica = new HashMap<>();
+    for (PartitionId partitionId : partitionIds) {
+      partitionToReplica.put(partitionId, new CloudReplica(partitionId, currentNode));
+    }
+    return partitionToReplica;
+  }
```
lightningrob: Can simplify this to:
`return partitionIds.stream().collect(Collectors.toMap(Function.identity(),
    partitionId -> new CloudReplica(partitionId, currentNode)));`

ambry-server/src/main/java/com.github.ambry.server/AmbryRequests.java
```diff
@@ -185,6 +181,19 @@ public void handleRequests(Request request) throws InterruptedException {
...
+   * Get the list of replicas on current node.
+   * @return list of {@link ReplicaId}s on current node.
+   */
+  protected Map<PartitionId, ReplicaId> createLocalPartitionToReplicaMap() {
+    Map<PartitionId, ReplicaId> partitionToReplica = new HashMap<>();
+    List<? extends ReplicaId> localReplicaIds = clusterMap.getReplicaIds(currentNode);
+    for (ReplicaId replicaId : localReplicaIds) {
+      partitionToReplica.put(replicaId.getPartitionId(), replicaId);
+    }
+    return partitionToReplica;
```
lightningrob: Same here.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageManager.java
```diff
@@ -109,20 +108,22 @@ public boolean removeBlobStore(PartitionId id) {
...
-    }
     return ServerErrorCode.No_Error;
   }
+
+  private void createAndStartBlobStoreIfAbsent(PartitionId partitionId) {
+    partitionToStore.computeIfAbsent(partitionId, value -> {
+      CloudBlobStore store = null;
+      try {
+        store = new CloudBlobStore(properties, partitionId, cloudDestination, clusterMap, vcrMetrics);
+        store.start();
```
lightningrob: We should avoid side effects like starting a store inside a computeIfAbsent lambda.

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrRequests.java
ankagrawal: fixed.

ambry-server/src/main/java/com.github.ambry.server/AmbryRequests.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageManager.java
ankagrawal: fixed.

zzmao: Add a log after `CloudBlobStore  created successfully?

zzmao:
One minor comment. 

LGTM.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageManager.java
```diff
@@ -67,8 +65,8 @@ public boolean shutdownBlobStore(PartitionId id) {
 
   @Override
   public Store getStore(PartitionId id) {
-    Store store = partitionToStore.get(id);
-    return (store != null && store.isStarted()) ? store : null;
+    createAndStartBlobStoreIfAbsent(id);
```
cgtz: As discussed offline, let's consider the calls to shutdownBlobStore and removeStore in VCR replication

ankagrawal: As a fix i have made the CloudBlobStore operations thread safe. Both removestore and shutdownstore are there as is in the add/remove replica flow. But a getstore will return null, if the store is not started.

PR #1277, opened 2019-10-09
Put (uploadBlob) in AzureCloudDestination should autocreate container if not exists

ankagrawal:


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -234,7 +234,7 @@ private boolean uploadIfNotExists(BlobId blobId, long inputLength, CloudBlobMeta
     azureMetrics.blobUploadRequestCount.inc();
     Timer.Context storageTimer = azureMetrics.blobUploadTime.time();
     try {
-      CloudBlockBlob azureBlob = getAzureBlobReference(blobId);
+      CloudBlockBlob azureBlob = getAzureBlobReference(blobId, true);
```
jsjtzyy: better to add piece of comment for this method when container doesn't exist.

ankagrawal: There are comments in lines 526 and 529, when we actually create the container. Let me know if you think we need more logging.

PR #1278, opened 2019-10-11
Refactor request classes from server to protocol module 

lightningrob:
(Preparation for cloud router)
Note: AmbryRequests has been refactored into two classes:
1. com.github.ambry.protocol.AmbryRequests: frontend and replication methods
2. com.github.ambry.server.AmbryServerRequests: adds admin methods


ambry-protocol/src/main/java/com.github.ambry.protocol/RequestHandler.java
```diff
@@ -25,10 +26,10 @@
 public class RequestHandler implements Runnable {
   private final int id;
   private final RequestResponseChannel requestChannel;
-  private final AmbryRequests requests;
+  private final RequestAPI requests;
   private Logger logger = LoggerFactory.getLogger(getClass());
```
jsjtzyy: change this logger to `static final`？

ambry-protocol/src/main/java/com.github.ambry.protocol/AmbryRequests.java
```diff
@@ -1206,85 +732,20 @@ protected void sendGetResponse(RequestResponseChannel requestResponseChannel, Ge
    */
```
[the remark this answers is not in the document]
lightningrob: Added "if any" here.

ambry-protocol/src/main/java/com.github.ambry.protocol/AmbryRequests.java
```diff
@@ -577,7 +528,11 @@ public void handleTtlUpdateRequest(Request request) throws IOException, Interrup
             metrics.updateBlobTtlTotalTimeInMs, null, null, totalTimeSpent));
   }
 
+  @Override
   public void handleReplicaMetadataRequest(Request request) throws IOException, InterruptedException {
+    if (replicationEngine == null) {
+      throw new IllegalStateException("Replication not supported on this node.");
```
cgtz: UnsupportedOperationException?

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrRequests.java
```diff
@@ -86,12 +81,4 @@ protected ServerErrorCode validateRequest(PartitionId partition, RequestOrRespon
   protected long getRemoteReplicaLag(Store store, long totalBytesRead) {
     return -1;
   }
-
-  @Override
-  protected Map<PartitionId, ReplicaId> createLocalPartitionToReplicaMap() {
```
cgtz: Where did this code go? The implementation defined in AmbryRequests looks like it only gets partitions on the local node, whereas this gets all partitions.

Also, shouldn't this extend AmbryServerRequests if it previously extended the old AmbryRequests (to get the request deserialization code)?

lightningrob: I removed this method because localPartitionToReplicaMap is only used for disk-based partitions, e.g. checking if a disk is bad or partition is not on this node.  I don't think the VCR needs this.  @ankagrawal or @zzmao can you weigh in here?

ambry-protocol/src/main/java/com.github.ambry.protocol/AmbryRequests.java
lightningrob: Sure.

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrRequests.java
cgtz: Ah, I see the change you made in AmbryRequests now to remove the check that used to be in `validateRequests`. It seems like the admin request support has been removed from VcrRequests as well (by having this not extend AmbryServerRequests). I think that is ok since most of those operations were not fully supported anyways (although some of them would be nice to haves in the future).

ambry-protocol/src/main/java/com.github.ambry.protocol/AmbryRequests.java
```diff
@@ -167,15 +125,14 @@ public void handleRequests(Request request) throws InterruptedException {
...
         case ReplicaMetadataRequest:
           handleReplicaMetadataRequest(request);
           break;
         case AdminRequest:
-          handleAdminRequest(request);
-          break;
-        case TtlUpdateRequest:
-          handleTtlUpdateRequest(request);
-          break;
+          throw new UnsupportedOperationException("Request type not supported");
```
ankagrawal: Why do we need a separate case for AdminRequest here. We can just let it goto default.

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
```diff
@@ -0,0 +1,653 @@
...
+    super(storeManager, requestResponseChannel, clusterMap, nodeId, registry, serverMetrics, findTokenHelper,
+        operationNotification, replicationEngine, storeKeyFactory, enableDataPrefetch, storeKeyConverterFactory);
+    this.statsManager = statsManager;
+
+    for (RequestOrResponseType requestType : EnumSet.of(RequestOrResponseType.PutRequest,
+        RequestOrResponseType.GetRequest, RequestOrResponseType.DeleteRequest,
+        RequestOrResponseType.ReplicaMetadataRequest, RequestOrResponseType.TtlUpdateRequest)) {
+      requestsDisableInfo.put(requestType, Collections.newSetFromMap(new ConcurrentHashMap<>()));
+    }
+    localPartitionToReplicaMap = createLocalPartitionToReplicaMap();
```
ankagrawal: Just curious.. Why did we remove Collections.unmodifiableMap from here? Seemed like a good practice to have it.

ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
```diff
@@ -21,6 +21,14 @@
...
+  /**
+   * Handle a request.
+   * @param request The request to handle.
+   * @throws InterruptedException
+   */
+  void handleRequests(Request request) throws InterruptedException;
+
   /**
    * Puts a blob into the store. It accepts a blob property, user metadata and the blob
    * as a stream and stores them
```
cgtz: If you want, you could actually use default methods in this interface to throw the UnsupportedOperationExceptions.

For example: 
```
default void handlePutRequest(Request) throws ... {
  throw new UnsupportedOperationException("Put request not supported");
}
```
You could also consider adding handleAdminRequest to this interface. Then AmbryServerRequests would not have to override the handleRequests of AmbryRequests.

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
jsjtzyy: I made the change in my previous PR. The intention is to support dynamically adding new replica to server node which requires `localPartitionToReplicaMap` to be mutable. I will take a second look if I misunderstood semantic of `unmodifiableMap` here.

ankagrawal: Right I get it now. With your changes the number of local replicas will change, hence we it shouldn't be a unmodifiable map anymore. Makes sense!

ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
lightningrob: Added admin handler as default method.  Left others the same.

ambry-cloud/src/main/java/com.github.ambry.cloud/VcrRequests.java
zzmao: The logic looks good.

ambry-protocol/src/main/java/com.github.ambry.protocol/AmbryRequests.java
```diff
@@ -167,15 +125,14 @@ public void handleRequests(Request request) throws InterruptedException {
...
           handleDeleteRequest(request);
           break;
+        case TtlUpdateRequest:
+          handleTtlUpdateRequest(request);
+          break;
         case ReplicaMetadataRequest:
           handleReplicaMetadataRequest(request);
           break;
         case AdminRequest:
           handleAdminRequest(request);
```
ankagrawal: are we missing a break statement here for AdminRequest case?

lightningrob: good catch, added.

PR #1280, opened 2019-10-14
Allow index to populate bloom filter with UUID only

jsjtzyy:
Instead of using whole blob id (including accountId, containerId,
partitionId etc) to populate bloom filter, this PR allows index to
generate bloom filter based on blobId's UUID only. Thus, bloom filter
logic becomes consistent with blobId comparison which uses UUID only to
distinguish blobs. Also, it helps to cleanup some A1 Deletes(if any)
that associated with A2 Puts.

### DIFF
--- a/ambry-commons/src/main/java/com.github.ambry.commons/BlobId.java
+++ b/ambry-commons/src/main/java/com.github.ambry.commons/BlobId.java
@@ -195,8 +195,8 @@ public class BlobId extends StoreKey {
    * @param blobDataType The blob data type.
    * @param uuidStr The uuid that is to be used to construct this id.
    */
-  private BlobId(short version, BlobIdType type, byte datacenterId, short accountId, short containerId,
-      PartitionId partitionId, boolean isEncrypted, BlobDataType blobDataType, String uuidStr) {
+  BlobId(short version, BlobIdType type, byte datacenterId, short accountId, short containerId, PartitionId partitionId,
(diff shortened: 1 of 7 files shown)


justinlin-linkedin:
what is your roll out plan for this logic? To delete all the bloomfilter file and rebuild them when ambry server starts?

ambry-commons/src/main/java/com.github.ambry.commons/BlobId.java
```diff
@@ -492,6 +492,30 @@ public String getUuid() {
...
+    ByteBuffer uuidBuf;
+    switch (version) {
+      case BLOB_ID_V1:
+      case BLOB_ID_V2:
+      case BLOB_ID_V3:
+      case BLOB_ID_V4:
+      case BLOB_ID_V5:
+        byte[] uuidBytes = getUuid().getBytes();
+        uuidBuf = ByteBuffer.allocate(UUID_SIZE_FIELD_LENGTH_IN_BYTES + (short) uuidBytes.length);
+        uuidBuf.putInt(uuidBytes.length);
```
cgtz: Is the length prefix needed in this case or could you return `uuidBytes` directly? Since there is only one item (the UUID) in the array.

jsjtzyy: Fair point. Length prefix is removed.

PR #1281, opened 2019-10-15
Share memory between blobdata and networking buffer

justinlin-linkedin:
For GetResponse, buffer at networking layer holds all the blob content and we are not sharing the buffer with the BlobData in the router, this would roughly double the memory footprint for GetResponse.

This PR is trying to share the memory between these two object and reduce the memory usage.

### DIFF
--- a/ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
+++ b/ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
@@ -13,9 +13,9 @@
  */
 package com.github.ambry.messageformat;
 
-import com.github.ambry.commons.CommonUtils;
 import com.github.ambry.store.StoreKey;
 import com.github.ambry.store.StoreKeyFactory;
+import com.github.ambry.utils.ByteBufferDataInputStream;
 import com.github.ambry.utils.ByteBufferInputStream;
 import com.github.ambry.utils.Crc32;
 import com.github.ambry.utils.CrcInputStream;
@@ -1639,7 +1639,7 @@ public class MessageFormatRecord {
 
   /**
    * Creating a {@link ByteBufferInputStream} from the {@link CrcInputStream} by sharing the underlying memory if the
-   * crcStream is built upon a {@link com.github.ambry.commons.CommonUtils.ByteBufferDataInputStream}.
+   * crcStream is built upon a {@link ByteBufferDataInputStream}.
    * @param crcStream The crcStream to read {@link ByteBuffer} out.
    * @param dataSize The size of {@link ByteBuffer}.
    * @return The {@link ByteBufferInputStream}
@@ -1647,30 +1647,23 @@ public class MessageFormatRecord {
    */
   static ByteBufferInputStream getByteBufferInputStreamForBlobRecord(CrcInputStream crcStream, int dataSize)
       throws IOException {
-    ByteBufferInputStream output = null;
+    ByteBufferInputStream output;
     InputStream inputStream = crcStream.getUnderlyingInputStream();
-    if (inputStream instanceof CommonUtils.ByteBufferDataInputStream) {
-      ByteBuffer byteBuffer = ((CommonUtils.ByteBufferDataInputStream) inputStream).getBuffer();
-      int startPosition = byteBuffer.position();
-      // exhaust dataSize bytes and update crc for it.
-      byte[] dummy = new byte[4];
-      int read = 0;
-      while (read < (int) dataSize) {
-        int toRead = Math.min(4, (int) dataSize - read);
-        int n = crcStream.read(dummy, 0, toRead);
-        read += n;
-      }
+    if (inputStream instanceof ByteBufferDataInputStream) {
+      ByteBuffer byteBuffer = ((ByteBufferDataInputStream) inputStream).getBuffer();
+      int startIndex = byteBuffer.position();
+      int oldLimit = byteBuffer.limit();
 
-      int endPosition = byteBuffer.position();
-      int endLimit = byteBuffer.limit();
-      byteBuffer.position(startPosition);
-      byteBuffer.limit(endPosition);
-      output = new ByteBufferInputStream(byteBuffer.slice());
+      byteBuffer.limit(startIndex + dataSize);
+      ByteBuffer dataBuffer = byteBuffer.slice();
+      crcStream.update(dataBuffer.duplicate());
+
+      output = new ByteBufferInputStream(dataBuffer);
       // resume the byteBuffer for future use.
-      byteBuffer.position(endPosition);
-      byteBuffer.limit(endLimit);
+      byteBuffer.position(startIndex+dataSize);
+      byteBuffer.limit(oldLimit);
     } else {
-      output = new ByteBufferInputStream(crcStream, (int) dataSize);
+      output = new ByteBufferInputStream(crcStream, dataSize);
     }
     return output;
   }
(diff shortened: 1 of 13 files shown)


ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1634,6 +1637,44 @@ public static ByteBuffer deserializeUserMetadataRecord(CrcInputStream crcStream)
...
+   * Creating a {@link ByteBufferInputStream} from the {@link CrcInputStream} by sharing the underlying memory if the
+   * crcStream is built upon a {@link com.github.ambry.commons.CommonUtils.ByteBufferDataInputStream}.
+   * @param crcStream The crcStream to read {@link ByteBuffer} out.
+   * @param dataSize The size of {@link ByteBuffer}.
+   * @return The {@link ByteBufferInputStream}
+   * @throws IOException Unexpected IO errors.
+   */
+  static ByteBufferInputStream getByteBufferInputStreamForBlobRecord(CrcInputStream crcStream, int dataSize)
+      throws IOException {
+    ByteBufferInputStream output = null;
```
cgtz: nit remove `= null`

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1634,6 +1637,44 @@ public static ByteBuffer deserializeUserMetadataRecord(CrcInputStream crcStream)
...
+      int endPosition = byteBuffer.position();
+      int endLimit = byteBuffer.limit();
+      byteBuffer.position(startPosition);
+      byteBuffer.limit(endPosition);
+      output = new ByteBufferInputStream(byteBuffer.slice());
+      // resume the byteBuffer for future use.
+      byteBuffer.position(endPosition);
+      byteBuffer.limit(endLimit);
+    } else {
+      output = new ByteBufferInputStream(crcStream, (int) dataSize);
```
cgtz: it also seems that the various casts to int are not needed since dataSize is already an int.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1634,6 +1637,44 @@ public static ByteBuffer deserializeUserMetadataRecord(CrcInputStream crcStream)
...
+   */
+  static ByteBufferInputStream getByteBufferInputStreamForBlobRecord(CrcInputStream crcStream, int dataSize)
+      throws IOException {
+    ByteBufferInputStream output = null;
+    InputStream inputStream = crcStream.getUnderlyingInputStream();
+    if (inputStream instanceof CommonUtils.ByteBufferDataInputStream) {
+      ByteBuffer byteBuffer = ((CommonUtils.ByteBufferDataInputStream) inputStream).getBuffer();
+      int startPosition = byteBuffer.position();
+      // exhaust dataSize bytes and update crc for it.
+      byte[] dummy = new byte[4];
```
cgtz: 4 is a pretty small size and _might_ impact crc calculation speed. I wonder if increasing it to something like 512 (still small compared to blob size) would help.

Another thing that we could look into would be introducing a method into Crc32 that can operate directly on bytebuffers without a copy. Java's CRC32 class has such a method.

cgtz: The latter appears doable with small changes to the existing code that deals with byte arrays:
```
  public void update(ByteBuffer b) {
    int localCrc = crc;

    while (b.remaining() > 7) {
      final int c0 = (b.get() ^ localCrc) & 0xff;
      final int c1 = (b.get() ^ (localCrc >>>= 8)) & 0xff;
      final int c2 = (b.get() ^ (localCrc >>>= 8)) & 0xff;
      final int c3 = (b.get() ^ (localCrc >>>= 8)) & 0xff;
      localCrc = (T[T8_7_start + c0] ^ T[T8_6_start + c1]) ^ (T[T8_5_start + c2] ^ T[T8_4_start + c3]);

      final int c4 = b.get() & 0xff;
      final int c5 = b.get() & 0xff;
      final int c6 = b.get() & 0xff;
      final int c7 = b.get() & 0xff;

      localCrc ^= (T[T8_3_start + c4] ^ T[T8_2_start + c5]) ^ (T[T8_1_start + c6] ^ T[T8_0_start + c7]);
    }

    /* loop unroll - duff's device style */
    switch (b.remaining()) {
      case 7:
        localCrc = (localCrc >>> 8) ^ T[T8_0_start + ((localCrc ^ b.get()) & 0xff)];
      case 6:
        localCrc = (localCrc >>> 8) ^ T[T8_0_start + ((localCrc ^ b.get()) & 0xff)];
      case 5:
        localCrc = (localCrc >>> 8) ^ T[T8_0_start + ((localCrc ^ b.get()) & 0xff)];
      case 4:
        localCrc = (localCrc >>> 8) ^ T[T8_0_start + ((localCrc ^ b.get()) & 0xff)];
      case 3:
        localCrc = (localCrc >>> 8) ^ T[T8_0_start + ((localCrc ^ b.get()) & 0xff)];
      case 2:
        localCrc = (localCrc >>> 8) ^ T[T8_0_start + ((localCrc ^ b.get()) & 0xff)];
      case 1:
        localCrc = (localCrc >>> 8) ^ T[T8_0_start + ((localCrc ^ b.get()) & 0xff)];
      default:
        /* nothing */
    }

    // Publish crc out to object
    crc = localCrc;
  }
```

You would need to add a method to get the `Crc32` method out of the `CrcInputStream` which is arguably a little hacky.

justinlin-linkedin: this might actually be a good idea given that random access to direct buffer seems to have a edge over heap buffer or just random byte array https://www.javamex.com/tutorials/io/nio_buffer_performance.shtml

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1669,7 +1710,7 @@ public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IO
       if (dataSize > Integer.MAX_VALUE) {
         throw new IOException("We only support data of max size == MAX_INT. Error while reading blob from store");
       }
-      ByteBufferInputStream output = new ByteBufferInputStream(crcStream, (int) dataSize);
+      ByteBufferInputStream output = getByteBufferInputStreamForBlobRecord(crcStream, (int) dataSize);
```
zzmao: Why cast dataSize to int? Not possible to be long?

ambry-utils/src/main/java/com.github.ambry.utils/Crc32.java
```diff
@@ -96,6 +97,51 @@ public void update(byte[] b, int off, int len) {
     crc = localCrc;
   }
 
+  public void update(ByteBuffer buffer) {
```
zzmao: `updateCrc`? or better a better name?

justinlin-linkedin: I am using the same name for other types. But I updated the method name in the CrcInputStream.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
justinlin-linkedin: updated.

cgtz: I guess I would argue that for this method, int is really the right type given that it materializes the stream contents into a single array, which has a max size of Integer.MAX_VALUE (the check which is enforced on the previous line). Functionally though, it shouldn't matter much b/c of that chunk

PR #1282, opened 2019-10-16
Add ByteBuf support in GCMCryptoService

justinlin-linkedin:
Adding ByteBuf support to CryptoService

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
+++ b/ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
@@ -13,6 +13,8 @@
  */
 package com.github.ambry.router;
 
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import java.nio.ByteBuffer;
 import java.security.GeneralSecurityException;
 
@@ -30,10 +32,26 @@ public interface CryptoService<T> {
    * @param toEncrypt {@link ByteBuffer} that needs to be encrypted
    * @param key the secret key (of type T) to use to encrypt
    * @return the {@link ByteBuffer} containing the encrypted content. Ensure the result has all
-   * the information like the IV along with the encrypted content, inorder to decrypt the content with a given key
+   * the information like the IV along with the encrypted content, in order to decrypt the content with a given key
    * @throws {@link GeneralSecurityException} on any exception with encryption
    */
   ByteBuffer encrypt(ByteBuffer toEncrypt, T key) throws GeneralSecurityException;
+
+  /**
+   * Encrypts the {@code toEncrypt} with the given key. The {@code toEncrypt} is a netty {@link ByteBuf}. This is used
+   * for encrypting data chunks.
+   * @param toEncrypt {@link ByteBuf} that needs to be encrypted
+   * @param key the secret key (of type T) to use to encrypt
+   * @return the {@link ByteBuf} containing the encrypted content. Ensure the result has all
+   * the information lik the IV along with the encrypted content, in order to decrypt the content with a given key
+   * @throws {@link GeneralSecurityException} on any exception with encryption
+   */
+  default ByteBuf encrypt(ByteBuf toEncrypt, T key) throws GeneralSecurityException {
+    ByteBuffer encrypted = encrypt(toEncrypt.nioBuffer(), key);
+    ByteBuf result = ByteBufAllocator.DEFAULT.ioBuffer(encrypted.remaining());
+    result.writeBytes(encrypted);
+    return result;
+  }
 
   /**
    * Decrypts the {@code toDecrypt} with the given key. This is used for decrypting data chunks.
@@ -43,6 +61,21 @@ public interface CryptoService<T> {
    * @throws {@link GeneralSecurityException} on any exception with decryption
    */
   ByteBuffer decrypt(ByteBuffer toDecrypt, T key) throws GeneralSecurityException;
+
+  /**
+   * Decrypts the {@code toDecrypt} with the given key. The {@code toDecrypt} is a netty {@link ByteBuf}. This is used
+   * for decrypting data chunks.
+   * @param toDecrypt {@link ByteBuf} that needs to be decrypted
+   * @param key the secret key (of type T) to use to decrypt
+   * @return the {@link ByteBuffer} containing the decrypted content
+   * @throws {@link GeneralSecurityException} on any exception with decryption
+   */
+  default ByteBuf decrypt(ByteBuf toDecrypt, T key) throws GeneralSecurityException {
+    ByteBuffer decrypted = decrypt(toDecrypt.nioBuffer(), key);
+    ByteBuf result = ByteBufAllocator.DEFAULT.ioBuffer(decrypted.remaining());
+    result.writeBytes(decrypted);
+    return result;
+  }
 
   /**
    * Returns the encrypted form of the key in bytes.
--- a/ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
+++ b/ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
@@ -14,7 +14,7 @@
 package com.github.ambry.router;
 
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.Unpooled;
 import java.nio.ByteBuffer;
 import java.security.GeneralSecurityException;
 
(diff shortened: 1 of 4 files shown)


ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
```diff
@@ -30,11 +32,27 @@
...
+   * for encrypting data chunks.
+   * @param toEncrypt {@link ByteBuf} that needs to be encrypted
+   * @param key the secret key (of type T) to use to encrypt
+   * @return the {@link ByteBuf} containing the encrypted content. Ensure the result has all
+   * the information lik the IV along with the encrypted content, in order to decrypt the content with a given key
+   * @throws {@link GeneralSecurityException} on any exception with encryption
+   */
+  default ByteBuf encrypt(ByteBuf toEncrypt, T key) throws GeneralSecurityException {
+    ByteBuffer encrypted = encrypt(toEncrypt.nioBuffer(), key);
+    ByteBuf result = ByteBufAllocator.DEFAULT.ioBuffer(encrypted.remaining());
```
cgtz: Why not use `Unpooled.wrappedBuffer` in this case instead of copying?

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
```diff
@@ -78,6 +98,52 @@ public ByteBuffer encrypt(ByteBuffer toEncrypt, SecretKeySpec key) throws Genera
...
+      } else {
+        byte[] toEncryptedArray = new byte[toEncrypt.readableBytes()];
+        toEncrypt.readBytes(toEncryptedArray);
+        int n = encrypter.doFinal(toEncryptedArray, 0, toEncryptedArray.length, encryptedContent.array(),
+            encryptedContent.arrayOffset() + encryptedContent.writerIndex());
+        encryptedContent.writerIndex(encryptedContent.writerIndex() + n);
+      }
+      return encryptedContent;
+    } catch (Exception e) {
+      throw new GeneralSecurityException("Expception thrown while encrypting data", e);
```
cgtz: typo in exception

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
```diff
@@ -78,6 +98,52 @@ public ByteBuffer encrypt(ByteBuffer toEncrypt, SecretKeySpec key) throws Genera
...
+      int outputSize = encrypter.getOutputSize(toEncrypt.readableBytes());
+      // We need to expose the array address to the encrypter, so here we need to allocate a ByteBuf at heap.
+      ByteBuf encryptedContent =
+          ByteBufAllocator.DEFAULT.heapBuffer(IVRecord_Format_V1.getIVRecordSize(iv) + outputSize);
+      IVRecord_Format_V1.serializeIVRecord(encryptedContent, iv);
+      if (toEncrypt.hasArray()) {
+        int n = encrypter.doFinal(toEncrypt.array(), toEncrypt.arrayOffset() + toEncrypt.readerIndex(),
+            toEncrypt.readableBytes(), encryptedContent.array(),
+            encryptedContent.arrayOffset() + encryptedContent.writerIndex());
+        toEncrypt.readerIndex(toEncrypt.readerIndex() + toEncrypt.readableBytes());
```
cgtz: I wonder if it would be possible to just use the nioBuffers() method on toEncrypt.
Then you could potentially make multiple calls to update (https://docs.oracle.com/javase/7/docs/api/javax/crypto/Cipher.html#update(java.nio.ByteBuffer,%20java.nio.ByteBuffer)) and end it with a call to doFinal (https://docs.oracle.com/javase/7/docs/api/javax/crypto/Cipher.html#doFinal(java.nio.ByteBuffer,%20java.nio.ByteBuffer))

Making the output buffer direct might be more challenging though so that one might have to stay as a heap buffer

build.gradle
```diff
@@ -138,6 +138,7 @@ project(':ambry-api') {
         compile "com.codahale.metrics:metrics-core:$metricsVersion"
         compile "org.codehaus.jackson:jackson-core-asl:$jacksonVersion"
         compile "org.codehaus.jackson:jackson-mapper-asl:$jacksonVersion"
+        compile "io.netty:netty-all:$nettyVersion"
```
cgtz: I wonder if we could limit the dependency to "io.netty:netty-buffer:$nettyVersion" in ambry-api. A lot of things (including java clients) depend on the ambry-api subproject.

ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
justinlin-linkedin: We should use the Unpooled ByteBuf since it's just default method which will be override by the actually implementation.

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
justinlin-linkedin: the actually implementation of those methods is pretty much what we are doing here. If bytebuffer doesn't have array (not heap buffer or read only), then it creates an temporary byte array and fill it with the data from byte buffer. I don't really see memory footprint would be reduced when if we use nioBuffer.

build.gradle
justinlin-linkedin: seems like a good idea, will update.

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
cgtz: Ah ok, that makes sense. I didn't take a look into the impls, only the docs.

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
```diff
@@ -58,14 +62,30 @@ public GCMCryptoService(CryptoServiceConfig cryptoServiceConfig) {
...
   @Override
   public ByteBuffer encrypt(ByteBuffer toEncrypt, SecretKeySpec key) throws GeneralSecurityException {
+    return encrypt(toEncrypt, key, false);
+  }
+
+  /**
+   * Helper function to encrypt ByteBuffer with the given key. When useFixedIv is true, don't use a random iv byte
+   * array, using a all zero byte array instead. Only set it to be true in test.
+   * @param toEncrypt {@link ByteBuffer} that needs to be encrypted
+   * @param key the secret key (of type T) to use to encrypt
```
zzmao: java doc for `useFixedIv`

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
```diff
@@ -58,14 +62,30 @@ public GCMCryptoService(CryptoServiceConfig cryptoServiceConfig) {
...
+   * @return the {@link ByteBuffer} containing the encrypted content. Ensure the result has all
+   * the information like the IV along with the encrypted content, in order to decrypt the content with a given key
+   * @throws {@link GeneralSecurityException} on any exception with encryption
+   */
+  ByteBuffer encrypt(ByteBuffer toEncrypt, SecretKeySpec key, boolean useFixedIv) throws GeneralSecurityException {
```
zzmao: Alternative, you can use `iv` as an argument. 
For test, pass in a fixedIv
For prod, pass in null. 

Depends on you. I am ok with both.

cgtz: I think I would prefer this way of passing in the `iv`. Then you could move the `fixedIvVal` constant out of this class and into the test class.

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
cgtz: One follow up question about this. Given how this method has to either access the array or copy the buffer, would this method impl save resources over just using the default method defined in the interface that calls `ByteBuf.nioBuffer()`?

ambry-router/src/main/java/com.github.ambry.router/GCMCryptoService.java
```diff
@@ -93,6 +160,33 @@ public ByteBuffer decrypt(ByteBuffer toDecrypt, SecretKeySpec key) throws Genera
     }
   }
 
+  @Override
+  public ByteBuf decrypt(ByteBuf toDecrypt, SecretKeySpec key) throws GeneralSecurityException {
+    try {
+      Cipher decrypter = Cipher.getInstance(GCM_CRYPTO_INSTANCE, "BC");
```
cgtz: Same question here, since we are using heap buffers

justinlin-linkedin: I take my words back about the nioBuffer. The implementation would actually save the memory when the buffer doesn't have array, since the temporary array it creates has a upper limit at 4096. If the buffer is larger than that, say 4MB, it will only create a 1/10 of what I did here, I will update.

PR #1286, opened 2019-10-21
Recovery fixes

ankagrawal:
This fixes two issues with recovery, and adds a more robust recovery test.

ISSUES
- CloudMessageReadSet should keep track of partial sends over network.
- OperationTime should be determined by creation time or delete time, and not ttl update time.

TEST
- Create a separate test for large blobs recovery.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -135,8 +135,7 @@ class CloudBlobStore implements Store {
       for (BlobId blobId : blobIdList) {
         CloudBlobMetadata blobMetadata = cloudBlobMetadataListMap.get(blobId.getID());
         MessageInfo messageInfo = new MessageInfo(blobId, blobMetadata.getSize(), blobMetadata.getExpirationTime(),
-            (short) blobMetadata.getAccountId(), (short) blobMetadata.getContainerId(),
-            getOperationTime(blobMetadata, currentTimeStamp));
+            (short) blobMetadata.getAccountId(), (short) blobMetadata.getContainerId(), getOperationTime(blobMetadata));
         messageInfos.add(messageInfo);
         blobReadInfos.add(new CloudMessageReadSet.BlobReadInfo(blobMetadata, blobId));
       }
@@ -195,7 +194,7 @@ class CloudBlobStore implements Store {
    * @param metadata blob metadata from which to derive operation time.
    * @return operation time.
    */
-  private long getOperationTime(CloudBlobMetadata metadata, long currentTimeStamp) {
+  private long getOperationTime(CloudBlobMetadata metadata) {
     if (isBlobDeleted(metadata)) {
       return metadata.getDeletionTime();
     }
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -198,8 +198,6 @@ class CloudBlobStore implements Store {
   private long getOperationTime(CloudBlobMetadata metadata, long currentTimeStamp) {
     if (isBlobDeleted(metadata)) {
       return metadata.getDeletionTime();
-    } else if (isBlobExpired(metadata, currentTimeStamp)) {
-      return metadata.getExpirationTime();
     }
     return (metadata.getCreationTime() == Utils.Infinite_Time) ? metadata.getUploadTime() : metadata.getCreationTime();
   }
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
@@ -52,6 +52,9 @@ class CloudMessageReadSet implements MessageReadSet {
       if (blobReadInfo.isPrefetched()) {
         ByteBuffer outputBuffer = blobReadInfo.getPrefetchedBuffer();
         outputBuffer.flip();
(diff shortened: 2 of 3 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -52,6 +52,9 @@ public long writeTo(int index, WritableByteChannel channel, long relativeOffset,
       if (blobReadInfo.isPrefetched()) {
         ByteBuffer outputBuffer = blobReadInfo.getPrefetchedBuffer();
         outputBuffer.flip();
```
jsjtzyy: Do we still need this flip()?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -198,8 +198,6 @@ private void validateCloudMetadata(Map<String, CloudBlobMetadata> cloudBlobMetad
   private long getOperationTime(CloudBlobMetadata metadata, long currentTimeStamp) {
     if (isBlobDeleted(metadata)) {
       return metadata.getDeletionTime();
-    } else if (isBlobExpired(metadata, currentTimeStamp)) {
-      return metadata.getExpirationTime();
```
lightningrob: What problem were these lines causing?

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
ankagrawal: No we dont. Good catch. Will remove them.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -198,8 +198,6 @@ private void validateCloudMetadata(Map<String, CloudBlobMetadata> cloudBlobMetad
   private long getOperationTime(CloudBlobMetadata metadata, long currentTimeStamp) {
     if (isBlobDeleted(metadata)) {
```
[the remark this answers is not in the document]
ankagrawal: Will remove.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
ankagrawal: These lines were not causing any problems. But based on how the operation time is calculated during replication between Ambry data nodes, this logic is inconsistent with that.

In PersistentIndex::findKey() which is called by findEntriesSince of BlobStore, the ttl entry doesn't change the operation time of the entry being replicated. In fact, in order to replicate a TTL_Entry record, the previous PUT record is found, and its ttl flag is set. Hence the operation time is not TTL update's time, rather its the previous PUT time. (Source: PersistentIndex::line609).

ambry-server/src/integration-test/java/com.github.ambry.server/VcrRecoveryTest.java
```diff
@@ -47,106 +65,224 @@
...
+
+    channel.send(getRequest);
+    GetResponse getResponse =
+        GetResponse.readFrom(new DataInputStream(channel.receive().getInputStream()), recoveryCluster.getClusterMap());
+
+    for (PartitionResponseInfo partitionResponseInfo : getResponse.getPartitionResponseInfoList()) {
+      assertEquals("Error in getting the recovered blobs", ServerErrorCode.No_Error,
+          partitionResponseInfo.getErrorCode());
+      for (MessageInfo messageInfo : partitionResponseInfo.getMessageInfoList()) {
+        assertEquals(blobIdToSizeMap.get(messageInfo.getStoreKey()) + 270, messageInfo.getSize());
```
jsjtzyy: nit: cast `messageInfo.getStoreKey()` to `BlobId` type

ambry-server/src/integration-test/java/com.github.ambry.server/VcrRecoveryTest.java
```diff
@@ -47,106 +65,224 @@
...
+   * @throws Exception If an exception happens.
+   */
+  @Test
+  public void cloudRecoveryTestForLargeBlob() throws Exception {
+    // Create blobs and upload to cloud destination.
+    int userMetaDataSize = 100;
+    byte[] userMetadata = new byte[userMetaDataSize];
+    TestUtils.RANDOM.nextBytes(userMetadata);
+    Map<BlobId, Integer> blobIdToSizeMap = new HashMap<>();
+    int blobSize = FOUR_MB_SZ;
```
jsjtzyy: So, we use 4MB to mock large blob, right? I guess this is biggest size for a chunk from ambry server's point of view. Maybe adding some comment about why we choose 4MB would be better.

ambry-server/src/integration-test/java/com.github.ambry.server/VcrRecoveryTest.java
```diff
@@ -47,106 +65,224 @@
...
+      CloudBlobMetadata cloudBlobMetadata =
+          new CloudBlobMetadata(blobId, time, Utils.Infinite_Time, putMessageFormatInputStream.getSize(),
+              CloudBlobMetadata.EncryptionOrigin.NONE);
+      latchBasedInMemoryCloudDestination.uploadBlob(blobId, putMessageFormatInputStream.getSize(), cloudBlobMetadata,
+          putMessageFormatInputStream);
+    }
+
+    // Waiting for download attempt
     assertTrue("Did not recover all blobs in 2 minutes",
         latchBasedInMemoryCloudDestination.awaitDownload(2, TimeUnit.MINUTES));
```
jsjtzyy: quick question: this read-after-write test is to ensure all blobs were successfully uploaded to cloud, right?
2 mins limit seems a little long, can we change it to smaller number?

PR #1287, opened 2019-10-22
Update snapshotVersion of Account in AccountUpdateTool

justinlin-linkedin:
Update snapshotVersion of Accounts in AccountUpdateTool before passing the Accounts to account service. This way, we don't have to change snapshot versions for every account to update.

### DIFF
--- a/ambry-tools/src/main/java/com.github.ambry/account/AccountUpdateTool.java
+++ b/ambry-tools/src/main/java/com.github.ambry/account/AccountUpdateTool.java
@@ -27,7 +27,9 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


ambry-tools/src/main/java/com.github.ambry/account/AccountUpdateTool.java
```diff
@@ -271,7 +273,23 @@ void updateAccountsFromFile(String accountJsonPath) throws IOException {
     long startTime = System.currentTimeMillis();
     Collection<Account> accountsToUpdate = getAccountsFromJson(accountJsonPath);
     if (!hasDuplicateAccountIdOrName(accountsToUpdate)) {
-      if (accountService.updateAccounts(accountsToUpdate)) {
+      Collection<Account> allAccounts = accountService.getAllAccounts();
+      // Update the snapshot version to resolve conflict
+      Map<Short, Account> existingAccountsMap = new HashMap<>();
```
cgtz: since we may not always want to avoid the version check, maybe keep the default behavior the same and add a --force option that will allow unconditional overrides

justinlin-linkedin: added a new option --ignoreSnapshotVersion

PR #1291, opened 2019-10-25
Local network client for cloud-facing router

lightningrob:
Note: NetworkClient made interface and impl moved to SocketNetworkClient, code is not new


ambry-network/src/main/java/com.github.ambry.network/SocketNetworkClient.java
```diff
@@ -0,0 +1,416 @@
...
+ * and receive responses for sent requests. Requests that come in via {@link #sendAndPoll(List, Set, int)} call,
+ * that could not be immediately sent are queued, and an attempt will be made in subsequent invocations of the call (or
+ * until they time out).
+ * (Note: We will empirically determine whether, rather than queueing a request,
+ * a request should be failed if connections could not be checked out if pool limit for its hostPort has been reached
+ * and all connections to the hostPort are unavailable).
+ *
+ * This class is not thread safe.
+ */
+public class SocketNetworkClient implements NetworkClient {
```
lightningrob: Note: code in this class copied from old NetworkClient class which is now an interface.

ambry-network/src/main/java/com.github.ambry.network/SocketRequestResponseChannel.java
```diff
@@ -30,7 +30,7 @@
   private final String connectionId;
   private final InputStream input;
   private final long startTimeInMs;
-  private Logger logger = LoggerFactory.getLogger(getClass());
+  private static final Logger logger = LoggerFactory.getLogger(SocketServerRequest.class);
```
lightningrob: Just noticed I put the wrong class here and will fix.

ambry-network/src/main/java/com.github.ambry.network/LocalNetworkClient.java
```diff
@@ -0,0 +1,116 @@
...
+   * Instantiates a LocalNetworkClient.
+   * @param channel the {@link RequestResponseChannel} to send requests to.
+   * @param processorId the id used to isolate requests and responses sent across the shared channel.
+   * @param networkMetrics the metrics to track the network related metrics
+   * @param time The Time instance to use.
+   */
+  public LocalNetworkClient(LocalRequestResponseChannel channel, int processorId, NetworkMetrics networkMetrics,
+      Time time) {
+    this.channel = channel;
+    this.processorId = processorId;
```
jsjtzyy: I still don't quite understand the `processorId`, what the difference between processorId and correlationId? Looks like the processorId is associated with LocalNetworkClient. So, there are several LocalNetworkClients sharing same channel?

ambry-network/src/main/java/com.github.ambry.network/LocalRequestResponseChannel.java
```diff
@@ -0,0 +1,157 @@
...
+      }
+    }
+  }
+
+  /**
+   * @return the response list corresponding to a processor id.
+   * @param processorId the processor id to match.
+   */
+  private List<ResponseInfo> getResponseList(int processorId) {
+    return responseMap.computeIfAbsent(processorId, p -> new ArrayList<>());
```
jsjtzyy: where does `responseMap` get populated?

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Construct and return a {@link NonBlockingRouter} that works with cloud storage.
+   * @return a {@link NonBlockingRouter}
+   */
+  @Override
+  public Router getRouter() {
+    try {
+      NonBlockingRouter router = new NonBlockingRouter(routerConfig, routerMetrics, networkClientFactory, notificationSystem, clusterMap,
+          kms, cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass);
```
jsjtzyy: minor: format this file.

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
```diff
@@ -0,0 +1,163 @@
...
+      router.setRequestHandlerPool(requestHandlerPool);
```
jsjtzyy: I don't see requestHandlerPool is called in router, will it be used in future PR?

ambry-network/src/main/java/com.github.ambry.network/LocalRequestResponseChannel.java
lightningrob: It is populated lazily/on-demand on this line.

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
lightningrob: The NonBlockingRouter closes it on shutdown.  It may not the the ideal place to do it but I couldn't think of a better place to put it.  The factory itself has no shutdown hook.

ambry-network/src/main/java/com.github.ambry.network/LocalNetworkClient.java
lightningrob: Yes, that is because each OperationController creates its own NetworkClient sharing the same channel, and when any of them retrieves its responses, it needs to get only the ones for the requests that it sent.  But it's not quite the same as correlationId since it doesn't correlate request with response.  It just segregates all the responses into separate buckets based on which client sent them.  I coped the name from the existing network client/channel classes.  I can change it to clientId if you prefer.

ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
```diff
@@ -14,81 +14,21 @@
...
 import java.util.List;
-import java.util.ListIterator;
-import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
+// TODO: move to api along with Request/ResponseInfo
```
ankagrawal: Is there a reason why we cant move now?

ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
```diff
@@ -97,322 +37,15 @@ public NetworkClient(Selector selector, NetworkConfig networkConfig, NetworkMetr
...
-    return successfulConnections;
-  }
-
-  /**
-   * Start creating a connection to a host using {@link Selector#connect}.
-   * @param host the hostname to connect to.
-   * @param port the port to connect to.
-   * @return the connection ID.
-   * @throws IOException if DNS resolution fails on the hostname or if the server is down
+   * @throws IllegalStateException if the SocketNetworkClient is closed.
```
ankagrawal: NetworkClient instead of SocketNetworkClient?

ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
```diff
@@ -97,322 +37,15 @@ public NetworkClient(Selector selector, NetworkConfig networkConfig, NetworkMetr
...
-   * @param responseInfoList the list to populate with {@link ResponseInfo} objects for responses created based on
-   *                         the selector events.
-   */
-  private void handleSelectorEvents(List<ResponseInfo> responseInfoList) {
-    for (String connId : selector.connected()) {
-      logger.trace("Checking in connection back to connection tracker for connectionId {} ", connId);
-      connectionTracker.checkInConnection(connId);
-      pendingConnectionsToAssociatedRequests.remove(connId);
-    }
+  int warmUpConnections(List<DataNodeId> dataNodeIds, int connectionWarmUpPercentagePerDataNode, long timeForWarmUp,
```
ankagrawal: Missing javadocs.

ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
```diff
@@ -97,322 +37,15 @@ public NetworkClient(Selector selector, NetworkConfig networkConfig, NetworkMetr
...
-              connId, dataNodeId);
-          // Explicitly set requestInfo = null in ResponseInfo, the OperationController should detect this and directly
-          // notify ResponseHandler without handing it over to PutManager/GetManager/DeleteManager/TtlUpdateManager.
-          responseInfoList.add(new ResponseInfo(null, NetworkClientErrorCode.NetworkError, null, dataNodeId));
-          // No need to call pendingRequests.remove() because it has been removed due to connection unavailability in prepareSends()
-        }
-      }
-      networkMetrics.connectionDisconnected.inc();
-    }
+  void wakeup();
```
ankagrawal: missing javadocs

ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
```diff
@@ -97,322 +37,15 @@ public NetworkClient(Selector selector, NetworkConfig networkConfig, NetworkMetr
...
-    /**
-     * Actions to be done on receiving response for the request sent
-     */
-    void onResponseReceive() {
-      networkMetrics.networkClientRoundTripTime.update(time.milliseconds() - requestDequeuedAtMs,
-          TimeUnit.MILLISECONDS);
-      networkMetrics.networkClientTotalTime.update(time.milliseconds() - requestQueuedAtMs, TimeUnit.MILLISECONDS);
-    }
-  }
+  public void close();
```
ankagrawal: public not needed.

ambry-network/src/main/java/com.github.ambry.network/NetworkClientFactory.java
```diff
@@ -13,53 +13,18 @@
...
    * @return return a new {@link NetworkClient}
-   * @throws IOException if the {@link Selector} could not be instantiated.
+   * @throws IOException if the {@link NetworkClient} could not be instantiated.
    */
-  public NetworkClient getNetworkClient() throws IOException {
-    Selector selector = new Selector(networkMetrics, time, sslFactory, networkConfig);
-    return new NetworkClient(selector, networkConfig, networkMetrics, maxConnectionsPerPortPlainText,
-        maxConnectionsPerPortSsl, connectionCheckoutTimeoutMs, time);
-  }
+  public NetworkClient getNetworkClient() throws IOException;
```
ankagrawal: public not needed.

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
```diff
@@ -0,0 +1,163 @@
...
+   * @param verifiableProperties the in-memory properties to use to construct configurations.
+   * @param clusterMap the {@link ClusterMap} to use to determine where operations should go.
+   * @param notificationSystem the {@link NotificationSystem} to use to log operations.
+   * @param sslFactory the {@link SSLFactory} to support SSL transmissions. Required if SSL is enabled for any
+   *                   datacenters.
+   * @param accountService the {@link AccountService} to use.
+   * @throws Exception if any of the arguments are null or if instantiation of KMS or CryptoService fails
+   */
+  public CloudRouterFactory(VerifiableProperties verifiableProperties, ClusterMap clusterMap,
+      NotificationSystem notificationSystem, SSLFactory sslFactory, AccountService accountService) throws Exception {
```
ankagrawal: Is throwing a generic exception ok? Should we more specific?

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
```diff
@@ -0,0 +1,163 @@
...
+
+    // TODO: need to shut pool down on exit, may need to move to RestServer
+    // Or pass it to router
+    requestHandlerPool = getRequestHandlerPool(verifiableProperties, clusterMap, notificationSystem);
+    networkClientFactory = new LocalNetworkClientFactory((LocalRequestResponseChannel) requestHandlerPool.getChannel(),
+        new NetworkConfig(verifiableProperties), new NetworkMetrics(registry), time);
+    logger.info("Instantiated CloudRouterFactory");
+  }
+
+  public static RequestHandlerPool getRequestHandlerPool(VerifiableProperties verifiableProperties,
```
ankagrawal: do we need this to be public static?

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
ankagrawal: Cant we just pass requestHandlerPool in router's constructor?

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
```diff
@@ -0,0 +1,163 @@
...
+   * Construct and return a {@link NonBlockingRouter} that works with cloud storage.
+   * @return a {@link NonBlockingRouter}
+   */
+  @Override
+  public Router getRouter() {
+    try {
+      NonBlockingRouter router = new NonBlockingRouter(routerConfig, routerMetrics, networkClientFactory, notificationSystem, clusterMap,
+          kms, cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass);
+      router.setRequestHandlerPool(requestHandlerPool);
+      return router;
```
ankagrawal: We have one instance of RequestHandlerPool here, which will be used for every getRouter() request. And looks like in router.close() we are doing a shutdown of the RequestHandlerPool. Can this cause unintended side-effect for other router instances? If we intend to create just one Router object, then we should make it a singleton.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -58,7 +58,7 @@ public ReplicationManager(ReplicationConfig replicationConfig, ClusterMapConfig
     // initialize all partitions
     for (ReplicaId replicaId : replicaIds) {
       PartitionId partition = replicaId.getPartitionId();
-      Store store = storageManager.getStore(partition, false);
```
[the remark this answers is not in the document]
ankagrawal: Ignore this comment. Just realized we have another getStore() method that does what the above code change does.

ambry-network/src/main/java/com.github.ambry.network/ResponseInfo.java
```diff
@@ -18,7 +18,7 @@
 
 
 /**
- * The response from a {@link NetworkClient} comes in the form of an object of this class.
+ * The response from a {@link SocketNetworkClient} comes in the form of an object of this class.
```
cgtz: This applies to AzureNetworkClient as well, right?

ambry-network/src/main/java/com.github.ambry.network/SocketNetworkClient.java
```diff
@@ -0,0 +1,416 @@
...
+   * @param requestsToSend the list of {@link RequestInfo} representing the requests that need to be sent out. This
+   *                     could be empty.
+   * @param requestsToDrop the list of correlation IDs representing the requests that can be dropped by
+   *                       closing the connection.
+   * @param pollTimeoutMs the poll timeout.
+   * @return a list of {@link ResponseInfo} representing the responses received for any requests that were sent out
+   * so far.
+   * @throws IllegalStateException if the SocketNetworkClient is closed.
+   */
+  public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Integer> requestsToDrop,
```
cgtz: add `@Override` annotations for all the methods that are defined in the interface

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
```diff
@@ -0,0 +1,163 @@
...
+    DataNodeId nodeId = new CloudDataNode(cloudConfig, clusterMapConfig);
+    CloudDestinationFactory cloudDestinationFactory =
+        Utils.getObj(cloudConfig.cloudDestinationFactoryClass, verifiableProperties, registry);
+    CloudDestination cloudDestination = cloudDestinationFactory.getCloudDestination();
+    VcrMetrics vcrMetrics = new VcrMetrics(registry);
+    StoreManager cloudStorageManager =
+        new CloudStorageManager(verifiableProperties, vcrMetrics, cloudDestination, clusterMap);
+    LocalRequestResponseChannel channel = new LocalRequestResponseChannel();
+    ServerMetrics serverMetrics = new ServerMetrics(registry, AmbryRequests.class);
+    StoreKeyFactory storeKeyFactory = Utils.getObj(storeConfig.storeKeyFactory, clusterMap);
```
cgtz: since there are other assumptions in the router that blobids will be used, you can just use BlobIdFactory directly w/o reflection.

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
```diff
@@ -0,0 +1,163 @@
...
+import com.github.ambry.utils.SystemTime;
+import com.github.ambry.utils.Time;
+import com.github.ambry.utils.Utils;
+import java.io.IOException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Implementation of {@link RouterFactory} that creates routers that work with cloud storage.
```
cgtz: Lets add a javadoc to say that this is experimental for now.

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
cgtz: instead of passing the requesthandlerpool in maybe we could pass in a lambda that represents a shutdown hook that will be called on close

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
lightningrob: for test

ambry-network/src/main/java/com.github.ambry.network/LocalNetworkClient.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @Override
+  public int warmUpConnections(List<DataNodeId> dataNodeIds, int connectionWarmUpPercentagePerDataNode,
+      long timeForWarmUp, List<ResponseInfo> responseInfoList) {
+    return 0;
+  }
+
+  @Override
+  public void wakeup() {
+
```
cgtz: If we introduced timed waiting to `channel.receiveResponses(processorId)` we would also need to implement `wakeup`

ambry-network/src/main/java/com.github.ambry.network/LocalNetworkClient.java
```diff
@@ -0,0 +1,116 @@
...
+      for (RequestInfo requestInfo : requestInfos) {
+        ByteBuffer buffer = LocalRequestResponseChannel.byteBufferFromPayload(requestInfo.getRequest());
+        channel.sendRequest(new LocalChannelRequest(requestInfo, processorId, new ByteBufferInputStream(buffer)));
+      }
+    } catch (Exception e) {
+      logger.error("Received an unexpected error during sendAndPoll(): ", e);
+      networkMetrics.networkClientException.inc();
+    }
+
+    List<ResponseInfo> responses = channel.receiveResponses(processorId);
```
cgtz: you might want to introduce some form of a timed wait mechanism here. In low qps situations this could run as a very tight loop when no responses are received

ambry-network/src/main/java/com.github.ambry.network/LocalRequestResponseChannel.java
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  /**
+   * Receive all queued responses corresponding to requests matching a processor id.
+   * @param processorId the processor id to match.
+   * @return the applicable responses.
+   */
+  public List<ResponseInfo> receiveResponses(int processorId) {
+    List<ResponseInfo> responseList = getResponseList(processorId);
+    synchronized (responseList) {
```
cgtz: maybe consider using a blocking queue instead of a plain ArrayList for `responseList` since it supports timed waits.

ambry-router/src/test/java/com.github.ambry.router/CloudRouterTest.java
```diff
@@ -0,0 +1,277 @@
...
+  /**
+   * Constructs and returns a VerifiableProperties instance with the defaults required for instantiating
+   * the {@link NonBlockingRouter}.
+   * @return the created VerifiableProperties instance.
+   */
+  @Override
+  protected Properties getNonBlockingRouterProperties(String routerDataCenter) {
+    Properties properties = super.getNonBlockingRouterProperties(routerDataCenter);
+    // Set parallelism and success quorum to 1 for all operations
+    properties.setProperty("router.put.request.parallelism", Integer.toString(REQUEST_PARALLELISM));
```
cgtz: adding a note that we will eventually have to consider changing how these configs work when we add support for addressing both LocalNetworkClient and SocketNetworkClient. This will require more significant changes to OperationTracker

ambry-network/src/main/java/com.github.ambry.network/LocalNetworkClient.java
```diff
@@ -0,0 +1,116 @@
...
+   * @return a list of {@link ResponseInfo} representing the responses received for any requests that were sent out
+   * so far.
+   * @throws IllegalStateException if the SocketNetworkClient is closed.
+   */
+  public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestInfos, Set<Integer> requestsToDrop,
+      int pollTimeoutMs) {
+    if (closed) {
+      throw new IllegalStateException("The client is closed.");
+    }
+    if (processorId < 0) {
```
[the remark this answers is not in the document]
lightningrob: no, that was for old code.

ambry-network/src/main/java/com.github.ambry.network/LocalRequestResponseChannel.java
```diff
@@ -0,0 +1,157 @@
...
+  private BlockingQueue<Request> requestQueue = new LinkedBlockingQueue<>();
+  private Map<Integer, List<ResponseInfo>> responseMap = new ConcurrentHashMap<>();
+  // buffer to hold size header that we strip off payloads
+  private static final byte[] sizeByteArray = new byte[Long.BYTES];
+
+  @Override
+  public void sendRequest(Request request) {
+    requestQueue.offer(request);
+    if (request instanceof LocalChannelRequest) {
+      LocalChannelRequest localRequest = (LocalChannelRequest) request;
```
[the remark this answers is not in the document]
lightningrob: Yes, they will get EmptyRequest on shutdown.

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
lightningrob: there are many to handle.

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
lightningrob: changing to create pool etc. in getRouter()

ambry-network/src/main/java/com.github.ambry.network/ResponseInfo.java
lightningrob: reverted the change

ambry-network/src/main/java/com.github.ambry.network/SocketNetworkClient.java
lightningrob: ok

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
lightningrob: ok

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
lightningrob: Storing it as list of Closeable and making ReqHandlerPool implement Closeable.

ambry-network/src/main/java/com.github.ambry.network/LocalRequestResponseChannel.java
lightningrob: Will do.

ambry-network/src/main/java/com.github.ambry.network/LocalNetworkClient.java
lightningrob: What would the wakeup() method do?

ambry-router/src/main/java/com.github.ambry.router/CloudRouterFactory.java
lightningrob: Done.

ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
lightningrob: Done.

ambry-network/src/main/java/com.github.ambry.network/LocalNetworkClient.java
lightningrob: Discussed with Casey.  It would have to either interrupt the main thread (ugly) or put a dummy response in the appropriate queue to make the poll() call return.  Will consider for a future PR.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudMessageReadSet.java
```diff
@@ -47,23 +46,21 @@ public long writeTo(int index, WritableByteChannel channel, long relativeOffset,
...
-      if (blobReadInfo.isPrefetched()) {
-        ByteBuffer outputBuffer = blobReadInfo.getPrefetchedBuffer();
-        long sizeToRead = Math.min(maxSize, blobReadInfo.getBlobSize() - relativeOffset);
-        outputBuffer.limit((int) (relativeOffset + sizeToRead));
-        outputBuffer.position((int) (relativeOffset));
-        written = channel.write(outputBuffer);
-      } else {
-        blobReadInfo.downloadBlob(blobStore, Channels.newOutputStream(channel));
-        written = blobReadInfo.getBlobSize();
+      if (!blobReadInfo.isPrefetched()) {
```
cgtz: could you add a comment explaining why we require prefetch to always happen

ambry-network/src/main/java/com.github.ambry.network/LocalRequestResponseChannel.java
```diff
@@ -0,0 +1,161 @@
...
+
+  @Override
+  public void shutdown() {
+  }
+
+  /**
+   * Utility to extract a byte buffer from a {@link Send} object, skipping the size header.
+   * @param payload the payload whose bytes we want.
+   */
+  static ByteBuffer byteBufferFromPayload(Send payload) throws IOException {
```
cgtz: I guess this method is temporary before we find a better way to avoid the extra serialization and then deserialization?

lightningrob: Yes, added a TODO.

PR #1293, opened 2019-10-28
Refactor the replication engine to create a separate replication mananger from cloud to store and a separate cloud replica token file

ankagrawal:
Add CloudToStoreReplicationManager to separate out replication from cloud replicas to data nodes.
Add Code to enable spectator on helix vcr cluster and to react to changes on helix.
Add listener for partition leadership changes.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterSpectator.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterSpectator.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2019 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.clustermap;
+
+import org.apache.helix.api.listeners.InstanceConfigChangeListener;
+import org.apache.helix.api.listeners.LiveInstanceChangeListener;
+
+
+/**
+ * {@link ClusterSpectator} adds itself as a spectator of a cluster and registers some cluster change listeners.
+ */
+public interface ClusterSpectator extends InstanceConfigChangeListener, LiveInstanceChangeListener {
+
+  /**
+   * Register with helix cluster as a spectator.
+   */
+  void spectate() throws Exception;
+
+  /**
+   * Register a listener for changes to instance config in cluster.
+   * @param instanceConfigChangeListener listener to add.
+   */
+  void registerInstanceConfigChangeListener(InstanceConfigChangeListener instanceConfigChangeListener);
+
+  /**
+   * Register a listener for changes to instance liveness information in cluster.
+   * @param liveInstanceChangeListener listener to add.
+   */
+  void registerLiveInstanceChangeListener(LiveInstanceChangeListener liveInstanceChangeListener);
+}
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/PartitionStateChangeListener.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/PartitionStateChangeListener.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2019 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.clustermap;
+
+/**
+ * {@link PartitionStateChangeListener} takes action when partition state changes.
+ */
+public interface PartitionStateChangeListener {
+
+  /**
+   * Action to take on becoming leader of a partition.
+   * @param partitionName of the partition.
+   */
+  void onPartitionLeadFromStandby(String partitionName);
+
+  /**
+   * Action to take on being removed as leader of a partition.
+   * @param partitionName of the partition.
+   */
+  void onPartitionStandbyFromLead(String partitionName);
+}
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/PartitionStateChangeListener.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/PartitionStateChangeListener.java
@@ -22,11 +22,11 @@ public interface PartitionStateChangeListener {
    * Action to take on becoming leader of a partition.
    * @param partitionName of the partition.
    */
-  void onPartitionLeadFromStandby(String partitionName);
+  void onPartitionStateChangeToLeaderFromStandby(String partitionName);
 
   /**
    * Action to take on being removed as leader of a partition.
    * @param partitionName of the partition.
    */
-  void onPartitionStandbyFromLead(String partitionName);
+  void onPartitionStateChangeToStandbyFromLeader(String partitionName);
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
@@ -23,7 +23,6 @@ public class ClusterMapConfig {
   public static final String DEFAULT_STATE_MODEL_DEF = LeaderStandbySMD.name;
   public static final String AMBRY_STATE_MODEL_DEF = "AmbryLeaderStandby";
   private static final String MAX_REPLICAS_ALL_DATACENTERS = "max-replicas-all-datacenters";
-  private static final String DEFAULT_VCR_DATACENTER_NAME = "ei4";
 
   /**
    * The factory class used to get the resource state policies.
@@ -184,10 +183,9 @@ public class ClusterMapConfig {
   public final boolean clustermapListenCrossColo;
 
   /**
-   * Name of the datacenter of vcrnodes. It is expected that all the vcr nodes will reside in the same datacenter.
+   * Name of the datacenter of vcr nodes. It is expected that all the vcr nodes will reside in the same datacenter.
    */
   @Config("clustermap.vcr.datacenter.name")
-  @Default(DEFAULT_VCR_DATACENTER_NAME)
   public final String clustermapVcrDatacenterName;
 
   /**
@@ -232,7 +230,7 @@ public class ClusterMapConfig {
     clustermapStateModelDefinition =
         verifiableProperties.getString("clustermap.state.model.definition", DEFAULT_STATE_MODEL_DEF);
     clustermapVcrDatacenterName =
-        verifiableProperties.getString("clustermap.vcr.datacenter.name", DEFAULT_VCR_DATACENTER_NAME);
+        verifiableProperties.getString("clustermap.vcr.datacenter.name", null);
     if (!clustermapStateModelDefinition.equals(DEFAULT_STATE_MODEL_DEF) && !clustermapStateModelDefinition.equals(
         AMBRY_STATE_MODEL_DEF)) {
       throw new IllegalArgumentException("Unsupported state model definition: " + clustermapStateModelDefinition);
--- a/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
@@ -23,6 +23,7 @@ public class ClusterMapConfig {
   public static final String DEFAULT_STATE_MODEL_DEF = LeaderStandbySMD.name;
   public static final String AMBRY_STATE_MODEL_DEF = "AmbryLeaderStandby";
   private static final String MAX_REPLICAS_ALL_DATACENTERS = "max-replicas-all-datacenters";
+  private static final String DEFAULT_VCR_DATACENTER_NAME = "ei4";
 
   /**
    * The factory class used to get the resource state policies.
@@ -182,6 +183,16 @@ public class ClusterMapConfig {
   @Default("true")
   public final boolean clustermapListenCrossColo;
 
+  /**
+   * Name of the datacenter of vcrnodes. It is expected that all the vcr nodes will reside in the same datacenter.
+   */
+  @Config("clustermap.vcr.datacenter.name")
+  @Default(DEFAULT_VCR_DATACENTER_NAME)
+  public final String clustermapVcrDatacenterName;
+
+  /**
+   * State model definition to register with helix cluster.
+   */
   @Config("clustermap.state.model.definition")
   @Default(DEFAULT_STATE_MODEL_DEF)
   public final String clustermapStateModelDefinition;
@@ -220,6 +231,8 @@ public class ClusterMapConfig {
     clustermapListenCrossColo = verifiableProperties.getBoolean("clustermap.listen.cross.colo", true);
     clustermapStateModelDefinition =
         verifiableProperties.getString("clustermap.state.model.definition", DEFAULT_STATE_MODEL_DEF);
+    clustermapVcrDatacenterName =
+        verifiableProperties.getString("clustermap.vcr.datacenter.name", DEFAULT_VCR_DATACENTER_NAME);
     if (!clustermapStateModelDefinition.equals(DEFAULT_STATE_MODEL_DEF) && !clustermapStateModelDefinition.equals(
         AMBRY_STATE_MODEL_DEF)) {
       throw new IllegalArgumentException("Unsupported state model definition: " + clustermapStateModelDefinition);
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -28,10 +28,13 @@ public class AmbryPartitionStateModel extends StateModel {
   private Logger logger = LoggerFactory.getLogger(getClass());
   private final String resourceName;
   private final String partitionName;
+  private final PartitionStateChangeListener partitionStateChangeListener;
 
-  AmbryPartitionStateModel(String resourceName, String partitionName) {
+  AmbryPartitionStateModel(String resourceName, String partitionName,
+      PartitionStateChangeListener partitionStateChangeListener) {
     this.resourceName = resourceName;
     this.partitionName = partitionName;
+    this.partitionStateChangeListener = partitionStateChangeListener;
     StateModelParser parser = new StateModelParser();
     _currentState = parser.getInitialState(DefaultLeaderStandbyStateModel.class);
   }
@@ -52,12 +55,14 @@ public class AmbryPartitionStateModel extends StateModel {
   public void onBecomeLeaderFromStandby(Message message, NotificationContext context) {
     logger.info("Partition {} in resource {} is becoming LEADER from STANDBY", message.getPartitionName(),
         message.getResourceName());
+    partitionStateChangeListener.onPartitionLeadFromStandby(message.getPartitionName());
   }
 
   @Transition(to = "STANDBY", from = "LEADER")
   public void onBecomeStandbyFromLeader(Message message, NotificationContext context) {
     logger.info("Partition {} in resource {} is becoming STANDBY from LEADER", message.getPartitionName(),
         message.getResourceName());
+    partitionStateChangeListener.onPartitionStandbyFromLead(message.getPartitionName());
   }
 
   @Transition(to = "INACTIVE", from = "STANDBY")
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -13,6 +13,7 @@
  */
 package com.github.ambry.clustermap;
 
+import java.util.Objects;
 import org.apache.helix.NotificationContext;
 import org.apache.helix.model.Message;
 import org.apache.helix.participant.statemachine.StateModel;
@@ -34,10 +35,7 @@ public class AmbryPartitionStateModel extends StateModel {
       PartitionStateChangeListener partitionStateChangeListener) {
     this.resourceName = resourceName;
     this.partitionName = partitionName;
-    if(partitionStateChangeListener == null) {
-      throw new IllegalArgumentException("partition state change listener cannot be null");
-    }
-    this.partitionStateChangeListener = partitionStateChangeListener;
+    this.partitionStateChangeListener = Objects.requireNonNull(partitionStateChangeListener);
     StateModelParser parser = new StateModelParser();
     _currentState = parser.getInitialState(DefaultLeaderStandbyStateModel.class);
   }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -34,6 +34,9 @@ public class AmbryPartitionStateModel extends StateModel {
       PartitionStateChangeListener partitionStateChangeListener) {
     this.resourceName = resourceName;
     this.partitionName = partitionName;
+    if(partitionStateChangeListener == null) {
+      throw new IllegalArgumentException("partition state change listener cannot be null");
+    }
     this.partitionStateChangeListener = partitionStateChangeListener;
     StateModelParser parser = new StateModelParser();
     _currentState = parser.getInitialState(DefaultLeaderStandbyStateModel.class);
@@ -55,14 +58,14 @@ public class AmbryPartitionStateModel extends StateModel {
   public void onBecomeLeaderFromStandby(Message message, NotificationContext context) {
     logger.info("Partition {} in resource {} is becoming LEADER from STANDBY", message.getPartitionName(),
         message.getResourceName());
-    partitionStateChangeListener.onPartitionLeadFromStandby(message.getPartitionName());
+    partitionStateChangeListener.onPartitionStateChangeToLeaderFromStandby(message.getPartitionName());
   }
 
   @Transition(to = "STANDBY", from = "LEADER")
   public void onBecomeStandbyFromLeader(Message message, NotificationContext context) {
     logger.info("Partition {} in resource {} is becoming STANDBY from LEADER", message.getPartitionName(),
         message.getResourceName());
-    partitionStateChangeListener.onPartitionStandbyFromLead(message.getPartitionName());
+    partitionStateChangeListener.onPartitionStateChangeToStandbyFromLeader(message.getPartitionName());
   }
 
   @Transition(to = "INACTIVE", from = "STANDBY")
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectatorFactory.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectatorFactory.java
@@ -21,9 +21,9 @@ import com.github.ambry.config.ClusterMapConfig;
  * A factory to create {@link ClusterSpectator} object.
  */
 public class HelixClusterSpectatorFactory implements ClusterSpectatorFactory {
-  
+
   @Override
   public ClusterSpectator getClusterSpectator(CloudConfig cloudConfig, ClusterMapConfig clusterMapConfig) {
-    return new VcrHelixClusterSpectator(cloudConfig, clusterMapConfig);
+    return new HelixClusterSpectator(cloudConfig, clusterMapConfig);
   }
 }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectatorFactory.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectatorFactory.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2019 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.clustermap;
+
+import com.github.ambry.config.CloudConfig;
+import com.github.ambry.config.ClusterMapConfig;
+
+
+/**
+ * A factory to create {@link ClusterSpectator} object.
+ */
+public class HelixClusterSpectatorFactory implements ClusterSpectatorFactory {
+  
+  @Override
+  public ClusterSpectator getClusterSpectator(CloudConfig cloudConfig, ClusterMapConfig clusterMapConfig) {
+    return new VcrHelixClusterSpectator(cloudConfig, clusterMapConfig);
+  }
+}
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
@@ -275,16 +275,16 @@ class HelixParticipant implements ClusterParticipant, PartitionStateChangeListen
   }
 
   @Override
-  public void onPartitionLeadFromStandby(String partitionName) {
+  public void onPartitionStateChangeToLeaderFromStandby(String partitionName) {
     for (PartitionStateChangeListener partitionStateChangeListener : partitionStateChangeListeners) {
-      partitionStateChangeListener.onPartitionLeadFromStandby(partitionName);
+      partitionStateChangeListener.onPartitionStateChangeToLeaderFromStandby(partitionName);
     }
   }
 
   @Override
-  public void onPartitionStandbyFromLead(String partitionName) {
+  public void onPartitionStateChangeToStandbyFromLeader(String partitionName) {
     for (PartitionStateChangeListener partitionStateChangeListener : partitionStateChangeListeners) {
-      partitionStateChangeListener.onPartitionLeadFromStandby(partitionName);
+      partitionStateChangeListener.onPartitionStateChangeToStandbyFromLeader(partitionName);
     }
   }
 }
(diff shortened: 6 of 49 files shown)


ambry-server/src/main/java/com.github.ambry.server/AmbryServer.java
```diff
@@ -156,6 +163,16 @@ public void startup() throws InstantiationException {
...
       replicationManager.start();
 
+      if (replicationConfig.isVcrClusterReplicationEnabled) {
+        vcrClusterSpectator = new HelixClusterSpectatorFactory().getClusterSpectator(cloudConfig, clusterMapConfig);
+        cloudToStoreReplicationManager =
+            new CloudToStoreReplicationManager(replicationConfig, clusterMapConfig, storeConfig, storageManager,
+                storeKeyFactory, clusterMap, scheduler, nodeId, connectionPool, registry, notificationSystem,
+                storeKeyConverterFactory, serverConfig.serverMessageTransformer, vcrClusterSpectator,
+                clusterParticipant);
+        cloudToStoreReplicationManager.start();
```
lightningrob: Need to shut this down too.

ambry-server/src/main/java/com.github.ambry.server/AmbryServer.java
```diff
@@ -156,6 +163,16 @@ public void startup() throws InstantiationException {
               serverConfig.serverMessageTransformer);
       replicationManager.start();
 
+      if (replicationConfig.isVcrClusterReplicationEnabled) {
+        vcrClusterSpectator = new HelixClusterSpectatorFactory().getClusterSpectator(cloudConfig, clusterMapConfig);
```
lightningrob: logger.info something here.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+    this.vcrNodes = new ConcurrentSkipListSet<>();
+    this.persistor =
+        new DiskTokenPersistor(cloudReplicaTokenFileName, mountPathToPartitionInfos, replicationMetrics, clusterMap,
+            tokenHelper);
+    this.localPartitionNameToPartition = mapPartitionNameToPartition(clusterMap, currentNode);
+  }
+
+  private ConcurrentHashMap<String, PartitionId> mapPartitionNameToPartition(ClusterMap clusterMap,
+      DataNodeId localNode) {
+    List<? extends ReplicaId> localReplicas = clusterMap.getReplicaIds(localNode);
```
lightningrob: Can fold this line into next one.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+        synchronized (notificationLock) {
+          for (LiveInstance liveInstance : liveInstances) {
+            if (instanceNameToCloudDataNode.containsKey(liveInstance.getInstanceName())) {
+              newVcrNodes.add(instanceNameToCloudDataNode.get(liveInstance.getInstanceName()));
+            }
+          }
+          Set<CloudDataNode> removedNodes = new HashSet<>(vcrNodes);
+          removedNodes.removeAll(newVcrNodes);
+          vcrNodes = newVcrNodes;
+          List<PartitionId> partitionsOnRemovedNodes = getPartitionsOnRemovedNodes(removedNodes);
```
lightningrob: This section looks identical or very similar to one in previous method.  Consolidate?

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+  public void start() throws ReplicationException {
+    // Add listener for vcr instance config changes
+    vcrClusterSpectator.registerInstanceConfigChangeListener(new InstanceConfigChangeListener() {
+      @Override
+      public void onInstanceConfigChange(List<InstanceConfig> instanceConfigs, NotificationContext context) {
+        ConcurrentSkipListSet<CloudDataNode> newVcrNodes = new ConcurrentSkipListSet<>();
+        ConcurrentHashMap<String, CloudDataNode> newInstanceNameToCloudDataNode = new ConcurrentHashMap<>();
+        Set<CloudDataNode> removedNodes = new HashSet<>(vcrNodes);
+        synchronized (notificationLock) {
+          for (InstanceConfig instanceConfig : instanceConfigs) {
```
lightningrob: Please add some comments here, including why you need to remove and then add the same partitions.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+      }
+    });
+
+    // start background persistent thread
+    // start scheduler thread to persist index in the background
+    scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
+        replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
+  }
+
+  private List<PartitionId> getPartitionsOnRemovedNodes(Set<CloudDataNode> removedNodes) {
```
lightningrob: javadoc

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+  }
+
+  /**
+   * Remove a replica of given partition and its {@link RemoteReplicaInfo}s from the backup list.
+   * @param partitionName the partition of the replica to removed.
+   * @throws ReplicationException if replicas initialization failed.
+   */
+  private void removeCloudReplica(String partitionName) throws ReplicationException {
+    if (!localPartitionNameToPartition.containsKey(partitionName)) {
+      logger.error("Got partition standby notification for partition {} that is not present on the node",
```
lightningrob: Just curious why you only log here but throw exception in two other places.  Should it be a warn instead?

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+          partitionName);
+      return;
+    }
+    PartitionId partitionId = localPartitionNameToPartition.get(partitionName);
+    PartitionInfo partitionInfo = partitionToPartitionInfo.remove(partitionId);
+    if (partitionInfo == null) {
+      logger.error("Partition {} not exist when remove from {}. ", partitionId, dataNodeId);
+      throw new ReplicationException("Partition not found");
```
lightningrob: Combine full error message into exception and remove log.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+    removeRemoteReplicaInfoFromReplicaThread(partitionInfo.getRemoteReplicaInfos());
+    if (replicationConfig.replicationPersistTokenOnShutdownOrReplicaRemove) {
+      try {
+        persistor.write(partitionInfo.getLocalReplicaId().getMountPath(), false);
+      } catch (IOException | ReplicationException e) {
+        logger.error("Exception on token write when remove Partition {} from {}: ", partitionId, dataNodeId, e);
+        throw new ReplicationException("Exception on token write.");
```
lightningrob: Here too.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/CloudReplicaTest.java
```diff
@@ -57,7 +55,7 @@ public void basicTest() {
     CloudReplica cloudReplica = new CloudReplica(new MockPartitionId(), cloudDataNode);
     assertEquals("Wrong mount path", mockPartitionId.toPathString(), cloudReplica.getMountPath());
     assertEquals("Wrong replica path",
-        GetRequest.Cloud_Replica_Keyword + File.separator + mockPartitionId.toPathString() + File.separator
+        CloudReplica.Cloud_Replica_Keyword + File.separator + mockPartitionId.toPathString() + File.separator
```
lightningrob: Remove the constant from GetRequest too.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/VcrHelixClusterSpectator.java
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Spectator for vcr helix cluster.
+ */
+public class VcrHelixClusterSpectator implements ClusterSpectator {
+
+  private final ClusterMapConfig clusterMapConfig;
+  private final CloudConfig cloudConfig;
+  private List<InstanceConfigChangeListener> registeredInstanceConfigChangeListeners;
+  private List<LiveInstanceChangeListener> registeredLiveInstanceChangeListeners;
```
lightningrob: Can't these two also be final?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterAgentsFactory.java
```diff
@@ -75,9 +76,15 @@ public StaticClusterManager getClusterMap() {
...
     if (clusterParticipant == null) {
       clusterParticipant = new ClusterParticipant() {
+        private final List<PartitionStateChangeListener> listeners = new LinkedList<>();
+
         @Override
         public void participate(List<AmbryHealthReport> ambryHealthReports) {
-
+          for (PartitionStateChangeListener listener : listeners) {
+            for (String partitionName : partitionLayout.getAllPartitionNames()) {
+              listener.onPartitionLeadFromStandby(partitionName);
```
lightningrob: Do we know that it was a standby before participate was called?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectatorFactory.java
```diff
@@ -0,0 +1,29 @@
...
+
+
+/**
+ * A factory to create {@link ClusterSpectator} object.
+ */
+public class HelixClusterSpectatorFactory implements ClusterSpectatorFactory {
+  
+  @Override
+  public ClusterSpectator getClusterSpectator(CloudConfig cloudConfig, ClusterMapConfig clusterMapConfig) {
+    return new VcrHelixClusterSpectator(cloudConfig, clusterMapConfig);
```
lightningrob: Name inconsistency: HelixClusterSpectatorFactory should create HelixClusterSpectator.

ambry-api/src/main/java/com.github.ambry/clustermap/ClusterSpectator.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * {@link ClusterSpectator} adds itself as a spectator of a cluster and registers some cluster change listeners.
+ */
+public interface ClusterSpectator extends InstanceConfigChangeListener, LiveInstanceChangeListener {
+
+  /**
+   * Register with helix cluster as a spectator.
+   */
+  void spectate() throws Exception;
```
jsjtzyy: I checked `VcrHelixClusterSpector`, it seems this method is more like a `start()` method that can be done in the constructor. For other implementations in the future, I wonder if this method is necessary. Maybe we can simplify this just like `ClusterChangeHandler` in `HelixClusterManager`.

ambry-api/src/main/java/com.github.ambry/clustermap/PartitionStateChangeListener.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * {@link PartitionStateChangeListener} takes action when partition state changes.
+ */
+public interface PartitionStateChangeListener {
+
+  /**
+   * Action to take on becoming leader of a partition.
+   * @param partitionName of the partition.
+   */
+  void onPartitionLeadFromStandby(String partitionName);
```
jsjtzyy: Could you change the name to `onPartitionBecomeLeaderFromStandby` ? I feel like this is more accurate.

ambry-api/src/main/java/com.github.ambry/clustermap/PartitionStateChangeListener.java
```diff
@@ -0,0 +1,32 @@
...
+
+  /**
+   * Action to take on being removed as leader of a partition.
+   * @param partitionName of the partition.
+   */
+  void onPartitionStandbyFromLead(String partitionName);
```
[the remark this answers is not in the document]
jsjtzyy: Same here, better to rename it to `onPartitionBecomeStandbyFromLeader`.
@lightningrob there might be some ambiguity between participant and partition. For partition, yes, it may have other transitions between states we defined. I will add rest of methods in this class because `PartitionStateChangeListener` will be used for "Move Replica" as well. We need to add/remove replicas on current node when state transition occurs from OFFLINE to BOOTSTRAP or from STANDBY to INACTIVE. 

Participant is basically datanode, it has multiple partitions and they may stay in different states. The state model applies to partition only.

ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
```diff
@@ -182,6 +183,16 @@
   @Default("true")
   public final boolean clustermapListenCrossColo;
 
+  /**
+   * Name of the datacenter of vcrnodes. It is expected that all the vcr nodes will reside in the same datacenter.
```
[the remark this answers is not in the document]
ankagrawal: No it wont be just the local one in that case. However, as previously discussed, we dcInfo isnt added to vcrclustermap in helix right now. So until we do that, this is a hack to get the datacenter name for vcr nodes.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -265,4 +273,18 @@ boolean setStoppedReplicas(List<String> stoppedReplicas) {
     instanceConfig.getRecord().setListField(ClusterMapUtils.STOPPED_REPLICAS_STR, stoppedReplicas);
     return helixAdmin.setInstanceConfig(clusterName, instanceName, instanceConfig);
   }
+
+  @Override
+  public void onPartitionLeadFromStandby(String partitionName) {
+    for (PartitionStateChangeListener partitionStateChangeListener : partitionStateChangeListeners) {
+      partitionStateChangeListener.onPartitionLeadFromStandby(partitionName);
```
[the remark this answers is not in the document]
ankagrawal: Thats true. As discussed offline, for enforcing ordering maybe we can define specific listeners (like storagelistener, replicationlistener) instead of defining a list of listeners. Can we consider this out of scope for this ticket?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterAgentsFactory.java
ankagrawal: This is the case for static cluster map. There will be no event for either standby from leader, or leader from standby in this case. The logic above just forces callbacks simulating adding partition leader. For this we don't care if partition was a standby or not.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/VcrHelixClusterSpectator.java
ankagrawal: will fix.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/VcrHelixClusterSpectator.java
```diff
@@ -0,0 +1,96 @@
...
+    registeredInstanceConfigChangeListeners = new LinkedList<>();
+    registeredLiveInstanceChangeListeners = new LinkedList<>();
+  }
+
+  @Override
+  public void spectate() throws Exception {
+    Map<String, DcZkInfo> dataCenterToZkAddress =
+        parseDcJsonAndPopulateDcInfo(clusterMapConfig.clusterMapDcsZkConnectStrings);
+    HelixFactory helixFactory = new HelixFactory();
+    String selfInstanceName = ClusterMapUtils.getInstanceName(clusterMapConfig.clusterMapHostName, 12111);
```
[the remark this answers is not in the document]
ankagrawal: yeah this shouldnt be hardcoded. Will fix.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/VcrHelixClusterSpectator.java
```diff
@@ -0,0 +1,96 @@
...
+
+    for (Map.Entry<String, DcZkInfo> entry : dataCenterToZkAddress.entrySet()) {
+      String zkConnectStr = entry.getValue().getZkConnectStr();
+      HelixManager helixManager =
+          helixFactory.getZKHelixManager(cloudConfig.vcrClusterName, selfInstanceName, InstanceType.SPECTATOR,
+              zkConnectStr);
+      helixManager.connect();
```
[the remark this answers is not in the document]
ankagrawal: We might need to think more here. In case of LI Ambry servers, we know that we have more than two datacenters. In case of cloud(azure) we are starting with just one fabric. If we add another fabric in future, such that there are only two fabrics, then I don't know if we should pass the startup with remote zk connection failure. Because if remote zk connection fails on both fabrics, then replication between fabrics might just stop.
For now, since we have only one fabric this will work. Once we add more fabrics, we should revisit this. Adding a comment in the code.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/CloudReplicaTest.java
ankagrawal: will do.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+ * {@link CloudToStoreReplicationManager} replicates data from Vcr nodes to ambry data nodes.
+ */
+public class CloudToStoreReplicationManager extends ReplicationEngine {
+  private final ClusterMapConfig clusterMapConfig;
+  private final StoreConfig storeConfig;
+  private final StoreManager storeManager;
+  private final ClusterSpectator vcrClusterSpectator;
+  private final ClusterParticipant clusterParticipant;
+  private static final String cloudReplicaTokenFileName = "cloudReplicaTokens";
+  private ConcurrentHashMap<String, CloudDataNode> instanceNameToCloudDataNode;
```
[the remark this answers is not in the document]
ankagrawal: will fix.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,333 @@
...
+  private ConcurrentSkipListSet<CloudDataNode> vcrNodes;
```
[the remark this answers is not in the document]
ankagrawal: will fix.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
ankagrawal: will fix.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
ankagrawal: will fix.

ambry-server/src/main/java/com.github.ambry.server/AmbryServer.java
ankagrawal: will fix.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
```diff
@@ -28,10 +28,13 @@
...
   private final String resourceName;
   private final String partitionName;
+  private final PartitionStateChangeListener partitionStateChangeListener;
 
-  AmbryPartitionStateModel(String resourceName, String partitionName) {
+  AmbryPartitionStateModel(String resourceName, String partitionName,
+      PartitionStateChangeListener partitionStateChangeListener) {
     this.resourceName = resourceName;
     this.partitionName = partitionName;
+    this.partitionStateChangeListener = partitionStateChangeListener;
```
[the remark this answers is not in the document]
lightningrob: Our practice is:
this.partitionStateChangeListener Objects.requireNonNull(partitionStateChangeListener, errorMsg);
Which throws a NPE instead of IAE, but we had discussion a while back and agreed on this.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
ankagrawal: fixed.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
ankagrawal: fixed logger.error to logger.warn where I am not throwing exception.
Also, I have changed the semantics about throwing Exception on remove such that if partition being removed doesn't exist in the data structures, we add a warning and return without throwing exception.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
ankagrawal: fixed.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
ankagrawal: fixed.

ambry-server/src/main/java/com.github.ambry.server/AmbryServer.java
ankagrawal: fixed.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,337 @@
...
+  private class PartitionStateChangeListenerImpl implements PartitionStateChangeListener {
+    @Override
+    public void onPartitionStateChangeToLeaderFromStandby(String partitionName) {
+      logger.debug("Partition state change notification from Standby to Leader received for partition {}",
+          partitionName);
+      synchronized (notificationLock) {
+        try {
+          addCloudReplica(partitionName);
+        } catch (ReplicationException rex) {
+          logger.error("Could not add replication for paritition {}", partitionName);
```
ankagrawal: NEED FEEDBACK: Note that here and for other listeners, I log an error in case the listener threw an exception, instead of throwing an exception here. With this approach, if anything fails, then cloud replication from only one partition or replica gets affected, and the notifications can keep coming for future change notifications. Let me know if you think this reasoning looks ok.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -0,0 +1,337 @@
...
+    }
+  }
+
+  /**
+   * {@link PartitionStateChangeListener} to capture changes in partition state.
+   */
+  private class PartitionStateChangeListenerImpl implements PartitionStateChangeListener {
+    @Override
+    public void onPartitionStateChangeToLeaderFromStandby(String partitionName) {
+      logger.debug("Partition state change notification from Standby to Leader received for partition {}",
```
jsjtzyy: suggest changing to INFO level

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
jsjtzyy: Do you know in which case, it may throw exception?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationEngine.java
```diff
@@ -400,4 +400,61 @@ protected boolean isTokenForRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInf
...
+    if (partitionInfo == null) {
+      logger.warn("Partition {} not exist when remove from {}. ", partitionId, dataNodeId);
+      return;
+    }
+    removeRemoteReplicaInfoFromReplicaThread(partitionInfo.getRemoteReplicaInfos());
+    if (replicationConfig.replicationPersistTokenOnShutdownOrReplicaRemove) {
+      try {
+        persistor.write(partitionInfo.getLocalReplicaId().getMountPath(), false);
+      } catch (IOException | ReplicationException e) {
+        throw new ReplicationException(
```
lightningrob: I don't think failure to write token should fail the stopping of replication, which can prevent a new replica from being added.  Just log the error and continue.  Some records may get reprocessed, but that is okay.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
lightningrob: I think it's okay since it's a listener, but will the partition no longer be replicated from cloud?  Maybe we should add the partition to an error list and retry the failed ones later?  And/or update an error metric.

jsjtzyy: ok, let's start with a metric here and implement possible re-try logic in future.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -310,8 +310,7 @@ public void emptyPartitionOverrideTest() throws Exception {
 
     Set<String> writableInClusterManager = new HashSet<>();
     for (PartitionId partition : clusterManagerWithEmptyRecord.getWritablePartitionIds(null)) {
-      String partitionStr =
-          useComposite ? ((Partition) partition).toPathString() : ((AmbryPartition) partition).toPathString();
+      String partitionStr = useComposite ? partition.toPathString() : partition.toPathString();
```
jsjtzyy: minor: looks like it can be simplified as `String partitionStr = partition.toPathString()`

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -1043,8 +1042,7 @@ private long getGaugeValue(String suffix) {
   private Pair<Set<String>, Set<String>> getWritablePartitions() {
     Set<String> writableInClusterManager = new HashSet<>();
     for (PartitionId partition : clusterManager.getWritablePartitionIds(null)) {
-      String partitionStr =
-          useComposite ? ((Partition) partition).toPathString() : ((AmbryPartition) partition).toPathString();
+      String partitionStr = useComposite ? partition.toPathString() : partition.toPathString();
```
jsjtzyy: same here

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
```diff
@@ -1061,8 +1059,7 @@ private long getGaugeValue(String suffix) {
   private void testAllPartitions() {
     Set<String> partitionsInClusterManager = new HashSet<>();
     for (PartitionId partition : clusterManager.getAllPartitionIds(null)) {
-      String partitionStr =
-          useComposite ? ((Partition) partition).toPathString() : ((AmbryPartition) partition).toPathString();
+      String partitionStr = useComposite ? partition.toPathString() : partition.toPathString();
```
jsjtzyy: same here

ambry-commons/src/main/java/com.github.ambry.commons/ResponseHandler.java
```diff
@@ -109,6 +110,8 @@ private void onException(ReplicaId replicaId, Exception e) {
    * {@link ServerErrorCode}.
    */
   public void onEvent(ReplicaId replicaId, Object event) {
+    if(replicaId instanceof CloudReplica)
+      return;
```
jsjtzyy: There is no backoff when replication with cloudReplica has an exception? (I guess it is based on Helix will notify which is next available cloud replica)

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
ankagrawal: fixed.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
ankagrawal: fixed.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/HelixClusterManagerTest.java
ankagrawal: fixed.

ambry-commons/src/main/java/com.github.ambry.commons/ResponseHandler.java
ankagrawal: Yes.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationEngine.java
```diff
@@ -400,4 +400,60 @@ protected boolean isTokenForRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInf
...
+    if (partitionInfo == null) {
+      logger.warn("Partition {} not exist when remove from {}. ", partitionId, dataNodeId);
+      return;
+    }
+    removeRemoteReplicaInfoFromReplicaThread(partitionInfo.getRemoteReplicaInfos());
+    if (replicationConfig.replicationPersistTokenOnShutdownOrReplicaRemove) {
+      try {
+        persistor.write(partitionInfo.getLocalReplicaId().getMountPath(), false);
+      } catch (IOException | ReplicationException e) {
+        logger.warn(
```
lightningrob: Minor: exceptions like this should still be logged as error, even if we're not rethrowing.

ambry-server/src/integration-test/java/com.github.ambry.server/CloudAndStoreReplicationTest.java
```diff
@@ -0,0 +1,325 @@
...
+  private MockDataNodeId partitionLeaderRecoveryNode;
+  private List<MockDataNodeId> allRecoveryNodes;
+  private VcrServer vcrServer;
+  private HelixControllerManager helixControllerManager;
+  private TestUtils.ZkInfo zkInfo;
+  private PartitionId partitionId;
+  private LatchBasedInMemoryCloudDestination latchBasedInMemoryCloudDestination;
+  private List<BlobId> cloudBlobIds;
+  private List<BlobId> serverBlobIds;
+  final private short accountId = Utils.getRandomShort(TestUtils.RANDOM);
```
lightningrob: Minor: private final.

ambry-server/src/integration-test/java/com.github.ambry.server/CloudAndStoreReplicationTest.java
```diff
@@ -0,0 +1,325 @@
...
+    recoveryProperties.setProperty("replication.enabled.with.vcr.cluster", "true");
+    recoveryProperties.setProperty("clustermap.vcr.datacenter.name", cloudDc);
+
+    // create vcr node
+    List<Port> vcrPortList = new ArrayList<>(2);
+    Port vcrClusterMapPort = new Port(12310, PortType.PLAINTEXT);
+    Port vcrSslPort = new Port(12410, PortType.SSL);
+    vcrPortList.add(vcrClusterMapPort);
+    vcrPortList.add(vcrSslPort);
+
```
lightningrob: This can all be one line via Arrays.toList() or similar.

ambry-server/src/integration-test/java/com.github.ambry.server/MockCluster.java
```diff
@@ -221,8 +251,44 @@ private void initializeServer(DataNodeId dataNodeId, Properties sslProperties, b
...
+   * @param enableHardDeletes {@code enableHardDeletes} flag.
+   * @param enableDataPrefetch {@code enableDataPrefetch} flag.
+   * @param notificationSystem {@link NotificationSystem} object.
+   * @param time {@link Time} object.
+   * @param mockClusterAgentsFactory {@link MockClusterAgentsFactory} object.
+   * @return {@link VerifiableProperties} object.
+   */
+  public void initializeServer(DataNodeId dataNodeId, Properties sslProperties, boolean enableHardDeletes,
+      boolean enableDataPrefetch, NotificationSystem notificationSystem, Time time,
+      MockClusterAgentsFactory mockClusterAgentsFactory) {
```
lightningrob: I don't think we need two constructors here since they both construct AmbryServer the same way.  Just have one that takes the MockClusterAgentsFactory arg, and if null is passed, use the class default.

ambry-server/src/integration-test/java/com.github.ambry.server/CloudAndStoreReplicationTest.java
```diff
@@ -0,0 +1,325 @@
...
+  }
+
+  /**
+   * Do a get on recovery server node to test that all the blobids that were uploaded to vcr node have been recovered on
+   * recovery node.
+   * @param blobIdToSizeMap {@link Map} of blobid to size uploaded to vcr node.
+   * @param node recovery server node
+   * @throws IOException on {@link IOException}
+   */
+  private void testGetOnServerNode(Map<BlobId, Integer> blobIdToSizeMap, DataNodeId node) throws IOException {
```
lightningrob: Move private methods to below public tests.

ambry-server/src/integration-test/java/com.github.ambry.server/CloudAndStoreReplicationTest.java
```diff
@@ -0,0 +1,325 @@
...
+            notificationSystem, cloudDestinationFactory);
+    vcrServer.startup();
+
+    // initialize and start ambry servers
+    for (MockDataNodeId serverNode : allRecoveryNodes) {
+      if (serverNode.equals(partitionLeaderRecoveryNode)) {
+        recoveryCluster.initializeServer(serverNode, recoveryProperties, false, true, notificationSystem,
+            SystemTime.getInstance(), leaderMockClusterAgentsFactory);
+      } else {
+        recoveryCluster.initializeServer(serverNode, recoveryProperties, false, true, notificationSystem,
```
lightningrob: After consolidating the two initializeServer methods, you can consolidate this into one call passing:
    serverNode.equals(partitionLeaderRecoveryNode) ? leaderMockClusterAgentsFactory : null

lightningrob:
The CloudAndStoreReplicationTest look very good.  It would be great to be able to run a longer version of it (with more blobs and operation types) against an actual Azure destination.

ambry-server/src/integration-test/java/com.github.ambry.server/MockCluster.java
```diff
@@ -221,8 +251,31 @@ private void initializeServer(DataNodeId dataNodeId, Properties sslProperties, b
...
+    AmbryServer server;
+    if (mockClusterAgentsFactory != null) {
+      server = new AmbryServer(createInitProperties(dataNodeId, enableDataPrefetch, enableHardDeletes, sslProperties),
+          mockClusterAgentsFactory, mockClusterSpectatorFactory, notificationSystem, time);
+    } else {
+      server = new AmbryServer(createInitProperties(dataNodeId, enableDataPrefetch, enableHardDeletes, sslProperties),
+          this.mockClusterAgentsFactory, mockClusterSpectatorFactory, notificationSystem, time);
+    }
```
lightningrob: Cleaner to say:
MockClusterAgentsFactory factory = mockClusterAgentsFactory != null ? mockClusterAgentsFactory : this.mockClusterAgentsFactory;
and then pass factory to the AmbryServer constructor.

PR #1294, opened 2019-10-29
Add netty ByteBuf to network layer

justinlin-linkedin:
Adding netty ByteBuf to network layer and add a new config option to enable or disable it.


ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,129 @@
...
+    return bytesRead;
+  }
+
+  /**
+   * Returns the payload as {@link ByteBuf}, at the same time release the current reference to this payload.
+   * It's not safe to call this function again.
+   * @return
+   */
+  @Override
+  public ByteBuf getAndRelease() {
```
cgtz: Should this method return null if the buffer was not yet created or the buffer was released because of an error? Right now, it may throw an exception upon calling `retainedDuplicate()`

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1662,6 +1664,15 @@ static ByteBufferInputStream getByteBufferInputStreamForBlobRecord(CrcInputStrea
       byteBuffer.limit(oldLimit);
       // Change the byte buffer's position as if the data is fetched.
       byteBuffer.position(startIndex + (int) dataSize);
+    } else if (inputStream instanceof NettyByteBufDataInputStream ) {
+      // getBuffer function doesn't increase the reference count on this ByteBuf.
+      ByteBuf nettyByteBuf = ((NettyByteBufDataInputStream)inputStream).getBuffer();
+      // construct a java.nio.ByteBuf to create a ByteBufferInputStream
+      int startIndex = nettyByteBuf.readerIndex();
+      ByteBuffer dataBuffer = nettyByteBuf.nioBuffer(startIndex, (int)dataSize);
```
cgtz: nit: add space between cast and variable: `(int) dataSize`

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1662,6 +1664,15 @@ static ByteBufferInputStream getByteBufferInputStreamForBlobRecord(CrcInputStrea
       byteBuffer.limit(oldLimit);
       // Change the byte buffer's position as if the data is fetched.
       byteBuffer.position(startIndex + (int) dataSize);
+    } else if (inputStream instanceof NettyByteBufDataInputStream ) {
```
cgtz: Could you add a check to this method to ensure that `dataSize` is within the int range? alternatively, maybe we change the dataSize parameter to an int since every part of this method requires it to be so.

ambry-api/src/main/java/com.github.ambry/network/BoundedByteBufferReceive.java
```diff
@@ -73,7 +73,8 @@ public long readFrom(ReadableByteChannel channel) throws IOException {
     return bytesRead;
   }
 
-  public ByteBuffer getPayload() {
+  @Override
+  public ByteBuffer getAndRelease() {
```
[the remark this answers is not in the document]
cgtz: Also, it might be good to keep the call once semantics the same between the `ByteBuf` and `ByteBuffer` implementations

ambry-network/src/main/java/com.github.ambry.network/ResponseInfo.java
```diff
@@ -63,10 +63,16 @@ public NetworkClientErrorCode getError() {
   /**
    * @return the response received for this request.
    */
-  public ByteBuffer getResponse() {
+  public Object getResponse() {
```
cgtz: Is there an opportunity to use generics here? It's ok if it's not possible. I realize it may not be easily done since the buffer type is determined by a runtime config.

ambry-network/src/test/java/com.github.ambry.network/BoundedByteBufferReceiveTest.java
```diff
@@ -40,9 +40,9 @@ public void testBoundedByteBufferReceive() throws Exception {
...
         set.readFrom(Channels.newChannel(new ByteBufferInputStream(buffer))));
     buffer.clear();
-    ByteBuffer payload = set.getPayload();
+    ByteBuffer payload = set.getAndRelease();
     for (int i = 8; i < bufferSize; i++) {
       Assert.assertEquals(buffer.array()[i], payload.get());
     }
   }
-}
+}
```
cgtz: add newline at end of file

ambry-network/src/test/java/com.github.ambry.network/NetworkClientTest.java
```diff
@@ -108,6 +124,12 @@ public NetworkClientTest() throws IOException {
     replicaOnSslNode = sslEnabledClusterMap.getReplicaIds(dataNodeId).get(0);
   }
 
+  @After
+  public void cleanup() {
+    // force JVM to gc to trigger more intense resource leak detection.
+    System.gc();
```
cgtz: If there was a leak, would the netty leak detector print messages in the log?

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,129 @@
...
+import java.nio.channels.ReadableByteChannel;
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A netty {@link ByteBuf} version of Receive to buffer the incoming request or response.
+ */
+public class BoundedNettyByteBufReceive implements BoundedReceive<ByteBuf> {
```
zzmao: `BoundedByteBufReceive` is sufficient.

ambry-network/src/main/java/com.github.ambry.network/PlainTextTransmission.java
```diff
@@ -59,7 +59,7 @@ public boolean ready() {
   @Override
   public boolean read() throws IOException {
     if (!hasReceive()) {
-      networkReceive = new NetworkReceive(getConnectionId(), new BoundedByteBufferReceive(), time);
+      createNetworkReceive();
```
zzmao: prefer to have a return value for explicitly set.

ambry-network/src/main/java/com.github.ambry.network/Transmission.java
```diff
@@ -166,6 +176,9 @@ public boolean hasReceive() {
   }
 
   public void clearReceive() {
+    if (networkReceive != null) {
+      ReferenceCountUtil.release(networkReceive.getReceivedBytes().getAndRelease());
```
cgtz: would this cause exceptions upon calling `getAndRelease` in `NetworkClient`? I think clearReceive is called from within the selector and then the receive is unboxed again in`NetworkClient:sendAndPoll`

ambry-network/src/main/java/com.github.ambry.network/SocketServer.java
```diff
@@ -17,10 +17,10 @@
 import com.github.ambry.commons.SSLFactory;
 import com.github.ambry.config.NetworkConfig;
 import com.github.ambry.config.SSLConfig;
-import com.github.ambry.utils.ByteBufferInputStream;
 import com.github.ambry.utils.SystemTime;
 import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
+import io.netty.buffer.ByteBufInputStream;
```
zzmao: not used.

ambry-network/src/main/java/com.github.ambry.network/NetworkClient.java
```diff
@@ -346,7 +345,7 @@ private void handleSelectorEvents(List<ResponseInfo> responseInfoList) {
       connectionTracker.checkInConnection(connId);
       RequestMetadata requestMetadata = connectionIdToRequestInFlight.remove(connId);
       correlationIdInFlightToConnectionId.remove(requestMetadata.requestInfo.getRequest().getCorrelationId());
-      responseInfoList.add(new ResponseInfo(requestMetadata.requestInfo, null, recv.getReceivedBytes().getPayload()));
+      responseInfoList.add(new ResponseInfo(requestMetadata.requestInfo, null, recv.getReceivedBytes().getAndRelease()));
```
cgtz: this was the second call to getAndRelease i was referring to.

ambry-network/src/test/java/com.github.ambry.network/NetworkClientTest.java
justinlin-linkedin: Netty resource leak detector works in a weird way. First, it only prints out the error message in the log, without any exception. So it's not considered as an error, thus will not fail the test. It's a bit troubling because if the unit test doesn't fail, we probably not going to read the log even if there is a bytebuf leak. Second, It heavily relies on the garbage collection. resource leak detector only sees a ByteBuf leaking when this ByteBuf is GC-ed, but not released yet. So it's not so predictable.

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
justinlin-linkedin: I prefer adding Netty in the middle of the name, since the counterpart's name is BoundedByteBufferReceive, which is too close to BoundedByteBufReceive. Adding Netty in the middle of it, would better serve intellij as well as human eyes.

ambry-network/src/main/java/com.github.ambry.network/Transmission.java
justinlin-linkedin: you are right, I did a test in the morning and it does mess up the network layer. Update is on the way.

ambry-api/src/main/java/com.github.ambry/network/BoundedByteBufferReceive.java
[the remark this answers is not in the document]
justinlin-linkedin: Make sense. Will update.

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,129 @@
...
+  /**
+   * Reading bytes from the {@link ReadableByteChannel} and writing those bytes to the given {@link ByteBuf}.
+   * This method would change the writerIndex of the given buffer.
+   * @param channel The {@link ReadableByteChannel} to read the bytes out.
+   * @param buffer The {@link ByteBuf} to write bytes to.
+   * @return The number of bytes read from channel.
+   * @throws IOException Any I/O error.
+   */
+  private int readBytesFromReadableByteChannel(ReadableByteChannel channel, ByteBuf buffer) throws IOException {
+    int n = channel.read(buffer.internalNioBuffer(buffer.writerIndex(), buffer.capacity() - buffer.writerIndex()));
```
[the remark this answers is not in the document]
justinlin-linkedin: I suppose use nioBuffer is better than internalNioBuffer since it's for internal usage. And the reason this method is private and not exposed, is that I want to make sure all the ByteBuf it receives is not composite buffer and allocated where we want it to. We can make an public method to deal with more generic ByteBuf later.

ambry-api/src/main/java/com.github.ambry/network/BoundedByteBufferReceive.java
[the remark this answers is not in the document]
justinlin-linkedin: I will break this method to two methods get() and release().

ambry-network/src/main/java/com.github.ambry.network/PlainTextTransmission.java
justinlin-linkedin: The networkReceive is a private field. createNetReceive is a private method. I suppose we can create it in the method. Is it the method name? Maybe I can change the method name to something like "initializeNetworkReceive"?

ambry-network/src/main/java/com.github.ambry.network/ResponseInfo.java
justinlin-linkedin: Tried a couple ways to use generic, but not very successfully. They are either quite intrusive, or adding too many encapsulation.

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * A netty {@link ByteBuf} version of Receive to buffer the incoming request or response.
+ */
+public class BoundedNettyByteBufReceive implements BoundedReceive<ByteBuf> {
+
+  private ByteBuf buffer = null;
+  private ByteBuf sizeBuffer = null;
+  private long sizeToRead = 0;
+  private long sizeRead = 0;
+  private Logger logger = LoggerFactory.getLogger(getClass());
```
lightningrob: make static.

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,130 @@
...
+
+  @Override
+  public boolean isReadComplete() {
+    return !(buffer == null || sizeRead < sizeToRead);
```
lightningrob: Maybe clearer to use AND and no NOT.

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+    return n;
+  }
+
+  @Override
+  public long readFrom(ReadableByteChannel channel) throws IOException {
+    long bytesRead = 0;
+    if (buffer == null) {
+      if (sizeBuffer == null) {
+        sizeBuffer = ByteBufAllocator.DEFAULT.heapBuffer(8);
```
lightningrob: replace 8 with Long.BYTES

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,130 @@
...
+          throw new EOFException();
+        }
+      } catch (IOException e) {
+        sizeBuffer.release();
+        sizeBuffer = null;
+        throw e;
+      }
+      if (sizeBuffer.writerIndex() == sizeBuffer.capacity()) {
+        sizeToRead = sizeBuffer.readLong();
+        sizeRead += 8;
```
lightningrob: same here

ambry-api/src/main/java/com.github.ambry/network/BoundedNettyByteBufReceive.java
```diff
@@ -0,0 +1,130 @@
...
+      sizeRead += bytesReadFromChannel;
+      bytesRead += bytesReadFromChannel;
+    }
+    logger.trace("size read from channel {}", sizeRead);
+    return bytesRead;
+  }
+
+  /**
+   * Returns the payload as {@link ByteBuf}, at the same time release the current reference to this payload.
+   * It's not safe to call this function again.
```
lightningrob: again -> multiple times.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -1654,16 +1656,25 @@ static ByteBufferInputStream getByteBufferInputStreamForBlobRecord(CrcInputStrea
...
 
       output = new ByteBufferInputStream(dataBuffer);
       byteBuffer.limit(oldLimit);
       // Change the byte buffer's position as if the data is fetched.
-      byteBuffer.position(startIndex + (int) dataSize);
+      byteBuffer.position(startIndex + dataSize);
+    } else if (inputStream instanceof NettyByteBufDataInputStream) {
+      // getBuffer function doesn't increase the reference count on this ByteBuf.
+      ByteBuf nettyByteBuf = ((NettyByteBufDataInputStream) inputStream).getBuffer();
+      // construct a java.nio.ByteBuf to create a ByteBufferInputStream
```
lightningrob: ByteBuffer

ambry-router/src/test/java/com.github.ambry.router/NonBlockingRouterTest.java
```diff
@@ -42,6 +42,7 @@
 import com.github.ambry.utils.SystemTime;
 import com.github.ambry.utils.TestUtils;
 import com.github.ambry.utils.Utils;
+import io.netty.buffer.ByteBuf;
```
cgtz: is this import used?

cgtz: Also, are there any router tests where Netty `ByteBuf`s are used? It might be worth parameterizing this test to test both.

justinlin-linkedin: Router should always expecting a ByteBuffer instead of ByteBuf. I will introduce ByteBuf to router at next pr.

PR #1295, opened 2019-10-30
Use cosmos generated last updatetime for tracking cloud replica token, rather than last uploadTime

ankagrawal:
This PR adds the functionality to track cloud tokens using following 
- lastUpdateTime field of the cosmos db (_ts)
- set of blobs read with the same lastUpdateTime


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudFindToken.java
```diff
@@ -160,21 +181,21 @@ public boolean equals(Object o) {
...
   }
 
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder();
     sb.append("version: ").append(version);
     sb.append(" lastUpdateTime: ").append(lastUpdateTime);
-    sb.append(" latestBlobId: ").append(latestBlobId);
     sb.append(" bytesRead: ").append(bytesRead);
+    sb.append(" numLastUpdateTimeBlobsRead: ").append(lastUpdateTimeReadBlobIds);
```
lightningrob: Are you trying to print the set of ids or just the size of the set?  (The header doesn't match.)

ambry-cloud/src/test/java/com.github.ambry.cloud/LatchBasedInMemoryCloudDestination.java
```diff
@@ -142,11 +143,34 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) {
...
+        }
         entries.add(map.get(blobId).getFirst());
       }
     }
-    return entries;
+    Collections.sort(entries, (CloudBlobMetadata o1, CloudBlobMetadata o2) -> {
+      if (o1.getLastUpdateTime() == o2.getLastUpdateTime()) {
+        return 0;
+      }
+      return o1.getLastUpdateTime() > o2.getLastUpdateTime() ? 1 : -1;
```
lightningrob: There should be a comparator lambda thing to do this all in one line.

ambry-cloud/src/test/java/com.github.ambry.cloud/LatchBasedInMemoryCloudDestination.java
```diff
@@ -142,11 +143,34 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) {
...
+        if (results.size() == 0) {
+          // We must add at least one regardless of size
+          results.add(metadata);
+        }
+        break;
+      }
+      results.add(metadata);
+      totalSize += metadata.getSize();
+    }
+    return results;
```
lightningrob: This section looks copied from the prod class, please reuse.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -344,25 +345,24 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
-        }
 
+      List<CloudBlobMetadata> cappedResults = new ArrayList<>();
+      long totalSize = 0;
+      for (CloudBlobMetadata metadata : queryResults) {
         // Cap results at max size
         if (totalSize + metadata.getSize() > maxTotalSizeOfEntries) {
-          // If we have no results yet, must add at least one regardless of size
-          if (cappedResults.isEmpty()) {
+          if (cappedResults.size() == 0) {
```
lightningrob: Why is size() == 0 preferable to isEmpty()?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -376,6 +376,27 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
     }
   }
 
+  /**
+   * Filter out {@link CloudBlobMetadata} objects from lastUpdateTime ordered {@code cloudBlobMetadataList} whose
+   * lastUpdateTime is {@code lastUpdateTime} and id is in {@code lastReadBlobIds}.
+   * @param cloudBlobMetadataList list of {@link CloudBlobMetadata} objects to filter out from.
```
lightningrob: Remove "from"

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -376,6 +376,27 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
+   * @param lastReadBlobIds set if blobIds which need to be filtered out.
+   * @param lastUpdateTime lastUpdateTime of the blobIds to filter out.
+   */
+  private void filterOutLastReadBlobs(List<CloudBlobMetadata> cloudBlobMetadataList, Set<String> lastReadBlobIds,
+      long lastUpdateTime) {
+    List<CloudBlobMetadata> removeList = new LinkedList<>();
+    for (CloudBlobMetadata cloudBlobMetadata : cloudBlobMetadataList) {
```
lightningrob: Minor: you can actually iterate through the input list and call iter.remove() along the way (a bit more efficient).

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -345,6 +353,46 @@ public CloudBlobMetadata setEncryptedSize(long encryptedSize) {
...
+   */
+  public void setLastUpdateTime(long lastUpdateTime) {
+    this.lastUpdateTime = lastUpdateTime;
+  }
+
+  /**
+   * Utility to cap specified {@link CloudBlobMetadata} list by specified size of its blobs.
+   * Always returns at least one metadata object irrespective of size.
+   * @param originalList List of {@link CloudBlobMetadata}.
+   * @param size size total size of metadata's blobs.
```
lightningrob: size repeated

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
```diff
@@ -345,6 +353,46 @@ public CloudBlobMetadata setEncryptedSize(long encryptedSize) {
...
+   * @return {@link List} of {@link CloudBlobMetadata} capped by size.
+   */
+  public static List<CloudBlobMetadata> capMetadataListBySize(List<CloudBlobMetadata> originalList, long size) {
```
lightningrob: I'd prefer to keep CloudBlobMetadata a pojo class and move this method to a utility class that both prod and test code can call.

ankagrawal: I couldn't find a better place for this code. I could maybe create a CloudUtil class, but it will have only one method for now.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
ankagrawal: Actually I think from is needed. "CloudBlobMetadatalist" is a list of objects to filter out "from". If we remove "from" it would mean these are the objects to be filtered out.

ankagrawal: Ignore this. I misunderstood. will fix it.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobMetadata.java
ankagrawal: As discussed we will leave it as is for now.

PR #1296, opened 2019-11-01
Adding reset() method to BlockingChannel

dniq:
The reset() method is forcing the socket to be closed by sending TCP RST instead of TCP FIN and is used in cases when we close the socket due to an error, for example - this prevents the socket from becoming an 'orphan'

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/ConnectionPoolConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ConnectionPoolConfig.java
@@ -64,7 +64,7 @@ public class ConnectionPoolConfig {
    * Use TCP RST instead of FIN on replication socket error
    */
   @Config("connectionpool.socket.reset.on.error")
-  @Default(true)
+  @Default("true")
   public final boolean connectionPoolSocketResetOnError;
 
   public ConnectionPoolConfig(VerifiableProperties verifiableProperties) {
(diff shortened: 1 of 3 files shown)


ambry-api/src/main/java/com.github.ambry/config/ConnectionPoolConfig.java
```diff
@@ -60,6 +60,13 @@
   @Default("2")
   public final int connectionPoolMaxConnectionsPerPortSSL;
 
+  /**
+   * Use TCP RST instead of FIN on replication socket error
+   */
+  @Config("connectionpool.socket.reset.on.error")
+  @Default(true)
```
jsjtzyy: change this to String:  `@Default("true")`

ambry-api/src/main/java/com.github.ambry/config/ConnectionPoolConfig.java
```diff
@@ -60,6 +60,13 @@
...
+  public final boolean connectionPoolSocketResetOnError;
+
   public ConnectionPoolConfig(VerifiableProperties verifiableProperties) {
     connectionPoolReadBufferSizeBytes =
```
jsjtzyy: Add this into ConnectionPoolConfig constructor (line 82).
```connectionPoolSocketResetOnError = verifiableProperties.getBoolean("connectionpool.socket.reset.on.error", true);```

File: ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java [hunk 3]
```diff
@@ -102,6 +102,20 @@ public void disconnect() {
     }
   }
 
+  public void reset() {
```
jsjtzyy: After revisiting the code, I think you need to make changes in both `BlockingChannel` and `SSLBlockingChannel`.

jsjtzyy: Never mind, I see `SSLBlockingChannel` extends `BlockingChannel`

dniq: The reset() is inherited by SSLBlockingChannel, and all reset() is doing is setting setSoLinger and then calling the disxonnet() method of the object, which, in case of SSLBlockingChannel would be its own disconnect().

PR #1300, opened 2019-11-05
Refactoring BlockingChannel and SSLBlockingChannel for better code reuse

dniq:
To unify code between BlockingChannel and SSLBlockingChannel I added 'socket' property to BlockingChannel and renamed 'sslSocket' of SSLBlockingChannel to 'socket'. With a few additional changes to the BlockingChannel method - SSLBlockingChannel now no longer need its own 'disconnect()' implementation, and 'reset()' method will now work properly for both classes.

(Also, 'channel' didn't need to be closed in BlockingChannel.disconnect(), since writeChannel is the same object and it's already being closed, so removed that).


jsjtzyy:
There are some NPEs in integration test `endToEndSSLReplicationWithMultiNodeMultiPartitionTest`

dniq:
Will resubmit again later

PR #1304, opened 2019-11-08
Rebuild bloom filter with appropriate size

jsjtzyy:
Previously, size of rebuilt bloom filter is based on the actual number of
entries in index segment. That number varies in different index segments
and could be very small. This PR ensures size of rebuilt bloom filter is
no less than "storeIndexMaxNumberOfInmemElements" in config and thus the
rebuilt bloom filters should have same performance (compared with previous
bloom filters) on false positive rate.

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
@@ -411,10 +411,9 @@ class IndexSegment {
    */
   private void generateBloomFilterAndPersist() throws StoreException {
     int numOfIndexEntries = numberOfEntries(serEntries);
-    // The number of elements that used to create bloom filter should be no less than "storeIndexMaxNumberOfInmemElements"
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
```diff
@@ -411,7 +411,12 @@ private ByteBuffer getStoreKeyBytes(StoreKey key) {
    */
   private void generateBloomFilterAndPersist() throws StoreException {
     int numOfIndexEntries = numberOfEntries(serEntries);
-    bloomFilter = FilterFactory.getFilter(numOfIndexEntries, config.storeIndexBloomMaxFalsePositiveProbability);
+    // The number of elements that used to create bloom filter should be no less than "storeIndexMaxNumberOfInmemElements"
```
cgtz: rephrase this to say that this is a workaround since we found higher than intended false positive rates with small bloom filter sizes instead of saying it "should be no less than ...".

PR #1305, opened 2019-11-08
Refactoring BlockingChannel and SSLBlockingChannel for better code reuse

dniq:
This change optimizes the code and allows for a better code reuse:

1. Removing SocketChannel from BlockingChannel class
2. Adding a private "socket" field to BlockingChannel
3. Adding "createSocket()" method, which returns a connected socket
4. Removing all overrides from SSLBlockingChannel and instead adding one override for "createSocket()" method, which calls super.createSocket() and "wraps" SSL around it.
5. Replacing "writeChannel" value in BlockingChannel from SocketChannel to socket.getOutputChannel() wrapped in a WriteByteChannel (same as used to be in SSLBlockingChannel)
6. Ditching a separate lock object in favor of locking on 'this'.

### DIFF
--- a/ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
+++ b/ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
@@ -16,9 +16,10 @@ package com.github.ambry.network;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetSocketAddress;
+import java.net.Socket;
 import java.nio.ByteBuffer;
+import java.nio.channels.Channels;
 import java.nio.channels.ClosedChannelException;
-import java.nio.channels.SocketChannel;
 import java.nio.channels.WritableByteChannel;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -28,7 +29,7 @@ import org.slf4j.LoggerFactory;
  * A blocking channel that is used to communicate with a server
  */
 public class BlockingChannel implements ConnectedChannel {
-  protected static final Logger logger = LoggerFactory.getLogger(BlockingChannel.class);
+  protected final Logger logger = LoggerFactory.getLogger(getClass());
   protected final String host;
   protected final int port;
   protected final int readBufferSize;
@@ -38,8 +39,7 @@ public class BlockingChannel implements ConnectedChannel {
   protected boolean connected = false;
   protected InputStream readChannel = null;
   protected WritableByteChannel writeChannel = null;
-  protected Object lock = new Object();
-  private SocketChannel channel = null;
+  private Socket socket = null;
 
   public BlockingChannel(String host, int port, int readBufferSize, int writeBufferSize, int readTimeoutMs,
       int connectTimeoutMs) {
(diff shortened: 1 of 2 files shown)


ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
```diff
@@ -54,37 +55,57 @@ public BlockingChannel(String host, int port, int readBufferSize, int writeBuffe
...
+        readChannel = socket.getInputStream();
+
         connected = true;
-        logger.debug("Created socket with SO_TIMEOUT = {} (requested {}), "
-                + "SO_RCVBUF = {} (requested {}), SO_SNDBUF = {} (requested {})", channel.socket().getSoTimeout(),
-            readTimeoutMs, channel.socket().getReceiveBufferSize(), readBufferSize,
-            channel.socket().getSendBufferSize(), writeBufferSize);
+
+        logger.debug(
+            "Connection established to {}({}):{}",
```
cgtz: Could you format the code in intellij?

ambry-network/src/main/java/com.github.ambry.network/SSLBlockingChannel.java
```diff
@@ -51,75 +48,38 @@ public SSLBlockingChannel(String host, int port, MetricRegistry registry, int re
...
-        if (readBufferSize > 0) {
-          socket.setReceiveBufferSize(readBufferSize);
-        }
-        if (writeBufferSize > 0) {
-          socket.setSendBufferSize(writeBufferSize);
-        }
-        socket.connect(new InetSocketAddress(host, port), connectTimeoutMs);
-        sslSocket = (SSLSocket) sslSocketFactory.createSocket(socket, host, port, true);
+  public Socket createSocket() throws IOException {
+    Socket tcpSocket = super.createSocket();
```
cgtz: Could you call `tcpSocket.close()` if there was an exception during SSL initialization?

ambry-network/src/main/java/com.github.ambry.network/SSLBlockingChannel.java
```diff
@@ -51,75 +48,38 @@ public SSLBlockingChannel(String host, int port, MetricRegistry registry, int re
...
+    // handshake in a blocking way
+    try {
+      sslSocket.startHandshake();
+      sslClientHandshakeCount.inc();
+    } catch (IOException e) {
+      sslClientHandshakeErrorCount.inc();
+      throw e;
     }
+
+    logger.debug(
```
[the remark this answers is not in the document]
dniq: You mean, not 'protected static'?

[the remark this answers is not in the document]
jsjtzyy: This comment is optional. If you decide to change, then `static` is not needed. `static` usually means the field is owned by class itself.

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
dniq: Done

ambry-network/src/main/java/com.github.ambry.network/SSLBlockingChannel.java
dniq: Done (with SO_LINGER set to 0, to issue TCP RST instead of FIN).

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
```diff
@@ -39,7 +40,7 @@
   protected InputStream readChannel = null;
   protected WritableByteChannel writeChannel = null;
   protected Object lock = new Object();
```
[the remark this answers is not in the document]
dniq: Done

ambry-network/src/main/java/com.github.ambry.network/SSLBlockingChannel.java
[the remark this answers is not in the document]
dniq: Done

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
```diff
@@ -54,37 +55,45 @@ public BlockingChannel(String host, int port, int readBufferSize, int writeBuffe
   public void connect() throws IOException {
```
lightningrob: Since you're modifying, please add javadoc.

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
```diff
@@ -54,37 +55,45 @@ public BlockingChannel(String host, int port, int readBufferSize, int writeBuffe
...
-            readTimeoutMs, channel.socket().getReceiveBufferSize(), readBufferSize,
-            channel.socket().getSendBufferSize(), writeBufferSize);
+
+        logger.debug("Connection established to {}({}):{}", host, socket.getRemoteSocketAddress(), socket.getPort());
+
       }
     }
   }
 
+  public Socket createSocket() throws IOException {
```
lightningrob: javadoc

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
```diff
@@ -54,37 +55,45 @@ public BlockingChannel(String host, int port, int readBufferSize, int writeBuffe
...
+    tcpSocket.connect(new InetSocketAddress(host, port), connectTimeoutMs);
+
+    logger.debug("Created socket with SO_TIMEOUT = {} (requested {}), "
+            + "SO_RCVBUF = {} (requested {}), SO_SNDBUF = {} (requested {})", tcpSocket.getSoTimeout(), readTimeoutMs,
+        tcpSocket.getReceiveBufferSize(), readBufferSize, tcpSocket.getSendBufferSize(), writeBufferSize);
+
+    return tcpSocket;
+  }
+
   public void disconnect() {
```
lightningrob: javadoc

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
```diff
@@ -38,8 +39,8 @@
   protected boolean connected = false;
   protected InputStream readChannel = null;
   protected WritableByteChannel writeChannel = null;
-  protected Object lock = new Object();
-  private SocketChannel channel = null;
+  protected final Object lock = new Object();
```
lightningrob: Since there's only one lock, you can remove it and synchronize the methods that use it.

ambry-network/src/main/java/com.github.ambry.network/SSLBlockingChannel.java
```diff
@@ -51,75 +48,42 @@ public SSLBlockingChannel(String host, int port, MetricRegistry registry, int re
...
-        writeChannel = Channels.newChannel(sslSocket.getOutputStream());
-        readChannel = sslSocket.getInputStream();
-        connected = true;
-        logger.debug(
-            "Created socket with SO_TIMEOUT = {} (requested {}), SO_RCVBUF = {} (requested {}), SO_SNDBUF = {} (requested {})",
-            sslSocket.getSoTimeout(), readTimeoutMs, sslSocket.getReceiveBufferSize(), readBufferSize,
-            sslSocket.getSendBufferSize(), writeBufferSize);
+      ArrayList<String> protocolsList = Utils.splitString(sslConfig.sslEnabledProtocols, ",");
+      if (!protocolsList.isEmpty()) {
+        String[] enabledProtocols = protocolsList.toArray(new String[protocolsList.size()]);
```
lightningrob: Why are you going from config string to ArrayList to array of Strings?  String.split() does that in one shot.

cgtz: There is a slight gotcha with `String.split()`. If the string passed to it is "", it will return an array with 1 empty string in it, so the check after will not work. 

I think it would be best if `SSLBlockingChannel` did not need to be given an `SSLConfig` object and parse the configs every time it creates a new connection.  Instead, the `SSLFactory` could provide a method for getting an `SSLSocket` instead of an `SSLContext` object. However, that may be out of the scope of this PR and should be done in a follow up.

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
dniq: I was trying to avoid making too many changes at once, but yes, I'm aware of locking on 'this' :) I can make the change, if everyone is OK with that

ambry-network/src/main/java/com.github.ambry.network/SSLBlockingChannel.java
dniq: I agree - lets not make too many changes this time around, as the refactoring alone is quite a significant change.

The string indeed should be split and stored as an array only once, though, and we definitely should make that change sooner rather than later.

lightningrob: That's fine.  It was hard to tell how much of the code is new vs moved around since I'm not that familiar with it.  I was just pointing out the deficits that I noticed.

PR #1307, opened 2019-11-11
Support getting end position of last PUT in BlobStore

jsjtzyy:
The PR introduces a new method in PersistentIndex to get end position of
last PUT record in BlobStore. Note that, the "last" PUT record is valid
only when the method is invoked. That is, there could be new PUTs
afterwards. The intention here is to help replica to check if its peer
replicas have caught up with latest PUT record when PUT is disabled on
this replica (this happens when current replica is being decommissioned)

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
@@ -778,6 +778,10 @@ class IndexSegment {
           if (indexValue.getFlags() == IndexValue.FLAGS_DEFAULT_VALUE && (indexValueOfLastPut == null
               || indexValue.compareTo(indexValueOfLastPut) > 0)) {
             indexValueOfLastPut = indexValue;
+            // note that values set contains all entries associated with specific key, so there are at most 3 entries in
+            // this set (one PUT, one TTL Update and one DELETE). Due to nature of log, PUT always comes first. And if we
+            // already find PUT, we can jump out of the inner loop.
+            break;
           }
         }
       }
@@ -788,6 +792,7 @@ class IndexSegment {
           if (indexValue.getFlags() == IndexValue.FLAGS_DEFAULT_VALUE && (indexValueOfLastPut == null
               || indexValue.compareTo(indexValueOfLastPut) > 0)) {
             indexValueOfLastPut = indexValue;
+            break;
           }
         }
       }
--- a/ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
@@ -761,6 +761,41 @@ class IndexSegment {
   }
 
   /**
+   * @return index value of last PUT record in this index segment. Return {@code null} if no PUT is found
+   */
+  IndexValue getIndexValueOfLastPut() throws StoreException {
+    IndexValue indexValueOfLastPut = null;
+    if (sealed.get()) {
+      ByteBuffer readBuf = serEntries.duplicate();
+      int numOfIndexEntries = numberOfEntries(readBuf);
+      NavigableSet<IndexValue> values = new TreeSet<>();
+      for (int i = 0; i < numOfIndexEntries; i++) {
+        StoreKey key = getKeyAt(readBuf, i);
+        values.clear();
+        getAllValuesFromMmap(readBuf, key, i, numOfIndexEntries, values);
+        for (IndexValue indexValue : values) {
+          // FLAGS_DEFAULT_VALUE means PUT record
+          if (indexValue.getFlags() == IndexValue.FLAGS_DEFAULT_VALUE && (indexValueOfLastPut == null
+              || indexValue.compareTo(indexValueOfLastPut) > 0)) {
+            indexValueOfLastPut = indexValue;
+          }
+        }
+      }
+    } else {
+      for (Map.Entry<StoreKey, ConcurrentSkipListSet<IndexValue>> entry : index.entrySet()) {
+        for (IndexValue indexValue : entry.getValue()) {
+          // FLAGS_DEFAULT_VALUE means PUT record
+          if (indexValue.getFlags() == IndexValue.FLAGS_DEFAULT_VALUE && (indexValueOfLastPut == null
+              || indexValue.compareTo(indexValueOfLastPut) > 0)) {
+            indexValueOfLastPut = indexValue;
+          }
+        }
+      }
+    }
+    return indexValueOfLastPut;
+  }
+
+  /**
    * Maps the segment of index either as a memory map or a in memory buffer depending on config.
    * @throws StoreException if there are problems with the index
    */
--- a/ambry-store/src/main/java/com.github.ambry.store/Journal.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/Journal.java
@@ -16,6 +16,7 @@ package com.github.ambry.store;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableMap;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
@@ -39,6 +40,18 @@ class JournalEntry {
 
   StoreKey getKey() {
     return key;
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    JournalEntry entry = (JournalEntry) o;
+    return this.offset == entry.offset && this.key == entry.key;
   }
 }
 
@@ -155,6 +168,23 @@ class Journal {
   }
 
   /**
+   * @return all journal entries at the time when this method is invoked. (Return empty list if journal is empty)
+   */
+  List<JournalEntry> getAllEntries() {
+    List<JournalEntry> result = new ArrayList<>();
+    // get current last Offset
+    Offset lastOffset = getLastOffset();
+    if (lastOffset != null) {
+      // get portion view of journal whose key is less than or equal to lastOffset
+      NavigableMap<Offset, StoreKey> journalView = journal.headMap(lastOffset, true);
+      for (Map.Entry<Offset, StoreKey> entry : journalView.entrySet()) {
+        result.add(new JournalEntry(entry.getKey(), entry.getValue()));
+      }
+    }
+    return result;
+  }
+
+  /**
    * @return the first/smallest offset in the journal or {@code null} if no such entry exists.
    */
   Offset getFirstOffset() {
@@ -200,6 +230,14 @@ class Journal {
    */
   void finishBootstrap() {
     inBootstrapMode = false;
+  }
+
+  boolean isInBootstrapMode(){
+    return inBootstrapMode;
+  }
+
+  int getMaxEntriesToJournal(){
+    return maxEntriesToJournal;
   }
 
   /**
--- a/ambry-store/src/main/java/com.github.ambry.store/Journal.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/Journal.java
@@ -17,6 +17,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
@@ -52,6 +53,11 @@ class JournalEntry {
     }
     JournalEntry entry = (JournalEntry) o;
     return this.offset == entry.offset && this.key == entry.key;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(offset, key);
   }
 }
 
@@ -174,6 +180,7 @@ class Journal {
     List<JournalEntry> result = new ArrayList<>();
     // get current last Offset
     Offset lastOffset = getLastOffset();
+    journal.entrySet();
     if (lastOffset != null) {
       // get portion view of journal whose key is less than or equal to lastOffset
       NavigableMap<Offset, StoreKey> journalView = journal.headMap(lastOffset, true);
@@ -232,11 +239,11 @@ class Journal {
     inBootstrapMode = false;
   }
 
-  boolean isInBootstrapMode(){
+  boolean isInBootstrapMode() {
     return inBootstrapMode;
   }
 
-  int getMaxEntriesToJournal(){
+  int getMaxEntriesToJournal() {
     return maxEntriesToJournal;
   }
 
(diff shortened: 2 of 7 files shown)


ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1063,6 +1063,47 @@ long getLogUsedCapacity() {
...
+        break;
+      }
+    }
+    if (!journalEntries.isEmpty() && indexValueOfLastPut == null) {
+      // 2. if not find in the journal, check index segments starting from most recent one (until latest PUT is found)
+      JournalEntry firstJournalEntry = journalEntries.get(0);
+      // generate a segment map in reverse order
+      ConcurrentNavigableMap<Offset, IndexSegment> segmentsMapToSearch =
+          indexSegments.subMap(indexSegments.firstKey(), true, indexSegments.lowerKey(firstJournalEntry.getOffset()),
+              true).descendingMap();
```
justinlin-linkedin: nit: toInclusive can be false.

ambry-store/src/main/java/com.github.ambry.store/Journal.java
```diff
@@ -154,6 +167,23 @@ void addEntry(Offset offset, StoreKey key) {
     return journalEntries;
   }
 
+  /**
+   * @return all journal entries at the time when this method is invoked. (Return empty list if journal is empty)
+   */
+  List<JournalEntry> getAllEntries() {
```
justinlin-linkedin: i am a little concerned about how often this is going to be invoked. It seems like a huge waste that most of the result would be thrown away.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1063,6 +1063,47 @@ long getLogUsedCapacity() {
...
 
+  /**
+   * @return absolute end position (in bytes) of latest PUT record when this method is invoked.
+   * @throws StoreException
+   */
+  long getAbsoluteEndPositionOfLastPut() throws StoreException {
+    IndexValue indexValueOfLastPut = null;
+    // 1. go through keys in journal in reverse order to see if there is a PUT index entry associated with the key
+    List<JournalEntry> journalEntries = journal.getAllEntries();
+    ConcurrentSkipListMap<Offset, IndexSegment> indexSegments = validIndexSegments;
```
dharju: unnecessary line?

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
```diff
@@ -760,6 +760,41 @@ void seal() throws StoreException {
     persistBloomFilter();
   }
 
+  /**
+   * @return index value of last PUT record in this index segment. Return {@code null} if no PUT is found
+   */
+  IndexValue getIndexValueOfLastPut() throws StoreException {
+    IndexValue indexValueOfLastPut = null;
+    if (sealed.get()) {
+      ByteBuffer readBuf = serEntries.duplicate();
```
zzmao: Is duplicate() needed?

jsjtzyy:
> It seems a complicated calculation. How about maintain a live variable and update it when PUT come?
> 
> If restart happens, set the value in index initialization.

I agree, the calculation is not cheap and I did consider your suggestion about maintaining variable and update when there is a new PUT. However, I decide current approach because this method is rarely called. Only when one replica is being decommissioned, we invoke this method and in most cases, the last PUT should be found in Journal already. Iterating index segments should be very rare. Even it happens, it's acceptable for decommissioning replica. With this approach, we don't have to introduce additional overhead for each PUT operation.

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
```diff
@@ -760,6 +760,41 @@ void seal() throws StoreException {
...
+      NavigableSet<IndexValue> values = new TreeSet<>();
+      for (int i = 0; i < numOfIndexEntries; i++) {
+        StoreKey key = getKeyAt(readBuf, i);
+        values.clear();
+        getAllValuesFromMmap(readBuf, key, i, numOfIndexEntries, values);
+        for (IndexValue indexValue : values) {
+          // FLAGS_DEFAULT_VALUE means PUT record
+          if (indexValue.getFlags() == IndexValue.FLAGS_DEFAULT_VALUE && (indexValueOfLastPut == null
+              || indexValue.compareTo(indexValueOfLastPut) > 0)) {
+            indexValueOfLastPut = indexValue;
```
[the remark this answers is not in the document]
jsjtzyy: Partially agree. True, I could add a `break` in the loop of  `for (IndexValue indexValue : values)` when PUT record is found.
Note that, the navigable set `values` contains all index entries associated with specific `key`. So, maximum size of this set is 3 ( one PUT entry, one TTL Update entry and one DELETE entry). Due to the nature of log semantic, PUT always comes first. Hence, there is no need to search in reverse order.

zzmao:
> > It seems a complicated calculation. How about maintain a live variable and update it when PUT come?
> > If restart happens, set the value in index initialization.
> 
> I agree, the calculation is not cheap and I did consider your suggestion about maintaining variable and update when there is a new PUT. However, I decide current approach because this method is rarely called. Only when one replica is being decommissioned, we invoke this method and in most cases, the last PUT should be found in Journal already. Iterating index segments should be very rare. Even it happens, it's acceptable for decommissioning replica. With this approach, we don't have to introduce additional overhead for each PUT operation.

I see, but 1. maintenance is also overhead. 2. people in the future may invoke it 3. how big is the addition overhead to PUT?

ambry-store/src/main/java/com.github.ambry.store/Journal.java
jsjtzyy: it's not cheap but this method is called rarely, only when replica is being decommissioned.

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
```diff
@@ -760,6 +760,41 @@ void seal() throws StoreException {
...
+          }
+        }
+      }
+    } else {
+      for (Map.Entry<StoreKey, ConcurrentSkipListSet<IndexValue>> entry : index.entrySet()) {
+        for (IndexValue indexValue : entry.getValue()) {
+          // FLAGS_DEFAULT_VALUE means PUT record
```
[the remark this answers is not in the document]
jsjtzyy: yeah, let me know what name you feel better. We can either make the change in this PR or rename it in future PR

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1063,6 +1063,47 @@ long getLogUsedCapacity() {
     return getAbsolutePositionInLogForOffset(getCurrentEndOffset(indexSegments), indexSegments);
   }
 
+  /**
+   * @return absolute end position (in bytes) of latest PUT record when this method is invoked.
```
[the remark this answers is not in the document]
jsjtzyy: will fix

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1063,6 +1063,47 @@ long getLogUsedCapacity() {
...
+   * @throws StoreException
+   */
+  long getAbsoluteEndPositionOfLastPut() throws StoreException {
```
cgtz: What do you mean by "absolute" in this context?

ambry-store/src/main/java/com.github.ambry.store/Journal.java
```diff
@@ -154,6 +167,23 @@ void addEntry(Offset offset, StoreKey key) {
...
+    List<JournalEntry> result = new ArrayList<>();
+    // get current last Offset
+    Offset lastOffset = getLastOffset();
+    if (lastOffset != null) {
+      // get portion view of journal whose key is less than or equal to lastOffset
+      NavigableMap<Offset, StoreKey> journalView = journal.headMap(lastOffset, true);
+      for (Map.Entry<Offset, StoreKey> entry : journalView.entrySet()) {
+        result.add(new JournalEntry(entry.getKey(), entry.getValue()));
```
cgtz: I talked offline with Yingyi about possibly passing in a lambda that would test if the journal entry corresponds to a put entry and returning a single JournalEntry instead of a full snapshot. We discussed that this would not work since the ordering of taking the snapshot before searching through validIndexSegments is important in this case.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
jsjtzyy: It could be false in some cases but if journal doesn't cover the entire index segment (which means some entries slip out of journal that we didn't check). Explicitly set true will help avoid such edge case and ensure all entries in index segment are checked, although some entries would be checked twice.

ambry-store/src/main/java/com.github.ambry.store/Journal.java
```diff
@@ -154,6 +167,23 @@ void addEntry(Offset offset, StoreKey key) {
...
+  /**
+   * @return all journal entries at the time when this method is invoked. (Return empty list if journal is empty)
+   */
+  List<JournalEntry> getAllEntries() {
+    List<JournalEntry> result = new ArrayList<>();
+    // get current last Offset
+    Offset lastOffset = getLastOffset();
+    if (lastOffset != null) {
+      // get portion view of journal whose key is less than or equal to lastOffset
+      NavigableMap<Offset, StoreKey> journalView = journal.headMap(lastOffset, true);
```
cgtz: might be able to just iterate through `journal.entrySet()` instead of finding the last offset on line 176. Not sure if the extra entries returned by doing so would cause any issues

jsjtzyy:
> > > It seems a complicated calculation. How about maintain a live variable and update it when PUT come?
> > > If restart happens, set the value in index initialization.
> > 
> > 
> > I agree, the calculation is not cheap and I did consider your suggestion about maintaining variable and update when there is a new PUT. However, I decide current approach because this method is rarely called. Only when one replica is being decommissioned, we invoke this method and in most cases, the last PUT should be found in Journal already. Iterating index segments should be very rare. Even it happens, it's acceptable for decommissioning replica. With this approach, we don't have to introduce additional overhead for each PUT operation.
> 
> I see, but 1. maintenance is also overhead. 2. people in the future may invoke it 3. how big is the addition overhead to PUT?

Talked with @zzmao offline. Updating offset (or blob id) for every PUT operation needs extra work in compaction. The reason is the last updated offset in PUT operation can be invalid due to compaction. Imagine a scenario where all PUTs go to previous log segment and last active segment only has DELETE/TTL update entries and partition itself is sealed. The last PUT blob can be compacted already and the associated offset is no longer valid.

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
```diff
@@ -760,6 +760,41 @@ void seal() throws StoreException {
...
+          if (indexValue.getFlags() == IndexValue.FLAGS_DEFAULT_VALUE && (indexValueOfLastPut == null
+              || indexValue.compareTo(indexValueOfLastPut) > 0)) {
+            indexValueOfLastPut = indexValue;
+          }
+        }
+      }
+    } else {
+      for (Map.Entry<StoreKey, ConcurrentSkipListSet<IndexValue>> entry : index.entrySet()) {
+        for (IndexValue indexValue : entry.getValue()) {
```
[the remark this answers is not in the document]
jsjtzyy: Basically your concern is when iterating on an entry, the `index` may have changed which makes last PUT invalid. This should not happen because we always check Journal first and Journal covers most recent two index segments. Hence, when we start to check index segment (it is very likely sealed but not guaranteed, seal operation is performed by background persistor), we can guarantee that this index segment being checked is not an active one, that is, no more entries will be added into this index segment. So it is safe to iterate over all entries in this index segment and validity of last PUT still holds

ambry-store/src/main/java/com.github.ambry.store/Journal.java
```diff
@@ -40,6 +41,18 @@ Offset getOffset() {
   StoreKey getKey() {
     return key;
   }
+
+  @Override
+  public boolean equals(Object o) {
```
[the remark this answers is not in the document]
jsjtzyy: right, will fix.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
jsjtzyy: Reference `validIndexSegments` might be changed to point to another ConcurrentSkipListMap (which is inFluxIndexSegments) during iteration. We still use old view to finish finding the last PUT.  (Although, in my move replica design, I will disable compaction first before calling this method, I still would like to keep a "dedicated" reference pointing to the indexSegments view at the time when the method is called)

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
jsjtzyy: The class `Offset` contains offset within specific log segment. Here we need offset in entire log.

ambry-store/src/main/java/com.github.ambry.store/Journal.java
jsjtzyy: I don't see any issues but I am slightly leaning towards current implementation. The `lastOffset` denotes a clear cutoff and headMap only contains keys that below the cutoff. The `entrySet()` keeps changing/growing during the iteration, I feel the last entry is not so deterministic.

jsjtzyy:
Gentle reminder to review.

dharju:
Sorry, will get to it today

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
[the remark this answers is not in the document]
dharju: So having a loop is just a safety mechanic (since you could technically just grab the first entry and throw an exception if it's not a PUT)?

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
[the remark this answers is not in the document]
dharju: yeah we can rename in the future, or I can do it in my changes

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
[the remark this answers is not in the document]
jsjtzyy: right, I probably could check first entry only but the current way doesn't introduce any more overhead. Technically, we can throw exception if it's not a PUT, which means the store is in a "incorrect" state violating log/index semantics. However, 1. this situation shouldn't occur with Ambry current implementation. 2. throwing exception makes transition fail because this method is called when decommissioning a replica. It's probably ok and replica will be put into ERROR state but I feel like not necessary because the intention here is to ensure peer replicas catch up with local replica, the order of PUT entries could be fixed by other peer replicas and local one is still safe to be removed. (My two cents)

ambry-store/src/main/java/com.github.ambry.store/IndexSegment.java
zzmao: solved. duplicate() create a bytebuffer which shares the same content but different limit, position and capacity.

PR #1310, opened 2019-11-14
Fix netty ByteBuf release bug in the ambry-frontend

justinlin-linkedin:
ResponseInfo release the ByteBuf when GetOperation finishes handleResponse, and it would effectively put ByteBuf back to the pool. But GetOperation might buffer those bytes in a BlobDataReadableStreamChannel so releasing the ByteBuf would eventually corrupt the response if the same ByteBuf is checked out to hold different bytes.

This PR fixes this bug, but retaining ResponseInfo and put it in a map and later release it. The reason to add retain method is to make logic easier to understand. After this PR, ResponseInfo would be released after GetOperation finishes handleResponse, but if it's retained, it will then again get released after BlobDataReadableStreamChannel finishes writing to HttpResponse.

### DIFF
--- a/ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
@@ -99,6 +99,7 @@ class GetBlobOperation extends GetOperation {
   private ListIterator<CompositeBlobInfo.ChunkMetadata> chunkIdIterator;
   // chunk index to retrieved chunk buffer mapping.
   private Map<Integer, ByteBuffer> chunkIndexToBuffer;
+  private Map<Integer, ResponseInfo> chunkIndexToResponseInfo;
   // To find the GetChunk to hand over the response quickly.
   private final Map<Integer, GetChunk> correlationIdToGetChunk = new HashMap<>();
   // the blob info that is populated on OperationType.BlobInfo or OperationType.All
@@ -336,6 +337,8 @@ class GetBlobOperation extends GetOperation {
         if (exception != null) {
           setOperationException(exception);
         }
+        chunkIndexToResponseInfo.get(numChunksWrittenOut).release();
+        chunkIndexToResponseInfo.remove(numChunksWrittenOut);
         numChunksWrittenOut++;
         routerCallback.onPollReady();
       }
@@ -377,6 +380,7 @@ class GetBlobOperation extends GetOperation {
     @Override
     public void close() throws IOException {
       if (isOpen.compareAndSet(true, false)) {
+        chunkIndexToResponseInfo.values().forEach(ResponseInfo::release);
         if (numChunksWrittenOut != numChunksTotal) {
           setOperationException(new RouterException(
               "The ReadableStreamChannel for blob data has been closed by the user before all chunks were written out.",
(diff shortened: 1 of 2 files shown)


ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -336,6 +337,10 @@ public void onCompletion(Long result, Exception exception) {
         if (exception != null) {
           setOperationException(exception);
         }
+        ResponseInfo responseInfo = chunkIndexToResponseInfo.remove(numChunksWrittenOut);
+        if (responseInfo != null) {
+          responseInfo.release();
+        }
         numChunksWrittenOut++;
```
jsjtzyy: optional: `numChunksWrittenOut` can be atomic.

justinlin-linkedin: updated.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -1235,14 +1262,15 @@ RouterErrorCode processServerError(ServerErrorCode errorCode) {
...
+     * @param responseInfo the response received for a request sent out on behalf of this chunk.
      * @param blobData the metadata blob's data.
      * @param userMetadata userMetadata of the blob
      * @param encryptionKey blob encryption key. Could be null for un-encrypted blobs
      * @throws IOException
      * @throws MessageFormatException
      */
-    private void handleMetadataBlob(BlobData blobData, byte[] userMetadata, ByteBuffer encryptionKey)
-        throws IOException, MessageFormatException {
+    private void handleMetadataBlob(ResponseInfo responseInfo, BlobData blobData, byte[] userMetadata,
```
cgtz: I noticed that responseInfo is passed into this method but not used.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -714,18 +729,21 @@ void checkAndMaybeComplete() {
...
      */
-    void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInfo messageInfo)
-        throws IOException, MessageFormatException {
+    void handleBody(ResponseInfo responseInfo, InputStream payload, MessageMetadata messageMetadata,
+        MessageInfo messageInfo) throws IOException, MessageFormatException {
       if (!successfullyDeserialized) {
         BlobData blobData = MessageFormatRecord.deserializeBlob(payload);
         ByteBuffer encryptionKey = messageMetadata == null ? null : messageMetadata.getEncryptionKey();
         ByteBuffer chunkBuffer = blobData.getStream().getByteBuffer();
+        responseInfo.retain();
```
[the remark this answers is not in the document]
justinlin-linkedin: It's one. When a buffer is created, the refcount will be initialized to one. And this method would increase the refcount by one.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -714,18 +730,21 @@ void checkAndMaybeComplete() {
...
+        chunkIndexToResponseInfo.put(chunkIndex, responseInfo);
 
         boolean launchedJob = maybeLaunchCryptoJob(chunkBuffer, null, encryptionKey, chunkBlobId);
         if (!launchedJob) {
```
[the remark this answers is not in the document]
justinlin-linkedin: It will be released in GetManager.

In GetManger, it calls GetBlobOperation.handleResponse and passes the ResponseInfo as the first parameter. After it finishes handleResponse method, it will release ResponseInfo. So if the responseInfo is not retained inside handleResponse function, then we don't need extra release call.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
justinlin-linkedin: Will update.

ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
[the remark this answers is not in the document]
cgtz: ah right, in `OperationController` that first ref will be decremented and this will increment the ref count when handing off the buffer to a different context

PR #1311, opened 2019-11-14
Add a NettyByteBufLeakHelper to help test ByteBuf leaking

justinlin-linkedin:
ByteBuf resource leak detection is currently relying on reading the test log and find out whether there is any leaking log. This is not very reliable and very tedious to do.

This PR provides a helper class to detect the ByteBuf leak. This helper class relies on metrics from Netty library. It first disables the caches (caches would mess up the deallocations), then it checks after test, if the number of active allocations remain the same. If not, it fails the test.


ambry-commons/src/test/java/com.github.ambry.commons/NettyByteBufLeakHelper.java
```diff
@@ -0,0 +1,93 @@
...
+  private long directDeallocations;
+
+  // when the cache is enable, just don't check the allocation and deallocation.
+  private boolean cachedEnabled;
+
+  /**
+   * Constructor to create a {@link NettyByteBufLeakHelper}.
+   */
+  public NettyByteBufLeakHelper() {
+    cachedEnabled = Integer.parseInt(System.getProperty("io.netty.alllocator.maxCachedBufferCapacity", "0")) != 0;
```
cgtz: This should be `allocator` instead of `alllocator`.
Does netty have a constant for the property name that can be used here?

ambry-network/src/test/java/com.github.ambry.network/SocketNetworkClientTest.java
```diff
@@ -18,12 +18,16 @@
...
 import com.github.ambry.clustermap.MockDataNodeId;
 import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.commons.NettyByteBufLeakHelper;
 import com.github.ambry.config.NetworkConfig;
 import com.github.ambry.config.VerifiableProperties;
 import com.github.ambry.utils.MockTime;
 import com.github.ambry.utils.Time;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.PoolArenaMetric;
```
zzmao: Not used.

ambry-commons/src/test/java/com.github.ambry.commons/NettyByteBufLeakHelper.java
justinlin-linkedin: good catch, It doesn't really have a constant for those property keys. Will update.

ambry-network/src/test/java/com.github.ambry.network/SocketNetworkClientTest.java
justinlin-linkedin: will update.

PR #1313, opened 2019-11-17
Skip persisting replica token if the disk is unavailable(I/O errors)

jsjtzyy:
The background replica token persistor should skip bad disk if there are
hardware issues with it. When a disk I/O error occurs, BlobStore will be
shutdown automatically. The persistor then checks if all BlobStores are shutdown 
on that disk to determine its availability. The unavailable disk is skipped by replica
token persistor to avoid triggering more IOExceptions.

### DIFF
--- a/ambry-replication/src/main/java/com.github.ambry.replication/DiskTokenPersistor.java
+++ b/ambry-replication/src/main/java/com.github.ambry.replication/DiskTokenPersistor.java
@@ -14,7 +14,8 @@
 package com.github.ambry.replication;
 
 import com.github.ambry.clustermap.ClusterMap;
-import com.github.ambry.store.StorageManager;
+import com.github.ambry.server.ServerErrorCode;
+import com.github.ambry.server.StoreManager;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
@@ -36,7 +37,7 @@ public class DiskTokenPersistor extends ReplicaTokenPersistor {
 
   private static final Logger logger = LoggerFactory.getLogger(DiskTokenPersistor.class);
   private final String replicaTokenFileName;
-  private final StorageManager storageManager;
+  private final StoreManager storeManager;
 
   /**
    * Constructor for {@link DiskTokenPersistor}.
@@ -45,14 +46,14 @@ public class DiskTokenPersistor extends ReplicaTokenPersistor {
    * @param replicationMetrics metrics including token persist time.
    * @param clusterMap the {@link ClusterMap} to deserialize tokens.
    * @param tokenHelper the {@link FindTokenHelper} to deserialize tokens.
-   * @param storageManager the {@link StorageManager} that manages disks and stores.
+   * @param storeManager the {@link StoreManager} that manages disks and stores.
    */
   DiskTokenPersistor(String replicaTokenFileName, Map<String, Set<PartitionInfo>> partitionGroupedByMountPath,
       ReplicationMetrics replicationMetrics, ClusterMap clusterMap, FindTokenHelper tokenHelper,
-      StorageManager storageManager) {
+      StoreManager storeManager) {
     super(partitionGroupedByMountPath, replicationMetrics, clusterMap, tokenHelper);
     this.replicaTokenFileName = replicaTokenFileName;
-    this.storageManager = storageManager;
+    this.storeManager = storeManager;
   }
 
   @Override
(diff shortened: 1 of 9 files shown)


jsjtzyy:
working on unit tests

ambry-replication/src/main/java/com.github.ambry.replication/DiskTokenPersistor.java
```diff
@@ -62,6 +67,10 @@ protected void persist(String mountPath, List<ReplicaTokenInfo> tokenInfoList) t
       logger.debug("Completed writing replica tokens to file {}", actual.getAbsolutePath());
     } catch (IOException e) {
       logger.error("IO error while persisting tokens to disk {}", temp.getAbsoluteFile());
+      // check disk state in storageManager. If there is a hardware issue, persistor should skip this bad disk next time.
+      if (!storageManager.isDiskAvailableAtMountPath(mountPath)) {
```
lightningrob: You don't need to add a new method to StorageManager to check by mount path.  This class already has the partitionGroupedByMountPath map that you can use to find the partition/replica for the mount path.  Then you can call StoreManager.checkLocalPartitionStatus() and avoid adding replication -> store dependency.

ambry-replication/src/main/java/com.github.ambry.replication/DiskTokenPersistor.java
```diff
@@ -62,6 +67,10 @@ protected void persist(String mountPath, List<ReplicaTokenInfo> tokenInfoList) t
...
+        mountPathsToSkip.add(mountPath);
```
lightningrob: Do we need logic to remove the mount path when the disk comes back up, or can we assume the server will be restarted in that case?

ankagrawal: I guess if a disk is bad, we might have to replace the disk and restart server so that mount path of the replicas get changed. Or we might have to trigger replica update via helix ideal state change.

ambry-replication/src/main/java/com.github.ambry.replication/DiskTokenPersistor.java
jsjtzyy: Good point, will refactor as you suggested

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaTokenPersistor.java
```diff
@@ -43,6 +45,7 @@
 
   private static final Logger logger = LoggerFactory.getLogger(DiskTokenPersistor.class);
```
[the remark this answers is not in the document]
jsjtzyy: fixed

build.gradle
```diff
@@ -255,7 +255,8 @@ project(':ambry-replication') {
                 project(':ambry-utils'),
                 project(':ambry-commons'),
                 project(':ambry-protocol'),
-                project(':ambry-network')
+                project(':ambry-network'),
+                project(':ambry-store')
```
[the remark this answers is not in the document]
jsjtzyy: will remove

ambry-replication/src/main/java/com.github.ambry.replication/DiskTokenPersistor.java
jsjtzyy: Thanks for both inputs. Right now, when replacing a bad disk, we have to restart the server and `mountPathsToSkip` gets re-initialized. If in the future, hot swap is supported, we can consider dynamically removing the path out of `mountPathsToSkip`.

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
```diff
@@ -202,11 +202,14 @@ DiskManager getDiskManager(PartitionId id) {
   }
 
   /**
-   * Check if a certain disk is available.
+   * Check if a certain disk is available at specific mount path
```
ankagrawal: Not sure what we mean by this comment. We are only passing the Disk object.

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
```diff
@@ -202,11 +202,14 @@ DiskManager getDiskManager(PartitionId id) {
...
    * @param disk the {@link DiskId} to check.
    * @return {@code true} if the disk is available. {@code false} if not.
    */
   public boolean isDiskAvailable(DiskId disk) {
+//    Optional<Map.Entry<DiskId, DiskManager>> diskAndDiskManager =
```
ankagrawal: Can we remove the commented lines?

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
jsjtzyy: Sorry, I forgot to update this. Will fix

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
jsjtzyy: Will remove.

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
```diff
@@ -202,12 +202,14 @@ DiskManager getDiskManager(PartitionId id) {
...
-   * @param disk the {@link DiskId} to check.
+   * Check if a certain disk is available at specific mount path
+   * @param mountPath the {@link DiskId}'s mount path.
    * @return {@code true} if the disk is available. {@code false} if not.
    */
-  public boolean isDiskAvailable(DiskId disk) {
-    DiskManager diskManager = diskToDiskManager.get(disk);
+  public boolean isDiskAvailableAtMountPath(String mountPath) {
+    Optional<Map.Entry<DiskId, DiskManager>> diskAndDiskManager =
+        diskToDiskManager.entrySet().stream().filter(e -> e.getKey().getMountPath().equals(mountPath)).findFirst();
```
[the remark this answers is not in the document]
jsjtzyy: True, I have refactored the implementation and this part is no longer needed.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaTokenPersistor.java
```diff
@@ -99,6 +104,10 @@ public final void write(String mountPath, boolean shuttingDown) throws IOExcepti
     }
   }
 
+  Set<String> getMountPathsToSkip() {
```
lightningrob: javadoc

PR #1315, opened 2019-11-20
Introduce current state in BlobStore to help validate requests

jsjtzyy:
This PR introduces current state in BlobStore that is updated by state
transition (Helix controller initiates the transition). Store's current
state is used to validate requests and reject some of them if frontends
have old view of the store. For example, store has transitted to
INACTIVE state but due to Helix notification delay, some frontends may
still believe store is in STANDBY. To avoid such inconsistent view, we
use current state (which is always up-to-date) to validate coming
requests.

### DIFF
--- a/ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
+++ b/ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
@@ -72,10 +72,9 @@ public class AmbryServerRequests extends AmbryRequests {
   private final ConcurrentHashMap<RequestOrResponseType, Set<PartitionId>> requestsDisableInfo =
       new ConcurrentHashMap<>();
   private final ConcurrentHashMap<PartitionId, ReplicaId> localPartitionToReplicaMap;
-  // POST requests are allowed on stores which are in LEADER or STANDBY state
-  static final Set<ReplicaState> POST_ALLOWED_STORE_STATES =
-      EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY);
-  // UPDATE requests (including DELETE, TTLUpdate) are allowed on stores that are LEADER/STANDBY/INACTIVE/BOOTSTRAP
+  // POST requests are allowed on stores states: { LEADER, STANDBY }
+  static final Set<ReplicaState> POST_ALLOWED_STORE_STATES = EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY);
+  // UPDATE requests (including DELETE, TTLUpdate) are allowed on stores states: { LEADER, STANDBY, INACTIVE, BOOTSTRAP }
   static final Set<ReplicaState> UPDATE_ALLOWED_STORE_STATES =
       EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY, ReplicaState.INACTIVE, ReplicaState.BOOTSTRAP);
   static final Set<RequestOrResponseType> UPDATE_REQUEST_TYPES =
--- a/ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
+++ b/ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
@@ -21,8 +21,10 @@ import com.github.ambry.clustermap.HardwareState;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.PartitionState;
 import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.clustermap.ReplicaState;
 import com.github.ambry.clustermap.ReplicaStatusDelegate;
 import com.github.ambry.commons.ServerMetrics;
+import com.github.ambry.config.ServerConfig;
 import com.github.ambry.network.Request;
 import com.github.ambry.network.RequestResponseChannel;
 import com.github.ambry.network.ServerNetworkResponseMetrics;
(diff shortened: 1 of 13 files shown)


ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -76,6 +77,7 @@
   private BlobStoreStats blobStoreStats;
   private boolean started;
   private FileLock fileLock;
+  private ReplicaState currentState;
```
cgtz: this should probably be volatile since it will be set from a different thread than where it is read

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
```diff
@@ -145,7 +155,27 @@ public void handleRequests(Request request) throws InterruptedException {
...
+    // 1. check if request is disabled by admin request
+    if (requestDisableInfo != null && requestDisableInfo.contains(id)) {
+      return false;
+    }
+    if (serverConfig.serverValidateRequestBasedOnStoreState) {
+      // 2. check if request is disabled due to current state of store
+      Store store = storeManager.getStore(id);
+      if (requestType == RequestOrResponseType.PutRequest && !POST_ALLOWED_STORE_STATES.contains(
+          store.getCurrentState())) {
+        logger.error("{} is not allowed because current state of store {} is {}", requestType, id,
```
cgtz: Is error the correct log level for this? How often will this occur during normal operation?

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -171,6 +173,7 @@ private BlobStore(ReplicaId replicaId, String storeId, StoreConfig config, Sched
     this.thresholdBytesLow = (long) (capacityInBytes * ((threshold - delta) / 100.0));
     ttlUpdateBufferTimeMs = TimeUnit.SECONDS.toMillis(config.storeTtlUpdateBufferTimeSeconds);
     errorCount = new AtomicInteger(0);
+    currentState = ReplicaState.OFFLINE;
```
cgtz: I don't see any callers to `setCurrentState()` in the production code. Will this be added in a future PR?

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
jsjtzyy: fair point, will fix

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
jsjtzyy: Should be infrequent. It happens in two cases:
1. remove replica: when local replica's state transition has completed but frontends still have old view and still issue PUT requests to the store. Once Helix state change notification arrives at frontend, this should die out quickly. 
2. shutdown server: when server is shut down, local replica is directly set to OFFLINE and some frontends may still attempt to write to this store due to Helix notification delay.

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
jsjtzyy: Right, this will be called in storage manager when state transition occurs. I was planning to make the changes in this PR but then decided to split it into a separate PR.

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
```diff
@@ -65,21 +67,29 @@
...
 public class AmbryServerRequests extends AmbryRequests {
-
-  private final ConcurrentHashMap<PartitionId, ReplicaId> localPartitionToReplicaMap;
+  private final ServerConfig serverConfig;
   private final StatsManager statsManager;
   private final ConcurrentHashMap<RequestOrResponseType, Set<PartitionId>> requestsDisableInfo =
       new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<PartitionId, ReplicaId> localPartitionToReplicaMap;
+  // POST requests are allowed on stores states: { LEADER, STANDBY }
+  static final Set<ReplicaState> POST_ALLOWED_STORE_STATES = EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY);
```
justinlin-linkedin: Shouldn't it be PUT instead of Post?

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
```diff
@@ -65,21 +67,29 @@
...
+  // UPDATE requests (including DELETE, TTLUpdate) are allowed on stores states: { LEADER, STANDBY, INACTIVE, BOOTSTRAP }
+  static final Set<ReplicaState> UPDATE_ALLOWED_STORE_STATES =
+      EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY, ReplicaState.INACTIVE, ReplicaState.BOOTSTRAP);
+  static final Set<RequestOrResponseType> UPDATE_REQUEST_TYPES =
+      EnumSet.of(RequestOrResponseType.DeleteRequest, RequestOrResponseType.TtlUpdateRequest);
```
justinlin-linkedin: we have UNDELETE now.

justinlin-linkedin: These can just be a Map from RequestOrResponseType to EnumSet of ReplicaState. So in isRequestEnabled method, you don't have to do if-else statement.

ambry-server/src/test/java/com.github.ambry.server/AmbryServerRequestsTest.java
```diff
@@ -183,6 +191,61 @@ public void after() throws InterruptedException {
...
+   * Tests that requests are validated based on local store state.
+   */
+  @Test
+  public void validateRequestsTest() {
+    // choose several replicas and make them in different states (there are 10 replicas on current node)
+    List<ReplicaId> localReplicas = clusterMap.getReplicaIds(dataNodeId);
+    Map<ReplicaState, ReplicaId> stateToReplica = new HashMap<>();
+    int cnt = 0;
+    for (ReplicaState state : EnumSet.complementOf(EnumSet.of(ReplicaState.ERROR))) {
+      stateToReplica.put(state, localReplicas.get(cnt++));
```
justinlin-linkedin: nit: I would discourage using ++ in the same place where you use cnt. This would cause confusion. Can you do
```
stateToReplica.put(state, localReplicas.get(cnt));
cnt++;
```

ambry-server/src/test/java/com.github.ambry.server/AmbryServerRequestsTest.java
```diff
@@ -183,6 +191,61 @@ public void after() throws InterruptedException {
...
+    for (Map.Entry<ReplicaState, ReplicaId> entry : stateToReplica.entrySet()) {
+      storageManager.getStore(entry.getValue().getPartitionId()).setCurrentState(entry.getKey());
+    }
+
+    for (RequestOrResponseType request : EnumSet.of(RequestOrResponseType.PutRequest, RequestOrResponseType.GetRequest,
+        RequestOrResponseType.DeleteRequest, RequestOrResponseType.TtlUpdateRequest)) {
+      for (Map.Entry<ReplicaState, ReplicaId> entry : stateToReplica.entrySet()) {
+        if (request == RequestOrResponseType.PutRequest) {
+          System.out.println("replica state = " + entry.getKey() + ", store state = " + storageManager.getStore(
+              entry.getValue().getPartitionId()).getCurrentState());
```
justinlin-linkedin: Remove System.out.println?

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
jsjtzyy: fixed

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
jsjtzyy: I will temporarily defer the UNDELETE until the logic is fully implemented.

ambry-server/src/test/java/com.github.ambry.server/AmbryServerRequestsTest.java
jsjtzyy: removed

ambry-server/src/test/java/com.github.ambry.server/AmbryServerRequestsTest.java
jsjtzyy: fixed

ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
jsjtzyy: This is actually a good idea but we only have two categories of request type to account for, I feel like we don't have to initialize a map and query it. The logic is already simple in `isRequestEnabled` method.

PR #1317, opened 2019-11-20
Simplify PutRequest by removing ReceivedPutRequest

justinlin-linkedin:
We don't really need it ReceivedPutRequest, just like we don't have ReceivedGetResponse.


ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
```diff
@@ -87,9 +89,41 @@ public PutRequest(int correlationId, String clientId, BlobId blobId, BlobPropert
...
+    this.blobType = blobType;
+    this.blobEncryptionKey = blobEncryptionKey;
+    this.blob = null;
+    this.crc = null;
+    this.crcBuf = null;
+    this.blobStream = blobStream;
+    this.crcValue = crc;
+  }
+
+  public static PutRequest readFrom(DataInputStream stream, ClusterMap map) throws IOException {
```
lightningrob: javadoc

PR #1320, opened 2019-11-21
Adding an option to enable/disable TCP_NODELAY socket option

dniq:
This change adds a configuration property to enable or disable TCP_NODELAY socket option in BlockingChannel (used for replication only ATM) and adds an additional (optional) parameter for the same. If the parameter is not provided - a default 'true' (enabled) is assumed. This means that the tests (which have not been updated) will have TCP_NODELAY socket option set to 'true'.

### DIFF
--- a/ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
+++ b/ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
@@ -13,6 +13,7 @@
  */
 package com.github.ambry.network;
 
+import com.github.ambry.config.ConnectionPoolConfig;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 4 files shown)


ambry-network/src/main/java/com.github.ambry.network/BlockingChannelInfo.java
```diff
@@ -168,11 +168,11 @@ private BlockingChannel getBlockingChannelBasedOnPortType(String host, int port)
     if (this.port.getPortType() == PortType.PLAINTEXT) {
       channel = new BlockingChannel(host, port, config.connectionPoolReadBufferSizeBytes,
           config.connectionPoolWriteBufferSizeBytes, config.connectionPoolReadTimeoutMs,
-          config.connectionPoolConnectTimeoutMs);
+          config.connectionPoolConnectTimeoutMs, config.connectionPoolSocketTcpNoDelay);
     } else if (this.port.getPortType() == PortType.SSL) {
       channel = new SSLBlockingChannel(host, port, registry, config.connectionPoolReadBufferSizeBytes,
           config.connectionPoolWriteBufferSizeBytes, config.connectionPoolReadTimeoutMs,
-          config.connectionPoolConnectTimeoutMs, sslSocketFactory, sslConfig);
+          config.connectionPoolConnectTimeoutMs, config.connectionPoolSocketTcpNoDelay, sslSocketFactory, sslConfig);
```
jsjtzyy: I wonder if it's worthwhile to pass the `config` into BlockingChannel and SSLBlockingChannel constructor. Thus, we can configure the parameters inside constructor and make function signature simpler.

dniq: That would require more substantial changes. That was my first thought, and I even started doing that, and once I saw how many changes that involved across the board (especially in the tests) - I opted for passing a parameter instead.

dniq: I've added a separate constructor, which takes ConnectionPoolConfig instance as a parameter and updated the BlockingChannelInfo accordingly.

ambry-network/src/main/java/com.github.ambry.network/BlockingChannel.java
```diff
@@ -39,16 +40,29 @@
...
     this.connectTimeoutMs = connectTimeoutMs;
+    this.enableTcpNoDelay = enableTcpNoDelay;
+  }
+
+  public BlockingChannel(String host, int port, int readBufferSize, int writeBufferSize, int readTimeoutMs,
+      int connectTimeoutMs) {
+    this(host, port, readBufferSize, writeBufferSize, readTimeoutMs, connectTimeoutMs, true);
+  }
+
+  public BlockingChannel(String host, int port, ConnectionPoolConfig config) {
```
jsjtzyy: optional: can be package-private (`public` can be removed)

PR #1321, opened 2019-11-25
Use async cosmos library and upgrade to 2.6.3

ankagrawal:
This changes replaces the cosmos documentdb library with async cosmosdb library.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
@@ -75,7 +75,7 @@ class AzureCloudDestination implements CloudDestination {
   private static final String TIME_SINCE_PARAM = "@timesince";
   private static final String BATCH_ID_QUERY_TEMPLATE = "SELECT * FROM c WHERE c.id IN (%s)";
   static final int ID_QUERY_BATCH_SIZE = 1000;
-  static final String DEAD_BLOBS_QUERY_TEMPLATE =
+  private static final String DEAD_BLOBS_QUERY_TEMPLATE =
       "SELECT TOP " + LIMIT_PARAM + " * FROM c WHERE (c." + CloudBlobMetadata.FIELD_DELETION_TIME + " BETWEEN 1 AND "
           + THRESHOLD_PARAM + ")" + " OR (c." + CloudBlobMetadata.FIELD_EXPIRATION_TIME + " BETWEEN 1 AND "
           + THRESHOLD_PARAM + ")" + " ORDER BY c." + CloudBlobMetadata.FIELD_UPLOAD_TIME + " ASC";
@@ -83,7 +83,7 @@ class AzureCloudDestination implements CloudDestination {
   // It is unlikely (but not impossible) for two blobs in same partition to have the same uploadTime (would have to
   // be multiple VCR's uploading same partition).  We track the lastBlobId in the CloudFindToken and skip it if
   // is returned in successive queries.
-  static final String ENTRIES_SINCE_QUERY_TEMPLATE =
+  private static final String ENTRIES_SINCE_QUERY_TEMPLATE =
       "SELECT TOP " + LIMIT_PARAM + " * FROM c WHERE c." + CosmosDataAccessor.COSMOS_LAST_UPDATED_COLUMN + " >= "
           + TIME_SINCE_PARAM + " ORDER BY c." + CosmosDataAccessor.COSMOS_LAST_UPDATED_COLUMN + " ASC";
   private static final String SEPARATOR = "-";
@@ -129,8 +129,6 @@ class AzureCloudDestination implements CloudDestination {
     }
     if (cloudConfig.vcrProxyHost != null) {
       connectionPolicy.setProxy(cloudConfig.vcrProxyHost, cloudConfig.vcrProxyPort);
-      //ANKUR TODO
-      //connectionPolicy.setHandleServiceUnavailableFromProxy(true);
     }
     // TODO: test option to set connectionPolicy.setEnableEndpointDiscovery(false);
     asyncDocumentClient = new AsyncDocumentClient.Builder().withServiceEndpoint(azureCloudConfig.cosmosEndpoint)
@@ -151,7 +149,6 @@ class AzureCloudDestination implements CloudDestination {
    * @param cosmosCollectionLink the CosmosDB collection link to use.
    * @param clusterName the name of the Ambry cluster.
    * @param azureMetrics the {@link AzureMetrics} to use.
-   * @throws CloudStorageException if the destination could not be created.
    */
   AzureCloudDestination(CloudStorageAccount azureAccount, AsyncDocumentClient asyncDocumentClient,
       String cosmosCollectionLink, String clusterName, AzureMetrics azureMetrics) {
@@ -309,15 +306,14 @@ class AzureCloudDestination implements CloudDestination {
       metadataList = getBlobMetadataChunked(blobIds);
     }
 
-    return metadataList.stream().collect(Collectors.toMap(m -> m.getId(), Function.identity()));
+    return metadataList.stream().collect(Collectors.toMap(CloudBlobMetadata::getId, Function.identity()));
   }
 
   private List<CloudBlobMetadata> getBlobMetadataChunked(List<BlobId> blobIds) throws CloudStorageException {
     if (blobIds.isEmpty() || blobIds.size() > ID_QUERY_BATCH_SIZE) {
       throw new IllegalArgumentException("Invalid input list size: " + blobIds.size());
     }
-    String quotedBlobIds =
-        String.join(",", blobIds.stream().map(s -> '"' + s.getID() + '"').collect(Collectors.toList()));
+    String quotedBlobIds = blobIds.stream().map(s -> '"' + s.getID() + '"').collect(Collectors.joining(","));
     String query = String.format(BATCH_ID_QUERY_TEMPLATE, quotedBlobIds);
     String partitionPath = blobIds.get(0).getPartition().toPathString();
     try {
@@ -392,7 +388,7 @@ class AzureCloudDestination implements CloudDestination {
    * @param fieldName The metadata field to modify.
    * @param value The new value.
    * @return {@code true} if the udpate succeeded, {@code false} if the metadata record was not found.
-   * @throws DocumentClientException
+   * @throws CloudStorageException
    */
   private boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value) throws CloudStorageException {
     Objects.requireNonNull(blobId, "BlobId cannot be null");
@@ -424,7 +420,6 @@ class AzureCloudDestination implements CloudDestination {
       }
 
       ResourceResponse<Document> response = cosmosDataAccessor.readMetadata(blobId);
-      //CloudBlobMetadata blobMetadata = response.getResource().toObject(CloudBlobMetadata.class);
       Document doc = response.getResource();
       if (doc == null) {
         logger.warn("Blob metadata record not found: {}", blobId.getID());
@@ -510,7 +505,8 @@ class AzureCloudDestination implements CloudDestination {
    * @param blobId the {@link BlobId} that needs a container.
    * @param autoCreate flag indicating whether to create the container if it does not exist.
    * @return the created {@link CloudBlobContainer}.
-   * @throws Exception
+   * @throws URISyntaxException
+   * @throws StorageException
    */
   private CloudBlobContainer getContainer(BlobId blobId, boolean autoCreate)
       throws URISyntaxException, StorageException {
@@ -594,7 +590,7 @@ class AzureCloudDestination implements CloudDestination {
    * @return the name of the Azure storage container where blobs in the specified partition are stored.
    * @param partitionPath the lexical path of the Ambry partition.
    */
-  String getAzureContainerName(String partitionPath) {
+  private String getAzureContainerName(String partitionPath) {
     // Include Ambry cluster name in case the same storage account is used to backup multiple clusters.
     // Azure requires container names to be all lower case
     String rawContainerName = clusterName + SEPARATOR + partitionPath;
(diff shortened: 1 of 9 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
```diff
@@ -166,96 +161,34 @@ void testConnectivity() {
...
 
-  /**
-   * Run the supplied DocumentClient action. If CosmosDB returns status 429 (TOO_MANY_REQUESTS),
-   * retry after the requested wait period, up to the configured retry limit.
-   * @param operation the DocumentClient resource operation to execute, wrapped in a {@link Callable}.
-   * @param timer the {@link Timer} to use to record execution time (excluding waiting).
-   * @return the {@link ResourceResponse} returned by the operation, if successful.
-   * @throws DocumentClientException if Cosmos returns a different error status, or if the retry limit is reached.
-   */
-  private ResourceResponse<Document> retryOperationWithThrottling(Callable<ResourceResponse<Document>> operation,
```
ankagrawal: Note that some default retry for http 429 happens implicity in the async SDK (As per [https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/cosmos-db/performance-tips-async-java.md] ). So I decided to do away with our own throttling logic.

_

> The SDKs all implicitly catch this response, respect the server-specified retry-after header, and retry the request. Unless your account is being accessed concurrently by multiple clients, the next retry will succeed.
> 
> If you have more than one client cumulatively operating consistently above the request rate, the default retry count currently set to 9 internally by the client may not suffice; in this case, the client throws a DocumentClientException with status code 429 to the application. The default retry count can be changed by using setRetryOptions on the ConnectionPolicy instance. By default, the DocumentClientException with status code 429 is returned after a cumulative wait time of 30 seconds if the request continues to operate above the request rate. This occurs even when the current retry count is less than the max retry count, be it the default of 9 or a user-defined value.

_

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -127,16 +125,18 @@
...
+      connectionPolicy.setProxy(cloudConfig.vcrProxyHost, cloudConfig.vcrProxyPort);
     }
     // TODO: test option to set connectionPolicy.setEnableEndpointDiscovery(false);
-    documentClient = new DocumentClient(azureCloudConfig.cosmosEndpoint, azureCloudConfig.cosmosKey, connectionPolicy,
-        ConsistencyLevel.Session);
-    cosmosDataAccessor = new CosmosDataAccessor(documentClient, azureCloudConfig, azureMetrics);
+    asyncDocumentClient = new AsyncDocumentClient.Builder().withServiceEndpoint(azureCloudConfig.cosmosEndpoint)
+        .withMasterKeyOrResourceToken(azureCloudConfig.cosmosKey)
+        .withConnectionPolicy(connectionPolicy)
+        .withConsistencyLevel(ConsistencyLevel.Session)
```
lightningrob: We probably need to change this but fine to keep it for now.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -127,16 +125,18 @@
     ConnectionPolicy connectionPolicy = new ConnectionPolicy();
     if (azureCloudConfig.cosmosDirectHttps) {
       logger.info("Using CosmosDB DirectHttps connection mode");
-      connectionPolicy.setConnectionMode(ConnectionMode.DirectHttps);
+      connectionPolicy.setConnectionMode(ConnectionMode.Direct);
     }
     if (cloudConfig.vcrProxyHost != null) {
-      connectionPolicy.setProxy(new HttpHost(cloudConfig.vcrProxyHost, cloudConfig.vcrProxyPort));
-      connectionPolicy.setHandleServiceUnavailableFromProxy(true);
```
lightningrob: Did this method go away?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -309,15 +306,14 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
+    return metadataList.stream().collect(Collectors.toMap(CloudBlobMetadata::getId, Function.identity()));
   }
 
   private List<CloudBlobMetadata> getBlobMetadataChunked(List<BlobId> blobIds) throws CloudStorageException {
     if (blobIds.isEmpty() || blobIds.size() > ID_QUERY_BATCH_SIZE) {
       throw new IllegalArgumentException("Invalid input list size: " + blobIds.size());
     }
-    String quotedBlobIds =
-        String.join(",", blobIds.stream().map(s -> '"' + s.getID() + '"').collect(Collectors.toList()));
+    String quotedBlobIds = blobIds.stream().map(s -> '"' + s.getID() + '"').collect(Collectors.joining(","));
```
lightningrob: nice

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -392,7 +388,7 @@ private void filterOutLastReadBlobs(List<CloudBlobMetadata> cloudBlobMetadataLis
    * @param fieldName The metadata field to modify.
    * @param value The new value.
    * @return {@code true} if the udpate succeeded, {@code false} if the metadata record was not found.
-   * @throws DocumentClientException
+   * @throws CloudStorageException
```
lightningrob: good catch.  please add "if the update fails."

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureCloudDestinationTest.java
```diff
@@ -502,62 +501,36 @@ public void testAzureConnection() throws Exception {
     }
   }
 
-  /** Test initializing AzureCloudDestination with a proxy */
-  @Test
-  public void testProxy() throws Exception {
```
lightningrob: Why was this deleted?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -127,16 +125,18 @@
...
-      connectionPolicy.setHandleServiceUnavailableFromProxy(true);
+      connectionPolicy.setProxy(cloudConfig.vcrProxyHost, cloudConfig.vcrProxyPort);
     }
     // TODO: test option to set connectionPolicy.setEnableEndpointDiscovery(false);
-    documentClient = new DocumentClient(azureCloudConfig.cosmosEndpoint, azureCloudConfig.cosmosKey, connectionPolicy,
-        ConsistencyLevel.Session);
-    cosmosDataAccessor = new CosmosDataAccessor(documentClient, azureCloudConfig, azureMetrics);
+    asyncDocumentClient = new AsyncDocumentClient.Builder().withServiceEndpoint(azureCloudConfig.cosmosEndpoint)
+        .withMasterKeyOrResourceToken(azureCloudConfig.cosmosKey)
+        .withConnectionPolicy(connectionPolicy)
```
lightningrob: should we make the number of retries configurable?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -359,6 +363,14 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
     }
   }
 
+  /**
+   * Getter for {@link AsyncDocumentClient} object.
+   * @return {@link AsyncDocumentClient} object.
+   */
+  public AsyncDocumentClient getAsyncDocumentClient() {
```
lightningrob: This can be package private, no?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -309,15 +310,14 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
...
   private List<CloudBlobMetadata> getBlobMetadataChunked(List<BlobId> blobIds) throws CloudStorageException {
     if (blobIds.isEmpty() || blobIds.size() > ID_QUERY_BATCH_SIZE) {
       throw new IllegalArgumentException("Invalid input list size: " + blobIds.size());
     }
-    String quotedBlobIds =
-        String.join(",", blobIds.stream().map(s -> '"' + s.getID() + '"').collect(Collectors.toList()));
+    String quotedBlobIds = blobIds.stream().map(s -> '"' + s.getID() + '"').collect(Collectors.joining(","));
     String query = String.format(BATCH_ID_QUERY_TEMPLATE, quotedBlobIds);
     String partitionPath = blobIds.get(0).getPartition().toPathString();
     try {
```
cgtz: This may be an action item for the future, but we should consider if we can get rid of a lot of the string concat/formatting when running these queries.

There is an alternate constructor for `SqlQuerySpec` that is used in `getDeadBlobs`: `SqlQuerySpec(String queryText, SqlParameterCollection parameters)`

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
```diff
@@ -16,102 +16,96 @@
...
-  public CosmosDataAccessor(DocumentClient documentClient, AzureCloudConfig azureCloudConfig,
+  CosmosDataAccessor(AsyncDocumentClient asyncDocumentClient, AzureCloudConfig azureCloudConfig,
       AzureMetrics azureMetrics) {
-    this(documentClient, azureCloudConfig.cosmosCollectionLink, azureCloudConfig.cosmosMaxRetries, azureMetrics);
+    this(asyncDocumentClient, azureCloudConfig.cosmosCollectionLink, azureMetrics);
   }
 
   /** Test constructor */
-  public CosmosDataAccessor(DocumentClient documentClient, String cosmosCollectionLink, int maxRetries,
+  public CosmosDataAccessor(AsyncDocumentClient asyncDocumentClient, String cosmosCollectionLink,
```
cgtz: why does the test constructor need to be public?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
```diff
@@ -166,96 +161,34 @@ void testConnectivity() {
...
-
   /**
    * Utility method to call a Cosmos method and extract any nested DocumentClientException.
    * @param action the action to call.
    * @return the result of the action.
-   * @throws Exception
+   * @throws DocumentClientException
    */
-  private Object executeCosmosAction(Callable<? extends Object> action) throws Exception {
+  private ResourceResponse<Document> executeCosmosAction(Callable<? extends Object> action, Timer timer)
```
cgtz: would making the parameter a `Callable<? extends ResourceResponse<? extends Document>>`, or `Callable<ResourceResponse<Document>>` work?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
```diff
@@ -166,96 +161,34 @@ void testConnectivity() {
...
-    } catch (DocumentClientException dex) {
-      throw dex;
-    } catch (IllegalStateException ex) {
-      if (ex.getCause() instanceof DocumentClientException) {
-        throw (DocumentClientException) ex.getCause();
-      } else {
-        throw ex;
+      Timer.Context operationTimer = timer.time();
+      resourceResponse = (ResourceResponse<Document>) action.call();
+      operationTimer.stop();
```
cgtz: put the stop in a finally clause?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
```diff
@@ -166,96 +161,34 @@ void testConnectivity() {
...
+    } catch (RuntimeException rex) {
+      if (rex.getCause() instanceof DocumentClientException) {
+        throw (DocumentClientException) rex.getCause();
       }
+      throw (RuntimeException) rex.getCause();
```
cgtz: this else case should just `throw rex`, right? since the cause may not necessarily be a `RuntimeException`

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
cgtz: Nevermind, I found that the client doesn't support making list `SqlParameter` that could be used for in clauses: https://github.com/Azure/azure-cosmosdb-java/issues/60

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
ankagrawal: done.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
ankagrawal: done.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
ankagrawal: done.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
ankagrawal: True. I dont know why I did that. fixed.

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureTestUtils.java
```diff
@@ -43,10 +53,52 @@ public static Document createDocumentFromCloudBlobMetadata(CloudBlobMetadata clo
...
+    when(mockResourceResponse.getResource()).thenReturn(metadataDoc);
+    return mockResponse;
+  }
+
+  /**
+   * Utility to mock the query call chain of {@link AsyncDocumentClient} such that query returns {@code documentList}.
+   * @param documentList {@link List <Document>} of documents to return from mocked call.
+   * @param mockResponse {@link Observable} mocked response.
+   */
+  static void mockObservableForQuery(List<Document> documentList, Observable<FeedResponse<Document>> mockResponse) {
```
lightningrob: Why doesn't this method create and return the mockResponse like the others?

PR #1323, opened 2019-11-27
Fix CryptoService to work with CompositeByteBuf

justinlin-linkedin:
PutOperation would read multiple ByteBuf as a Composite from NettyServer layer. If the blob should be encrypted, then the crypto service has to work with CompositeByteBuf. This PR adds such support.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
+++ b/ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
@@ -14,6 +14,7 @@
 package com.github.ambry.router;
 
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.Unpooled;
 import java.nio.ByteBuffer;
 import java.security.GeneralSecurityException;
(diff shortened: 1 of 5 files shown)


ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
```diff
@@ -69,8 +79,17 @@ default ByteBuf encrypt(ByteBuf toEncrypt, T key) throws GeneralSecurityExceptio
    * @throws {@link GeneralSecurityException} on any exception with decryption
    */
   default ByteBuf decrypt(ByteBuf toDecrypt, T key) throws GeneralSecurityException {
-    ByteBuffer decrypted = decrypt(toDecrypt.nioBuffer(), key);
-    return Unpooled.wrappedBuffer(decrypted);
+    if (toDecrypt.nioBufferCount() == 1) {
```
cgtz: are there any unit tests that cover the default methods?

ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
```diff
@@ -69,8 +79,17 @@ default ByteBuf encrypt(ByteBuf toEncrypt, T key) throws GeneralSecurityExceptio
...
+      return Unpooled.wrappedBuffer(decrypt(toDecrypt.nioBuffer(), key));
+    } else {
+      ByteBuf temp = ByteBufAllocator.DEFAULT.heapBuffer(toDecrypt.readableBytes());
+      try {
+        temp.writeBytes(toDecrypt);
+        return Unpooled.wrappedBuffer(decrypt(temp.nioBuffer(), key));
```
zzmao: Is it possible to create common util function to handle the logic in these two default method?

PR #1325, opened 2019-11-27
Change AsyncWritableChannel to support netty ByteBuf

justinlin-linkedin:
Adding a default function to AsyncWritableChannel to support netty ByteBuf.

Adding this method as a default function because of backward compatibility. By doing so, all the implementations would be compatible with the new interface.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
+++ b/ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
@@ -17,6 +17,7 @@ import io.netty.buffer.ByteBuf;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channel;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 
@@ -60,39 +61,35 @@ public interface AsyncWritableChannel extends Channel {
    *         written).
    */
   default Future<Long> write(ByteBuf src, Callback<Long> callback) {
+    if (src == null) {
+      throw new IllegalArgumentException("Source ByteBuf cannot be null");
+    }
     int numBuffers = src.nioBufferCount();
     if (numBuffers == 1) {
       return write(src.nioBuffer(), (result, exception) -> {
         if (result != 0) {
           src.readerIndex(src.readerIndex() + (int) result.longValue());
         }
-        callback.onCompletion(result, exception);
+        if (callback != null) {
+          callback.onCompletion(result, exception);
+        }
       });
     } else {
       ByteBuffer[] buffers = src.nioBuffers();
       AtomicLong size = new AtomicLong(0);
       AtomicInteger index = new AtomicInteger(0);
-      FutureResult<Long> futureResult = new FutureResult<Long>();
-      Callback<Long> cb = new Callback<Long>() {
-        @Override
-        public void onCompletion(Long result, Exception exception) {
-          index.addAndGet(1);
-          size.addAndGet(result);
-          if (result != 0) {
-            src.readerIndex(src.readerIndex() + (int) result.longValue());
-          }
-          if (exception != null) {
-            futureResult.done(size.get(), exception);
-            if (callback != null) {
-              callback.onCompletion(size.get(), exception);
-            }
-          } else {
-            if (index.get() == buffers.length) {
-              futureResult.done(size.get(), exception);
-              if (callback != null) {
-                callback.onCompletion(size.get(), exception);
-              }
-            }
+      AtomicBoolean callbackInvoked = new AtomicBoolean(false);
+      FutureResult<Long> futureResult = new FutureResult<>();
+      Callback<Long> cb = (result, exception) -> {
+        index.addAndGet(1);
+        size.addAndGet(result);
+        if (result != 0) {
+          src.readerIndex(src.readerIndex() + (int) result.longValue());
+        }
+        if ((exception != null || index.get() == buffers.length) && callbackInvoked.compareAndSet(false, true)) {
+          futureResult.done(size.get(), exception);
+          if (callback != null) {
+            callback.onCompletion(size.get(), exception);
           }
         }
       };
--- a/ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
+++ b/ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
@@ -65,15 +65,20 @@ public interface AsyncWritableChannel extends Channel {
       throw new IllegalArgumentException("Source ByteBuf cannot be null");
     }
     int numBuffers = src.nioBufferCount();
-    if (numBuffers == 1) {
-      return write(src.nioBuffer(), (result, exception) -> {
-        if (result != 0) {
-          src.readerIndex(src.readerIndex() + (int) result.longValue());
-        }
-        if (callback != null) {
-          callback.onCompletion(result, exception);
-        }
-      });
+    Callback<Long> singleBufferCallback = (result, exception) -> {
+      if (result != 0) {
+        src.readerIndex(src.readerIndex() + (int) result.longValue());
+      }
+      if (callback != null) {
+        callback.onCompletion(result, exception);
+      }
+    };
+    if (numBuffers < 1) {
+      ByteBuffer byteBuffer = ByteBuffer.allocate(src.readableBytes());
+      src.getBytes(src.readerIndex(), byteBuffer);
+      return write(byteBuffer, singleBufferCallback);
+    } else if (numBuffers == 1) {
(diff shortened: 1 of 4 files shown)


ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
```diff
@@ -44,5 +47,59 @@
...
+        if (result != 0) {
+          src.readerIndex(src.readerIndex() + (int) result.longValue());
+        }
+        callback.onCompletion(result, exception);
+      });
+    } else {
+      ByteBuffer[] buffers = src.nioBuffers();
+      AtomicLong size = new AtomicLong(0);
+      AtomicInteger index = new AtomicInteger(0);
+      FutureResult<Long> futureResult = new FutureResult<Long>();
```
cgtz: `new FutureResult<>()`

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
```diff
@@ -44,5 +47,59 @@
...
+      Callback<Long> cb = new Callback<Long>() {
+        @Override
+        public void onCompletion(Long result, Exception exception) {
+          index.addAndGet(1);
+          size.addAndGet(result);
+          if (result != 0) {
+            src.readerIndex(src.readerIndex() + (int) result.longValue());
+          }
+          if (exception != null) {
+            futureResult.done(size.get(), exception);
```
cgtz: this could result in the callback and futureResult being called multiple times if multiple chunks got an exception.

How about:
```java
        if (exception != null || index.get() == buffers.length) {
          if (callbackInvoked.compareAndSet(false, true)) {
            futureResult.done(size.get(), exception);
            if (callback != null) {
              callback.onCompletion(size.get(), exception);
            }
          }
        }
```

ambry-commons/src/main/java/com.github.ambry.commons/ByteBufferAsyncWritableChannel.java
```diff
@@ -142,9 +168,71 @@ public ByteBuffer getNextChunk() throws InterruptedException {
    * @throws InterruptedException if the wait for a chunk is interrupted.
    */
   public ByteBuffer getNextChunk(long timeoutInMs) throws InterruptedException {
-    ByteBuffer chunkBuf = null;
     if (isOpen()) {
-      chunkBuf = getChunkBuffer(chunks.poll(timeoutInMs, TimeUnit.MILLISECONDS));
```
cgtz: Call to getNextByteBuf instead here and in `getNextChunk()`?

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
```diff
@@ -44,5 +47,59 @@
...
+   *         written).
+   */
+  default Future<Long> write(ByteBuf src, Callback<Long> callback) {
+    int numBuffers = src.nioBufferCount();
+    if (numBuffers == 1) {
+      return write(src.nioBuffer(), (result, exception) -> {
+        if (result != 0) {
+          src.readerIndex(src.readerIndex() + (int) result.longValue());
+        }
+        callback.onCompletion(result, exception);
```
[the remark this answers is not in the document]
justinlin-linkedin: good catch, will update with the null check.

With regard to the fact the readerIndex might be updated after future is completed. There is really no good way of dealing with it even when the nioBufferCount is greater than 1. Do you have any suggestions?

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
justinlin-linkedin: this is better, will update.

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
```diff
@@ -44,5 +47,59 @@
...
+              futureResult.done(size.get(), exception);
+              if (callback != null) {
+                callback.onCompletion(size.get(), exception);
+              }
+            }
+          }
+        }
+      };
+      for (int i = 0; i < buffers.length; i++) {
+        write(buffers[i], cb);
```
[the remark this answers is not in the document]
justinlin-linkedin: The comments of AsyncWritableChannel#write(ByteBuffer, Callback<Long>) states that " Every single write is guaranteed to be acknowledged as either succeeded or failed even if the channel is closed. Further, writes will be acknowledged in the same order that they were received." I take this as the guarantee that the order of write will not be messed up in the implementation?

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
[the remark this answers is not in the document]
cgtz: I think for this we could do what was done in the else case and create a secondary `FutureResult` object:

```java
  FutureResult<Long> futureResult = new FutureResult<>();
  Callback<Long> singleBufferCallback = (result, exception) -> {
      if (result != 0) {
        src.readerIndex(src.readerIndex() + (int) result.longValue());
      }
      futureResult.done(result, exception);
      if (callback != null) {
        callback.onCompletion(result, exception);
      }
  };

if (numBuffers <1) {
  ...
  write(byteBuffer, singleBufferCallback);
  return futureResult
```

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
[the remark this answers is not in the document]
cgtz: Yep, I think you're right. Most likely the production impls all satisfy these requirements.

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
```diff
@@ -44,5 +48,60 @@
...
+   * Write data in {@code src} to the channel and the {@code callback} will be invoked once the write succeeds or fails.
+   * This method is the counterpart for {@link ByteBuf}. It shares the same guarantee as the {@link #write(ByteBuffer, Callback)}.
+   * Whoever implements this interface, shouldn't release the {@code src}. If releasing is expected after finishing writing
+   * to channel, please releas this {@link ByteBuf} in the callback method.
+   * @param src The data taht needs to be written to the channel.
+   * @param callback The {@link Callback} that will be invoked once the write succeeds/fails. This can be null.
+   * @return a {@link Future} that will eventually contain the result of the write operation (the number of bytes
+   *         written).
+   */
+  default Future<Long> write(ByteBuf src, Callback<Long> callback) {
```
cgtz: Could you write a unit test for this default method to check that it works as expected in all of the 3 cases nioBufferCount cases?

ambry-commons/src/test/java/com.github.ambry.commons/CopyingAsyncWritableChannelTest.java
```diff
@@ -56,6 +73,41 @@ public void basicsTest() throws Exception {
     writeAndCheckCallback(ByteBuffer.allocate(0), channel, 0, ClosedChannelException.class, null);
   }
 
+  /**
+   * Test that {@link CopyingAsyncWritableChannel} behaves as expected: chunks are copied, callback completed
+   * immediately after {@link CopyingAsyncWritableChannel#write} method completes.
+   */
+  @Test
+  public void basicsTestWithNettyByteBuf() throws Exception {
+    for (boolean useCompositeByteBuf : Arrays.asList(false, true)) {
```
cgtz: Is this meant to test the default method in `AsyncWritableChannel`? We could also consider overriding the write(ByteBuf) method for `CopyingAsyncWritableChannel`, which should be relatively straightforward since it just copies data to byte arrays internally.

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
justinlin-linkedin: the case where nioBufferCount is -1 is pretty hard to construct so I will not do that. And I have unit tests for CopyingAsyncWritableChannel, which didn't have it's owner implementation of write(ByteBuf), that would serve to test the write(ByteBuf) as default method.

ambry-commons/src/test/java/com.github.ambry.commons/CopyingAsyncWritableChannelTest.java
justinlin-linkedin: you are right. The implementations are driven by the need. There is no need for CopyingAsyncWritableChannel to implement write(ByteBuf) for now. But later we will have to change all the implementations for it.

ambry-api/src/main/java/com.github.ambry/router/AsyncWritableChannel.java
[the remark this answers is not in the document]
justinlin-linkedin: will update.

PR #1331, opened 2019-12-06
Recovery partition override config

ankagrawal:
This PR implements the following:

- Add vcr.recovery.partitions config to override partition assignment for recovery. This will allow to recover only specified partitions with both helix and static cluster.

- Fix assigning token on adding cloud replica so that cloud replication can restart from where it left off. In replication from vcr, datanode id and replica id do not matter in replication token and their check should be ignored while restarting replication.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
@@ -195,7 +195,6 @@ public class ReplicationConfig {
         verifiableProperties.getShortInRange("replication.metadata.request.version", (short) 1, (short) 1, (short) 2);
     replicationEnabledWithVcrCluster = verifiableProperties.getBoolean("replication.enabled.with.vcr.cluster", false);
     String vcrRecoveryPartitions = verifiableProperties.getString("replication.vcr.recovery.partitions", "");
-    replicationVcrRecoveryPartitions =
-        vcrRecoveryPartitions.isEmpty() ? new HashSet<>() : Utils.splitString(vcrRecoveryPartitions, ",", HashSet::new);
+    replicationVcrRecoveryPartitions = Utils.splitString(vcrRecoveryPartitions, ",", HashSet::new);
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
@@ -145,6 +145,14 @@ public class ReplicationConfig {
   @Default("false")
   public final boolean replicationEnabledWithVcrCluster;
 
+  /**
+   * Comma separated list of partitions to recover.
+   * This config will ignore the partition leadership information for recovering partitions.
+   */
+  @Config("vcr.recovery.partitions")
+  @Default("")
+  public final String vcrRecoveryPartitions;
+
   public ReplicationConfig(VerifiableProperties verifiableProperties) {
 
     replicationStoreTokenFactory =
@@ -181,5 +189,6 @@ public class ReplicationConfig {
     replicaMetadataRequestVersion =
         verifiableProperties.getShortInRange("replication.metadata.request.version", (short) 1, (short) 1, (short) 2);
     replicationEnabledWithVcrCluster = verifiableProperties.getBoolean("replication.enabled.with.vcr.cluster", false);
+    vcrRecoveryPartitions = verifiableProperties.getString("vcr.recovery.partitions", "");
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
@@ -149,9 +149,9 @@ public class ReplicationConfig {
    * Comma separated list of partitions to recover.
    * This config will ignore the partition leadership information for recovering partitions.
    */
-  @Config("vcr.recovery.partitions")
+  @Config("replication.vcr.recovery.partitions")
   @Default("")
-  public final String vcrRecoveryPartitions;
+  public final String replicationVcrRecoveryPartitions;
 
   public ReplicationConfig(VerifiableProperties verifiableProperties) {
 
@@ -189,6 +189,6 @@ public class ReplicationConfig {
     replicaMetadataRequestVersion =
         verifiableProperties.getShortInRange("replication.metadata.request.version", (short) 1, (short) 1, (short) 2);
     replicationEnabledWithVcrCluster = verifiableProperties.getBoolean("replication.enabled.with.vcr.cluster", false);
-    vcrRecoveryPartitions = verifiableProperties.getString("vcr.recovery.partitions", "");
+    replicationVcrRecoveryPartitions = verifiableProperties.getString("vcr.recovery.partitions", "");
   }
 }
--- a/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
@@ -13,6 +13,11 @@
  */
 package com.github.ambry.config;
 
+import com.github.ambry.utils.Utils;
+import java.util.HashSet;
+import java.util.Set;
+
+
 /**
  * The configs for the replication layer
  */
@@ -151,7 +156,7 @@ public class ReplicationConfig {
    */
   @Config("replication.vcr.recovery.partitions")
   @Default("")
-  public final String replicationVcrRecoveryPartitions;
+  public final Set<String> replicationVcrRecoveryPartitions;
 
   public ReplicationConfig(VerifiableProperties verifiableProperties) {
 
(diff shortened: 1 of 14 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterAgentsFactory.java
```diff
@@ -79,9 +81,13 @@ public ClusterParticipant getClusterParticipant() throws IOException {
         private final List<PartitionStateChangeListener> listeners = new LinkedList<>();
 
         @Override
-        public void participate(List<AmbryHealthReport> ambryHealthReports) {
+        public void participate(List<AmbryHealthReport> ambryHealthReports, DataNodeId currentNode) {
+          Set<String> partitionsOnNode = getClusterMap().getReplicaIds(currentNode)
+              .stream()
+              .map(r -> r.getPartitionId().toPathString())
+              .collect(Collectors.toSet());
```
jsjtzyy: To get `currentNode`, we can use following code:
``` DataNodeId currentNode = getClusterMap().getDataNodeId(clusterMapConfig.clusterMapHostName, clusterMapConfig.clusterMapPort);```

File: ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java [hunk 2]
```diff
@@ -334,6 +374,10 @@ public void onPartitionBecomeStandbyFromBootstrap(String partitionName) {
     public void onPartitionBecomeLeaderFromStandby(String partitionName) {
       logger.info("Partition state change notification from Standby to Leader received for partition {}",
           partitionName);
+      if (!partitionsToReplicate.isEmpty() && !partitionsToReplicate.contains(partitionName)) {
```
lightningrob: Since you do this multiple times, suggest utility like:
boolean shouldReplicatePartition(String partitionName)
which does the check and the logging.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -215,6 +219,42 @@ private void addCloudReplica(String partitionName) throws ReplicationException {
...
+    int tokenReloadFailureCount = 0;
+    List<RemoteReplicaInfo.ReplicaTokenInfo> tokenInfos = persistor.retrieve(localReplicaId.getMountPath());
+    if (tokenInfos.size() != 0) {
+      for (RemoteReplicaInfo remoteReplicaInfo : remoteReplicaInfos) {
+        boolean tokenReloaded = false;
+        for (RemoteReplicaInfo.ReplicaTokenInfo tokenInfo : tokenInfos) {
+          // Note that in case of cloudReplicaTokens, the actual remote data node might not match as the node is chosen at
+          // random during initialization. So its enough to just match the partitionId in the token so that replication
+          // can start from cloud from where it left off.
+          if (tokenInfo.getReplicaInfo()
```
lightningrob: I don't see the need to override the whole method which is mostly the same.  Can't you override isTokenForRemoteReplicaInfo() instead?

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -113,6 +115,8 @@ public CloudToStoreReplicationManager(ReplicationConfig replicationConfig, Clust
         new DiskTokenPersistor(cloudReplicaTokenFileName, mountPathToPartitionInfos, replicationMetrics, clusterMap,
             tokenHelper, storeManager);
     this.localPartitionNameToPartition = mapPartitionNameToPartition(clusterMap, currentNode);
+    this.partitionsToReplicate = replicationConfig.vcrRecoveryPartitions.isEmpty() ? new HashSet<>()
+        : Arrays.stream(replicationConfig.vcrRecoveryPartitions.split(",")).collect(Collectors.toSet());
```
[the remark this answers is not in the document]
cgtz: If you want to make `Utils.splitString` more flexible, since it seems like you want to put the parts into a set, you can extend the util method to something like this:
```java
  public static ArrayList<String> splitString(String data, String delimiter) {
    return splitString(data, delimiter, ArrayList::new);
  }

  public static <C extends Collection<String>> C splitString(String data, String delimiter,
      Supplier<C> collectionFactory) {
    if (data == null) {
      throw new IllegalArgumentException("Passed in string is null ");
    }
    return Arrays.stream(data.split(delimiter))
        .filter(s -> !s.isEmpty())
        .collect(Collectors.toCollection(collectionFactory));
  }
```

```java
Utils.splitString(vcrRecoveryPartitions, ",", HashSet::new)
```

ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
```diff
@@ -181,5 +189,6 @@ public ReplicationConfig(VerifiableProperties verifiableProperties) {
     replicaMetadataRequestVersion =
         verifiableProperties.getShortInRange("replication.metadata.request.version", (short) 1, (short) 1, (short) 2);
     replicationEnabledWithVcrCluster = verifiableProperties.getBoolean("replication.enabled.with.vcr.cluster", false);
+    vcrRecoveryPartitions = verifiableProperties.getString("vcr.recovery.partitions", "");
```
cgtz: +1 to Yingyi's suggestion about doing the string splitting here and making the config a `Set<String>`

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterAgentsFactory.java
```diff
@@ -79,9 +81,13 @@ public ClusterParticipant getClusterParticipant() throws IOException {
...
           for (PartitionStateChangeListener listener : listeners) {
-            for (String partitionName : partitionLayout.getAllPartitionNames()) {
+            for (String partitionName : partitionsOnNode) {
```
cgtz: What are the implications of calling this callback on the partitions not present on this node?

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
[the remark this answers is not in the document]
ankagrawal: I thought of making this a set and split in the config class itself, but my concern with making anything in config a set is that I will have to write same logic in the closed source code to get the string config and then split it there again. Thats why I moved the split to where it is really used.
I like the suggestion of extending Utils.splitString though.

ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
ankagrawal: see reply below for CloudToReplicationManager.

ambry-utils/src/main/java/com.github.ambry.utils/Utils.java
```diff
@@ -851,12 +851,26 @@ public static DataInputStream createDataInputStreamFromBuffer(Object buffer, boo
...
+  }
+
+  /**
+   * Split the input string "data" using the delimiter and return as collection of strings for the slices obtained.
+   * This method will ignore empty segments. That is, a call like {@code splitString(",a1,,b2,c3,", ","}} will return
+   * {@code ["a1","b2","c3]}. Since this is used for reading list-style configs, this is usually the desired behavior.
+   * @param data the string to split.
+   * @param delimiter the delimiter for splitting.
+   * @param collectionFactory functional interface to get the required collection.
+   * @return a mutable list of items.
```
jsjtzyy: minor: a mutable collection of items

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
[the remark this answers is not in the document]
jsjtzyy: I don't quite understand "writing same logic in closed source code". Let's discuss this offline.

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
```diff
@@ -215,6 +215,24 @@ private void addCloudReplica(String partitionName) throws ReplicationException {
...
+   * Check if a token is for the given {@link RemoteReplicaInfo} based on partition id.
+   * @param remoteReplicaInfo The remoteReplicaInfo to check.
+   * @param tokenInfo The tokenInfo to check.
+   * @return true if partition id matches. false otherwise.
+   */
+  @Override
+  protected boolean isTokenForRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInfo,
+      RemoteReplicaInfo.ReplicaTokenInfo tokenInfo) {
+    // Note that in case of cloudReplicaTokens, the actual remote vcr node might not match as the vcr node is chosen at
+    // random during initialization. So its enough to just match the partitionId in the token so that replication
```
lightningrob: Minor: its -> it's

ambry-api/src/main/java/com.github.ambry/config/ReplicationConfig.java
```diff
@@ -181,5 +194,8 @@ public ReplicationConfig(VerifiableProperties verifiableProperties) {
     replicaMetadataRequestVersion =
         verifiableProperties.getShortInRange("replication.metadata.request.version", (short) 1, (short) 1, (short) 2);
     replicationEnabledWithVcrCluster = verifiableProperties.getBoolean("replication.enabled.with.vcr.cluster", false);
+    String vcrRecoveryPartitions = verifiableProperties.getString("replication.vcr.recovery.partitions", "");
+    replicationVcrRecoveryPartitions =
+        vcrRecoveryPartitions.isEmpty() ? new HashSet<>() : Utils.splitString(vcrRecoveryPartitions, ",", HashSet::new);
```
cgtz: `Utils.splitString` would actually handle the empty case.

cgtz: Could we add a test case for the empty string case and the new generic method

ambry-replication/src/main/java/com.github.ambry.replication/CloudToStoreReplicationManager.java
ankagrawal: done.

PR #1334, opened 2019-12-09
Introduce ReplicaSyncUpService to help Bootstrap-To-Standby transition

jsjtzyy:
This PR introduces ReplicaSyncUpService interface and its implementation
for HelixParticipant. The service provides a way to track catchup status
between local replica and peer replicas. It contains a count down latch
that blocks replica in bootstrap or being decommissioned state until it
has caught up with peers or peers have synced up with local one.
Also this PR implements Bootstrap-To-Standby state transition logic in
replication manager and state model.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
@@ -68,6 +68,12 @@ public interface ClusterParticipant extends AutoCloseable {
       PartitionStateChangeListener partitionStateChangeListener);
 
   /**
+   * @return {@link ReplicaSyncUpService} that is used to determine new replica has caught up with peers or peer replicas
+   * have caught up with old replica that is being decommissioned.
+   */
+  ReplicaSyncUpService getReplicaSyncUpService();
+
+  /**
    * Terminate the participant.
    */
   @Override
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
@@ -68,10 +68,11 @@ public interface ClusterParticipant extends AutoCloseable {
       PartitionStateChangeListener partitionStateChangeListener);
 
   /**
-   * @return {@link ReplicaSyncUpService} that is used to determine new replica has caught up with peers or peer replicas
+   * Gets the {@link ReplicaSyncUpManager} object.
+   * @return {@link ReplicaSyncUpManager} that is used to determine new replica has caught up with peers or peer replicas
    * have caught up with old replica that is being decommissioned.
    */
-  ReplicaSyncUpService getReplicaSyncUpService();
+  ReplicaSyncUpManager getReplicaSyncUpManager();
 
   /**
    * Terminate the participant.
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpService.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpService.java
@@ -1,0 +1,70 @@
+/**
+ * Copyright 2019 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.clustermap;
+
+/**
+ * A class helps check if replicas have synced up.
+ * There are two use cases:
+ *     1. determine if new added replica has caught up with peers (occurs within BOOTSTRAP -> STANDBY transition)
+ *     2. determine peer replicas have caught up with old replica that is being decommissioned; (occurs in STANDBY ->
+ *     INACTIVE and INACTIVE -> OFFLINE transitions)
+ */
+public interface ReplicaSyncUpService {
+
+  /**
+   * Initiate bootstrap process if the replica is newly added and needs to catch up with peer ones.
+   * @param replicaId the replica to bootstrap
+   */
+  void initiateBootstrap(ReplicaId replicaId);
+
+  /**
+   * Wait until bootstrap for given replica is complete. The method is blocked on a {@link java.util.concurrent.CountDownLatch}
+   * until given replica has caught up with enough peer replicas either in local DC or remote DCs
+   * @param partitionName partition name of replica that in bootstrap state
+   * @throws InterruptedException
+   */
+  void waitBootstrapCompleted(String partitionName) throws InterruptedException;
+
+  /**
+   * Update replica lag (in byte) between two replicas (chaser and precursor)
+   * @param chaser the replica which is catching up with precursor
+   * @param precursor the replica which is in leading position
+   * @param lagInBytes replica lag bytes
+   * @return whether the lag is updated or not. If {@code false}, it means the chaser is not tracked in this service.
+   *         Either the replica has caught up and removed from service or it is an existing replica that doesn't need catchup.
+   */
+  boolean updateLagBetweenReplicas(ReplicaId chaser, ReplicaId precursor, long lagInBytes);
+
+  /**
+   * Whether given replica has synced up with its peers.
+   * @param replicaId replica to check
+   * @return {@code true} if given replica has caught up with peers or peer replicas have synced up with given replica
+   *         (this occurs when given replica is being decommissioned)
+   */
+  boolean isSyncUpComplete(ReplicaId replicaId);
+
+  /**
+   * Bootstrap on given replica is complete. This method will count down the latch.
+   * @param partitionName partition name of replica on which bootstrap completes.
+   */
+  void onBootstrapComplete(String partitionName);
+
+  /**
+   * When exception/error occurs during bootstrap. This method will count down latch and terminates bootstrap.
+   * @param partitionName partition name of replica which encounters error.
+   */
+  void onBootstrapError(String partitionName);
+
+  // TODO introduce decommission logic in sync-up service. For example, initiateDecommission(String partitionName)
+}
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -468,7 +468,7 @@ class CloudBlobStore implements Store {
 
   @Override
   public boolean isBootstrapInProgress() {
-    throw new UnsupportedOperationException("Method not supported");
+    return false;
   }
 
   @Override
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -472,6 +472,11 @@ class CloudBlobStore implements Store {
   }
 
   @Override
+  public void completeBootstrap() {
+    // no op
+  }
+
+  @Override
   public void setCurrentState(ReplicaState state) {
     throw new UnsupportedOperationException("Method not supported");
   }
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -467,6 +467,11 @@ class CloudBlobStore implements Store {
   }
 
   @Override
+  public boolean isBootstrapInProgress() {
+    throw new UnsupportedOperationException("Method not supported");
+  }
+
+  @Override
   public void setCurrentState(ReplicaState state) {
     throw new UnsupportedOperationException("Method not supported");
   }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -66,7 +66,7 @@ public class AmbryPartitionStateModel extends StateModel {
     try {
       replicaSyncUpManager.waitBootstrapCompleted(partitionName);
     } catch (InterruptedException | IllegalStateException e) {
-      logger.info("Bootstrap failed or was interrupted on partition {}", partitionName);
+      logger.error("Bootstrap failed or was interrupted on partition {}", partitionName);
       throw new StateTransitionException("Bootstrap failed or was interrupted",
           StateTransitionException.TransitionErrorCode.BootstrapFailure);
     }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -32,7 +32,7 @@ public class AmbryPartitionStateModel extends StateModel {
   private final String partitionName;
   private final PartitionStateChangeListener partitionStateChangeListener;
   private final ClusterMapConfig clusterMapConfig;
-  private final ReplicaSyncUpManager _replicaSyncUpManager;
+  private final ReplicaSyncUpManager replicaSyncUpManager;
 
   AmbryPartitionStateModel(String resourceName, String partitionName,
       PartitionStateChangeListener partitionStateChangeListener, ClusterMapConfig clusterMapConfig,
@@ -41,7 +41,7 @@ public class AmbryPartitionStateModel extends StateModel {
     this.partitionName = partitionName;
     this.partitionStateChangeListener = Objects.requireNonNull(partitionStateChangeListener);
     this.clusterMapConfig = Objects.requireNonNull(clusterMapConfig);
-    this._replicaSyncUpManager = Objects.requireNonNull(replicaSyncUpManager);
+    this.replicaSyncUpManager = Objects.requireNonNull(replicaSyncUpManager);
     StateModelParser parser = new StateModelParser();
     _currentState = parser.getInitialState(AmbryPartitionStateModel.class);
   }
@@ -64,7 +64,7 @@ public class AmbryPartitionStateModel extends StateModel {
       partitionStateChangeListener.onPartitionBecomeStandbyFromBootstrap(partitionName);
     }
     try {
-      _replicaSyncUpManager.waitBootstrapCompleted(partitionName);
+      replicaSyncUpManager.waitBootstrapCompleted(partitionName);
     } catch (InterruptedException | IllegalStateException e) {
       logger.info("Bootstrap failed or was interrupted on partition {}", partitionName);
       throw new StateTransitionException("Bootstrap failed or was interrupted",
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -65,10 +65,13 @@ public class AmbryPartitionStateModel extends StateModel {
     }
     try {
       replicaSyncUpManager.waitBootstrapCompleted(partitionName);
-    } catch (InterruptedException | IllegalStateException e) {
-      logger.error("Bootstrap failed or was interrupted on partition {}", partitionName);
+    } catch (InterruptedException e) {
+      logger.error("Bootstrap was interrupted on partition {}", partitionName);
       throw new StateTransitionException("Bootstrap failed or was interrupted",
           StateTransitionException.TransitionErrorCode.BootstrapFailure);
+    } catch (StateTransitionException e) {
+      logger.error("Bootstrap didn't complete.", e);
+      throw e;
     }
   }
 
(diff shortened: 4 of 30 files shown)


ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
```diff
@@ -67,6 +67,12 @@
   void registerPartitionStateChangeListener(StateModelListenerType listenerType,
       PartitionStateChangeListener partitionStateChangeListener);
 
+  /**
+   * @return {@link ReplicaSyncUpService} that is used to determine new replica has caught up with peers or peer replicas
```
ankagrawal: minor: lets also add a line for the method comment saying "Gets the {@link ReplicaSyncUpService} object"

ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpService.java
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Whether given replica has synced up with its peers.
+   * @param replicaId replica to check
+   * @return {@code true} if given replica has caught up with peers or peer replicas have synced up with given replica
+   *         (this occurs when given replica is being decommissioned)
+   */
+  boolean isSyncUpComplete(ReplicaId replicaId);
+
+  /**
+   * Bootstrap on given replica is complete. This method will count down the latch.
```
ankagrawal: Here and elsewhere in this interface, the comments talk about using a latch. I think whether replica complete is tracked by latch or something else is an implementation detail. Lets us not put implementation details in an interface.

ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpService.java
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * A class helps check if replicas have synced up.
+ * There are two use cases:
+ *     1. determine if new added replica has caught up with peers (occurs within BOOTSTRAP -> STANDBY transition)
+ *     2. determine peer replicas have caught up with old replica that is being decommissioned; (occurs in STANDBY ->
+ *     INACTIVE and INACTIVE -> OFFLINE transitions)
+ */
+public interface ReplicaSyncUpService {
+
```
ankagrawal: Should we create a unit test for this class? like ReplicaSyncUpServiceTest?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
```diff
@@ -0,0 +1,197 @@
...
+  private void countDownLatch(String partitionName) {
+    CountDownLatch latch = partitionToBootstrapLatch.get(partitionName);
+    if (latch == null) {
+      throw new IllegalStateException("No bootstrap latch is found for partition " + partitionName);
+    } else {
+      latch.countDown();
+    }
+  }
+
+  private class RemoteReplicaLagInfos {
```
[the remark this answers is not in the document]
jsjtzyy: yeah, `LocalReplicaLagInfo` makes more sense, will rename it.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
```diff
@@ -0,0 +1,197 @@
...
+      localDcName = localReplica.getDataNodeId().getDatacenterName();
+      for (ReplicaId peerReplica : peerReplicas) {
+        // put peer replicas into local/remote DC maps (initial value is Long.MAX_VALUE)
+        if (peerReplica.getDataNodeId().getDatacenterName().equals(localDcName)) {
+          localDcPeerReplicaAndLag.put(peerReplica, Long.MAX_VALUE);
+        } else {
+          remoteDcPeerReplicaAndLag.put(peerReplica, Long.MAX_VALUE);
+        }
+      }
+      catchupTarget = localDcPeerReplicaAndLag.size();
```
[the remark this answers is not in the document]
jsjtzyy: good idea, it gives us some flexibility. Will do.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
```diff
@@ -0,0 +1,197 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class AmbryReplicaSyncUpService implements ReplicaSyncUpService {
```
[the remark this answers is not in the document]
jsjtzyy: Sounds good to me, will rename it.

ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpService.java
lightningrob: +1

jsjtzyy: Fair point, will add details.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+  }
+
+  private class RemoteReplicaLagInfos {
+    // keep lag map here for tracking progress
+    private final ConcurrentHashMap<ReplicaId, Long> localDcPeerReplicaAndLag = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<ReplicaId, Long> remoteDcPeerReplicaAndLag = new ConcurrentHashMap<>();
+    // refer to those replicas that current replica has caught up with
+    private final Set<ReplicaId> localDcCaughtUpReplicas = ConcurrentHashMap.newKeySet();
+    private final Set<ReplicaId> remoteDcCaughtUpReplicas = ConcurrentHashMap.newKeySet();
```
lightningrob: If the catchup threshold is a total across local and remote replicas, is there any benefit to tracking them separately?  Or conversely, would it make sense to split the target into local and remote targets and require both to be true?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -437,6 +442,18 @@ public void replicate() {
...
               exchangeMetadataResponseList.add(exchangeMetadataResponse);
+              // update replication lag in ReplicaSyncUpService
+              if (replicaSyncUpService != null) {
+                ReplicaId localReplica = remoteReplicaInfo.getLocalReplicaId();
+                ReplicaId remoteReplica = remoteReplicaInfo.getReplicaId();
+                boolean updated = replicaSyncUpService.updateLagBetweenReplicas(localReplica, remoteReplica,
+                    exchangeMetadataResponse.localLagFromRemoteInBytes);
+                if (updated && replicaSyncUpService.isSyncUpComplete(localReplica)) {
+                  // complete BOOTSTRAP -> STANDBY transition
+                  remoteReplicaInfo.getLocalStore().setCurrentState(ReplicaState.STANDBY);
```
lightningrob: I'm confused - how do we know this replica was in bootstrap state before the check?  Couldn't it be primary?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -241,9 +242,24 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
...
+      // if code arrives here, it means local replica has completed OFFLINE -> BOOTSTRAP transition. We don't have to
+      // check if local replica exists or not.
+      ReplicaId localReplica = storeManager.getReplica(partitionName);
+      Store store = storeManager.getStore(localReplica.getPartitionId());
+      // 1. check if store is started
+      if (store == null) {
+        throw new StateTransitionException("Store " + partitionName + " is not started",
+            StateTransitionException.TransitionErrorCode.StoreNotStarted);
+      }
+      store.setCurrentState(ReplicaState.BOOTSTRAP);
```
lightningrob: Move this line inside the if clause

ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
jsjtzyy: added

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -466,6 +466,11 @@ public boolean isEmpty() {
     return false;
   }
 
+  @Override
+  public boolean isBootstrapInProgress() {
+    throw new UnsupportedOperationException("Method not supported");
```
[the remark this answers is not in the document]
jsjtzyy: This method within CloudBlobStore is not supposed to be called in practice but in case logic is changed I will return false directly because there is no bootstrap stage for `CloudBlobStore`.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
```diff
@@ -0,0 +1,197 @@
...
+  public void waitBootstrapCompleted(String partitionName) throws InterruptedException {
+    CountDownLatch latch = partitionToBootstrapLatch.get(partitionName);
+    if (latch == null) {
+      logger.info("Skipping bootstrap for existing partition {}", partitionName);
+    } else {
+      logger.info("Waiting for new partition {} to complete bootstrap", partitionName);
+      latch.await();
+      partitionToBootstrapLatch.remove(partitionName);
+      if (!partitionToBootstrapSuccess.remove(partitionName)) {
+        throw new IllegalStateException("Partition " + partitionName + " failed on bootstrap.");
```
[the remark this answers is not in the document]
jsjtzyy: This is needed. The latch is counted down in two cases: (1) bootstrap completes successfully, (2) exception/error occurs during replication (i.e. the disk on which new replica is crashes). For case (2) we terminate bootstrap by counting down the latch but don't mark it success. The transition in state model catches this exception and puts replica into `ERROR` state.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
```diff
@@ -0,0 +1,197 @@
...
+      }
+      logger.info("Bootstrap is complete on partition {}", partitionName);
+    }
+  }
+
+  @Override
+  public boolean updateLagBetweenReplicas(ReplicaId chaser, ReplicaId precursor, long lagInBytes) {
```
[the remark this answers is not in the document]
jsjtzyy: sure, will rename them.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
jsjtzyy: Usually, for new replica, catching up with all local peer replicas should suffice. In case that some local replicas are down (especially when it occurs in DC that only has one peer replica), new replica then is supposed to catch up with enough peers in remote DCs. We didn't separate local DC and remote DC because it usually catches up with local DC replicas first.  As mentioned before, we add remote replicas here in case local replicas are unavailable.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
jsjtzyy: As discussed offline, whether it's a bootstrap replica is determined by `updated` boolean value. I will add comment to clarify this.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: will do

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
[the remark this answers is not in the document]
jsjtzyy: added

jsjtzyy:
Comments are addressed. @ankagrawal @lightningrob gentle reminder to review. Thanks.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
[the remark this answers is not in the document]
lightningrob: Minor: I see your point; however, it seems questionable to use IllegalStateException to handle normal failures.  A different RuntimeException might be more appropriate.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
lightningrob: That makes sense.  But then separating the caught-up-replica set into local and remote doesn't seem to be add much.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -0,0 +1,213 @@
...
+      logger.info("Bootstrap is complete on partition {}", partitionName);
+    }
+  }
+
+  @Override
+  public boolean updateLagBetweenReplicas(ReplicaId source, ReplicaId target, long lagInBytes) {
+    boolean updated = false;
+    if (replicaToLagInfos.containsKey(source)) {
+      replicaToLagInfos.get(source).updateLagInfo(target, lagInBytes);
+      logger.debug(replicaToLagInfos.get(source).toString());
```
lightningrob: You need to suround this inside a "if (Logger.isDebugEnabled())" check, otherwise the expensive toString() method will get called before the logger level is checked.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
```diff
@@ -437,6 +442,19 @@ public void replicate() {
                   new ExchangeMetadataResponse(missingStoreKeys, replicaMetadataResponseInfo.getFindToken(),
                       replicaMetadataResponseInfo.getRemoteReplicaLagInBytes());
               exchangeMetadataResponseList.add(exchangeMetadataResponse);
+              // update replication lag in ReplicaSyncUpManager
+              if (replicaSyncUpManager != null) {
+                ReplicaId localReplica = remoteReplicaInfo.getLocalReplicaId();
+                ReplicaId remoteReplica = remoteReplicaInfo.getReplicaId();
+                boolean updated = replicaSyncUpManager.updateLagBetweenReplicas(localReplica, remoteReplica,
+                    exchangeMetadataResponse.localLagFromRemoteInBytes);
+                // if updated is false, it means local replica is not found in replicaSyncUpManager
```
lightningrob: and is therefore in bootstrap state.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
lightningrob: Minor: If isSyncUpComplete() can throw an expected IllegalStateException, the error code below should not be UnknownError.  Unless we lost track of why the bootstrap failed, in which case it might be okay.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterAgentsFactory.java
```diff
@@ -117,6 +117,11 @@ public void registerPartitionStateChangeListener(StateModelListenerType listener
             PartitionStateChangeListener partitionStateChangeListener) {
           listeners.add(partitionStateChangeListener);
         }
+
+        @Override
+        public ReplicaSyncUpManager getReplicaSyncUpManager() {
+          return null;
```
lightningrob: Minor (thinking out load): for optional fields like this that aren't already set, perhaps we should use Optional<>.  don't need to change it now.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManagerTest.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.stream.Collectors;
+import org.apache.helix.model.Message;
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+
+public class AmbryReplicaSyncUpManagerTest {
```
lightningrob: javadoc

ambry-clustermap/src/test/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManagerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    @Override
+    public void onPartitionBecomeStandbyFromBootstrap(String partitionName) {
+      replicaState = ReplicaState.BOOTSTRAP;
+      replicaSyncUpService.initiateBootstrap(currentReplica);
+      listenerLatch.countDown();
+    }
+
+    @Override
+    public void onPartitionBecomeLeaderFromStandby(String partitionName) {
+
```
lightningrob: Minor: can remove empty lines.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
```diff
@@ -54,10 +57,18 @@ public void onBecomeBootstrapFromOffline(Message message, NotificationContext co
...
+    logger.info("Partition {} in resource {} is becoming STANDBY from BOOTSTRAP", partitionName,
         message.getResourceName());
     if (clusterMapConfig.clustermapEnableStateModelListener) {
-      partitionStateChangeListener.onPartitionBecomeStandbyFromBootstrap(message.getPartitionName());
+      partitionStateChangeListener.onPartitionBecomeStandbyFromBootstrap(partitionName);
+    }
+    try {
+      replicaSyncUpManager.waitBootstrapCompleted(partitionName);
+    } catch (InterruptedException | IllegalStateException e) {
+      logger.info("Bootstrap failed or was interrupted on partition {}", partitionName);
```
lightningrob: log exceptions at ERROR level.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManager.java
jsjtzyy: fixed, thanks for the suggestion.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
jsjtzyy: added

ambry-clustermap/src/test/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManagerTest.java
jsjtzyy: added

ambry-clustermap/src/test/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManagerTest.java
jsjtzyy: removed

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterAgentsFactory.java
jsjtzyy: Fair point, would probably change this in future PR.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
jsjtzyy: changed

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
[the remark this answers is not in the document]
jsjtzyy: Let me think about this. I'll probably change it to `StateTransitionException`.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
jsjtzyy: I tried to separate IllegalStateException from Exception for this case but then I found `catch (Exception e)` capture any exception from `getMissingStoreKeys`, `processReplicaMetadataResponse` and `isSyncUpComplete` and set error code to a general `Unknown_Error`. I guess previous author of this code combined all exception cases into single error code and drove us dig into the root cause. The cause is actually logged in this catch block:
```
logger.error(
                  "Remote node: " + remoteNode + " Thread name: " + threadName + " Remote replica: " + remoteReplicaInfo
                      .getReplicaId(), e);```
So, I temporarily leave it as it is, to refactor the exception handling may be worth a separate PR.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpService.java
jsjtzyy: You are right, we technically can combine them together. For now, I track them separately for debugging purpose. Since there is not too much overheard by doing so, let's temporarily keep them here and simplify them if needed in the future.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicaThread.java
lightningrob: I think it works if you catch StateTransitionException before Exception, but we can punt this to a future PR.
Even if you just catch Exception, you can have a utility that converts exception to error code.

PR #1337, opened 2019-12-13
Allow PutOperation to use dynamic success target

jsjtzyy:
In the intermediate state of "move replica", when decommission of old replicas is initiated(but hasn't transited to INACTIVE yet), the PUT requests should be rejected on old replicas. For frontends, they are seeing both old and new replicas(lets say 3 old and 3 new) and the success target should be 6 - 1 = 5. In the aforementioned scenario, PUT request failed on 3 old replicas and we are supposed to fail whole PUT operation because number of remaining requests is already less than success target. From another point of view, however, PUT request is highly likely to succeed on 3 new replicas and we actually could consider it success without generating "slip put" (which makes PUT latency worse)
In this PR, we propose a new success criterion:
             success count > Math.max(parallelism - disabled count - 1, 2)
The 2 here is original PUT success target from config, which is the lower bound of PUT operation target.

### DIFF
--- a/ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
@@ -1279,7 +1279,7 @@ class PutOperation {
         int correlationId = entry.getKey();
         ChunkPutRequestInfo info = entry.getValue();
         if (time.milliseconds() - info.startTimeMs > routerConfig.routerRequestTimeoutMs) {
-          onErrorResponse(info.replicaId);
+          onErrorResponse(info.replicaId, RouterErrorCode.OperationTimedOut);
           logger.debug("PutRequest with correlationId {} in flight has expired for replica {} ", correlationId,
               info.replicaId.getDataNodeId());
           // Do not notify this as a failure to the response handler, as this timeout could simply be due to
@@ -1383,18 +1383,21 @@ class PutOperation {
       routerMetrics.getDataNodeBasedMetrics(chunkPutRequestInfo.replicaId.getDataNodeId()).putRequestLatencyMs.update(
           requestLatencyMs);
       boolean isSuccessful;
+      RouterErrorCode routerErrorCode = null;
       if (responseInfo.getError() != null) {
         logger.debug("PutRequest with response correlationId {} timed out for replica {} ", correlationId,
             chunkPutRequestInfo.replicaId.getDataNodeId());
-        setChunkException(new RouterException("Operation timed out", RouterErrorCode.OperationTimedOut));
+        routerErrorCode = RouterErrorCode.OperationTimedOut;
+        setChunkException(new RouterException("Operation timed out", routerErrorCode));
         isSuccessful = false;
       } else {
         if (putResponse == null) {
           logger.debug(
               "PutRequest with response correlationId {} received an unexpected error on response deserialization from replica {} ",
               correlationId, chunkPutRequestInfo.replicaId.getDataNodeId());
-          setChunkException(new RouterException("Response deserialization received an unexpected error",
-              RouterErrorCode.UnexpectedInternalError));
+          routerErrorCode = RouterErrorCode.UnexpectedInternalError;
+          setChunkException(
+              new RouterException("Response deserialization received an unexpected error", routerErrorCode));
           isSuccessful = false;
         } else {
           if (putResponse.getCorrelationId() != correlationId) {
@@ -1405,8 +1408,8 @@ class PutOperation {
             routerMetrics.unknownReplicaResponseError.inc();
             logger.error("The correlation id in the PutResponse " + putResponse.getCorrelationId()
                 + " is not the same as the correlation id in the associated PutRequest: " + correlationId);
-            setChunkException(
-                new RouterException("Unexpected internal error", RouterErrorCode.UnexpectedInternalError));
+            routerErrorCode = RouterErrorCode.UnexpectedInternalError;
+            setChunkException(new RouterException("Unexpected internal error", routerErrorCode));
             isSuccessful = false;
             // we do not notify the ResponseHandler responsible for failure detection as this is an unexpected error.
           } else {
@@ -1421,6 +1424,8 @@ class PutOperation {
                   chunkPutRequestInfo.replicaId, putResponse.getError(), putResponse.getCorrelationId(), blobId);
               processServerError(putResponse.getError());
               isSuccessful = false;
+              routerErrorCode = putError == ServerErrorCode.Temporarily_Disabled ? RouterErrorCode.OperationDisabled
+                  : RouterErrorCode.AmbryUnavailable;
             }
           }
         }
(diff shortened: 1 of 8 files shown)


ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -238,9 +243,29 @@
         routerOperation, successTarget, parallelism, originatingDcNotFoundFailureThreshold);
   }
 
+  /**
+   * The dynamic success target is introduced mainly for following use case:
+   * In the intermediate state of "move replica", when decommission of old replicas is initiated(but hasn't transited to
```
justinlin-linkedin: Does this apply to the scenario when adding replicas to the partition? When it's bootstrapping, ambry server would return Temporary_Disable to put request as well.

jsjtzyy: Good question, we actually don't have to worry about new replicas that are bootstrapping because from frontends' point view, they are either in OFFLINE or BOOTSTRAP state which won't be picked for PUT operation. (For PUT, the eligible states are STANDBY and LEADER)
The problem addressed by this PR is because frontends are still seeing old view in which some replicas are still STANDBY/LEADER but their actual state is being decommissioned.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -257,17 +282,22 @@ public boolean isDone() {
...
-        originatingDcNotFoundCount++;
-      }
+    switch (trackedRequestFinalState) {
+      case SUCCESS:
+        succeededCount++;
+        break;
+      case REQUEST_DISABLED:
+        disabledCount++;
+        break;
+      default:
```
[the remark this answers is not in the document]
jsjtzyy: It actually occurs for DELETE and TTLUpdate as well but the original success target/ failure condition still holds for these two types. Only PUT operation should be adjusted accordingly. I will add some comments here.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1421,6 +1438,8 @@ void handleResponse(ResponseInfo responseInfo, PutResponse putResponse) {
                   chunkPutRequestInfo.replicaId, putResponse.getError(), putResponse.getCorrelationId(), blobId);
               processServerError(putResponse.getError());
```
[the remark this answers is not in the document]
jsjtzyy: yeah, we could but I feel we may not get any benefit by doing that. Distinguishing `OperationDisabled` from `AmbryUnavailable` is to help operation tracker to make decisions only. For external components (netty layer), OperationDisabled is also a kind of AmbryUnavailable. Hence, I probably would not touch processServerError to avoid any logic depending on AmrbyUnavailable router error code.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1433,17 +1452,19 @@ void handleResponse(ResponseInfo responseInfo, PutResponse putResponse) {
...
     }
 
     /**
      * Perform the necessary actions when a request to a replica fails.
      * @param replicaId the {@link ReplicaId} associated with the failed response.
+     * @param routerErrorCode the {@link RouterErrorCode} associated with failed response.
      */
-    private void onErrorResponse(ReplicaId replicaId) {
-      operationTracker.onResponse(replicaId, TrackedRequestFinalState.FAILURE);
+    private void onErrorResponse(ReplicaId replicaId, RouterErrorCode routerErrorCode) {
```
cgtz: If the new `RouterErrorCode`,  `OperationDisabled` is only used to assign `TrackedRequestFinalState`, why not pass `TrackedRequestFinalState` into this method instead of adding a new router code?

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
```diff
@@ -159,6 +161,127 @@ public void localFailTest() {
...
+   * Test cases:
+   * Case 1: 1 LEADER and 2 STANDBY replicas (regular case)
+   * Case 2: 1 LEADER and 3 STANDBY replicas. One of them returns REQUEST_DISABLED error code. There are two combinations
+   *         for remaining 3 replicas:
+   *         (1) two success and one failure (whole operation should succeed)
+   *         (2) one success and two failure (whole operation should fail)
+   * Case 3: 1 LEADER and 5 STANDBY replicas. Several combinations to discuss:
+   *         (1) 3 REQUEST_DISABLED, 2 success and 1 failure (operation should succeed)
+   *         (2) 2 REQUEST_DISABLED, 2 failure (operation should fail no matter what results are from remaining replicas)
+   *         (3) 1 REQUEST_DISABLED, 1 failure, 4 success (operation should succeed)
```
cgtz: should cases where we have 6 replicas but 0 replicas return REQUEST_DISABLED be tested in this test?

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -238,9 +243,29 @@
...
+   * able to see new replicas and subsequent READ/DELETE/TtlUpdate request should succeed on at least 2 aforementioned
+   * new replicas.
+   */
   @Override
   public boolean hasSucceeded() {
-    return succeededCount >= successTarget;
+    boolean hasSucceeded;
+    if (routerOperation == RouterOperation.PutOperation && routerConfig.routerPutUseDynamicSuccessTarget) {
+      hasSucceeded =
+          succeededCount >= Math.max(totalReplicaCount - disabledCount - 1, routerConfig.routerPutSuccessTarget);
```
cgtz: i think we could use `successTarget` in these places

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
jsjtzyy: added

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
jsjtzyy: You're right. Will make the change.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
jsjtzyy: Unfortunately, we can't. Here is a counter example:  
 We have 6 replicas(3 old, 3 new) and they are all eligible from frontend's view.  `successTarget` is now 5 (= 6 - 1). Meanwhile, decommission is initiated on 3 old replicas and PutOperation gets 3 success (from new replicas) and 3 disabled result. We should consider this a success. However, if we use `succeededCount >= Math.max(totalReplicaCount - disabledCount - 1, successTarget)`, then `succeededCount` is less than `successTarget` (3 < 5) in this case, which make the operation fail.

PR #1339, opened 2019-12-16
Publish test jars to maven repo

cgtz:
Ambry library users may want to depend on test utilities. This PR
removes the unused legacy maven publications and includes the test jar
in the javaLibrary publication if shipkit (which instantiates the
javaLibrary publication) is enabled.

Tested with `./gradlew publishToMavenLocal`:
```
~/.m2/repository/com/github/ambry/ambry-router/0.3.72
❯ ls
ambry-router-0.3.72-javadoc.jar  ambry-router-0.3.72-test.jar     ambry-router-0.3.72.pom
ambry-router-0.3.72-sources.jar  ambry-router-0.3.72.jar
```


lightningrob:
The change looks fine, but what is the use case for Ambry users to use our test code?

cgtz:
@lightningrob in our linkedin codebase, we have some test cases that depend on useful test utilities that are defined in open source ambry. This will facilitate the switch to depending on the open source artifacts in AmbryLI.

PR #1341, opened 2019-12-17
Fix potential empty routing table snapshot during HelixClusterManager initialization

jsjtzyy:
After routing table provider is created, initial instance config change event is added into the queue. However, when the event will be dequeued is not able to be determined (dequeue is called by background thread). There could be an edge case where even after clustermap is initialized, the initial event is still in the queue which causes routing table snapshot to be empty. This PR explicitly
checks the snapshot and waits until it gets populated. Otherwise, there may be no available replicas for router to send requests.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
@@ -154,8 +154,7 @@ public class HelixClusterManager implements ClusterMap {
               logger.info("Established connection to Helix manager at {}", zkConnectStr);
             }
 
-            CountDownLatch routingTableInitLatch = new CountDownLatch(1);
-            ClusterChangeHandler clusterChangeHandler = new ClusterChangeHandler(dcName, routingTableInitLatch);
+            ClusterChangeHandler clusterChangeHandler = new ClusterChangeHandler(dcName);
             // Create RoutingTableProvider of each DC to keep track of partition(replicas) state. Here, we use current
             // state based RoutingTableProvider to remove dependency on Helix's pipeline and reduce notification latency.
             logger.info("Creating routing table provider associated with Helix manager at {}", zkConnectStr);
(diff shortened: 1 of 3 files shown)


jsjtzyy:
Also, this PR change the test from latch-based to periodic polling approach.

jsjtzyy:
@cgtz gentle reminder to review

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -546,13 +560,17 @@ long getErrorCount() {
     private final AtomicBoolean instanceConfigInitialized = new AtomicBoolean(false);
     private final AtomicBoolean liveStateInitialized = new AtomicBoolean(false);
     private final AtomicBoolean idealStateInitialized = new AtomicBoolean(false);
+    private final AtomicBoolean routingTableInitialized = new AtomicBoolean(false);
```
cgtz: Don't need a second variable for this, you can just check that `routingTableInitLatch.getCount() == 0`. This should have roughly equivalent performance since it is just checking the value of a `volatile int`

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -153,6 +153,9 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
               manager.connect();
               logger.info("Established connection to Helix manager at {}", zkConnectStr);
             }
+
+            CountDownLatch routingTableInitLatch = new CountDownLatch(1);
```
cgtz: I would recommend instantiating the `CountDownLatch` within `ClusterChangeHandler` since instantiating it outside leaks some internal details, like the count you are waiting for. Maybe add a method like `CountDownLatch::waitForStartup(int minutesToWait)`

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
jsjtzyy: fair point, will remove this to make it simple.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
jsjtzyy: I guess you are referring to `ClusterChangeHandler.waitForStartup(int waitTimeout)`? (Correct me if I am wrong)

PR #1345, opened 2019-12-23
Implement replica Standby-To-Inactive transition on server side

jsjtzyy:
Standby-To-Inactive transition is triggered when replica is being
decommissioned. To avoid race condition due to inconsisten view and to
ensure other peer replicas have caught up with last PUT in local store,
we take following actions in order during the transition:
1. server sets local replica to INACTIVE state
2. disable compaction on local store
3. reject new PUTs but accept DELETE/TTLUpdates on this replica
4. check remote replica lag to ensure they catch up with last PUT
5. complete the transition

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/StateTransitionException.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/StateTransitionException.java
@@ -13,6 +13,9 @@
  */
 package com.github.ambry.clustermap;
 
+/**
+ * An extension of {@link RuntimeException} used to record exceptions occurred during state transition.
+ */
 public class StateTransitionException extends RuntimeException {
   private static final long serialVersionUID = 1L;
   private final TransitionErrorCode error;
@@ -26,6 +29,11 @@ public class StateTransitionException extends RuntimeException {
     return error;
   }
 
+  /**
+   * All types of error code that associate with {@link StateTransitionException}. The error code is currently used by
+   * tests to determine location of exception. In production environment, if transition exception occurs, the message
+   * together with error code should be recorded in Helix log which helps us investigate failure cause.
+   */
   public enum TransitionErrorCode {
     /**
      * If replica is not present in Helix and not found on current node.
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/StateTransitionException.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/StateTransitionException.java
@@ -42,6 +42,10 @@ public class StateTransitionException extends RuntimeException {
     /**
      * If bootstrap process fails at some point for specific replica.
      */
-    BootstrapFailure
+    BootstrapFailure,
+    /**
+     * If failure occurs during Standby-To-Inactive transition.
+     */
+    DeactivationFailure
   }
 }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
@@ -62,16 +62,16 @@ public class AmbryPartitionStateModel extends StateModel {
         message.getResourceName());
     if (clusterMapConfig.clustermapEnableStateModelListener) {
       partitionStateChangeListener.onPartitionBecomeStandbyFromBootstrap(partitionName);
-    }
-    try {
-      replicaSyncUpManager.waitBootstrapCompleted(partitionName);
-    } catch (InterruptedException e) {
-      logger.error("Bootstrap was interrupted on partition {}", partitionName);
-      throw new StateTransitionException("Bootstrap failed or was interrupted",
-          StateTransitionException.TransitionErrorCode.BootstrapFailure);
-    } catch (StateTransitionException e) {
-      logger.error("Bootstrap didn't complete.", e);
-      throw e;
+      try {
+        replicaSyncUpManager.waitBootstrapCompleted(partitionName);
+      } catch (InterruptedException e) {
+        logger.error("Bootstrap was interrupted on partition {}", partitionName);
+        throw new StateTransitionException("Bootstrap failed or was interrupted",
+            StateTransitionException.TransitionErrorCode.BootstrapFailure);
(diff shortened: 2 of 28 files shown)


ambry-api/src/main/java/com.github.ambry/clustermap/StateTransitionException.java
```diff
@@ -42,6 +42,10 @@ public TransitionErrorCode getErrorCode() {
     /**
      * If bootstrap process fails at some point for specific replica.
      */
-    BootstrapFailure
+    BootstrapFailure,
```
lightningrob: Could you add javadoc for the enum class that explains how TransitionErrorCode is used by Helix or elsewhere?  I only see them used by the tests.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -336,4 +348,32 @@ public void onPartitionBecomeStandbyFromLeader(String partitionName) {
...
+    if (replicationManagerListener != null) {
+      replicationManagerListener.onPartitionBecomeInactiveFromStandby(partitionName);
+      // after deactivation is initiated in ReplicationManager, transition is blocked here and wait until enough peer
+      // replicas have caught up with last PUT in local store.
+      try {
+        replicaSyncUpManager.waitDeactivationCompleted(partitionName);
+      } catch (InterruptedException e) {
+        logger.error("Deactivation was interrupted on partition {}", partitionName);
+        throw new StateTransitionException("Deactivation failed or was interrupted",
+            StateTransitionException.TransitionErrorCode.DeactivationFailure);
```
lightningrob: Minor: import TransitionErrorCode statically.

ambry-clustermap/src/test/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManagerTest.java
```diff
@@ -63,18 +63,23 @@ public AmbryReplicaSyncUpManagerTest() throws IOException {
     replicas.removeAll(localDcPeers);
     localDcPeerReplicas = new ArrayList<>(localDcPeers);
     remoteDcPeerReplicas = new ArrayList<>(replicas);
+    List<com.github.ambry.utils.TestUtils.ZkInfo> zkInfoList = new ArrayList<>();
```
lightningrob: Please import static the ZkInfo class.

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
```diff
@@ -428,5 +429,35 @@ public void onPartitionBecomeLeaderFromStandby(String partitionName) {
...
+        Store localStore = getStore(replica.getPartitionId());
+        if (localStore != null) {
+          // 1. set state to INACTIVE
+          localStore.setCurrentState(ReplicaState.INACTIVE);
+          logger.info("Store {} is set to INACTIVE", partitionName);
+          // 2. disable compaction on this store
+          if (!controlCompactionForBlobStore(replica.getPartitionId(), false)) {
+            logger.error("Failed to disable compaction on store {}", partitionName);
+            throw new StateTransitionException("Couldn't disable compaction on replica " + replica.getReplicaPath(),
+                StateTransitionException.TransitionErrorCode.ReplicaOperationFailure);
```
lightningrob: Minor: consider renaming ReplicaOperationFailure or using one of the other error codes.

ambry-replication/src/test/java/com.github.ambry.replication/MockReplicationManager.java
```diff
@@ -48,6 +49,8 @@
   // Variables for controlling getRemoteReplicaLagFromLocalInBytes()
   // the key is partitionId:hostname:replicaPath
   public Map<String, Long> lagOverrides = null;
+  CountDownLatch listenerExcuctionLatch = null;
```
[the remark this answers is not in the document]
jsjtzyy: good catch, will fix

ambry-api/src/main/java/com.github.ambry/clustermap/StateTransitionException.java
jsjtzyy: Sure, for now it is used by tests to determine which step has failed. In the future, it will probably be used by Helix State Model when investigating specific replica that is in `ERROR` state.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
jsjtzyy: good point, will do

ambry-store/src/main/java/com.github.ambry.store/StorageManager.java
jsjtzyy: Right, I will probably change it to `ReplicaNotFound` because it is the only reason disabling compaction may fail.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartitionStateModel.java
```diff
@@ -91,8 +78,12 @@ public void onBecomeStandbyFromLeader(Message message, NotificationContext conte
 
   @Transition(to = "INACTIVE", from = "STANDBY")
   public void onBecomeInactiveFromStandby(Message message, NotificationContext context) {
+    String partitionName = message.getPartitionName();
```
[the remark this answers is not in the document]
jsjtzyy: Good catch! The logger.info should be: (getResourceName)
```logger.info("Partition {} in resource {} is becoming INACTIVE from STANDBY", partitionName, message.getResourceName());```
I will fix this. Thanks

ambry-clustermap/src/test/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManagerTest.java
lightningrob: Minor: since you are importing TestUtils.*, the code can just say ZkInfo not TestUtils.ZkInfo.

PR #1346, opened 2019-12-26
Implement replica Inactive-To-Offline transition on server side

jsjtzyy:
The major change in this PR is to implement Inactive-To-Offline logic
during the transition. Specfically, there are 3 steps to go:
1. set local store state to OFFLINE and initiate disconnection
2. wait until enough peer replicas have fully caught up with local
store (include DELETE/TTLUpdate records)
3. storage manager stop the store
4. (TODO) update instanceConfig in Helix (will be in a separate PR)
In addition, this PR also allows store to create a decommission file to
indicate if local store is in decommission process. This is to
distinguish such store from regular stores and it also helps failure
recovery.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpManager.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpManager.java
@@ -41,7 +41,7 @@ public interface ReplicaSyncUpManager {
    * @param localReplica the replica that resides on current node
    * @param peerReplica the peer replica of local one.
    * @param lagInBytes replica lag bytes
-   * @return whether the lag is updated or not. If {@code false}, it means the source replica is not tracked in this service.
+   * @return whether the lag is updated or not. If {@code false}, it means the local replica is not tracked in this service.
    *         Either the replica has caught up and removed from service or it is an existing replica that doesn't need catchup.
    */
   boolean updateLagBetweenReplicas(ReplicaId localReplica, ReplicaId peerReplica, long lagInBytes);
@@ -90,4 +90,29 @@ public interface ReplicaSyncUpManager {
    * @throws InterruptedException
    */
   void waitDeactivationCompleted(String partitionName) throws InterruptedException;
+
+  /**
+   * Initiate disconnection process to stop replica and make it offline. This happens when replica is going to be
+   * removed from current node. For Ambry, it disconnection occurs in Inactive-To-Offline transition.
+   * @param replicaId the {@link ReplicaId} to be disconnected.
+   */
+  void initiateDisconnection(ReplicaId replicaId);
+
+  /**
+   * Wait until disconnection on given partition is complete.
+   * @param partitionName partition name of replica that in disconnection process
+   */
+  void waitDisconnectionCompleted(String partitionName) throws InterruptedException;
+
+  /**
+   * When disconnection completes on given replica and then it becomes offline.
+   * @param replicaId the {@link ReplicaId} on which disconnection completes
+   */
+  void onDisconnectionComplete(ReplicaId replicaId);
+
+  /**
+   * When exception/error occurs during disconnection.
+   * @param replicaId the {@link ReplicaId} which encounters error.
+   */
+  void onDisconnectionError(ReplicaId replicaId);
 }
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManager.java
@@ -40,9 +40,9 @@ import static com.github.ambry.clustermap.StateTransitionException.TransitionErr
 public class AmbryReplicaSyncUpManager implements ReplicaSyncUpManager {
   private final ConcurrentHashMap<String, CountDownLatch> partitionToBootstrapLatch = new ConcurrentHashMap<>();
   private final ConcurrentHashMap<String, CountDownLatch> partitionToDeactivationLatch = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<String, CountDownLatch> partitionToDisconnectionLatch = new ConcurrentHashMap<>();
   private final ConcurrentHashMap<String, Boolean> partitionToBootstrapSuccess = new ConcurrentHashMap<>();
   private final ConcurrentHashMap<String, Boolean> partitionToDeactivationSuccess = new ConcurrentHashMap<>();
-  private final ConcurrentHashMap<String, CountDownLatch> partitionToDisconnectionLatch = new ConcurrentHashMap<>();
   private final ConcurrentHashMap<String, Boolean> partitionToDisconnectionSuccess = new ConcurrentHashMap<>();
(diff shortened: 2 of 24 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -105,18 +108,62 @@ public boolean isSyncUpComplete(ReplicaId replicaId) {
...
+    replicaToLagInfos.put(replicaId, new LocalReplicaLagInfos(replicaId, 0, ReplicaState.OFFLINE));
+  }
+
+  @Override
+  public void waitDisconnectionCompleted(String partitionName) throws InterruptedException {
+    CountDownLatch latch = partitionToDisconnectionLatch.get(partitionName);
+    if (latch == null) {
+      logger.error("Partition {} is not found for disconnection", partitionName);
+      throw new StateTransitionException("No disconnection latch is found for partition " + partitionName,
+          StateTransitionException.TransitionErrorCode.ReplicaNotFound);
```
justinlin-linkedin: this makes waitDisconnectionCompleted method non-reentrant? Is this intentional?

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationEngine.java
```diff
@@ -141,7 +147,25 @@ public void updateTotalBytesReadByRemoteReplica(PartitionId partitionId, String
...
+        Store localStore = storeManager.getStore(partitionId);
+        if (localStore.getCurrentState() == ReplicaState.INACTIVE) {
+          // TODO incorporate changes for Standby-To-Inactive transition.
+        } else if (localStore.isDecommissionInProgress() && localStore.getCurrentState() == ReplicaState.OFFLINE) {
+          // if local store is in OFFLINE state, we need more info to determine if replica is really in Inactive-To-Offline
+          // transition. So we check if decommission file is present. If present, we update SyncUpManager by peer's lag
+          // from end offset in local store.
+          boolean updated =
+              replicaSyncUpManager.updateLagBetweenReplicas(localReplica, remoteReplicaInfo.getReplicaId(),
+                  localStore.getSizeInBytes() - totalBytesRead);
```
justinlin-linkedin: I am confused by the calculation here. localStore.getSizeInBytes() might actually have more bytes than the totalBytesRead since we ignore the delete blob in replication. Is this an accurate difference or just a estimate?

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -683,6 +684,13 @@ public boolean isBootstrapInProgress() {
     return (new File(dataDir, BOOTSTRAP_FILE_NAME)).exists();
   }
 
+  @Override
+  public boolean isDecommissionInProgress() {
+    // note that, the decommission file will be removed by calling deleteStoreFiles() when replica is being dropped. We
+    // don't need to explicitly delete it. The file is also used for failure recovery to resume decommission process.
+    return (new File(dataDir, DECOMMISSION_FILE_NAME)).exists();
```
justinlin-linkedin: decommission is per partition, does this work with more than on partition? if not, shouldn't we pass the partition name as part of the filename?

ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpManager.java
```diff
@@ -56,15 +56,38 @@
...
-  void onBootstrapError(String partitionName);
+  void onBootstrapError(ReplicaId replicaId);
 
-  // TODO introduce decommission logic in sync-up service. For example, initiateDecommission(String partitionName)
+  /**
+   * Initiate disconnection process to stop replica and make it offline. This happens when replica is going to be
+   * removed from current node. For Ambry, it disconnection occurs in Inactive-To-Offline transition.
+   * @param replicaId the {@link ReplicaId} to be disconnected.
+   */
+  void initiateDisconnection(ReplicaId replicaId);
```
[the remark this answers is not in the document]
jsjtzyy: I am not good at naming :(  but I think `disconnection` contains the meaning: make sth offline. Offline doesn't sound like a process (more like a state). In another PR, I also use a similar word `deactivation` to denote process from Standby to Inactive. In my humble opinion, we are able to use these two words to distinguish STANDBY->INACTIVE from INACTIVE->OFFLINE transition.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplicaSyncUpManager.java
jsjtzyy: will discuss this offline (my first thought is this method is not supposed to be reentrant)

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
jsjtzyy: Each store represents one partition. We put file into each store's directory (`dataDir`). This method is within BlobStore, when it is called, it indicates if the partition associated with this store is in decommission process. (I need to apologize for confusing naming for state transition. `Decommission` has two phases: `deactivation` and `disconnection`.  Any partition that is being decommissioned first gets deactivated and then gets disconnected.)

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationEngine.java
jsjtzyy: It's actually accurate. `totalBytesRead` is calculated in metadata exchange phase and is computed based on remote replica token (offset in this token). The bytes read by remote replica contain all the records in local store (PUT/DELETE/TTL Update), although remote replica is able to drop the delete record if original put record is not found in its store.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -336,4 +348,34 @@ public void onPartitionBecomeStandbyFromLeader(String partitionName) {
...
+    try {
+      replicaSyncUpManager.waitDisconnectionCompleted(partitionName);
+    } catch (InterruptedException e) {
+      logger.error("Disconnection was interrupted on partition {}", partitionName);
+      throw new StateTransitionException("Disconnection failed or was interrupted",
+          StateTransitionException.TransitionErrorCode.DisconnectionFailure);
+    } catch (StateTransitionException e) {
+      logger.error("Disconnection didn't complete ", e);
+      throw e;
+    }
```
justinlin-linkedin: should move this block to the if statement above.

ambry-api/src/main/java/com.github.ambry/clustermap/ReplicaSyncUpManager.java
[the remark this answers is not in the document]
zzmao: When waitDisconnectionCompleted returns, is offline transition completed?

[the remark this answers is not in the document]
jsjtzyy: The transition is pretty much done (except for one more step in storage manager to shut down the replica, should be very quick).

jsjtzyy:
Fair point, will rename them in a separate PR.

PR #1347, opened 2019-12-28
Implement replica Offline-To-Dropped transition on server side

jsjtzyy:
This PR implements the last step to remove old replica. Specifically, it
removes old replica from in-mem data structures (in StatsManager,
ReplicationManager and StorageManager). Also, during the transition, all
store files would be deleted in the end and seal/stop list in Helix
would be updated as well.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
@@ -19,7 +19,6 @@ import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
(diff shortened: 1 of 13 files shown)


jsjtzyy:
@zzmao  this PR is ready for further review.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
```diff
@@ -166,7 +165,7 @@ static int getSchemaVersion(InstanceConfig instanceConfig) {
    */
   static List<String> getSealedReplicas(InstanceConfig instanceConfig) {
     List<String> sealedReplicas = instanceConfig.getRecord().getListField(ClusterMapUtils.SEALED_STR);
-    return sealedReplicas == null ? Collections.emptyList() : sealedReplicas;
+    return sealedReplicas == null ? new ArrayList<>() : sealedReplicas;
```
zzmao: Just curious why.

jsjtzyy: Collections.emptyList() is immutable, we cannot call list.remove() method.

PR #1348, opened 2019-12-30
Allow ambry server to update replica info in Helix InstanceConfig

jsjtzyy:
Once replica addition or removal is complete on particular datanode,
ambry server is supposed to update InstanceConfig in Helix to broadcast
the replica related changes. This PR introduced a new method that allows
participant to update its node info in cluster. It supports adding new
replica info or removing old replica info. For rollback purpose, we also
add a config to switch on/off update operation.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
+++ b/ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
@@ -75,6 +75,16 @@ public interface ClusterParticipant extends AutoCloseable {
   ReplicaSyncUpManager getReplicaSyncUpManager();
 
   /**
+   * Update disk/replica infos associated with current data node in cluster (this occurs when replica addition/removal
+   * on current node is complete and local changes will be broadcast to all listeners in this cluster)
+   * @param replicaId the {@link ReplicaId} whose info should be updated on current node
+   * @param shouldExist Whether the replica info should exist or not. When {@code true}, replica info will be added if
+   *                      it is missing in current node info. When {@code false}, replica info will be removed if present.
+   * @return if {@code true}, node info is successfully updated. {@code false} otherwise.
+   */
+  boolean updateDataNodeInfoInCluster(ReplicaId replicaId, boolean shouldExist);
+
+  /**
    * Terminate the participant.
    */
   @Override
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
@@ -18,6 +18,7 @@ import com.github.ambry.server.AmbryHealthReport;
 import com.github.ambry.utils.Utils;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -334,23 +335,19 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
     if (diskInfo != null) {
       String replicasStr = diskInfo.get(ClusterMapUtils.REPLICAS_STR);
       if (!replicasStr.isEmpty()) {
-        String[] replicaInfos = replicasStr.split(ClusterMapUtils.REPLICAS_DELIM_STR);
-        for (String replicaInfo : replicaInfos) {
-          String[] infos = replicaInfo.split(ClusterMapUtils.REPLICAS_STR_SEPARATOR);
-          if (infos[0].equals(partitionName)) {
-            replicaFound = true;
-            break;
-          }
-        }
+        List<String> replicaInfos =
+            new ArrayList<>(Arrays.asList(replicasStr.split(ClusterMapUtils.REPLICAS_DELIM_STR)));
+        // if any element is removed, that means old replica is found in replicasStr.
+        replicaFound = replicaInfos.removeIf(
+            info -> (info.split(ClusterMapUtils.REPLICAS_STR_SEPARATOR)[0]).equals(partitionName));
+
         // We update InstanceConfig only when replica is found in current instanceConfig. (This is to avoid unnecessary
         // notification traffic due to InstanceConfig change)
         if (replicaFound) {
           StringBuilder newReplicasStrBuilder = new StringBuilder();
+          // note that old replica info has been removed from "replicaInfos"
           for (String replicaInfo : replicaInfos) {
-            String[] infos = replicaInfo.split(ClusterMapUtils.REPLICAS_STR_SEPARATOR);
-            if (!infos[0].equals(partitionName)) {
-              newReplicasStrBuilder.append(replicaInfo).append(ClusterMapUtils.REPLICAS_DELIM_STR);
-            }
+            newReplicasStrBuilder.append(replicaInfo).append(ClusterMapUtils.REPLICAS_DELIM_STR);
           }
           // update diskInfo and MountPathToDisk map
           diskInfo.put(ClusterMapUtils.REPLICAS_STR, newReplicasStrBuilder.toString());
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
@@ -190,7 +190,6 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
   }
 
   /**
-   * Get the list of sealed replicas from the HelixAdmin.
    * @return list of sealed replicas from HelixAdmin.
    */
   @Override
@@ -204,7 +203,6 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
   }
 
   /**
-   * Get the list of stopped replicas from the HelixAdmin.
    * @return list of stopped replicas from HelixAdmin
    */
   @Override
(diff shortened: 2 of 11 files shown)


ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
```diff
@@ -74,6 +74,16 @@ void registerPartitionStateChangeListener(StateModelListenerType listenerType,
    */
   ReplicaSyncUpManager getReplicaSyncUpManager();
 
+  /**
+   * Update disk/replica infos associated with current data node in cluster (this occurs when replica addition/removal
+   * on current node is complete and local changes will be broadcast to all listeners in this cluster)
+   * @param replicaId the {@link ReplicaId} whose info should be updated on current node
+   * @param shouldPresent Whether the replica info should be present. When {@code true}, replica info will be added if
```
zzmao: Minor: Feel `addOrRemove` is better.

jsjtzyy: `addOrRemove` seems to require an enum. We cannot tell if it should be added or removed by looking at the boolean value (true/false).  Aside from `shouldPresent`, I also consider `shouldExist`.  I am still open to other proper names.

zzmao: `shouldExist` feels better.

jsjtzyy: ok, will make the change.

config/server.properties
```diff
@@ -16,3 +16,4 @@ host.name=localhost
 clustermap.cluster.name=Ambry_Dev
 clustermap.datacenter.name=Datacenter
 clustermap.host.name=localhost
+clustermap.port=6667
```
zzmao: Why is this newly added?

jsjtzyy: ah, I just added it here to fix issue #1357  (probably should be in a separate PR but let's get it fixed together with this PR merged)

jsjtzyy: Feel free to ignore this, not related to this PR.

zzmao: It's OK. I also need this change. :)

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -221,13 +221,154 @@ public ReplicaSyncUpManager getReplicaSyncUpManager() {
...
+    Map<String, Map<String, String>> mountPathToDiskInfos = instanceConfig.getRecord().getMapFields();
+    Map<String, String> diskInfo = mountPathToDiskInfos.get(replicaId.getMountPath());
+    if (diskInfo != null) {
+      String replicasStr = diskInfo.get(ClusterMapUtils.REPLICAS_STR);
+      if (!replicasStr.isEmpty()) {
+        String[] replicaInfos = replicasStr.split(ClusterMapUtils.REPLICAS_DELIM_STR);
+        for (String replicaInfo : replicaInfos) {
+          String[] infos = replicaInfo.split(ClusterMapUtils.REPLICAS_STR_SEPARATOR);
+          if (infos[0].equals(partitionName)) {
+            replicaFound = true;
```
zzmao: split string to arrayList and then `removeIf` probably needs less code.

jsjtzyy:
@ankagrawal  gentle reminder to review, thanks!

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -221,13 +222,150 @@ public ReplicaSyncUpManager getReplicaSyncUpManager() {
...
+      mountPathToDiskInfos.put(replicaId.getMountPath(), diskInfoToAdd);
+      newReplicaInfoAdded = true;
+    }
+    if (newReplicaInfoAdded) {
+      // we update InstanceConfig only when new replica info is added (skip updating if replica is already present)
+      instanceConfig.getRecord().setMapFields(mountPathToDiskInfos);
+      logger.info("Updating config: {} in Helix by adding partition {}", instanceConfig, partitionName);
+      additionResult = helixAdmin.setInstanceConfig(clusterName, instanceName, instanceConfig);
+    }
+    return additionResult;
```
ankagrawal: looks like additionResult is true even if partition is already present in an instance. Is that expected behavior? Should we return false in this case, as really we didnt add anything here.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -221,13 +222,150 @@ public ReplicaSyncUpManager getReplicaSyncUpManager() {
...
+        // this means new replica id is larger than all existing replicas' ids
+        replicasStrBuilder.append(newReplicaInfo);
+        newReplicaInfoAdded = true;
+      }
+      if (newReplicaInfoAdded) {
+        diskInfo.put(ClusterMapUtils.REPLICAS_STR, replicasStrBuilder.toString());
+        mountPathToDiskInfos.put(replicaId.getMountPath(), diskInfo);
+      }
+    } else {
+      // add replica onto a brand new disk
```
ankagrawal: minor:
Maybe add a log here saying that this is a brand new disk being added.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -221,13 +222,150 @@ public ReplicaSyncUpManager getReplicaSyncUpManager() {
...
+          mountPathToDiskInfos.put(replicaId.getMountPath(), diskInfo);
+          // update InstanceConfig
+          instanceConfig.getRecord().setMapFields(mountPathToDiskInfos);
+          logger.info("Updating config: {} in Helix by removing partition {}", instanceConfig, partitionName);
+          removalResult = helixAdmin.setInstanceConfig(clusterName, instanceName, instanceConfig);
+        }
+      }
+    }
+    if (!replicaFound) {
+      logger.warn("Partition {} is not found on instance {}, skipping removing it from InstanceConfig in Helix.",
```
ankagrawal: here too, should we return false, as we actually didnt find anything to remove?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -221,13 +222,150 @@ public ReplicaSyncUpManager getReplicaSyncUpManager() {
...
+        // We update InstanceConfig only when replica is found in current instanceConfig. (This is to avoid unnecessary
+        // notification traffic due to InstanceConfig change)
+        if (replicaFound) {
+          StringBuilder newReplicasStrBuilder = new StringBuilder();
+          // note that old replica info has been removed from "replicaInfos"
+          for (String replicaInfo : replicaInfos) {
+            newReplicasStrBuilder.append(replicaInfo).append(ClusterMapUtils.REPLICAS_DELIM_STR);
+          }
+          // update diskInfo and MountPathToDisk map
+          diskInfo.put(ClusterMapUtils.REPLICAS_STR, newReplicasStrBuilder.toString());
```
ankagrawal: If there are no more replicas on this disk (mount path) after removal, should we remove the disk from mountPathToDiskInfos?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -221,13 +222,150 @@ public ReplicaSyncUpManager getReplicaSyncUpManager() {
...
+        replicaFound = replicaInfos.removeIf(
+            info -> (info.split(ClusterMapUtils.REPLICAS_STR_SEPARATOR)[0]).equals(partitionName));
+
+        // We update InstanceConfig only when replica is found in current instanceConfig. (This is to avoid unnecessary
+        // notification traffic due to InstanceConfig change)
+        if (replicaFound) {
+          StringBuilder newReplicasStrBuilder = new StringBuilder();
+          // note that old replica info has been removed from "replicaInfos"
+          for (String replicaInfo : replicaInfos) {
+            newReplicasStrBuilder.append(replicaInfo).append(ClusterMapUtils.REPLICAS_DELIM_STR);
```
ankagrawal: We have already removed the replica from repliaInfos above. Not sure why we are doing this loop again?

config/server.properties
ankagrawal: Thanks for fixing this.

ambry-api/src/main/java/com.github.ambry/clustermap/ClusterParticipant.java
jsjtzyy: @ankagrawal  I guess you are suggesting `addReplicaToDataNodeInfo` and `removeReplicaFromDataNodeInfo`.  Yes, we could but a single method with boolean variable would suffice and seems to be more concise. In addition, updating data node info only applies to `HelixParticipant`. That is, we don't need to introduce one more unnecessary method to other implementation of `ClusterParticipant` (i.e. StaticClusterParticipant).

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
```diff
@@ -221,13 +222,150 @@ public ReplicaSyncUpManager getReplicaSyncUpManager() {
...
+  /**
+   * Add new replica info into {@link InstanceConfig} of current data node.
+   * @param replicaId new replica whose info should be added into {@link InstanceConfig}.
+   * @param instanceConfig the {@link InstanceConfig} to update.
+   * @return {@code true} replica info is successfully added. {@code false} otherwise.
+   */
+  private boolean addNewReplicaInfo(ReplicaId replicaId, InstanceConfig instanceConfig) {
+    boolean additionResult = true;
+    String partitionName = replicaId.getPartitionId().toPathString();
+    String newReplicaInfo = partitionName + ClusterMapUtils.REPLICAS_STR_SEPARATOR + replicaId.getCapacityInBytes()
```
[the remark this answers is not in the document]
jsjtzyy: sure, let me make the change.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
jsjtzyy: refer to previous explanation. Thanks for pointing it out. It's actually intended behavior in that case.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
jsjtzyy: A fair point. We temporarily don't remove the empty disk because it might be reused for holding other new replicas. If the node will be decommissioned, we directly remove InstanceConfig associated with this node in Helix.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixParticipant.java
jsjtzyy: We did remove the replica from `replicaInfos` and the loop is to generate a new string without removed replica. We then replace previous replica string with the new string.

PR #1349, opened 2020-01-03
Make writable partition selection account for partitions with more than 3 replicas

jsjtzyy:
Currently, selecting writable partition for PUT operation only picks partitions with highest local replica count. This will become a problem when "move replica" is rolled out. In the intermediate state of "move replica", particular partition may have 6 replicas and if we stick with current logic. All the write traffic will be routed to this particular partition. Hence, this PR we make changes to allow partition selection to pick partitions with replicas count >= min replica count(specified in
ClusterMapConfig). In most cases, min replica count = 3.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
+++ b/ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
@@ -216,6 +216,14 @@ public class ClusterMapConfig {
   @Config("clustermap.replica.catchup.target")
   public final int clustermapReplicaCatchupTarget;
 
+  /**
+   * The minimum number of replicas in local datacenter required for a partition to serve PUT request. This is used to
+   * get writable partitions for PUT operation. Any partition with replica count larger than or equal to this number is
+   * acceptable to be considered as a candidate.
+   */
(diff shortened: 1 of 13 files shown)


ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
```diff
@@ -261,5 +269,6 @@ public ClusterMapConfig(VerifiableProperties verifiableProperties) {
         verifiableProperties.getLongInRange("clustermap.replica.catchup.acceptable.lag.bytes", 0L, 0L, Long.MAX_VALUE);
     clustermapReplicaCatchupTarget =
         verifiableProperties.getIntInRange("clustermap.replica.catchup.target", 0, 0, Integer.MAX_VALUE);
+    clustermapWritablePartitionMinReplicaCount = verifiableProperties.getIntInRange("clustermap.writable.partition.min.replica.count", 3, 0, Integer.MAX_VALUE);
```
lightningrob: Minor: might need formatting.

jsjtzyy: sure, I am on it

PR #1354, opened 2020-01-09
Dont return bytes on simple blob when getChunkIdsOnly is true

justinlin-linkedin:
This would fixes a bytebuf leaking issue. When getChunkIdsOnly is true, if the blob is a simple blob, we should just return the id, not the data.

### DIFF
--- a/ambry-router/src/test/java/com.github.ambry.router/GetBlobOperationTest.java
+++ b/ambry-router/src/test/java/com.github.ambry.router/GetBlobOperationTest.java
@@ -408,6 +408,22 @@ public class GetBlobOperationTest {
   }
 
   /**
+   * Test gets of composite blob with getChunkIdsOnly being true.
+   */
+  @Test
+  public void testCompositeBlobGetChunkIdsOnly() throws Exception {
+    options =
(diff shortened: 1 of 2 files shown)


ambry-router/src/test/java/com.github.ambry.router/GetBlobOperationTest.java
```diff
@@ -363,6 +363,19 @@ public void testSimpleBlobRawMode() throws Exception {
...
+   */
+  @Test
+  public void testSimpleBlobGetChunkIdsOnly() throws Exception {
+    blobSize = maxChunkSize;
+    doPut();
+    options =
+        new GetBlobOptionsInternal(new GetBlobOptionsBuilder().operationType(GetBlobOptions.OperationType.All).build(),
+            true, routerMetrics.ageAtGet);
+    getAndAssertSuccess();
+  }
```
cgtz: Thanks for adding this test. Is there a test case getChunkIdsOnly with composite blobs anywhere? If not, it may be worth adding

justinlin-linkedin: added.

PR #1355, opened 2020-01-09
Introduce dynamic cluster change handler in HelixClusterManager

jsjtzyy:
Dynamic cluster change handler is another implementation of cluster change handler. Unlike simple cluster change handler which only create instance/partition/replica during initialization, dynamic cluster change handler supports adding new partition/node/replica at runtime. Also it supports removing replica from existing node. With this change, spectators in cluster are able to update membership list of instances/partitions and location of replicas without reboot. Moreover, the cluster metrics are also updated at runtime (i.e. cluster wide capacity, number of partitions etc)

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
@@ -16,16 +16,17 @@ package com.github.ambry.clustermap;
 import com.github.ambry.config.ClusterMapConfig;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
 import org.apache.helix.NotificationContext;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
@@ -298,12 +299,14 @@ public class DynamicClusterChangeHandler implements ClusterChangeHandler {
     ConcurrentHashMap<String, AmbryReplica> currentReplicasOnNode = ambryDataNodeToAmbryReplicas.get(dataNode);
     ConcurrentHashMap<String, AmbryReplica> replicasFromConfig = new ConcurrentHashMap<>();
     Map<String, Map<String, String>> diskInfos = instanceConfig.getRecord().getMapFields();
+    Map<String, AmbryDisk> mountPathToDisk = ambryDataNodeToAmbryDisks.get(dataNode)
+        .stream()
+        .collect(Collectors.toMap(AmbryDisk::getMountPath, disk -> disk));
+    Map<AmbryPartition, List<AmbryReplica>> replicaToAddByPartition = new HashMap<>();
     for (Map.Entry<String, Map<String, String>> entry : diskInfos.entrySet()) {
       String mountPath = entry.getKey();
       Map<String, String> diskInfo = entry.getValue();
-      Optional<AmbryDisk> potentialDisk =
-          ambryDataNodeToAmbryDisks.get(dataNode).stream().filter(d -> d.getMountPath().equals(mountPath)).findFirst();
-      AmbryDisk disk = potentialDisk.orElse(null);
+      AmbryDisk disk = mountPathToDisk.getOrDefault(mountPath, null);
       if (disk == null) {
         logger.info("Temporarily don't support adding new disk to existing node.");
         // TODO support dynamically adding disk in the future
@@ -340,23 +343,37 @@ public class DynamicClusterChangeHandler implements ClusterChangeHandler {
             updateReplicaStateAndOverrideIfNeeded(replica, sealedReplicas, stoppedReplicas);
             // add new created replica to "replicasFromConfig" map
             replicasFromConfig.put(partitionName, replica);
-            // add new replica to specific partition
-            clusterChangeHandlerCallback.addReplicasToPartition(mappedPartition, Collections.singletonList(replica));
+            // Put new replica into partition-to-replica map temporarily (this is to avoid any exception thrown within the
+            // loop before updating "ambryDataNodeToAmbryReplicas" map. If we update call addReplicasToPartition here
+            // immediately, the exception may cause inconsistency between "ambryPartitionToAmbryReplicas" and
+            // "ambryDataNodeToAmbryReplicas")
+            replicaToAddByPartition.put(mappedPartition, Collections.singletonList(replica));
           }
         }
       }
     }
+    // update ambryDataNodeToAmbryReplicas map
+    replicaToAddByPartition.forEach(clusterChangeHandlerCallback::addReplicasToPartition);
     // update ambryDataNodeToAmbryReplicas map by adding "replicasFromConfig"
     ambryDataNodeToAmbryReplicas.put(instanceNameToAmbryDataNode.get(instanceName), replicasFromConfig);
-    // compare replicasFromConfig with current replica set to derive old replicas that are removed
-    Set<AmbryReplica> previousReplicas = new HashSet<>(currentReplicasOnNode.values());
+    // Derive old replicas that are removed and delete them from partition
+/*    Set<AmbryReplica> previousReplicas = new HashSet<>(currentReplicasOnNode.values());
     previousReplicas.removeAll(replicasFromConfig.values());
     for (AmbryReplica ambryReplica : previousReplicas) {
       logger.info("Removing replica {} from existing node {}", ambryReplica.getPartitionId().toPathString(),
           instanceName);
       clusterChangeHandlerCallback.removeReplicasFromPartition(ambryReplica.getPartitionId(),
           Collections.singletonList(ambryReplica));
-    }
+    }*/
+    currentReplicasOnNode.keySet()
+        .stream()
+        .filter(partitionName -> !replicasFromConfig.containsKey(partitionName))
+        .forEach(pName -> {
+          logger.info("Removing replica {} from existing node {}", pName, instanceName);
+          AmbryReplica ambryReplica = currentReplicasOnNode.get(pName);
+          clusterChangeHandlerCallback.removeReplicasFromPartition(ambryReplica.getPartitionId(),
+              Collections.singletonList(ambryReplica));
+        });
   }
 
   /**
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
@@ -357,14 +357,6 @@ public class DynamicClusterChangeHandler implements ClusterChangeHandler {
     // update ambryDataNodeToAmbryReplicas map by adding "replicasFromConfig"
     ambryDataNodeToAmbryReplicas.put(instanceNameToAmbryDataNode.get(instanceName), replicasFromConfig);
     // Derive old replicas that are removed and delete them from partition
-/*    Set<AmbryReplica> previousReplicas = new HashSet<>(currentReplicasOnNode.values());
-    previousReplicas.removeAll(replicasFromConfig.values());
-    for (AmbryReplica ambryReplica : previousReplicas) {
-      logger.info("Removing replica {} from existing node {}", ambryReplica.getPartitionId().toPathString(),
-          instanceName);
-      clusterChangeHandlerCallback.removeReplicasFromPartition(ambryReplica.getPartitionId(),
-          Collections.singletonList(ambryReplica));
-    }*/
     currentReplicasOnNode.keySet()
         .stream()
         .filter(partitionName -> !replicasFromConfig.containsKey(partitionName))
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
@@ -37,6 +37,10 @@ import org.slf4j.LoggerFactory;
 import static com.github.ambry.clustermap.ClusterMapUtils.*;
 
 
+/**
+ * A more dynamic implementation of {@link ClusterChangeHandler} which supports adding new nodes/partitions at runtime.
+ * It is also able to absorb replica location changes in cluster.
+ */
 public class DynamicClusterChangeHandler implements ClusterChangeHandler {
   private final String dcName;
   private final Object notificationLock = new Object();
@@ -250,8 +254,8 @@ public class DynamicClusterChangeHandler implements ClusterChangeHandler {
   }
 
   /**
-   *
-   * @param instanceConfigs
+   * Add new instance or update existing instance based on {@link InstanceConfig}(s).
+   * @param instanceConfigs the {@link InstanceConfig}(s) used to update in-mem cluster map.
    * @throws Exception
    */
   private void addOrUpdateInstanceInfos(List<InstanceConfig> instanceConfigs) throws Exception {
@@ -343,10 +347,11 @@ public class DynamicClusterChangeHandler implements ClusterChangeHandler {
   }
 
   /**
-   *
-   * @param replica
-   * @param sealedReplicas
-   * @param stoppedReplicas
+   * If partition override is enabled, we override replica SEAL/UNSEAL state based on partitionOverrideMap. If disabled,
+   * update replica state according to the info from {@link InstanceConfig}.
+   * @param replica the {@link ReplicaId} whose states (seal,stop) should be updated.
+   * @param sealedReplicas a list of {@link ReplicaId}(s) that are in SEALED state.
+   * @param stoppedReplicas a list of {@link ReplicaId}(s) that are in STOPPED state.
    */
   private void updateReplicaStateAndOverrideIfNeeded(AmbryReplica replica, List<String> sealedReplicas,
       List<String> stoppedReplicas) {
@@ -364,8 +369,8 @@ public class DynamicClusterChangeHandler implements ClusterChangeHandler {
   }
 
   /**
-   *
-   * @param instanceConfig
+   * Create a new instance(node) and initialize disks/replicas on it.
+   * @param instanceConfig the {@link InstanceConfig} to create new instance
    * @throws Exception
    */
   private void createNewInstance(InstanceConfig instanceConfig) throws Exception {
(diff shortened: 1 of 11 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+      new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<AmbryDataNode, Set<AmbryDisk>> ambryDataNodeToAmbryDisks = new ConcurrentHashMap<>();
+  private final AtomicLong errorCount = new AtomicLong(0);
+
+  private volatile ConcurrentHashMap<String, String> partitionNameToResource = new ConcurrentHashMap<>();
+  private AtomicReference<RoutingTableSnapshot> routingTableSnapshotRef = new AtomicReference<>();
+  private ConcurrentHashMap<String, AmbryDataNode> instanceNameToAmbryDataNode = new ConcurrentHashMap<>();
+  private static final Logger logger = LoggerFactory.getLogger(DynamicClusterChangeHandler.class);
+
+  DynamicClusterChangeHandler(ClusterMapConfig clusterMapConfig, String dcName, String selfInstanceName,
```
justinlin-linkedin: javadoc please.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+
+  /**
+   * Handle any {@link InstanceConfig} related change in current datacenter. Several events will trigger instance config
+   * change: (1) replica's seal or stop state has changed; (2) new node or new partition is added; (3) new replica is
+   * added to existing node; (4) old replica is removed from existing node; (5) data node is deleted from cluster.
+   * For now, {@link DynamicClusterChangeHandler} supports (1)~(4). We may consider supporting (5) in the future.
+   * @param configs all the {@link InstanceConfig}(s) in current data center. (Note that PreFetch is enabled by default
+   *                in Helix, which means all instance configs under "participants" ZNode will be sent to this method)
+   * @param changeContext the {@link NotificationContext} associated.
+   */
```
justinlin-linkedin: can we add some comments with regards to the znode path in the zookeeper? I think other engineers would benefit a lot from this information.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+      synchronized (notificationLock) {
+        if (!instanceConfigInitialized.get()) {
+          logger.info("Received initial notification for instance config change from {}", dcName);
+        } else {
+          logger.info("Instance config change triggered from {}", dcName);
+        }
+        logger.debug("Detailed instance configs in {} are: {}", dcName, configs);
+        try {
+          addOrUpdateInstanceInfos(configs);
+          instanceConfigInitialized.set(true);
```
justinlin-linkedin: nit: just do `instanceConfigInitalized.set(true)` in an finally block.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+  /**
+   * Add new instance or update existing instance based on {@link InstanceConfig}(s).
+   * @param instanceConfigs the {@link InstanceConfig}(s) used to update in-mem cluster map.
+   * @throws Exception
+   */
+  private void addOrUpdateInstanceInfos(List<InstanceConfig> instanceConfigs) throws Exception {
+    for (InstanceConfig instanceConfig : instanceConfigs) {
+      int schemaVersion = getSchemaVersion(instanceConfig);
+      if (schemaVersion != 0) {
+        logger.error("Unknown InstanceConfig schema version: {}, ignoring.", schemaVersion);
```
justinlin-linkedin: throwing an exception here, just like SimpleClusterChangeHandler.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+
+  /**
+   * Initialize the disks and replicas on the given node. Create partitions if this is the first time a replica of
+   * that partition is being constructed. If partition override is enabled, the seal state of replica is determined by
+   * partition info in HelixPropertyStore, if disabled, the seal state is determined by instanceConfig.
+   * @param datanode the {@link AmbryDataNode} that is being initialized.
+   * @param instanceConfig the {@link InstanceConfig} associated with this datanode.
+   * @throws Exception if creation of {@link AmbryDisk} throws an Exception.
+   */
+  private void initializeDisksAndReplicasOnNode(AmbryDataNode datanode, InstanceConfig instanceConfig)
```
justinlin-linkedin: this function is quite similar to the one in the SimpleClusterChangeHandler. If we are going to keep two implementations for a long time, we can find a way to reuse the code.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -589,6 +614,68 @@ long getErrorCount() {
...
+        // into the map
+        currentPartition = partition;
+        // it doesn't really need to synchronize this method. "partitionNameToAmbryPartition" guarantees each thread
+        // will get same instance of ambry partition. The first one that succeeds adding partition into
+        // "partitionNameToAmbryPartition" will update "partitionMap".
+        partitionMap.put(ByteBuffer.wrap(currentPartition.getBytes()), currentPartition);
+        // update cluster-wide capacity
+        clusterWideAllocatedUsableCapacityBytes.getAndAdd(capacityBytes);
+      }
+      ambryPartitionToAmbryReplicas.putIfAbsent(currentPartition, ConcurrentHashMap.newKeySet());
```
justinlin-linkedin: we can move this statement to the if block as well, just like partitionMap.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+    if (clusterMapConfig.clusterMapEnablePartitionOverride && partitionOverrideInfoMap.containsKey(partitionName)) {
+      isSealed = partitionOverrideInfoMap.get(partitionName)
+          .get(ClusterMapUtils.PARTITION_STATE)
+          .equals(ClusterMapUtils.READ_ONLY_STR);
+    } else {
+      isSealed = sealedReplicas.contains(partitionName);
+    }
+    replica.setSealedState(isSealed);
+    replica.setStoppedState(stoppedReplicas.contains(partitionName));
+  }
```
justinlin-linkedin: same here, share this method with  SimpleClusterChangeHandler.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,487 @@
...
+    String instanceName = instanceConfig.getInstanceName();
+    logger.info("Updating replicas info for existing node {}", instanceName);
+    List<String> sealedReplicas = getSealedReplicas(instanceConfig);
+    List<String> stoppedReplicas = getStoppedReplicas(instanceConfig);
+    ConcurrentHashMap<String, AmbryReplica> currentReplicasOnNode =
+        ambryDataNodeToAmbryReplicas.get(instanceNameToAmbryDataNode.get(instanceName));
+    ConcurrentHashMap<String, AmbryReplica> replicasFromConfig = new ConcurrentHashMap<>();
+    Map<String, Map<String, String>> diskInfos = instanceConfig.getRecord().getMapFields();
+    for (Map.Entry<String, Map<String, String>> entry : diskInfos.entrySet()) {
+      String mountPath = entry.getKey();
```
justinlin-linkedin: I suggest check if the mountPath can map to a AmbryDisk in the ambryDataNodeToAmbryDisks first. If not, just throw an exception. We can't want to deal with the case of adding a disk.

And also check if it removes a disk or not.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: sure, will do

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: make sense

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: Yes, both of cluster change handlers should be able to share the code. I will probably do this in future PR after dynamic handler gets rolled out to production.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: will do

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
jsjtzyy: Unfortunately we can't. There could be an edge case where two threads from two dcs are attempting to modify the map concurrently. Will discuss with you offline.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: Removing disk is not in the plan but adding disk is actually supported in StorageManager. For now, we temporarily skip adding new disk in dynamic handler. We can support this if needed in the future.
I will take your suggestion to check if the mount path is valid.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,504 @@
...
+          logger.info("Instance config change triggered from {}", dcName);
+        }
+        logger.debug("Detailed instance configs in {} are: {}", dcName, configs);
+        try {
+          addOrUpdateInstanceInfos(configs);
+        } catch (Exception e) {
+          if (!instanceConfigInitialized.get()) {
+            logger.error("Exception occurred when initializing instances in {}: ", dcName, e);
+            initializationFailureMap.putIfAbsent(dcName, e);
+            instanceConfigInitialized.set(true);
```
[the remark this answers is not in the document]
jsjtzyy: correct, removed.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,504 @@
...
+
+
+/**
+ * A more dynamic implementation of {@link ClusterChangeHandler} which supports adding new nodes/partitions at runtime.
+ * It is also able to absorb replica location changes in cluster.
+ */
+public class DynamicClusterChangeHandler implements ClusterChangeHandler {
+  private final String dcName;
+  private final Object notificationLock = new Object();
+  private final AtomicBoolean instanceConfigInitialized = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
jsjtzyy: right, I have changed them to `volatile`

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,504 @@
...
+  /**
+   * Triggered whenever the IdealState in current data center has changed (for now, it is usually updated by Helix
+   * Bootstrap tool).
+   * @param idealState a list of {@link IdealState} that specifies ideal location of replicas.
+   * @param changeContext the {@link NotificationContext} associated.
+   * @throws InterruptedException
+   */
+  @Override
+  public void onIdealStateChange(List<IdealState> idealState, NotificationContext changeContext)
+      throws InterruptedException {
```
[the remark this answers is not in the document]
jsjtzyy: removed

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,500 @@
...
+            // if this is a new replica and doesn't exist on node
+            logger.info("Adding new replica {} to existing node {}", partitionName, instanceName);
+            long replicaCapacity = Long.valueOf(info[1]);
+            String partitionClass = info.length > 2 ? info[2] : clusterMapConfig.clusterMapDefaultPartitionClass;
+            // this can be a brand new partition that is added to an existing node
+            AmbryPartition mappedPartition =
+                new AmbryPartition(Long.valueOf(partitionName), partitionClass, helixClusterManagerCallback);
+            // Ensure only one AmbryPartition instance exists for specific partition.
+            mappedPartition = clusterChangeHandlerCallback.addPartitionIfAbsent(mappedPartition, replicaCapacity);
+            ensurePartitionAbsenceOnNodeAndValidateCapacity(mappedPartition, dataNode, replicaCapacity);
```
cgtz: General question: What happens when exceptions get thrown inside of helix listener callbacks? Would future notifications be received?

Throwing at this point in the middle of the loop could create some inconsistency since clusterChangeHandlerCallback.addReplicasToPartition is called on every iteration of the loop, but ambryDataNodeToAmbryReplicas is updated after the loop. If this is an issue, you could move the clusterChangeHandlerCallback.addReplicasToPartition call to after the loop.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,500 @@
...
+            clusterChangeHandlerCallback.addReplicasToPartition(mappedPartition, Collections.singletonList(replica));
+          }
+        }
+      }
+    }
+    // update ambryDataNodeToAmbryReplicas map by adding "replicasFromConfig"
+    ambryDataNodeToAmbryReplicas.put(instanceNameToAmbryDataNode.get(instanceName), replicasFromConfig);
+    // compare replicasFromConfig with current replica set to derive old replicas that are removed
+    Set<AmbryReplica> previousReplicas = new HashSet<>(currentReplicasOnNode.values());
+    previousReplicas.removeAll(replicasFromConfig.values());
```
cgtz: could this be done with partition IDs without the copy? 
```java
currentReplicasOnNode.keySet()
  .stream()
  .filter(partitionId -> !replicasFromConfig.containsKey(partitionId))
  .forEach(do delete);
```

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
cgtz: Could we remove SimpleClusterChangeHandler after we have gained enough confidence with DynamicClusterChangeHandler?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -596,6 +621,68 @@ long getErrorCount() {
...
+      return currentPartition;
+    }
+
+    /**
+     * Add a list of {@link AmbryReplica} to given {@link AmbryPartition} and update cluster-wide capacity stats
+     * @param partition the {@link AmbryPartition} which replicas should be added to.
+     * @param replicas list of {@link AmbryReplica} to be added.
+     */
+    void addReplicasToPartition(AmbryPartition partition, List<AmbryReplica> replicas) {
+      ambryPartitionToAmbryReplicas.computeIfAbsent(partition, k -> ConcurrentHashMap.newKeySet()).addAll(replicas);
```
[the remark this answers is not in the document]
jsjtzyy: It seems this method is called after `addPartitionIfAbsent` method is invoked but you're right it's better to call `addPartitionIfAbsent` if partition is not present (this should happen though).

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -0,0 +1,500 @@
...
+    List<String> stoppedReplicas = getStoppedReplicas(instanceConfig);
+    AmbryDataNode dataNode = instanceNameToAmbryDataNode.get(instanceName);
+    ConcurrentHashMap<String, AmbryReplica> currentReplicasOnNode = ambryDataNodeToAmbryReplicas.get(dataNode);
+    ConcurrentHashMap<String, AmbryReplica> replicasFromConfig = new ConcurrentHashMap<>();
+    Map<String, Map<String, String>> diskInfos = instanceConfig.getRecord().getMapFields();
+    for (Map.Entry<String, Map<String, String>> entry : diskInfos.entrySet()) {
+      String mountPath = entry.getKey();
+      Map<String, String> diskInfo = entry.getValue();
+      Optional<AmbryDisk> potentialDisk =
+          ambryDataNodeToAmbryDisks.get(dataNode).stream().filter(d -> d.getMountPath().equals(mountPath)).findFirst();
```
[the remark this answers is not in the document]
jsjtzyy: Good point. Let's create a map out of loop for now.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: I need to check with Helix team. Yeah, we need to move `addReplicasToPartition` out of loop.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: Changed

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -340,23 +343,37 @@ private void updateInstanceInfo(InstanceConfig instanceConfig) throws Exception
...
       }
     }
+    // update ambryDataNodeToAmbryReplicas map
+    replicaToAddByPartition.forEach(clusterChangeHandlerCallback::addReplicasToPartition);
     // update ambryDataNodeToAmbryReplicas map by adding "replicasFromConfig"
     ambryDataNodeToAmbryReplicas.put(instanceNameToAmbryDataNode.get(instanceName), replicasFromConfig);
-    // compare replicasFromConfig with current replica set to derive old replicas that are removed
-    Set<AmbryReplica> previousReplicas = new HashSet<>(currentReplicasOnNode.values());
+    // Derive old replicas that are removed and delete them from partition
+/*    Set<AmbryReplica> previousReplicas = new HashSet<>(currentReplicasOnNode.values());
```
cgtz: remove commented code

PR #1362, opened 2020-01-21
Fix the refCnt exception

justinlin-linkedin:
Fix the concurrency issue that causes release to be called for the object twice.


cgtz (approved):
I will merge after the build finishes

jsjtzyy (approved):

PR #1365, opened 2020-01-24
Azure batch deletes

lightningrob:


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -302,13 +317,47 @@ public boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value)
...
+      blobBatchClient.submitBatchWithResponse(blobBatch, false, Duration.ofHours(1), Context.NONE);
+      for (int j = 0; j < responseList.size(); j++) {
+        Response<Void> response = responseList.get(j);
+        CloudBlobMetadata blobMetadata = someBlobs.get(j);
+        // Note: Response.getStatusCode() throws exception on any error.
+        int statusCode;
+        try {
+          statusCode = response.getStatusCode();
+        } catch (BlobStorageException bex) {
+          statusCode = bex.getStatusCode();
```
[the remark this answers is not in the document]
lightningrob: They throw exception if the http status is not the "expected" value, which in this case is Accepted.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -55,12 +63,15 @@
...
   private final BlobServiceClient storageClient;
+  private final BlobBatchClient blobBatchClient;
   private final Configuration storageConfiguration;
   private final AzureMetrics azureMetrics;
   private final String clusterName;
   // Containers known to exist in the storage account
   private final Set<String> knownContainers = ConcurrentHashMap.newKeySet();
   private ProxyOptions proxyOptions;
+  // TODO: add to AzureCloudConfig
+  private int purgeBatchSize = 100;
```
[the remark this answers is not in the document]
lightningrob: Added to config class and moved comment there.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -302,13 +317,47 @@ public boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value)
...
+    List<List<CloudBlobMetadata>> partitionedLists = Lists.partition(blobMetadataList, purgeBatchSize);
+    for (List<CloudBlobMetadata> someBlobs : partitionedLists) {
+      BlobBatch blobBatch = blobBatchClient.getBlobBatch();
+      List<Response<Void>> responseList = new ArrayList<>();
+      for (CloudBlobMetadata blobMetadata : someBlobs) {
+        String containerName = getAzureContainerName(blobMetadata);
+        String blobName = getAzureBlobName(blobMetadata);
+        responseList.add(blobBatch.deleteBlob(containerName, blobName));
+      }
+      blobBatchClient.submitBatchWithResponse(blobBatch, false, Duration.ofHours(1), Context.NONE);
```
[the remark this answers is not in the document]
lightningrob: Made private final and changed to 1 minute.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -349,51 +352,48 @@ private boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value
...
+        deleteFromCosmos(blobMetadata);
       }
-      azureMetrics.blobDeletedCount.inc(deletionDone ? 1 : 0);
-      return deletionDone;
-    } catch (Exception e) {
-      azureMetrics.blobDeleteErrorCount.inc();
-      String error = (e instanceof DocumentClientException) ? "Failed to delete metadata document for blob " + blobId
-          : "Failed to delete blob " + blobId + ", storage path: " + containerName + "/" + blobFileName;
-      throw toCloudStorageException(error, e);
+      return deletedBlobs.size();
```
cgtz: I just want to confirm my understanding of the compaction logic: In the case where `deletedBlobs.size() < blobMetadataList.size()`, we rely on the fact that not deleting the record from cosmos will allow the dead blobs query to return the same blob ID in future compaction iterations?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -32,12 +36,15 @@
 import com.github.ambry.commons.BlobId;
 import com.github.ambry.config.CloudConfig;
 import com.github.ambry.utils.Utils;
+import com.google.common.collect.Lists;
```
cgtz: If we want to use guava directly in Ambry (I'm not against this since it has a lot of useful things), it might be good to explicitly reference the desired version in our build files. Interestingly, our dependencies seem to transitively bring in 3 versions of guava.

cgtz: If we decide not to use guava directly we can add a partition method to `Utils.java` (the impl is pretty simple), or just use `List::subList`

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -302,13 +318,46 @@ public boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value)
...
+        switch (statusCode) {
+          case HttpURLConnection.HTTP_OK:
+          case HttpURLConnection.HTTP_ACCEPTED:
+          case HttpURLConnection.HTTP_NOT_FOUND:
+          case HttpURLConnection.HTTP_GONE:
+            // blob was deleted or already gone
+            deletedBlobs.add(blobMetadata);
+            break;
+          default:
+            logger.error("Deleting blob {} got status {}", blobMetadata.getId(), statusCode);
```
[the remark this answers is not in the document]
lightningrob: Right.  They will be returned by the Cosmos query on the next compaction run (since we are not deleting those records).

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -302,13 +318,46 @@ public boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value)
...
+    List<List<CloudBlobMetadata>> partitionedLists = Lists.partition(blobMetadataList, purgeBatchSize);
+    for (List<CloudBlobMetadata> batchOfBlobs : partitionedLists) {
+      BlobBatch blobBatch = blobBatchClient.getBlobBatch();
+      List<Response<Void>> responseList = new ArrayList<>();
+      for (CloudBlobMetadata blobMetadata : batchOfBlobs) {
+        String containerName = getAzureContainerName(blobMetadata);
+        String blobName = getAzureBlobName(blobMetadata);
+        responseList.add(blobBatch.deleteBlob(containerName, blobName));
+      }
+      blobBatchClient.submitBatchWithResponse(blobBatch, false, Duration.ofSeconds(batchPurgeTimeoutSec), Context.NONE);
```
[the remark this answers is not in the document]
lightningrob: Good point.  I  think it's okay to let the RuntimeException be thrown, but will add to the javadoc.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
lightningrob: Right.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
lightningrob: Added Utils.partitionList() and refactored AzureCloudDestination.getBlobMetadata() to use it as well.

ambry-utils/src/main/java/com.github.ambry.utils/Utils.java
```diff
@@ -912,6 +913,30 @@ public static DataInputStream createDataInputStreamFromBuffer(Object buffer, boo
...
+   * @param batchSize the maximum size of the returned sublists.
+   * @return the partitioned list of sublists.
+   */
+  public static <T> List<List<T>> partitionList(List<T> inputList, int batchSize) {
+    Objects.requireNonNull(inputList, "Input list cannot be null");
+    if (batchSize < 1) {
+      throw new IllegalArgumentException("Invalid batchSize: " + batchSize);
+    }
+    List<List<T>> partitionedLists = new ArrayList<>();
+    for (int j = 0; j < inputList.size() / batchSize + 1; j++) {
```
cgtz: could this loop be simplified to:
```
for (int start = 0; start < inputList.size(); start += batchSize) {
  int end = Math.min(start + batchSize, inputList.size());
  partitionedLists.add(inputList.subList(start, end));
}
```
I may be missing an edge case here though

cgtz: So I was thinking about this partition thing and it seems that we don't really need a full list and just need an iterable that we can do a for loop on. Something like this could work, I think:
```java
  private static <T> Iterable<List<T>> partitionIterable(List<T> list, int batchSize) {
    return () -> new Iterator<List<T>>() {
      int nextStart = 0;

      @Override
      public boolean hasNext() {
        return nextStart < list.size();
      }

      @Override
      public List<T> next() {
        if (!hasNext()) {
          throw new NoSuchElementException();
        }
        int start = nextStart;
        nextStart += batchSize;
        return list.subList(start, Math.min(start + batchSize, list.size()));
      }
    };
  }
```

lightningrob: I'm not sure this is a big improvement.  I prefer to mimic the guava utility.  Once we upgrade to a newer JDK, we can probably retire the utility and use the Collections one.

lightningrob: I made the simplification you suggested.

PR #1366, opened 2020-01-24
Adding undelete record to persistent index

justinlin-linkedin:
Adding undelete record to persistent index. 

This would introduces new assumptions to the system.

1. Deletes are not the final state of blobs any more, instead, deletes and undeletes are the final possible state.
2. Deletes will be receiving less special treatment from the code. In findDeleteEntriesSince method, we have to exam if a deleted record is revived later.
3. Deletes following undelete will not have original message offset anymore.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/store/StoreErrorCodes.java
+++ b/ambry-api/src/main/java/com.github.ambry/store/StoreErrorCodes.java
@@ -35,5 +35,7 @@ public enum StoreErrorCodes {
   Already_Updated,
   Update_Not_Allowed,
   File_Not_Found,
-  Channel_Closed
+  Channel_Closed,
+  Life_Version_Conflict,
+  ID_Not_Deleted
 }
--- a/ambry-api/src/main/java/com.github.ambry/store/StoreErrorCodes.java
+++ b/ambry-api/src/main/java/com.github.ambry/store/StoreErrorCodes.java
@@ -37,5 +37,7 @@ public enum StoreErrorCodes {
   File_Not_Found,
   Channel_Closed,
   Life_Version_Conflict,
-  ID_Not_Deleted
+  ID_Not_Deleted,
+  ID_Undeleted,
+  ID_Deleted_Permanently
 }
--- a/ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
@@ -189,6 +189,15 @@ class IndexValue implements Comparable<IndexValue> {
   }
 
   /**
+   * Constructor to copy all data from a given {@link IndexValue}.
+   * @param other the given {@link IndexValue}.
+   */
+  IndexValue(IndexValue other) {
+    this(other.getSize(), other.getOffset(), other.getFlags(), other.getExpiresAtMs(), other.getOriginalMessageOffset(),
+        other.getOperationTimeInMs(), other.getAccountId(), other.getContainerId(), other.getLifeVersion());
+  }
+
+  /**
    * Constructs IndexValue based on the args passed
    * @param size the size of the blob that this index value refers to
    * @param offset the {@link Offset} in the {@link Log} where the blob that this index value refers to resides
--- a/ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
@@ -55,6 +55,8 @@ class IndexValue implements Comparable<IndexValue> {
 
   final static byte FLAGS_DEFAULT_VALUE = (byte) 0;
   final static long UNKNOWN_ORIGINAL_MESSAGE_OFFSET = -1;
+  // The life version when the operation is trigger by the requests from frontend.
+  final static short LIFE_VERSION_FROM_FRONTEND = -1;
 
   private final static int BLOB_SIZE_IN_BYTES = 8;
   private final static int OFFSET_SIZE_IN_BYTES = 8;
@@ -210,7 +212,7 @@ class IndexValue implements Comparable<IndexValue> {
     this.accountId = accountId;
     this.containerId = containerId;
     this.lifeVersion = lifeVersion;
-    formatVersion = PersistentIndex.CURRENT_VERSION;
+    formatVersion = PersistentIndex.getVersion();
   }
 
   /**
@@ -241,6 +243,38 @@ class IndexValue implements Comparable<IndexValue> {
    */
   boolean isFlagSet(Flags flag) {
     return ((getFlags() & (1 << flag.ordinal())) != 0);
+  }
+
+  /**
+   * Helper function for isFlagSet(Flags.Ttl_Update_Index).
+   * @return true when the Ttl_Update_Index is set.
+   */
+  boolean isTTLUpdate() {
+    return isFlagSet(Flags.Ttl_Update_Index);
+  }
+
+  /**
+   * Helper function for isFlagSet(Flags.Delete_Index).
+   * @return true when the Delete_Index is set.
+   */
+  boolean isDelete() {
+    return isFlagSet(Flags.Delete_Index);
+  }
+
+  /**
+   * Helper function for isFlagSet(Flags.Undelete_Index).
+   * @return true when the Undelete_Index is set.
+   */
+  boolean isUndelete() {
+    return isFlagSet(Flags.Undelete_Index);
+  }
+
+  /**
+   * Helper function to decide if this value is a put value or not.
+   * @return true when it's not a put record.
+   */
+  boolean isPut() {
+    return !(isTTLUpdate() || isDelete() || isUndelete());
   }
 
   /**
@@ -277,6 +311,15 @@ class IndexValue implements Comparable<IndexValue> {
    */
   short getContainerId() {
     return containerId;
+  }
+
+  /**
+   * True when the life version is not from frontend requests.
+   * @param lifeVersion the given life version.
+   * @return true when it's not from frontend requests.
+   */
+  static boolean hasLifeVersion(short lifeVersion) {
+    return lifeVersion > LIFE_VERSION_FROM_FRONTEND;
   }
 
   /**
@@ -386,11 +429,11 @@ class IndexValue implements Comparable<IndexValue> {
   @Override
   public String toString() {
     return "Offset: " + offset + ", Size: " + getSize() + ", Deleted: " + isFlagSet(Flags.Delete_Index)
-        + ", TTL Updated: " + isFlagSet(Flags.Ttl_Update_Index) + ", Undelete: " + isFlagSet(
-        Flags.Undelete_Index) + ", ExpiresAtMs: " + getExpiresAtMs() + ", Original Message Offset: "
-        + getOriginalMessageOffset() + (formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeAtSecs "
-        + getOperationTimeInMs() + ", AccountId " + getAccountId() + ", ContainerId " + getContainerId())
-        : "") + (formatVersion > PersistentIndex.VERSION_2 ? ", Life Version:" + lifeVersion : "");
+        + ", TTL Updated: " + isFlagSet(Flags.Ttl_Update_Index) + ", Undelete: " + isFlagSet(Flags.Undelete_Index)
+        + ", ExpiresAtMs: " + getExpiresAtMs() + ", Original Message Offset: " + getOriginalMessageOffset() + (
+        formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeAtSecs " + getOperationTimeInMs() + ", AccountId "
+            + getAccountId() + ", ContainerId " + getContainerId()) : "") + (formatVersion > PersistentIndex.VERSION_2 ?
+        ", Life Version:" + lifeVersion : "");
   }
 
   /**
--- a/ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
@@ -212,7 +212,7 @@ class IndexValue implements Comparable<IndexValue> {
     this.accountId = accountId;
     this.containerId = containerId;
     this.lifeVersion = lifeVersion;
-    formatVersion = PersistentIndex.getVersion();
+    formatVersion = PersistentIndex.CURRENT_VERSION;
   }
 
   /**
@@ -274,7 +274,7 @@ class IndexValue implements Comparable<IndexValue> {
    * @return true when it's not a put record.
    */
   boolean isPut() {
-    return !(isTTLUpdate() || isDelete() || isUndelete());
+    return flags == FLAGS_DEFAULT_VALUE;
   }
 
   /**
@@ -428,12 +428,11 @@ class IndexValue implements Comparable<IndexValue> {
 
   @Override
   public String toString() {
-    return "Offset: " + offset + ", Size: " + getSize() + ", Deleted: " + isFlagSet(Flags.Delete_Index)
-        + ", TTL Updated: " + isFlagSet(Flags.Ttl_Update_Index) + ", Undelete: " + isFlagSet(Flags.Undelete_Index)
-        + ", ExpiresAtMs: " + getExpiresAtMs() + ", Original Message Offset: " + getOriginalMessageOffset() + (
-        formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeAtSecs " + getOperationTimeInMs() + ", AccountId "
-            + getAccountId() + ", ContainerId " + getContainerId()) : "") + (formatVersion > PersistentIndex.VERSION_2 ?
-        ", Life Version:" + lifeVersion : "");
+    return "Offset: " + offset + ", Size: " + getSize() + ", Deleted: " + isDelete() + ", TTL Updated: " + isTTLUpdate()
+        + ", Undelete: " + isUndelete() + ", ExpiresAtMs: " + getExpiresAtMs() + ", Original Message Offset: "
+        + getOriginalMessageOffset() + (formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeAtSecs "
+        + getOperationTimeInMs() + ", AccountId " + getAccountId() + ", ContainerId " + getContainerId()) : "") + (
+        formatVersion > PersistentIndex.VERSION_2 ? ", Life Version:" + lifeVersion : "");
   }
 
   /**
--- a/ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
@@ -323,6 +323,17 @@ class PersistentIndex {
               dataDir, info.getStoreKey(), info.getSize(), info.getExpirationTimeInMs(), info.getLifeVersion());
           // removes from the tracking structure if a delete was being expected for the key
           deleteExpectedKeys.remove(info.getStoreKey());
+        } else if (info.isUndeleted()) {
+          markAsUndeleted(info.getStoreKey(), new FileSpan(runningOffset, infoEndOffset), info.getOperationTimeMs(),
+              info.getLifeVersion());
+          logger.info(
+              "Index : {} updated message with key {} by inserting undelete entry of size {} ttl {} lifeVersion {}",
+              dataDir, info.getStoreKey(), info.getSize(), info.getExpirationTimeInMs(), info.getLifeVersion());
+          if (value == null) {
+            // Undelete record indicates that a put and/or a delete record were expected.
+            throw new StoreException("Put record were expected but were not encountered for key: " + info.getStoreKey(),
+                StoreErrorCodes.Initialization_Error);
+          }
         } else if (info.isTtlUpdated()) {
           markAsPermanent(info.getStoreKey(), new FileSpan(runningOffset, infoEndOffset), info,
               info.getOperationTimeMs(), info.getLifeVersion());
@@ -334,17 +345,6 @@ class PersistentIndex {
             // a DELETE for this key (because the PUT record is gone, compaction must have cleaned it up because a
             // DELETE must have been present)
             deleteExpectedKeys.add(info.getStoreKey());
-          }
-        } else if (info.isUndeleted()) {
-          markAsUndeleted(info.getStoreKey(), new FileSpan(runningOffset, infoEndOffset), info.getOperationTimeMs(),
-              info.getLifeVersion());
-          logger.info(
-              "Index : {} updated message with key {} by inserting undelete update entry of size {} ttl {} lifeVersion {}",
-              dataDir, info.getStoreKey(), info.getSize(), info.getExpirationTimeInMs(), info.getLifeVersion());
-          if (value == null) {
-            // Undelete record indicates that there might be a put and delete record before it.
-            throw new StoreException("Put record were expected but were not encountered for key: " + info.getStoreKey(),
-                StoreErrorCodes.Initialization_Error);
           }
         } else if (value != null) {
           throw new StoreException("Illegal message state during recovery. Duplicate PUT record",
@@ -648,7 +648,8 @@ class PersistentIndex {
 
   /**
    * Finds all the {@link IndexValue}s associated with the given {@code key} that matches any of the provided {@code types}
-   * if present in the index with the given {@code fileSpan} and return them in reversed chronological order.
+   * if present in the index with the given {@code fileSpan} and return them in reversed chronological order. If there is
+   * no matched {@link IndexValue}, this method would return null;
    * @param key the {@link StoreKey} whose {@link IndexValue} is required.
    * @param fileSpan {@link FileSpan} which specifies the range within which search should be made.
    * @param types the types of {@link IndexEntryType} to look for.
@@ -663,12 +664,12 @@ class PersistentIndex {
     try {
       ConcurrentNavigableMap<Offset, IndexSegment> segmentsMapToSearch;
       if (fileSpan == null) {
-        logger.trace("Searching for " + key + " in the entire index");
+        logger.trace("Searching all indexes for " + key + " in the entire index");
         segmentsMapToSearch = indexSegments.descendingMap();
       } else {
         logger.trace(
-            "Searching for " + key + " in index with filespan ranging from " + fileSpan.getStartOffset() + " to "
-                + fileSpan.getEndOffset());
+            "Searching all indexes for " + key + " in index with filespan ranging from " + fileSpan.getStartOffset()
+                + " to " + fileSpan.getEndOffset());
         segmentsMapToSearch = indexSegments.subMap(indexSegments.floorKey(fileSpan.getStartOffset()), true,
             indexSegments.floorKey(fileSpan.getEndOffset()), true).descendingMap();
         metrics.segmentSizeForExists.update(segmentsMapToSearch.size());
@@ -676,7 +677,7 @@ class PersistentIndex {
       int segmentsSearched = 0;
       for (Map.Entry<Offset, IndexSegment> entry : segmentsMapToSearch.entrySet()) {
         segmentsSearched++;
-        logger.trace("Index : {} searching index with start offset {}", dataDir, entry.getKey());
+        logger.trace("Index : {} searching all indexes with start offset {}", dataDir, entry.getKey());
         NavigableSet<IndexValue> values = entry.getValue().find(key);
         if (values != null) {
           if (result == null) {
@@ -687,13 +688,10 @@ class PersistentIndex {
             IndexValue value = it.next();
             if ((types.contains(IndexEntryType.DELETE) && value.isDelete()) || (types.contains(IndexEntryType.UNDELETE)
                 && value.isUndelete()) || (types.contains(IndexEntryType.TTL_UPDATE) && !value.isDelete()
-                && value.isTTLUpdate()) || (types.contains(IndexEntryType.PUT) && value.isPut())) {
+                && !value.isUndelete() && value.isTTLUpdate()) || (types.contains(IndexEntryType.PUT)
+                && value.isPut())) {
               // Add a copy of the value to the result since we return a modifiable list to the caller.
-              IndexValue newValue =
-                  new IndexValue(value.getSize(), value.getOffset(), value.getFlags(), value.getExpiresAtMs(),
-                      value.getOperationTimeInMs(), value.getAccountId(), value.getContainerId(),
-                      value.getLifeVersion());
-              result.add(newValue);
+              result.add(new IndexValue(value));
             }
           }
         }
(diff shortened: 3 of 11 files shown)


ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -712,37 +897,106 @@ IndexValue markAsPermanent(StoreKey id, FileSpan fileSpan, long operationTimeMs)
    * @return the {@link IndexValue} of the ttl update record
    * @throws StoreException if there is any problem writing the index record
    */
-  private IndexValue markAsPermanent(StoreKey id, FileSpan fileSpan, MessageInfo info, long operationTimeMs)
-      throws StoreException {
+  private IndexValue markAsPermanent(StoreKey id, FileSpan fileSpan, MessageInfo info, long operationTimeMs,
+      short lifeVersion) throws StoreException {
```
jsjtzyy: update the java doc of this method by adding `lifeVersion`

ambry-api/src/main/java/com.github.ambry/store/StoreErrorCodes.java
```diff
@@ -35,5 +35,9 @@
   Already_Updated,
   Update_Not_Allowed,
   File_Not_Found,
-  Channel_Closed
+  Channel_Closed,
+  Life_Version_Conflict,
+  ID_Not_Deleted,
+  ID_Undeleted,
+  ID_Deleted_Permanently
```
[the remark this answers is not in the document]
justinlin-linkedin: we still need it, it will be returned when ambry-frontend requests to delete a already deleted blob.

ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
```diff
@@ -55,6 +55,8 @@
 
   final static byte FLAGS_DEFAULT_VALUE = (byte) 0;
   final static long UNKNOWN_ORIGINAL_MESSAGE_OFFSET = -1;
+  // The life version when the operation is trigger by the requests from frontend.
+  final static short LIFE_VERSION_FROM_FRONTEND = -1;
```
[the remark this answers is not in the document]
justinlin-linkedin: frontend requests don't carry any lifeVersion information. This just indicates that markAsPermanent, markAsDeleted and markAsUndeleted is called by the requests from ambry frontend. When it's not from frontend, it should be from recovery and replication.

In the end, we will remove this variable. In the end, after we start to support lifeVersion, we will have to know lifeVersion for every frontend and replication requests.

ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
```diff
@@ -279,6 +313,15 @@ short getContainerId() {
     return containerId;
   }
 
+  /**
+   * True when the life version is not from frontend requests.
+   * @param lifeVersion the given life version.
+   * @return true when it's not from frontend requests.
+   */
+  static boolean hasLifeVersion(short lifeVersion) {
+    return lifeVersion > LIFE_VERSION_FROM_FRONTEND;
```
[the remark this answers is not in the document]
justinlin-linkedin: and recovery also

ambry-store/src/main/java/com.github.ambry.store/IndexValue.java
```diff
@@ -243,6 +245,38 @@ boolean isFlagSet(Flags flag) {
...
+  boolean isUndelete() {
+    return isFlagSet(Flags.Undelete_Index);
+  }
+
+  /**
+   * Helper function to decide if this value is a put value or not.
+   * @return true when it's not a put record.
+   */
+  boolean isPut() {
+    return !(isTTLUpdate() || isDelete() || isUndelete());
```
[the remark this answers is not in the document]
justinlin-linkedin: make sense.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -712,37 +897,106 @@ IndexValue markAsPermanent(StoreKey id, FileSpan fileSpan, long operationTimeMs)
...
     }
     long size = fileSpan.getEndOffset().getOffset() - fileSpan.getStartOffset().getOffset();
     IndexValue newValue;
+
     if (value == null) {
       // It is possible that the PUT has been cleaned by compaction
+      // but the TTL update is going to still be placed?
+      if (!hasLifeVersion) {
+        throw new StoreException("MessageInfo of ttlUpdate carries invalid lifeVersion",
+            StoreErrorCodes.Initialization_Error);
```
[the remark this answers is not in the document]
justinlin-linkedin: If it doesn't have lifeVersion, it mean lifeVersion is -1. This will never happen, for 1 when the method is called in recovery or replication, we already the lifeVersion from MessageInfo 2 when the method is called in handling frontend requests, we will have a put record in the persistent index so the value == null will be false.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -332,14 +357,24 @@ private void recover(MessageStoreRecovery recovery) throws StoreException, IOExc
             // DELETE must have been present)
             deleteExpectedKeys.add(info.getStoreKey());
           }
+        } else if (info.isUndeleted()) {
```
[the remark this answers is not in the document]
justinlin-linkedin: It does, I have to rebase.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -518,17 +532,18 @@ private boolean needToRollOverIndex(IndexEntry entry) {
...
-   * Finds a key in the index and returns the blob index value associated with it. If not found,
-   * returns null
+   * Finds {@link IndexValue} that represents the latest state of the given {@code key} in the index and return it. If
+   * not found, returns null.
    * <br>
-   * This method only returns PUT or DELETE index entries. It does not return TTL_UPDATE entries but accounts for
-   * TTL updates by updating the flag and expiry time (if applicable).
+   * This method returns the final state of the given {@code key}. The final state of a key can be a Put value, a Delete
+   * value and a Undelete value. Ttl_update isn't considered as final state as it just update the expiration date.
+   * {@link IndexValue} returned by this method would carry expiration date from Ttl_update if there is one.
```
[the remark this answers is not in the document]
zzmao: Or, can we set it permanent if a blob with TTL is undeleted?  To make TTL_UPDATE easier.

ambry-api/src/main/java/com.github.ambry/store/StoreErrorCodes.java
```diff
@@ -35,5 +35,9 @@
   Already_Updated,
   Update_Not_Allowed,
   File_Not_Found,
-  Channel_Closed
+  Channel_Closed,
+  Life_Version_Conflict,
```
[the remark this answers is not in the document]
justinlin-linkedin: Feel like conflict is better here, invalid seems to indicate the lifeVersion is not a valid number.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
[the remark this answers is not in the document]
justinlin-linkedin: talked offline.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+   * Finds all the {@link IndexValue}s associated with the given {@code key} that matches any of the provided {@code types}
+   * if present in the index with the given {@code fileSpan} and return them in reversed order.
+   * @param key the {@link StoreKey} whose {@link IndexValue} is required.
+   * @param fileSpan {@link FileSpan} which specifies the range within which search should be made.
+   * @param types the types of {@link IndexEntryType} to look for.
+   * @param indexSegments the map of index segment start {@link Offset} to {@link IndexSegment} instances
+   * @return The list of the {@link IndexValue}s for {@code key} conforming to one of the types {@code types}.
+   * @throws StoreException any error.
+   */
+  List<IndexValue> findAllIndexValuesForKeyInReverseOrder(StoreKey key, FileSpan fileSpan,
```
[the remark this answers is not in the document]
justinlin-linkedin: I can do that. Given that this is the only order we present the index values.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -839,6 +1099,33 @@ private BlobReadOptions getDeletedBlobReadOptions(IndexValue value, StoreKey key
...
 
+  /**
+   * Gets {@link BlobReadOptions} for a undeleted blob.
+   * @param value the {@link IndexValue} of the delete index entry for the blob.
+   * @param key the {@link StoreKey} for which {@code value} is the delete {@link IndexValue}
+   * @param indexSegments the map of index segment start {@link Offset} to {@link IndexSegment} instances
+   * @return the {@link BlobReadOptions} that contains the information for the given {@code id}
+   * @throws StoreException
+   */
+  private BlobReadOptions getUndeletedBlobReadOptions(IndexValue value, StoreKey key,
```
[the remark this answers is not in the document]
justinlin-linkedin: talked offline.

ambry-store/src/test/java/com.github.ambry.store/CuratedLogIndexState.java
```diff
@@ -77,7 +77,10 @@
   // deliberately do not divide the capacities perfectly.
   static final long PUT_RECORD_SIZE = 53;
   static final long DELETE_RECORD_SIZE = 29;
-  static final long TTL_UPDATE_RECORD_SIZE = 37;
+  static final long TTL_UPDATE_RECORD_SIZE = 29;
```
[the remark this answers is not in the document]
justinlin-linkedin: That's minor change, size doesn't really matter, it seems like those are just random prime number greater than 24. I added some new records for undelete and the log segment is over flowed. I will change this back and change the log segment size instead.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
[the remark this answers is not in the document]
zzmao: yes.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -590,15 +605,17 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+            if (types.contains(IndexEntryType.DELETE) && value.isDelete()) {
+              retCandidate = value;
+              break;
+            } else if (types.contains(IndexEntryType.UNDELETE) && value.isUndelete()) {
               retCandidate = value;
               break;
-            } else if (types.contains(IndexEntryType.TTL_UPDATE) && !value.isFlagSet(IndexValue.Flags.Delete_Index)
-                && value.isFlagSet(IndexValue.Flags.Ttl_Update_Index)) {
+            } else if (types.contains(IndexEntryType.TTL_UPDATE) && !value.isDelete() && !value.isUndelete()
+                && value.isTTLUpdate()) {
```
jsjtzyy: Can you present a case where we do need to check `!value.isDelete() && !value.isUndelete()`? I know the piece of code comes from previous version, I just don't quite understand why we need such additional check.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+   * Finds all the {@link IndexValue}s associated with the given {@code key} that matches any of the provided {@code types}
+   * if present in the index with the given {@code fileSpan} and return them in reversed chronological order.
+   * @param key the {@link StoreKey} whose {@link IndexValue} is required.
+   * @param fileSpan {@link FileSpan} which specifies the range within which search should be made.
+   * @param types the types of {@link IndexEntryType} to look for.
+   * @param indexSegments the map of index segment start {@link Offset} to {@link IndexSegment} instances
+   * @return The list of the {@link IndexValue}s for {@code key} conforming to one of the types {@code types}.
+   * @throws StoreException any error.
+   */
+  List<IndexValue> findAllIndexValuesForKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryType> types,
```
jsjtzyy: can be private

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+      ConcurrentSkipListMap<Offset, IndexSegment> indexSegments) throws StoreException {
+    List<IndexValue> result = null;
+    final Timer.Context context = metrics.findTime.time();
+    try {
+      ConcurrentNavigableMap<Offset, IndexSegment> segmentsMapToSearch;
+      if (fileSpan == null) {
+        logger.trace("Searching for " + key + " in the entire index");
```
jsjtzyy: minor: change to `Searching all index values for` to distinguish from findKey method

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+        segmentsMapToSearch = indexSegments.descendingMap();
+      } else {
+        logger.trace(
+            "Searching for " + key + " in index with filespan ranging from " + fileSpan.getStartOffset() + " to "
```
jsjtzyy: same here

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+                + fileSpan.getEndOffset());
+        segmentsMapToSearch = indexSegments.subMap(indexSegments.floorKey(fileSpan.getStartOffset()), true,
+            indexSegments.floorKey(fileSpan.getEndOffset()), true).descendingMap();
+        metrics.segmentSizeForExists.update(segmentsMapToSearch.size());
+      }
+      int segmentsSearched = 0;
+      for (Map.Entry<Offset, IndexSegment> entry : segmentsMapToSearch.entrySet()) {
+        segmentsSearched++;
+        logger.trace("Index : {} searching index with start offset {}", dataDir, entry.getKey());
```
jsjtzyy: same here

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+        if (values != null) {
+          if (result == null) {
+            result = new LinkedList<>();
+          }
+          Iterator<IndexValue> it = values.descendingIterator();
+          while (it.hasNext()) {
+            IndexValue value = it.next();
+            if ((types.contains(IndexEntryType.DELETE) && value.isDelete()) || (types.contains(IndexEntryType.UNDELETE)
+                && value.isUndelete()) || (types.contains(IndexEntryType.TTL_UPDATE) && !value.isDelete()
+                && value.isTTLUpdate()) || (types.contains(IndexEntryType.PUT) && value.isPut())) {
```
jsjtzyy: `types.contains(IndexEntryType.TTL_UPDATE` logic doesn't consider !value.isUndelete() like that in `findKey` method. Any specific reason?

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+              // Add a copy of the value to the result since we return a modifiable list to the caller.
+              IndexValue newValue =
+                  new IndexValue(value.getSize(), value.getOffset(), value.getFlags(), value.getExpiresAtMs(),
```
jsjtzyy: Optional: probably worth adding a constructor for IndexValue that takes in another index value for deep copy purpose. (Pass in many parameters is error-prone)

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+            }
+          }
+        }
+      }
+      metrics.segmentsAccessedPerBlobCount.update(segmentsSearched);
+    } finally {
+      context.stop();
+    }
+    if (result != null) {
+      logger.trace("Index: {} Returninng values {}", dataDir, result);
```
jsjtzyy: typo `Returning`

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+    }
+    return result;
```
jsjtzyy: Looks like result may be null. Can you explicitly mention this in java doc?

Also, if null means not find index value related to given key. We can make result = new ArrayList<>() at the very beginning and in the end, `return result.isEmpty() ? null : result;` 

Then we can remove
```
 if (result == null) {
            result = new LinkedList<>();
  }
```

jsjtzyy: Or probably we can accept an empty result as a return value.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+  void validateSanityForUndelete(StoreKey key, List<IndexValue> values, short lifeVersion) throws StoreException {
+    if (values == null || values.isEmpty()) {
+      throw new StoreException("Id " + key + " not present in index " + dataDir, StoreErrorCodes.ID_Not_Found);
+    }
+    if (!IndexValue.hasLifeVersion(lifeVersion)) {
+      validateSanityForUndeleteWithoutLifeVersion(key, values);
+      return;
+    }
+    IndexValue firstValue = values.get(values.size() - 1);
+    IndexValue lastValue = values.get(0);
```
jsjtzyy: I get you point here but I would suggest reverse values before hand which makes logic easier to understand and more consistent with your java docs of this method.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+    if (!firstValue.isPut()) {
+      throw new StoreException("Id " + key + " requires first value to be a put in index " + dataDir,
+          StoreErrorCodes.ID_Deleted_Permanently);
+    }
+    if (lastValue.getLifeVersion() >= lifeVersion) {
+      throw new StoreException(
+          "LifeVersion conflict in index. Id " + key + " LifeVersion: " + lastValue.getLifeVersion()
+              + " Undelete LifeVersion: " + lifeVersion, StoreErrorCodes.Life_Version_Conflict);
+    }
```
jsjtzyy: Why we didn't check last value type and expiration time of PUT in this method?

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,165 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+  }
+
+  /**
+   * Change the target's expiration date and set the ttl_update_index to be true if there is a ttl index in given list.
+   * @param target the {@link IndexValue} to change expiration date.
+   * @param allValues the given list of {@link IndexValue}s.
+   */
+  void maybeChangeExpirationDate(IndexValue target, List<IndexValue> allValues) {
+    if (target.isTTLUpdate()) {
+      return;
```
jsjtzyy: if target is already updated or expiration time = -1, in either case we can skip update

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -658,33 +834,55 @@ IndexValue markAsDeleted(StoreKey id, FileSpan fileSpan, long deletionTimeMs) th
...
+          throw new StoreException("Id " + id + " already deleted in index " + dataDir, StoreErrorCodes.ID_Deleted);
+        }
+      }
     }
     long size = fileSpan.getEndOffset().getOffset() - fileSpan.getStartOffset().getOffset();
     IndexValue newValue;
     if (value == null) {
       // It is possible that the PUT has been cleaned by compaction
+      if (!hasLifeVersion) {
+        throw new StoreException("MessageInfo of delete carry invalid lifeVersion",
```
jsjtzyy: nit: `carries`

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -658,33 +834,55 @@ IndexValue markAsDeleted(StoreKey id, FileSpan fileSpan, long deletionTimeMs) th
...
-              value.getAccountId(), value.getContainerId(), (short) 0);
+              value.getAccountId(), value.getContainerId(), lifeVersion);
       newValue.setNewOffset(fileSpan.getStartOffset());
+      // Only set the original message offset when the value is put
+      if (!value.isPut()) {
+        newValue.clearOriginalMessageOffset();
+      }
       newValue.setNewSize(size);
     }
+    newValue.clearFlag(IndexValue.Flags.Undelete_Index);
```
jsjtzyy: should we clear TTL Update flag as well?

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -315,31 +316,44 @@ private void recover(MessageStoreRecovery recovery) throws StoreException, IOExc
...
           }
+        } else if (info.isUndeleted()) {
+          markAsUndeleted(info.getStoreKey(), new FileSpan(runningOffset, infoEndOffset), info.getOperationTimeMs(),
+              info.getLifeVersion());
+          logger.info(
+              "Index : {} updated message with key {} by inserting undelete update entry of size {} ttl {} lifeVersion {}",
+              dataDir, info.getStoreKey(), info.getSize(), info.getExpirationTimeInMs(), info.getLifeVersion());
+          if (value == null) {
+            // Undelete record indicates that there might be a put and delete record before it.
+            throw new StoreException("Put record were expected but were not encountered for key: " + info.getStoreKey(),
```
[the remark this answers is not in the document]
justinlin-linkedin: it's not possible, if a put is already compacted, then an undelete operation would be rejected since undelete doesn't do anything.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: A delete or undelete entry will carry TTL_UPDATE_INDEX flag if there is a ttl_update record comes before it. So here if the search types only contain TTL_UPDATE_INDEX and we are seeing a delete record, then we have to make sure it doesn't get returned, therefore, we have to check `!value.isDelete() && !value.isUndelete()`.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: I have to expose this method in package scope, since in later PR, blobstore will call this method to do a precheck before adding undelete record to persistent index.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: good catch, it's my bad, will update.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: make sense, will add a new constructor.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: I think returning a null is find here. I will comment it.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: I think the order in list is correct, i will change the variable name.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: you are right, we should check the the expiration date.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: we don't, a delete or undelete can carry ttl_update_index flag and it's expiration date, since a delete and a undelete should represent the final state of a key. And expiration date is the final state of this key.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -709,40 +909,100 @@ IndexValue markAsPermanent(StoreKey id, FileSpan fileSpan, long operationTimeMs)
...
       newValue.setNewSize(size);
     }
+    newValue.clearFlag(IndexValue.Flags.Undelete_Index);
     newValue.setFlag(IndexValue.Flags.Ttl_Update_Index);
     addToIndex(new IndexEntry(id, newValue, null), fileSpan);
     return newValue;
   }
 
+  /**
+   * Marks a blob as permanent
```
jsjtzyy: typo: as undelete

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -709,40 +909,100 @@ IndexValue markAsPermanent(StoreKey id, FileSpan fileSpan, long operationTimeMs)
...
+    } else if (hasLifeVersion && retrievedLifeVersion > lifeVersion) {
+      throw new StoreException("LifeVersion conflict in index. Id " + id + " LifeVersion: " + retrievedLifeVersion
+          + " Undelete LifeVersion: " + lifeVersion, StoreErrorCodes.Life_Version_Conflict);
     }
     long size = fileSpan.getEndOffset().getOffset() - fileSpan.getStartOffset().getOffset();
     IndexValue newValue;
+
     if (value == null) {
       // It is possible that the PUT has been cleaned by compaction
+      // but the TTL update is going to still be placed?
```
jsjtzyy: I have same question here, do you have any idea why TTL update is placed even when PUT has been cleaned up?

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -629,6 +646,167 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
+            StoreErrorCodes.ID_Not_Deleted);
+      } else {
+        throw new StoreException("Id " + key + " is already undeleted in index" + dataDir,
+            StoreErrorCodes.ID_Undeleted);
+      }
+    }
+    // First item has to be put and last item has to be a delete.
+    // PutRecord can't expire and delete record can't be older than the delete retention time.
+    IndexValue firstValue = values.get(0);
+    IndexValue lastValue = values.get(values.size() - 1);
```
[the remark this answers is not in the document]
justinlin-linkedin: updated.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: i don't think it's possible. But I will take another look at later PR

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -839,6 +1098,33 @@ private BlobReadOptions getDeletedBlobReadOptions(IndexValue value, StoreKey key
     return readOptions;
   }
 
+  /**
+   * Gets {@link BlobReadOptions} for a undeleted blob.
+   * @param value the {@link IndexValue} of the delete index entry for the blob.
+   * @param key the {@link StoreKey} for which {@code value} is the delete {@link IndexValue}
```
jsjtzyy: please update comment for `value` and `key` parameters (change delete to undelete)

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -839,6 +1098,33 @@ private BlobReadOptions getDeletedBlobReadOptions(IndexValue value, StoreKey key
...
+  private BlobReadOptions getUndeletedBlobReadOptions(IndexValue value, StoreKey key,
+      ConcurrentSkipListMap<Offset, IndexSegment> indexSegments) throws StoreException {
+    IndexValue putValue =
+        findKey(key, new FileSpan(getStartOffset(indexSegments), value.getOffset()), EnumSet.of(IndexEntryType.PUT),
+            indexSegments);
+    if (putValue != null) {
+      // use the expiration time from the original value because it may have been updated
+      // since we are here dealing with undelete blob, we have to return the right life version
+      return new BlobReadOptions(log, putValue.getOffset(),
+          new MessageInfo(key, putValue.getSize(), false, value.isFlagSet(IndexValue.Flags.Ttl_Update_Index), true,
```
jsjtzyy: nit: `value.isTTLUpdate()`

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1555,6 +1839,10 @@ FindInfo findDeletedEntriesSince(FindToken token, long maxTotalSizeOfEntries, lo
         }
       }
     }
+    // Filter out all the messages that are not "deleted", then update state for remaining deleted message.
+    filterDeleteEntries(messageEntries);
+    updateStateForMessages(messageEntries);
+
     filterDeleteEntries(messageEntries);
```
jsjtzyy: why here we filter delete entries again? Is this because messageInfo is updated in `updateStateForMessages`?

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1342,17 +1627,16 @@ private void updateStateForMessages(List<MessageInfo> messageEntries) throws Sto
...
+      // for all the message info, we use most recent ref even for delete. since a deleted entry can be undeleted.
+      // ok to use most recent ref
+      IndexValue indexValue = findKey(messageInfo.getStoreKey(), null,
+          EnumSet.of(IndexEntryType.TTL_UPDATE, IndexEntryType.DELETE, IndexEntryType.UNDELETE));
+      if (indexValue != null) {
+        messageInfo = new MessageInfo(messageInfo.getStoreKey(), indexValue.getSize(), indexValue.isDelete(),
+            indexValue.isTTLUpdate(), indexValue.isUndelete(), indexValue.getExpiresAtMs(), null,
+            indexValue.getAccountId(), indexValue.getContainerId(), indexValue.getOperationTimeInMs(),
+            indexValue.getLifeVersion());
+        messageEntriesIterator.set(messageInfo);
```
jsjtzyy: minor: please update java doc for this method.

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
justinlin-linkedin: talked offline

PR #1367, opened 2020-01-24
Fix additional release bug

justinlin-linkedin:
There is another place where forEach method is used to release the ResponseInfo. This PR fixes that.


ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -449,16 +448,24 @@ void completeRead() {
...
+        releaseResourse();
       }
       operationCompleted = true;
     }
 
+    /**
+     * Release all the {@link ResponseInfo} in the map. Use {@link ConcurrentHashMap#remove(Object)} method to avoid
+     * conflict with the release call in the chunk async callback.
+     */
+    private void releaseResourse() {
```
cgtz: minor spelling: `releaseResource`.

PR #1368, opened 2020-01-28
Introduce clustermap change listener to propagate remote replica addition/removal events

jsjtzyy:
With "move replica", Helix cluster manager should be able to dynamically add/remove remote replica infos and therefore clustermap will be dynamically changed. External components like ReplicationManager needs to update thread assignment and some in-mem data structures in repsonse to such changes. This PR introduces clustermap change listener which allows ReplicationManager to listen to any change in 
clustermap and take actions accordingly. For now, there is only one method in clustermap
change listener: onReplicaAddedOrRemoved(), we could add more methods like onReplicaLeadershipChange etc if needed.

---
Remote replica addition/removal in clustermap is currently supported by dynamic cluster change handler only.( #1355 )

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
@@ -153,6 +153,11 @@ class StaticClusterManager implements ClusterMap {
     return metricRegistry;
   }
 
+  @Override
+  public void registerClusterMapListener(ClusterMapChangeListener clusterMapChangeListener) {
+    // no op for static cluster map.
+  }
+
   // Administrative API
   // -----------------------
 
@@ -535,8 +540,7 @@ class StaticClusterManager implements ClusterMap {
 
   @Override
   public ReplicaId getBootstrapReplica(String partitionIdStr, DataNodeId dataNodeId) {
-    throw new UnsupportedOperationException(
-        "Adding new replica is currently not supported in static cluster manager.");
+    throw new UnsupportedOperationException("Adding new replica is currently not supported in static cluster manager.");
   }
 
   @Override
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
@@ -155,7 +155,7 @@ class StaticClusterManager implements ClusterMap {
 
   @Override
   public void registerClusterMapListener(ClusterMapChangeListener clusterMapChangeListener) {
-    // no op for static cluster map.
+    throw new UnsupportedOperationException("Registering clustermap listener is not supported in static clustermap");
   }
 
   // Administrative API
--- a/ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
+++ b/ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
@@ -82,9 +82,9 @@ public class PartitionInfo {
   }
 
   /**
-   * Remove {@link RemoteReplicaInfo} from this {@link PartitionInfo} if it is present.
-   * @param remoteReplica the {@link RemoteReplicaInfo} to remove.
-   * @return {@code true} if given remote replica info previously existed and is successfully removed. {@code false}, otherwise.
+   * Remove {@link RemoteReplicaInfo} of given replica from this {@link PartitionInfo} if it is present.
+   * @param remoteReplica the {@link ReplicaId} whose info should be removed.
+   * @return {@link RemoteReplicaInfo} that is removed, can be null if it is not present.
    */
   RemoteReplicaInfo removeReplicaInfoIfPresent(ReplicaId remoteReplica) {
     rwLock.writeLock().lock();
--- a/ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
+++ b/ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
@@ -53,6 +53,11 @@ public class PartitionInfo {
     return this.localReplicaId;
   }
 
+  /**
+   * Add {@link RemoteReplicaInfo} to this {@link PartitionInfo} if it is previously absent.
+   * @param remoteReplicaInfo the {@link RemoteReplicaInfo} to add.
+   * @return {@code true} if remote replica info is added. {@code false} if it is already present
+   */
   boolean addReplicaInfoIfAbsent(RemoteReplicaInfo remoteReplicaInfo) {
     lock.lock();
     boolean isAdded = false;
@@ -70,6 +75,11 @@ public class PartitionInfo {
     return isAdded;
   }
 
+  /**
+   * Remove {@link RemoteReplicaInfo} from this {@link PartitionInfo} if it is present.
+   * @param remoteReplicaInfo the {@link RemoteReplicaInfo} to remove.
+   * @return {@code true} if given remote replica info previously existed and is successfully removed. {@code false}, otherwise.
+   */
   boolean removeRelicaInfoIfPresent(RemoteReplicaInfo remoteReplicaInfo) {
     lock.lock();
     boolean isRemoved;
--- a/ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
+++ b/ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
@@ -17,7 +17,8 @@ import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.store.Store;
 import java.util.List;
-import java.util.concurrent.locks.ReentrantLock;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import java.util.stream.Collectors;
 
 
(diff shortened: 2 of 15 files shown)


ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -115,8 +127,89 @@ public void start() throws ReplicationException {
...
+    }
+  }
+
+  /**
+   * {@inheritDoc}
+   * Note that, this method should be thread-safe because multiple threads (from different cluster change handlers) may
+   * concurrently update remote replica infos.
+   */
+  @Override
+  public void onReplicaAddedOrRemoved(List<ReplicaId> addedReplicas, List<ReplicaId> removedReplicas) {
```
ankagrawal: Looks like this method will be called either for replicas being added or removed. We should use a boolean then (e.g, isRemove), instead of two lists.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -115,8 +127,89 @@ public void start() throws ReplicationException {
...
+    // 1. wait for start() to complete
+    try {
```
ankagrawal: Also it looks like we have some duplicate code between this method and addReplica/removeReplica methods. Can we try to refactor?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
```diff
@@ -153,6 +153,11 @@ public MetricRegistry getMetricRegistry() {
     return metricRegistry;
   }
 
+  @Override
+  public void registerClusterMapListener(ClusterMapChangeListener clusterMapChangeListener) {
+    // no op for static cluster map.
```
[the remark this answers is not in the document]
jsjtzyy: sure, will make the change

ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
```diff
@@ -50,6 +53,44 @@ public ReplicaId getLocalReplicaId() {
     return this.localReplicaId;
   }
 
+  /**
+   * Add {@link RemoteReplicaInfo} to this {@link PartitionInfo} if it is previously absent.
+   * @param remoteReplicaInfo the {@link RemoteReplicaInfo} to add.
+   * @return {@code true} if remote replica info is added. {@code false} if it is already present
+   */
+  boolean addReplicaInfoIfAbsent(RemoteReplicaInfo remoteReplicaInfo) {
+    lock.lock();
```
justinlin-linkedin: please lock getRemoveReplicaInfos method as well, this lock is protecting remoteReplicas list, we should lock it everywhere we use it.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -115,8 +127,89 @@ public void start() throws ReplicationException {
...
+              new RemoteReplicaInfo(remoteReplica, partitionInfo.getLocalReplicaId(), partitionInfo.getStore(),
+                  findToken,
+                  TimeUnit.SECONDS.toMillis(storeConfig.storeDataFlushIntervalSeconds) * Replication_Delay_Multiplier,
+                  SystemTime.getInstance(), remoteReplica.getDataNodeId().getPortToConnectTo());
+          logger.info("Adding remote replica {} on {} to partition info.", remoteReplica.getReplicaPath(),
+              remoteReplica.getDataNodeId());
+          if (partitionInfo.addReplicaInfoIfAbsent(remoteReplicaInfo)) {
+            replicationMetrics.addMetricsForRemoteReplicaInfo(remoteReplicaInfo);
+          }
+          replicaInfosToAdd.add(remoteReplicaInfo);
```
justinlin-linkedin: we should only add remoteReplicaInfo to the "toAdd" list when it was successfully added to partitionInfo, so we should move this statement in the if-else statement above.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -115,8 +127,89 @@ public void start() throws ReplicationException {
...
+              logger.info("Removing remote replica {} on {} from replica threads.", remoteReplica.getReplicaPath(),
+                  remoteReplica.getDataNodeId());
+              replicaInfosToRemove.add(remoteReplicaInfo);
+              if (partitionInfo.removeRelicaInfoIfPresent(remoteReplicaInfo)) {
+                replicationMetrics.removeMetricsForRemoteReplicaInfo(remoteReplicaInfo);
+              }
+              break;
+            }
+          }
+        }
```
justinlin-linkedin: why do we need this? we can just change the PartitionInfo's remove method to take a replica id as parameter, and looping the list in PartitionInfo, it's more aligned with addReplicaInfo.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -115,8 +127,89 @@ public void start() throws ReplicationException {
...
+        // 4. remove replicaInfo from existing partitionInfo and replica-threads
+        List<RemoteReplicaInfo> replicaInfosToRemove = new ArrayList<>();
+        for (ReplicaId remoteReplica : removedPeerReplicas) {
+          PartitionInfo partitionInfo = partitionToPartitionInfo.get(remoteReplica.getPartitionId());
+          List<RemoteReplicaInfo> remoteReplicaInfos = new ArrayList<>(partitionInfo.getRemoteReplicaInfos());
+          for (RemoteReplicaInfo remoteReplicaInfo : remoteReplicaInfos) {
+            if (remoteReplicaInfo.getReplicaId().getDataNodeId() == remoteReplica.getDataNodeId()) {
+              logger.info("Removing remote replica {} on {} from replica threads.", remoteReplica.getReplicaPath(),
+                  remoteReplica.getDataNodeId());
+              replicaInfosToRemove.add(remoteReplicaInfo);
```
justinlin-linkedin: And we should put this statement within if-else statement as well.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
```diff
@@ -115,8 +127,89 @@ public void start() throws ReplicationException {
...
+   * Note that, this method should be thread-safe because multiple threads (from different cluster change handlers) may
+   * concurrently update remote replica infos.
+   */
+  @Override
+  public void onReplicaAddedOrRemoved(List<ReplicaId> addedReplicas, List<ReplicaId> removedReplicas) {
+    // 1. wait for start() to complete
+    try {
+      startupLatch.await();
+    } catch (InterruptedException e) {
+      logger.warn("Waiting for startup is interrupted.");
```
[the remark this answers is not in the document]
jsjtzyy: Will throw an IllegalStateException here. Note that, it doesn't help catch up missed event because next time this method is invoked, the pass-in parameters have changed. (For replication manager this is not critical, we still allow whole server to serve traffic temporarily. Missed replica event eventually will generate some error logs and then we can restart this node)

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: `addReplica()`, `removeReplica()` are methods for changing replicas on local node. They create a new PartitionInfo along with all peer RemoteReplicaInfos. `onReplicaAddedOrRemoved()` basically modifies existing PartitionInfo by adding/removing certain RemoteReplicaInfo. Refactoring will be considered in a separate PR after we have verified functionality of this change.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: I see your point but the method is invoked by InstanceConfig change and there could be both added and removed replica within a single InstanceConfig change. We pass in two lists (can be empty) and let replication manager handle both cases at the same time.

ambry-replication/src/main/java/com.github.ambry.replication/PartitionInfo.java
jsjtzyy: make sense, i will use read-write lock instead

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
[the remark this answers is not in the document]
jsjtzyy: Any benefits we can get by doing so? One probably reason I come up with is: when unifying all existing replication managers into a general replication manager, it requires 2 clustermap change listeners taking actions against changes in VCR and regular Ambry cluster. Feel free to add more thoughts here.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: Correct, will fix this.

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: will fix

ambry-replication/src/main/java/com.github.ambry.replication/ReplicationManager.java
jsjtzyy: right, will change `removeRelicaInfoIfPresent `

ambry-clustermap/src/main/java/com.github.ambry.clustermap/StaticClusterManager.java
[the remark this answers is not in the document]
ankagrawal: I think this isnt addressed yet.

PR #1370, opened 2020-01-28
Add undelete request and response

justinlin-linkedin:
Adding undelete request and response and prepare RequestAPI to handle undelete request.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
+++ b/ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
@@ -80,4 +80,14 @@ public interface RequestAPI {
   default void handleAdminRequest(NetworkRequest request) throws InterruptedException, IOException {
     throw new UnsupportedOperationException("Admin request not supported on this node");
   }
+
+  /**
+   * Undelete the blob from the store.
+   * @param request the request that contains the partitioni and the id of the blob that needs to be undeleted.
+   * @throws IOException if there are I/O errors carrying our the required operation.
+   * @throws InterruptedException if request processing is interrupted.
+   */
+  default void handleUndeleteRequest(NetworkRequest request) throws InterruptedException, IOException {
+    throw new UnsupportedOperationException("Undelete request not supported on this node");
+  }
 }
--- a/ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
+++ b/ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
@@ -83,7 +83,7 @@ public interface RequestAPI {
 
   /**
    * Undelete the blob from the store.
-   * @param request the request that contains the partitioni and the id of the blob that needs to be undeleted.
+   * @param request the request that contains the partition and the id of the blob that needs to be undeleted.
    * @throws IOException if there are I/O errors carrying our the required operation.
    * @throws InterruptedException if request processing is interrupted.
    */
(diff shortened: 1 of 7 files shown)


ambry-server/src/main/java/com.github.ambry.server/AmbryServerRequests.java
```diff
@@ -100,39 +100,6 @@
     localPartitionToReplicaMap = createLocalPartitionToReplicaMap();
   }
 
-  @Override
-  public void handleRequests(NetworkRequest request) throws InterruptedException {
```
zzmao: Why this method removed?

justinlin-linkedin: AmbryServerRequests is a direct child of AmbryRequests and this method is trying to override the same method from AmbryRequests, but they are exactly the same method. So we can remove them.

ambry-protocol/src/main/java/com.github.ambry.protocol/UndeleteResponse.java
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Constructs a {@link UndeleteResponse} with an {@link ServerErrorCode}.
+   * @param correlationId correlationId of the undelete response.
+   * @param clientId clientId of the undelete response.
+   * @param error error code returned in this undelete response.
+   */
+  public UndeleteResponse(int correlationId, String clientId, ServerErrorCode error) {
+    super(RequestOrResponseType.UndeleteResponse, UNDELETE_RESPONSE_VERSION_1, correlationId, clientId, error);
+    if (error == ServerErrorCode.No_Error) {
+      throw new IllegalArgumentException("NoError is not a valid error code");
```
jsjtzyy: Can you explain a little more about this exception?

ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
```diff
@@ -80,4 +80,14 @@
...
     throw new UnsupportedOperationException("Admin request not supported on this node");
   }
+
+  /**
+   * Undelete the blob from the store.
+   * @param request the request that contains the partitioni and the id of the blob that needs to be undeleted.
+   * @throws IOException if there are I/O errors carrying our the required operation.
+   * @throws InterruptedException if request processing is interrupted.
+   */
+  default void handleUndeleteRequest(NetworkRequest request) throws InterruptedException, IOException {
```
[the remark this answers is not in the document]
justinlin-linkedin: Interface's abstract method can't have it's owner body unless it's a default method.

ambry-protocol/src/main/java/com.github.ambry.protocol/UndeleteResponse.java
justinlin-linkedin: there are two possible responses for undelete request: 1 Normal response with a lifeVersion and NO_ERROR code, 2 error response with a invalid lifeVersion number and any error code rather than NO_ERROR. That's why I am throwing out an exception here.

zzmao:
test failure?

ambry-api/src/main/java/com.github.ambry/protocol/RequestAPI.java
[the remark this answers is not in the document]
jsjtzyy: I see.

PR #1373, opened 2020-02-05
Add undelete method to BlobStore

justinlin-linkedin:
Adding undelete method to blobstore and other Store implementation.
This PR will have no change w.r.t AmbryRequests.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
+++ b/ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
@@ -58,6 +58,15 @@ public interface NotificationSystem extends Closeable {
   void onBlobDeleted(String blobId, String serviceId, Account account, Container container);
 
   /**
+   * Notifies the underlying system when a deleted blob is undeleted.
+   * @param blobId The id of the blob whose deleted state has been replicated
+   * @param serviceId The service ID of the service deleting the blob. This can be null if unknown.
+   * @param account The {@link Account} for the blob
+   * @param container The {@link Container} for the blob
+   */
+  void onBlobUndeleted(String blobId, String serviceId, Account account, Container container);
+
+  /**
    * Notifies the underlying system when a blob is replicated to a node
    * @param sourceHost The source host from where the notification is being invoked
    * @param port The port of the source host from where the notification is being invoked.
@@ -86,4 +95,14 @@ public interface NotificationSystem extends Closeable {
    */
   void onBlobReplicaUpdated(String sourceHost, int port, String blobId, BlobReplicaSourceType sourceType,
       UpdateType updateType, MessageInfo info);
+
+  /**
+   * Notifies the underlying system when a undeleted state of a blob is replicated to a node
+   * @param sourceHost The source host from where the notification is being invoked
+   * @param port The port of the source host from where the notification is being invoked.
+   * @param blobId The id of the blob whose deleted state has been replicated
+   * @param sourceType The source that undeleted the blob replica
+   */
+  void onBlobReplicaUndeleted(String sourceHost, int port, String blobId, BlobReplicaSourceType sourceType);
+
 }
--- a/ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
+++ b/ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
@@ -104,5 +104,4 @@ public interface NotificationSystem extends Closeable {
    * @param sourceType The source that undeleted the blob replica
    */
   void onBlobReplicaUndeleted(String sourceHost, int port, String blobId, BlobReplicaSourceType sourceType);
-
 }
--- a/ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
+++ b/ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
@@ -58,13 +58,14 @@ public interface NotificationSystem extends Closeable {
   void onBlobDeleted(String blobId, String serviceId, Account account, Container container);
 
   /**
-   * Notifies the underlying system when a deleted blob is undeleted.
-   * @param blobId The id of the blob whose deleted state has been replicated
-   * @param serviceId The service ID of the service deleting the blob. This can be null if unknown.
+   * Notifies the underlying system when the blob is undeleted.
+   * @param blobId The id of the blob whose undeleted state has been replicated
+   * @param serviceId The service ID of the service undeleting the blob. This can be null if unknown.
    * @param account The {@link Account} for the blob
    * @param container The {@link Container} for the blob
    */
-  void onBlobUndeleted(String blobId, String serviceId, Account account, Container container);
+  default void onBlobUndeleted(String blobId, String serviceId, Account account, Container container) {
+  }
 
   /**
    * Notifies the underlying system when a blob is replicated to a node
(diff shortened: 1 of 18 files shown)


ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
```diff
@@ -57,6 +57,15 @@ void onBlobCreated(String blobId, BlobProperties blobProperties, Account account
    */
   void onBlobDeleted(String blobId, String serviceId, Account account, Container container);
 
+  /**
+   * Notifies the underlying system when a deleted blob is undeleted.
+   * @param blobId The id of the blob whose deleted state has been replicated
+   * @param serviceId The service ID of the service deleting the blob. This can be null if unknown.
```
jsjtzyy: minor: `undeleting`

ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
```diff
@@ -57,6 +57,15 @@ void onBlobCreated(String blobId, BlobProperties blobProperties, Account account
    */
   void onBlobDeleted(String blobId, String serviceId, Account account, Container container);
 
+  /**
+   * Notifies the underlying system when a deleted blob is undeleted.
+   * @param blobId The id of the blob whose deleted state has been replicated
```
jsjtzyy: `the blob was undeleted`

ambry-api/src/main/java/com.github.ambry/notification/NotificationSystem.java
```diff
@@ -86,4 +95,13 @@ void onBlobCreated(String blobId, BlobProperties blobProperties, Account account
    */
   void onBlobReplicaUpdated(String sourceHost, int port, String blobId, BlobReplicaSourceType sourceType,
       UpdateType updateType, MessageInfo info);
+
+  /**
+   * Notifies the underlying system when a undeleted state of a blob is replicated to a node
+   * @param sourceHost The source host from where the notification is being invoked
+   * @param port The port of the source host from where the notification is being invoked.
+   * @param blobId The id of the blob whose deleted state has been replicated
```
jsjtzyy: `undeleted state`

ambry-api/src/main/java/com.github.ambry/store/Store.java
```diff
@@ -54,6 +54,13 @@
    */
   void delete(MessageWriteSet messageSetToDelete) throws StoreException;
 
+  /**
+   * Undelete the blob identified by {@code id}.
+   * @param info The {@link MessageInfo} that carries some basic information about this operation.
+   * @return the lifeVersion of the undeleted blob.
+   */
+  short undelete(MessageInfo info) throws StoreException;
```
jsjtzyy: Any reason why we don't pass in `MessageWriteSet`? Is this because we undelete only one blob at a time?

ambry-commons/src/main/java/com.github.ambry.commons/LoggingNotificationSystem.java
```diff
@@ -64,22 +64,36 @@ public void onBlobDeleted(String blobId, String serviceId, Account account, Cont
...
+
   @Override
   public void onBlobReplicaCreated(String sourceHost, int port, String blobId, BlobReplicaSourceType sourceType) {
     logger.debug("onBlobReplicaCreated " + sourceHost + ", " + port + ", " + blobId + ", " + sourceType);
   }
 
   @Override
   public void onBlobReplicaDeleted(String sourceHost, int port, String blobId, BlobReplicaSourceType sourceType) {
-    logger.debug("onBlobReplicaCreated " + sourceHost + ", " + port + ", " + blobId + ", " + sourceType);
+    logger.debug("onBlobReplicaDeleted " + sourceHost + ", " + port + ", " + blobId + ", " + sourceType);
```
jsjtzyy: good catch

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -612,6 +615,70 @@ public void updateTtl(MessageWriteSet messageSetToUpdate) throws StoreException
...
 
+  @Override
+  public short undelete(MessageInfo info) throws StoreException {
+    checkStarted();
+    final Timer.Context context = metrics.undeleteResponse.time();
+    try {
+      StoreKey id = info.getStoreKey();
+      Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
+      List<IndexValue> values = index.findAllIndexValuesForKey(id, null);
+      index.validateSanityForUndelete(id, values, IndexValue.LIFE_VERSION_FROM_FRONTEND);
```
jsjtzyy: Note that, other methods (put, ttlUpdate, delete) in BlobStore are also invoked in Replication. When you implement undelete replication part, the `LIFE_VERSION_FROM_FRONTEND` check here probably needs to change.

ambry-api/src/main/java/com.github.ambry/store/Store.java
justinlin-linkedin: MessgeWriteSet includes a serialized undelete record, which will be persisted on disk in the log. The problem is that this serialized record requires a lifeVersion, but we don't know the lifeVersion before we query the index. So here I pass a MessageInfo instead of MessageWriteSet. You can see in the BlobStore.undelete method, I create a MessageWriteSet in the end, with a lifeVersion.

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
justinlin-linkedin: I actually have to change ttlupdate and delete later to make them aware of undelete in replication.

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -612,6 +615,70 @@ public void updateTtl(MessageWriteSet messageSetToUpdate) throws StoreException
...
+          if (value != null) {
+            throw new StoreException("Cannot undelete id " + info.getStoreKey() + " since concurrent operation occurs",
+                StoreErrorCodes.Life_Version_Conflict);
+          }
+        }
+        Offset endOffsetOfLastMessage = log.getEndOffset();
+        writeSet.writeTo(log);
+        logger.trace("Store : {} undelete mark written to log", dataDir);
+        FileSpan fileSpan = log.getFileSpanForMessage(endOffsetOfLastMessage, info.getSize());
+        index.markAsUndeleted(info.getStoreKey(), fileSpan, info.getOperationTimeMs());
```
jsjtzyy: Add a TODO here: update `blobStoreStats` for Undelete.

ambry-store/src/main/java/com.github.ambry.store/BlobStore.java
```diff
@@ -612,6 +615,70 @@ public void updateTtl(MessageWriteSet messageSetToUpdate) throws StoreException
...
+  @Override
+  public short undelete(MessageInfo info) throws StoreException {
+    checkStarted();
+    final Timer.Context context = metrics.undeleteResponse.time();
+    try {
+      StoreKey id = info.getStoreKey();
+      Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
+      List<IndexValue> values = index.findAllIndexValuesForKey(id, null);
+      index.validateSanityForUndelete(id, values, IndexValue.LIFE_VERSION_FROM_FRONTEND);
+      IndexValue latestValue = values.get(0);
```
zzmao: It's clear when oldest/latest is used in variable. I would like to suggest also use var oldest/latest in validateSanityForUndelete, instead of first/last. To be consistent.

PR #1380, opened 2020-02-10
Add precondition to metadata updates to fail on conflicts

lightningrob:
When we update blob metadata in ABS and Cosmos, we have to first pull the whole metadata record (a property map in ABS and a document in Cosmos), update the field of interest, and upload the modified record.  (There is no single-field update operation.)  In live-serving mode, both the ambry-frontend and ambry-vcr may try to update the same blob metadata concurrently.  Hence, it is possible for updates to get clobbered due to race conditions.  To prevent this, we make use of the resource etag, along with the If-Match request condition, to fail conflicting updates and force them to retry.
 
Note: the mechanism used to test conflicts is non-standard as it relies on a callback in the production code for the test to inject a conflicting modification.  I'm open to suggestions for cleaner alternatives.


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -295,7 +301,13 @@ public boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value)
         String textValue = String.valueOf(value);
         if (!textValue.equals(metadata.get(fieldName))) {
           metadata.put(fieldName, textValue);
-          // Set condition to ensure we don't clobber another update
+          if (updateCallback != null) {
+            try {
+              updateCallback.call();
+            } catch (Exception ex) {
```
cgtz: log exception?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -120,6 +121,11 @@
     this.purgeBatchSize = AzureCloudConfig.DEFAULT_PURGE_BATCH_SIZE;
   }
 
+  /** Visible for testing */
+  void setUpdateCallback(Callable callback) {
```
cgtz: use generics in both the member variable and method arg. If the return type of call doesn't matter, you can use `Callable<?>` in both places.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
```diff
@@ -58,6 +61,11 @@
     this.azureMetrics = azureMetrics;
   }
 
+  /** Visible for testing */
+  void setUpdateCallback(Callable callback) {
+    this.updateCallback = callback;
```
cgtz: same comments for CosmosDataAccessor as AzureBlobDataAccessor.

PR #1382, opened 2020-02-12
Minor changes to support recovering from replica addition failure

jsjtzyy:
This PR adds minor changes to resume dynamic replica addition if it failed last time before updating InstanceConfig(clustermap update in Helix). The logic is simple: if diskManager finds there is a existing
directory associated with replica to add, it first deletes it and recreates a new one. (The files in previous directory may not be reliable)

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
@@ -24,6 +24,7 @@ import com.github.ambry.utils.Throttler;
 import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.io.File;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


ambry-store/src/main/java/com.github.ambry.store/DiskManager.java
```diff
@@ -311,11 +312,23 @@ boolean addBlobStore(ReplicaId replica) {
       if (!running) {
         logger.error("Failed to add {} because disk manager is not running", replica.getPartitionId());
       } else {
+        // Clean up existing dir associated with this replica to add. Here we re-create a new store because we don't
+        // know the state of files in old directory. (The old directory was created last time when adding this replica
```
ankagrawal: Just validating my though?
Is the assumption here that, if there is some old state of a replica on a node's storage, and storage manager or disk manager don't know about it, then it could be a corrupted state, and hence remove it.

The existing state might help in case where let's say a node crashed and comes back up after sometime, and is assigned same replica back again. In this case, even though there is some state on disk, we will delete it. The state on disk might have helped this replica in catching up with existing replicas quicker. But I assuming that we are ignoring this scenario because we don't have a way to identify this scenario. Is this correct?

jsjtzyy: You are right. `The state on disk might have helped this replica in catching up with existing replicas quicker` this is true when the added replica was previously in BOOTSTRAP -> STANDBY and was in the middle of caught-up. Note that, if this added replica was ready in BOOTSTRAP -> STANDBY before node crashed, that means it should have already updated InstanceConfig in Helix (this is the last step in OFFLINE -> BOOTSTRAP). So, after node reboots, storage manager/disk manager should be aware of this replica and don't attempt to add it again. 
The failure I want to address in this PR occurs before new added replica updates InstanceConfig, so the new replica may be created but still empty. We don't have any loss by deleting its dir when node is restarted. (we can discuss it offline if this didn't answer your question)

PR #1384, opened 2020-02-13
Add 3 basic classes for HTTP2 based NetworkClient implementation.

zzmao:
Implemented Http2ChannelPoolMap, MultiplexedChannelRecord and Http2MultiplexedChannelPool.
Moved HTTP/2 related classes to com.github.ambry.network.http2
Moved SSLFactory implementations to com.github.ambry.commons


ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelInitializer.java
```diff
@@ -0,0 +1,63 @@
...
+import io.netty.handler.codec.http2.Http2FrameCodecBuilder;
+import io.netty.handler.codec.http2.Http2MultiplexHandler;
+import io.netty.handler.codec.http2.Http2Settings;
+import io.netty.handler.ssl.SslHandler;
+
+
+/**
+ * A {@link ChannelInitializer} to be used with {@link Http2ChannelPoolMap}.
+ */
+public class Http2ChannelInitializer extends AbstractChannelPoolHandler {
```
cgtz: Could you call this `Http2ClientChannelInitializer` and replace the one used for `Http2BlockingChannel` with this one?

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelInitializer.java
```diff
@@ -0,0 +1,63 @@
...
+  public Http2ChannelInitializer(SSLConfig sslConfig, RequestInfo requestInfo) {
+    this.sslConfig = sslConfig;
+    this.host = requestInfo.getHost();
+    this.port = requestInfo.getPort().getPort();
+  }
+
+  @Override
+  public void channelCreated(Channel ch) throws Exception {
+    ChannelPipeline pipeline = ch.pipeline();
+    SSLFactory sslFactory = new NettySslHttp2Factory(sslConfig);
```
cgtz: You should only need one shared sslFactory for all channels. I would prefer only parsing the configs once

ambry-rest/src/main/java/com.github.ambry.rest/Http2ClientChannelInitializer.java
```diff
@@ -26,8 +26,7 @@
 
 
 /**
- * A {@link ChannelInitializer} to be used with {@link Http2BlockingChannel}. Calling {@link #initChannel(SocketChannel)}
- * adds the necessary handlers to a channel's pipeline so that it may handle requests.
+ * A {@link ChannelInitializer} to be used with {@link Http2ChannelPoolMap}.
```
cgtz: looks like this is not used with `Http2ChannelPoolMap`

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelPoolMap.java
```diff
@@ -0,0 +1,46 @@
...
+import io.netty.channel.pool.AbstractChannelPoolMap;
+import io.netty.channel.pool.ChannelPool;
+import io.netty.channel.socket.nio.NioSocketChannel;
+
+
+public class Http2ChannelPoolMap extends AbstractChannelPoolMap<RequestInfo, ChannelPool> {
+  private final EventLoopGroup eventLoopGroup;
+  private final SSLConfig sslConfig;
+
+  public Http2ChannelPoolMap(SSLConfig sslConfig, EventLoopGroup eventLoopGroup) {
```
cgtz: Maybe construct `SSLFactory` outside (i.e. in the `NetworkClientFactory` impl) and pass it into this constructor since it should be able to shared across all instances.

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelPoolMap.java
```diff
@@ -0,0 +1,46 @@
...
+import com.github.ambry.network.RequestInfo;
+import io.netty.bootstrap.Bootstrap;
+import io.netty.channel.ChannelOption;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.pool.AbstractChannelPoolMap;
+import io.netty.channel.pool.ChannelPool;
+import io.netty.channel.socket.nio.NioSocketChannel;
+
+
+public class Http2ChannelPoolMap extends AbstractChannelPoolMap<RequestInfo, ChannelPool> {
```
cgtz: It seems that RequestInfo isn't exactly the right key to use for `Http2ChannelPoolMap` and `Http2ChannelInitializer` for a couple reasons:
1. It doesn't have hashcode and equals defined, and a separate object is created for each request, so a new pool will created for each request, instead of each host:port combo
2. It includes info about the request body, which is not relevant to the pool (i think).

Instead, you could use `InetSocketAddress` and pass this directly into `Bootstrap::remoteAddress`. The impl of `Bootstrap::remoteAddress(String inetHost, int inetPort)` would call `InetSocketAddress.createUnresolved(inetHost, inetPort)` anyways.

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelInitializer.java
zzmao: Http2ClientChannelInitializer is `Channelnitializer`, which is not a `AbstractChannelPoolHandler`.
We probably need to keep two classes. 
I can rename this(Http2ChannelInitializer) to `Http2ChannelPoolHanlder` and try to avoid dup code.

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelInitializer.java
zzmao: ok

ambry-rest/src/main/java/com.github.ambry.rest/Http2ClientChannelInitializer.java
zzmao: fixed. For `Http2BlockingChannel`

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelInitializer.java
zzmao: Http2ClientChannelInitializer class is removed. Do inline initialize in Http2BlockingChannel::channel().

ambry-rest/src/main/java/com.github.ambry.rest/Http2ClientChannelInitializer.java
zzmao: This class is removed.

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelPoolMap.java
zzmao: SSLFactory is used.

ambry-rest/src/main/java/com.github.ambry.rest/Http2ChannelPoolMap.java
zzmao: InetSocketAddress is used.

zzmao:
Moved http2 realted classes to `com.github.ambry.network.http2`.

This also requires to move `NettyHttp2SslFactory` to `com.github.ambry.commons`. Http2 needs `NettyHttp2SslFactory`.

So I moved both `NettyHttp2SslFactory` and `NettySslFactory` to `com.github.ambry.commons` and ambry-commons now also depends on netty-all. (if you have any suggestion, let me know.)

ambry-server/src/integration-test/java/com.github.ambry.server/ServerTestUtil.java
```diff
@@ -122,6 +124,7 @@
 import java.util.concurrent.atomic.AtomicReference;
 import javax.net.ssl.SSLSocketFactory;
 import org.junit.Assert;
+import org.junit.Test;
```
cgtz: Is this import needed?

zzmao:
Addressed @cgtz 's comments.

zzmao:
@cgtz  @justinlin-linkedin   The build finally passed. Ready to merge. Thanks!

PR #1386, opened 2020-02-14
Change Store interface to not use MessageWriteSet

justinlin-linkedin:
delete and updateTtl method in interface Store take MessageWriteSet as its only parameter. This is not ideal since the data from MessageWriteSet would be persisted in the LogSegment files on disk, which means there is a chance that outside of the BlobStore implementation, we can construct any arbitrary bytes and persist them in the log files. This is not very bad, for we control how to construct those bytes, but from interface's perspective, it's not a good one.

And in order to support persisting undelete record, we need to get the lifeVersion from the PersistentIndex before we can construct a UndeleteFormatInputStream. The same goes to delete and ttlUpdate record. But in AmbryRequests, we don't yet know the lifeVersion before creating FormatInputStream for those record because of Store interface requires MessageWriteSet as the parameter. Changing the interface to take MessageInfo not MessageWriteSet solves this problem.

This PR changes that to make sure BlobStore would construct all the delete and updateTtl records by itself. We should change put method as well, but later.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
@@ -129,7 +129,7 @@ class CloudBlobStore implements Store {
   @Override
   public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> storeGetOptions) throws StoreException {
     checkStarted();
-    checkDuplicates(ids);
+    checkStoreKeyDuplicates(ids);
     List<CloudMessageReadSet.BlobReadInfo> blobReadInfos = new ArrayList<>(ids.size());
     List<MessageInfo> messageInfos = new ArrayList<>(ids.size());
     try {
@@ -253,7 +253,7 @@ class CloudBlobStore implements Store {
     if (messageSetToWrite.getMessageSetInfo().isEmpty()) {
       throw new IllegalArgumentException("Message write set cannot be empty");
     }
-    checkDuplicates(messageSetToWrite);
+    checkDuplicates(messageSetToWrite.getMessageSetInfo());
 
     // Write the blobs in the message set
     CloudWriteChannel cloudWriter = new CloudWriteChannel(this, messageSetToWrite.getMessageSetInfo());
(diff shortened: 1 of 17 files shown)


ambry-protocol/src/main/java/com.github.ambry.protocol/AmbryRequests.java
```diff
@@ -398,16 +396,11 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
-        MessageFormatInputStream stream =
-            new DeleteMessageFormatInputStream(convertedStoreKey, convertedBlobId.getAccountId(),
-                convertedBlobId.getContainerId(), deleteRequest.getDeletionTimeInMs());
-        MessageInfo info = new MessageInfo(convertedStoreKey, stream.getSize(), convertedBlobId.getAccountId(),
-            convertedBlobId.getContainerId(), deleteRequest.getDeletionTimeInMs());
-        ArrayList<MessageInfo> infoList = new ArrayList<MessageInfo>();
-        infoList.add(info);
-        MessageFormatWriteSet writeSet = new MessageFormatWriteSet(stream, infoList, false);
+        MessageInfo info =
+            new MessageInfo(convertedStoreKey, 0, convertedBlobId.getAccountId(), convertedBlobId.getContainerId(),
```
zzmao: -1 might be better, to indicate invalid or meaningless.

ambry-messageformat/src/main/java/com.github.ambry.messageformat/DeleteMessageFormatInputStream.java
```diff
@@ -55,4 +55,22 @@ public DeleteMessageFormatInputStream(StoreKey key, short accountId, short conta
     messageLength = buffer.capacity();
     buffer.flip();
   }
+
+  public DeleteMessageFormatInputStream(StoreKey key, short accountId, short containerId, long deletionTimeMs,
```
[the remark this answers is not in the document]
justinlin-linkedin: The reason I create a different constructor is because of dependency issue.
The new constructor requires lifeVersion to be not -1, which is a constant from IndexValue.Life_Version_From_Frontend. And I don't want to create a new constant here and I can't import the IndexValue constant either. That's why I create a new constructor and leave if-else statement in the ambry-store package, where IndexValue constant can be used.

ambry-protocol/src/main/java/com.github.ambry.protocol/AmbryRequests.java
justinlin-linkedin: sure

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -367,18 +367,12 @@ public short undelete(MessageInfo info) throws StoreException {
     throw new UnsupportedOperationException("Undelete not supported in cloud store");
   }
 
-  /**
-   * {@inheritDoc}
-   * Currently, the only supported operation is to set the TTL to infinite (i.e. no arbitrary increase or decrease)
-   * @param messageSetToUpdate The list of messages that need to be updated
-   * @throws StoreException if there is a problem persisting the operation in the store.
-   */
```
lightningrob: Not sure why you removed the javadoc.  Please restore.

lightningrob:
The change looks fine.  I suggest adding a bit more context on why we are doing this change now.  Is it needed for undelete feature?

justinlin-linkedin:
> The change looks fine. I suggest adding a bit more context on why we are doing this change now. Is it needed for undelete feature?

I added more details in this PR's description. Part of the description does mention the support for undelete.

PR #1389, opened 2020-02-18
Support ByteBuf writing in NettyResponseChannel

justinlin-linkedin:
Start to support writing ByteBuf in the NettyResponseChannel.

This is another step to stop supporting writing java ByteBuffer in the AsyncWritableChannel.

### DIFF
--- a/ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
+++ b/ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
@@ -86,7 +86,7 @@ class NettyResponseChannel implements RestResponseChannel {
   private final ChunkedWriteHandler chunkedWriteHandler;
   private final PerformanceConfig perfConfig;
 
-  private final static Logger logger = LoggerFactory.getLogger(NettyResponseChannel.class);
+  private static final Logger logger = LoggerFactory.getLogger(NettyResponseChannel.class);
   private final HttpResponse responseMetadata = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);
   // tracks whether onResponseComplete() has been called. Helps make it idempotent and also treats response channel as
   // closed if this is true.
@@ -127,19 +127,19 @@ class NettyResponseChannel implements RestResponseChannel {
 
   @Override
   public Future<Long> write(ByteBuffer src, Callback<Long> callback) {
-    return write(Unpooled.wrappedBuffer(src), new Callback<Long>() {
+    FutureResult<Long> futureResult = new FutureResult<>();
+    write(Unpooled.wrappedBuffer(src), new Callback<Long>() {
       @Override
       public void onCompletion(Long result, Exception exception) {
-        if (exception == null) {
-          src.position(src.limit());
-        } else {
-          src.position(result.intValue());
-        }
+        long r = result == null ? (long) 0 : result.longValue();
+        src.position((int) r);
+        futureResult.done(r, exception);
         if (callback != null) {
           callback.onCompletion(result, exception);
         }
       }
     });
+    return futureResult;
   }
 
   @Override
@@ -460,7 +460,7 @@ class NettyResponseChannel implements RestResponseChannel {
 
   /**
    * Builds and sends an error response to the client based on {@code cause}.
-   *j@param exception the cause of the request handling failure.
+   * @param exception the cause of the request handling failure.
    * @return {@code true} if error response was scheduled to be sent. {@code false} otherwise.
    */
   private boolean maybeSendErrorResponse(Exception exception) {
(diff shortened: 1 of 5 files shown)


ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -443,7 +443,7 @@ private boolean maybeWriteResponseMetadata(HttpResponse responseMetadata,
 
   /**
    * Builds and sends an error response to the client based on {@code cause}.
-   * @param exception the cause of the request handling failure.
+   *j@param exception the cause of the request handling failure.
```
xuhao417347761: Maybe a typo here for j added before @param.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -127,6 +127,23 @@
 
   @Override
   public Future<Long> write(ByteBuffer src, Callback<Long> callback) {
```
cgtz: i would recommend doing the thing that was done with in the default impl in AsyncWritableChannel where a second future is created and completed from the callback. This will ensure that the future returned is completed after the buffer position is changed.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -877,14 +893,13 @@ public HttpContent readChunk(ByteBufAllocator allocator) throws Exception {
       Chunk chunk = chunksToWrite.poll();
       if (chunk != null) {
         chunk.onDequeue();
-        ByteBuf buf = Unpooled.wrappedBuffer(chunk.buffer);
-        progress.addAndGet(chunk.buffer.remaining());
+        progress.addAndGet(chunk.buffer.readableBytes());
         chunksAwaitingCallback.add(chunk);
         if (chunk.isLast) {
-          content = new DefaultLastHttpContent(buf);
+          content = new DefaultLastHttpContent(chunk.buffer);
```
cgtz: should these buffers be retained before sending as content? If the chunk callback called by `operationProgressed` releases the buffer and the content object is also released by downstream channel handlers (not sure if this part happens or not), it may get an IllegalReferenceCountException

cgtz: Confirmed that the outgoing ByteBuf will be released once it reaches the end of the pipeline: https://netty.io/wiki/reference-counted-objects.html#outbound-messages

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -127,6 +127,23 @@
 
   @Override
   public Future<Long> write(ByteBuffer src, Callback<Long> callback) {
+    return write(Unpooled.wrappedBuffer(src), new Callback<Long>() {
+      @Override
+      public void onCompletion(Long result, Exception exception) {
+        if (exception == null) {
```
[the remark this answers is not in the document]
justinlin-linkedin: I suppose when result == null then the exception would be not null.

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -127,6 +127,23 @@
 
   @Override
   public Future<Long> write(ByteBuffer src, Callback<Long> callback) {
+    FutureResult<Long> futureResult = new FutureResult<>();
+    write(Unpooled.wrappedBuffer(src), new Callback<Long>() {
+      @Override
+      public void onCompletion(Long result, Exception exception) {
+        long r = result == null ? (long) 0 : result.longValue();
+        src.position((int) r);
```
zzmao: Why 0 instead of previous src.position()?

ambry-rest/src/test/java/com.github.ambry.rest/NettyResponseChannelTest.java
```diff
@@ -17,10 +17,12 @@
 import com.github.ambry.config.PerformanceConfig;
 import com.github.ambry.config.VerifiableProperties;
 import com.github.ambry.router.Callback;
+import com.github.ambry.utils.NettyByteBufLeakHelper;
 import com.github.ambry.utils.TestUtils;
 import com.github.ambry.utils.Utils;
 import com.github.ambry.utils.UtilsTest;
 import io.netty.buffer.ByteBuf;
+import io.netty.buffer.PooledByteBufAllocator;
 import io.netty.buffer.Unpooled;
```
zzmao: not used.

ambry-router/src/test/java/com.github.ambry.router/InMemoryRouter.java
```diff
@@ -157,7 +157,11 @@ public ByteBuffer getBlob(ByteRange range) throws RouterException {
         } catch (IllegalArgumentException e) {
           throw new RouterException("Invalid range for blob", e, RouterErrorCode.RangeNotSatisfiable);
         }
-        buf = ByteBuffer.wrap(blob.array(), (int) resolvedRange.getStartOffset(), (int) resolvedRange.getRangeSize());
+        byte[] bytes = new byte[(int) resolvedRange.getRangeSize()];
+        ByteBuffer duplicate = blob.duplicate();
+        duplicate.position((int) resolvedRange.getStartOffset());
+        duplicate.get(bytes);
+        buf = ByteBuffer.wrap(bytes);
       }
```
justinlin-linkedin: I have seen a really weird bug, where a range request would somehow pollute the blob data here. I change the way we are dealing with range request data to copying it out. And it fixes the problem. This is weird, but I won't get into too much detail.

PR #1390, opened 2020-02-18
Fix early close refcnt error

justinlin-linkedin:
This pr should fix this test error:
image

In order to prevent concurrent access of chunkIndexToBuf, this pr separates ByteBufs waiting to be written and the ByteBuf that already written but waiting for callback to comeback.

### DIFF
--- a/ambry-router/src/main/java/com.github.ambry.router/CryptoJobHandler.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/CryptoJobHandler.java
@@ -18,6 +18,7 @@ import java.security.GeneralSecurityException;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -68,6 +69,11 @@ class CryptoJobHandler implements Closeable {
         } else {
           logger.error("Unknown type of job seen : " + task.getClass());
         }
+      }
+      try {
+        scheduler.awaitTermination(1000, TimeUnit.MILLISECONDS);
+      } catch (Exception e) {
+        logger.error("Unexpected exception while waiting for crypto jobs to terminate", e);
       }
     }
   }
--- a/ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
@@ -153,6 +153,12 @@ class GetBlobOperation extends GetOperation {
     if (chunkIndexToBuf == null) {
       return;
     }
+    for (Integer key : chunkIndexToBufWaitingForRelease.keySet()) {
+      ByteBuf byteBuf = chunkIndexToBufWaitingForRelease.remove(key);
+      if (byteBuf != null) {
+        byteBuf.release();
+      }
+    }
     for (Integer key : chunkIndexToBuf.keySet()) {
       ByteBuf byteBuf = chunkIndexToBuf.remove(key);
       if (byteBuf != null) {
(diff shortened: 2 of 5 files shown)


ambry-router/src/main/java/com.github.ambry.router/GetBlobOperation.java
```diff
@@ -442,9 +449,12 @@ private void maybeWriteToChannel() {
       // if there are chunks available to be written out, do now.
       if (firstChunk.isComplete() && readCalled) {
         while (operationException.get() == null && chunkIndexToBuf.containsKey(indexOfNextChunkToWriteOut)) {
-          ByteBuf byteBuf = chunkIndexToBuf.get(indexOfNextChunkToWriteOut);
-          asyncWritableChannel.write(byteBuf.nioBuffer(), chunkAsyncWriteCallback);
-          indexOfNextChunkToWriteOut++;
+          ByteBuf byteBuf = chunkIndexToBuf.remove(indexOfNextChunkToWriteOut);
+          if (byteBuf != null) {
+            asyncWritableChannel.write(byteBuf.nioBuffer(), chunkAsyncWriteCallback);
+            chunkIndexToBufWaitingForRelease.put(indexOfNextChunkToWriteOut, byteBuf);
```
cgtz: should this call come before `asyncWritableChannel.write` in case write calls `chunkAsyncWriteCallback` before the buff is added to the map?

justinlin-linkedin: good catch.

ambry-router/src/main/java/com.github.ambry.router/CryptoJobHandler.java
```diff
@@ -69,6 +70,11 @@ public void close() {
           logger.error("Unknown type of job seen : " + task.getClass());
         }
       }
+      try {
+        scheduler.awaitTermination(1000, TimeUnit.MILLISECONDS);
```
jsjtzyy: Could you explain a little why we need wait for 1 sec here?

ambry-router/src/test/java/com.github.ambry.router/GetBlobOperationTest.java
```diff
@@ -1215,6 +1212,9 @@ public void run() {
                   chunksLeftToRead--;
                 }
                 result.getBlobResult.getBlobDataChannel().close();
+                while (writableChannel.getNextChunk(100) != null) {
```
jsjtzyy: In which case, there might be a chunk left? I feel like chunks have been read in previous loop.

ambry-router/src/main/java/com.github.ambry.router/CryptoJobHandler.java
justinlin-linkedin: This is trying to fix the earlyCloseReadableChannel test method.
In that particular test, we would close the blobDataReadableChannel before we finish fetching data chunk from server and decrypting them. So there is a chance when the test method is finished, the decryption of some data chunks are still ongoing in the decryption thread. Since in the decrypt job, we still have a ByteBuf not released, we will run into ByteBuf leak error.
So here we wait for some time for all the decrypt jobs to finish and release all the ByteBufs they hold, and we won't have the leak error.

ambry-router/src/test/java/com.github.ambry.router/GetBlobOperationTest.java
justinlin-linkedin: They are not all read in the previous loop since this test case is to test when we close the ReadableChannel before we finishing read all the data from it, as you can see we close the channel before this line. But the AsyncWritableChannel's interface guarantees that whenever you pass a callback to the write method, this callback will be invoked in the same order they are passed to the channel. So this loop is doing exactly that. It loops over all the left  data chunk in the asyncWritableChannel and calls those callbacks.

jsjtzyy: I see. I thought `chunksLeftToRead` was the total number of chunks that written into the channel.

PR #1391, opened 2020-02-19
Remove getMessageInfo from MessageStoreHardDelete interface

justinlin-linkedin:
Remove unnecessary method getMessageInfo from MessageStoreHardDelete interface.

When the log files are not segmented, each delete index entry has a field "originalMessageOffset" pointing to the original Put record, so that when fetching content for a deleted blob, we don't have to search the Put record from the index again, we simple read the content from the Put record located by the "originalMessageOffset".

This is not true after we segmented log files and start compaction. Now a Delete index entry might not have a "originalMessageOffset" when the Put entry is not located in the same segment as the delete entry, so we have to search the Put entry from the index again not matter what. This totally obsolete the "originalMessageOffset" and thus the logic to read the content from the "origianlMessageOffset".

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
@@ -1055,57 +1055,20 @@ class PersistentIndex {
   private BlobReadOptions getDeletedBlobReadOptions(IndexValue value, StoreKey key,
       ConcurrentSkipListMap<Offset, IndexSegment> indexSegments) throws StoreException {
     BlobReadOptions readOptions;
-    try {
-      IndexValue putValue =
-          findKey(key, new FileSpan(getStartOffset(indexSegments), value.getOffset()), EnumSet.of(IndexEntryType.PUT),
-              indexSegments);
-      if (value.getOriginalMessageOffset() != IndexValue.UNKNOWN_ORIGINAL_MESSAGE_OFFSET
-          && value.getOriginalMessageOffset() != value.getOffset().getOffset()) {
-        // PUT record in the same log segment.
-        String logSegmentName = value.getOffset().getName();
-        // The delete entry in the index might not contain the information about the size of the original blob. So we
-        // use the Message format to read and provide the information. The range in log that we provide starts at the
-        // original message offset and ends at the delete message's start offset (the original message surely cannot go
-        // beyond the start offset of the delete message).
-        MessageInfo deletedBlobInfo =
-            hardDelete.getMessageInfo(log.getSegment(logSegmentName), value.getOriginalMessageOffset(), factory);
-        if (putValue != null && putValue.getOffset().getName().equals(value.getOffset().getName())) {
-          if (putValue.getOffset().getOffset() != value.getOriginalMessageOffset()) {
-            logger.error(
-                "Offset in PUT index entry {} is different from original message offset in delete entry {} for key {}",
-                putValue.getOffset().getOffset(), value.getOriginalMessageOffset(), key);
-            metrics.putEntryDeletedInfoMismatchCount.inc();
-          }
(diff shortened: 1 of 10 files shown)


ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1055,57 +1055,20 @@ BlobReadOptions getBlobReadInfo(StoreKey id, EnumSet<StoreGetOptions> getOptions
...
-      StoreErrorCodes errorCode = StoreException.resolveErrorCode(e);
-      throw new StoreException(errorCode.toString() + " when reading delete blob info from the log " + dataDir, e,
-          errorCode);
+    IndexValue putValue =
+        findKey(key, new FileSpan(getStartOffset(indexSegments), value.getOffset()), EnumSet.of(IndexEntryType.PUT),
+            indexSegments);
+    if (putValue != null) {
+      // use the expiration time from the original value because it may have been updated
+      readOptions = new BlobReadOptions(log, putValue.getOffset(),
+          new MessageInfo(key, putValue.getSize(), true, value.isFlagSet(IndexValue.Flags.Ttl_Update_Index),
```
jsjtzyy: `isFlagSet` -> `isTTLUpdate`

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1055,57 +1055,20 @@ BlobReadOptions getBlobReadInfo(StoreKey id, EnumSet<StoreGetOptions> getOptions
...
-        // original message offset and ends at the delete message's start offset (the original message surely cannot go
-        // beyond the start offset of the delete message).
-        MessageInfo deletedBlobInfo =
-            hardDelete.getMessageInfo(log.getSegment(logSegmentName), value.getOriginalMessageOffset(), factory);
-        if (putValue != null && putValue.getOffset().getName().equals(value.getOffset().getName())) {
-          if (putValue.getOffset().getOffset() != value.getOriginalMessageOffset()) {
-            logger.error(
-                "Offset in PUT index entry {} is different from original message offset in delete entry {} for key {}",
-                putValue.getOffset().getOffset(), value.getOriginalMessageOffset(), key);
-            metrics.putEntryDeletedInfoMismatchCount.inc();
```
jsjtzyy: also, this metric can be deleted from `StoreMetrics`

ambry-store/src/main/java/com.github.ambry.store/PersistentIndex.java
```diff
@@ -1055,57 +1055,20 @@ BlobReadOptions getBlobReadInfo(StoreKey id, EnumSet<StoreGetOptions> getOptions
...
+      // use the expiration time from the original value because it may have been updated
+      readOptions = new BlobReadOptions(log, putValue.getOffset(),
+          new MessageInfo(key, putValue.getSize(), true, value.isFlagSet(IndexValue.Flags.Ttl_Update_Index),
+              value.getExpiresAtMs(), putValue.getAccountId(), putValue.getContainerId(),
+              putValue.getOperationTimeInMs()));
+    } else {
+      // PUT record no longer available.
+      throw new StoreException("Did not find PUT index entry for key [" + key
+          + "] and the the original offset in value of the DELETE entry was [" + value.getOriginalMessageOffset() + "]",
+          StoreErrorCodes.ID_Deleted);
```
jsjtzyy: I wonder why the error code is ID_Deleted rather than ID_Not_Found?

justinlin-linkedin: I think it can be further detailed. It means the original Put Record is already compacted, or this node only replicates the Delete record from it's peer, which are different error code and ambry frontend would react differently. But this is what we have before so I will just keep as it's.

PR #1393, opened 2020-02-20
Cloud compaction retries and throttling

lightningrob:
Goal is to eliminate the Cosmos 429 errors by breaking query up into smaller ones.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageCompactor.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageCompactor.java
@@ -15,9 +15,11 @@ package com.github.ambry.cloud;
 
 import com.codahale.metrics.Timer;
 import com.github.ambry.clustermap.PartitionId;
+import com.github.ambry.config.CloudConfig;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -26,8 +28,11 @@ public class CloudStorageCompactor implements Runnable {
   private static final Logger logger = LoggerFactory.getLogger(CloudStorageCompactor.class);
   private final CloudDestination cloudDestination;
   private final Set<PartitionId> partitions;
+  private final int queryLimit;
+  private final long compactionTimeLimitMs;
   private final VcrMetrics vcrMetrics;
   private final boolean testMode;
+  private final CloudRequestAgent requestAgent;
 
   /**
    * Public constructor.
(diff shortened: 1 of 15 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
```diff
@@ -155,6 +157,28 @@ void testConnectivity() {
...
+   * @param timer the {@link Timer} to use to record query time (excluding waiting).
+   * @return the number of matching blobs.
+   */
+  int countMetadata(String partitionPath, SqlQuerySpec querySpec, Timer timer)
+      throws DocumentClientException {
+    FeedOptions feedOptions = new FeedOptions();
+    feedOptions.setPartitionKey(new PartitionKey(partitionPath));
+    try {
+      FeedResponse<Document> response = executeCosmosQuery(querySpec, feedOptions, timer).single();
+      return ((Number) response.getResults().get(0).getHashMap().values().stream().findFirst().get()).intValue();
```
cgtz: what is the format of results here? Instead of findFirst could you get the relevant entry by looking up its key?

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
```diff
@@ -298,23 +269,58 @@ public void testPurgeDeadBlobs() throws Exception {
...
+   */
+  @Test
+  public void testFindEntriesSinceByUpdateTime() throws Exception {
+    testFindEntriesSince("com.github.ambry.cloud.azure.CosmosUpdateTimeFindTokenFactory");
+  }
+
+  /**
+   * Test findEntriesSince with CosmosChangeFeedFindTokenFactory.
    */
+  @Ignore // Fails with wrong number of queries.
```
cgtz: let's make a github issue to track that this test needs a fix

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageCompactor.java
```diff
@@ -91,14 +99,25 @@ public int compactPartitions() {
...
-      return deadBlobs.size();
+  public int compactPartition(String partitionPath, long cutoffTime) throws CloudStorageException {
+    if (testMode) {
+      return cloudDestination.getDeadBlobCount(partitionPath, cutoffTime);
     }
+
+    int numDeadBlobs = 0;
+    // Iterate until returned list size < limit or time runs out
+    while (System.currentTimeMillis() < cutoffTime + compactionTimeLimitMs) {
+      List<CloudBlobMetadata> deadBlobs =
```
cgtz: do we need to sleep in between successive calls to getDeadBlobs, or do you feel that the cosmo's retry-with-backoff will be good enough for rate limiting this query?

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosDataAccessor.java
lightningrob: The returned document is: {"_aggregate":20}
I can change it to lookup by the key, it just involves ugly hardcoding and I could break if they decide to change the key name.

lightningrob: Never mind, Cosmos does have a constant for it: Constants.Properties.AGGREGATE, so I'll make the change.

ambry-cloud/src/main/java/com.github.ambry.cloud/CloudStorageCompactor.java
lightningrob: There is an expensive call to cloudDestination.purgeBlobs() between calls to getDeadBlobs, and that should provide more than enough throttling between queries.

ambry-cloud/src/test/java/com.github.ambry.cloud/azure/AzureIntegrationTest.java
lightningrob: Good point.  Filed https://github.com/linkedin/ambry/issues/1394 to track.

PR #1399, opened 2020-02-25
Improve Helix bootstrap tool for moving replica

jsjtzyy:
1. Assigned dedicated thread for each dc to speed up upgrading clustermap in Helix. It allows tool to bootstrap/upgrade multiple dcs concurrently.
2. Introduced an option that only update resource (IdealState) in Helix without changing InstanceConfig. This is used when initiating replica movement (We change IdealState in Helix to trigger moving replica and Helix is responsible for notifying nodes that are involved). The InstanceConfig will be updated eventually by ambry nodes themselves.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
@@ -33,6 +33,10 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import org.apache.helix.AccessOption;
 import org.apache.helix.HelixAdmin;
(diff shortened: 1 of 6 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -94,22 +98,28 @@
...
-  private Map<String, Set<String>> partitionsNotForceRemovedByDc = new HashMap<>();
-  private int instancesAdded = 0;
-  private int instancesUpdated = 0;
-  private int instancesDropped = 0;
-  private int resourcesAdded = 0;
-  private int resourcesUpdated = 0;
-  private int resourcesDropped = 0;
+  private ConcurrentHashMap<String, Set<String>> instancesNotForceRemovedByDc = new ConcurrentHashMap<>();
+  private ConcurrentHashMap<String, Set<String>> partitionsNotForceRemovedByDc = new ConcurrentHashMap<>();
+  private final AtomicInteger instancesAdded = new AtomicInteger();
```
xuhao417347761: Since instancesAdded is changed during execution, so does it need to be final?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -645,35 +674,41 @@ private void addUpdateInstances(String dcName, Map<String, Set<String>> partitio
...
             expectMoreInHelixDuringValidate = true;
-            partitionsNotForceRemovedByDc.computeIfAbsent(dcName, k -> new HashSet<>()).add(partitionName);
+            partitionsNotForceRemovedByDc.computeIfAbsent(dcName, k -> ConcurrentHashMap.newKeySet())
+                .add(partitionName);
           }
         } else if (!instanceSetInStatic.equals(instanceSetInHelix)) {
-          info("Different instance sets for partition {} under resource {}. Updating Helix using static.",
-              partitionName, resourceName);
+          info(
+              "[{}] Different instance sets for partition {} under resource {}. Updating Helix using static. Previous instance set: [{}], new instance set: [{}]",
```
xuhao417347761: Too long, better to split it.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
jsjtzyy: the `final` means the reference doesn't change. The value of this object is allowed to be mutable.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixBootstrapUpgradeUtil.java
jsjtzyy: Sorry, I don't quite understand, could you offer some suggestions here? Thanks in advance.

xuhao417347761: I mean this line or string is too long. It is better to split into two lines.

PR #1400, opened 2020-02-27
Enable publishing to main maven repo instead of test

cgtz:
- The new pipeline has been verified, so we can enable publishing to the
  main maven repo in the linkedin bintray acct instead of the test repo.
- Remove scripts/artifacts related to old release pipeline (shipkit
  plugin)
- Add metadata to maven pom from old shipkit pom: license, scm, issues, CI


gradle/java-publishing.gradle
```diff
@@ -33,10 +33,31 @@ publishing {
     publications {
         maven(MavenPublication) {
             from components.java
-            
+
             artifact javadocJar
             artifact sourcesJar
             artifact testJar
+
+            pom {
```
mockitoguy: Looks great!

PR #1402, opened 2020-02-27
Add more tests to hard deleter for undelete record

justinlin-linkedin:
Adding more tests to hard deleter to test undelete record.
Change how hard deleter prune the recovery range.

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
@@ -89,9 +89,9 @@ public class HardDeleter implements Runnable {
    * startTokenBeforeLogFlush: This token is set to the current start token just before log flush and once the log is
    *                           flushed, this is used to set startTokenSafeToPersist.
    */
-  private FindToken startToken;
   private FindToken startTokenBeforeLogFlush;
-  private FindToken startTokenSafeToPersist;
+  private volatile FindToken startTokenSafeToPersist;
+  private FindToken startToken;
   private FindToken endToken;
   private StoreFindToken recoveryEndToken;
   private HardDeletePersistInfo hardDeleteRecoveryRange = new HardDeletePersistInfo();
@@ -269,9 +269,21 @@ public class HardDeleter implements Runnable {
              correctness issue. Additionally, since this token is journal based, it is highly likely that this token
              will soon become equal to endtoken, in which case we will prune everything (in the "if case" above).
              If the token is index based, safely prune off entries that have already been flushed in the log */
-        hardDeleteRecoveryRange.pruneTill(logFlushedTillToken.getStoreKey());
+        hardDeleteRecoveryRange.pruneTill(logFlushedTillToken);
       }
     }
+    if (hardDeleteRecoveryRange.getSize() > 0) {
+      logger.trace("Pruned Hard delete recovery range : {} to {} with size {} for {}",
+          hardDeleteRecoveryRange.blobReadOptionsList.get(0),
+          hardDeleteRecoveryRange.getBlobReadOptionsList().get(hardDeleteRecoveryRange.getSize() - 1),
+          hardDeleteRecoveryRange.getSize(), dataDir);
+    } else {
+      logger.trace("Hard delete recovery range is empty for {}", dataDir);
+    }
+  }
+
+  HardDeletePersistInfo getHardDeleteRecoveryRange() {
+    return hardDeleteRecoveryRange;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -653,22 +680,39 @@ private void performHardDeletes(List<MessageInfo> messageInfoList) throws StoreE
     }
   }
 
+  class HardDeletePersistItem {
```
jsjtzyy: java doc for this class please

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -352,6 +359,24 @@ void preLogFlush() {
   void postLogFlush() {
     /* start token saved before the flush is now safe to be persisted */
     startTokenSafeToPersist = startTokenBeforeLogFlush;
+
+    hardDeleteLock.lock();
+    try {
+      // PersistCleanupToken because startTokenSafeToPersist changed.
+      pruneHardDeleteRecoveryRange();
+      persistCleanupToken();
```
jsjtzyy: What if recovery range has been pruned but failed to persist cleanup token, will it be a problem?

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -625,9 +652,9 @@ private void performHardDeletes(List<MessageInfo> messageInfoList) throws StoreE
           throw new StoreException("Aborting hard deletes as store is shutting down",
               StoreErrorCodes.Store_Shutting_Down);
         }
+        diskIOScheduler.getSlice(HARD_DELETE_CLEANUP_JOB_NAME, HARD_DELETE_CLEANUP_JOB_NAME, logWriteInfo.size);
```
jsjtzyy: any reason to change the ordering here? Throttle the hard delete I/O ?

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -705,52 +751,64 @@ int getSize() {
...
-    private List<BlobReadOptions> getBlobReadOptionsList() {
-      return blobReadOptionsList;
+    /**
+     * Compare two StoreFindTokens and return the result as an integer like compareTo interface.
+     * These two tokens have to be IndexBased tokens.
+     * @param token1 The first token to compare.
+     * @param token2 The second tokent to compare.
+     * @return 0 means they are equal. negative number means token1 is less than token2. postive number means the opposite.
+     */
+    int compareTwoTokens(StoreFindToken token1, StoreFindToken token2) {
```
jsjtzyy: The method looks good, minor suggestion is to align with your comment and do a token type check at the very beginning. (Reject Journal based token)

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -705,52 +751,64 @@ int getSize() {
...
-         later occurrence is the one really associated with this token, it does not affect the safety.
-         Persisting more than what is required is okay as hard deleting a blob is an idempotent operation. */
-        messageStoreRecoveryListIterator.next();
-        if (blobReadOptionsListIterator.next().getMessageInfo().getStoreKey().equals(storeKey)) {
-          break;
+    void pruneTill(StoreFindToken token) {
+      Iterator<HardDeletePersistItem> itemsIterator = items.iterator();
+      while (itemsIterator.hasNext()) {
+        HardDeletePersistItem item = itemsIterator.next();
+        if (item.startTokenForBlobReadOptions.getType() == FindTokenType.Uninitialized) {
```
jsjtzyy: Could you explain a little bit about this case?

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -89,9 +89,9 @@
    * startTokenBeforeLogFlush: This token is set to the current start token just before log flush and once the log is
    *                           flushed, this is used to set startTokenSafeToPersist.
    */
-  private FindToken startToken;
   private FindToken startTokenBeforeLogFlush;
-  private FindToken startTokenSafeToPersist;
+  private volatile FindToken startTokenSafeToPersist;
```
jsjtzyy: For concurrency, let's make sure all variables shared by HardDeleter thread and Index persistor thread to be volatile.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: updated

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: It's fine if it's fails to persist, then we will wait for the next one. 
The assumption is that the file on the disk contains a valid range and all the records between this range. And every time we change the file (prune and then hard delete), we replace the old range with a new one. If we fail to persist it, then next time when we restart, we are going to recover from a older range, which is totally fine, since the hard delete operation is idempotent.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: the reason for changing the order is to use diskIOScheduler.getSlice to throw an exception before  modify the log file. The reason why I want to throw an exception is to mock the situation when we persist the HardDeleteRecoveryRange but fail to actually hard delete the records in the log file. Under such situation, we can test if the recovery actually works or not.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: good catch, updated.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: The very first time when we starting scanning through the log file, we are using a Uninitialized FindToken. And since it's at the very first beginning of a log file, we can assume they are always prior to any IndexBased FindToken.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: sure.

PR #1404, opened 2020-02-27
Fix for ResponseMetadata race condition.

zzmao:
Before this change, responseMetadata is accessed without confirming if it is ready or not.
This fix introduces a latch to avoid such race condition.

### DIFF
--- a/ambry-rest/src/main/java/com.github.ambry.rest/NettyMetrics.java
+++ b/ambry-rest/src/main/java/com.github.ambry.rest/NettyMetrics.java
@@ -37,7 +37,6 @@ public class NettyMetrics {
   public final Meter channelDestructionRate;
   public final Meter requestArrivalRate;
   public final Meter multipartPostRequestRate;
-  public final Meter channelStatusInconsistent;
   // NettyResponseChannel
   public final Meter bytesWriteRate;
   public final Meter requestCompletionRate;
@@ -135,6 +134,7 @@ public class NettyMetrics {
   public final Counter requestTooLargeCount;
   public final Counter throwableCount;
   public final Counter unknownResponseStatusCount;
+  public final Counter channelStatusInconsistentCount;
   // NettyServer
   public final Histogram nettyServerShutdownTimeInMs;
   public final Histogram nettyServerStartTimeInMs;
@@ -171,8 +171,6 @@ public class NettyMetrics {
     publicAccessLogRequestRate =
         metricRegistry.meter(MetricRegistry.name(PublicAccessLogHandler.class, "RequestArrivalRate"));
     healthCheckRequestRate = metricRegistry.meter(MetricRegistry.name(HealthCheckHandler.class, "RequestArrivalRate"));
-    channelStatusInconsistent =
-        metricRegistry.meter(MetricRegistry.name(NettyMessageProcessor.class, "ChannelStatusInconsistent"));
 
     // Latencies
     // NettyMessageProcessor
@@ -321,6 +319,8 @@ public class NettyMetrics {
     throwableCount = metricRegistry.counter(MetricRegistry.name(NettyResponseChannel.class, "ThrowableCount"));
     unknownResponseStatusCount =
         metricRegistry.counter(MetricRegistry.name(NettyResponseChannel.class, "UnknownResponseStatusCount"));
+    channelStatusInconsistentCount =
+        metricRegistry.counter(MetricRegistry.name(NettyResponseChannel.class, "ChannelStatusInconsistentCount"));
     // NettyServer
     nettyServerShutdownTimeInMs = metricRegistry.histogram(MetricRegistry.name(NettyServer.class, "ShutdownTimeInMs"));
     nettyServerStartTimeInMs = metricRegistry.histogram(MetricRegistry.name(NettyServer.class, "StartTimeInMs"));
--- a/ambry-rest/src/main/java/com.github.ambry.rest/NettyMetrics.java
+++ b/ambry-rest/src/main/java/com.github.ambry.rest/NettyMetrics.java
@@ -37,6 +37,7 @@ public class NettyMetrics {
   public final Meter channelDestructionRate;
   public final Meter requestArrivalRate;
   public final Meter multipartPostRequestRate;
+  public final Meter channelStatusInconsistent;
   // NettyResponseChannel
   public final Meter bytesWriteRate;
   public final Meter requestCompletionRate;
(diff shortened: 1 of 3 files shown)


zzmao:
Internal Ticket: EXC-38958

zzmao:
Need to figure out why this happen.

zzmao:
Couldn't figure out where is the concurrency. GetBlobOperation::maybeWriteToChannel is the caller, but it uses while loop instead of multi thread. @cgtz  any insight?

cgtz:
I will look through the code a little more to see why this may happen. I worry that the latch may never get counted down if there is some other bug.

justinlin-linkedin:
> I will look through the code a little more to see why this may happen. I worry that the latch may never get counted down if there is some other bug. I guess the NPE could happen if there are two calls to `maybeWriteToChannel` in close proximity (concurrent write() calls?)

write call should be a blocking call that creates the finalRespoonseMetadata and maybeWriteToChannel seems like it's only invoked in a single thread.
Is it possible that the http channel is not active when we start the first write? in this case the finalResponseMetadata will not be constructed but yet we expect it in the Chunk constructor. Just. a thought.

zzmao:
I have same guess with @justinlin-linkedin ,

```
 if (responseMetadataWriteInitiated.compareAndSet(false, true) && ctx.channel().isActive()) {
```

It sets responseMetadataWriteInitiated to true, but channel is not active. 
@cgtz

zzmao:
Made another change. Please take a second look at. @cgtz @justinlin-linkedin

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -148,6 +148,18 @@ public void onCompletion(Long result, Exception exception) {
     if (!responseMetadataWriteInitiated.get()) {
       maybeWriteResponseMetadata(responseMetadata, new ResponseMetadataWriteListener());
     }
+    if (finalResponseMetadata == null) {
```
justinlin-linkedin: this looks good, but can you add a test case for this logic?

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -148,6 +148,18 @@ public void onCompletion(Long result, Exception exception) {
...
+      // If finalResponseMetadata is still null, it indicates channel becomes inactive.
+      if (ctx.channel().isActive()) {
+        logger.warn("Channel should be inactive status. {}", ctx.channel());
+      } else {
+        logger.debug("Scheduling a chunk cleanup on channel {} because response channel is closed.", ctx.channel());
+        writeFuture.addListener(new CleanupCallback(new ClosedChannelException()));
+      }
+      FutureResult<Long> future = new FutureResult<Long>();
+      future.done(-1L, null);
```
justinlin-linkedin: the exception should be ClosedChannelException, not null.

cgtz: Also call the callback here?

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
cgtz: +1, I guess we can at least test the case where the channel becomes inactive

ambry-rest/src/main/java/com.github.ambry.rest/NettyResponseChannel.java
```diff
@@ -148,6 +148,18 @@ public void onCompletion(Long result, Exception exception) {
     if (!responseMetadataWriteInitiated.get()) {
       maybeWriteResponseMetadata(responseMetadata, new ResponseMetadataWriteListener());
     }
+    if (finalResponseMetadata == null) {
+      // If finalResponseMetadata is still null, it indicates channel becomes inactive.
+      if (ctx.channel().isActive()) {
+        logger.warn("Channel should be inactive status. {}", ctx.channel());
```
cgtz: Could you add a metric for this? We can set up an alert if this case ever occurs.

zzmao:
@cgtz  @justinlin-linkedin 

Addressed comments and added test. Please take another round of review. Thanks!

The new test can reproduce the issue without fix and test passes with the fix.

ambry-rest/src/test/java/com.github.ambry.rest/NettyResponseChannelTest.java
```diff
@@ -1713,3 +1746,221 @@ void compareWithFuture() throws InterruptedException, TimeoutException {
     }
   }
 }
+
+/**
+ * Mock class for ChannelHandlerContext used in .
```
cgtz: add "in channelInactiveTest()"

ambry-rest/src/test/java/com.github.ambry.rest/NettyResponseChannelTest.java
```diff
@@ -636,6 +647,28 @@ public void closeTest() {
...
+    ChunkedWriteHandler chunkedWriteHandler = new ChunkedWriteHandler();
+    EmbeddedChannel channel = new EmbeddedChannel(chunkedWriteHandler);
+    NettyResponseChannel nettyResponseChannel =
+        new NettyResponseChannel(new MockChannelHandlerContext(channel), new NettyMetrics(new MetricRegistry()),
+            new PerformanceConfig(new VerifiableProperties(new Properties())));
+    channel.disconnect().awaitUninterruptibly();
+    assertFalse("Channel should be closed", channel.isOpen());
+    Future<Long> future = nettyResponseChannel.write(Unpooled.buffer(1), null);
+    try {
+      future.get();
```
justinlin-linkedin: please add a fail("") after future.get();

ambry-rest/src/main/java/com.github.ambry.rest/NettyMetrics.java
```diff
@@ -170,6 +171,8 @@ public NettyMetrics(MetricRegistry metricRegistry) {
     publicAccessLogRequestRate =
         metricRegistry.meter(MetricRegistry.name(PublicAccessLogHandler.class, "RequestArrivalRate"));
     healthCheckRequestRate = metricRegistry.meter(MetricRegistry.name(HealthCheckHandler.class, "RequestArrivalRate"));
+    channelStatusInconsistent =
```
cgtz: It seems like this may fit better as a counter, looking at other similar metrics (see line 107-113). However, they both should work about the same in practice, so if you prefer Meter keep it as is.

zzmao:
Comments addressed.

PR #1405, opened 2020-02-28
Make PartitionSelectionHelper incorporate clustermap changes

jsjtzyy:
Previously, data structures in PartitionSelectionHelper are constant after initialization. With "Move Replica" feature on, this should dynamically incorporate any changes in clustermap (i.e. new replica / partition added or old replica removed). In this PR, we make PartitionSelectionHelper as a listener of clustermap and re-populate all in-mem data structures if number of replicas / partitions has changed.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
@@ -20,6 +20,7 @@ import java.io.InputStream;
 import java.util.List;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 import org.json.JSONArray;
 import org.json.JSONObject;
 
@@ -46,7 +47,7 @@ public class AmbryPartition implements PartitionId {
    * Instantiate an AmbryPartition instance.
    * @param id the id associated with this partition.
    * @param partitionClass the partition class that this partition belongs to
-   * @param clusterManagerCallback the {@link ClusterManagerCallback} to use to make callbacks
+   * @param clusterManagerCallback the {@link HelixClusterManager.HelixClusterManagerCallback} to use to make callbacks
    *                               to the {@link HelixClusterManager}
    * The initial state defaults to {@link PartitionState#READ_WRITE}.
    */
@@ -64,12 +65,18 @@ public class AmbryPartition implements PartitionId {
 
   @Override
   public List<AmbryReplica> getReplicaIds() {
-    return clusterManagerCallback.getReplicaIdsForPartition(this);
+    return clusterManagerCallback.getReplicaIdsForPartition(this)
+        .stream()
+        .map(r -> (AmbryReplica) r)
+        .collect(Collectors.toList());
   }
 
   @Override
   public List<AmbryReplica> getReplicaIdsByState(ReplicaState state, String dcName) {
-    return clusterManagerCallback.getReplicaIdsByState(this, state, dcName);
+    return clusterManagerCallback.getReplicaIdsByState(this, state, dcName)
+        .stream()
+        .map(r -> (AmbryReplica) r)
+        .collect(Collectors.toList());
   }
 
   @Override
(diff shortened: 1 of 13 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
```diff
@@ -64,12 +65,18 @@
 
   @Override
   public List<AmbryReplica> getReplicaIds() {
-    return clusterManagerCallback.getReplicaIdsForPartition(this);
+    return clusterManagerCallback.getReplicaIdsForPartition(this)
```
cgtz: Instead of copying the list and requiring a (potentially) unsafe cast. I think we could make more use of generics in `ClusterManagerCallback`.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterManagerCallback.java
```diff
@@ -18,26 +18,25 @@
 
 
 /**
- * A callback that needs to be implemented by dynamic implementations of the cluster manager which can be
- * used by dynamic cluster manager components such as {@link AmbryDataNode}, {@link AmbryDisk},
- * {@link AmbryPartition}, and {@link AmbryReplica}
+ * A callback that needs to be implemented by different implementations of the cluster manager. External components may
+ * get resources (i.e. {@link ReplicaId}, {@link PartitionId}, {@link DiskId}) via this callback.
  */
 interface ClusterManagerCallback {
```
cgtz: `ClusterManagerCallback<R extends ReplicaId, D extends DiskId, P extends PartitionId>`

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
```diff
@@ -46,7 +47,7 @@
    * Instantiate an AmbryPartition instance.
    * @param id the id associated with this partition.
    * @param partitionClass the partition class that this partition belongs to
-   * @param clusterManagerCallback the {@link ClusterManagerCallback} to use to make callbacks
+   * @param clusterManagerCallback the {@link HelixClusterManager.HelixClusterManagerCallback} to use to make callbacks
    *                               to the {@link HelixClusterManager}
    * The initial state defaults to {@link PartitionState#READ_WRITE}.
    */
```
cgtz: And then pass in `ClusterManagerCallback<AmbryReplica, AmbryDisk, AmbryPartition>` here and in other similar classes, which will get rid of the need for the copies/casts.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterManagerCallback.java
jsjtzyy: good point, changes are made

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -510,7 +512,9 @@ public ReplicaId getBootstrapReplica(String partitionIdStr, DataNodeId dataNodeI
 
   @Override
   public void registerClusterMapListener(ClusterMapChangeListener clusterMapChangeListener) {
-    clusterMapChangeListeners.add(clusterMapChangeListener);
+    for (DcInfo dcInfo : dcToDcZkInfo.values()) {
+      dcInfo.clusterChangeHandler.registerClusterMapListener(clusterMapChangeListener);
```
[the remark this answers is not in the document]
jsjtzyy: Discussed offline. Basically, the `ClusterChangeHandlerCallback. addPartitionIfAbsent()` might be called multiple times in the for-loop (see `updateInstanceInfo` method in DynamicClusterChangeHandler). What I am doing is to aggregate the added replicas and invoke callback in `PartitionSelectionHelper` only once.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -393,11 +421,12 @@ private void updateReplicaStateAndOverrideIfNeeded(AmbryReplica replica, List<St
   /**
    * Create a new instance(node) and initialize disks/replicas on it.
    * @param instanceConfig the {@link InstanceConfig} to create new instance
+   * @return a pair of lists: (1) newly added replicas; (2) removed old replicas(in this method, it should be empty)
```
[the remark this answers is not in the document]
xuhao417347761: Return a pair instead of return multiple values in java function(not allowed)? Just curious about this.

PR #1406, opened 2020-02-28
Fix a bug in the segment range request

justinlin-linkedin:
When we resolve the range request, we should pass the chunk size instead of total blob size to the method.

### DIFF
--- a/ambry-router/src/test/java/com.github.ambry.router/GetManagerTest.java
+++ b/ambry-router/src/test/java/com.github.ambry.router/GetManagerTest.java
@@ -72,7 +72,7 @@ public class GetManagerTest {
   private ReadableStreamChannel putChannel;
   private GetBlobOptions options = new GetBlobOptionsBuilder().build();
   private List<ChunkInfo> chunkInfos;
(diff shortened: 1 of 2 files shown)


cgtz:
Thank you for fixing this and improving the test case

ambry-router/src/test/java/com.github.ambry.router/GetManagerTest.java
```diff
@@ -72,7 +72,7 @@
   private ReadableStreamChannel putChannel;
   private GetBlobOptions options = new GetBlobOptionsBuilder().build();
   private List<ChunkInfo> chunkInfos;
-  private static final int CHUNK_SIZE = new Random().nextInt(1024 * 1024) + 1;
+  private static final int CHUNK_SIZE = new Random().nextInt(1024 * 1024) + 8191;
```
xuhao417347761: 1. The CHUNK_SIZE and chunkSize are confusing here. It seems that CHUNK_SIZE is used for initializing LARGE_BLOB_SIZE here. 
2. Why we need to change CHUNK_SIZE into a value: rand + 8191?

PR #1407, opened 2020-03-01
Change blob type from java.nio.ByteBuffer to netty ByteBuf in PutRequest

justinlin-linkedin:
Change the blob's type in PutRequest from ByteBuffer to ByteBuf.
Later will make the change to pass a ByteBuf from PutOperation to PutRequest.

### DIFF
--- a/ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
+++ b/ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
@@ -171,35 +171,35 @@ public class PutRequest extends RequestOrResponse {
     return sizeExcludingBlobAndCrc;
   }
 
+  /**
+   * Construct the bufferToSend to serialize request metadata and other blob related information. The newly constructed
+   * bufferToSend will not include the blob content as it's carried by the {@code blob} field in this class.
+   */
   private void prepareBuffer() {
-    if (bufferToSend == null) {
-      // this is the first time this method was called, prepare the buffer to send the header and other metadata
-      // (everything except the blob content).
-      bufferToSend = ByteBuffer.allocate(sizeExcludingBlobAndCrcSize());
-      writeHeader();
-      int crcStart = bufferToSend.position();
-      bufferToSend.put(blobId.toBytes());
-      BlobPropertiesSerDe.serializeBlobProperties(bufferToSend, properties);
-      bufferToSend.putInt(usermetadata.capacity());
-      bufferToSend.put(usermetadata);
-      bufferToSend.putShort((short) blobType.ordinal());
-      short keyLength = blobEncryptionKey == null ? 0 : (short) blobEncryptionKey.remaining();
-      bufferToSend.putShort(keyLength);
-      if (keyLength > 0) {
-        bufferToSend.put(blobEncryptionKey);
-      }
-      bufferToSend.putLong(blobSize);
-      crc.update(bufferToSend.array(), bufferToSend.arrayOffset() + crcStart, bufferToSend.position() - crcStart);
-      nioBuffersFromBlob = blob.nioBuffers();
-      for (ByteBuffer bb : nioBuffersFromBlob) {
-        crc.update(bb);
-        // change it back to 0 since we are going to write it to the channel later.
-        bb.position(0);
-      }
-      crcBuf.putLong(crc.getValue());
-      crcBuf.flip();
-      bufferToSend.flip();
+    bufferToSend = ByteBuffer.allocate(sizeExcludingBlobAndCrcSize());
+    writeHeader();
+    int crcStart = bufferToSend.position();
+    bufferToSend.put(blobId.toBytes());
+    BlobPropertiesSerDe.serializeBlobProperties(bufferToSend, properties);
+    bufferToSend.putInt(usermetadata.capacity());
+    bufferToSend.put(usermetadata);
+    bufferToSend.putShort((short) blobType.ordinal());
+    short keyLength = blobEncryptionKey == null ? 0 : (short) blobEncryptionKey.remaining();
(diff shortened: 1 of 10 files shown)


ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
```diff
@@ -168,51 +171,80 @@ private int sizeExcludingBlobAndCrcSize() {
...
 
+  private void prepareBuffer() {
+    if (bufferToSend == null) {
+      // this is the first time this method was called, prepare the buffer to send the header and other metadata
+      // (everything except the blob content).
+      bufferToSend = ByteBuffer.allocate(sizeExcludingBlobAndCrcSize());
+      writeHeader();
+      int crcStart = bufferToSend.position();
+      bufferToSend.put(blobId.toBytes());
+      BlobPropertiesSerDe.serializeBlobProperties(bufferToSend, properties);
```
zzmao: Is it possible to convert BlobProperties and userMeatadata to bytebuf first, and then use `Bytebuf bufferToSend`?

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1346,8 +1347,8 @@ private void fetchRequests(RequestRegistrationCallback<PutOperation> requestRegi
      */
     protected PutRequest createPutRequest() {
       return new PutRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
-          chunkBlobId, chunkBlobProperties, ByteBuffer.wrap(chunkUserMetadata), buf.duplicate(), buf.remaining(),
-          BlobType.DataBlob, encryptedPerBlobKey != null ? encryptedPerBlobKey.duplicate() : null);
+          chunkBlobId, chunkBlobProperties, ByteBuffer.wrap(chunkUserMetadata), Unpooled.wrappedBuffer(buf.duplicate()),
```
zzmao: Unpooled.wrappedBuffer(chunkUserMetadata)?

ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
justinlin-linkedin: Use bytebuf as usermetadata's data type is currently in my plan, but I would do blob data first since blob data is usually much larger than the user metadata.

zzmao: Yes, but if you can make the change in one shot, you can eliminate ByteBuffer bufferToSend. Everything can be ByteBuf based in PutRequest.

justinlin-linkedin: The problem is bufferToSend belongs to the it's parent class RequestOrResponse. To change the type of bufferToSend would trigger lots of changes for other classes as well.

zzmao: Any plan to change it?

justinlin-linkedin: We can change it later. I have a jira to change user metadata to use netty bytebuf. will do it with this jira.

ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
```diff
@@ -168,51 +171,80 @@ private int sizeExcludingBlobAndCrcSize() {
...
-        bufferToSend.putInt(usermetadata.capacity());
-        bufferToSend.put(usermetadata);
-        bufferToSend.putShort((short) blobType.ordinal());
-        short keyLength = blobEncryptionKey == null ? 0 : (short) blobEncryptionKey.remaining();
-        bufferToSend.putShort(keyLength);
-        if (keyLength > 0) {
-          bufferToSend.put(blobEncryptionKey);
+    try {
+      if (sentBytes < sizeInBytes()) {
+        prepareBuffer();
```
jsjtzyy: optional:  I would suggest renaming this method and moving the "if" statement out of method. Something like this:
```
if (bufferToSend == null) {
 // this is the first time this method was called, prepare the buffer to send the header and other metadata
   prepareBufferToSend();
}
```

ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
```diff
@@ -168,51 +171,80 @@ private int sizeExcludingBlobAndCrcSize() {
...
-      }
 
-      // If the header and metadata are not yet written out completely, try and write out as much of it now.
-      if (bufferToSend.hasRemaining()) {
-        written = channel.write(bufferToSend);
-      }
+        // If the header and metadata were written out completely (in this call or a previous call),
+        // try and write out as much of the blob now.
+        if (!bufferToSend.hasRemaining() && blob != null) {
+          int n = 0, currentWritten = -1;
```
jsjtzyy: Could you rename `n` to something like `oneTimeBytesWritten`?  This indicates that, when current method is invoked, how many bytes have been written during this call. This may contain one or more nio buffers that are successfully written into channel.

ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
justinlin-linkedin: sure

ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
justinlin-linkedin: will update.

PR #1412, opened 2020-03-02
Add basic support for cloud replicas in clustermap

cgtz:
- Add a representation of a cloud datacenter to configs
- Refactor AmbryReplica into abstract class to allow for cloud and disk
  backed replicas in HelixClusterManager.
- Implement listener logic to create a CloudAmbryReplica whenever a new
  partition is registered in the clustermap. This makes the assumption
  that any partition that exists in Ambry will be accessible via a cloud
  store if a cloud datacenter is present.

Note that this PR does not add first-class support for VCR clusters in
HelixClusterManager. That will come later.

**This is currently a draft PR. I still have to add test cases and fix some issues**


ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -254,8 +179,11 @@ private HelixManager initializeHelixManagerAndPropertyStoreInLocalDC(Map<String,
       String instanceName, HelixFactory helixFactory) throws Exception {
     DcZkInfo dcZkInfo = dataCenterToZkAddress.get(clusterMapConfig.clusterMapDatacenterName);
     String zkConnectStr = dcZkInfo.getZkConnectStr();
-    HelixManager manager;
-    manager = helixFactory.getZKHelixManager(clusterName, instanceName, InstanceType.SPECTATOR, zkConnectStr);
+    if (dcZkInfo.getReplicaType() == ReplicaType.CLOUD_BACKED) {
+      return null;
```
lightningrob: Minor: add javadoc that it returns null in this case.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudChangeListener.java
```diff
@@ -0,0 +1,66 @@
...
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Currently, this just adds a single {@link ReplicaType#CLOUD_BACKED} replica every time a partition is added to allow
+ * for requests to be routed to cloud services. Currently it is only registered as a {@link ClusterMapChangeListener},
+ * but once the clustermap natively supports VCR clusters, this class can implement {@link ClusterChangeHandler} and
+ * listen for changes in the VCR cluster.
+ */
+class CloudChangeListener implements ClusterMapChangeListener {
```
lightningrob: Minor: suggest CloudClusterMapChangeListener.

ambry-utils/src/main/java/com.github.ambry.utils/Utils.java
```diff
@@ -310,7 +310,7 @@ public static ByteBuf readNettyByteBufFromCrcInputStream(CrcInputStream crcStrea
    * @param <T> The exception to throw in this operation.
    */
   @FunctionalInterface
-  public static interface ByteBufferFunction<T extends Throwable> {
+  public interface ByteBufferFunction<T extends Throwable> {
```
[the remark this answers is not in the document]
cgtz: It was a suggested change from intellij. The static is redundant since inner interfaces are inherently static (no outer class instance variables can be accessed)

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectator.java
```diff
@@ -63,15 +63,17 @@ public void spectate() throws Exception {
...
     // Once we add more fabrics, we should revisit this.
-    for (Map.Entry<String, DcZkInfo> entry : dataCenterToZkAddress.entrySet()) {
-      String zkConnectStr = entry.getValue().getZkConnectStr();
-      HelixManager helixManager =
-          helixFactory.getZKHelixManager(cloudConfig.vcrClusterName, selfInstanceName, InstanceType.SPECTATOR,
-              zkConnectStr);
-      helixManager.connect();
+    for (DcZkInfo dcZkInfo: dataCenterToZkAddress.values()) {
+      // only handle vcr clusters for now
+      if (dcZkInfo.getReplicaType() == ReplicaType.CLOUD_BACKED) {
```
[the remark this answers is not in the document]
cgtz: I think it worked based on the assumption that the dczkinfo config would be set to only include the VCR cluster and no native clusters, which is fine for recovery, but not live replication.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudChangeListener.java
```diff
@@ -0,0 +1,66 @@
...
+  @Override
+  public void onReplicaAddedOrRemoved(Collection<? extends ReplicaId> addedReplicas,
+      Collection<? extends ReplicaId> removedReplicas) {
+    // This impl assumes that no partition will ever be completely removed.
+    for (ReplicaId replica : addedReplicas) {
+      AmbryPartition partition = (AmbryPartition) replica.getPartitionId();
+      String partitionName = replica.getPartitionId().toPathString();
+      if (knownPartitions.add(partitionName)) {
+        try {
+          logger.debug("Adding cloud replica: dc={}, partition={}", dcName, partitionName);
```
jsjtzyy: I think `INFO` level would be better here.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -663,6 +592,9 @@ void addReplicasToPartition(AmbryPartition partition, List<AmbryReplica> replica
         return v;
       });
       clusterWideAllocatedRawCapacityBytes.getAndAdd(replicas.get(0).getCapacityInBytes() * replicas.size());
+      for (ClusterMapChangeListener listener : clusterMapChangeListeners) {
+        listener.onReplicaAddedOrRemoved(replicas, Collections.emptyList());
```
jsjtzyy: After rebasing master branch, we should be able to remove this. In #1405, clustermap change listeners are called within DynamicClusterChangeHandler. If you have additional concern, we can talk offline.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudChangeListener.java
```diff
@@ -0,0 +1,66 @@
...
+          // Copy the capacity from an existing replica
+          AmbryReplica cloudAmbryReplica =
+              new CloudAmbryReplica(clusterMapConfig, partition, replica.getCapacityInBytes());
+          changeHandlerCallback.addReplicasToPartition(partition, Collections.singletonList(cloudAmbryReplica));
```
jsjtzyy: After adding CloudAmbryReplica to AmbryPartition, when `getReplicaIdsForPartition` is called, is there any check on if the replica is `DiskAmbryReplica` or `CloudAmbryReplica` ?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -870,4 +802,132 @@ long getAllocatedUsableCapacity() {
...
+     */
+    private void initializeCloudDatacenter() {
+      DcInfo dcInfo = new DcInfo(dcName, dcZkInfo, null, null);
+      dcToDcInfo.put(dcName, dcInfo);
+      dcIdToDcName.put(dcZkInfo.getDcId(), dcName);
+      CloudChangeListener listener =
+          new CloudChangeListener(dcName, clusterMapConfig, clusterChangeHandlerCallback);
+      registerClusterMapListener(listener);
+      // seed existing replicas.
+      for (Set<AmbryReplica> replicas : ambryPartitionToAmbryReplicas.values()) {
```
jsjtzyy: I don't quite understand this part. Will talk with you offline.

cgtz: This logic will have to change a bit since I am rebasing on your changes to listening. However, the main goal is to add any ambry partitions that already exist.

I am roughly trying to accomplish what ReplicationManager does where it first adds the listener, and then iterates through all existing replicas right afterwards so that it knows about all replicas added before the listener was added.

I may just make a new listener for just newly added partitions for this use case since I envision it would be somewhat temporary, as eventually ClusterMap will get the set of managed partitions from the VCR.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
cgtz: One concern now is that we can't add a listener from the background startup thread since the DcInfo object for other datacenters may not be present yet. Here I can either add the listener after all of the startup threads are joined, or make a new type of listener that is notified on new partitions added.

ambry-api/src/main/java/com.github.ambry/config/ClusterMapConfig.java
```diff
@@ -111,10 +111,15 @@
    *       "datacenter":"dc2",
    *       "id" : "2",
    *       "zkConnectStr":"def.example.com:2300",
+   *     },
+   *     {
+   *       "datacenter":"cloud-dc",
+   *       "id" : "3",
+   *       "replicaType": "CLOUD_BACKED"
```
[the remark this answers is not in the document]
cgtz: Here I wanted to leave the option open to have `HelixClusterManager` connect to VCRs in the future, where the replicaType will be CLOUD_BACKED and the zookeeper connect string will be set.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudAmbryReplica.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+package com.github.ambry.clustermap;
+
+import com.github.ambry.config.ClusterMapConfig;
+import org.json.JSONObject;
+
+import static com.github.ambry.clustermap.ClusterMapSnapshotConstants.*;
+
+
+class CloudAmbryReplica extends AmbryReplica {
```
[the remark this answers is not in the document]
cgtz: I changed the name to hopefully be more descriptive of the difference too.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudChangeListener.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Currently, this just adds a single {@link ReplicaType#CLOUD_BACKED} replica every time a partition is added to allow
```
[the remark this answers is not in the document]
cgtz: No, will fix.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudAmbryReplica.java
```diff
@@ -0,0 +1,90 @@
...
+  public String toString() {
+    return "Replica[cloud:" + getPartitionId().toPathString() + "]";
+  }
+
+  @Override
+  public void markDiskDown() {
+  }
+
+  @Override
+  public void markDiskUp() {
```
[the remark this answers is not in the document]
cgtz: Changed it to throw UnsupportedOperationException. Before I was unsure since i didn't know how many callers it had, but all of them are specific to AmbryServer/BlobStore, so its ok to do so.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -870,4 +802,132 @@ long getAllocatedUsableCapacity() {
       return clusterWideAllocatedUsableCapacityBytes.get();
     }
   }
+
+  private class DatacenterInitializer implements Runnable {
```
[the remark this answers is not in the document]
cgtz: Moved it out into its own class. Unfortunately, I had to pass a lot of variables in to make it work, but this can probably be refactored a bit further in the future when SimpleClusterChangeHandler is removed

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudChangeListener.java
cgtz: Setting it to info would log for every partition in the cluster on startup. Is that okay, or does it seem too noisy?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
cgtz: I chose to connect the listeners after the other startup threads are finished.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudChangeListener.java
cgtz: There are no such checks right now. Do you know of any places where including cloud replicas in the list of partitions will break something in the frontend? This feature should not be used for the server (until VCR support is integrated)

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudChangeListener.java
cgtz: Changed it to CloudServiceClusterChangeHandler, since it now implements that interface as well.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -134,6 +134,7 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
       }
 
       for (DcInfo dcInfo : dcToDcInfo.values()) {
+        System.out.println("AKAKAKA" + dcInfo.clusterChangeHandler);
```
lightningrob: ?

ambry-clustermap/src/test/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandlerTest.java
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private List<ReplicaId> makeReplicas(AmbryPartition... partitionIds) throws Exception {
+    List<ReplicaId> replicaIds = new ArrayList<>(partitionIds.length);
+    for (AmbryPartition partitionId : partitionIds) {
+      replicaIds.add(new AmbryServerReplica(config, partitionId, mock(AmbryDisk.class), false, replicaCapacity, false));
+    }
+    return replicaIds;
+  }
+}
```
lightningrob: newline

ambry-clustermap/src/test/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandlerTest.java
```diff
@@ -0,0 +1,98 @@
...
+    assertEquals(Collections.emptyList(), handler.getReplicaIds(mock(AmbryDataNode.class)));
+    assertEquals(Collections.emptySet(), handler.getDisks(cloudServiceDataNode));
+    assertEquals(Collections.emptyMap(), handler.getPartitionToResourceMap());
+    assertEquals(0, handler.getErrorCount());
+
+    // simulate failure
+    doThrow(new RuntimeException()).when(clusterChangeHandlerCallback).addReplicasToPartition(any(), any());
+    handler.onReplicaAddedOrRemoved(makeReplicas(p5), makeReplicas());
+    assertEquals(1, handler.getErrorCount());
+
```
lightningrob: remove blank line

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectator.java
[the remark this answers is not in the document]
lightningrob: According to codecov this code is never tested (though presumably does run).

ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterChangeHandler.java
```diff
@@ -10,42 +10,36 @@
...
-
-  /**
-   * @return current snapshot held by this {@link ClusterChangeHandler}.
+   * Get replicas of given partition from this datacenter that are in required state
+   * @param partition the {@link PartitionId} for which to get the list of replicas.
+   * @param state {@link ReplicaState} associated with replica
+   * @return the {@link ReplicaId}s satisfying requirements.
    */
-  RoutingTableSnapshot getRoutingTableSnapshot();
+  Stream<AmbryReplica> getReplicaIdsByState(AmbryPartition partition, ReplicaState state);
```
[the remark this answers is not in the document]
cgtz: Incidentally, most of the methods in ClusterChangeHandler are about getting a view of the datanodes, and replicas in that datacenter. This was the case before this change. I added this method to make the interface more general to things that do not have a helix `RoutingTableProvider`.

Perhaps a better name would be `DatacenterView` for this interface, but I would probably prefer changing the name in a separate PR since this one is already big.

ambry-api/src/main/java/com.github.ambry/router/CryptoService.java
```diff
@@ -15,8 +15,6 @@
 
 import com.github.ambry.utils.Utils;
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufAllocator;
-import io.netty.buffer.Unpooled;
```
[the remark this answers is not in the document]
cgtz: Will do in the future. I'm not sure why I even did this cleanup on this branch.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryDatanode.java
```diff
@@ -30,53 +26,35 @@
 /**
  * {@link DataNodeId} implementation to use within dynamic cluster managers.
  */
-class AmbryDataNode implements DataNodeId {
+abstract class AmbryDataNode implements DataNodeId {
   private final String hostName;
-  private final Port plainTextPort;
-  private final Port sslPort;
-  private final Port http2Port;
+  // exposed for subclass access
```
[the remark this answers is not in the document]
cgtz: Well, protected provides more access than package private, so I went with the modifier that gave the least access needed (no access needed for extenders outside of the package).

Modifier | Class | Package | Subclass | World
-- | -- | -- | -- | --
public | Y | Y | Y | Y
protected | Y | Y | Y | N
no modifier | Y | Y | N | N
private | Y | N | N | N

But I will make them protected, since it probably makes it clearer what the intent is.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandler.java
```diff
@@ -0,0 +1,139 @@
...
+ * for requests to be routed to cloud services. Currently it is only registered as a {@link ClusterMapChangeListener},
+ * but once the clustermap natively supports VCR clusters, this class can implement {@link HelixAwareClusterChangeHandler} and
+ * listen for changes in the VCR cluster.
+ */
+class CloudServiceClusterChangeHandler implements ClusterMapChangeListener, ClusterChangeHandler {
+  private static final Logger logger = LoggerFactory.getLogger(CloudServiceClusterChangeHandler.class);
+  private final Set<String> knownPartitions = ConcurrentHashMap.newKeySet();
+  private final ConcurrentHashMap<String, CloudServiceReplica> partitionToReplica = new ConcurrentHashMap<>();
+  private final List<ClusterMapChangeListener> listeners = new CopyOnWriteArrayList<>();
+  private final AtomicLong errorCount = new AtomicLong(0);
```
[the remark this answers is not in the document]
cgtz: Yes, this method is a bit weird to me, but apparently the method isn't used for codahale metrics, but for HelixBootstrapUpgradeTool. Maybe I can get rid of it from the interface in the future.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DatacenterInitializer.java
```diff
@@ -0,0 +1,248 @@
...
+  private final String selfInstanceName;
+  private final Map<String, Map<String, String>> partitionOverrideInfoMap;
+  private final HelixClusterManager.ClusterChangeHandlerCallback clusterChangeHandlerCallback;
+  private final HelixClusterManager.HelixClusterManagerCallback helixClusterManagerCallback;
+  private final HelixClusterManagerMetrics helixClusterManagerMetrics;
+  private final AtomicLong sealedStateChangeCounter;
+  // Fields to pass into only SimpleClusterChangeHandler (These can be removed if SimpleClusterChangeHandler is removed)
+  private final ConcurrentHashMap<ByteBuffer, AmbryPartition> partitionMap;
+  private final ConcurrentHashMap<String, AmbryPartition> partitionNameToAmbryPartition;
+  private final ConcurrentHashMap<AmbryPartition, Set<AmbryReplica>> ambryPartitionToAmbryReplicas;
```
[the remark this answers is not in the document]
cgtz: The child class constructors require `ConcurrentHashMap`s. I believe it was done this way to indicate that the implementations will only work correctly if the instances are `ConcurrentHashMap`, due to their usage of various CAS-style methods where atomicity is expected. @jsjtzyy may be able to explain more.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplica.java
```diff
@@ -15,84 +15,58 @@
...
-class AmbryReplica implements ReplicaId {
+abstract class AmbryReplica implements ReplicaId {
   private final AmbryPartition partition;
-  private final AmbryDisk disk;
   private final long capacityBytes;
   private volatile boolean isSealed;
-  private volatile boolean isStopped;
-  private final ResourceStatePolicy resourceStatePolicy;
+  volatile boolean isStopped;
+  final ResourceStatePolicy resourceStatePolicy;
```
jsjtzyy: I wonder if `onReplicaUnavailable()` and `onReplicaResponse()` in this class are really needed for `CloudServiceReplica`. If not, we can move it into `AmbryServerReplica`.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceReplica.java
```diff
@@ -0,0 +1,104 @@
...
+
+import static com.github.ambry.clustermap.ClusterMapSnapshotConstants.*;
+
+
+/**
+ * An implementation of {@link AmbryReplica} that is meant to route requests towards cloud destinations using a standard
+ * client library. The difference between this and {@link CloudReplica} is that {@link CloudReplica} is meant to
+ * represent a VCR host instead of just being used to route requests towards a local client library. Additionally,
+ * implementing {@link AmbryReplica} currently allows for easier interop with the rest of the helix cluster management
+ * codebase.
```
jsjtzyy: Thanks for explaining the difference. So, `ClouldServiceReplica` is used by frontend for now and primarily used to route requests to cloud, right? For server, it interacts with `CloudReplica` to perform two-way replication. (Correct me if I have misunderstood)

jsjtzyy: Since `CloudServiceReplica` will be in HelixClusterManager, so server is able to see it. I guess there is no server logic directly depends on `CloudServiceReplica`.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceDataNode.java
```diff
@@ -0,0 +1,69 @@
...
+ * This is a place for {@link CloudServiceReplica}s to live on. This represents a location for in-process managed
+ * service clients (not a VCR like {@link CloudDataNode}). This will not contain any ports, as it is not meant to
+ * represent a remote network location. The hostname will just be the datacenter name. For logging purposes, the
+ * plaintext port will show as -1.
+ */
+class CloudServiceDataNode extends AmbryDataNode {
+
+  /**
+   * Instantiate a {@link CloudServiceDataNode}.
+   * @param dataCenterName the name of the dataCenter associated with this data node.
```
jsjtzyy: Could this dataCenter be the same with regular datacenter (where regular ambry data nodes sit)? Or this should be a special datacenter representing the cloud?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandler.java
```diff
@@ -0,0 +1,140 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.stream.Stream;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Currently, this just adds a single {@link ReplicaType#CLOUD_BACKED} replica every time a partition is added to allow
+ * for requests to be routed to cloud services. Currently it is only registered as a {@link ClusterMapChangeListener},
+ * but once the clustermap natively supports VCR clusters, this class can implement {@link HelixAwareClusterChangeHandler} and
```
jsjtzyy: Is there a doc briefly describing how to support VCR clusters in clustermap? I am interested in this part and probably can work on some of its implementation. Can I envision that clustermap will incorporate both CloudReplica (for server's two-way replication) and CloudServiceReplica (for router to issue request)?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandler.java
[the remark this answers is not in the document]
jsjtzyy: This is used by Validating Cluster Manager in HelixBootstrapTool to verify boostrap/upgrade succeeds.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectator.java
```diff
@@ -63,15 +63,17 @@ public void spectate() throws Exception {
...
+        HelixManager helixManager =
+            helixFactory.getZKHelixManager(cloudConfig.vcrClusterName, selfInstanceName, InstanceType.SPECTATOR,
+                dcZkInfo.getZkConnectStr());
```
jsjtzyy: Does `dcZkInfo.getZkConnectStr()` return empty string for now?

ambry-clustermap/src/test/java/com.github.ambry.clustermap/MockHelixCluster.java
```diff
@@ -158,6 +158,11 @@ void setReplicaState(AmbryPartition partition, String instance, TestUtils.Replic
     for (MockHelixAdmin helixAdmin : helixAdmins.values()) {
       upInstances.addAll(helixAdmin.getUpInstances());
     }
+    // add cloud datanodes
+    dataCenterToZkAddress.values()
+        .stream()
+        .filter(info -> info.getReplicaType() == ReplicaType.CLOUD_BACKED)
+        .forEach(info -> upInstances.add(ClusterMapUtils.getInstanceName(info.getDcName(), -1)));
```
jsjtzyy: Suggest using a constant for `-1` and put it into somewhere like TestUtil that can be shared by all tests.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -234,30 +172,35 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
...
    */
   private HelixManager initializeHelixManagerAndPropertyStoreInLocalDC(Map<String, DcZkInfo> dataCenterToZkAddress,
       String instanceName, HelixFactory helixFactory) throws Exception {
     DcZkInfo dcZkInfo = dataCenterToZkAddress.get(clusterMapConfig.clusterMapDatacenterName);
     String zkConnectStr = dcZkInfo.getZkConnectStr();
-    HelixManager manager;
-    manager = helixFactory.getZKHelixManager(clusterName, instanceName, InstanceType.SPECTATOR, zkConnectStr);
+    if (dcZkInfo.getReplicaType() == ReplicaType.CLOUD_BACKED) {
+      return null;
+    }
```
jsjtzyy: I am not quite sure returning null for cloud dc is reasonable here.  For now, it makes some sense but in the future we need to reconsider. 
Let's say a frontend is brought up in cloud dc and this method should populate `partitionOverrideInfoMap` for the frontend. With current logic, `partitionOverrideInfoMap` is empty and frontend may route PUT to those partitions which should be temporarily sealed.
If frontend in cloud will be initialized with different code path, then just ignore this comment.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryDatanode.java
```diff
@@ -95,7 +72,7 @@ public int getSSLPort() {
       return sslPort.getPort();
     } else {
       throw new IllegalStateException(
-          "No HTTP2 port exists for the Data Node " + hostName + ":" + plainTextPort.getPort());
+          "No HTTP2 port exists for the Data Node " + getHostname() + ":" + plainTextPort.getPort());
```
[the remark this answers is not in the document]
cgtz: changed back

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryReplica.java
cgtz: I suppose they are at least needed as abstract methods in AmbryReplica since `onReplicaEvent` calls them and it would be unclean IMO to have `onReplicaEvent` have to check the replica type and cast before calling the method.

Whether the `ResourceStatePolicy` is needed for cloud replicas is another question. One could just make the CloudServiceReplica always up and not have any failure tracking. For our preliminary testing environment I am inclined to leave it in and let the CloudServiceReplica go "down" when faced with service unavailability and see how it behaves. We definitely would need to take a second look at this though.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceReplica.java
cgtz: Yes it is for frontends, but it would only be present for the nodes that are configured to see it, using `clusterMapDcsZkConnectStrings`. Frontends can be configured to include it, and servers could be configured without it.

Yes a `CloudReplica` currently represents a replica that is present on a VCR host. In the long run, I would hope to also fold this one into the regular clustermap as a `VcrReplica`. I think that would help to clean up our 2-way replication story and allow us to share more code between the many ReplicationEngine variants that currently exist.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceDataNode.java
cgtz: No, a cloud DC is currently modeled as a separate datacenter.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandler.java
cgtz: No, there is no up to date doc, but I think our end goal should be that ClusterMap provides full support for spectating vcr and server state such that we can unify `CloudToStoreReplicationManager` and `ReplicationManager` (since an ambry-server will soon have to be able to replicate from both servers and vcrs simultaneously).

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandler.java
```diff
@@ -0,0 +1,140 @@
...
+  private final String cloudServiceInstanceName;
+
+  public CloudServiceClusterChangeHandler(String dcName, ClusterMapConfig clusterMapConfig,
+      ClusterChangeHandlerCallback changeHandlerCallback) throws Exception {
+    this.dcName = dcName;
+    this.clusterMapConfig = clusterMapConfig;
+    this.changeHandlerCallback = changeHandlerCallback;
+    cloudServiceDataNode = new CloudServiceDataNode(dcName, clusterMapConfig);
+    cloudServiceInstanceName =
+        ClusterMapUtils.getInstanceName(cloudServiceDataNode.getDatacenterName(), cloudServiceDataNode.getPort());
```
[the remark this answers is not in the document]
cgtz: I tried this. The problem I ran into is that the method `ClusterMap.getDataNodeId(...)` requires an int port number to be passed in and generates the instance name before calling `ClusterChangeHandler.getDataNode`.

The options would be:
1. Keep it as it is with a -1 port for CloudServiceDataNodes
2. Don't support `getDataNode` in `CloudServiceClusterChangeHandler`. This could be a possibility since it probably has very limited value in this case.
3. Change `ClusterMap.getDataNodeId` method to take a nullable `Integer` instead of `int`

Which do you prefer, @jsjtzyy?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
cgtz: Where should a frontend in a datacenter without a helix ambry-server cluster get partitionOverrideInfo from? Some other datacenter? VCR cluster?

I think that not having partition override is not a blocker for to start testing so I would prefer to keep this as is. In the future, I feel that storing partition override info in the VCR cluster would be the right way forward, since configuring an alternate zk to get this from seems like a hack. However, that would require the extra dev effort to support spectating VCR clusters described in some of the other comments.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterSpectator.java
cgtz: No, this would be a case where it would be set to the vcr cluster address. Hopefully, we can find a way to unify `HelixClusterSpectator` with the rest of the `ClusterMap` eventually.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -106,105 +101,48 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
...
+        }
+      }
+
+      for (DcInfo dcInfo : dcToDcInfo.values()) {
+        if (dcInfo.clusterChangeHandler instanceof ClusterMapChangeListener) {
+          ClusterMapChangeListener listener = (ClusterMapChangeListener) dcInfo.clusterChangeHandler;
+          registerClusterMapListener(listener);
+          List<ReplicaId> allReplicas = new ArrayList<>();
+          for (Set<AmbryReplica> replicas : ambryPartitionToAmbryReplicas.values()) {
+            allReplicas.addAll(replicas);
```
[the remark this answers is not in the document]
cgtz: That is true. I will do that! However, we should be careful in making it clear that this is specifically tailored to `CloudServiceClusterChangeHandler, since in most cases one would expect all replicas to be provided.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -234,30 +172,35 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
...
     }
-    localDatacenterId = dcToDcZkInfo.get(clusterMapConfig.clusterMapDatacenterName).dcZkInfo.getDcId();
+    localDatacenterId = dcToDcInfo.get(clusterMapConfig.clusterMapDatacenterName).dcZkInfo.getDcId();
     partitionSelectionHelper =
         new PartitionSelectionHelper(helixClusterManagerCallback, clusterMapConfig.clusterMapDatacenterName,
             clusterMapConfig.clustermapWritablePartitionMinReplicaCount);
     // register partition selection helper as a listener of cluster map changes.
-    dcToDcZkInfo.values()
-        .forEach(info -> info.clusterChangeHandler.registerClusterMapListener(partitionSelectionHelper));
+    registerClusterMapListener(partitionSelectionHelper);
```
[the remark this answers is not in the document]
cgtz: I feel like it would, otherwise there would be no way to make a put request to newly added partitions in a cloud datacenter since the partitionSelectionHelper will not be fed CloudServiceReplicas as new partitions get added? Please correct me if my understanding is incorrect.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandler.java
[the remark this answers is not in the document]
jsjtzyy: If `getDataNode` is unlikely to be used in `CloudServiceClusterChangeHandler`, yeah, we don't have to support it. (Let's go with option 2, what do you think?)

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -106,105 +101,48 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
...
-            initializationAttemptComplete.countDown();
-          }
-        }, false).start();
+        DatacenterInitializer initializer =
+            new DatacenterInitializer(clusterMapConfig, localManager, helixFactory, dcZkInfo, selfInstanceName,
+                partitionOverrideInfoMap, clusterChangeHandlerCallback, helixClusterManagerCallback,
+                helixClusterManagerMetrics, sealedStateChangeCounter, partitionMap, partitionNameToAmbryPartition,
+                ambryPartitionToAmbryReplicas);
+        initializer.start();
+        initializers.add(initializer);
```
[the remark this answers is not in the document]
cgtz: Yes, that is right.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
jsjtzyy: Correct, VCR cluster is the right place to store partition override info.  (I agree this is not a blocker now, we can keep this in mind if we are doing cluster expansion in the future. Previously, I just want to mention a case where some partitions are administratively sealed in linkedin dc. However, if frontends in cloud still believe them unsealed, the PUT requests are still able to write into these partitions via cloud frontend)

ambry-clustermap/src/main/java/com.github.ambry.clustermap/ClusterMapUtils.java
```diff
@@ -141,8 +149,12 @@ public static String getInstanceName(String host, Integer port) {
     JSONArray all = root.getJSONArray(ZKINFO_STR);
     for (int i = 0; i < all.length(); i++) {
       JSONObject entry = all.getJSONObject(i);
+      String name = entry.getString(DATACENTER_STR);
       byte id = (byte) entry.getInt(DATACENTER_ID_STR);
-      DcZkInfo dcZkInfo = new DcZkInfo(entry.getString(DATACENTER_STR), id, entry.getString(ZKCONNECTSTR_STR));
+      ReplicaType replicaType = entry.optEnum(ReplicaType.class, REPLICA_TYPE_STR, ReplicaType.DISK_BACKED);
```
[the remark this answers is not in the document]
cgtz: That would break backwards compatibility with existing configs by adding a new required field, replicaType. Personally, I don't find having default values too complicated. I also wanted to resist having a dummy value for cloud_backed datacenters, since the work I'm doing here is admittedly a little transient. Eventually, the zk string for cloud dcs will likely be set to the VCR zk for things like getting partition override info and additional metadata.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/CloudServiceClusterChangeHandler.java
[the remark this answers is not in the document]
cgtz: implemented 2

Issue #1414, opened 2020-03-03
NullPointerException when removing replica from thread in VCR cluster


jsjtzyy:
java.lang.NullPointerException
```
com.github.ambry.replication.ReplicationEngine.removeRemoteReplicaInfoFromReplicaThread - Line 215 (ReplicationEngine.java)
com.github.ambry.replication.ReplicationEngine.stopPartitionReplication - Line 449 (ReplicationEngine.java)
com.github.ambry.cloud.VcrReplicationManager.removeReplica - Line 206 (VcrReplicationManager.java)
com.github.ambry.cloud.VcrReplicationManager$1.onPartitionRemoved - Line 111 (VcrReplicationManager.java)
com.github.ambry.cloud.HelixVcrCluster.removePartition - Line 117 (HelixVcrCluster.java)
com.github.ambry.cloud.HelixVcrStateModel.onBecomeStandbyFromLeader - Line 55 (HelixVcrStateModel.java)
sun.reflect.GeneratedMethodAccessor595.invoke - Line -1 ()
sun.reflect.DelegatingMethodAccessorImpl.invoke - Line 43 (DelegatingMethodAccessorImpl.java)
java.lang.reflect.Method.invoke - Line 498 (Method.java)
org.apache.helix.messaging.handling.HelixStateTransitionHandler.invoke - Line 430 (HelixStateTransitionHandler.java)
org.apache.helix.messaging.handling.HelixStateTransitionHandler.handleMessage - Line 344 (HelixStateTransitionHandler.java)
org.apache.helix.messaging.handling.HelixTask.call - Line 96 (HelixTask.java)
org.apache.helix.messaging.handling.HelixTask.call - Line 48 (HelixTask.java)
java.util.concurrent.FutureTask.run - Line 266 (FutureTask.java)
java.util.concurrent.ThreadPoolExecutor.runWorker - Line 1149 (ThreadPoolExecutor.java)
java.util.concurrent.ThreadPoolExecutor$Worker.run - Line 624 (ThreadPoolExecutor.java)
java.lang.Thread.run - Line 748 (Thread.java)
```

jsjtzyy:
Closing this issue as it is fixed in #1417

PR #1417, opened 2020-03-05
Minor changes in CloudToStoreReplicationManager for move replica

jsjtzyy:
CloudToStoreReplicationManager is not really affected by "move replica" except for the local partition map. This PR removes local partition map and makes it fetch the partition from storage manager which maintains an up-to-date list of current local partitions.
Also, this PR fixes NPE issue #1414 .

### DIFF
--- a/ambry-replication/src/test/java/com.github.ambry.replication/CloudToStoreReplicationManagerTest.java
+++ b/ambry-replication/src/test/java/com.github.ambry.replication/CloudToStoreReplicationManagerTest.java
@@ -1,0 +1,210 @@
+/**
+ * Copyright 2020 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.replication;
+
+import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.clustermap.MockClusterMap;
+import com.github.ambry.clustermap.MockClusterSpectator;
+import com.github.ambry.clustermap.MockDataNodeId;
+import com.github.ambry.clustermap.MockHelixParticipant;
+import com.github.ambry.clustermap.MockPartitionId;
+import com.github.ambry.clustermap.PartitionId;
+import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.config.DiskManagerConfig;
+import com.github.ambry.config.ReplicationConfig;
+import com.github.ambry.config.ServerConfig;
+import com.github.ambry.config.StoreConfig;
+import com.github.ambry.config.VerifiableProperties;
+import com.github.ambry.network.Port;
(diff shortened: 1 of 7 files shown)


ambry-replication/src/test/java/com.github.ambry.replication/CloudToStoreReplicationManagerTest.java
```diff
@@ -0,0 +1,210 @@
...
+            storeKeyFactory, clusterMap, mockScheduler, currentNode, null, clusterMap.getMetricRegistry(), null,
+            storeKeyConverterFactory, serverConfig.serverMessageTransformer, mockClusterSpectator,
+            mockHelixParticipant);
+    storageManager.start();
+    cloudToStoreReplicationManager.start();
+    mockClusterSpectator.spectate();
+    // 1. test adding cloud replica that is not present locally
+    mockHelixParticipant.onPartitionBecomeLeaderFromStandby(NEW_PARTITION_NAME);
+    assertNull("Thread pool for DC1 should be null",
+        cloudToStoreReplicationManager.getReplicaThreadPoolByDc().get("DC1"));
```
lightningrob: I don't think we should expose an internal thread pool just for the test to check if an operation succeeded.  Perhaps we can expose a more specific method like hasStoreForDatacenter() in the ReplicationEngine base class?

ambry-replication/src/test/java/com.github.ambry.replication/CloudToStoreReplicationManagerTest.java
```diff
@@ -0,0 +1,207 @@
...
+
+/**
+ * Tests for {@link CloudToStoreReplicationManager} when adding/removing cloud replica. The tests also cover the case
+ * where new replica is added due to "move replica".
+ */
+public class CloudToStoreReplicationManagerTest {
+  private static final String NEW_PARTITION_NAME = "12";
+  private static final String CLOUD_DC_NAME = "CloudDc";
+  private static final String VCR_MOUNT_PATH = "/vcr/1";
+  private static final String VCR_REPLICA_THREAD_PREFIX = "VcrReplicaThread-";
```
lightningrob: I was wondering where this hardcoded constant was coming from and found this horrible line of code in ReplicationEngine.createThreadPool():
      String threadIdentity =  (startThread ? "Vcr" : "") + "ReplicaThread-" + ...
I'm adding a TODO for myself to fix that and this constant as well.  It's okay for now.

ambry-replication/src/test/java/com.github.ambry.replication/CloudToStoreReplicationManagerTest.java
```diff
@@ -139,8 +140,9 @@ public void cloudReplicaAdditionTest() throws Exception {
     mockClusterSpectator.spectate();
     // 1. test adding cloud replica that is not present locally
     mockHelixParticipant.onPartitionBecomeLeaderFromStandby(NEW_PARTITION_NAME);
-    assertNull("Thread pool for DC1 should be null",
-        cloudToStoreReplicationManager.getReplicaThreadPoolByDc().get("DC1"));
+    assertNull("Cloud replica thread should not be created", TestUtils.getThreadByThisName(VCR_REPLICA_THREAD_PREFIX));
+    //assertNull("Thread pool for DC1 should be null",
+    //cloudToStoreReplicationManager.getReplicaThreadPoolByDc().get("DC1"));
```
lightningrob: Remove these lines?

zzmao:
looks like there is a conflict.

ambry-replication/src/test/java/com.github.ambry.replication/CloudToStoreReplicationManagerTest.java
jsjtzyy: yeah, we probably need to fix it in ReplicationEngine as well.

PR #1419, opened 2020-03-08
Support netty bytebuf in PutOperation

justinlin-linkedin:
This PR starts to support netty bytebuf in the PutOperation. There are some benefits out of it.
1. We don't have to copy the content from http channel to a java bytebuffer 
2. We can avoid a memory allocation for java bytebuffer.

### DIFF
--- a/ambry-api/src/main/java/com.github.ambry/network/Send.java
+++ b/ambry-api/src/main/java/com.github.ambry/network/Send.java
@@ -52,7 +52,7 @@ public interface Send {
   long sizeInBytes();
 
   /**
-   * Release all the resource this object holds.
+   * Release all the resource this object holds. Make this a default method so subclasses don't have to override it.
    */
   default void release() {}
 }
--- a/ambry-api/src/main/java/com.github.ambry/network/Send.java
+++ b/ambry-api/src/main/java/com.github.ambry/network/Send.java
@@ -50,4 +50,9 @@ public interface Send {
    * @return The size of the data in bytes to be written
    */
   long sizeInBytes();
+
+  /**
+   * Release all the resource this object holds.
+   */
+  default void release() {}
 }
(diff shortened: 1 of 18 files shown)


ambry-api/src/main/java/com.github.ambry/network/Send.java
```diff
@@ -50,4 +50,9 @@ default void writeTo(AsyncWritableChannel channel, Callback<Long> callback) {
    * @return The size of the data in bytes to be written
    */
   long sizeInBytes();
+
+  /**
+   * Release all the resource this object holds.
+   */
+  default void release() {}
```
zzmao: Minor: throw UnSupported Exception in case something wrong is released?

ambry-protocol/src/main/java/com.github.ambry.protocol/PutRequest.java
```diff
@@ -225,8 +225,14 @@ public long writeTo(WritableByteChannel channel) throws IOException {
             ByteBuffer byteBuffer = nioBuffersFromBlob[bufferIndex];
             currentWritten = channel.write(byteBuffer);
             totalWrittenBytesForBlob += currentWritten;
-            if (!byteBuffer.hasRemaining()) {
+            while (!byteBuffer.hasRemaining()) {
```
cgtz: Would something like this work to eliminate the nested while loop?
```java
          while (bufferIndex < nioBuffersFromBlob.length && currentWritten != 0) {
            currentWritten = -1;
            ByteBuffer byteBuffer = nioBuffersFromBlob[bufferIndex];
            if (!byteBuffer.hasRemaining()) {
              bufferIndex++;
            } else {
              currentWritten = channel.write(byteBuffer);
              totalWrittenBytesForBlob += currentWritten;
            }
          }
```

PR #1420, opened 2020-03-09
Refactor test utils into ambry-test-utils subproj

cgtz:
There are many test utilities that are shared across many subprojects
and even dependended on when writing tests for wrappers that extend open
source Ambry (like AmbryLI for LinkedIn). Generally, depending on test
artifacts across projects is not a great practice (issues handling
transitive dependencies with non-standard classifiers), so this project
moves the test utilities used across repos into an ambry-test-utils
subproject.

For now, this targets moving only the test dependencies that are
depended on by LinkedIn internal projects to ambry-test-utils. Going
forward, we can simplify our test dependencies within the open source
repo by moving more classes into ambry-test-utils.

For tests that still depend on other subprojects' tests, this commit
switches from using source sets to a configuration based approach. This
ensures that test classes are always built after their dependencies are
compiled and ensures that transitive dependencies are placed on the 
classpath at runtime. 
See https://softnoise.wordpress.com/2014/09/07/gradle-sub-project-test-dependencies-in-multi-project-builds/

*Reviewer notes:*
* To test this, I published to ~/local-repo using `./gradlew -Dmaven.repo.local=$HOME/local-repo publishToMavenLocal`. 
  I then added the ambry libraries as external dependencies in AmbryLI. See https://rb.corp.linkedin.com/r/2003587/ for reference.
* Most of the actual changes in the diff are either from the new way of declaring test dependencies
  in build.gradle or from moving the `getRandomString` method from `UtilsTest` to `TestUtils`.


ambry-test-utils/src/main/java/com/github/ambry/router/InMemoryRouter.java
```diff
@@ -391,195 +398,175 @@ protected static void completeOperation(FutureResult futureResult, Callback call
       callback.onCompletion(operationResult, exception);
     }
   }
-}
-
-/**
- * Thread to read the post data async and store it.
- */
-class InMemoryBlobPoster implements Runnable {
```
cgtz: The diff looks weird in the class, but the change I made is to change the package-private classes below InMemoryRouter in this file into private static inner classes, since they are not used anywhere outside of InMemoryRouter

ambry-clustermap/src/test/java/com.github.ambry.clustermap/ClusterMapUtilsTest.java
```diff
@@ -26,6 +25,7 @@
 import org.junit.Test;
 import org.mockito.Mockito;
 
+import static com.github.ambry.utils.TestUtils.*;
```
lightningrob: Why static import in this class?

ambry-test-utils/src/main/java/com/github/ambry/router/InMemoryRouter.java
```diff
@@ -376,6 +375,14 @@ private boolean handlePrechecks(FutureResult futureResult, Callback callback) {
     return continueOp;
   }
 
+  private void checkBlobId(String blobId) throws RouterException {
```
lightningrob: javadoc

ambry-test-utils/src/main/java/com/github/ambry/utils/HelixControllerManager.java
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
lightningrob: Is this needed, or even correct?  Nearly all our javadocs have the two star convention.

ambry-test-utils/src/main/java/com/github/ambry/utils/TestUtils.java
```diff
@@ -232,6 +233,14 @@ public static String getRandomKey(int size) {
     return sb.toString();
   }
 
+  public static String getRandomString(int length) {
```
lightningrob: javadoc

ambry-test-utils/src/main/java/com/github/ambry/utils/HelixControllerManager.java
cgtz: The single star is more correct according to intellij since the license comment is not a javadoc, just a block comment. When the two stars are present it gives a warning "Dangling javadoc comment".

ambry-clustermap/src/test/java/com.github.ambry.clustermap/ClusterMapUtilsTest.java
cgtz: For the tests in the clustermap package, there is another class called `com.github.ambry.clustermap.TestUtils`. To avoid having to write out the fully qualified classname every time getRandomString is used, I did a static import. I suppose the best thing to do would be to change the other package's name to `ClustermapTestUtils`, but I didn't want to do that in the same PR to prevent this one from getting even bigger.

PR #1421, opened 2020-03-09
Netty pr6 for review

zzmao:
Implemented Http2NetworkClient and other http2 client related class.
Removed ClusterMapHttp2NetworkClientEndbaled, because replication http2 request is not supported.
Added RouterEnableHttp2NetworkClient, to limit http2 request in router only, for now.


zzmao:
@jsjtzyy  You are optional reviewer.

zzmao:
There is an interesting problem when I tested the code: 

Client cert should be created before Server, otherwise, client doesn't send cert to server in SSL handshake.

For example, change the router/server properties creation order, SSLHandshake failure happen with the error that peer(client) can't provide cert.
```
    // Set up router properties
    Properties routerProps = getRouterProperties("DC1");
    TestSSLUtils.addSSLProperties(routerProps, "DC1,DC2,DC3", SSLFactory.Mode.CLIENT, trustStoreFile, "http2-router");
    TestSSLUtils.addHttp2Properties(routerProps, SSLFactory.Mode.CLIENT, false);
    routerProps.setProperty(RouterConfig.ROUTER_ENABLE_HTTP2_NETWORK_CLIENT, "true");

    // Set up server properties
    Properties serverSSLProps = new Properties();
    TestSSLUtils.addSSLProperties(serverSSLProps, "DC1,DC2,DC3", SSLFactory.Mode.SERVER, trustStoreFile,
        "http2-server");
    TestSSLUtils.addHttp2Properties(serverSSLProps, SSLFactory.Mode.SERVER, false);
    serverSSLProps.setProperty("server.enable.store.data.prefetch", "true");
```

zzmao:
@cgtz @justinlin-linkedin   reminder to review.

ambry-network/src/main/java/com.github.ambry.network/http2/Http2ClientResponseHandler.java
```diff
@@ -0,0 +1,91 @@
...
+
+
+/**
+ * Process {@link io.netty.handler.codec.http.FullHttpResponse} translated from HTTP/2 frames
+ */
+@ChannelHandler.Sharable
+public class Http2ClientResponseHandler extends SimpleChannelInboundHandler<FullHttpResponse> {
+  protected final Logger logger = LoggerFactory.getLogger(getClass());
+  private Queue<ResponseInfo> responseInfoQueue0 = new ConcurrentLinkedQueue<>();
+  private Queue<ResponseInfo> responseInfoQueue1 = new ConcurrentLinkedQueue<>();
```
justinlin-linkedin: what are the purpose of having two queues here?

ambry-network/src/main/java/com.github.ambry.network/http2/Http2NetworkClient.java
```diff
@@ -17,38 +17,83 @@
...
   public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Integer> requestsToDrop,
       int pollTimeoutMs) {
-    return null;
+    long startTime = SystemTime.getInstance().milliseconds();
+    List<ResponseInfo> readyResponseInfo = new ArrayList<>();
+    // Send request
+    for (RequestInfo requestInfo : requestsToSend) {
+      requestInfo.streamAcquireTime = SystemTime.getInstance().milliseconds();
+      this.pools.get(InetSocketAddress.createUnresolved(requestInfo.getHost(), requestInfo.getPort().getPort()))
+          .acquire()
```
justinlin-linkedin: can we add pollTimeout at acquire method?

ambry-rest/src/main/java/com.github.ambry.rest/Http2StreamHandler.java
```diff
@@ -31,21 +33,25 @@
...
     this.nettyConfig = nettyConfig;
     this.performanceConfig = performanceConfig;
+    this.http2ClientConfig = http2ClientConfig;
     this.requestHandler = requestHandler;
   }
 
   @Override
   public void handlerAdded(ChannelHandlerContext ctx) throws Exception {
     ctx.pipeline().addLast(new Http2StreamFrameToHttpObjectCodec(true));
+    // TODO: Remove HttpObjectAggregator. Currently, without HttpObjectAggregator, last httpContent can't be recognized as LastHttpContent.
```
justinlin-linkedin: what do you want to remove it in the TODO? Keep HttpObjectAggregator seems fine to me.

ambry-server/src/integration-test/java/com.github.ambry.server/Http2NetworkClientTest.java
```diff
@@ -0,0 +1,196 @@
+/**
+ * Copyright 2016 LinkedIn Corp. All rights reserved.
```
justinlin-linkedin: the year should be 2020

ambry-network/src/main/java/com.github.ambry.network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -0,0 +1,65 @@
...
+    // TODO: Make ByteBufChannel a RetainingByteBufChannel.
+    ByteBufChannel byteBufChannel = new ByteBufChannel();
+    try {
+      send.writeTo(byteBufChannel);
+    } catch (IOException e) {
+      promise.setFailure(e);
+    }
+    DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(byteBufChannel.getBuf(), true);
+    ctx.write(dataFrame);
+    ctx.flush();
```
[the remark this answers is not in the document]
zzmao: why not?

ambry-network/src/main/java/com.github.ambry.network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -0,0 +1,65 @@
...
+
+  }
+
+  /**
+   * Handles conversion of {@link Send} to HTTP/2 frames.
+   */
+  @Override
+  public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) {
+    if (!(msg instanceof Send)) {
+      ctx.write(msg, promise);
```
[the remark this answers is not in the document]
zzmao: good catch

ambry-network/src/main/java/com.github.ambry.network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+    Send send = (Send) msg;
+    Http2Headers http2Headers = new DefaultHttp2Headers().method(HttpMethod.POST.asciiName()).scheme("https").path("/");
```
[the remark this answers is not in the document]
zzmao: pass path in constructor?

ambry-network/src/main/java/com.github.ambry.network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -0,0 +1,65 @@
...
+    DefaultHttp2HeadersFrame headersFrame = new DefaultHttp2HeadersFrame(http2Headers, false);
+    ctx.write(headersFrame);
+    // TODO: Make ByteBufChannel a RetainingByteBufChannel.
+    ByteBufChannel byteBufChannel = new ByteBufChannel();
```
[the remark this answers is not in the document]
zzmao: Make it a retaining bytebufchannel is also something I want to sync with you. 
If we can support Send.getByteBut() eventually, then we don't need retaining bytebufchannel.

ambry-network/src/main/java/com.github.ambry.network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -0,0 +1,65 @@
...
+    try {
+      send.writeTo(byteBufChannel);
+    } catch (IOException e) {
+      promise.setFailure(e);
+    }
```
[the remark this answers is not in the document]
zzmao: sure.

ambry-network/src/main/java/com.github.ambry.network/http2/Http2ClientResponseHandler.java
```diff
@@ -0,0 +1,91 @@
+/*
+ * Copyright 2020 The Netty Project
```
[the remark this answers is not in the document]
zzmao: No. just typo. Fixed.

ambry-network/src/main/java/com.github.ambry.network/http2/Http2ClientResponseHandler.java
zzmao: 1. to reduce lock contention in this produce consumer problem. While `channelRead0` produces to one queue, `sendAndPoll` consumes the other one.
2. sendAndPoll() needs a stop while processing response. switching between two queue achieves this naturally.

ambry-network/src/main/java/com.github.ambry.network/http2/AmbrySendToHttp2Adaptor.java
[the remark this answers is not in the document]
zzmao: tried to remove this flush. Everything stuck.

ambry-rest/src/main/java/com.github.ambry.rest/Http2StreamHandler.java
zzmao: HttpObjectAggregator caused an extra copy and extra cpu resource.

I am actually thinking to remove NettyMessageProcessor. 

HttpObjectAggregator is reliable and simple.

ambry-rest/src/main/java/com.github.ambry.rest/AsyncRequestResponseHandler.java
```diff
@@ -579,7 +579,7 @@ protected void submitResponse(RestRequest restRequest, RestResponseChannel restR
         } else {
           response.readInto(restResponseChannel, responseWriteCallback);
           inFlightResponsesCount.incrementAndGet();
-          logger.trace("Response of size {} for request {} is scheduled to be sent", response.getSize(),
+          logger.info("Response of size {} for request {} is scheduled to be sent", response.getSize(),
```
cgtz: should this be changed back to trace before merging?

ambry-rest/src/main/java/com.github.ambry.rest/RestServerMain.java
```diff
@@ -80,10 +81,10 @@ public static void main(String[] args) {
    * @throws IOException
    */
   private static SSLFactory getSSLFactoryIfRequired(VerifiableProperties verifiableProperties) throws Exception {
-    ClusterMapConfig clusterMapConfig = new ClusterMapConfig(verifiableProperties);
-    if (clusterMapConfig.clusterMapHttp2NetworkClientEnabled) {
+    if (new RouterConfig(verifiableProperties).routerEnableHttp2NetworkClient) {
```
cgtz: We should change this to instantiate the SslFactory within NetworkClientFactory, since there can be some incompatibilities between the HTTP2 factory and the one used for RestServer.

ambry-router/src/main/java/com.github.ambry.router/DeleteOperation.java
```diff
@@ -123,7 +124,12 @@ private void fetchRequests(RequestRegistrationCallback<DeleteOperation> requestR
     while (replicaIterator.hasNext()) {
       ReplicaId replica = replicaIterator.next();
       String hostname = replica.getDataNodeId().getHostname();
-      Port port = replica.getDataNodeId().getPortToConnectTo();
+      Port port;
```
cgtz: Could you make a static method in RouterUtils to do the port selection?

ambry-rest/src/main/java/com.github.ambry.rest/Http2StreamHandler.java
cgtz: What would be the pathway to removing HttpObjectAggregator? is it a bug in the Netty http2 handlers or a bug in NettyMessageProcessor?

ambry-server/src/integration-test/java/com.github.ambry.server/RouterServerHttp2Test.java
```diff
@@ -0,0 +1,207 @@
...
+  @Before
+  public void before() {
+    Map<String, Meter> meters = routerMetricRegistry.getMeters();
+    transmissionSendBytesCountBeforeTest = meters.get(transmissionSendBytesMetricName).getCount();
+    transmissionReceiveBytesCountBeforeTest = meters.get(transmissionReceiveBytesMetricName).getCount();
+  }
+
+  @After
+  public void after() {
+//    TODO: assert metrics.
```
cgtz: commented code here

ambry-utils/src/main/java/com.github.ambry.utils/ByteBufChannel.java
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Copies bytes from {@code src} into the composite buf.
+   * @param src the source {@link ByteBuf} to copy bytes from.
+   * @return the number of bytes copied.
+   * @throws ClosedChannelException if the channel is closed when this function was called.
+   */
+  @Override
+  public int write(ByteBuffer src) throws ClosedChannelException {
+    int size = src.remaining();
+    compositeBuffer.addComponent(true, Unpooled.wrappedBuffer(src));
```
cgtz: Generally for implementations of `WritableByteChannel`, I think you are supposed to copy the bytes instead of just wrapping them since `ByteBuffer`s have no ref count system and may be reused by the writer afterwards. (This is what many of the code snippets for copying from ReadableByteChannel to WritableByteChannel do).

I think for the router, it may be okay to keep this optimization present but I did find some cases that I was a little unsure about. The suspicious case I found was PutOperation where the buffer is cleared in `prepareForBuilding`. I think this method is only called after a chunk is cleared. This happens when a chunk operation is completed, so for requests that we care about, it hopefully won't cause issues, but a chunk operation being completed does not necessarily mean all requests have been successfully sent out (some could be stuck in the queue).

ambry-network/src/main/java/com.github.ambry.network/http2/Http2NetworkClient.java
```diff
@@ -17,38 +17,83 @@
...
+            }
+          });
+    }
+    logger.trace("HTTP2 send request time: " + (SystemTime.getInstance().milliseconds() - startTime));
+    // TODO: close stream channel for requestsToDrop. Need a hashmap from corelationId to streamChannel
+
+    // Add the good responses to readyResponseInfo
+    Queue<ResponseInfo> queue = http2ClientResponseHandler.getQueueToConsume();
+    ResponseInfo responseInfo = queue.poll();
+    while (responseInfo != null) {
```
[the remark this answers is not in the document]
zzmao: To use one queue, I also prefer `LinkedBlockingQueue ` because it has a method `drainTo`, which takes all current available elements. This fits our use case. 

(To make sure we are in the same page): The busy loop happens when there is no/low request. 
```
Router.run { 
     while() {
         ....
         client.sendAndPool()
         ....
     }
}
```
So either one queue or two queues, we need a condition wait in `sendAndPool()` and be waken up when responses come.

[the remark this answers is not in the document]
cgtz: Yep, I think we are on the same page. For a blocking queue example, you can look at `LocalRequestResponseChannel`. The only missing part is the implementation of `wakeup`, but I think that is doable with a method that puts a dummy element in the queue.

With the 2 queue system, you would need a condition variable to signal when the list is not empty.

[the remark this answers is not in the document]
zzmao: @cgtz , I am changing to use one blocking queue. But still, I think a cv wait needed when there is no response in h2 client and a signal should be sent when a h2 response arrive, right?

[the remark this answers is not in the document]
cgtz: I think there may be a way around having an additional condition variable, since LinkedBlockingQueue already has one inside. It already supports doing a timed wait and getting notified when a new element is present. The only case that would have to be handled differently would be the wakeup call. Here is an idea I had for how to make it work. Tell me if it makes sense (adapting some code from LocalRequestResponseChannel):

For the NetworkClient (consumer) side:
```java
  private static final ResponseInfo WAKEUP_MARKER = new ResponseInfo(null, null, null);
  BlockingQueue<ResponseInfo> responseQueue = new LinkedBlockingQueue<>();
  
  List<ResponseInfo> sendAndPoll() {
    // handle requests
    
    ResponseInfo firstResponse = null;
    try {
      firstResponse = responseQueue.poll(pollTimeoutMs, TimeUnit.MILLISECONDS);
    } catch (InterruptedException ie) {
      logger.debug("Interrupted polling responses");
    }
    if (firstResponse == null) {
      return Collections.emptyList();
    }
    // using a linked list to make removal easier (maybe there's a better way to handle this)
    List<ResponseInfo> responseList = new LinkedList<>();
    responseList.add(firstResponse);
    responseQueue.drainTo(responseList);
    // remove any wakeup markers, since they are just there to signal this thread to unblock itself
    responseList.removeIf(responseInfo -> responseInfo == WAKEUP_MARKER);
    return responseList;
  }

  void wakeup() {
    // if sendAndPoll is currently executing a timed poll on the blocking queue, we need to put something there
    // to wake it up before a response comes
    try {
      responseQueue.put(WAKEUP_MARKER);
    } catch (InterruptedException e) {
      logger.debug("Interrupted while waking up");
    }
  }
```

For the ResponseHandler (producer) side:
```java
  void produce(ResponseInfo responseInfo) throws InterruptedException {
    responseQueue.put(responseInfo);
  }
```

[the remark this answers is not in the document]
zzmao: Cool idea. Will use it.

ambry-network/src/main/java/com.github.ambry.network/http2/Http2NetworkClient.java
zzmao: The pollTimeout is used for response waiting and timeout.

ambry-network/src/main/java/com.github.ambry.network/http2/AmbrySendToHttp2Adaptor.java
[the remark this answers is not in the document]
zzmao: got you. Will postpone this to follow up PRs.

ambry-rest/src/main/java/com.github.ambry.rest/Http2StreamHandler.java
zzmao: It looks like a compatible issue between http2 impl and NettyMessageProcessor. 
I need to re-consider what to do next step. Two choices:
1. fix the bug 
2. do not use NettyMessageProcessor

Will fix this is follow up PRs.

ambry-utils/src/main/java/com.github.ambry.utils/ByteBufChannel.java
zzmao: I noticed some PUT issue in EI test. Not sure yet.

To make everything safe, the latest change use copy in ByteBufChannel. Next step is to use Use  RetainingAsyncWritableChannel after PUT/GET request's `writeTo(AsyncWritableChannel channel, Callback<Long> callback)` is fully implemented.

ambry-network/src/main/java/com.github.ambry.network/http2/Http2NetworkClient.java
```diff
@@ -17,38 +17,94 @@
...
+      this.pools.get(InetSocketAddress.createUnresolved(requestInfo.getHost(), requestInfo.getPort().getPort()))
+          .acquire()
+          .addListener((GenericFutureListener<Future<Channel>>) future -> {
+            if (future.isSuccess()) {
+              http2ClientMetrics.http2StreamAcquireTime.update(System.currentTimeMillis() - streamInitiateTime);
+              Channel streamChannel = future.getNow();
+              streamChannel.pipeline().addLast(http2ClientResponseHandler);
+              streamChannel.pipeline().addLast(new AmbrySendToHttp2Adaptor());
+              streamChannel.attr(REQUEST_INFO).set(requestInfo);
+              streamChannel.write(requestInfo.getRequest());
```
justinlin-linkedin: I think If you do writeAndFlush here, you can remove ctx.flush in AmbrySendToHttp2Adaptor.write method.

There is one more thing to consider, is that when the write method fails, it's not going to be captured by anything. We need to deal with that.

ambry-rest/src/main/java/com.github.ambry.rest/Http2StreamHandler.java
justinlin-linkedin: Yeah, the NettyMessageProcessor doesn't expect a FullHttpRequest, it expects different http components. If we can get rid of NettyMessageProcessor here, I would totaly vote for it. NettyMessageProcessor does look like over complicated processor that was designed for frontend.

justinlin-linkedin: in the ambry server, we don't have multiple chunks, we don't have requests or response larger than 4MB. Lots of logic in the frontend just don't make too much sense in the ambry server.

cgtz: We do have requests and responses larger than 4 MB for replication batch-gets (and that period of time last year when we ran with 8 MB chunks), so there can be value from streaming. However, a lot of things, such as the way we prefetch blobs in the server, to get tangible benefit (other than abstract goals like sharing code btwn frontend/server) out of NettyMessageProcessor or similar pieces.

ambry-network/src/main/java/com.github.ambry.network/http2/Http2NetworkClient.java
zzmao: It works.

Let me see how to handle write failure.

ambry-network/src/main/java/com.github.ambry.network/http2/MultiplexedChannelRecord.java
```diff
@@ -131,10 +135,11 @@ protected void initChannel(Channel ch) throws Exception {
   }
 
   private void enableCloseIfIdleTask() {
+    log.info("enableCloseIfIdleTask is enabled.");
     NettyUtils.warnIfNotInEventLoop(parentChannel.eventLoop());
 
     // Don't poll more frequently than 1 second. Being overly-conservative is okay. Blowing up our CPU is not.
```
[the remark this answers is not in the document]
zzmao: The reason I changed it to 800ms is I saw some stream acquire latency is 100X ms. The change is to rule out it caused by this line. (and after the change I still see 100x ms, but anyway will keep this 800ms)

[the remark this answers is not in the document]
cgtz: Does the stream acquire time potentially include SSL handshakes on new connections leading to the high latency?

ambry-network/src/main/java/com.github.ambry.network/http2/Http2NetworkClient.java
zzmao: failure handling added.

ambry-network/src/main/java/com.github.ambry.network/http2/Http2NetworkClient.java
```diff
@@ -17,38 +17,94 @@
...
+    } catch (InterruptedException ie) {
+      logger.debug("Interrupted polling responses");
+    }
+    if (firstResponse == null) {
+      return Collections.emptyList();
+    }
+    // using a linked list to make removal easier (maybe there's a better way to handle this)
+    readyResponseInfos.add(firstResponse);
+    http2ClientResponseHandler.getResponseInfoQueue().drainTo(readyResponseInfos);
+    // wakeup markers will be detected in NonBlockingRouter#onResponse()
```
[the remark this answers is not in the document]
zzmao: thanks! FilteredInserter used.

zzmao:
@cgtz @justinlin-linkedin  addressed your last comments.

zzmao:
@justinlin-linkedin  poll is blocking call, so it waits until a new element come in. Send a wakeup_marker is to do so. 

Will a do a conflict fix.

justinlin-linkedin:
> @justinlin-linkedin poll is blocking call, so it waits until a new element come in. Send a wakeup_marker is to do so.
> 
> Will a do a conflict fix.

Oh ok, got it.

PR #1424, opened 2020-03-13
Add new version to MessageInfoAndMetadataListSerDe to delivery lifeVersion on wire

justinlin-linkedin:
currently we don't serialize lifeVersion in message info at GetResponse and ReplicaMetadataRespone. 
This PR adds a new version to MessageInfoAndMetadataListSerDe class to pass lifeVersion on wire.

After this is merged, we have to first deploy it to all the frontend so the frontend can deserialize newer version. Then we have to deploy to all server, so server can deserialize newer version.

Then we can push a new PR to write message info with newer version from ambryserver.

### DIFF
--- a/ambry-protocol/src/main/java/com.github.ambry.protocol/MessageInfoAndMetdataListSerde.java
+++ b/ambry-protocol/src/main/java/com.github.ambry.protocol/MessageInfoAndMetdataListSerde.java
@@ -41,8 +41,10 @@ class MessageInfoAndMetadataListSerde {
   static final short VERSION_3 = 3;
   static final short VERSION_4 = 4;
   static final short VERSION_5 = 5;
+  static final short VERSION_6 = 6;
+  static final short VERSION_MAX = VERSION_6;
 
-  static final short AUTO_VERSION = VERSION_5;
+  static short AUTO_VERSION = VERSION_5;
 
   private final short version;
 
@@ -93,11 +95,15 @@ class MessageInfoAndMetadataListSerde {
       size += Long.BYTES;
       // whether deleted
       size += 1;
-      if (version < VERSION_1 || version > VERSION_5) {
+      if (version < VERSION_1 || version > VERSION_MAX) {
         throw new IllegalArgumentException("Unknown version in MessageInfoList " + version);
       }
       if (version >= VERSION_5) {
         // whether ttl updated
+        size += 1;
+      }
+      if (version >= VERSION_6) {
+        // whether undelete
         size += 1;
       }
       if (version > VERSION_1) {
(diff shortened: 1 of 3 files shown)


ambry-protocol/src/main/java/com.github.ambry.protocol/MessageInfoAndMetdataListSerde.java
```diff
@@ -116,6 +122,10 @@ int getMessageInfoAndMetadataListSize() {
         // operationTime
         size += Long.BYTES;
       }
+      if (version >= VERSION_6) {
+        // lifeVersion
+        size += 2;
```
zzmao: Short.BYTES?

ambry-protocol/src/main/java/com.github.ambry.protocol/MessageInfoAndMetdataListSerde.java
```diff
@@ -93,13 +95,17 @@ int getMessageInfoAndMetadataListSize() {
...
+      if (version < VERSION_1 || version > VERSION_MAX) {
         throw new IllegalArgumentException("Unknown version in MessageInfoList " + version);
       }
       if (version >= VERSION_5) {
         // whether ttl updated
         size += 1;
       }
+      if (version >= VERSION_6) {
+        // whether undelete
+        size += 1;
```
zzmao: Change all +=1 ->  += Byte.Bytes?

ambry-protocol/src/main/java/com.github.ambry.protocol/MessageInfoAndMetdataListSerde.java
```diff
@@ -146,12 +156,15 @@ void serializeMessageInfoAndMetadataList(ByteBuffer outputBuffer) {
         outputBuffer.putLong(messageInfo.getSize());
         outputBuffer.putLong(messageInfo.getExpirationTimeInMs());
         outputBuffer.put(messageInfo.isDeleted() ? UPDATED : (byte) ~UPDATED);
-        if (version < VERSION_1 || version > VERSION_5) {
+        if (version < VERSION_1 || version > VERSION_MAX) {
           throw new IllegalArgumentException("Unknown version in MessageInfoList " + version);
```
jsjtzyy: Probably not related to your change. Just wonder why this check is not performed at the beginning of this method as  opposed to doing it in the whole loop.

justinlin-linkedin: I didn't notice that, I suppose checking at the beginning of the method makes more sense. I will update.

Issue #1425, opened 2020-03-15
NullPointerException when persisting cleanup token in HardDelete


jsjtzyy:
```
2020/03/15 08:45:22.989 ERROR [HardDeleter] [ambry-scheduler-7] [ambry-server] [] Failed to persist cleanup token
java.lang.NullPointerException
        at com.github.ambry.store.HardDeleter$HardDeletePersistInfo.pruneTill(HardDeleter.java:789)
        at com.github.ambry.store.HardDeleter.pruneHardDeleteRecoveryRange(HardDeleter.java:277)
        at com.github.ambry.store.HardDeleter.postLogFlush(HardDeleter.java:367)
        at com.github.ambry.store.PersistentIndex$IndexPersistor.write(PersistentIndex.java:1949)
        at com.github.ambry.store.PersistentIndex$IndexPersistor.run(PersistentIndex.java:1989)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:294)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:748)
```

PR #1426, opened 2020-03-15
Fix the null pointer exception in hard deleter

justinlin-linkedin:
This should fix the null pointer exception in hard deleter that recorded in this issue https://github.com/linkedin/ambry/issues/1425

The NPE is caused by performing recovery. When we persist HardDeletePersistItem to the disk, we only write the BlobReadOptions and the recoveryInfo bytes, without the start token. So when we perform the recovery, we read the BlobReadOptions and the recovery info bytes out, but now the HardDeletePersistItem doesn't have a start token. Luckily, we don't need start token for recovery, we only have to use the BlobReadOption and recovery info bytes. After performing the recovery, we have to clear the hardDeleteRecoveryRange, so later all the items added to the list would have all three fields.

Before the PR, the hardDeleteRecoveryRange is not cleared after performing the recovery, so next time when calling pruneTill method, some of the item doesn't have a startToken. Clearing it would fix this issue.

### DIFF
--- a/ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
+++ b/ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
@@ -38,6 +38,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -217,6 +221,7 @@ void performRecovery() throws StoreException {
         readOptionsList.add(item.blobReadOptions);
         messageStoreRecoveryInfoList.add(item.messagesStoreRecoveryInfo);
       }
+      hardDeleteRecoveryRange.clear();
```
jsjtzyy: Could you briefly explain why this would fix NPE in `pruneTill()`? You can add the explanation to the commit message of this PR.

justinlin-linkedin: updated.

jsjtzyy: Got it. I wonder why `readCleanupTokenAndPopulateRecoveryRange` didn't pass in `startToken` when creating recoverRange previously? Looks like it's viable to do so. Nvm, will approve this PR.

PR #1430, opened 2020-03-16
Revert the message header version back to 2 and fix a hard deleter race condition

justinlin-linkedin:
Bump the version from 2 to 3 without deploying a version of ambry server that knows how to transform version 3 message is a mistake. And it causes replication to fail. 

When we deploy a new version of ambry server that has version 3 to one machine, the other machines doesn't know how to transform the message in version 3. So in replication, they all fail. The right way of changing version should be
1. Add version 3 to all transformers, open source and close source.
2. Deploy this version to all ambry servers
3. Change the version to 3
4. Deploy the new version to all ambry servers.

### DIFF
--- a/ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
+++ b/ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
@@ -59,7 +59,7 @@ public class MessageFormatRecord {
   public static final short Metadata_Content_Version_V3 = 3;
   public static final int Message_Header_Invalid_Relative_Offset = -1;
 
-  static short headerVersionToUse = Message_Header_Version_V3;
+  static short headerVersionToUse = Message_Header_Version_V2;
 
   private static final short Delete_Subrecord_Version_V1 = 1;
   private static final short Undelete_Subrecord_Version_V1 = 1;
--- a/ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
+++ b/ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
@@ -59,6 +59,13 @@ public class MessageFormatRecord {
   public static final short Metadata_Content_Version_V3 = 3;
   public static final int Message_Header_Invalid_Relative_Offset = -1;
 
+  // Bumping version to a high number requires several things
+  // 1. Change transformers to support newer version, in open source and closed source
+  // 2. Release a newer version to all ambry-frontend and ambry-server
+  // 3. Change the version to a high value.
+  //
+  // We have to do first two steps, then can we bump the version, otherwise, other ambry-server would
+  // fail to replicate blobs through GetRequest/GetResponse, they don't know how to decode the bytes.
   static short headerVersionToUse = Message_Header_Version_V2;
 
   private static final short Delete_Subrecord_Version_V1 = 1;
--- a/ambry-server/src/integration-test/java/com.github.ambry.server/ServerHardDeleteTest.java
+++ b/ambry-server/src/integration-test/java/com.github.ambry.server/ServerHardDeleteTest.java
@@ -311,9 +311,9 @@ public class ServerHardDeleteTest {
     // For each future change to this offset, add to this variable and write an explanation of why the number changed.
     // old value: 198728. Increased by 4 to 198732 because the format for delete record went from 2 to 3 which adds
     // 4 bytes (two shorts) extra. The last record is a delete record so its extra 4 bytes are not (yet) added
-    // Add 14 here, since the message header version went from 2 to 3 and adds a short to every record, which include 6
+    // Add 14 here for version 3, since the message header version went from 2 to 3 and adds a short to every record, which include 6
     // puts and 1 delete. (last delete is not included).
-    int expectedTokenValueT1 = 198732 + 14;
+    int expectedTokenValueT1 = 198732;// + 14
     ensureCleanupTokenCatchesUp(chosenPartition.getReplicaIds().get(0).getReplicaPath(), mockClusterMap,
         expectedTokenValueT1);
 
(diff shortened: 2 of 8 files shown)


ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -172,8 +174,6 @@ void pause() throws StoreException, IOException {
       if (paused.compareAndSet(false, true)) {
         logger.info("HardDelete thread has been paused ");
         index.persistIndex();
-        pruneHardDeleteRecoveryRange();
-        persistCleanupToken();
```
justinlin-linkedin: we don't need to persistCleanupToken in pause method, since there are only two cases we have to persist cleanup token to disk. One is when the startTokenSafeToPersist is changed, which is dealt with postLogFlush method. The other one is when the hardDeleteRecoveryRange is changed, which is dealt with performHardDelete.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -464,8 +464,6 @@ void shutdown() throws InterruptedException, StoreException, IOException {
         }
         shutdownLatch.await();
         logger.info("Hard delete shutdown complete for store {} ", dataDir);
-        pruneHardDeleteRecoveryRange();
-        persistCleanupToken();
```
justinlin-linkedin: same above

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
jsjtzyy: Does this mean when `pause` method gets the undelete lock, the `performHardDelete` should be complete and token should have been persisted already?

jsjtzyy: If so, `IOException` in method signature can be removed

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -648,35 +646,36 @@ private void performHardDeletes(List<MessageInfo> messageInfoList) throws StoreE
       Iterator<HardDeleteInfo> hardDeleteIterator = hardDelete.getHardDeleteMessages(readSet, factory, null);
       Iterator<BlobReadOptions> readOptionsIterator = readOptionsList.iterator();
 
-      /* Next, get the information to persist hard delete recovery info. Get all the information and save it, as only
-       * after the whole range is persisted can we start with the actual log write */
-      while (hardDeleteIterator.hasNext()) {
-        if (!enabled.get()) {
-          throw new StoreException("Aborting hard deletes as store is shutting down",
-              StoreErrorCodes.Store_Shutting_Down);
```
jsjtzyy: any reason to remove this check?

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
jsjtzyy: same here, you can remove `StoreException, IOException` in method signature

ambry-messageformat/src/main/java/com.github.ambry.messageformat/MessageFormatRecord.java
```diff
@@ -59,7 +59,7 @@
   public static final short Metadata_Content_Version_V3 = 3;
   public static final int Message_Header_Invalid_Relative_Offset = -1;
 
-  static short headerVersionToUse = Message_Header_Version_V3;
```
zzmao: Add some notes here to remind people how to increase this version? just like you description.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: that's right. Persisting recovery information to disk is not a periodical action, it's triggered on demand. When 'startTokenSafeToPersist' changes, we should write it to file. And this is taking place in postLogFlush method. When the hardDeleteRecoveryRange or endToken changes, we should write it to file. And this is taking place in performHardDelete method. So there is no need to persist it here or in shutdown.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: explained.

ambry-server/src/integration-test/java/com.github.ambry.server/ServerHardDeleteTest.java
```diff
@@ -311,9 +311,9 @@ public void endToEndTestHardDeletes() throws Exception {
     // For each future change to this offset, add to this variable and write an explanation of why the number changed.
     // old value: 198728. Increased by 4 to 198732 because the format for delete record went from 2 to 3 which adds
     // 4 bytes (two shorts) extra. The last record is a delete record so its extra 4 bytes are not (yet) added
-    // Add 14 here, since the message header version went from 2 to 3 and adds a short to every record, which include 6
+    // Add 14 here for version 3, since the message header version went from 2 to 3 and adds a short to every record, which include 6
     // puts and 1 delete. (last delete is not included).
-    int expectedTokenValueT1 = 198732 + 14;
+    int expectedTokenValueT1 = 198732;// + 14
```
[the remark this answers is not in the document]
justinlin-linkedin: updated.

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
justinlin-linkedin: I actually don't have to, I will add it back

ambry-store/src/main/java/com.github.ambry.store/HardDeleter.java
```diff
@@ -368,16 +368,18 @@ void preLogFlush() {
...
   void postLogFlush() {
     /* start token saved before the flush is now safe to be persisted */
+    persistFileLock.lock();
     startTokenSafeToPersist = startTokenBeforeLogFlush;
-    hardDeleteLock.lock();
     try {
-      // PersistCleanupToken because startTokenSafeToPersist changed.
-      pruneHardDeleteRecoveryRange();
-      persistCleanupToken();
+      if (!isPaused()) {
```
jsjtzyy: good catch

PR #1431, opened 2020-03-18
Change query size to AzureCloudConfig.cosmosQueryBatchSize

ankagrawal:
Change replication feed query size to AzureCloudConfig.cosmosQueryBatchSize from the hardcoded value.

### DIFF
--- a/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosChangeFeedBasedReplicationFeed.java
+++ b/ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosChangeFeedBasedReplicationFeed.java
@@ -140,9 +140,11 @@ public final class CosmosChangeFeedBasedReplicationFeed implements AzureReplicat
    * Constructor to create a {@link CosmosChangeFeedBasedReplicationFeed} object.
    * @param cosmosDataAccessor {@link CosmosDataAccessor} object.
    * @param azureMetrics{@link {@link AzureMetrics} object.
+   * @param changeFeedBatchSize batch size for each change feed request.
    */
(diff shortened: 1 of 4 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/azure/CosmosChangeFeedBasedReplicationFeed.java
```diff
@@ -140,9 +140,11 @@ boolean isExpired() {
    * Constructor to create a {@link CosmosChangeFeedBasedReplicationFeed} object.
    * @param cosmosDataAccessor {@link CosmosDataAccessor} object.
    * @param azureMetrics{@link {@link AzureMetrics} object.
+   * @param changeFeedBatchSize batch size for each change feed request.
    */
-  public CosmosChangeFeedBasedReplicationFeed(CosmosDataAccessor cosmosDataAccessor, AzureMetrics azureMetrics) {
-    this.defaultCacheSize = AzureCloudDestination.getFindSinceQueryLimit();
+  public CosmosChangeFeedBasedReplicationFeed(CosmosDataAccessor cosmosDataAccessor, AzureMetrics azureMetrics,
+      int changeFeedBatchSize) {
+    this.defaultCacheSize = changeFeedBatchSize;
```
lightningrob: Why is this called default CacheSize?  Can it be overridden?

ankagrawal: Its true that it cannot be overridden in the current state of the code. 
I named this variable "default" because in future we might want a feature to tune change feed to fetch more aggressively based on if we are just catching up on the replica tokens, and the blobs are already present.

PR #1433, opened 2020-03-21
Remove dots from java package directory names

cgtz:
One odd issue I have always seen is that, when depending on ambry as a
library, intellij is unable to associate the source files with package
files even though both jars are present. I found that the reason for
this is because we include dots (.) in our package directory names
instead of the traditional way of having a separate directory for each
package name part.

This commit replaces the directory names using a script. It ignores directories that are already formatted correctly.
(e.g. com.github.ambry becomes com/github/ambry)

This is the script that I used is `remove-dot-dirs.sh` (in this PR):
```bash
#! /bin/bash 

function mvdotdir() {
  OLDDIR=$(dirname $1)
  NEWDIR=${OLDDIR//./\/}
  if [ "$OLDDIR" != "$NEWDIR" ]
  then
    echo moving $1 to $NEWDIR
    mkdir -p $NEWDIR
    git mv $1 $NEWDIR
  fi
}

find -f **/src/*/java -type f | while read file; do mvdotdir "$file"; done
```
If there are merge conflicts after this is merged with other in-progress changes, this script can hopefully be used to resolve them. Also, if a following commit uses the old format, you can use this script to convert to the new folder naming scheme in one step.


cgtz:
@zzmao rebased to fix conflicts.

For other people that are facing merge conflicts after this PR is merged, this process seems to work for me:
1. Get latest master and go back to your branch: `git checkout master && git pull --rebase upstream master && git checkout -`
2. Rebase on top of master. Most conflicts seemed to be resolved automatically by the rebase logic for me. `git rebase master`
3. To fix any new files added in your change run the script: `./remove-dot-dirs.sh`

PR #1435, opened 2020-03-24
Fix two instances of same replica during replica addition

jsjtzyy:
When adding a new replica to certain node, the node first creates a bootstrap replica and uses it to perform bootstrap work(i.e. allocate files and create new store). Afterwards, the node updates Helix
InstanceConfig which notifies all listeners in cluster (including itself). When dealing with InstanceConfig change, this node creates a second instance of same replica and adds it into in-mem clustermap.

The consequence is storage manager holds a different instance from Helix Cluster Manager, which mistakenly adds self replica into peer replicas when performing replication catchup. It blocks decommission process because there is always a self replica that will never catch up.

This PR fixes this problem by maintaining bootstrap replica set in cluster manager and it checks if there are existing replica instances when handling InstanceConfig change.

### DIFF
--- a/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
+++ b/ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
@@ -53,6 +53,7 @@ public class DynamicClusterChangeHandler implements HelixClusterChangeHandler {
   private final ClusterMapConfig clusterMapConfig;
   private final String selfInstanceName;
   private final Map<String, Map<String, String>> partitionOverrideInfoMap;
+  private final ConcurrentHashMap<String, ReplicaId> bootstrapReplicas;
   private final HelixClusterManager.HelixClusterManagerCallback helixClusterManagerCallback;
   private final HelixClusterManager.ClusterChangeHandlerCallback clusterChangeHandlerCallback;
   private final CountDownLatch routingTableInitLatch = new CountDownLatch(1);
@@ -83,13 +84,14 @@ public class DynamicClusterChangeHandler implements HelixClusterChangeHandler {
    * @param helixClusterManagerMetrics metrics to keep track of changes and status of {@link HelixClusterManager}.
    * @param onInitializationFailure callback to be called if initialization fails in a listener call.
    * @param sealedStateChangeCounter a counter indicating if sealed state of any partition has changed.
+   * @param bootstrapReplicas a map recording bootstrap replicas created on current node (only applicable to server)
    */
   DynamicClusterChangeHandler(ClusterMapConfig clusterMapConfig, String dcName, String selfInstanceName,
       Map<String, Map<String, String>> partitionOverrideInfoMap,
       HelixClusterManager.HelixClusterManagerCallback helixClusterManagerCallback,
       HelixClusterManager.ClusterChangeHandlerCallback clusterChangeHandlerCallback,
       HelixClusterManagerMetrics helixClusterManagerMetrics, Consumer<Exception> onInitializationFailure,
-      AtomicLong sealedStateChangeCounter) {
+      AtomicLong sealedStateChangeCounter, ConcurrentHashMap<String, ReplicaId> bootstrapReplicas) {
     this.clusterMapConfig = clusterMapConfig;
     this.dcName = dcName;
     this.selfInstanceName = selfInstanceName;
@@ -99,6 +101,7 @@ public class DynamicClusterChangeHandler implements HelixClusterChangeHandler {
     this.helixClusterManagerMetrics = helixClusterManagerMetrics;
     this.onInitializationFailure = onInitializationFailure;
     this.sealedStateChangeCounter = sealedStateChangeCounter;
+    this.bootstrapReplicas = bootstrapReplicas;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -360,10 +363,20 @@ private void addOrUpdateInstanceInfos(List<InstanceConfig> instanceConfigs) thro
...
-            AmbryReplica replica =
-                new AmbryServerReplica(clusterMapConfig, mappedPartition, disk, stoppedReplicas.contains(partitionName),
-                    replicaCapacity, sealedReplicas.contains(partitionName));
+            // create new replica belonging to this partition or find the existing replica from bootstrapReplicas map.
+            AmbryReplica replica;
+            if (selfInstanceName.equals(instanceName)) {
+              // if this is a newly added replica on current instance, it should be present in bootstrapReplicas map.
+              replica = (AmbryReplica) bootstrapReplicas.remove(mappedPartition.toPathString());
+              if (replica == null) {
+                logger.warn("Replica {} is not present in bootstrap replica set, skip adding it to clustermap",
```
cgtz: Would it be worth having a metric for this event? Would this mean that the replica never gets added until the machine is restarted or will it get added eventually?

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
```diff
@@ -360,10 +363,20 @@ private void addOrUpdateInstanceInfos(List<InstanceConfig> instanceConfigs) thro
...
             ensurePartitionAbsenceOnNodeAndValidateCapacity(mappedPartition, dataNode, replicaCapacity);
-            // create new replica belonging to this partition
-            AmbryReplica replica =
-                new AmbryServerReplica(clusterMapConfig, mappedPartition, disk, stoppedReplicas.contains(partitionName),
-                    replicaCapacity, sealedReplicas.contains(partitionName));
+            // create new replica belonging to this partition or find the existing replica from bootstrapReplicas map.
+            AmbryReplica replica;
+            if (selfInstanceName.equals(instanceName)) {
+              // if this is a newly added replica on current instance, it should be present in bootstrapReplicas map.
+              replica = (AmbryReplica) bootstrapReplicas.remove(mappedPartition.toPathString());
```
cgtz: Instead of passing the map directly to DynamicClusterChangeHandler, could you add a `AmbryReplica popBootstrapReplica(String partition)` method to ClusterChangeHandlerCallback?

ambry-clustermap/src/test/java/com.github.ambry.clustermap/TestUtils.java
```diff
@@ -39,20 +39,24 @@
...
     StoppedState
+    // @formatter:om
   }
 
   /**
    * Resource state associated with datanode, disk and replica.
    */
   enum ResourceState {
+    // @formatter:off
     Node_Up,
```
cgtz: No need to remove these, but for some reason, my formatter keeps the enum like this without compressing it onto one line like it used to. It may be different on different versions of LI intellij.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: Yeah, we can do that. Actually, it is quite safe in terms of concurrency. There should be only one dc (where the current node sits) calling this method.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/DynamicClusterChangeHandler.java
jsjtzyy: Actually, `replica == null` should not happen in practice. If everything goes well, then the replica should be present in bootstrap replica map. If failure occurs during bootstrap, there are two cases to discuss:
(1)  new replica is added but node crashes before updating Helix InstanceConfig
       There is no InstanceConfig change. When machine is restarted, node will re-do replica addition and newly added replica should be in the map.
(2) InstanceConfig is updated and then node crashes immediately
       Since new replica is already in InstanceConfig, once machine is restarted the new replica will be added to in-mem clustermap when handling initial InstanceConfig change (a different code path within `createNewInstance()` method)

There is only one rare case (human error) where we mistakenly update InstanceConfig of existing node by adding new replica rather than update IdealState, which may trigger this `replica == null` statement.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -73,6 +73,7 @@
   private final AtomicLong sealedStateChangeCounter = new AtomicLong(0);
   private final PartitionSelectionHelper partitionSelectionHelper;
   private final Map<String, Map<String, String>> partitionOverrideInfoMap = new HashMap<>();
+  private final ConcurrentHashMap<String, ReplicaId> bootstrapReplicas = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
jsjtzyy: Yep, will make the change.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -617,6 +637,16 @@ void removeReplicasFromPartition(AmbryPartition partition, List<AmbryReplica> re
...
       clusterWideRawCapacityBytes.getAndAdd(diskRawCapacityBytes);
     }
+
+    /**
+     * Pop out bootstrap replica (if any) on current instance. A bootstrap replica is a replica dynamically added to
+     * current node at runtime.
+     * @param partitionName the partition name of bootstrap replica.
+     * @return bootstrap replica or {@code null} if not found.
+     */
+    AmbryReplica popBootstrapReplica(String partitionName) {
```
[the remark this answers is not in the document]
jsjtzyy: Considering the context of this method, I think I will rename it to fetchBootstrapReplica. It fetches a replica out (if any) from bootstrap replica map and add it into in-memory clustermap.
(Just like fetchRequest() in GetBlobOperation, DeleteOperation, etc)

PR #1436, opened 2020-03-25
Getting single blob should avoid Cosmos

lightningrob:
Add cache hit counter for CloudBlobStore
Frontend uploads blob regardless of TTL
Handle IllegalStateException on ABS timeouts
AmbryPartition overrides equals and hashcode

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobMetadata.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobMetadata.java
@@ -434,6 +434,14 @@ public class CloudBlobMetadata {
     });
   }
 
+  /**
+   * @return a {@link CloudBlobMetadata} from a property map.
+   * @param properties the key-value property map.
+   */
+  public static CloudBlobMetadata fromMap(Map<String, String> properties) {
+    return objectMapper.convertValue(properties, CloudBlobMetadata.class);
+  }
+
   /** Custom serializer for CloudBlobMetadata class that omits fields with default values. */
   static class MetadataSerializer extends StdSerializer<CloudBlobMetadata> {
 
@@ -469,7 +477,7 @@ public class CloudBlobMetadata {
         gen.writeNumberField(FIELD_NAME_SCHEME_VERSION, value.nameSchemeVersion);
       }
       // Encryption fields that may or may not apply
-      if (value.encryptionOrigin != EncryptionOrigin.NONE) {
+      if (value.encryptionOrigin != null && value.encryptionOrigin != EncryptionOrigin.NONE) {
         gen.writeStringField(FIELD_ENCRYPTION_ORIGIN, value.encryptionOrigin.toString());
         if (value.encryptionOrigin == EncryptionOrigin.VCR) {
           gen.writeStringField(FIELD_VCR_KMS_CONTEXT, value.vcrKmsContext);
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -525,7 +525,7 @@ class CloudBlobStore implements Store {
   public boolean isKeyDeleted(StoreKey key) throws StoreException {
     checkStarted();
     // Not definitive, but okay for some deletes to be replayed.
-    return (checkCacheState(key.getID(),BlobState.DELETED));
+    return checkCacheState(key.getID(),BlobState.DELETED);
   }
 
   @Override
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -303,8 +303,9 @@ class CloudBlobStore implements Store {
         // If buffer was encrypted, we no longer know its size
         long bufferLen = (encryptedSize == -1) ? size : encryptedSize;
         InputStream uploadInputStream = new ByteBufferInputStream(messageBuf);
-        requestAgent.doWithRetries(() -> cloudDestination.uploadBlob(blobId, bufferLen, blobMetadata, uploadInputStream),
-            "Upload", partitionId.toPathString());
+        requestAgent.doWithRetries(
+            () -> cloudDestination.uploadBlob(blobId, bufferLen, blobMetadata, uploadInputStream), "Upload",
+            partitionId.toPathString());
       } else {
         // Upload blob as is
         CloudBlobMetadata blobMetadata =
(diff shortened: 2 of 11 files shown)


ambry-cloud/src/main/java/com.github.ambry.cloud/CloudBlobStore.java
```diff
@@ -341,12 +342,20 @@ private boolean shouldUpload(MessageInfo messageInfo) {
...
-        || messageInfo.getExpirationTimeInMs() - messageInfo.getOperationTimeMs() >= minTtlMillis);
+    if (isVcr) {
+      // VCR only backs up blobs with expiration time above threshold.
+      // Expired blobs are blocked by ReplicaThread.
+      // TODO: VCR for non-backup also needs to replicate everything
+      // We can change default cloudConfig.vcrMinTtlDays to 0 and override in config
+      return (messageInfo.getExpirationTimeInMs() == Utils.Infinite_Time
+          || messageInfo.getExpirationTimeInMs() - messageInfo.getOperationTimeMs() >= minTtlMillis);
+    } else {
+      // Upload all live blobs
```
lightningrob: Will update test case for this.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -161,7 +161,12 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
       return Collections.emptyMap();
     }
 
-    // TODO: For single blob GET request, get metadata from ABS
+    // For single blob GET request, get metadata from ABS instead of Cosmos
+    if (blobIds.size() == 1) {
+      CloudBlobMetadata metadata = azureBlobDataAccessor.getBlobMetadata(blobIds.get(0));
```
ankagrawal: Why do we want to avoid cosmos here? Because of consistency concern between ABS and cosmos (but that should apply to all cases) OR coz cosmos is more costly?

lightningrob: We don't want the extra round trip of going to cosmos (note that with this change there will still be two round trips to ABS, so there is additional work to do).  And it will be expensive to query Cosmos on every GET/Getinfo request.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -127,7 +127,7 @@ public boolean uploadBlob(BlobId blobId, long inputLength, CloudBlobMetadata clo
         azureMetrics.backupSuccessByteRate.mark(inputLength);
       }
       return uploaded;
-    } catch (BlobStorageException | DocumentClientException | IOException e) {
+    } catch (Exception e) {
```
zzmao: Use `Exception` to catch RunTimeException?

lightningrob: Yes.  Azure client throws IllegalStateException on timeouts.  Given the amount of different exceptions being thrown, I gave up and decided to catch them all.  If we do get a RuntimeException from our own code, we should still see it in the log.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
```diff
@@ -84,6 +84,24 @@ public boolean isEqual(String other) {
...
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    AmbryPartition partition = (AmbryPartition) o;
+    return id.equals(partition.id);
```
cgtz, line 97: not directly related to this patch, but can you change Long id into long id. No need to use a boxed long if it is non-nullable.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
```diff
@@ -84,6 +84,24 @@ public boolean isEqual(String other) {
...
+  }
+
+  @Override
+  public int hashCode() {
+    return id.hashCode();
```
cgtz, line 102: Did we find a case where a clustermap was returning a non-singleton AmbryPartition instance to require a hashcode? Not necessarily against having it since this can be a big gotcha when writing cluster aware code or test cases. However, if there are cases where PartitionId is found to have multiple instances for the same partition number, we should track them down since this class holds some state regarding sealed/stopped replicas.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -262,13 +262,41 @@ public void downloadBlob(BlobId blobId, OutputStream outputStream) throws BlobSt
...
+   * @return The {@link CloudBlobMetadata} if the blob was found, or null otherwise.
+   * @throws BlobStorageException
+   */
+  public CloudBlobMetadata getBlobMetadata(BlobId blobId) throws BlobStorageException {
+    BlockBlobClient blobClient = getBlockBlobClient(blobId, false);
+    BlobProperties blobProperties = null;
+    try {
+      blobProperties =
+          blobClient.getPropertiesWithResponse(defaultRequestConditions, requestTimeout, Context.NONE).getValue();
+      if (blobProperties == null) {
```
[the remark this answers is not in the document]
lightningrob: Technically it shouldn't, but I figured it was safer to check.

ambry-cloud/src/main/java/com.github.ambry.cloud/azure/AzureCloudDestination.java
```diff
@@ -379,9 +384,13 @@ private static CloudStorageException toCloudStorageException(String message, Exc
     } else if (e instanceof DocumentClientException) {
       statusCode = ((DocumentClientException) e).getStatusCode();
       retryDelayMs = ((DocumentClientException) e).getRetryAfterInMilliseconds();
+    } else {
+      // Note: catch-all since ABS can throw things like IOException, IllegalStateException
+      statusCode = StatusCodes.INTERNAL_SERVER_ERROR;
     }
+    logger.info("{} status {}, {}", message, statusCode, e.toString());
```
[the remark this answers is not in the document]
lightningrob: I added this mainly to log the status code which is useful for debugging and which gets lost in the CloudStorageException.  I will remove it if there are too many of them.

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
lightningrob: Done

ambry-clustermap/src/main/java/com.github.ambry.clustermap/AmbryPartition.java
lightningrob: I'm not sure.  The frequency of cache misses (even repeated ones causing 3 retries to fail) made me suspicious but I have no proof of that.

PR #1439, opened 2020-03-25
Support netty bytebuf in PutOperation

justinlin-linkedin:
This PR starts to support netty bytebuf in the PutOperation. There are some benefits out of it.

We don't have to copy the content from http channel to a java bytebuffer
We can avoid a memory allocation for java bytebuffer.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/CryptoJobHandler.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CryptoJobHandler.java
@@ -71,7 +71,7 @@ class CryptoJobHandler implements Closeable {
         }
       }
       try {
-        scheduler.awaitTermination(10000, TimeUnit.MILLISECONDS);
+        scheduler.awaitTermination(1000, TimeUnit.MILLISECONDS);
       } catch (Exception e) {
         logger.error("Unexpected exception while waiting for crypto jobs to terminate", e);
       }
--- a/ambry-router/src/main/java/com/github/ambry/router/CryptoJobHandler.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CryptoJobHandler.java
@@ -71,7 +71,7 @@ class CryptoJobHandler implements Closeable {
         }
       }
       try {
-        scheduler.awaitTermination(1000, TimeUnit.MILLISECONDS);
+        scheduler.awaitTermination(10000, TimeUnit.MILLISECONDS);
       } catch (Exception e) {
         logger.error("Unexpected exception while waiting for crypto jobs to terminate", e);
       }
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -585,7 +585,7 @@ class PutOperation {
         }
       }
       if (chunkFillingCompletedSuccessfully) {
-        // If the blob size if less then 4MB or the last chunk size is less than 4MB, than this lastChunk will be
+        // If the blob size is less then 4MB or the last chunk size is less than 4MB, than this lastChunk will be
         // the chunk above.
         PutChunk lastChunk = getBuildingChunk();
         if (lastChunk != null) {
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -553,8 +553,6 @@ class PutOperation {
         // Attempt to fill a chunk
         if (channelReadBuf == null) {
           channelReadBuf = chunkFillerChannel.getNextByteBuf(0);
-          if (channelReadBuf != null) {
-          }
         }
         if (channelReadBuf != null) {
           maybeStopTrackingWaitForChannelDataTime();
@@ -603,10 +601,11 @@ class PutOperation {
         }
       }
       if (operationCompleted) {
-        logger.info("Clear unfinished chunk since operation is completed");
         PutChunk lastChunk = getBuildingChunk();
         if (lastChunk != null) {
-          lastChunk.clear();
+          logger.info("Clear unfinished chunk since operation is completed");
+          // call release blob content, not clear, since clear should only be used in the main thread.
+          lastChunk.releaseBlobContent();
         }
       }
     } catch (Exception e) {
@@ -615,7 +614,8 @@ class PutOperation {
               RouterErrorCode.UnexpectedInternalError);
       PutChunk lastChunk = getBuildingChunk();
       if (lastChunk != null) {
-        lastChunk.clear();
+        // call release blob content, not clear, since clear should only be used in the main thread.
+        lastChunk.releaseBlobContent();
       }
       routerMetrics.chunkFillerUnexpectedErrorCount.inc();
       setOperationExceptionAndComplete(routerException);
@@ -1135,10 +1135,62 @@ class PutOperation {
     }
 
     /**
+     * The callback method to handle the result and exception from an encryption job.
+     * @param result The result of Encryption job.
+     * @param exception The exception of Encryption job.
+     */
+    private void encryptionCallback(EncryptJob.EncryptJobResult result, Exception exception) {
+      logger.trace("Processing encrypt job callback for chunk at index {}", chunkIndex);
+      if (!isMetadataChunk()) {
+        // If this is a data blob, then release the content with or without exception.
+        // When there is no exception, then the encrypted data will be used.
+        // When there is an exception, then PutOperation will be terminated, and there is no use for this blob anymore.
+        releaseBlobContent();
+      }
+      encryptJobMetricsTracker.onJobResultProcessingStart();
+      if (exception == null && !isOperationComplete()) {
+        if (!isMetadataChunk()) {
+          buf = result.getEncryptedBlobContent();
+        }
+        encryptedPerBlobKey = result.getEncryptedKey();
+        chunkUserMetadata = result.getEncryptedUserMetadata().array();
+        logger.trace("Completing encrypt job result for chunk at index {}", chunkIndex);
+        prepareForSending();
+        chunkReadyAtMs = time.milliseconds();
+      } else {
+        encryptJobMetricsTracker.incrementOperationError();
+        if (!isOperationComplete()) {
+          logger.trace("Setting exception from encrypt of chunk at index {} ", chunkIndex, exception);
+          // If we are here, then the result is null. no need to release it.
+          setOperationExceptionAndComplete(
+              new RouterException("Exception thrown on encrypting the content for chunk at index " + chunkIndex,
+                  exception, RouterErrorCode.UnexpectedInternalError));
+        } else {
+          logger.trace(
+              "Ignoring exception from encrypt job for chunk at index {} as operation exception {} is set already",
+              chunkIndex, getOperationException(), exception);
+          // If we are here, then the operation is completed and the exception could be null, in this case,
+          // we have to release the content in the result.
+          if (result != null) {
+            result.release();
+          }
+        }
+      }
+      routerMetrics.encryptTimeMs.update(time.milliseconds() - chunkEncryptReadyAtMs);
+      encryptJobMetricsTracker.onJobResultProcessingComplete();
+      routerCallback.onPollReady();
+      // double check if the operation is not completed. If so, we have to release the buf here, since in
+      // main thread, chunk might already be released.
+      if (isOperationComplete()) {
+        logger.info("Release blob content for put chunk in encryption callback since operation is completed");
+        releaseBlobContent();
+      }
+    }
+
+    /**
      * Submits encrypt job for the given {@link PutChunk} and processes the callback for the same
      */
     private void encryptChunk() {
-      ByteBuf toEncrypt = buf.retain();
       try {
         logger.trace("Chunk at index {} moves to {} state", chunkIndex, ChunkState.Encrypting);
         state = ChunkState.Encrypting;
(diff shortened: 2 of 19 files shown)


ambry-router/src/test/java/com.github.ambry.router/NonBlockingRouterTest.java
```diff
@@ -411,6 +413,7 @@ public void testRouterNoPartitionInLocalDC() throws Exception {
    */
   @Test
   public void testRequestResponseHandlerThreadExitFlow() throws Exception {
+    nettyByteBufLeakHelper.setDisabled(true);
```
cgtz: why was the leak helper disabled for this test case?

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -562,9 +587,11 @@ void fillChunks() {
         }
       }
       if (chunkFillingCompletedSuccessfully) {
+        // If the blob size if less then 4MB or the last chunk size is less than 4MB, than this lastChunk will be
```
cgtz: nit: if less then -> is less than, than this -> then this

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1152,42 +1147,62 @@ private void encryptChunk() {
...
+                    }
+                  }
+                  routerMetrics.encryptTimeMs.update(time.milliseconds() - chunkEncryptReadyAtMs);
+                  encryptJobMetricsTracker.onJobResultProcessingComplete();
+                  routerCallback.onPollReady();
+                  // double check if the operation is not completed. If so, we have to release the buf here, since in
+                  // main thread, chunk might already be released.
+                  if (isOperationComplete()) {
+                    logger.info("Clear put chunk in encryption callback since operation is completed");
+                    clear();
```
cgtz: The clear() method does not seem like it was meant to be called from non-main threads. Is it necessary to do all of the cleanup or just call `releaseBlobContent` here and in `fillChunks()`?

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1152,42 +1147,62 @@ private void encryptChunk() {
...
-                      chunkIndex, getOperationException(), exception);
-                }
-              }
-              routerMetrics.encryptTimeMs.update(time.milliseconds() - chunkEncryptReadyAtMs);
-              encryptJobMetricsTracker.onJobResultProcessingComplete();
-              routerCallback.onPollReady();
-            }));
+                isMetadataChunk() ? null : toEncrypt.retainedDuplicate(), ByteBuffer.wrap(chunkUserMetadata),
+                kms.getRandomKey(), cryptoService, kms, encryptJobMetricsTracker,
+                (EncryptJob.EncryptJobResult result, Exception exception) -> {
```
cgtz: This lambda is quite long. Could you move it to a helper function. e.g. `void onEncryptionComplete(EncryptJobResult, Exception)`?

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1144,6 +1138,7 @@ private void prepareForSending() {
      * Submits encrypt job for the given {@link PutChunk} and processes the callback for the same
      */
     private void encryptChunk() {
+      ByteBuf toEncrypt = buf.retain();
```
cgtz: It seems like this buffer is retained and then retained again with toEncrypt.retainedDuplicate() and then released at the end of this method. Could the retain from this line and release on 1204 be removed?

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1211,22 +1225,43 @@ void onFillComplete(boolean updateMetric) {
...
-    int fillFrom(ByteBuffer channelReadBuffer) {
-      int toWrite = Math.min(channelReadBuffer.remaining(), buf.remaining());
-      if (channelReadBuffer.remaining() > buf.remaining()) {
-        // Manipulate limit of the source buffer in order to read only enough to fill the chunk
-        int savedLimit = channelReadBuffer.limit();
-        channelReadBuffer.limit(channelReadBuffer.position() + buf.remaining());
-        buf.put(channelReadBuffer);
-        channelReadBuffer.limit(savedLimit);
+    int fillFrom(ByteBuf channelReadBuf) {
+      int toWrite = 0;
```
cgtz: for declaring`toWrite` and `remaininSlice` you can remove the assignment part because of each of them are assigned by all the cases afterwards

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1211,22 +1225,43 @@ void onFillComplete(boolean updateMetric) {
...
+        int remainingSize = routerConfig.routerMaxPutChunkSizeBytes - buf.readableBytes();
+        ByteBuf remainingSlice = null;
+        if (channelReadBuf.readableBytes() > remainingSize) {
+          toWrite = remainingSize;
+          remainingSlice = channelReadBuf.readRetainedSlice(remainingSize);
+        } else {
+          toWrite = channelReadBuf.readableBytes();
+          remainingSlice = channelReadBuf.readRetainedSlice(toWrite);
+        }
+        int size = buf.readableBytes();
```
cgtz: size unused

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1211,22 +1225,43 @@ void onFillComplete(boolean updateMetric) {
...
+        // buf already has some bytes
+        if (buf instanceof CompositeByteBuf) {
+          // Buf is already a CompositeByteBuf, then just add the slice from
+          ((CompositeByteBuf) buf).addComponent(true, remainingSlice);
+        } else {
+          CompositeByteBuf composite =
+              new CompositeByteBuf(buf.alloc(), buf.isDirect(), routerConfig.routerMaxPutChunkSizeBytes);
```
cgtz: The javadocs for CompositeByteBuf recommend using the `alloc.compositeBuffer()` methods if possible. Probably since the allocators seem to wrap the buffer with extra leak detection stuff (see `AbstractByteBufAllocator`). Also, the third param is the max number of child components in the buffer, not the max size in bytes. I guess it is still valuable to raise the limit since the default limit seems to be 16 from `AbstractByteBufAllocator`.

This can probably be `composite = buf.isDirect()? buf.alloc().compositeDirectBuffer(maxComponents) : buf.alloc().compositeHeapBuffer(maxComponents)` instead.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
```diff
@@ -1211,22 +1225,43 @@ void onFillComplete(boolean updateMetric) {
...
+        if (channelReadBuf.readableBytes() > remainingSize) {
+          toWrite = remainingSize;
+          remainingSlice = channelReadBuf.readRetainedSlice(remainingSize);
+        } else {
+          toWrite = channelReadBuf.readableBytes();
+          remainingSlice = channelReadBuf.readRetainedSlice(toWrite);
+        }
+        int size = buf.readableBytes();
+        // buf already has some bytes
+        if (buf instanceof CompositeByteBuf) {
```
cgtz: What if `channelReadBuf` was already a `CompositeByteBuf`? Would it be safer to just initialize the buffer in `prepareForBuilding` instead of use type checking logic here? The only thing that I think would be lost would be if the buffer came with a non-default allocator and the logic about direct vs not, but I don't think either of these will be used in practice unless the composite buffers capacity is expanded or components are squashed together inside.

ambry-router/src/test/java/com.github.ambry.router/ChunkFillTest.java
```diff
@@ -319,20 +319,20 @@ private void fillChunksAndAssertSuccess() throws Exception {
   private void assertDataIdentity(ClusterMap clusterMap) throws IOException {
     if (!testEncryption) {
       ByteBuffer dest = ByteBuffer.allocate(totalSizeWritten);
-      for (ByteBuffer buf : compositeBuffers) {
+      for (ByteBuf buf : compositeBuffers) {
         Assert.assertNotNull("All chunks should have come in", buf);
-        buf.flip();
-        dest.put(buf);
+        for (ByteBuffer buffer: buf.nioBuffers()) {
```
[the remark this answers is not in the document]
justinlin-linkedin: This is easier. dest will read all the ByteBufs in the array of compositeBuffers, so `buf.readBytes(dest)` would fail. We have to set limit on `dest` before reading bytes from `buf`. But in this way, we don't have to.

ambry-router/src/test/java/com.github.ambry.router/NonBlockingRouterTest.java
justinlin-linkedin: There are tests in `NonBlockingRouterTest` that needs creates new ByteBuf by putting a blob to the in memory router but never release it. Those tests are going to break the leak detector helper. That's why I disable the leak detector helper for those tests.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
justinlin-linkedin: make sense

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
justinlin-linkedin: legacy code.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
justinlin-linkedin: updated to new code.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
justinlin-linkedin: make sense.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
justinlin-linkedin: I was under the impression that channelReadBuf is passed to PutOperation from HttpContent and shouldn't be composite bytebuf. Is there cases where it will be composite bytebuf?

cgtz: You're probably right that there are no cases in practice with the chunked HttpContents we use. I just wanted to not make too many assumptions about the input in case something changes down the line

ambry-router/src/main/java/com/github/ambry/router/CryptoJobHandler.java
```diff
@@ -71,7 +71,7 @@ public void close() {
         }
       }
       try {
-        scheduler.awaitTermination(1000, TimeUnit.MILLISECONDS);
+        scheduler.awaitTermination(10000, TimeUnit.MILLISECONDS);
```
zzmao, line 74: any reason to increase it?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1211,22 +1229,33 @@ void onFillComplete(boolean updateMetric) {
...
+      if (buf == null) {
+        // If current buf is null, then only read the up to routerMaxPutChunkSizeBytes.
+        toWrite = Math.min(channelReadBuf.readableBytes(), routerConfig.routerMaxPutChunkSizeBytes);
+        buf = channelReadBuf.readRetainedSlice(toWrite);
       } else {
-        buf.put(channelReadBuffer);
+        int remainingSize = routerConfig.routerMaxPutChunkSizeBytes - buf.readableBytes();
+        toWrite = Math.min(channelReadBuf.readableBytes(), remainingSize);
+        ByteBuf remainingSlice = channelReadBuf.readRetainedSlice(toWrite);
+        // buf already has some bytes
```
zzmao, line 1246: Can you explain a little bit how fillFrom triggered?  I am not familiar with this area.

For 1 MB blob put, will it trigger fillFrom multiple times? 

My concern here is if this will cause many small ByteBufs in `buf`.

ambry-router/src/main/java/com.github.ambry.router/PutOperation.java
justinlin-linkedin: If we keep this code as it's, then when the `channelReadBuf` is a `CompositeByteBuf`, then when we do `channelReadBuf.readRetainedSlice`, it returns a unpooled ByteBuf. I suppose the code would still operate, but just not with the optimized memory arranagement.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
justinlin-linkedin: It depends on how the http content works. For http put request, frontend would receive a bunch of netty httpContents, each http content carries a netty bytebuf, which is `channelReadBuf`. PutManager has a running thread that calls fillFrom periodically.

justinlin-linkedin: If we have a 1MB blob, if the frontend receives 10 100K bytebuf, then the final `buf` would have 10 small ByteBuf. But the important thing is this is zero-copy bytebuf. We don't have to move any bytes from one place to another.

ambry-router/src/main/java/com/github/ambry/router/CryptoJobHandler.java
justinlin-linkedin: for testing purpose, i will revert.

PR #1443, opened 2020-03-27
Changing the MessageInfoAndMetadataListSerde version to 6

justinlin-linkedin:
Changing the MessageInfoAndMetadataListSerde's AutoVersion to version 6. In the new version, we can send out the lifeVersion in the message info so the other server can replicate based on the lifeVersion.

This has to be merged after the deployment of server 16.1.399 on both frontend and server.


lightningrob:
Any reason not to merge this now?

PR #1440, opened 2020-03-26
Integrate CompositeNetworkClient with CloudRouterFactory

cgtz:
Prior to this change, CloudRouterFactory could only communicate with
cloud services. This integrates CompositeNetworkClient which lets it
address both cloud service replicas and ambry-server replicas.

- Fix a bug in CompositeNetworkClient. sendAndPoll on the child clients
  needs to be called regardless of if there are new requests to send so
  that any responses that come back are received in a timely manner.
- Add support for the wakeup call to
  LocalNetworkClient/LocalRequestResponseChannel. This allows a
  sendAndPoll call to end before the timeout if the router indicates
  that there is extra work to do (e.g. processing decrypted blobs)
- Some MockClusterMap changes to support adding a datacenter with
  CLOUD_BACKED replicas to the cluster. This will help test retry
  policies with heterogenous clusters down the line and makes the
  existing tests work with CompositeNetworkClient for now.

PR #1445, opened 2020-03-28
Add support for cloud replicas to OperationTracker

cgtz:
CLOUD_BACKED replicas differ from DISK_BACKED replicas in that they
only require one succesful response to deem the operation a success.
The changes in this commit tune the success target and parallelism based
on the type of replica requests are sent to. For DISK_BACKED replicas,
the policies remain the same, but for CLOUD_BACKED replicas, one
successful response will complete the operation. Additionally, only one
request will be sent by SimpleOperationTracker at a time when a cloud
request is in flight (since this one request would be able to
successfully complete the operation).

Note to reviewers, this PR branch branches from #1440, so you can ignore the
first commit, "Integrate CompositeNetworkClient with CloudRouterFactory",
when reviewing this change.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -194,7 +194,6 @@ class SimpleOperationTracker implements OperationTracker {
     if (shuffleReplicas) {
       Collections.shuffle(replicas);
     }
-    int numReplicasInLocalDc = 0;
     // While iterating through the replica list, count the number of replicas from the originating DC. Subtract
     // 1 from this count to get the not found failure threshold.
     int numReplicasInOriginatingDc = 0;
@@ -209,9 +208,6 @@ class SimpleOperationTracker implements OperationTracker {
       String replicaDcName = replicaId.getDataNodeId().getDatacenterName();
       boolean isLocalDcReplica = replicaDcName.equals(datacenterName);
       boolean isOriginatingDcReplica = replicaDcName.equals(originatingDcName);
-      if (isLocalDcReplica) {
-        numReplicasInLocalDc++;
-      }
       if (isOriginatingDcReplica) {
         numReplicasInOriginatingDc++;
       }
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -16,6 +16,7 @@ package com.github.ambry.router;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.clustermap.ReplicaState;
+import com.github.ambry.clustermap.ReplicaType;
 import com.github.ambry.config.RouterConfig;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -26,6 +27,8 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Set;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -66,27 +69,32 @@ import org.slf4j.LoggerFactory;
 class SimpleOperationTracker implements OperationTracker {
   protected final String datacenterName;
   protected final String originatingDcName;
-  protected final int successTarget;
-  protected final int parallelism;
-  protected final LinkedList<ReplicaId> replicaPool = new LinkedList<>();
-
-  protected int totalReplicaCount = 0;
-  protected int inflightCount = 0;
-  protected int succeededCount = 0;
-  protected int failedCount = 0;
-  protected int disabledCount = 0;
-
+  protected final int diskSuccessTarget;
+  protected final int diskParallelism;
+  protected final int cloudSuccessTarget;
+  protected final int cloudParallelism;
+  protected final boolean cloudReplicasPresent;
   // How many NotFound responses from originating dc will terminate the operation.
   // It's decided by the success target of each mutation operations, including put, delete, update ttl etc.
-  protected int originatingDcNotFoundFailureThreshold = 0;
+  protected final int originatingDcNotFoundFailureThreshold;
+  protected final int totalReplicaCount;
+  protected final LinkedList<ReplicaId> replicaPool = new LinkedList<>();
+
+  protected int inflightCount = 0;
+  protected int diskSucceededCount = 0;
+  protected int cloudSucceededCount = 0;
+  protected int failedCount = 0;
+  protected int disabledCount = 0;
   protected int originatingDcNotFoundCount = 0;
   protected int totalNotFoundCount = 0;
   protected int diskDownCount = 0;
+  protected ReplicaId lastReturnedByIterator = null;
+  protected ReplicaType inFlightReplicaType;
 
   private final OpTrackerIterator otIterator;
   private final RouterOperation routerOperation;
   private final RouterConfig routerConfig;
-  private boolean crossColoEnabled = false;
+  private final boolean crossColoEnabled;
   private Iterator<ReplicaId> replicaIterator;
   private static final Logger logger = LoggerFactory.getLogger(SimpleOperationTracker.class);
 
@@ -127,13 +135,16 @@ class SimpleOperationTracker implements OperationTracker {
     int numOfReplicasRequired = Integer.MAX_VALUE;
     this.routerConfig = routerConfig;
     this.routerOperation = routerOperation;
+    this.originatingDcName = originatingDcName;
     datacenterName = routerConfig.routerDatacenterName;
+    cloudSuccessTarget = routerConfig.routerCloudSuccessTarget;
+    cloudParallelism = routerConfig.routerCloudRequestParallelism;
     List<ReplicaId> eligibleReplicas;
     switch (routerOperation) {
       case GetBlobOperation:
       case GetBlobInfoOperation:
-        successTarget = routerConfig.routerGetSuccessTarget;
-        parallelism = routerConfig.routerGetRequestParallelism;
+        diskSuccessTarget = routerConfig.routerGetSuccessTarget;
+        diskParallelism = routerConfig.routerGetRequestParallelism;
         crossColoEnabled = routerConfig.routerGetCrossDcEnabled;
         includeNonOriginatingDcReplicas = routerConfig.routerGetIncludeNonOriginatingDcReplicas;
         numOfReplicasRequired = routerConfig.routerGetReplicasRequired;
@@ -143,21 +154,22 @@ class SimpleOperationTracker implements OperationTracker {
       case PutOperation:
         eligibleReplicas =
             getEligibleReplicas(partitionId, datacenterName, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER));
-        successTarget = routerConfig.routerGetEligibleReplicasByStateEnabled ? Math.max(eligibleReplicas.size() - 1,
+        diskSuccessTarget = routerConfig.routerGetEligibleReplicasByStateEnabled ? Math.max(eligibleReplicas.size() - 1,
             routerConfig.routerPutSuccessTarget) : routerConfig.routerPutSuccessTarget;
-        parallelism = routerConfig.routerGetEligibleReplicasByStateEnabled ? eligibleReplicas.size()
+        diskParallelism = routerConfig.routerGetEligibleReplicasByStateEnabled ? eligibleReplicas.size()
             : routerConfig.routerPutRequestParallelism;
+        crossColoEnabled = false;
         break;
       case DeleteOperation:
-        successTarget = routerConfig.routerDeleteSuccessTarget;
-        parallelism = routerConfig.routerDeleteRequestParallelism;
+        diskSuccessTarget = routerConfig.routerDeleteSuccessTarget;
+        diskParallelism = routerConfig.routerDeleteRequestParallelism;
         crossColoEnabled = true;
         eligibleReplicas = getEligibleReplicas(partitionId, null,
             EnumSet.of(ReplicaState.BOOTSTRAP, ReplicaState.STANDBY, ReplicaState.LEADER));
         break;
       case TtlUpdateOperation:
-        successTarget = routerConfig.routerTtlUpdateSuccessTarget;
-        parallelism = routerConfig.routerTtlUpdateRequestParallelism;
+        diskSuccessTarget = routerConfig.routerTtlUpdateSuccessTarget;
+        diskParallelism = routerConfig.routerTtlUpdateRequestParallelism;
         crossColoEnabled = true;
         eligibleReplicas = getEligibleReplicas(partitionId, null,
             EnumSet.of(ReplicaState.BOOTSTRAP, ReplicaState.STANDBY, ReplicaState.LEADER));
@@ -165,10 +177,9 @@ class SimpleOperationTracker implements OperationTracker {
       default:
         throw new IllegalArgumentException("Unsupported operation: " + routerOperation);
     }
-    if (parallelism < 1) {
-      throw new IllegalArgumentException("Parallelism has to be > 0. Configured to be " + parallelism);
+    if (diskParallelism < 1) {
+      throw new IllegalArgumentException("Parallelism has to be > 0. Configured to be " + diskParallelism);
     }
-    this.originatingDcName = originatingDcName;
 
     // Order the replicas so that local healthy replicas are ordered and returned first,
     // then the remote healthy ones, and finally the possibly down ones.
@@ -191,19 +202,21 @@ class SimpleOperationTracker implements OperationTracker {
     for (ReplicaId replicaId : replicas) {
       examinedReplicas.add(replicaId);
       String replicaDcName = replicaId.getDataNodeId().getDatacenterName();
-      if (replicaDcName.equals(originatingDcName)) {
+      boolean localDcReplica = replicaDcName.equals(datacenterName);
+      boolean originatingDcReplica = replicaDcName.equals(originatingDcName);
+      if (originatingDcReplica) {
         numReplicasInOriginatingDc++;
       }
       if (!replicaId.isDown()) {
-        if (replicaDcName.equals(datacenterName)) {
+        if (localDcReplica) {
           replicaPool.addFirst(replicaId);
-        } else if (crossColoEnabled && replicaDcName.equals(originatingDcName)) {
+        } else if (crossColoEnabled && originatingDcReplica) {
           replicaPool.addLast(replicaId);
         } else if (crossColoEnabled) {
           backupReplicas.addFirst(replicaId);
         }
       } else {
-        if (replicaDcName.equals(datacenterName)) {
+        if (localDcReplica) {
           downReplicas.addFirst(replicaId);
         } else if (crossColoEnabled) {
           downReplicas.addLast(replicaId);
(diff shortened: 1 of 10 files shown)


ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -323,12 +352,13 @@ public void onResponse(ReplicaId replicaId, TrackedRequestFinalState trackedRequ
   private class OpTrackerIterator implements Iterator<ReplicaId> {
     @Override
     public boolean hasNext() {
-      return inflightCount < parallelism && replicaIterator.hasNext();
+      return inflightCount < getCurrentParallelism() && replicaIterator.hasNext();
```
jsjtzyy: Do we need to keep different `inflightCount` for diskReplica and cloudReplica separately? If inflightCount already contains request cloudReplica and we compare this number to parallelism of diskReplica might be wrong. Let's say diskReplica's parallelism is 3 and now there are 2 inflight diskReplica and 1 inflight cloudReplica, then `hasNext` may return false. However, the actual number of inflight diskReplica is less than parallelism of disk replica.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -66,27 +69,32 @@
...
-  protected final int successTarget;
-  protected final int parallelism;
+  protected final int diskSuccessTarget;
+  protected final int diskParallelism;
+  protected final int cloudSuccessTarget;
+  protected final int cloudParallelism;
+  protected final boolean cloudReplicasPresent;
+  // How many NotFound responses from originating dc will terminate the operation.
+  // It's decided by the success target of each mutation operations, including put, delete, update ttl etc.
+  protected final int originatingDcNotFoundFailureThreshold;
```
jsjtzyy: Sorry, I am not sure if I wrote the comments, but this can be applied to GET as well

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -357,20 +388,40 @@ public ReplicaId next() {
...
+        && inFlightReplicaType == ReplicaType.DISK_BACKED) {
       hasFailed = totalReplicaCount - failedCount < Math.max(totalReplicaCount - 1,
           routerConfig.routerPutSuccessTarget + disabledCount);
     } else {
-      hasFailed = (totalReplicaCount - failedCount) < successTarget || hasFailedOnNotFound();
+      // if one type of replica has a lower success target, we can only call it a complete failure when there are fewer
+      // than that target left to check. This could be optimized by saving the replica type of the remaining replicas
+      // to try and choosing the success target based on their type.
+      int minSuccessTarget = cloudReplicasPresent ? Math.min(diskSuccessTarget, cloudSuccessTarget) : diskSuccessTarget;
+      hasFailed = (totalReplicaCount - failedCount) < minSuccessTarget || hasFailedOnNotFound();
```
lightningrob: I'm having trouble following this logic.  Assuming there is a cloud replica present, totalReplicaCount = 13, minSuccessTarget will be 1, in which case failedCount has to equal totalReplicaCount to declare failure.  But if the cloud replica has failed, we will still require diskSuccessTarget successes among disk replicas - assume that is 2.  Then the most failures we can tolerate is 11 (10 disk + 1 cloud).  With failedCount = 12, hasFailed = (13-12 < 1) will be false but should be true, no?

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -261,11 +284,13 @@
...
-      hasSucceeded =
-          succeededCount >= Math.max(totalReplicaCount - disabledCount - 1, routerConfig.routerPutSuccessTarget);
+    if (routerOperation == RouterOperation.PutOperation && routerConfig.routerPutUseDynamicSuccessTarget
+        && inFlightReplicaType == ReplicaType.DISK_BACKED) {
+      // this logic only applies to disk replicas where the quorum can change during replica movement
+      int dynamicSuccessTarget = Math.max(totalReplicaCount - disabledCount - 1, routerConfig.routerPutSuccessTarget);
+      hasSucceeded = diskSucceededCount >= dynamicSuccessTarget;
     } else {
-      hasSucceeded = succeededCount >= successTarget;
+      hasSucceeded = diskSucceededCount >= diskSuccessTarget || cloudSucceededCount >= cloudSuccessTarget;
```
lightningrob: This logic looks correct.  I'm not convinced about the hasFailed logic, seems like that needs to use a similar OR condition.

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
```diff
@@ -923,11 +925,59 @@ public void operationClassTest() {
...
+      ot.onResponse(inflightReplicas.poll(), TrackedRequestFinalState.FAILURE);
+      // parallelism should have changed to 1 after cloud request was sent
+      sendRequests(ot, 0, false);
+    }
+    assertFalse("Operation should not have been done.", ot.isDone());
+    ot.onResponse(inflightReplicas.poll(), TrackedRequestFinalState.SUCCESS);
+    assertTrue("Operation should have succeeded", ot.hasSucceeded());
+    assertTrue("Operation should be done", ot.isDone());
+  }
+
```
lightningrob: Since both of these test cases succeed, shouldn't we also have one that fails?

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
```diff
@@ -944,19 +994,64 @@ private void initialize() {
...
   }
 
+  /**
+   * @param replicaType the type of replica to filter by.
+   * @return the datacenter names with replicas of type {@code replicaType}.
+   */
+  private Set<String> getDatacenters(ReplicaType replicaType) {
+    return mockPartition.getReplicaIds()
+        .stream()
+        .filter(r -> r.getReplicaType() == ReplicaType.CLOUD_BACKED)
```
lightningrob: Should compare to replicaType, no?

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -143,32 +154,32 @@
...
         crossColoEnabled = true;
         eligibleReplicas = getEligibleReplicas(partitionId, null,
             EnumSet.of(ReplicaState.BOOTSTRAP, ReplicaState.STANDBY, ReplicaState.LEADER));
         break;
       default:
         throw new IllegalArgumentException("Unsupported operation: " + routerOperation);
     }
-    if (parallelism < 1) {
-      throw new IllegalArgumentException("Parallelism has to be > 0. Configured to be " + parallelism);
+    if (diskParallelism < 1) {
```
jsjtzyy: I think this check also applies to cloudParallelism.
Also, could you add additional info in the exception message to record operation type? (so far, only PUT operation has dynamic parallelism but it's still good to add type into error message)

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -227,22 +240,32 @@
...
-          generateErrorMessage(partitionId, examinedReplicas, replicaPool, backupReplicasToCheck, downReplicasToCheck));
+
+    // MockPartitionId.getReplicaIds() is returning a shared reference which may cause race condition.
+    // Please report the test failure if you run into this exception.
+    Supplier<IllegalArgumentException> notEnoughReplicasException = () -> new IllegalArgumentException(
+        generateErrorMessage(partitionId, examinedReplicas, replicaPool, backupReplicasToCheck, downReplicasToCheck));
+    // initialize this to the replica type of the first request to send so that parallelism is set correctly for the
+    // first request
+    inFlightReplicaType =
+        replicaPool.stream().findFirst().map(ReplicaId::getReplicaType).orElseThrow(notEnoughReplicasException);
```
jsjtzyy: If I remember correctly, we are supposed to try replicas in local dc first, then replica in linkedin dcs, and then cloud replica. So this logic has changed?

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -227,22 +240,32 @@
         replicaPool.add(downReplicas.pollFirst());
       }
     }
+    cloudReplicasPresent =
+        replicaPool.stream().map(ReplicaId::getReplicaType).anyMatch(Predicate.isEqual(ReplicaType.CLOUD_BACKED));
```
jsjtzyy: If `routerGetEligibleReplicasByStateEnabled` is true, the replicaPool won't include cloudReplica. Do we need to explicitly add it into pool?

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
```diff
@@ -66,27 +69,32 @@
...
+  // It's decided by the success target of each mutation operations, including put, delete, update ttl etc.
+  protected final int originatingDcNotFoundFailureThreshold;
+  protected final int totalReplicaCount;
   protected final LinkedList<ReplicaId> replicaPool = new LinkedList<>();
 
-  protected int totalReplicaCount = 0;
   protected int inflightCount = 0;
-  protected int succeededCount = 0;
+  protected int diskSucceededCount = 0;
+  protected int cloudSucceededCount = 0;
```
jsjtzyy: I would suggest renaming these variables to something like  `diskReplicaSuccessCount` to distinguish it from `diskDownCount` more easily.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
cgtz: This logic has not changed. The order of the items in replica pool is still the same. This line just gets the replica type of the first item in the linked list, which should be a replica in the local dc.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
cgtz: I was under the impression that it would be returned since `CloudServiceClusterChangeHandler` has the following implementation (return the replica unconditionally if it exists):
```
  @Override
  public Stream<AmbryReplica> getReplicaIdsByState(AmbryPartition partition, ReplicaState state) {
    AmbryReplica replica = partitionToReplica.get(partition.toPathString());
    return replica == null ? Stream.empty() : Stream.of(replica);
  }
```

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
jsjtzyy: ah, I get your point. The frontend that uses operation tracker might be Azure frontend. In that case, first replica type is cloud replica, right? Did we ensure cloud replica will be picked in first place if when populating the replica pool?

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
cgtz: Yes, will fix this in my commit to add more tests and address comments.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
cgtz: This is an interesting case. There is a question here about whether the cloud and disk parallelism can be summed or not. In your case the effective parallelism could be 4 since you are allowing both disk and cloud requests to reach full parallelism.

However, I feel that although this increases the chance of success for the disk replicas, it is not completely necessary as a single acknowledgement from a cloud replica could satisfy the operation. The reason I decided to shrink the parallelism instead of summing it was to not waste a single cloud request by also sending requests to disk replicas at the same time. I notice that I also made an assumption that the requests are ordered by datacenter, which sometimes isn't true for the backup and down nodes.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
cgtz: Changed the comment. The success target no longer has a part in this calculation after the dynamic target changes

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
cgtz: I guess this is a little rough to use the min. The reason I did that was to account for the case where the 13th replica was a cloud replica that could still satisfy the request with a success target of 1. However, this has the downside of wasting a request if that last replica was a disk replica.

To address this, I could do what the comment suggests and figure out the replica types of the remaining replicas. To avoid iterating through the entire list every time hasFailed is called, I could record the index of the last cloud replica and the last disk replica in the pool. If we are past the last replica of a certain type, switch to the other success target. Or, in this method, compare against the max success target and then if that returns true, make a second comparison by inspecting the remaining replicas in the pool (this should be a relatively smaller # of items to iterate through assuming success targets are small).

Do either of these approaches sound better?

ambry-router/src/test/java/com.github.ambry.router/OperationTrackerTest.java
cgtz: Added some more test cases

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
cgtz: The cloud replica will be returned in the list of replica IDs if it exists. Since each cloud replica has a virtual "datanode" it will follow the same datacenter ordering logic that other replicas do. For azure fabrics, it will come first.

jsjtzyy: Make sense. Thanks for the explanation.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
jsjtzyy: Right, I forgot cloud dc is added to `dcToDcInfo`

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
jsjtzyy: I feel ok with current logic. Perhaps we can add debug log to help determine if the request being sent is cloudReplica request or disk one.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
jsjtzyy: Sounds like 1st approach is more efficient.

jsjtzyy: In addition, let's also account for PUT operation when this is an Azure frontend. I feel like we need an extra `else if` for this case because the first `if` only applies to DiskReplica in regular dc(s).

ambry-clustermap/src/main/java/com.github.ambry.clustermap/HelixClusterManager.java
```diff
@@ -316,10 +316,14 @@ public void onReplicaEvent(ReplicaId replicaId, ReplicaEventType event) {
         replica.getDataNodeId().onNodeTimeout();
         break;
       case Disk_Error:
-        replica.getDiskId().onDiskError();
+        if (replica.getReplicaType() == ReplicaType.DISK_BACKED) {
+          replica.getDiskId().onDiskError();
+        }
```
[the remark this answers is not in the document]
cgtz: Right now, I only implemented CloudServiceReplicas for HelixClusterManager, but I will add this in in case we have to add this later on.

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
```diff
@@ -83,7 +81,7 @@
       localDcResourceToHistogram = getResourceToLatencyMap(routerOperation, true);
       crossDcResourceToHistogram = getResourceToLatencyMap(routerOperation, false);
     }
-    if (parallelism > routerConfig.routerOperationTrackerMaxInflightRequests) {
+    if (Math.max(diskParallelism, cloudParallelism) > routerConfig.routerOperationTrackerMaxInflightRequests) {
```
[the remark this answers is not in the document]
jsjtzyy: I guess the reason we introduced `routerOperationTrackerMaxInflightRequests` is to throttle total # of request in regular data centers. So, although current number of disk requests equals `MaxInflightRequests`, router is still able to route request to cloud which is not restricted by `MaxInflightRequests` for regular dc(s). 
I agree with one point that we'd better break it into different checks and probably use separate `MaxInflightRequests` for cloud dc and regular dc.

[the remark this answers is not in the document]
cgtz: Yep, to expand on this, the adaptive operation tracker has the ability to go beyond the parallelism setting if requests are past due. This limit makes sure that it does not snowball and keep sending out parallel requests.

ambry-router/src/main/java/com.github.ambry.router/SimpleOperationTracker.java
cgtz: Hi @jsjtzyy, for the put operation in azure, what condition were you thinking of where the logic will be different from the fallback policy? I feel like there is no concept of disabled for cloud replicas.

cgtz: For this condition I added a check in `hasFailedOnNotFound` to always return false on put operations. The condition would have returned false anyways given how we instantiate the operationtracker in PutOperation, but this makes it clearer.

For the failure threshold I changed the logic around a bit by counting how many replicas of each type are left. Using the success count for each type and this number, we can tell if a success is still possible or not for each type without requiring the hacky min logic

ambry-router/src/main/java/com.github.ambry.router/AdaptiveOperationTracker.java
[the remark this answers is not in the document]
cgtz: I added a second if for the constructor validations but kept the logic as is. Since max in flight requests is used when the number of requests made needs to go beyond the initial parallelism for performance reasons but isn't related to the quorum, I kept the config as is so that all in flight requests count towards the same limit.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -191,19 +207,24 @@
     for (ReplicaId replicaId : replicas) {
       examinedReplicas.add(replicaId);
       String replicaDcName = replicaId.getDataNodeId().getDatacenterName();
-      if (replicaDcName.equals(originatingDcName)) {
+      boolean isLocalDcReplica = replicaDcName.equals(datacenterName);
+      boolean isOriginatingDcReplica = replicaDcName.equals(originatingDcName);
+      if (isLocalDcReplica) {
+        numReplicasInLocalDc++;
```
jsjtzyy, line 213: minor: `numReplicasInLocalDc` is updated but never accessed. Consider removing it?

cgtz: Sorry left that in there since I was thinking about adding in some changes to the logic for adding backup replicas when includeOriginating is false to just use the total number of replicas in the local and remote dcs instead of hardcoding it to 6. I ultimately decided not to couple these changes though since it affects our current production cluster the most.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -261,21 +288,32 @@
...
+      return false;
+    }
+    if (originatingDcNotFoundFailureThreshold > 0
+        && originatingDcNotFoundCount >= originatingDcNotFoundFailureThreshold) {
+      return true;
+    }
     // To account for GET operation, the threshold should be  >= totalReplicaCount - (success target - 1)
+    // Right now, this only applies for disk replica only partitions and may not be completely accurate if there are
+    // failures responses other than not found.
+    // TODO support cloud replicas in this condition, also account for failures other than not found
```
lightningrob, line 314: Is there still a TODO here?

cgtz: Yes that is still a todo. I will make github issues for the todos that I added to this file

PR #1455, opened 2020-04-03
Start supporting undelete in compaction

justinlin-linkedin:
Please reference it to https://docs.google.com/document/d/1rm6NBOUeZRMbuF26116lerytiW3tde38I23cZVr3SG0/edit?pli=1#heading=h.hdj8rn37deoq

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -1356,15 +1356,15 @@ class BlobStoreCompactor {
       // ----------------------------------------------------------------------------------------
       List<IndexEntry> validEntries = new ArrayList<>();
       Iterator<IndexEntry> iterator = indexSegment.getIterator();
-      StoreKey lastKey = null;
-      IndexValue lastFinalState = null;
+      StoreKey previousKey = null;
+      IndexValue previousLatestState = null;
       while (iterator.hasNext()) {
         IndexEntry entry = iterator.next();
         StoreKey currentKey = entry.getKey();
         IndexValue currentValue = entry.getValue();
-        IndexValue currentFinalState;
+        IndexValue currentLatestState;
 
-        if (lastKey == null) {
+        if (previousKey == null) {
           // save a token for restart (the key gets ignored but is required to be non null for construction)
           StoreFindToken safeToken =
               new StoreFindToken(currentKey, indexSegment.getStartOffset(), sessionId, incarnationId);
@@ -1373,42 +1373,43 @@ class BlobStoreCompactor {
         }
         // If an IndexSegment contains more than one IndexValue for the same StoreKey, then they must follow each other
         // since IndexSegment store IndexValues based on StoreKey. If the current key equals to the previous key, then
-        // we don't have to query the final state again.
-        if (currentKey.equals(lastKey)) {
-          currentFinalState = lastFinalState;
+        // we don't have to query the latest state again.
+        if (currentKey.equals(previousKey)) {
+          currentLatestState = previousLatestState;
         } else {
-          lastKey = currentKey;
-          lastFinalState = currentFinalState = srcIndex.findKey(currentKey);
+          previousKey = currentKey;
+          previousLatestState = currentLatestState = srcIndex.findKey(currentKey);
         }
 
         if (currentValue.isUndelete()) {
-          if (currentFinalState.isPut()) {
+          if (currentLatestState.isPut()) {
             throw new IllegalStateException(
-                "Undelete's final state can't be put for key" + currentKey + " in store " + dataDir);
+                "Undelete's latest state can't be put for key" + currentKey + " in store " + dataDir);
           }
-          if (currentFinalState.isUndelete() && currentFinalState.getLifeVersion() == currentValue.getLifeVersion()
+          if (currentLatestState.isUndelete() && currentLatestState.getLifeVersion() == currentValue.getLifeVersion()
               && !srcIndex.isExpired(currentValue)) {
             validEntries.add(entry);
           }
         } else if (currentValue.isDelete()) {
-          if (currentFinalState.isPut()) {
+          if (currentLatestState.isPut()) {
             throw new IllegalStateException(
-                "Delete's final state can't be put for key" + currentKey + " in store " + dataDir);
+                "Delete's latest state can't be put for key" + currentKey + " in store " + dataDir);
           }
-          if (currentFinalState.isDelete() && currentFinalState.getLifeVersion() == currentValue.getLifeVersion()) {
+          if (currentLatestState.isDelete() && currentLatestState.getLifeVersion() == currentValue.getLifeVersion()) {
             validEntries.add(entry);
           }
         } else if (currentValue.isTtlUpdate()) {
-          if (currentFinalState.isPut()) {
-            // If isPut returns true, when the final state doesn't carry ttl_update flag, this is wrong
+          if (currentLatestState.isPut()) {
+            // If isPut returns true, when the latest state doesn't carry ttl_update flag, this is wrong
             throw new IllegalStateException(
-                "TtlUpdate's final state can't be put for key" + currentKey + " in store " + dataDir);
+                "TtlUpdate's latest state can't be put for key" + currentKey + " in store " + dataDir);
           }
           // This is a TTL_UPDATE record, then the blob can't be expired. Only check if it's deleted or not.
-          if (currentFinalState.isDelete()) {
-            if (currentFinalState.getOperationTimeInMs() >= compactionLog.getCompactionDetails().getReferenceTimeMs()) {
+          if (currentLatestState.isDelete()) {
+            if (currentLatestState.getOperationTimeInMs() >= compactionLog.getCompactionDetails()
+                .getReferenceTimeMs()) {
               validEntries.add(entry);
-            } else if (isTtlUpdateEntryValidWhenFinalStateIsDeleteAndRetention(currentKey,
+            } else if (isTtlUpdateEntryValidWhenLatestStateIsDeleteAndRetention(currentKey,
                 indexSegment.getStartOffset())) {
               validEntries.add(entry);
             }
@@ -1416,19 +1417,20 @@ class BlobStoreCompactor {
             validEntries.add(entry);
           }
         } else {
-          if (srcIndex.isExpired(currentFinalState)) {
+          if (srcIndex.isExpired(currentLatestState)) {
             logger.trace("{} in index segment with start offset {} in {} is not valid because it is an expired PUT",
                 entry, indexSegment.getStartOffset(), storeId);
             continue;
           }
-          if (currentFinalState.isPut()) {
-            if (currentFinalState.getLifeVersion() != currentValue.getLifeVersion()) {
+          if (currentLatestState.isPut()) {
+            if (currentLatestState.getLifeVersion() != currentValue.getLifeVersion()) {
               throw new IllegalStateException(
                   "Two different lifeVersions  for puts key" + currentKey + " in store " + dataDir);
             }
             validEntries.add(entry);
-          } else if (currentFinalState.isDelete()) {
-            if (currentFinalState.getOperationTimeInMs() >= compactionLog.getCompactionDetails().getReferenceTimeMs()) {
+          } else if (currentLatestState.isDelete()) {
+            if (currentLatestState.getOperationTimeInMs() >= compactionLog.getCompactionDetails()
+                .getReferenceTimeMs()) {
               validEntries.add(entry);
             } else {
               logger.trace("{} in index segment with start offset {} in {} is not valid because it is a deleted PUT",
@@ -1450,7 +1452,7 @@ class BlobStoreCompactor {
      * @return {@code true} if the TTL update entry is valid
      * @throws StoreException if there are problems reading the index
      */
-    private boolean isTtlUpdateEntryValidWhenFinalStateIsDeleteAndRetention(StoreKey key,
+    private boolean isTtlUpdateEntryValidWhenLatestStateIsDeleteAndRetention(StoreKey key,
         Offset indexSegmentStartOffset) throws StoreException {
       boolean valid = false;
       //  A TTL update entry is "valid" if the corresponding PUT is still alive
@@ -1462,7 +1464,7 @@ class BlobStoreCompactor {
         logger.trace("TTL update of {} in segment with start offset {} in {} is not valid the corresponding PUT entry "
             + "does not exist anymore", key, indexSegmentStartOffset, storeId);
       } else {
-        // PUT exists in the source index even when the final state for this StoreKey is delete and delete already reaches
+        // PUT exists in the source index even when the latest state for this StoreKey is delete and delete already reaches
         // retention date. This can happen either because
         // 1. The log segment to which srcValue(PUT) belongs is not under compaction, thus PUT is not compacted.
         // 2. srcValue(PUT) will be compacted in this cycle (because it has been determined that the PUT is not valid. Since
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -632,17 +632,36 @@ class BlobStoreCompactor {
    * @throws StoreException if {@link BlobReadOptions} could not be obtained from the store for deleted blobs.
    */
   private List<IndexEntry> getValidIndexEntries(IndexSegment indexSegment) throws StoreException {
-    // save a token for restart (the key gets ignored but is required to be non null for construction)
-    // Assumed preference order from IndexSegment (current impl)
-    // (Legend: entry/entries in segment -> output from IndexSegment#getIndexEntriesSince())
-    // PUT entry only -> PUT entry
-    // TTL update entry only -> TTL update entry
-    // DELETE entry only -> DELETE entry
-    // PUT + DELETE -> DELETE
-    // TTL update + DELETE -> DELETE
-    // PUT + TTL update -> PUT (the one relevant to this comment)
-    // PUT + TTL update + DELETE -> DELETE
-    // TODO: move this blob store stats
+    // Validity of a IndexValue is determined by itself and it's latest state. For example, if the current IndexValue
+    // is a Put, and the latest IndexValue for this blob is a Delete, and the operation is done out of retention duration,
+    // then this Put IndexValue is considered as invalid.
+    //
+    // There is one exception, TtlUpdate IndexValue. A TtlUpdate IndexValue's validity is not only depends on the latest
+    // state of the blob, it's also affected by the Put IndexValue. If the latest IndexValue for this blob is a Delete,
+    // and it's operation time is out of the retention duration, this TtlUpdate record would be considered as invalid.
+    // However if the Put IndexValue exists for this TtlUpdate and is not under compaction, then we will keep this TtlUpdate
+    // anyway.
+    //
+    // This is a table that shows how the validity of the current IndexValue is determined.
+    // ----------------------------------------------------------------------------------------
+    // Current IndexValue  | Latest IndexValue | Action                                       |
+    // --------------------+-------------------+----------------------------------------------|
+    // Put(verion c)       | Put(version f)    | isExpired(Pc)?false:true                     |
+    //                     | Delete(f)         | reachRetention(Df)||isExpired(Df)?false:true |
+    //                     | Undelete(f)       | isExpired(Uf)?false:true                     |
+    // --------------------+-------------------+----------------------------------------------|
+    // TtlUpdate(c)        | Put(f)            | Exception                                    |
+    //                     | Delete(f)         | reachRetention(Df)?false:true                |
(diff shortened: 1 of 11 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -556,27 +530,15 @@ private boolean copyDataByLogSegment(LogSegment logSegmentToCopy, FileSpan dupli
...
       FileSpan duplicateSearchSpan) throws IOException, StoreException {
     logger.debug("Copying data from {}", indexSegmentToCopy.getFile());
-    List<IndexEntry> allIndexEntries = new ArrayList<>();
     // call into diskIOScheduler to make sure we can proceed (assuming it won't be 0).
     diskIOScheduler.getSlice(INDEX_SEGMENT_READ_JOB_NAME, INDEX_SEGMENT_READ_JOB_NAME, 1);
-    // get all entries. We get one entry per key
-    indexSegmentToCopy.getIndexEntriesSince(null, new FindEntriesCondition(Long.MAX_VALUE), allIndexEntries,
-        new AtomicLong(0), true);
-
-    // save a token for restart (the key gets ignored but is required to be non null for construction)
```
zzmao, line 566: Can you explain why this logic is moved?

justinlin-linkedin: It's not removed, it's relocated, to the method `getValidIndexEntries`.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+        IndexValue currentFinalState;
+
+        if (lastKey == null) {
+          // save a token for restart (the key gets ignored but is required to be non null for construction)
+          StoreFindToken safeToken =
+              new StoreFindToken(currentKey, indexSegment.getStartOffset(), sessionId, incarnationId);
+          compactionLog.setSafeToken(safeToken);
+          logger.debug("Set safe token for compaction in {} to {}", storeId, safeToken);
+        }
+        if (currentKey.equals(lastKey)) {
```
jsjtzyy, line 1377: When will this happen?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+
+    /**
+     * Determines whether a TTL update entry is valid. A TTL update entry is valid as long as the associated PUT record
+     * is still present in the store (the validity of the PUT record does not matter - only its presence/absence does).
+     * @param key the {@link StoreKey} being examined
+     * @param indexSegmentStartOffset the start offset of the {@link IndexSegment} that the TTL update record is in
+     * @return {@code true} if the TTL update entry is valid
+     * @throws StoreException if there are problems reading the index
+     */
+    private boolean isTtlUpdateEntryValidWhenFinalStateIsDeleteAndRetention(StoreKey key,
```
jsjtzyy, line 1453: what does the retention mean, within retention time?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+          }
+        } else if (currentValue.isTtlUpdate()) {
+          if (currentFinalState.isPut()) {
+            // If isPut returns true, when the final state doesn't carry ttl_update flag, this is wrong
+            throw new IllegalStateException(
+                "TtlUpdate's final state can't be put for key" + currentKey + " in store " + dataDir);
+          }
+          // This is a TTL_UPDATE record, then the blob can't be expired. Only check if it's deleted or not.
+          if (currentFinalState.isDelete()) {
+            if (currentFinalState.getOperationTimeInMs() >= compactionLog.getCompactionDetails().getReferenceTimeMs()) {
```
jsjtzyy, line 1409: Can you explain the intention of `currentFinalState.getOperationTimeInMs() >= compactionLog.getCompactionDetails().getReferenceTimeMs()`?
Also looks like if `CurrentFinalState.isDelete()`, the if/else if can combined together.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -125,6 +127,13 @@ public boolean accept(File dir, String name) {
     this.sessionId = sessionId;
     this.incarnationId = incarnationId;
     this.useDirectIO = Utils.isLinux() && config.storeCompactionEnableDirectIO;
+    IndexSegmentValidEntryFilter filter;
+    if (config.storeCompactionFilter.equals(IndexSegmentValidEntryFilterWithoutUndelete.class.getSimpleName())) {
+      filter = new IndexSegmentValidEntryFilterWithoutUndelete();
+    } else {
+      filter = new IndexSegmentValidEntryFilterWithUndelete();
+    }
+    validEntryFilter = filter;
```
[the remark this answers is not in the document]
justinlin-linkedin: the validEntryFilter field is a final field, it has to be initialized in the constructor. If we initialize it in the if-else statement, they might be some compile error. This way, the `validEntryFilter` is sure to be initialized.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+    private List<IndexEntry> getValidIndexEntries(IndexSegment indexSegment, List<IndexEntry> allIndexEntries)
+        throws StoreException {
+      // Assumed preference order from IndexSegment (current impl)
+      // (Legend: entry/entries in segment -> output from IndexSegment#getIndexEntriesSince())
+      // PUT entry only -> PUT entry
+      // TTL update entry only -> TTL update entry
+      // DELETE entry only -> DELETE entry
+      // PUT + DELETE -> DELETE
+      // TTL update + DELETE -> DELETE
+      // PUT + TTL update -> PUT (the one relevant to this comment)
```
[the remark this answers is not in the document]
justinlin-linkedin: These comments are a bit wired, it's not the rules for compaction, it's actually the rules for returning IndexValues from an index segment. Like this Rule `PUT + TTL -> PUT`, it means if a index segment has PUT and TTL for the same key, `indexSegment.getIndexEntriesSince` would return PUT.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: This method would iterate through all the IndexValue in an IndexSegment. For same key, they might be multiple IndexValue, like. PUT + TTL + DELETE, since IndexSegment sorts IndexValues based on their keys, so if we have multiple IndexValues for the same key in the same IndexSegment, then they must follow each other. So here, if we detect the current key equals to the previous one, we don't have to search for the final state for the same key again.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: Yes, current retention time is 7 days.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: `compactionLog.getCompactionDetails().getReferenceTimeMs()` is set to be the current time - retention duration when starting compaction. So if the retention duration is 7 days, then the reference time is now - 7d. And this if statement means if the delete's operation is later than now - 7d (e.g. now - 1d), the delete and ttl update should not be compacted.

I have two if statement under this `if (currentFinalState.isDelete())`, it's just easier to read in this way.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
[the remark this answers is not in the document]
jsjtzyy: I actually tried to initialize it in if-else statement, the tests passed and intellij didn't complain. (might worth giving it a shot)

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+          // if srcValue is not a delete, it is a duplicate.
+          return !srcValue.isDelete();
+        } else if (value.isUndelete()) {
+          if (srcValue.isPut()) {
+            throw new IllegalStateException(
+                "An Undelete[" + value + "] and a Put[" + srcValue + "] can't be at the same lifeVersion at store "
+                    + dataDir);
+          }
+          // value is a UNDELETE without a TTL update or a DELETE
+          return !srcValue.isDelete() && !srcValue.isTtlUpdate();
```
jsjtzyy, line 1517: this can be simplified to `return srcValue.isUndelete()`. We have precluded other cases.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+        } else {
+          if (srcValue.isUndelete()) {
+            throw new IllegalStateException(
+                "A Put[" + value + "] and an Undelete[" + srcValue + "] can't be at the same lifeVersion at store "
+                    + dataDir);
+          }
+          // value is a PUT without a TTL update or a DELETE
+          return !srcValue.isDelete() && !srcValue.isTtlUpdate();
```
jsjtzyy, line 1525: same here, this can be simplified (as long as comment is here to help us understand the only possible duplicate case is PUT)

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
jsjtzyy: Makes sense. Better to add a comment here.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+      //
+      // There is one exception, TtlUpdate IndexValue. A TtlUpdate IndexValue's validity is not only depends on the latest
+      // state of the blob, it's also affected by the Put IndexValue. If the latest IndexValue for this blob is a Delete,
+      // and it's operation time is out of the retention duration, this TtlUpdate record would be considered as invalid.
+      // However if the Put IndexValue exists for this TtlUpdate and is not under compaction, then we will keep this TtlUpdate
+      // anyway.
+      //
+      // This is a table that shows how the validity of the current IndexValue is determined.
+      // ----------------------------------------------------------------------------------------
+      // Current IndexValue  | Latest IndexValue | Action                                       |
```
jsjtzyy, line 1342: better to change `Action` to `Is Valid`

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+          if (currentFinalState.isUndelete() && currentFinalState.getLifeVersion() == currentValue.getLifeVersion()
+              && !srcIndex.isExpired(currentValue)) {
+            validEntries.add(entry);
+          }
+        } else if (currentValue.isDelete()) {
+          if (currentFinalState.isPut()) {
+            throw new IllegalStateException(
+                "Delete's final state can't be put for key" + currentKey + " in store " + dataDir);
+          }
+          if (currentFinalState.isDelete() && currentFinalState.getLifeVersion() == currentValue.getLifeVersion()) {
```
jsjtzyy, line 1398: why we don't check retention time here?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+      // The PUT entry, if it exists, must be "before" this TTL update entry.
+      FileSpan srcSearchSpan = new FileSpan(srcIndex.getStartOffset(), indexSegmentStartOffset);
+      IndexValue srcValue = srcIndex.findKey(key, srcSearchSpan, EnumSet.of(PersistentIndex.IndexEntryType.PUT));
+      if (srcValue == null) {
+        // PUT is not in the source - it might be compacted
+        logger.trace("TTL update of {} in segment with start offset {} in {} is not valid the corresponding PUT entry "
+            + "does not exist anymore", key, indexSegmentStartOffset, storeId);
+      } else {
+        // Exists in the source. This can happen either because
+        // 1. The FileSpan to which srcValue belongs is not under compaction (so there is no reason for tgt to have it)
```
jsjtzyy, line 1466: To make the comment clearer. could you change `srcValue` to `PUT`?
(same for `it`, if I understand correctly, `it` is also referring to `PUT`)

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +998,533 @@ private void endCompaction() {
...
+              validEntries.add(entry);
+            } else if (isTtlUpdateEntryValidWhenFinalStateIsDeleteAndRetention(currentKey,
+                indexSegment.getStartOffset())) {
+              validEntries.add(entry);
+            }
+          } else {
+            validEntries.add(entry);
+          }
+        } else {
+          if (srcIndex.isExpired(currentFinalState)) {
```
jsjtzyy, line 1419: I probably asked this before, here I just want to confirm again: both delete and undelete will carry the TTL from original PUT, right?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -1170,5 +1183,34 @@ static Offset getIndexSegmentStartOffset(String filename) {
...
+   * only be used in the compaction so that the {@link IndexSegment} should be sealed already.
+   */
+  class SealedIndexSegmentEntryIterator implements Iterator<IndexEntry> {
+    private int currentIdx = 0;
+    private ByteBuffer mmap = serEntries.duplicate();
+    private byte[] valueBuf = new byte[valueSize];
+
+    @Override
+    public boolean hasNext() {
+      return currentIdx < numberOfEntries(mmap);
```
jsjtzyy, line 1198: if this is a sealed segment, is `numberOfEntries(mmap)` a fixed value? If yes, it's worth keeping it somewhere without calculating multiple times.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -1170,5 +1183,34 @@ static Offset getIndexSegmentStartOffset(String filename) {
...
+      try {
+        StoreKey key = getKeyAt(mmap, currentIdx);
+        mmap.get(valueBuf);
+        return new IndexEntry(key, new IndexValue(startOffset.getName(), ByteBuffer.wrap(valueBuf), getVersion()));
+      } catch (Exception e) {
+        logger.error("Failed to read index entry at " + currentIdx, e);
+      } finally {
+        currentIdx++;
+      }
+      return null;
```
jsjtzyy, line 1212: This will cause NullPointException if it failed to read index. Is this intended?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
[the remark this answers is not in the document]
justinlin-linkedin: You are right, I will update.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: We don't have to. The current value is delete, and it's only valid when it's the last delete for this blob. eg. P -> D0 -> U1 -> D1. If D0 is the current value, then it's not valid, since D1 is the last delete. eg. P -> D0 -> U1 -> D1 -> U2. If D1 is the current value, then it's not valid, since final state is not a delete.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: sure

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: They will carry the latest expiration date, which means if there is no TTL_UPDATE after PUT, then they will carry the expiration date of PUT.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
justinlin-linkedin: sure

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
justinlin-linkedin: that's right, will update.

ambry-store/src/main/java/com/github/ambry/store/IndexSegmentValidEntryFilter.java
```diff
@@ -0,0 +1,46 @@
...
+   * @param checkAlreadyCopied {@code true} if a check for existence in the swap spaces has to be executed (due to
+   *                                       crash/shutdown), {@code false} otherwise.
+   * @return the list of valid {@link IndexEntry} sorted by their offset.
+   */
+  List<IndexEntry> getValidEntry(IndexSegment indexSegment, FileSpan duplicateSearchSpan, boolean checkAlreadyCopied)
+      throws StoreException;
+
+  /**
+   * Checks if a record already exists in {@code idx}.
+   * @param idx the {@link PersistentIndex} to search in
```
zzmao, line 37: minor: index

ambry-store/src/main/java/com/github/ambry/store/IndexSegmentValidEntryFilter.java
```diff
@@ -0,0 +1,46 @@
...
+package com.github.ambry.store;
+
+import java.util.List;
+
+
+/**
+ * Interface to return valid entries from a index segment. Those valid entries will be copied to a new segment
+ * in compaction process.
+ */
+public interface IndexSegmentValidEntryFilter {
```
zzmao, line 23: I like it.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +996,536 @@ private void endCompaction() {
...
+      //                     | Delete(f)         | c==f?true:false                              |
+      //                     | Undelete(f)       | false                                        |
+      // --------------------+-------------------+----------------------------------------------|
+      // Undelete(c)         | Put(f)            | Exception                                    |
+      //                     | Delete(f)         | false                                        |
+      //                     | Undelete(f)       | c==f&&!isExpired(Uc)?true:false              |
+      // ----------------------------------------------------------------------------------------
+      List<IndexEntry> validEntries = new ArrayList<>();
+      Iterator<IndexEntry> iterator = indexSegment.getIterator();
+      StoreKey lastKey = null;
```
zzmao, line 1359: final -> latest   to align with comments.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
zzmao: +1 , comment to explain why don't need to query latest state: because they share same latest state.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
zzmao: final -> latest

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1295,4 +996,536 @@ private void endCompaction() {
...
+        if (lastKey == null) {
+          // save a token for restart (the key gets ignored but is required to be non null for construction)
+          StoreFindToken safeToken =
+              new StoreFindToken(currentKey, indexSegment.getStartOffset(), sessionId, incarnationId);
+          compactionLog.setSafeToken(safeToken);
+          logger.debug("Set safe token for compaction in {} to {}", storeId, safeToken);
+        }
+        // If an IndexSegment contains more than one IndexValue for the same StoreKey, then they must follow each other
+        // since IndexSegment store IndexValues based on StoreKey. If the current key equals to the previous key, then
+        // we don't have to query the final state again.
```
zzmao, line 1376: final -> latest

justinlin-linkedin:
> Core logic looks good.
> 
> Suggest to change all final to latest.

Comments addressed

PR #1456, opened 2020-04-07
Add metrics to track progress of replication from cloud.

ankagrawal:
Since a metric like replication lag in bytes is not possible from cloud, a metric that can be used to track replication progress from cloud can be the point of time up to which the local replica has caught up from the cloud replica. Below are considerations for implementing such a metric. 

There are two options to emit metrics for recovery/vcr-replication progress

1. Progress is reported by the VCR that is the recovery source.
2. Progress is reported by the storage node itself that is recovering.

## Progress Reported By Vcr

When a VCR gets a replica metadata request, then it can look at the token to see the lastUpdateTime (which needs to be added to the CloudFindToken) in the incoming token, and can add it to a metric called *partition_<partitionid_datanodeid>_catchup_point*. This metric is the time up to which the remote is caught up from the VCR.

A major assumption here is that one storage node recovers/replicates from only one VCR node. If there is a possibility of one storage node replicating from multiple VCR nodes (which basically doesn't make sense, since the source of data is same for all VCRs), then it will not be possible to determine which VCR node has most up to date progress. Since such a possibility is not expected, this assumption is expected to hold true.

A major issue with approach is that there will be no way to know when to drop the metrics from a VCR node. Since this happens in the request path, a VCR node will never be able to drop the metrics. Due to this issue, the next approach is a better approach to report progress.

## Progress Reported by the Storage Node

Just like a storage node emits metrics for *replicationLagInBytes*, it can also emit metrics for *replicationCatchupTime* (time up to which a replica has caught up). The response of replica metadata exchange consists of an ordered list of *MessageInfo*, representing the metadata of a blob. The operation time of the last MessageInfo, can be used as the point in time up to which the local replica has caught up from cloud replica. While this information can be obtained for both replication from cloud as well as from store, it makes more sense to track this only for replication from cloud, because there will be three peer store replicas for a local replica, and it is difficult to determine how to merge the three catch up timestamps to get accurate information about catch up point (there can be a remote store peer replica, which is new, and is still catching up with the existing replicas. It will be far behind the other replicas). With the assumption that there will be only one cloud replica per local replica in a store, this approach can be used to track progress of replication from cloud, and hence progress of disaster recovery.

Due to the reasons above, approach 2 (progress reported by the storage node), is used to get the recovery progress metric.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
@@ -40,7 +40,7 @@ public class ReplicationMetrics {
 
   private final static String MAX_LAG_FROM_PEERS_IN_BYTE_METRIC_NAME_TEMPLATE = "Partition-%s-maxLagFromPeersInBytes";
   private final static String CATCH_POINT_FROM_CLOUD_METRIC_NAME_TEMPLATE = "Partition-%s-catchupPointFromCloud";
-  
+
   public final Map<String, Meter> interColoReplicationBytesRate = new HashMap<String, Meter>();
   public final Meter intraColoReplicationBytesRate;
   public final Map<String, Meter> plainTextInterColoReplicationBytesRate = new HashMap<String, Meter>();
@@ -736,7 +736,7 @@ public class ReplicationMetrics {
     // update this metric only for cloud peer replica. There will only be one cloud replica peer per partition.
     if (remoteReplicaInfo.getReplicaId().getReplicaType() == ReplicaType.CLOUD_BACKED
         && cloudReplicaCatchUpPoint.containsKey(remoteReplicaInfo.getLocalReplicaId().getPartitionId())) {
-      // update the partition's lag if and only if it was tracked.
+      // update the partition's catch up point if and only if it was tracked.
       cloudReplicaCatchUpPoint.put(remoteReplicaInfo.getLocalReplicaId().getPartitionId(), catchUpPoint);
     }
   }
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
@@ -22,6 +22,7 @@ import com.codahale.metrics.Timer;
 import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.clustermap.ReplicaType;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -136,6 +137,7 @@ public class ReplicationMetrics {
   private final Map<PartitionId, Counter> partitionIdToInvalidMessageStreamErrorCounter = new HashMap<>();
   // ConcurrentHashMap is used to avoid cache incoherence.
   private final Map<PartitionId, Map<DataNodeId, Long>> partitionLags = new ConcurrentHashMap<>();
+  private Map<PartitionId, Long> partitionCatchUpPoint = new ConcurrentHashMap<>();
   private final Map<String, Set<RemoteReplicaInfo>> remoteReplicaInfosByDc = new ConcurrentHashMap<>();
   private final Map<String, LongSummaryStatistics> dcToReplicaLagStats = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -671,6 +699,20 @@ public void updateLagMetricForRemoteReplica(RemoteReplicaInfo remoteReplicaInfo,
...
   }
 
+  /**
+   * Update catch up point of local replica from the cloud replica.
+   * @param remoteReplicaInfo {@link RemoteReplicaInfo} of the cloud replica.
+   * @param catchUpPoint timestamp upto which local replica has caught with the cloud replica.
+   */
+  public void updateCatchupPointMetricForRemoteReplica(RemoteReplicaInfo remoteReplicaInfo, long catchUpPoint) {
+    // update this metric only for cloud peer replica. There will only be one cloud replica peer per partition.
+    if (remoteReplicaInfo.getReplicaId().getReplicaType() == ReplicaType.CLOUD_BACKED
```
lightningrob, line 709: If this is only for cloud replicas, should the metric name reflect that?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -671,6 +699,20 @@ public void updateLagMetricForRemoteReplica(RemoteReplicaInfo remoteReplicaInfo,
...
 
+  /**
+   * Update catch up point of local replica from the cloud replica.
+   * @param remoteReplicaInfo {@link RemoteReplicaInfo} of the cloud replica.
+   * @param catchUpPoint timestamp upto which local replica has caught with the cloud replica.
+   */
+  public void updateCatchupPointMetricForCloudReplica(RemoteReplicaInfo remoteReplicaInfo, long catchUpPoint) {
+    // update this metric only for cloud peer replica. There will only be one cloud replica peer per partition.
+    if (remoteReplicaInfo.getReplicaId().getReplicaType() == ReplicaType.CLOUD_BACKED
+        && cloudReplicaCatchUpPoint.containsKey(remoteReplicaInfo.getLocalReplicaId().getPartitionId())) {
```
cgtz, line 710: Not sure if its possible for there to be a conflict with removeCatchupPointMetricForPartition being called concurrently, but if there is, you could use `cloudReplicaCatchUpPoint.computeIfPresent(remoteReplicaInfo.getLocalReplicaId().getPartitionId(), (k, v) -> catchUpPoint)` instead.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -434,6 +437,20 @@ public void addLagMetricForPartition(PartitionId partitionId) {
...
+   * Add catchup point metric(local from cloud) for given partitionId.
+   * @param partitionId partition to add metric for.
+   */
+  public void addCatchUpPointMetricForPartition(PartitionId partitionId) {
+    if (!cloudReplicaCatchUpPoint.containsKey(partitionId)) {
+      cloudReplicaCatchUpPoint.put(partitionId, 0L);
+      // Set up metrics if and only if no mapping for this partition before.
+      Gauge<Long> catchUpPoint = () -> cloudReplicaCatchUpPoint.get(partitionId);
+      registry.register(MetricRegistry.name(ReplicaThread.class,
+          "Partition-" + partitionId.toPathString() + "-catchupPointFromCloud"), catchUpPoint);
```
cgtz, line 450: minor: I know it wasn't done for the other similar metrics, but since this name format is used in two methods you could probably make a helper method to generate the name
```java
String getCatchupPointFromCloudMetricName(PartitionId partitionId) { 
  return MetricRegistry.name(ReplicaThread.class, "Partition-" + partitionId.toPathString() + "-catchupPointFromCloud");
}
```

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
ankagrawal: done.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
ankagrawal: done.

lightningrob:
Will merge once the build passes.
The PR description is very thorough - recommend adding it to the github project wiki if there's an appropriate place for it.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
ankagrawal: As discussed offline, since remove only removes the metric from the registry and not from the map, there shouldn't be any conflict here.

PR #1459, opened 2020-04-07
Add some metrics for cloud replication feed

ankagrawal:
Add metrics to check cosmos change feed cache hit, miss and refresh rate.
Add metrics to check time taken for cloud replication feed query.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosChangeFeedBasedReplicationFeed.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosChangeFeedBasedReplicationFeed.java
@@ -172,11 +172,11 @@ public final class CosmosChangeFeedBasedReplicationFeed implements AzureReplicat
       if (changeFeedCacheEntry == null || !isCacheValid(partitionPath, cosmosChangeFeedFindToken,
           changeFeedCacheEntry)) {
         // the cache may not be valid. So we cannot use session id
+        azureMetrics.changeFeedCacheMissRate.mark();
+        cacheHit = false;
         changeFeedCacheEntry = getNextChangeFeed(partitionPath, cosmosChangeFeedFindToken.getStartContinuationToken());
         // invalidate the previous token's cache
         changeFeedCache.remove(cosmosChangeFeedFindToken.getCacheSessionId());
-        azureMetrics.changeFeedCacheMissRate.mark();
-        cacheHit = false;
         index = 0;
       }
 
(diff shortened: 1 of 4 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosChangeFeedBasedReplicationFeed.java
```diff
@@ -169,49 +170,64 @@ public CosmosChangeFeedBasedReplicationFeed(CosmosDataAccessor cosmosDataAccesso
...
+          changeFeedCacheEntry = getNextChangeFeed(partitionPath, changeFeedCacheEntry.getEndContinuationToken(),
+              changeFeedCacheEntry.getCacheSessionId());
+          fetchedEntries = changeFeedCacheEntry.getFetchedEntries();
+          if (fetchedEntries.isEmpty()) {
+            // this means that there are no new changes
+            break;
+          } else {
+            azureMetrics.changeFeedCacheRefreshRate.mark();
+          }
+          index = 0;
```
jsjtzyy, line 218: I probably misunderstand this but I am thinking if there is a case where `cacheHit` is true before entering the `while(true)` loop.  The first time code enters in while loop, `index 
 == fetchedEntries.size()` and jumps to `else` branch. It does fetch new changes and index is reset to 0. Then it goes back to beginning of while(true) and `azureMetrics.changeFeedCacheHitRate.mark();` will be called. I just want to confirm this is a possible case and updating `changeFeedCacheHitRate` is what we expect.

ankagrawal: First time code enters in while loop, and if index == fetchedEntries.size(), then we refresh the cache as you mentioned. In this case we don't want to say cacheHit just yet, because we haven't yet hit the cache. We will refresh the cache and mark it so.

Now once the refresh is successful, there can be either no results (fetchedEntries.isEmpty()), in which case we will come out of loop without marking cache hit (as expected because it isn't technically a cache hit). If there are entries returned  in cache refresh, then loop's next iteration will run, which will mark the cache hit. 

In short unless we are able to consume something from the cache, we don't mark a cache hit. Hope this makes sense.

lightningrob:
Looks like you need to merge one conflict before merging.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosChangeFeedBasedReplicationFeed.java
jsjtzyy: make sense

PR #1460, opened 2020-04-07
Fix http2 bytebuf leak issue

zzmao:
ByteBuf from PutRequest(or other requests) should be released by networkClient if exception happens in networkClient.


ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -99,6 +99,8 @@ public void operationComplete(ChannelFuture future) throws Exception {
                     http2ClientMetrics.http2StreamWriteAndFlushErrorCount.inc();
                     logger.warn("Stream writeAndFlush fail: {}", future.cause());
                   }
+                  // release related bytebuf
+                  requestInfo.getRequest().release();
```
justinlin-linkedin, line 103: in line 108, we should release the request as well.

PR #1466, opened 2020-04-09
Treat 'SSLEngine closed already' as a client termination exception.

cgtz:
When clients close the connection from their side, the router may
still be delivering data. When using SSL, the write call in
NettyResponseChannel may encounter an exception indicating the SSLEngine
has already been closed, which can happen when the connection is in the
process of being closed.

This change treats such exceptions as client-side connection
terminations so they will not be logged so verbosely. See
NettyResponseChannel::log for how Utils::isPossibleClientTermination
is used.

### DIFF
--- a/ambry-utils/src/main/java/com/github/ambry/utils/Utils.java
+++ b/ambry-utils/src/main/java/com/github/ambry/utils/Utils.java
@@ -58,6 +58,7 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import javax.net.ssl.SSLException;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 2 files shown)


ambry-utils/src/main/java/com/github/ambry/utils/Utils.java
```diff
@@ -1026,8 +1030,14 @@ public static void shutDownExecutorService(ExecutorService executorService, long
    * @return {@code true} this cause indicates a possible early termination from the client. {@code false} otherwise.
    */
   public static boolean isPossibleClientTermination(Throwable cause) {
-    return cause instanceof IOException && (CLIENT_RESET_EXCEPTION_MSG.equals(cause.getMessage())
-        || CLIENT_BROKEN_PIPE_EXCEPTION_MSG.equals(cause.getMessage()));
+    if (cause instanceof SSLException) {
+      return SSL_ENGINE_CLOSED_EXCEPTION_MSG.equals(cause.getMessage());
```
justinlin-linkedin, line 1034: Actually SSLException is an IOException, so we don't have to do this if-else statement.

cgtz: That is true, it is an IOException. I will simplify this code block by just checking the message. The one thing we lose out on by doing that is if there is a rogue IOException that is not an SSLException with the same message, but that is not likely (and we may want to catch it anyways).

PR #1468, opened 2020-04-09
Increase retention time to fix the hard deleter test failure

justinlin-linkedin:
This PR increases the retention day in the hard delete integration test to fix this failure:
```
com.github.ambry.server.ServerHardDeleteTest > endToEndTestHardDeletes FAILED
1321    java.lang.AssertionError: expected:<Blob_Deleted_Permanently> but was:<No_Error>
1322        at org.junit.Assert.fail(Assert.java:91)
1323        at org.junit.Assert.failNotEquals(Assert.java:645)
1324        at org.junit.Assert.assertEquals(Assert.java:126)
1325        at org.junit.Assert.assertEquals(Assert.java:145)
1326        at com.github.ambry.server.ServerHardDeleteTest.undeleteBlob(ServerHardDeleteTest.java:416)
1327        at com.github.ambry.server.ServerHardDeleteTest.endToEndTestHardDeletes(ServerHardDeleteTest.java:349)
```

The problem is when we delete a blob in the test case, we undelete it later. There are some time gap between these two operations and somehow the time instance runs too far to surpass one day, which is the retention day for deleted blob. So when we undelete it, it's already too late.

This PR changes the retention day to 7 and hopefully this would fix this the failure.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/IndexValue.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/IndexValue.java
@@ -437,7 +437,7 @@ class IndexValue implements Comparable<IndexValue> {
   public String toString() {
     return "Offset: " + offset + ", Size: " + getSize() + ", Deleted: " + isDelete() + ", TTL Updated: " + isTtlUpdate()
         + ", Undelete: " + isUndelete() + ", ExpiresAtMs: " + getExpiresAtMs() + ", Original Message Offset: "
-        + getOriginalMessageOffset() + (formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeAtSecs "
+        + getOriginalMessageOffset() + (formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeInMs "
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/IndexValue.java
```diff
@@ -437,7 +437,7 @@ ByteBuffer getBytes() {
   public String toString() {
     return "Offset: " + offset + ", Size: " + getSize() + ", Deleted: " + isDelete() + ", TTL Updated: " + isTtlUpdate()
         + ", Undelete: " + isUndelete() + ", ExpiresAtMs: " + getExpiresAtMs() + ", Original Message Offset: "
-        + getOriginalMessageOffset() + (formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeAtSecs "
+        + getOriginalMessageOffset() + (formatVersion != PersistentIndex.VERSION_0 ? (", OperationTimeInMs "
```
jsjtzyy, line 440: good catch

PR #1467, opened 2020-04-09
Add skeleton for router to support undelete

justinlin-linkedin:
First step to support undelete in the frontend. This PR adds skeleton code in the router for undelete manager and undelete operation.

PR #1469, opened 2020-04-09
Adding frontend support for undelete

justinlin-linkedin:
Supporting undelete in frontend. There is no storage call to actually undelete blob, it's added as a todo. This PR will be merged after https://github.com/linkedin/ambry/pull/1467.


ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
```diff
@@ -201,6 +203,10 @@
   @Default("20 * 1024 * 1024")
   public final int maxJsonRequestSizeBytes;
 
+  @Config(ENABLE_UNDELETE)
```
zzmao, line 206: Why create this frontend config `ENABLE_UNDELETE`? Looks like you already have a  `ROUTER_ENABLE_UNDELETE` in the other PR?

justinlin-linkedin: I've removed the configuration in the router. This is the gateway for undelete, to control it here is better.

zzmao:
Good to merge?

Issue #1369, opened 2020-01-28
Intermittent test failure: com.github.ambry.router.CloudRouterTest » testRouterBasic[1]


cgtz:
This failed once for me on travis CI and succeeded on the following attempt:
```
com.github.ambry.router.RouterException: Server returned: Blob_Expired. Chunk ID: AAYQAgfuY7wAAQAAAAAAAAAHLlADt2RqSkmfDIZM8yc6KQ Error: BlobExpired
java.util.concurrent.ExecutionException: com.github.ambry.router.RouterException: Server returned: Blob_Expired. Chunk ID: AAYQAgfuY7wAAQAAAAAAAAAHLlADt2RqSkmfDIZM8yc6KQ Error: BlobExpiredOpen stacktrace
Caused by: com.github.ambry.router.RouterException: Server returned: Blob_Expired. Chunk ID: AAYQAgfuY7wAAQAAAAAAAAAHLlADt2RqSkmfDIZM8yc6KQ Error: BlobExpiredClose stacktrace
at com.github.ambry.router.GetBlobOperation$GetChunk.buildChunkException(GetBlobOperation.java:1068)
at com.github.ambry.router.GetBlobOperation$GetChunk.processGetBlobResponse(GetBlobOperation.java:937)
at com.github.ambry.router.GetBlobOperation$GetChunk.handleResponse(GetBlobOperation.java:828)
at com.github.ambry.router.GetBlobOperation.handleResponse(GetBlobOperation.java:247)
at com.github.ambry.router.GetManager.handleResponse(GetManager.java:226)
at com.github.ambry.router.NonBlockingRouter$OperationController.onResponse(NonBlockingRouter.java:856)
at com.github.ambry.router.NonBlockingRouter$OperationController.run(NonBlockingRouter.java:894)
at java.lang.Thread.run(Thread.java:748)
```

Build scan link: https://scans.gradle.com/s/gfipjv2jcdigm/tests/vw7cjzzt7rohe-s5avd6yx7dauy?open-stacktraces=WzFd

cgtz:
Happened here too: https://gradle.com/s/rzds3uifocoby
```
com.github.ambry.router.CloudRouterTest > testRouterBasic[3] FAILED
[...]

PR #1471, opened 2020-04-11
Check cache state before updating blobs including retries

lightningrob:
No actual changes to Cosmos consistency level.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -67,6 +67,7 @@ class CloudBlobStore implements Store {
   private static final Logger logger = LoggerFactory.getLogger(CloudBlobStore.class);
   private static final int cacheInitialCapacity = 1000;
   private static final float cacheLoadFactor = 0.75f;
+  private static final int STATUS_NOT_FOUND = 404;
   private final PartitionId partitionId;
   private final CloudDestination cloudDestination;
   private final ClusterMap clusterMap;
@@ -357,6 +358,8 @@ class CloudBlobStore implements Store {
     }
   }
 
+  // TODO: delete and update should throw StoreErrorCodes.ID_Not_Found on 404
+
   @Override
   public void delete(List<MessageInfo> infos) throws StoreException {
     checkStarted();
@@ -370,7 +373,9 @@ class CloudBlobStore implements Store {
             partitionId.toPathString());
       }
     } catch (CloudStorageException ex) {
-      throw new StoreException(ex, StoreErrorCodes.IOError);
+      StoreErrorCodes errorCode =
+          (ex.getStatusCode() == STATUS_NOT_FOUND) ? StoreErrorCodes.ID_Not_Found : StoreErrorCodes.IOError;
+      throw new StoreException(ex, errorCode);
     }
   }
 
@@ -422,7 +427,9 @@ class CloudBlobStore implements Store {
         }
       }
     } catch (CloudStorageException ex) {
-      throw new StoreException(ex, StoreErrorCodes.IOError);
+      StoreErrorCodes errorCode =
+          (ex.getStatusCode() == STATUS_NOT_FOUND) ? StoreErrorCodes.ID_Not_Found : StoreErrorCodes.IOError;
+      throw new StoreException(ex, errorCode);
     }
   }
 
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -358,8 +358,6 @@ class CloudBlobStore implements Store {
     }
   }
 
-  // TODO: delete and update should throw StoreErrorCodes.ID_Not_Found on 404
-
   @Override
   public void delete(List<MessageInfo> infos) throws StoreException {
     checkStarted();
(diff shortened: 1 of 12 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -357,6 +358,8 @@ private boolean shouldUpload(MessageInfo messageInfo) {
     }
   }
 
+  // TODO: delete and update should throw StoreErrorCodes.ID_Not_Found on 404
```
lightningrob, line 361: I'll remove this comment

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -248,22 +254,43 @@ private boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value
...
-    // 1) the CosmosDB metadata collection
-    // 2) the blob storage entry metadata (to enable rebuilding the database)
-
+    // 1) the blob storage entry metadata (so GET's can be served entirely from ABS)
+    // 2) the CosmosDB metadata collection
     try {
-      if (!azureBlobDataAccessor.updateBlobMetadata(blobId, fieldName, value)) {
-        // TODO: what if this is a retry where ABS has been updated but Cosmos has not?
-        return false;
+      boolean updatedStorage, updatedCosmos = false;
```
cgtz, line 260: nit: could you separate these two assignments onto separate lines? it looks a little confusing since one of the values is assigned/mutated but the other is not assigned until line 265

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -162,7 +167,8 @@ public boolean updateBlobExpiration(BlobId blobId, long expirationTime) throws C
     }
 
     // For single blob GET request, get metadata from ABS instead of Cosmos
-    if (blobIds.size() == 1) {
+    // Note: findMissingKeys needs to query Cosmos regardless
```
cgtz, line 170: Could you expand on why the cosmos query is needed for findMissingKeys in this comment?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
```diff
@@ -144,47 +141,37 @@ void testConnectivity() {
...
-  }
-
-  /**
-   * Update the blob metadata document in the CosmosDB collection, retrying as necessary.
+   * Update the blob metadata document in the CosmosDB collection.
    * @param blobId the {@link BlobId} for which metadata is replaced.
    * @param fieldName the metadata field to update.
    * @param value the new value for the field.
    * @return the {@link ResourceResponse} returned by the operation, if successful.
    * Returns {@Null} if the document is not found or the field already has the specified value.
```
cgtz, line 154: This comment says that null will be returned if the document is not found. I assume this has been changed to throw an exception in this case?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
```diff
@@ -144,47 +141,37 @@ void testConnectivity() {
...
-      return null;
-    }
+
+    // Read the existing record
+    String docLink = getDocumentLink(blobId.getID());
+    RequestOptions options = getRequestOptions(blobId.getPartition().toPathString());
+    ResourceResponse<Document> readResponse =
+        executeCosmosAction(() -> asyncDocumentClient.readDocument(docLink, options).toBlocking().single(),
+            azureMetrics.documentReadTime);
+    Document doc = readResponse.getResource();
```
cgtz, line 166: how is the doc == null case handled now?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -248,22 +254,43 @@ private boolean updateBlobMetadata(BlobId blobId, String fieldName, Object value
...
+      // Note: even if nothing changed in blob storage, still attempt to update Cosmos since this could be a retry
+      // of a request where ABS was updated but Cosmos update failed.
+      try {
+        ResourceResponse<Document> response = cosmosDataAccessor.updateMetadata(blobId, fieldName, value);
+        updatedCosmos = response != null;
+      } catch (DocumentClientException dex) {
+        if (dex.getStatusCode() == HttpConstants.StatusCodes.NOTFOUND) {
+          // blob exists in ABS but not Cosmos - inconsistent state
+          // Recover by inserting the updated map into cosmos
+          cosmosDataAccessor.upsertMetadata(CloudBlobMetadata.fromMap(metadataMap));
```
cgtz, line 276: I just want to check if there are any conditions where it may not be good to create a record. For example, what if compaction already cleaned up the record (or is that sequence of events impossible)?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
lightningrob: Per Cosmos it can't happen.  It either returns a document response or throws exception.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
lightningrob: Good question.  It is possible for a blob to be purged but the Cosmos record not yet purged.  In this case, is it even possible for a lifecycle event to be passed here?  I would think the ReplicaThread would cut it off earlier, but not sure if the undelete feature changes that.  In that specific case, either expirationTime or deletionTime should be older than the retention period.

cgtz:
Restarted the failed build. I saw an occurence of this intermittent test failure: #1369
Let's see if it occurs again.

PR #1472, opened 2020-04-13
Add basic support for multiple participants on server side

jsjtzyy:
In some cases where single datanode has to parcipate into mulitiple clusters
and it needs to update node info in these clusters separately. This requires
muptiple participants on same node and each of them is responsible for interacting
with corresponding cluster. With this feature, we are able to build a mirror cluster
that helps us perform some monitoring and adminstration operations (i.e. ZK migration).

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/clustermap/ClusterAgentsFactory.java
+++ b/ambry-api/src/main/java/com/github/ambry/clustermap/ClusterAgentsFactory.java
@@ -32,8 +32,8 @@ public interface ClusterAgentsFactory {
   /**
    * Construct and return the references or return the references to the previously constructed
    * {@link ClusterParticipant}(s). We extend this method to support multiple participants on same node. In some special
-   * cases we require a data node to participate into multiple clusters and each participant interacts with corresponding
-   * cluster independently.
+   * cases (i.e. migrating cluster to another Zookeeper), we require a data node to participate into multiple ZK clusters
+   * and each participant interacts with corresponding ZK independently.
    * @return a list of {@link ClusterParticipant}(s) on current node.
    */
   List<ClusterParticipant> getClusterParticipants() throws IOException;
--- a/ambry-api/src/main/java/com/github/ambry/clustermap/ClusterAgentsFactory.java
+++ b/ambry-api/src/main/java/com/github/ambry/clustermap/ClusterAgentsFactory.java
@@ -15,6 +15,7 @@
 package com.github.ambry.clustermap;
 
 import java.io.IOException;
+import java.util.List;
 
 
 /**
@@ -29,9 +30,12 @@ public interface ClusterAgentsFactory {
   ClusterMap getClusterMap() throws IOException;
 
   /**
-   * Construct and return the reference or return the reference to the previously constructed
-   * {@link ClusterParticipant}
+   * Construct and return the references or return the references to the previously constructed
+   * {@link ClusterParticipant}(s). We extend this method to support multiple participants on same node. In some special
+   * cases we require a data node to participate into multiple clusters and each participant interacts with corresponding
+   * cluster independently.
+   * @return a list of {@link ClusterParticipant}(s) on current node.
    */
-  ClusterParticipant getClusterParticipant() throws IOException;
+  List<ClusterParticipant> getClusterParticipants() throws IOException;
 }
 
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
@@ -14,11 +14,11 @@
 package com.github.ambry.clustermap;
 
 import com.github.ambry.network.Port;
+import com.github.ambry.utils.Utils;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
@@ -72,8 +72,8 @@ public class ClusterMapUtils {
   static final String AVAILABLE_STR = "AVAILABLE";
   static final String READ_ONLY_STR = "RO";
   static final String READ_WRITE_STR = "RW";
-  static final String ZKCONNECTSTR_STR = "zkConnectStr";
-  static final String ZKCONNECTSTR_DELIMITER = ",";
+  static final String ZKCONNECT_STR = "zkConnectStr";
+  static final String ZKCONNECT_STR_DELIMITER = ",";
   static final String ZKINFO_STR = "zkInfo";
   static final String DATACENTER_STR = "datacenter";
   static final String DATACENTER_ID_STR = "id";
@@ -155,10 +155,10 @@ public class ClusterMapUtils {
       String name = entry.getString(DATACENTER_STR);
       byte id = (byte) entry.getInt(DATACENTER_ID_STR);
       ReplicaType replicaType = entry.optEnum(ReplicaType.class, REPLICA_TYPE_STR, ReplicaType.DISK_BACKED);
-      String[] zkConnectStrs =
-          (replicaType == ReplicaType.DISK_BACKED) ? entry.getString(ZKCONNECTSTR_STR).split(ZKCONNECTSTR_DELIMITER)
-              : entry.optString(ZKCONNECTSTR_STR).split(ZKCONNECTSTR_DELIMITER);
-      DcZkInfo dcZkInfo = new DcZkInfo(name, id, Arrays.asList(zkConnectStrs), replicaType);
+      ArrayList<String> zkConnectStrs =
+          (replicaType == ReplicaType.DISK_BACKED) ? Utils.splitString(entry.getString(ZKCONNECT_STR),
+              ZKCONNECT_STR_DELIMITER) : Utils.splitString(entry.optString(ZKCONNECT_STR), ZKCONNECT_STR_DELIMITER);
+      DcZkInfo dcZkInfo = new DcZkInfo(name, id, zkConnectStrs, replicaType);
       dataCenterToZkAddress.put(dcZkInfo.dcName, dcZkInfo);
     }
     return dataCenterToZkAddress;
@@ -243,7 +243,7 @@ public class ClusterMapUtils {
    */
   static long getXid(InstanceConfig instanceConfig) {
     String xid = instanceConfig.getRecord().getSimpleField(XID_STR);
-    return xid == null ? DEFAULT_XID : Long.valueOf(xid);
+    return xid == null ? DEFAULT_XID : Long.parseLong(xid);
   }
 
   /**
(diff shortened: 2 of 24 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/DatacenterInitializer.java
```diff
@@ -158,7 +158,10 @@ private void onInitializationFailure(Exception e) {
    * @throws Exception if something went wrong during startup
    */
   private DcInfo initializeHelixDatacenter() throws Exception {
-    String zkConnectStr = dcZkInfo.getZkConnectStr();
+    // For now, the first ZK endpoint (if there are more than one endpoints) will be adopted by default for initialization.
+    // If we really need multiple HelixClusterManagers on same node in the future, we can use separate zk connect str configs
+    // for HelixClusterManager and HelixParticipant.
```
justinlin-linkedin, line 163: This comment is confusing to me. DatacenterInitializer is not part of the ClusterParticipant. It's used by HelixClusterManager to fetch the instance config and fill the clustermap. Why do we need to support multiple of them?

ambry-api/src/main/java/com/github/ambry/clustermap/ClusterAgentsFactory.java
```diff
@@ -29,9 +30,12 @@
   ClusterMap getClusterMap() throws IOException;
 
   /**
-   * Construct and return the reference or return the reference to the previously constructed
-   * {@link ClusterParticipant}
+   * Construct and return the references or return the references to the previously constructed
+   * {@link ClusterParticipant}(s). We extend this method to support multiple participants on same node. In some special
```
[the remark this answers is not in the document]
jsjtzyy: sure, will add

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
```diff
@@ -165,9 +164,11 @@ public void startup() throws InstantiationException {
       }
 
       StoreKeyFactory storeKeyFactory = Utils.getObj(storeConfig.storeKeyFactory, clusterMap);
+      // TODO make StorageManager, ReplicationManager, CloudToStoreReplicationManager and StatsManager support multiple participants
+      // For now, we assume there is only one element in clusterParticipants list.
```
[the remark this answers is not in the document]
jsjtzyy: Good question. ReplicaitonManager may not need actions against multiple participants but StatsManager and StorageManager are supposed to interact with multiple participants to upload local info to multiple clusters (i.e sealing partition info and stats info)

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
```diff
@@ -152,9 +155,10 @@ public static String getInstanceName(String host, Integer port) {
       String name = entry.getString(DATACENTER_STR);
       byte id = (byte) entry.getInt(DATACENTER_ID_STR);
       ReplicaType replicaType = entry.optEnum(ReplicaType.class, REPLICA_TYPE_STR, ReplicaType.DISK_BACKED);
-      String zkConnectStr = (replicaType == ReplicaType.DISK_BACKED) ? entry.getString(ZKCONNECTSTR_STR)
-          : entry.optString(ZKCONNECTSTR_STR);
-      DcZkInfo dcZkInfo = new DcZkInfo(name, id, zkConnectStr, replicaType);
+      String[] zkConnectStrs =
+          (replicaType == ReplicaType.DISK_BACKED) ? entry.getString(ZKCONNECTSTR_STR).split(ZKCONNECTSTR_DELIMITER)
+              : entry.optString(ZKCONNECTSTR_STR).split(ZKCONNECTSTR_DELIMITER);
```
cgtz, line 160: this can result in a NPE, since we use optString for cloud backed replicas.

Instead, you could do `Utils.splitString(entry.optString(ZKCONNECT_STR, ""), ZKCONNECTSTR_DELIMITER)`

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
```diff
@@ -152,9 +155,10 @@ public static String getInstanceName(String host, Integer port) {
       String name = entry.getString(DATACENTER_STR);
       byte id = (byte) entry.getInt(DATACENTER_ID_STR);
       ReplicaType replicaType = entry.optEnum(ReplicaType.class, REPLICA_TYPE_STR, ReplicaType.DISK_BACKED);
-      String zkConnectStr = (replicaType == ReplicaType.DISK_BACKED) ? entry.getString(ZKCONNECTSTR_STR)
-          : entry.optString(ZKCONNECTSTR_STR);
-      DcZkInfo dcZkInfo = new DcZkInfo(name, id, zkConnectStr, replicaType);
+      String[] zkConnectStrs =
+          (replicaType == ReplicaType.DISK_BACKED) ? entry.getString(ZKCONNECTSTR_STR).split(ZKCONNECTSTR_DELIMITER)
```
cgtz, line 159: Use `Utils.splitString` here?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -197,7 +197,9 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
   private HelixManager initializeHelixManagerAndPropertyStoreInLocalDC(Map<String, DcZkInfo> dataCenterToZkAddress,
       String instanceName, HelixFactory helixFactory) throws Exception {
     DcZkInfo dcZkInfo = dataCenterToZkAddress.get(clusterMapConfig.clusterMapDatacenterName);
-    String zkConnectStr = dcZkInfo.getZkConnectStr();
+    // For now, the first ZK endpoint (if there are more than one endpoints) will be adopted by default. We can update
+    // this if we really need to initialize mulitiple HelixManager in local dc.
+    String zkConnectStr = dcZkInfo.getZkConnectStrs().get(0);
```
cgtz, line 202: this needs to go below line 203 since CLOUD_BACKED replicas do not have a zk connect string

ambry-clustermap/src/main/java/com/github/ambry/clustermap/DatacenterInitializer.java
jsjtzyy: True, we don't need support multiple HelixClusterManager(which is also spectator in the context of Helix). I added the comment to clarify that we have to pick one (and only one) ZK endpoint to connect, as `clusterMapDcsZkConnectStrings` contains several ZK endpoints in each dc to support multiple participants. I will remove the comment and make it clear.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
jsjtzyy: Yeah, `Utils.splitString` is better. I previously used `split` with the impression that `optString` returns empty `""` if not found, which doesn't cause NPE. Anyway, I will go with `Utils.splitString`.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
jsjtzyy: sure

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
cgtz: Oh interesting, you are right about it returning empty string instead of null:
```
    public String optString(String key) {
        return this.optString(key, "");
    }
```
You can ignore the comment about the NPE. Still it will be good to use Utils.splitString since it will filter out the empty string. i.e. return the list `[]` instead of `[""]`

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -197,10 +197,12 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
   private HelixManager initializeHelixManagerAndPropertyStoreInLocalDC(Map<String, DcZkInfo> dataCenterToZkAddress,
       String instanceName, HelixFactory helixFactory) throws Exception {
     DcZkInfo dcZkInfo = dataCenterToZkAddress.get(clusterMapConfig.clusterMapDatacenterName);
-    String zkConnectStr = dcZkInfo.getZkConnectStr();
     if (dcZkInfo.getReplicaType() == ReplicaType.CLOUD_BACKED) {
       return null;
     }
+    // For now, the first ZK endpoint (if there are more than one endpoints) will be adopted by default. We can update
+    // this if we really need to initialize multiple HelixManagers in local dc.
```
justinlin-linkedin, line 204: I don't think we need to support multiple helix instances here, maybe we should change the comments.

jsjtzyy: hmm, I don't see any use cases for now, will remove the comment.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipantMetrics.java
```diff
@@ -32,26 +32,27 @@
   // no need to record exact number of "dropped" partition, a counter to track partition-dropped events would suffice
   final Counter partitionDroppedCount;
 
-  HelixParticipantMetrics(MetricRegistry metricRegistry) {
+  HelixParticipantMetrics(MetricRegistry metricRegistry, String zkConnectStr) {
```
lightningrob, line 35: Minor: my only concern here is that the metric names will look convoluted.  But I don't have a better suggestion either for making them distinct.

lightningrob: Perhaps in the common case where there is only one participant, you can skip adding the zkConnectStr to the metrics (assuming the caller knows this).

ambry-clustermap/src/test/java/com/github/ambry/clustermap/HelixParticipantTest.java
```diff
@@ -301,21 +307,45 @@ public void testBadCases() throws IOException {
...
+    zkInfoJson.put(ClusterMapUtils.ZKCONNECT_STR, "localhost:2199" + ZKCONNECT_STR_DELIMITER + "localhost:2299");
+    zkInfosJson.put(zkInfoJson);
+    JSONObject jsonObject = new JSONObject().put(ClusterMapUtils.ZKINFO_STR, zkInfosJson);
+    props.setProperty("clustermap.dcs.zk.connect.strings", jsonObject.toString(2));
+    ClusterMapConfig clusterMapConfig = new ClusterMapConfig(new VerifiableProperties(props));
+    List<ClusterParticipant> participants =
+        new HelixClusterAgentsFactory(clusterMapConfig, helixManagerFactory).getClusterParticipants();
+    assertEquals("Number of participants is not expected", 2, participants.size());
+    // restore previous setup
+    props.setProperty("clustermap.dcs.zk.connect.strings", zkJson.toString(2));
```
lightningrob, line 337: Minor: It's generally better to let either the test constructor or @Before method do this kind of initialization in case an exception happens in test case.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipantMetrics.java
jsjtzyy: Good idea, will try to skip zkConnectStr if there is only one participant.

PR #1474, opened 2020-04-14
Supporting Undelete in BlobStoreStats class

justinlin-linkedin:
Adding code to support Undelete in BlobStoreStats class.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -78,7 +78,7 @@ class BlobStoreStats implements StoreStats, Closeable {
   private final StoreMetrics metrics;
   private final ReentrantLock scanLock = new ReentrantLock();
   private final Condition waitCondition = scanLock.newCondition();
-  private final Queue<Pair<IndexValue, IndexValue>> recentEntryQueue = new LinkedBlockingQueue<>();
+  private final Queue<RecentQueueItem> recentEntryQueue = new LinkedBlockingQueue<>();
   private final AtomicInteger queueEntryCount = new AtomicInteger(0);
   private final AtomicBoolean enabled = new AtomicBoolean(true);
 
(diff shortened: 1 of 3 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -1062,18 +1090,46 @@ private void forwardScan(Offset startOffset, Offset endOffset) throws StoreExcep
...
-              } else if (indexValue.isFlagSet(IndexValue.Flags.Delete_Index)) {
-                processNewDelete(newScanResults, indexValue, originalPut);
-              } else if (indexValue.isFlagSet(IndexValue.Flags.Ttl_Update_Index)) {
+              } else if (indexValue.isDelete()) {
+                if (indexValue.getLifeVersion() == (short) 0) {
+                  processNewDelete(newScanResults, indexValue, originalPut, originalPut);
+                } else {
+                  List<IndexValue> allValues = index.findAllIndexValuesForKey(entry.getKey(), null);
+                  IndexValue latestValue = null;
+                  for (int i = allValues.size() - 1; i >= 0; i++) {
```
zzmao, line 1099: i++?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -388,7 +389,20 @@ void handleNewDeleteEntry(IndexValue deleteValue, IndexValue originalPutValue) {
    */
   void handleNewTtlUpdateEntry(IndexValue ttlUpdateValue, IndexValue originalPutValue) {
     if (recentEntryQueueEnabled) {
-      recentEntryQueue.offer(new Pair<>(ttlUpdateValue, originalPutValue));
+      recentEntryQueue.offer(new RecentQueueItem(ttlUpdateValue, originalPutValue, null));
```
jsjtzyy, line 392: I think the passed in `originalPutValue` may not be the actual original put because `updateTtl` method in BlobStore.java doesn't guarantee the found index value is PUT, could be UNDELETE as well. Double check this please.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
zzmao: This is my only comment. Others looks good.

PR #1475, opened 2020-04-15
Create config for datacenter from which cloud replication is allowed.

ankagrawal:
We need a datacenter to replicate from. For DR, we currently deploy VCR nodes on same datacenter as ambry-server. So we use local datacenter replicas as remote replicas.

For serving data from cloud, we must specify the ambry-server datacenter(s) from which to replicate. This PR introduces configs and code changes for getting the peer replicas for cloud.


ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -79,6 +77,12 @@ public VcrReplicationManager(VerifiableProperties properties, CloudConfig cloudC
     this.cloudStorageCompactor =
         cloudConfig.cloudBlobCompactionEnabled ? new CloudStorageCompactor(cloudDestination, cloudConfig,
             partitionToPartitionInfo.keySet(), vcrMetrics) : null;
+    // We need a datacenter to replicate from. For DR, we currently deploy VCR nodes on same datacenter as ambry-server.
+    // For serving data from cloud, we must specify the ambry-server datacenter(s) from which to replicate.
+    if (!cloudConfig.cloudIsVcr && cloudConfig.vcrCrossColoReplicationPeerDatacenters.isEmpty()) {
```
jsjtzyy, line 82: I don't have much context of cloud but I am trying to understand the config `cloudIsVcr`, so there are two types of nodes related to Cloud. One is VCR, in my impression, it is supposed to replicate blobs from regular ambry server to Azure (correct me if I am wrong). What is the other type of cloud node?

lightningrob: cloudIsVcr is used to tell the CloudBlobStore if it is being used by the VCR (true) or by the frontend (live serving).  VcrReplicationManager should never be created when isVcr = false

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -45,6 +50,7 @@
   public static final String VCR_PROXY_HOST = "vcr.proxy.host";
   public static final String VCR_PROXY_PORT = "vcr.proxy.port";
   public static final String VCR_CLUSTER_SPECTATOR_FACTORY_CLASS = "vcr.cluster.spectator.factory.class";
+  public static final String VCR_CROSS_COLO_REPLICATION_PEER_DATACENTERS = "vcr.cross.colo.replication.peer.datacenter";
```
lightningrob, line 53: This is very long.  How about VCR_SOURCE_DATACENTERS and a comment explaining it?

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -262,6 +268,13 @@
   @Default(DEFAULT_VCR_CLUSTER_SPECTATOR_FACTORY_CLASS)
   public final String vcrClusterSpectatorFactoryClass;
 
+  /**
+   * Comma separated set of datacenters which can act as peer for cross colo replication.
+   */
+  @Config(VCR_CROSS_COLO_REPLICATION_PEER_DATACENTERS)
+  @Default("")
+  public final Set<String> vcrCrossColoReplicationPeerDatacenters;
```
lightningrob, line 276: vcrSourceDataCenters

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
jsjtzyy: I see. Then, I have some difficulties understanding the `cloudIsVcr` being checked in `VcrReplicationManager`. I suppose it should be VCR already before instantiating the `VcrReplicationManager`.

lightningrob, line 82: If cloudIsVcr is false here, that alone would be an illegal state.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -79,6 +77,12 @@ public VcrReplicationManager(VerifiableProperties properties, CloudConfig cloudC
...
+      throw new IllegalStateException(
+          "Either this should be a vcr replicator OR one or more vcr cross colo replication peer datacenter should be specified");
```
lightningrob, line 84: Simplify to "One or more VCR cross colo replication peer datacenter should be specified"

jsjtzyy:
I guess the intention of this PR is to allow VCR to selectively replicate with remote datacenters. Replicating with local dc is always enabled. (Correct me if I am wrong)

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -236,4 +240,18 @@ public long getRemoteReplicaLagFromLocalInBytes(PartitionId partitionId, String
...
     return -1;
   }
+
+  /**
+   * Check if replication is allowed from given datacenter.
+   * @param datacenterName datacenter name to check.
+   * @return true if replication is allowed. false otherwise.
+   */
+  private boolean shouldReplicateFromDc(String datacenterName) {
+    if (cloudConfig.cloudIsVcr) {
```
lightningrob, line 250: cloudIsVcr will be true in both cases.  Perhaps it's time to add another config property like isBackup?
But I think we can use the new logic for both scenarios, and update the DR config to tell it which DC to use.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
ankagrawal: I was under the wrong impression that cloudIsVcr would be true for all cases related to DR, and false otherwise (even though I now realize that comments in CloudBlobStore say so). 
Since thats not the case, I guess the suggestion to always use vcrSourceDatacenters (vcrCrossColoReplicationPeerDatacenters) to get the list of DCs to replicate from makes sense.
I will make the required changes.

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
ankagrawal: sure.

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
ankagrawal: ok

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
ankagrawal: true

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
ankagrawal: ok

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
ankagrawal: sure.. as of now I don't see any other explicit need for a separate config for backup, so I will use the same config for all cases. We can revisit if we need a separate config for backup if an explicit need arises.

PR #1476, opened 2020-04-16
Command line tool to reset replica tokens in Azure

lightningrob:


ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
```diff
@@ -128,25 +133,25 @@
...
-  @Config("clustermap.host.name")
+  @Config(CLUSTERMAP_HOST_NAME)
   public final String clusterMapHostName;
 
   /**
    * The port number associated with this node.
    */
-  @Config("clustermap.port")
+  @Config(CLUSTERMAP_PORT)
   @Default("null")
```
ankagrawal, line 155: Maybe we can either remove Default from here, or change it to empty string.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -129,6 +129,10 @@
     batchTimeout = null;
   }
 
+  public BlobServiceClient getStorageClient() {
```
ankagrawal, line 132: javadocs.

ambry-tools/src/main/java/com/github/ambry/tools/admin/AzureTokenResetTool.java
```diff
@@ -0,0 +1,93 @@
...
+      int tokensDeleted = resetTokens(clusterName);
+      logger.info("Deleted tokens for {} partitions", tokensDeleted);
+    } catch (Exception ex) {
+      ex.printStackTrace();
+      logger.error("Command {} failed", commandName, ex);
+      System.exit(1);
+    }
+  }
+
+  // Reset the offset token by deleting the blob from the container
```
ankagrawal, line 73: Nit: Not sure about how we maintain javadocs for tools. But if we maintain it like prod code, then we might need javadocs for methods in this class.

ambry-tools/src/main/java/com/github/ambry/tools/admin/AzureTokenResetTool.java
```diff
@@ -0,0 +1,93 @@
...
+  public static int resetTokens(String containerPrefix) throws BlobStorageException {
+    AtomicInteger tokensDeleted = new AtomicInteger(0);
```
ankagrawal, line 75: Curious as to why we need AtomicInteger here? Does forEachRemaining for PagedIterable's returned by listBlobContainers, return a parallel stream?

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
```diff
@@ -282,10 +287,10 @@ public ClusterMapConfig(VerifiableProperties verifiableProperties) {
     clusterMapClusterChangeHandlerType =
         verifiableProperties.getString("clustermap.cluster.change.handler.type", "SimpleClusterChangeHandler");
     clusterMapDcsZkConnectStrings = verifiableProperties.getString("clustermap.dcs.zk.connect.strings", "");
```
jsjtzyy, line 289: minor: in future PR, we can replace all config names with string constants (like what you did in this PR to make them consistent)

lightningrob: Yes.  I didn't do it in this PR to keep the scope limited.

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
lightningrob: I don't know why it's there, but I'd prefer not to touch it right now.

ambry-tools/src/main/java/com/github/ambry/tools/admin/AzureTokenResetTool.java
lightningrob: I had it as an int but the compiler complained that I couldn't pass a non-final variable to a lambda, and this seemed like the easiest way to work around it.

PR #1477, opened 2020-04-17
Use DelayedAutoRebalancer in Helix VCR cluster

lightningrob:
Add tool option to update existing resources

### DIFF
--- a/ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
+++ b/ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
@@ -112,7 +112,6 @@ public class HelixVcrPopulateTool {
         System.out.println("Updating cluster: " + destClusterName + "by checking " + srcClusterName);
         updateResourceAndPartition(srcZkString, srcClusterName, destZkString, destClusterName, dryRun);
       } else {
-        // Just update cluster config, not resources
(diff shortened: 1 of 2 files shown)


ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -92,11 +101,26 @@ public static void main(String[] args) {
...
+        String[] srcZkAndCluster = options.valueOf(srcOpt).split(SEPARATOR);
+        if (srcZkAndCluster.length != 2) {
+          errorAndExit("src argument must have form 'zkString/clusterName'");
+        }
+        String srcZkString = srcZkAndCluster[0];
+        String srcClusterName = srcZkAndCluster[1];
+        System.out.println("Updating cluster: " + destClusterName + "by checking " + srcClusterName);
+        updateResourceAndPartition(srcZkString, srcClusterName, destZkString, destClusterName, dryRun);
+      } else {
+        // Just update cluster config, not resources
```
jsjtzyy, line 115: I am a little confused about this comment because I see resource ideal state is updated in the following lines.

lightningrob: Yeah, I meant the resources would not be updated with any new partitions, though their idealstate is still updated.  I removed the comment, it's not needed.

PR #1479, opened 2020-04-21
Re-commit PR 1472 to support multiple participants on server side

jsjtzyy:
This PR fixes the conflict and keeps the same logic in PR 1472.


jsjtzyy:
Addressed Rob's comments and should be ready for review or be merged

PR #1480, opened 2020-04-21
Complete UndeleteOperation and add UndeleteOperationTracker

justinlin-linkedin:
Finish the implementation of UndeleteOperation.

1. Add handleResponse and poll method
2. Add UndeleteOperationTracker
3. Add tests for UndeleteOperation and UndeleteOperationTracker.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -70,6 +70,7 @@ public class RouterConfig {
   public static final String ROUTER_CRYPTO_JOBS_WORKER_COUNT = "router.crypto.jobs.worker.count";
   public static final String ROUTER_TTL_UPDATE_REQUEST_PARALLELISM = "router.ttl.update.request.parallelism";
   public static final String ROUTER_TTL_UPDATE_SUCCESS_TARGET = "router.ttl.update.success.target";
+  public static final String ROUTER_UNDELETE_REQUEST_PARALLELISM = "router.undelete.request.parallelism";
   public static final String ROUTER_USE_GET_BLOB_OPERATION_FOR_BLOB_INFO =
       "router.use.get.blob.operation.for.blob.info";
   public static final String ROUTER_OPERATION_TRACKER_CUSTOM_PERCENTILES =
@@ -325,6 +326,13 @@ public class RouterConfig {
   public final int routerTtlUpdateSuccessTarget;
 
   /**
+   * The maximum number of parallel requests issued at a time by the undelete manager for a blob.
+   */
+  @Config(ROUTER_UNDELETE_REQUEST_PARALLELISM)
+  @Default("3")
+  public final int routerUndeleteRequestParallelism;
+
+  /**
    * If this config is set to {@code true} the router will use {@code GetBlobOperation} instead of
    * {@code GetBlobInfoOperation} for {@code getBlobInfo} calls. This allows the router to correct some blob size
    * corruptions that may have arisen from using older versions of {@code BlobIdTransformer} with the downside of
@@ -541,6 +549,8 @@ public class RouterConfig {
         verifiableProperties.getIntInRange(ROUTER_TTL_UPDATE_REQUEST_PARALLELISM, 3, 1, Integer.MAX_VALUE);
     routerTtlUpdateSuccessTarget =
         verifiableProperties.getIntInRange(ROUTER_TTL_UPDATE_SUCCESS_TARGET, 2, 1, Integer.MAX_VALUE);
+    routerUndeleteRequestParallelism =
+        verifiableProperties.getIntInRange(ROUTER_UNDELETE_REQUEST_PARALLELISM, 3, 1, Integer.MAX_VALUE);
     routerUseGetBlobOperationForBlobInfo =
         verifiableProperties.getBoolean(ROUTER_USE_GET_BLOB_OPERATION_FOR_BLOB_INFO, false);
     List<String> customPercentiles =
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -122,10 +122,11 @@ class SimpleOperationTracker implements OperationTracker {
    * |     PUT          |           N                     |       N - 1      |
    * |    DELETE        |          3~N                    |         2        |
    * |   TTLUpdate      |          3~N                    |         2        |
-   * |   UNDELETE       |          3~N                    |         2        |
+   * |   UNDELETE       |          3~N                    |  Global Quorum   |
    *  -----------------------------------------------------------------------
    *  Note: for now, we still use 3 as parallelism for DELETE/TTLUpdate/UNDELETE even though there are N eligible replicas, this
    *        can be adjusted to any number between 3 and N (inclusive)
+   *        For Undelete, it needs to reach global quorum to succeed. A dedicated operation tracker is created to check that.
    * @param routerConfig The {@link RouterConfig} containing the configs for operation tracker.
    * @param routerOperation The {@link RouterOperation} which {@link SimpleOperationTracker} is associated with.
    * @param partitionId The partition on which the operation is performed.
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -112,6 +112,7 @@ class SimpleOperationTracker implements OperationTracker {
    * |      GET         | STANDBY, LEADER, BOOTSTRAP, INACTIVE |
    * |    DELETE        | STANDBY, LEADER, BOOTSTRAP           |
    * |   TTLUpdate      | STANDBY, LEADER, BOOTSTRAP           |
+   * |   UNDELETE       | STANDBY, LEADER, BOOTSTRAP           |
    *  ---------------------------------------------------------
    * Following are dynamic configs when replica state is taken into consideration: (N is number of eligible replicas)
    *  -----------------------------------------------------------------------
@@ -121,8 +122,9 @@ class SimpleOperationTracker implements OperationTracker {
    * |     PUT          |           N                     |       N - 1      |
    * |    DELETE        |          3~N                    |         2        |
    * |   TTLUpdate      |          3~N                    |         2        |
+   * |   UNDELETE       |          3~N                    |         2        |
    *  -----------------------------------------------------------------------
-   *  Note: for now, we still use 3 as parallelism for DELETE/TTLUpdate even though there are N eligible replicas, this
+   *  Note: for now, we still use 3 as parallelism for DELETE/TTLUpdate/UNDELETE even though there are N eligible replicas, this
    *        can be adjusted to any number between 3 and N (inclusive)
    * @param routerConfig The {@link RouterConfig} containing the configs for operation tracker.
    * @param routerOperation The {@link RouterOperation} which {@link SimpleOperationTracker} is associated with.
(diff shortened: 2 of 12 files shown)


ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -121,8 +122,9 @@
    * |     PUT          |           N                     |       N - 1      |
    * |    DELETE        |          3~N                    |         2        |
    * |   TTLUpdate      |          3~N                    |         2        |
+   * |   UNDELETE       |          3~N                    |         2        |
```
zzmao, line 125: According to the design doc, the quorum target for the UNDELETE should be 2 per DC. Is there a follow up PR for this?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -176,6 +179,15 @@
         eligibleReplicas = getEligibleReplicas(partitionId, null,
             EnumSet.of(ReplicaState.BOOTSTRAP, ReplicaState.STANDBY, ReplicaState.LEADER));
         break;
+      case UndeleteOperation:
+        diskParallelism = routerConfig.routerUndeleteRequestParallelism;
+        crossColoEnabled = true;
+        eligibleReplicas = getEligibleReplicas(partitionId, null,
+            EnumSet.of(ReplicaState.BOOTSTRAP, ReplicaState.STANDBY, ReplicaState.LEADER));
```
cgtz, line 186: if there are machines that are offline, they won't be included in this set. Is it okay to ignore them? I believe this policy currently says that an undelete needs 100% success from replicas that are currently up.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
```diff
@@ -0,0 +1,139 @@
...
+ * An implementation of {@link OperationTracker}. It internally maintains the status of a corresponding operation, and
+ * returns information that decides if the operation should continue or terminate.
+ *
+ * This OperationTracker only works for {@link UndeleteOperation}. Since an {@link UndeleteOperation} requires a global
+ * quorum for success, a map of datacenter to number of success requests is used to keep track of how many success requests
+ * tracker has seen in each datacenter. Once it reaches quorum in all the datacenter, it returns true for success. Or if
+ * any datacenter see failure requests reach the quorum, it return true for failure.
+ */
+public class UndeleteOperationTracker extends SimpleOperationTracker {
+  private Map<String, Integer> numReplicasInDcs = new HashMap<>();
```
cgtz, line 34: final

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -324,6 +325,13 @@
   @Default("2")
   public final int routerTtlUpdateSuccessTarget;
 
+  /**
+   * The maximum number of parallel requests issued at a time by the undelete manager for a blob.
+   */
+  @Config(ROUTER_UNDELETE_REQUEST_PARALLELISM)
+  @Default("3")
+  public final int routerUndeleteRequestParallelism;
```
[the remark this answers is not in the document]
justinlin-linkedin: it's the same meaning as other request's parallelism. If we have more than one DC, then it's on all DCs.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
justinlin-linkedin: non-eligible hosts would be considered as a failure. for instance, if a dc has three hosts, and one is offline, then starting from the beginning, we already have a failure hosts. If we see another hosts returning a failure response, then the operation is considered as failed. But if the other two both return success, then undelete operation still is valid.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
```diff
@@ -0,0 +1,139 @@
...
+    if (!hasReachedGlobalQuorum(numReplicasInDcs, numEligibleReplicasInDcs)) {
+      throw new IllegalArgumentException(
+          "Eligible replicas are not sufficient for undelete operation for partition " + partitionId);
+    }
+    // Consider not-eligible hosts as failure
+    for (String dcName : numReplicasInDcs.keySet()) {
+      int totalNum = numReplicasInDcs.get(dcName);
+      if (numEligibleReplicasInDcs.containsKey(dcName)) {
+        int eligibleNum = numEligibleReplicasInDcs.get(dcName);
+        if (eligibleNum != totalNum) {
```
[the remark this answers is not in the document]
justinlin-linkedin: updated.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
```diff
@@ -0,0 +1,139 @@
...
+   * @return true if current numbers reach the global quorum.
+   */
+  private static boolean hasReachedGlobalQuorum(Map<String, Integer> totalNumberInDcs,
+      Map<String, Integer> currentNumberInDcs) {
+    boolean hasReached = true;
+    if (totalNumberInDcs.size() == currentNumberInDcs.size()) {
+      for (String dcName : totalNumberInDcs.keySet()) {
+        int totalNum = totalNumberInDcs.get(dcName);
+        int currentNum = currentNumberInDcs.get(dcName);
+        if (totalNum != 1 && currentNum <= totalNum / 2) {
```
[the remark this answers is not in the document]
justinlin-linkedin: you are right, updated.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
cgtz: Ah, I think I got confused by the setting of disk success target, but that seems to just be a dummy value and isn't used anywhere in UndeleteOperationTracker.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
```diff
@@ -0,0 +1,135 @@
...
+
+  @Override
+  public boolean hasFailed() {
+    return hasReachedAnyLocalQuorum(numReplicasInDcs, numFailureInDcs);
+  }
+
+  /**
+   * Return true if the {@code currentNumberInDcs}'s each value has reached the quorum of corresponding value
+   * in {@code totalNumberInDcs}.
+   * @param totalNumberInDcs The total number of members in each datacenter.
```
zzmao, line 93: total number of replicas?

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
```diff
@@ -0,0 +1,135 @@
...
+   * @param currentNumberInDcs The current number of members in each datacenter.
```
zzmao, line 94: member -> replica?

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
```diff
@@ -0,0 +1,135 @@
...
+   * @return true if current numbers reach the global quorum.
+   */
+  private static boolean hasReachedGlobalQuorum(Map<String, Integer> totalNumberInDcs,
+      Map<String, Integer> currentNumberInDcs) {
+    boolean hasReached = true;
+    if (totalNumberInDcs.size() == currentNumberInDcs.size()) {
+      for (String dcName : totalNumberInDcs.keySet()) {
+        int totalNum = totalNumberInDcs.get(dcName);
+        int currentNum = currentNumberInDcs.getOrDefault(dcName, 0);
```
zzmao, line 103: as discussed, adding some tests will be good.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
```diff
@@ -104,6 +151,89 @@ void poll(RequestRegistrationCallback<UndeleteOperation> requestRegistrationCall
...
+              LOGGER.trace("Cross colo request successful for remote replica {} in {} ", replica.getDataNodeId(),
+                  replica.getDataNodeId().getDatacenterName());
+              routerMetrics.crossColoSuccessCount.inc();
+            }
+            if (lifeVersion == null) {
+              // This is first successful response.
+              lifeVersion = undeleteResponse.getLifeVersion();
+              firstResponseReplicaId = replica;
+            } else {
+              if (lifeVersion.shortValue() != undeleteResponse.getLifeVersion()) {
```
jsjtzyy, line 202: minor: unnecessary unboxing for `lifeVersion.shortValue()`

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
```diff
@@ -104,6 +151,89 @@ void poll(RequestRegistrationCallback<UndeleteOperation> requestRegistrationCall
...
+                    firstResponseReplicaId, replica, lifeVersion, undeleteResponse.getLifeVersion());
+                LOGGER.error(message);
+                // this is a successful response and one that completes the operation regardless of whether the
+                // success target has been reached or not.
+                operationCompleted = true;
+                onErrorResponse(replica, new RouterException(message, RouterErrorCode.LifeVersionConflict));
+              }
+            }
+          } else if (serverError == ServerErrorCode.Disk_Unavailable) {
+            LOGGER.trace("Replica {} returned Disk_Unavailable for a delete request with correlationId : {} ", replica,
```
jsjtzyy, line 214: minor: `undelete request`

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
```diff
@@ -104,6 +151,89 @@ void poll(RequestRegistrationCallback<UndeleteOperation> requestRegistrationCall
...
+                undeleteRequest.getCorrelationId());
+            operationTracker.onResponse(replica, TrackedRequestFinalState.DISK_DOWN);
+            setOperationException(
+                new RouterException("Server returned: " + serverError, RouterErrorCode.AmbryUnavailable));
+            routerMetrics.routerRequestErrorCount.inc();
+            routerMetrics.getDataNodeBasedMetrics(replica.getDataNodeId()).undeleteRequestErrorCount.inc();
+          } else {
+            LOGGER.trace("Replica {} returned an error {} for a delete request with response correlationId : {} ",
```
jsjtzyy, line 222: `undelete request`

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
```diff
@@ -141,19 +361,19 @@ private int getPrecedenceLevel(RouterErrorCode routerErrorCode) {
...
+        return 3;
       case UnexpectedInternalError:
-        return 5;
+        return 4;
       case OperationTimedOut:
-        return 6;
+        return 5;
       case BlobDoesNotExist:
+        return 6;
+      case BlobNotDeleted:
```
jsjtzyy, line 376: I feel `BlobNotDeleted` may have higher precedence level than `OperationTimedOut`

justinlin-linkedin: this one is quite tricky. Two errors in one dc would render undelete operation a failure. Say if we have one `BlobNotDeleted` and one `OperationTimeOut`, it might be because there is only one host that doesn't have this blob deleted, but the other two are deleted and unfortunately one of these two returns `OperationTimeOut`. In this case, I think it's better to return `OperationTimeOut`.

jsjtzyy: I think `BlobNotDeleted` means one server holds the blob but this blob is not deleted yet. If undelete succeeds on all servers except for one server that returns `BlobNotDeleted`, we still consider whole operation a success?

justinlin-linkedin: you are right, `BlobNotDeleted` does mean that. And `BlobNotDeleted` is not a terminating error, which means if we see this error, the undelete operation will continue. If undelete operation sees more than one error on one dc (assume there are 3 replicas), then it will fail the operation. But if undelete operation only sees one `BlobNotDeleted` and two success in one dc, then it's still considered a success.

PR #1481, opened 2020-04-21
Limit the number of partitions per ReplicaMetadataRequest

jsjtzyy:
Two peer nodes may share many partitions. Imagine an edge case where
one node has been down for several weeks and is being brought up. If
the two nodes share 250 partitions and each partition has one 4 MB new
blob, in worst case, the previously down node will attempt to fetch 250
blobs which is roughly 1 GB in single get request. This causes memory
pressure on source node and may induce Out-of-Memory issue. This pr
introduces a config to limit number of partitions per replication request
with aim to split partitions into small batches and replicate in sequential
order.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -114,6 +114,15 @@ public class ReplicationConfig {
   public final long replicationFetchSizeInBytes;
 
   /**
+   * The maximum number of partitions on remote node can be replicated with in a single replication request. If local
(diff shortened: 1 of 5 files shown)


jsjtzyy:
> LGTM, just a nitpicking
> 
> 1. Do we have to keep option open for non max, like using 0 as a way to indicate there is no chunking
> 2. 20 seems large to me. In production, we have some partition requests reaching 400MB right? I think it makes more sense to comment the reason of choosing 20 here. So later if we want to change it, we know the criteria.

1. It's actually a good idea, will make the changes accordingly.
2. This config will be adopted together with the fix in findEntriesSince. After that fix, `fetchSizeInBytes` should be correctly honored when PUT, TTLUpdate are present in same segment.

PR #1484, opened 2020-04-22
Use conservative findEntriesSince size limit when update and put are both in the journal

cgtz:
In the case where both a put and a TTL update show up when finding
entries in the journal, the most recent index entry found will always be
the ttl update entry. This means that the size of the TTL update entry
will be checked against the size limit of the findEntriesSince query.
However, a replicator will often have to fetch the original blob
anyways if they don't already have the put entry, which means that
they could do a much larger batch get than expected, leading to
excessive memory consumption on the batch-get server side.

To address this issue, this commit adds logic to check if there is a
put that the caller of findEntriesSince has likely not encountered yet,
and, if so, counts its size towards the limit. This behavior is not used
for deletes, since there is no need to fetch a deleted blob's data.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -17,6 +17,7 @@ import com.codahale.metrics.Timer;
 import com.github.ambry.config.StoreConfig;
 import com.github.ambry.replication.FindToken;
 import com.github.ambry.replication.FindTokenType;
+import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.io.File;
@@ -59,7 +60,10 @@ class PersistentIndex {
    * Represents the different types of index entries.
    */
   enum IndexEntryType {
-    TTL_UPDATE, PUT, DELETE, UNDELETE
+    TTL_UPDATE,
+    PUT,
+    DELETE,
+    UNDELETE
   }
 
   static final short VERSION_0 = 0;
@@ -665,7 +669,8 @@ class PersistentIndex {
    * @param fileSpan {@link FileSpan} which specifies the range within which search should be made.
    * @param types the types of {@link IndexEntryType} to look for.
    * @param indexSegments the map of index segment start {@link Offset} to {@link IndexSegment} instances
-   * @return The list of the {@link IndexValue}s for {@code key} conforming to one of the types {@code types}.
+   * @return The list of the {@link IndexValue}s for {@code key} conforming to one of the types {@code types} ordered
+   *         from most to least recent.
    * @throws StoreException any error.
    */
   List<IndexValue> findAllIndexValuesForKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryType> types,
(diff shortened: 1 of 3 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1162,14 +1167,12 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
                   + entries.size());
           Offset offsetEnd = offsetToStart;
           long currentTotalSizeOfEntries = 0;
+          Offset endOffsetOfSnapshot = getCurrentEndOffset(indexSegments);
           for (JournalEntry entry : entries) {
```
justinlin-linkedin, line 1171: nit: Now we are no this, I think we can do a de-dup before going through the for loop.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1574,6 +1575,41 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
     }
   }
 
+  /**
+   * Helper method for getting a {@link MessageInfo} and an estimate of fetch size corresponding to the provided
+   * {@link JournalEntry}. This size may include the size of the put record if the put record is at or past the offset
+   * in the journal entry and the latest value in the index is a delete or undelete. In this case, we can make a safe
```
jsjtzyy, line 1581: nit: `ttlUpdate or undelete`

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -2455,6 +2457,7 @@ private void doFindEntriesSinceTest(StoreFindToken startToken, long maxTotalSize
       StoreFindToken expectedEndToken) throws StoreException {
     FindInfo findInfo = state.index.findEntriesSince(startToken, maxTotalSizeOfEntries);
     StoreFindToken token = (StoreFindToken) findInfo.getFindToken();
+    System.out.println("GOT token: " + token);
```
jsjtzyy, line 2460: removed?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
cgtz: addressed this with caching of messageinfos

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1574,6 +1573,52 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
...
+      }
+      messageInfo =
+          new MessageInfo(entry.getKey(), latestValue.getSize(), latestValue.isDelete(), latestValue.isTtlUpdate(),
+              latestValue.isUndelete(), latestValue.getExpiresAtMs(), null, latestValue.getAccountId(),
+              latestValue.getContainerId(), latestValue.getOperationTimeInMs(), latestValue.getLifeVersion());
+      messageInfoCache.put(entry.getKey(), messageInfo);
+    }
+    // We may add duplicate MessageInfos to the list here since the ordering of the list is used to calculate
+    // bytes read for the token. The duplicates will be cleaned up by eliminateDuplicates.
+    currentTotalSizeOfEntries.addAndGet(messageInfo.getSize());
```
jsjtzyy, line 1618: Logic looks good. I just need some clarification on PUT. If the entry is a PUT, how can we avoid its size being added `currentTotalSizeOfEntries` twice ?

cgtz: Sorry, I missed this question this morning. If the most recent entry is a put, the check on line 1604 will be false, so the size will only be accounted for once. The only case where extra size will be added if there are more than one entry found past `entry.getOffset()` and the latest entry is a TTL update or undelete and the earliest value at or past the journal entry is a put (e.g. not something else that doesn't indicate a need for blob fetching like a delete).

jsjtzyy: got it

PR #1485, opened 2020-04-24
Prevent compaction partial failure if possible, recover from it if necessary.

lightningrob:
Two complementary changes here:

1. Add shutdown hook to let compaction operations to complete.
2. Check for inconsistency in update operation and delete stray Cosmos record.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
@@ -21,7 +21,10 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -65,6 +73,29 @@ public void run() {
...
+   */
+  public void shutdown() {
+    shuttingDown.set(true);
+    logger.info("Compactor received shutdown request, waiting up to {} seconds for in-flight operations to finish",
+        shutDownTimeoutSecs);
+    try {
+      doneLatch.get().await(shutDownTimeoutSecs, TimeUnit.SECONDS);
+      logger.info("Compactor shut down successfully.");
+    } catch (InterruptedException ex) {
+      logger.warn("Timed out waiting for operations to finish.  If cloud provider uses separate stores for "
```
ankagrawal, line 87: Can we log a metric here? Will be easier to track. Otherwise this will get lost in log messages.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -65,6 +73,29 @@ public void run() {
...
 
+  /**
+   * Shut down the compactor waiting for in progress operations to complete.
+   */
+  public void shutdown() {
+    shuttingDown.set(true);
+    logger.info("Compactor received shutdown request, waiting up to {} seconds for in-flight operations to finish",
+        shutDownTimeoutSecs);
+    try {
+      doneLatch.get().await(shutDownTimeoutSecs, TimeUnit.SECONDS);
```
cgtz, line 84: When a latch times out, it actually doesn't throw an exception, it just returns false, so the return value has to be checked. The InterruptedException is only thrown if the thread is actually interrupted.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -248,7 +249,7 @@ public FindResult findEntriesSince(String partitionPath, FindToken findToken, lo
    * @param blobId The {@link BlobId} to update.
    * @param fieldName The metadata field to modify.
    * @param value The new value.
-   * @return {@code true} if the udpate succeeded, {@code false} if the metadata record was not found.
+   * @return {@code true} if the udpate succeeded, {@code false} if no update was needed.
```
cgtz, line 252: udpate -> update

Was the previous javadoc inaccurate? It seems like the old code would have also thrown an exception on not found.

lightningrob: yes I fixed the doc.

PR #1486, opened 2020-04-24
Make partition class more permissive for migrations

cgtz:
If a user is migrating between ambry clusters where the partition classes
are different, they may run into a situation where they either change
replicationPolicy on their container first, leading to puts in the
original cluster fail, or change the cluster used first, which leads to
puts in the new cluster failing. To aid in migration, this PR changes
the partition class selection behavior to fall back to the default
partition class. In this case, a metric will be emitted and a warn log
message will be written.

Additionally, this PR treats an empty string in replicationPolicy the
same as a null string, as this case was ambiguous before.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
@@ -559,7 +559,8 @@ public class ClusterMapUtils {
           }
           if (partitionsByReplicaCount == null) {
             throw new IllegalArgumentException(
-                "No partitions for default partition class " + partitionClass + " found");
+                "No partitions for partition class = '" + partitionClass + "' or default partition class = '"
+                    + defaultPartitionClass + "' found");
           }
           if (minimumReplicaCountRequired) {
             // get partitions with replica count >= min replica count specified in ClusterMapConfig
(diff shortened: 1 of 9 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
```diff
@@ -548,22 +551,26 @@ private boolean areAllLocalReplicasForPartitionUp(PartitionId partitionId) {
...
-        } else if (partitionIdsByClassAndLocalReplicaCount.containsKey(partitionClass)) {
+        } else {
           SortedMap<Integer, List<PartitionId>> partitionsByReplicaCount =
               partitionIdsByClassAndLocalReplicaCount.get(partitionClass);
+          if (partitionsByReplicaCount == null) {
+            partitionsByReplicaCount = partitionIdsByClassAndLocalReplicaCount.get(defaultPartitionClass);
+          }
+          if (partitionsByReplicaCount == null) {
+            throw new IllegalArgumentException(
+                "No partitions for default partition class " + partitionClass + " found");
```
jsjtzyy, line 562: minor: `"No partitions for partition class = '" + partitionClass + "' and default partition class = '" + defaultPartitionClass + "' found"`

ambry-router/src/main/java/com/github/ambry/router/PutManager.java
```diff
@@ -176,7 +176,7 @@ private String getPartitionClass(BlobProperties blobProperties) {
     Account account = accountService.getAccountById(blobProperties.getAccountId());
     if (account != null) {
       Container container = account.getContainerById(blobProperties.getContainerId());
-      if (container != null && container.getReplicationPolicy() != null) {
+      if (container != null && !Utils.isNullOrEmpty(container.getReplicationPolicy())) {
```
jsjtzyy, line 179: good catch

PR #1487, opened 2020-04-24
Return http status conflict when blob is undeleted or not deleted in undelete operation

justinlin-linkedin:
When the blob is already undeleted or not yet deleted, ambry frontend would return CONFLICT as http status.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/rest/RestServiceErrorCode.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RestServiceErrorCode.java
@@ -168,7 +168,12 @@ public enum RestServiceErrorCode {
   /**
    * Action not allowed
    */
-  NotAllowed;
+  NotAllowed,
+
+  /**
+   * Request conflicts with the current state of the server.
+   */
+  Conflict;
 
   /**
    * Gets the RestServiceErrorCode that corresponds to the {@code routerErrorCode}.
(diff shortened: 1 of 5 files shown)


ambry-api/src/main/java/com/github/ambry/rest/RestServiceErrorCode.java
```diff
@@ -199,8 +204,12 @@ public static RestServiceErrorCode getRestServiceErrorCode(RouterErrorCode route
         return ServiceUnavailable;
       case InsufficientCapacity:
         return InsufficientCapacity;
+      case BlobUndeleted:
+      case BlobNotDeleted:
+        return Conflict;
       case UnexpectedInternalError:
       case ChannelClosed:
+      case LifeVersionConflict:
```
cgtz, line 212: Why is LifeVersionConflict counted as an InternalServerError? does this error code mean that the operation has to be retried after replication converges?

justinlin-linkedin: First of all it's not a client error for sure. If we see this, then something is wrong with delete and undelete operations. Maybe we have some delete operations performed between undeletes. I would not recommend a retry if we see this, we should try to figure out what happened first.

cgtz: Thanks for the explanation. I thought that it would equate to a conflict http status at first

zzmao: Add some comments to explain why it's `InternalServerError` ?

PR #1488, opened 2020-04-27
Updating replica status(sealed/stopped) in multiple clusters

jsjtzyy:
As an effort for migrating Ambry from one ZK cluster to another, storage manager
should update replica status in both clusters to make sure they are consistent.
This PR allows storage manager to take in a list of participants and generate
corresponding replica status delegates. BlobStore will update sealing/unsealing
state via these delegates.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -943,7 +943,7 @@ public class BlobStore implements Store {
   }
 
   /**
-   * @return {@link ReplicaStatusDelegate} associated with this store
+   * @return a {@link ReplicaStatusDelegate}(s) associated with this store
    */
   public List<ReplicaStatusDelegate> getReplicaStatusDelegates() {
     return replicaStatusDelegates;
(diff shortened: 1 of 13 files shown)


ambry-store/src/test/java/com/github/ambry/store/StorageManagerTest.java
```diff
@@ -1236,15 +1274,23 @@ private void generateConfigs(boolean segmentedLog, boolean updateInstanceConfig)
...
     Set<ReplicaId> sealedReplicas = new HashSet<>();
     Set<ReplicaId> stoppedReplicas = new HashSet<>();
+    private Boolean setSealStateReturnVal;
+    private Boolean setStopStateReturnVal;
 
     MockClusterParticipant() throws IOException {
+      this(null, null);
+    }
+
+    MockClusterParticipant(Boolean setSealStateReturnVal, Boolean setStopStateReturnVal) {
```
lightningrob, line 1288: Add javadoc explaining what the return vals are and that they can be null.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -941,8 +945,8 @@ public long getEndPositionOfLastPut() throws StoreException {
   /**
    * @return {@link ReplicaStatusDelegate} associated with this store
    */
-  public ReplicaStatusDelegate getReplicaStatusDelegate() {
```
ankagrawal, line 944: javadoc needs to be changed here.

PR #1490, opened 2020-04-28
Introduce dc-level replication lag metrics

jsjtzyy:
Currently ReplicationMetrics tracks each remote replica lag from local by emitting a dedicated metric. As ambry grows, it brings more than one million metrics and it's hard to get a whole view of replication lag from remote dcs. This PR introduces a config that allows us to track remote replica lags dc level. That is, it provides statistics regarding replicas' lag in certain dc (min/max/avg) which gives a decent estimation of overall replication lag in each dc.
Also this PR introduces two new metrics on source node to track bytes rate of cross-colo fetch and metadata exchange. These metrics present a more accurate view of WAN bandwidth usage.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
@@ -56,7 +56,7 @@ public class VcrReplicationManager extends ReplicationEngine {
   private final VirtualReplicatorCluster virtualReplicatorCluster;
   private final CloudStorageCompactor cloudStorageCompactor;
   private final Map<String, Store> partitionStoreMap = new HashMap<>();
-  private final boolean trackPerPartitionLagInMetric;
+  private final boolean trackPerDatacenterLagInMetric;
 
   public VcrReplicationManager(CloudConfig cloudConfig, ReplicationConfig replicationConfig,
       ClusterMapConfig clusterMapConfig, StoreConfig storeConfig, StoreManager storeManager,
@@ -78,7 +78,7 @@ public class VcrReplicationManager extends ReplicationEngine {
     this.cloudStorageCompactor =
         cloudConfig.cloudBlobCompactionEnabled ? new CloudStorageCompactor(cloudDestination, cloudConfig,
             partitionToPartitionInfo.keySet(), vcrMetrics) : null;
-    trackPerPartitionLagInMetric = replicationConfig.replicationTrackRemoteFromLocalPerDatacenterLag;
+    trackPerDatacenterLagInMetric = replicationConfig.replicationTrackPerDatacenterLagFromLocal;
     // We need a datacenter to replicate from, which should be specified in the cloud config.
     if (cloudConfig.vcrSourceDatacenters.isEmpty()) {
       throw new IllegalStateException("One or more VCR cross colo replication peer datacenter should be specified");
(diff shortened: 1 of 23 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -77,6 +78,7 @@ public VcrReplicationManager(CloudConfig cloudConfig, ReplicationConfig replicat
     this.cloudStorageCompactor =
         cloudConfig.cloudBlobCompactionEnabled ? new CloudStorageCompactor(cloudDestination, cloudConfig,
             partitionToPartitionInfo.keySet(), vcrMetrics) : null;
+    trackPerPartitionLagInMetric = replicationConfig.replicationTrackRemoteFromLocalPerDatacenterLag;
```
cgtz, line 81: looks like this variable name may need adjusting since it is assigned to the per datacenter metric config

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -318,7 +318,7 @@ public void scheduleCompactionFailureTest() throws InterruptedException, IOExcep
    * @throws InterruptedException
    * @throws IOException
    */
-  @Test
+  //@Test
```
zzmao, line 321: Why comment it?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
jsjtzyy: ah, good catch, will change. 

For naming, on second thought, I probably won't change the name of `replicationTrackPerPartitionLagFromRemote` (this is introduced by @zzmao ). As you can see, I renamed it in this PR but since it requires to change config name/ sensor name in closed source, I feel like I would change it back. What do you think? @cgtz @zzmao

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
jsjtzyy: Good catch, it was previously disabled for debugging. I will remove it. Thanks.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
cgtz: It probably isn't used in too many places and isn't a required config so it may be okay to change if you think the new names are significantly more clear. Of course, you may need to keep both configs in src files temporarily so that it works before and after the deployment of this change and then remove the old config after this change is fully deployed. Up to you if you want to do this.

ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
```diff
@@ -211,12 +215,14 @@
   public final Counter ttlAlreadyUpdatedError;
   public final Counter ttlUpdateRejectedError;
   public final Counter replicationResponseMessageSizeTooHigh;
+  public final Map<String, Meter> crossColoFetchBytesRate = new HashMap<>();
+  public final Map<String, Meter> crossColoMetadataExchangeBytesRate = new HashMap<>();
```
jsjtzyy, line 219: @Arun-LinkedIn these two metrics are what you need to track realtime cross-colo fetch/metadata request bytes rate. Note that, these metrics are emitted on source node, which are more accurate than existing metrics on puller to indicate cross-colo bandwidth usage.

Issue #1489, opened 2020-04-28
com.github.ambry.messageformat.MessageSievingInputStreamTest > testInValidDeletedAndExpiredBlobsAgainstCorruption[2] FAILED


justinlin-linkedin:
```832com.github.ambry.messageformat.MessageSievingInputStreamTest > testInValidDeletedAndExpiredBlobsAgainstCorruption[2] FAILED
833    java.io.IOException: Encountered exception during transformation
834        at com.github.ambry.messageformat.MessageSievingInputStream.validateAndTransform(MessageSievingInputStream.java:221)
835        at com.github.ambry.messageformat.MessageSievingInputStream.<init>(MessageSievingInputStream.java:118)
836        at com.github.ambry.messageformat.MessageSievingInputStreamTest.testInValidDeletedAndExpiredBlobs(MessageSievingInputStreamTest.java:594)
837        at com.github.ambry.messageformat.MessageSievingInputStreamTest.testInValidDeletedAndExpiredBlobsAgainstCorruption(MessageSievingInputStreamTest.java:403)
838
839        Caused by:
840        java.lang.IllegalArgumentException: readShortString : the size cannot be negative
841            at com.github.ambry.utils.Utils.readShortString(Utils.java:110)
842            at com.github.ambry.store.MockId.<init>(MockId.java:41)
843            at com.github.ambry.store.MockIdFactory.getStoreKey(MockIdFactory.java:24)
844            at com.github.ambry.messageformat.ValidatingKeyConvertingTransformer.transform(MessageSievingInputStreamTest.java:1076)
845            at com.github.ambry.messageformat.MessageSievingInputStream.validateAndTransform(MessageSievingInputStream.java:206)
846            ... 3 more```

PR #1491, opened 2020-04-28
Fix a MessageSievingInputStream test failure

justinlin-linkedin:
This PR is to fix this issue https://github.com/linkedin/ambry/issues/1489.

The root cause for this failure is we create a corrupted data stream for message info 2, and sometimes the header version is valid since it's all a random number. When the header version is valid, it will might fail because of StoreKey is not valid. Since when the header version is valid, the entire header bytes are valid, and store key is the item following the header, and it will fail.

In this PR, we start verifying the header's sanity after reading the header out. This will make header invalid even if the header version is valid.

### DIFF
--- a/ambry-messageformat/src/test/java/com/github/ambry/messageformat/MessageSievingInputStreamTest.java
+++ b/ambry-messageformat/src/test/java/com/github/ambry/messageformat/MessageSievingInputStreamTest.java
@@ -474,11 +474,13 @@ public class MessageSievingInputStreamTest {
     MessageInfo msgInfo2 =
         new MessageInfo(key2, messageFormatStream2.getSize(), accountId2, containerId2, prop2.getCreationTimeInMs());
 
-    // corrupt the message stream
-    byte[] corruptMessageStream = new byte[(int) messageFormatStream2.getSize()];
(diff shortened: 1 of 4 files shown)


ambry-messageformat/src/test/java/com/github/ambry/messageformat/MessageSievingInputStreamTest.java
```diff
@@ -581,6 +618,10 @@ private void testInValidDeletedAndExpiredBlobs(short blobVersion, BlobType blobT
         (int) messageFormatStream1.getSize() + (int) messageFormatStream2.getSize()
             + (int) messageFormatStream3.getSize() + (int) messageFormatStream4.getSize(),
         (int) messageFormatStream5.getSize());
+    corruptStream6.read(totalMessageStreamContent,
+        (int) messageFormatStream1.getSize() + (int) messageFormatStream2.getSize()
+            + (int) messageFormatStream3.getSize() + (int) messageFormatStream4.getSize()
+            + (int) messageFormatStream5.getSize(), (int) messageFormatStream6.getSize());
```
jsjtzyy, line 624: minor: we have lots of redundant calculation of the offset in `read` method, suggest simplifying them.

PR #1493, opened 2020-04-29
Prevent inactive child clients from slowing down sendAndPoll

### LINKED ISSUE #1492: Fix CompositeNetworkClient polling logic when one child client is inactive
Currently, [`CompositeNetworkClient.sendAndPoll`](https://github.com/linkedin/ambry/blob/master/ambry-network/src/main/java/com/github/ambry/network/CompositeNetworkClient.java#L82) iterates over each child `NetworkClient` and calls `sendAndPoll` on each of them. This is alright if both child clients are active and with many requests/responses to handle. However, if one of the clients is less active, it may wait for up to 1 second before timing out. Since this iteration is done from a single loop, waiting for 1 second prevents responses produced by the other child client from being handled.

To fix this issue, one idea is to have each `CompositeNetworkClient` create a child thread for each type of `NetworkClient` where polling can be performed without blocking the other client. This requires an extra queue layer for coordination.

cgtz:
Previously, CompositeNetworkClient.sendAndPoll iterates over each child
NetworkClient and calls sendAndPoll on each of them. This is alright if
both child clients are active and with many requests/responses to handle.
However, if one of the clients is less active, it may wait for up to the
configured poll timeout. Since this iteration is done from a single loop,
waiting for the inactive client prevents responses produced by the other
child client from being handled.

To fix this, this change introduces background threads so that each
sendAndPoll can run in parallel (similar to parallelism in Selector).
Whichever child client finishes its sendAndPoll first will call wakeup
on the other child clients to indicate that they should return from
sendAndPoll immediately.

Fixes #1492

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/CompositeNetworkClient.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/CompositeNetworkClient.java
@@ -17,6 +17,7 @@ package com.github.ambry.network;
 
 import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.clustermap.ReplicaType;
+import com.github.ambry.utils.Utils;
 import java.util.ArrayList;
 import java.util.EnumMap;
 import java.util.HashMap;
@@ -24,6 +25,12 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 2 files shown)


ambry-network/src/main/java/com/github/ambry/network/CompositeNetworkClient.java
```diff
@@ -71,23 +81,46 @@
...
+        List<ResponseInfo> responseInfoList = future.get();
+        for (ResponseInfo responseInfo : responseInfoList) {
+          // clean up correlation ids for completed requests
+          if (responseInfo.getRequestInfo() != null) {
+            correlationIdToReplicaType.remove(responseInfo.getRequestInfo().getRequest().getCorrelationId());
+          }
+          responses.add(responseInfo);
+        }
+      } catch (InterruptedException | ExecutionException e) {
+        logger.error("Hit unexpected exception on parallel sendAndPoll.", e);
```
lightningrob, line 121: If we eat this exception, what will happen to the response consumers, will they just time out?

cgtz: Yep, they will just time out in the router after the configured routerRequestTimeout and a request will be sent to another replica. 

This isn't the greatest, but i noticed that if we rethrow the exception (or encounter any other RuntimeException in sendAndPoll), the router will be closed: https://github.com/linkedin/ambry/blob/master/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java#L1012
I think swallowing the error is better than doing this. However, one could also argue that such errors are truly unexpected and require the frontend to be restarted.

PR #1497, opened 2020-04-30
Cloud undelete Implementation for VCR

ankagrawal:
Note that this is Phase 1 of undelete support in cloud, for VCR and DR cases only. 
Support for frontend related cases coming soon with other cloud blob store changes for supporting serving data in azure.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -67,7 +67,7 @@ class CloudBlobStore implements Store {
   private static final Logger logger = LoggerFactory.getLogger(CloudBlobStore.class);
   private static final int cacheInitialCapacity = 1000;
   private static final float cacheLoadFactor = 0.75f;
-  private static final int STATUS_NOT_FOUND = 404;
+  static final int STATUS_NOT_FOUND = 404;
   private final PartitionId partitionId;
   private final CloudDestination cloudDestination;
   private final ClusterMap clusterMap;
@@ -402,10 +402,16 @@ class CloudBlobStore implements Store {
     checkStarted();
     try {
       BlobId blobId = (BlobId) info.getStoreKey();
-      return requestAgent.doWithRetries(() -> cloudDestination.updateBlobLifeVersion(blobId, info.getLifeVersion()),
-          "Undelete", partitionId.toPathString());
+      // We do not have enough information in cache to check life version.
+      // It is expected that un-deletes will be very rare, so its ok to trade off occasional hits to azure,
+      // in lieu of maintaining live version along with blob id in cache.
+      short newLifeVersion =
+          requestAgent.doWithRetries(() -> cloudDestination.undeleteBlob(blobId, info.getLifeVersion()), "Undelete",
+              partitionId.toPathString());
+      // But we do need to update the cache.
+      addToCache(blobId.getID(), BlobState.CREATED);
+      return newLifeVersion;
     } catch (CloudStorageException ex) {
-      // TODO check the appropriate error code for InvalidABSOperation exception
       StoreErrorCodes errorCode =
           (ex.getStatusCode() == STATUS_NOT_FOUND) ? StoreErrorCodes.ID_Not_Found : StoreErrorCodes.IOError;
       throw new StoreException(ex, errorCode);
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -301,7 +301,7 @@ class CloudBlobStore implements Store {
         CloudBlobMetadata blobMetadata =
             new CloudBlobMetadata(blobId, messageInfo.getOperationTimeMs(), messageInfo.getExpirationTimeInMs(),
                 messageInfo.getSize(), EncryptionOrigin.VCR, cryptoAgent.getEncryptionContext(),
-                cryptoAgentFactory.getClass().getName(), encryptedSize);
+                cryptoAgentFactory.getClass().getName(), encryptedSize, messageInfo.getLifeVersion());
         // If buffer was encrypted, we no longer know its size
         long bufferLen = (encryptedSize == -1) ? size : encryptedSize;
         InputStream uploadInputStream = new ByteBufferInputStream(messageBuf);
@@ -494,10 +494,10 @@ class CloudBlobStore implements Store {
   /**
    * Check the blob state in the recent blob cache against one or more desired states and life version.
    * @param blobKey the blob key to lookup.
-   * @param lifeVersion the desired state(s) to check.  If empty, any cached state is accepted.
-   * @param desiredStates true if the blob key is in the cache in one of the desired states and has appropriate life
-   *                      version, otherwise false.
-   * @return true if the blob key is in the cache in one of the desired states, otherwise false.
+   * @param lifeVersion the life version to check.
+   * @param desiredStates the desired state(s) to check. If empty, any cached state is accepted.
+   * @return true if the blob key is in the cache in one of the desired states and has appropriate life
+   * version, otherwise false.
    */
   private boolean checkCacheState(String blobKey, short lifeVersion, BlobState... desiredStates) {
     if (isVcr) {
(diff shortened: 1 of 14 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -398,7 +398,24 @@ private boolean deleteIfNeeded(BlobId blobId, long deletionTime) throws CloudSto
...
-    throw new UnsupportedOperationException("Undelete not supported in cloud store");
+    // TODO: Currently this is implemented for undeletes via replication only for DR.
+    checkStarted();
+    try {
+      BlobId blobId = (BlobId) info.getStoreKey();
+      // We do not have enough information in cache to check life version.
+      // It is expected that un-deletes will be very rare, so its ok to trade off occasional hits to azure,
+      // in lieu of maintaining live version along with blob id in cache.
+      short newLifeVersion =
+          requestAgent.doWithRetries(() -> cloudDestination.undeleteBlob(blobId, info.getLifeVersion()), "Undelete",
```
justinlin-linkedin, line 409: I want to point out that info.getLifeVersion might return -1 (MessageInfo.LIFE_VERSION_FROM_FRONTEND). If so, it means we have to fetch the current lifeVersion and increment it and then put it back to the azure and cosmo. Can you add a todo here so later we can deal with it.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -398,7 +398,24 @@ private boolean deleteIfNeeded(BlobId blobId, long deletionTime) throws CloudSto
 
   @Override
   public short undelete(MessageInfo info) throws StoreException {
-    throw new UnsupportedOperationException("Undelete not supported in cloud store");
+    // TODO: Currently this is implemented for undeletes via replication only for DR.
```
justinlin-linkedin, line 401: BTW, delete method would also have to be take lifeVersion into consideration. For instance, if the current lifeVersion is 0 and replication thread calls delete with lifeVersion being 2, then we have to update lifeVersion as well, even if we don't change the deletion_time.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -292,15 +293,15 @@ public CloudBlobMetadata getBlobMetadata(BlobId blobId) throws BlobStorageExcept
...
    * @return a {@link UpdateResponse} with the updated metadata.
    * @throws BlobStorageException if the blob does not exist or an error occurred.
    * @throws IllegalStateException on request timeout.
    */
-  public UpdateResponse updateBlobMetadata(BlobId blobId, String fieldName, Object value) throws BlobStorageException {
+  public UpdateResponse updateBlobMetadata(BlobId blobId, Map<String, Object> updateFields)
+      throws BlobStorageException {
     Objects.requireNonNull(blobId, "BlobId cannot be null");
-    Objects.requireNonNull(fieldName, "Field name cannot be null");
+    updateFields.keySet().stream().forEach(field -> Objects.requireNonNull(updateFields.get(field)));
```
justinlin-linkedin, line 304: nit: can you add the field name as the second parameter to requireNoNull method.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
ankagrawal: done

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
ankagrawal: done

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -381,24 +382,52 @@ public void delete(List<MessageInfo> infos) throws StoreException {
...
+   * @param lifeVersion life version of the deleted blob.
+   * @return final updated life version of the blob.
+   * @throws CloudStorageException in case any exception happens during undelete.
+   */
+  private short undeleteIfNeeded(BlobId blobId, short lifeVersion) throws CloudStorageException {
+    // TODO: Currently this is implemented for undeletes via replication only for DR.
+    // See note in deleteIfNeeded.
+    if (!checkCacheState(blobId.getID(), lifeVersion, BlobState.CREATED)) {
+      short newLifeVersion = cloudDestination.undeleteBlob(blobId, lifeVersion);
+      addToCache(blobId.getID(), newLifeVersion, BlobState.CREATED);
```
lightningrob, line 426: What if the blob went from state TTL_UPDATED to DELETED before getting undeleted.  Does it matter that we are switching it back to CREATED?  It seems okay, just checking.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -442,21 +471,37 @@ private boolean updateTtlIfNeeded(BlobId blobId) throws CloudStorageException {
...
    */
   private boolean checkCacheState(String blobKey, BlobState... desiredStates) {
+    return checkCacheState(blobKey, IGNORE_LIFE_VERSION, desiredStates);
+  }
+
+  /**
+   * Check the blob state in the recent blob cache against one or more desired states and life version.
+   * @param blobKey the blob key to lookup.
+   * @param lifeVersion the desired state(s) to check.  If empty, any cached state is accepted.
+   * @param desiredStates true if the blob key is in the cache in one of the desired states and has appropriate life
```
lightningrob, line 498: The javadoc for these 2 lines looks messed up.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -466,7 +511,9 @@ private boolean checkCacheState(String blobKey, BlobState... desiredStates) {
         return true;
```
lightningrob, line 511: What lifeVersion doesn't match?

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -480,12 +527,21 @@ private boolean checkCacheState(String blobKey, BlobState... desiredStates) {
...
+  // Visible for test.
+  void addToCache(String blobKey, short lifeVersion, BlobState blobState) {
     if (isVcr) {
-      recentBlobCache.put(blobKey, blobState);
+      if (blobState == BlobState.TTL_UPDATED) {
+        // In case of ttl update we update the ttl without taking into account the life version.
+        // So make sure that we do not decrease the lifeVersion in cache due to an incoming ttl update.
+        if (recentBlobCache.containsKey(blobKey)) {
+          lifeVersion = (lifeVersion > recentBlobCache.get(blobKey).getLifeVersion()) ? lifeVersion
+              : recentBlobCache.get(blobKey).getLifeVersion();
```
lightningrob, line 541: Can this just be Math.max(lifeVersion, recentBlobCache.get(blobKey).getLifeVersion()) ?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -151,13 +152,30 @@ public void downloadBlob(BlobId blobId, OutputStream outputStream) throws CloudS
...
+        Collections.singletonMap(CloudBlobMetadata.FIELD_EXPIRATION_TIME, expirationTime));
+  }
+
+  @Override
+  public short undeleteBlob(BlobId blobId, short lifeVersion) throws CloudStorageException {
+    Map<String, Object> updateFields = new HashMap<>();
+    updateFields.put(CloudBlobMetadata.FIELD_LIFE_VERSION, lifeVersion);
+    updateFields.put(CloudBlobMetadata.FIELD_DELETION_TIME, Utils.Infinite_Time);
+    updateBlobMetadata(blobId, updateFields);
+    // We either update lifeVersion or throw error. So this should work for now.
```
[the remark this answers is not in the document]
ankagrawal: in `updateBlobMetadata`, there are three cases possible
1. there is an exception during the update.
2. update is successful and we return true.
3. update returns false, because metadata is already updated to the asked values.

So, in effect, either the lifeversion gets updated, or an error is thrown.
Anyways, with the upcoming cloudblob store changes to support frontend request, I will handle this case to return the correct lifeVersion. So this is a temporary fix to enable replication, and unblock undelete testing, while I am working on cloud blob store related changes.

PR #1498, opened 2020-04-30
Support generating partition override map from config file in bootstrap tool

jsjtzyy:
Previously, we have to manually edit the state of partition in static clustermap. If there are thousands of partitions to be marked as ReadOnly, it will take a lot of effort to change clustermap(either by script or manually). This PR allows user to feed tool with a static config file that contains a list of partition ids to override. These partitions are separated by comma and bootstrap tool will check if id is valid (i.e. numeric, within valid range).

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
@@ -55,6 +55,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import static com.github.ambry.clustermap.ClusterMapUtils.*;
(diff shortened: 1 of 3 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -494,6 +500,51 @@ private HelixBootstrapUpgradeUtil(String hardwareLayoutPath, String partitionLay
     return partitionOverrideByDc;
   }
 
+  /**
+   * Generate partition override map a static config file. It requires file to be a list of partition ids (separated by
+   * comma). These partitions will overridden to READ_ONLY state.
```
xuhao417347761, line 505: Minor: will be overriden to to READ_ONLY state.

PR #1500, opened 2020-05-04
Account for replica state when randomly selecting writable partition

jsjtzyy:
Previously writable partition is selected based on 1. partition state(read-write)
2.replicas of partition are up. This logic is correct in most cases but may have
two minor issues: (1) replicas are up but not in a required state for PUT. (2)
replica is actually down but failure detector temporarily marks it up after backoff.
This PR introduces 3rd criterion that ensures alive replicas are in eligible states
for PUT operation. Together with previous two criteria, the partition selection logic
is able to preclude some unsatisfied partitions.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
@@ -20,6 +20,7 @@ import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedList;
@@ -487,7 +488,7 @@ public class ClusterMapUtils {
               selected)) {
             if (selected.getPartitionState() == PartitionState.READ_WRITE) {
               anyWritablePartition = selected;
-              if (areEnoughReplicasForPartitionUp(selected)) {
+              if (hasEnoughEligibleReplicasForPut(selected)) {
                 return selected;
               }
             }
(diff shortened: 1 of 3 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
```diff
@@ -505,17 +506,19 @@ PartitionId getRandomWritablePartition(String partitionClass, List<PartitionId>
...
+     * Check whether the partition has enough eligible replicas for write operations to try. Here, "eligible" means
+     * replica is up and in required states for PUT request. Enough replicas is considered to be all local replicas if
+     * such information is available. In case localDatacenterName is not available, all of the partition's replicas
+     * should be up.
      * @param partitionId the {@link PartitionId} to check.
-     * @return true if enough replicas are up; false otherwise.
+     * @return true if enough replicas are eligible; false otherwise.
      */
-    private boolean areEnoughReplicasForPartitionUp(PartitionId partitionId) {
+    private boolean hasEnoughEligibleReplicasForPut(PartitionId partitionId) {
```
ankagrawal, line 516: This is a nitpick. Feel free to ignore, if you don't feel like it.

I would suggest that either we name both methods (`getRandomWritablePartition`, `hasEnoughEligibleReplicasForPut`) to one of these names
1.  (`getRandomWritablePartition`, `hasEnoughEligibleWritableReplicas`)
2.  (`getRandomPartitionForPut`, `hasEnoughEligibleReplicasForPut`)

But if we decide to take approach 2, then we should also rename these:
`ClusterMap::getRandomWritablePartition` to `ClusterMap::getRandomPartitionForPut`
`PartitionLayout::getRandomWritablePartition` to `PartitionLayout::getRandomPartitionForPut`

cgtz: Option 1 sounds like a good name to me without requiring many public interface changes

jsjtzyy: sure, let's take option 1

PR #1504, opened 2020-05-05
Add config to true on added channel handlers

justinlin-linkedin:
Use a config to true on and off the added channel handlers.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RestServerConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RestServerConfig.java
@@ -17,6 +17,8 @@ package com.github.ambry.config;
  * Configuration parameters required by RestServer and Rest infrastructure
  */
 public class RestServerConfig {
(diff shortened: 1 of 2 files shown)


ambry-api/src/main/java/com/github/ambry/config/RestServerConfig.java
```diff
@@ -17,6 +17,8 @@
  * Configuration parameters required by RestServer and Rest infrastructure
  */
 public class RestServerConfig {
+  public final static String CONFIG_PREFIX = "rest.server.";
```
cgtz, line 20: nit: final static -> static final

justinlin-linkedin:
> Changes LGTM
> However, one alternative way to "configure" would be to have the caller of the RestServer constructor pass an empty list of channel handlers. I don't have a strong preference for either way.

The caller of the RestServer constructor is in closed source. I don't know how to add configuration to closed source code.

PR #1505, opened 2020-05-06
Improving POST request failure handling

ssen-li:
Introducing a delay in closing the netty channel when a POST request fails due to any reason.
Adding a config to enable this behaviour. By default this is disabled.
Also adding a metric to keep a count of such delayed close event.
Fix for some existing UT and and new UT to test the new code flow.


ambry-rest/src/main/java/com/github/ambry/rest/NettyResponseChannel.java
```diff
@@ -110,16 +114,37 @@
...
   private ResponseStatus errorResponseStatus = null;
 
+  /**
+   * A {@link ChannelFutureListener} that closes the {@link Channel} which is
+   * associated with the specified {@link ChannelFuture} after certain delay which is configured through {@link NettyConfig}.
+   */
+  public final ChannelFutureListener DELAYED_CLOSE = new ChannelFutureListener() {
+    @Override
+    public void operationComplete(ChannelFuture future) {
+      logger.trace("scheduling closeure of channel {}", future.channel());
```
lightningrob, line 124: closure

ambry-rest/src/main/java/com/github/ambry/rest/NettyResponseChannel.java
```diff
@@ -681,10 +706,16 @@ private HttpResponseStatus getHttpResponseStatus(ResponseStatus responseStatus)
    * </p>
    * May also close the channel if the class internally is forcing a close (i.e. if {@link #close()} is called.
    * @param closeNetworkChannel network channel is closed if {@code true}.
+   * @param shouldDelay network channel may be closed after certain delay if {@code true}.
    */
-  private void completeRequest(boolean closeNetworkChannel) {
+  private void completeRequest(boolean closeNetworkChannel, boolean shouldDelay) {
     if ((closeNetworkChannel || forceClose) && ctx.channel().isOpen()) {
-      writeFuture.addListener(ChannelFutureListener.CLOSE);
+      if (!shouldDelay || (request != null && !request.getRestMethod().equals(RestMethod.POST)) || (this.nettyConfig.nettyServerCloseDelayTimeoutMs == 0)) {
```
lightningrob, line 713: Need to format this class for line wrapping.  In LI Intellij, do Code/Reformat Code or key shortcut Cmd-Alt-L

lightningrob: also it's a bit hard to read the line, I suggest reversing this line and changing the OR's to AND's and removing NOT's (meaning we do want to delay the close).

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
```diff
@@ -658,7 +659,8 @@ public void channelInactiveTest() {
     EmbeddedChannel channel = new EmbeddedChannel(chunkedWriteHandler);
     NettyResponseChannel nettyResponseChannel =
         new NettyResponseChannel(new MockChannelHandlerContext(channel), new NettyMetrics(new MetricRegistry()),
-            new PerformanceConfig(new VerifiableProperties(new Properties())));
+            new PerformanceConfig(new VerifiableProperties(new Properties())),
+            new NettyConfig(new VerifiableProperties(new Properties())));
```
lightningrob, line 663: Minor: please create VerifiableProperties once since it's being used multiple times.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
```diff
@@ -1591,7 +1626,7 @@ private void setNullHeaders() throws RestServiceException {
    */
   private void setRequestTest() throws RestServiceException {
     ResponseStatus status = ResponseStatus.Accepted;
-    restResponseChannel = new NettyResponseChannel(ctx, new NettyMetrics(new MetricRegistry()), PERFORMANCE_CONFIG);
+    restResponseChannel = new NettyResponseChannel(ctx, new NettyMetrics(new MetricRegistry()), PERFORMANCE_CONFIG, this.nettyConfig);
```
lightningrob, line 1629: Line formatting

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
```diff
@@ -1082,6 +1084,33 @@ private void sendRequestAndEvaluateResponsePerformance(HttpRequest httpRequest,
...
+    HttpHeaders httpHeaders = new DefaultHttpHeaders();
+    httpHeaders.set(MockNettyMessageProcessor.REST_SERVICE_ERROR_CODE_HEADER_NAME, REST_ERROR_CODE);
+    httpHeaders.set(MockNettyMessageProcessor.INCLUDE_EXCEPTION_MESSAGE_IN_RESPONSE_HEADER_NAME, "true");
+    HttpRequest httpRequest = RestTestUtils.createFullRequest(HttpMethod.POST,
+        TestingUri.OnResponseCompleteWithRestException.toString(), httpHeaders, content.getBytes());
+    channel.writeInbound(httpRequest);
+    HttpResponse response = (HttpResponse) channel.readOutbound();
+    assertEquals("Unexpected response status", getExpectedHttpResponseStatus(REST_ERROR_CODE), response.status());
+    //channel should not be closed right away.
+    assertTrue("Channel closed on the server", channel.isActive());
```
lightningrob, line 1112: Can also wait 500ms and verify that it does get closed.  Also check the counters were incremented.

ambry-rest/src/main/java/com/github/ambry/rest/NettyResponseChannel.java
ssen-li: Done.

ambry-rest/src/main/java/com/github/ambry/rest/NettyResponseChannel.java
ssen-li: Done. Thanks much for the formatting help.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
ssen-li: Done.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
ssen-li: Done.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
ssen-li: I actually tried waiting but it did not work. There is no dedicated threads running the netty channel handlers here.  Everything is getting executed in a single Junit test runner thread. The netty eventloop.scheduler() task never gets executed because netty does not have any real threads to schedule this task. Even if i add a Thread.sleep() here, after the sleep is over the the test runner thread simply exits. I could not find any other way of verifying this. Let me know if there is any way to test this. 

I added check for the metrics. This is fine.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
ssen-li: actually this line is within the 120 char limit. no change needed.

ambry-rest/src/main/java/com/github/ambry/rest/NettyResponseChannel.java
```diff
@@ -681,10 +706,17 @@ private HttpResponseStatus getHttpResponseStatus(ResponseStatus responseStatus)
    * </p>
    * May also close the channel if the class internally is forcing a close (i.e. if {@link #close()} is called.
    * @param closeNetworkChannel network channel is closed if {@code true}.
+   * @param shouldDelay network channel may be closed after certain delay if {@code true}.
    */
-  private void completeRequest(boolean closeNetworkChannel) {
+  private void completeRequest(boolean closeNetworkChannel, boolean shouldDelay) {
     if ((closeNetworkChannel || forceClose) && ctx.channel().isOpen()) {
-      writeFuture.addListener(ChannelFutureListener.CLOSE);
+      if (shouldDelay && (request != null && request.getRestMethod().equals(RestMethod.POST))
```
cgtz, line 713: I wonder if we need to only do this on post requests? It is probably less likely to experience this issue on non-POST requests since those don't usually have a body and closeNetworkChannel is rarely true, but I don't see an issue with doing the delayed close on all types of requests where closeNetworkChannel is true

ssen-li: Thanks @cgtz for the comment. Later I was thinking about doing this for all http method as well but since the ticket specifically talks about to solve it for POST method I made it like so. Please let me know if this fine or should i remove the request method check?

ssen-li: Thanks @cgtz for the comment. Later I was thinking about doing this for all http method as well but since the ticket specifically talks about to solve it for POST method I made it like so. Please let me know if this fine or should i remove the request method check?

cgtz: I guess it is fine to keep it as is, since this issue probably happens when there is more request data to send, which is only posts for now. We can revisit this if we support other operations that contain request bodies in the future.

ssen-li: Thanks for the confirmation. So will keep it as it is for now.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
lightningrob: I'm kind of surprised by this.  Looking at the code, I see EmbeddedChannel uses EmbeddedEventLoop which extends AbstractScheduledEventExecutor and I would have thought the runScheduledTasks() method would get the scheduled task and run it.  Did you try stepping through with the debugger to see where the task ends up?  I see that Yingyi @jsjtzyy has worked on this test class, he might have suggestions on how to trigger it.

ssen-li: Thanks so much @lightningrob for the pointers. Since EmbeddedChannel does not have its own thread pool the UT thread has to call the runPendingTask() api to fire the scheduled tasks on this channel. This is very different from the NioSocketChannel. I was not aware that the api has to be explicitly triggered. I have enhanced the UT to do all the validations now. Please take a look. Thanks.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
```diff
@@ -1082,6 +1084,43 @@ private void sendRequestAndEvaluateResponsePerformance(HttpRequest httpRequest,
...
+        RestTestUtils.createFullRequest(HttpMethod.POST, TestingUri.OnResponseCompleteWithRestException.toString(),
+            httpHeaders, content.getBytes());
+    channel.writeInbound(httpRequest);
+    HttpResponse response = (HttpResponse) channel.readOutbound();
+    assertEquals("Unexpected response status", getExpectedHttpResponseStatus(REST_ERROR_CODE), response.status());
+    //channel should not be closed right away.
+    assertTrue("Channel closed on the server", channel.isActive());
+    //wait for delayed time * 2 times (to rule out timing out on border) and then check again.
+    Thread.sleep(delayMs * 2);
+    channel.runPendingTasks();
```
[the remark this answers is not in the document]
cgtz: Since the task is delayed, I would feel that it should be called after, otherwise the task wouldn't be ready to run.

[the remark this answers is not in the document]
jsjtzyy: make sense, feel free to ignore this comment.

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
```diff
@@ -1234,6 +1273,7 @@ public static TestingUri getTestingURI(String uri) {
   static final byte[] CHUNK = TestUtils.getRandomBytes(1024);
   static final String CHUNK_COUNT_HEADER_NAME = "chunkCount";
   static PerformanceConfig PERFORMANCE_CONFIG = new PerformanceConfig(new VerifiableProperties(new Properties()));
+  static NettyConfig NETTY_CONFIG = new NettyConfig(new VerifiableProperties(new Properties()));
```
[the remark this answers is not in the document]
ssen-li: Done

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
```diff
@@ -1082,6 +1084,43 @@ private void sendRequestAndEvaluateResponsePerformance(HttpRequest httpRequest,
       assertFalse(httpRequest.method() + " request should be unsatisfied", MockNettyRequest.mockTracker.isSatisfied());
     }
   }
+
+  /**
+   * Tests the invocation of DELAYED_CLOSE when post failures happen in {@link NettyResponseChannel}.
+   */
+  @Test
+  public void completeRequestWithDelayedCloseTest() throws Exception {
```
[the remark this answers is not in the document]
ssen-li: Done

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
```diff
@@ -1082,6 +1084,43 @@ private void sendRequestAndEvaluateResponsePerformance(HttpRequest httpRequest,
...
+    RestServiceErrorCode REST_ERROR_CODE = RestServiceErrorCode.BadRequest;
+    String content = "@@randomContent@@@";
+    HttpHeaders httpHeaders = new DefaultHttpHeaders();
+    httpHeaders.set(MockNettyMessageProcessor.REST_SERVICE_ERROR_CODE_HEADER_NAME, REST_ERROR_CODE);
+    httpHeaders.set(MockNettyMessageProcessor.INCLUDE_EXCEPTION_MESSAGE_IN_RESPONSE_HEADER_NAME, "true");
+    HttpRequest httpRequest =
+        RestTestUtils.createFullRequest(HttpMethod.POST, TestingUri.OnResponseCompleteWithRestException.toString(),
+            httpHeaders, content.getBytes());
+    channel.writeInbound(httpRequest);
+    HttpResponse response = (HttpResponse) channel.readOutbound();
```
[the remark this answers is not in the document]
ssen-li: Done

ambry-rest/src/test/java/com/github/ambry/rest/NettyResponseChannelTest.java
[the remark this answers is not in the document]
ssen-li: No problem, Yeah we have to trigger the runPendingTasks() after sleeping for the delay time. Otherwise the task will not be fired.

ssen-li:
Addressed Yingyi's comment in latest commit.

PR #1506, opened 2020-05-06
Break compaction query into smaller time buckets

lightningrob:
Improve logging in a few places


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -212,6 +214,28 @@ public int compactPartition(String partitionPath, String fieldName, long querySt
 
     // Iterate until returned list size < limit, time runs out or we get shut down
     int totalPurged = 0;
+    long chunkTimeRange = TimeUnit.DAYS.toMillis(queryBucketDays);
+    if (queryEndTime - queryStartTime > chunkTimeRange) {
+      logger.debug("Dividing compaction query for {} into buckets of {} days", partitionPath, queryBucketDays);
+      long chunkedStartTime = queryStartTime;
+      while (chunkedStartTime < queryEndTime) {
+        long chunkedEndTime = Math.min(chunkedStartTime + chunkTimeRange, queryEndTime);
+        int numPurged = compactPartition(partitionPath, fieldName, chunkedStartTime, chunkedEndTime, timeToQuit);
```
ankagrawal, line 223: We should move the logic that actually does the compaction (which is basically everything outside the if statement in line 218), in a separate method of its own, unless we have reason not to do so. Makes code easy to read, keeps methods small and separates out the control part of compaction logic from the actual compaction.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -124,30 +128,30 @@ public int compactPartitions() {
...
       if (!partitions.contains(partitionId)) {
         // Looks like partition was reassigned since the loop started, so skip it
         continue;
       }
-      logger.info("Running compaction on partition {}", partitionPath);
+      logger.info("Compacting partition {} over time range {} - {}", partitionPath, queryStartDate, queryEndDate);
       try {
-        // TODO: before compacting, call getOldestBlob to get queryStartTime
         int numPurged =
             compactPartition(partitionPath, CloudBlobMetadata.FIELD_DELETION_TIME, queryStartTime, queryEndTime,
```
ankagrawal, line 147: Can there be a case where we cannot complete compaction of all partitions in 24 hours?  In that case, it's possible that same that partitions that are at the end of the list of partitions, miss the compaction every time. There are atleast 2 ways we can approach this:

1. If instead of draining out each partition in a loop, we did n batches of `compactPartittion` calls for each partition, before looping through the list of partitions again, it will keep every partition compacted uniformly.
2. Another way we can solve this could be by setting cloudConfig.cloudBlobCompactionIntervalHours large enough so that this scenario can never happen. For this, it might be good to have metrics that can tell us how many times we were able to successfully loop through all partitions, and how many times we couldn't loop through all partitions.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -124,30 +128,30 @@ public int compactPartitions() {
     long compactionStartTime = now;
     long timeToQuit = now + compactionTimeLimitMs;
     long queryEndTime = now - retentionPeriodMs;
-    // TODO: we can cache the latest timestamp that we know we have cleared and use that on subsequent calls
+    // TODO: checkpoint the latest timestamp and use that on subsequent compactions
     // Starting from beginning of time is too expensive
     // Order partitions by earliest time at which dead blob exists
```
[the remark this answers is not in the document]
lightningrob: I'm removing these comments.  I will file a ticket for the checkpointing enhancement

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
lightningrob: I think the first suggestion is a good idea.  I will file a separate ticket to do that in a follow up PR.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
lightningrob: I had it that way initially but then consolidated it into single method.  I can change it back if it helps readability.

PR #1510, opened 2020-05-07
Allow router to send cross-colo request to dc with most replicas

jsjtzyy:
There is a case where a certain dc is decommissioned and blobs previously uploaded to this dc now have an unrecognizable dc id. In our current logic, clustermap will treat originating dc of this blob as null and router randomly picks a remote replica to route cross-colo request without accounting for replica distribution. This can be improved by prioritizing remote dc with 3 replicas.

This PR introduces a minor change: when request didn't succeed in local dc, operation tracker should be allowed to try on remote dc with most replicas first so as to improve success rate of cross-colo requests(i.e GET/DELETE etc). This is useful in cluster with "unbalanced" replica distribution (i.e. 3 replicas in local dc and 1 replica per remote dc).

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -70,10 +70,10 @@ import org.slf4j.LoggerFactory;
 class SimpleOperationTracker implements OperationTracker {
   protected final String datacenterName;
   protected final String originatingDcName;
-  protected final int diskSuccessTarget;
-  protected final int diskParallelism;
-  protected final int cloudSuccessTarget;
-  protected final int cloudParallelism;
+  protected final int diskReplicaSuccessTarget;
+  protected final int diskReplicaParallelism;
+  protected final int cloudReplicaSuccessTarget;
+  protected final int cloudReplicaParallelism;
   protected final boolean cloudReplicasPresent;
   protected final boolean diskReplicasPresent;
   // How many NotFound responses from originating dc will terminate the operation.
@@ -144,14 +144,14 @@ class SimpleOperationTracker implements OperationTracker {
     this.routerOperation = routerOperation;
     this.originatingDcName = originatingDcName;
     datacenterName = routerConfig.routerDatacenterName;
-    cloudSuccessTarget = routerConfig.routerCloudSuccessTarget;
-    cloudParallelism = routerConfig.routerCloudRequestParallelism;
+    cloudReplicaSuccessTarget = routerConfig.routerCloudSuccessTarget;
+    cloudReplicaParallelism = routerConfig.routerCloudRequestParallelism;
     List<ReplicaId> eligibleReplicas;
     switch (routerOperation) {
       case GetBlobOperation:
       case GetBlobInfoOperation:
-        diskSuccessTarget = routerConfig.routerGetSuccessTarget;
-        diskParallelism = routerConfig.routerGetRequestParallelism;
+        diskReplicaSuccessTarget = routerConfig.routerGetSuccessTarget;
+        diskReplicaParallelism = routerConfig.routerGetRequestParallelism;
         crossColoEnabled = routerConfig.routerGetCrossDcEnabled;
         includeNonOriginatingDcReplicas = routerConfig.routerGetIncludeNonOriginatingDcReplicas;
         numOfReplicasRequired = routerConfig.routerGetReplicasRequired;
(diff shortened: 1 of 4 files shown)


justinlin-linkedin:
> There is a case where a certain dc is decommissioned and blobs previously uploaded to this dc now have an unrecognizable dc id. In our current logic, clustermap will treat originating dc of this blob as null and router randomly picks a remote replica to route cross-colo request without accounting for replica distribution. This can be improved by prioritizing remote dc with 3 replicas.
> 
> This PR introduces a minor change: when request didn't succeed in local dc, operation tracker should be allowed to try on remote dc with most replicas first so as to improve success rate of cross-colo requests(i.e GET/DELETE etc). This is useful in cluster with "unbalanced" replica distribution (i.e. 3 replicas in local dc and 1 replica per remote dc).

I understand the logic, but not the motivation. If a blob's originating dc is decommissioned, then all the other remote dcs should all have 3 replicas or one replica. In video cluster, if a blob is uploaded to sg1, then all the other colos only have 1 replicas. In main cluster, if a blob is uploaded to sg1, than all the other colos would have 3 replicas. So there is really no need to do this if this is only for decommissioned colos.

jsjtzyy:
> > There is a case where a certain dc is decommissioned and blobs previously uploaded to this dc now have an unrecognizable dc id. In our current logic, clustermap will treat originating dc of this blob as null and router randomly picks a remote replica to route cross-colo request without accounting for replica distribution. This can be improved by prioritizing remote dc with 3 replicas.
> > This PR introduces a minor change: when request didn't succeed in local dc, operation tracker should be allowed to try on remote dc with most replicas first so as to improve success rate of cross-colo requests(i.e GET/DELETE etc). This is useful in cluster with "unbalanced" replica distribution (i.e. 3 replicas in local dc and 1 replica per remote dc).
> 
> I understand the logic, but not the motivation. If a blob's originating dc is decommissioned, then all the other remote dcs should all have 3 replicas or one replica. In video cluster, if a blob is uploaded to sg1, then all the other colos only have 1 replicas. In main cluster, if a blob is uploaded to sg1, than all the other colos would have 3 replicas. So there is really no need to do this if this is only for decommissioned colos.

Your point is correct if we do nothing after retiring certain colo, but we actually have replenished the replicas to lor1 colo in video cluster for lsg1. Now, for those partitions that originally have 3 replicas in lsg1 now have 3 replicas in lor1 and this PR will help router prioritize lor1 if blob id originated in lsg1.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -159,48 +161,66 @@
...
-              + ", routerOperation=" + routerOperation);
+          "Parallelism has to be > 0. diskParallelism=" + diskReplicaParallelism + ", cloudParallelism="
+              + cloudReplicaParallelism + ", routerOperation=" + routerOperation);
     }
 
     // Order the replicas so that local healthy replicas are ordered and returned first,
     // then the remote healthy ones, and finally the possibly down ones.
     List<? extends ReplicaId> replicas =
         routerConfig.routerGetEligibleReplicasByStateEnabled ? eligibleReplicas : partitionId.getReplicaIds();
+    // In a case where a certain dc is decommissioned and blobs previously uploaded to this dc now have a unrecognizable
```
lightningrob, line 207: Minor: please add blank lines before and after this new section for readability.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -159,48 +161,66 @@
...
+    // replicas first. This is useful in cluster with "unbalanced" replica distribution (i.e. 3 replicas in local dc and
+    // 1 replica per remote dc)
+    if (originatingDcName == null && routerConfig.routerCrossColoRequestToDcWithMostReplicas) {
+      Map<String, Long> dcToReplicaCnt = replicas.stream()
+          .collect(Collectors.groupingBy(e -> e.getDataNodeId().getDatacenterName(), Collectors.counting()));
+      List<Map.Entry<String, Long>> entryList = new ArrayList<>(dcToReplicaCnt.entrySet());
+      entryList.sort(Map.Entry.comparingByValue());
+      // we assign a dc with most replicas to "originatingDcName", which only takes effect when populating replica pool
+      // (replicas in that colo have higher priority than other remote colos). Note that, "this.originatingDcName" still
+      // keeps the actual originating dc name (which is null). This value forces operation track to go through replicas
```
lightningrob, line 219: Minor: operation tracker

lightningrob: I recommend using a different variable name than originatingDcName.  Making it different from the class variable is too confusing.  Also, I don't see where local originatingDcName is being used other than in the debug line.

PR #1513, opened 2020-05-07
Return lifeVersion from get request

justinlin-linkedin:
Returning life version in the get request. LifeVersion is a piece of information that would further verify the success of an undelete operation. Since for now undelete doesn't return the lifeVersion, we should expose a way to get the lifeVersion before and after undelete. There is no better way than a get.

I plan in the closed source code, to not return lifeVersion when the client is not internal client. And maybe we should return the revised lifeVersion to the client in an undelete response.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/messageformat/BlobInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/messageformat/BlobInfo.java
@@ -22,7 +22,7 @@ public class BlobInfo {
   private short lifeVersion;
 
   /**
-   * Constructor to creat a {@link BlobInfo}.
+   * Constructor to create a {@link BlobInfo}.
    * @param blobProperties The {@link BlobProperties} of this blob.
    * @param userMetadata The user metadata of this blob.
    */
@@ -31,7 +31,7 @@ public class BlobInfo {
   }
 
   /**
-   * Constructor to creat a {@link BlobInfo}.
+   * Constructor to create a {@link BlobInfo}.
    * @param blobProperties The {@link BlobProperties} of this blob.
    * @param userMetadata The user metadata of this blob.
    * @param lifeVersion The lifeVersion of this blob.
--- a/ambry-api/src/main/java/com/github/ambry/messageformat/BlobInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/messageformat/BlobInfo.java
@@ -22,9 +22,12 @@ public class BlobInfo {
 
   private byte[] userMetadata;
 
+  private short lifeVersion;
+
   public BlobInfo(BlobProperties blobProperties, byte[] userMetadata) {
     this.blobProperties = blobProperties;
     this.userMetadata = userMetadata;
+    this.lifeVersion = 0;
   }
 
   public BlobProperties getBlobProperties() {
@@ -33,5 +36,20 @@ public class BlobInfo {
 
   public byte[] getUserMetadata() {
     return userMetadata;
+  }
+
+  /**
+   * Set the lifeVersion of this blob.
+   * @param lifeVersion The lifeVersion to set.
+   */
+  public void setLifeVersion(short lifeVersion) {
+    this.lifeVersion = lifeVersion;
+  }
+
+  /**
+   * @return The lifeVersion of this blob.
+   */
+  public short getLifeVersion() {
+    return lifeVersion;
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/messageformat/BlobInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/messageformat/BlobInfo.java
@@ -14,26 +14,44 @@
 package com.github.ambry.messageformat;
 
 /**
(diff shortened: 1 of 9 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
```diff
@@ -353,9 +353,9 @@ private void handleBody(InputStream payload, MessageMetadata messageMetadata, Me
     ByteBuffer userMetadata = MessageFormatRecord.deserializeUserMetadata(payload);
     if (encryptionKey == null) {
       // if blob is not encrypted, move the state to Complete
-      operationResult =
-          new GetBlobResultInternal(new GetBlobResult(new BlobInfo(serverBlobProperties, userMetadata.array()), null),
-              null);
+      BlobInfo blobInfo = new BlobInfo(serverBlobProperties, userMetadata.array());
```
cgtz, line 356: It seems like in most places, `setLifeVersion` is called right after `new BlobInfo`. Instead of adding a new setter, one alternative could be making the BlobInfo.lifeVersion final and adding a new constructor that takes life version as an argument.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1100,6 +1100,7 @@ private RouterException buildChunkException(String message, RouterErrorCode erro
     private BlobType blobType;
     private List<CompositeBlobInfo.ChunkMetadata> chunkMetadataList;
     private BlobProperties serverBlobProperties;
+    private MessageInfo serverMessageInfo;
```
cgtz, line 1103: just save life version instead of the entire messageInfo?

ambry-api/src/main/java/com/github/ambry/messageformat/BlobInfo.java
```diff
@@ -14,26 +14,44 @@
...
 public class BlobInfo {
-
   private BlobProperties blobProperties;
-
   private byte[] userMetadata;
-
   private short lifeVersion;
 
+  /**
+   * Constructor to creat a {@link BlobInfo}.
```
cgtz, line 25: creat -> create (also line 34)

zzmao:
What about put lifeVersion in `BlobProperties`?

justinlin-linkedin:
> What about put lifeVersion in `BlobProperties`?

The problem of having lifeVersion in `BlobProperties` is that it's not consistent with what's being persisted on disk. `BlobProperties` is persisted with `PutRecord` on disk. and if we add lifeVersion in `BlobProperties`, we should be thinking about adding it to the serialized bytes on disk as well and that's just too much overhead for now.

zzmao:
@justinlin-linkedin   we have a similar case: `externalAssetTag`
```
public class BlobProperties {
  private final String serviceId;
  private final String ownerId;
  private final String contentType;
  private final boolean isPrivate;
  private final long creationTimeInMs;
  private final short accountId;
  private final short containerId;
  private final boolean isEncrypted;
  private long blobSize;
  private long timeToLiveInSeconds;
  // Non persistent blob properties.
  private final String externalAssetTag;
```

justinlin-linkedin:
> fo needs ACL.

Hmm, you have a point here. I am not super into the idea of having it on `BlobProperties` but I am not oppose it either. I will try to add it BlobProperties and see it goes with the code. It might be a better idea to just stuff it in the `BlobProperties`.

cgtz:
> > fo needs ACL.
> 
> Hmm, you have a point here. I am not super into the idea of having it on `BlobProperties` but I am not oppose it either. I will try to add it BlobProperties and see it goes with the code. It might be a better idea to just stuff it in the `BlobProperties`.

One could still add it to the BlobInfo/HEAD response even if it is not in BlobProperties. The objects in BlobProperties don't get automatically transformed into header. The code in security service reads from BlobInfo and turns it into response headers, so it would work both ways.

If I remember correctly, the reason for putting externalAssetTag in BlobProperties was because it needed to be passed to Router.putBlob and then to the NotificationSystem, and we did not have PutBlobOptions in the interface at the time which would allow us to pass such non-persistent fields to the router.

justinlin-linkedin:
> @justinlin-linkedin we have a similar case: `externalAssetTag`
> 
> ```
> public class BlobProperties {
>   private final String serviceId;
>   private final String ownerId;
>   private final String contentType;
>   private final boolean isPrivate;
>   private final long creationTimeInMs;
>   private final short accountId;
>   private final short containerId;
>   private final boolean isEncrypted;
>   private long blobSize;
>   private long timeToLiveInSeconds;
>   // Non persistent blob properties.
>   private final String externalAssetTag;
> ```
> 
> Being `BlobProperties` is easy to query because HEAD doesn't require ACL, but blobInfo needs ACL.

As I started to materialize this idea, it doesn't really make any difference. LifeVersion has to be added to `BlobProperties` after deserialization just like `BlobInfo`. And HEAD request would return the lifeVersion as well without ACL.

PR #1514, opened 2020-05-09
Make BlobUndeleted not an error in router and more tests

justinlin-linkedin:
BlobAlreadyUndeleted is not an error response in router from this PR on.
There are several ways ambry server might return BlobAlreadyUndeleted.
1. Frontend makes several attempts to undelete a blob
2. Frontend and replication concurrently undelete a blob (this happens a lot)

In both cases, we can ignore the BlobAlreadyUndeleted error and just treat it as an NoError.

Also more integration tests are added.


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -727,22 +735,32 @@ public short undelete(MessageInfo info) throws StoreException {
...
-              index.findKey(info.getStoreKey(), fileSpan, EnumSet.allOf(PersistentIndex.IndexEntryType.class));
+          IndexValue value = index.findKey(info.getStoreKey(), fileSpan,
+              EnumSet.of(PersistentIndex.IndexEntryType.DELETE, PersistentIndex.IndexEntryType.UNDELETE));
           if (value != null) {
-            throw new StoreException("Cannot undelete id " + info.getStoreKey() + " since concurrent operation occurs",
-                StoreErrorCodes.Life_Version_Conflict);
+            if (value.isUndelete() && value.getLifeVersion() == revisedLifeVersion) {
+              // Might get an concurrent undelete from both replication and frontend. This is considered as an
+              // successful operation and the exception will be captured by the catch statement below.
+              throw new StoreException(
```
cgtz, line 744: Instead of the logic in AmbryRequests to look up the life version again, could the life version be carried inside of the `StoreException`? You could create an `UndeleteStoreException` that extends `StoreException` and then `AmbryRequests` can handle it with an instanceof check:
```
if (e instanceof UndeleteStoreException) {
  response = Response(e.getErrorCode(), ((UndeleteStoreException) e).getLifeVersion());
}
```
I feel like this would keep the boundary between store logic and request handling logic more clear.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
```diff
@@ -195,6 +194,7 @@ void handleResponse(ResponseInfo responseInfo, UndeleteResponse undeleteResponse
               // This is first successful response.
               lifeVersion = undeleteResponse.getLifeVersion();
               firstResponseReplicaId = replica;
+              operationTracker.onResponse(replica, TrackedRequestFinalState.SUCCESS);
```
cgtz, line 197: How is the life version field in the error UndeleteResponse used?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: this is a good idea, updated.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
justinlin-linkedin: when the error is Blob_Already_Undelete, we treat it as a successful operation, just like in DeleteOperation, when the error is Blob_Already_Deleted, it's considered as a successful operation. The lifeVersion will be used to compare with other ambry servers' lifeVersions.

ambry-api/src/main/java/com/github/ambry/store/IdUndeletedStoreException.java
```diff
@@ -0,0 +1,39 @@
...
+    this.lifeVersion = lifeVersion;
+  }
+
+  /**
+   * @return The lifeVersion for the current undelete record.
+   */
+  public short getLifeVersion() {
+    return lifeVersion;
+  }
+}
```
cgtz, line 39: nit: end file with newline

ambry-protocol/src/main/java/com/github/ambry/protocol/UndeleteResponse.java
```diff
@@ -62,6 +62,18 @@ public UndeleteResponse(int correlationId, String clientId, short lifeVersion) {
...
 
+  /**
+   * Constructs a {@link UndeleteResponse} with a valid lifeVersion. The error code will be set to {@link ServerErrorCode#No_Error}.
+   * @param correlationId correlationId of the undelete response.
+   * @param clientId clientId of the undelete response.
+   * @param lifeVersion a valid lifeVersion to return to client.
+   * @param error error code returned in this undelete response.
+   */
+  public UndeleteResponse(int correlationId, String clientId, short lifeVersion, ServerErrorCode error) {
+    super(RequestOrResponseType.UndeleteResponse, UNDELETE_RESPONSE_VERSION_1, correlationId, clientId, error);
```
cgtz, line 73: since there is no live site traffic performing undeletes yet, this schema change should be ok, right?

justinlin-linkedin: that's right.

PR #1516, opened 2020-05-11
Support monitoring inconsistency between participants on same node

jsjtzyy:
As required by ZK cluster migration, ambry-server has to host two participants
that update sealed/stopped replicas at same time. In case any participant fails
to update corresponding ZK cluster(i.e. connection lost), ambry-server needs to
monitor and alert inconsistency of sealed/stopped replicas from two participants.
This PR introduces a consistency checker that periodically checks sealed/stopped
replica lists from two participants and emits metric if there is a mismatch.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -422,7 +422,7 @@ public class RouterConfig {
   public final boolean routerOperationTrackerHistogramDumpEnabled;
 
   /**
-   * The period of dumping resource-level histogram (if enabled).
+   * The period of dumping resource-level histogram in second(if enabled).
    */
   @Config(ROUTER_OPERATION_TRACKER_HISTOGRAM_DUMP_PERIOD)
   @Default("600")
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -422,7 +422,7 @@ public class RouterConfig {
   public final boolean routerOperationTrackerHistogramDumpEnabled;
 
   /**
-   * The period of dumping resource-level histogram in second(if enabled).
+   * The period of dumping resource-level histogram in seconds(if enabled).
    */
   @Config(ROUTER_OPERATION_TRACKER_HISTOGRAM_DUMP_PERIOD)
   @Default("600")
--- a/ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
@@ -1,0 +1,60 @@
+/**
+ * Copyright 2020 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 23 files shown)


jsjtzyy:
@justinlin-linkedin @lightningrob gentle reminder to review

ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
```diff
@@ -0,0 +1,60 @@
...
+  public void run() {
+    // when code reaches here, it means there are at least two participants.
+    ClusterParticipant clusterParticipant = participants.get(0);
+    Set<String> sealedReplicas1 = new HashSet<>(clusterParticipant.getSealedReplicas());
+    Set<String> stoppedReplicas1 = new HashSet<>(clusterParticipant.getStoppedReplicas());
+    for (int i = 1; i < participants.size(); ++i) {
+      Set<String> sealedReplicas2 = new HashSet<>(participants.get(i).getSealedReplicas());
+      if (!sealedReplicas1.equals(sealedReplicas2)) {
+        logger.warn("Mismatch in sealed replicas. Set {} is different from set {}", sealedReplicas1, sealedReplicas2);
+        metrics.sealedReplicasMismatchCount.inc();
```
lightningrob, line 50: If this task is going to run every X seconds, do we want to increment the counter each time?  Once we have a mismatch, the metric count will increase proportionally to how often the schedule is run, which doesn't seem very meaningful.  Wondering if it should be a Guage instead.
Btw, what action can be taken if an alert on this metric is seen?

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -422,7 +422,7 @@
   public final boolean routerOperationTrackerHistogramDumpEnabled;
 
   /**
-   * The period of dumping resource-level histogram (if enabled).
+   * The period of dumping resource-level histogram in second(if enabled).
```
lightningrob, line 425: Minor: in seconds.

ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
jsjtzyy: Using counter seems inappropriate but is actually ok because Ingraph converts it to `per second rate` when it is read. (Reference: https://iwww.corp.linkedin.com/wiki/cf/display/SOP/inGraphs+FAQ#inGraphsFAQ). An example is `instanceConfigChangeTriggerCount`, if we look at the dashboard, it actually records number of events per second.

jsjtzyy: If alert is fired, we can try to restart the server and attempt to correct the number. Sometimes the inconsistency comes from intermittent failure due to connection issue. 
If it doesn't work, we can use "partition override" feature to override certain partition to the correct state.

ambry-server/src/test/java/com/github/ambry/server/ParticipantsConsistencyCheckerTest.java
```diff
@@ -0,0 +1,238 @@
...
+import org.mockito.Mockito;
+
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+
+/**
+ * Test class to verify participant consistency checker.
+ */
+public class ParticipantsConsistencyCheckerTest {
```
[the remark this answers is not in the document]
jsjtzyy: Fair point. I will probably just call it `ParticipantsConsistencyTest` and rename several test method as well.

ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
lightningrob: I understand that counter is converted to rate.  My point was that the reported rate would be governed by how often the checker task kicks off, rather than how frequently an error happened or how many discrepancies there are, which might be confusing.  It's not a deal breaker, just wanted to point that out.

PR #1523, opened 2020-05-19
Adding a new metrics to track delayed close for netty channels.

ssen-li:
adding another metric for tracking to make sure all the sheduled delayed close gets expired and the callbacks are called. 
Modified the UT as well.


ambry-rest/src/main/java/com/github/ambry/rest/NettyResponseChannel.java
```diff
@@ -123,6 +123,7 @@
     public void operationComplete(ChannelFuture future) {
       logger.trace("scheduling closure of channel {}", future.channel());
       future.channel().eventLoop().schedule(() -> {
+        nettyMetrics.delayedCloseExpiredCount.inc();
```
lightningrob, line 126: Nit: nothing has really expired here.  Suggest renaming this delayedCloseActivatedCount or delayedCloseTriggerCount.

ssen-li: changed the metric name to delayedCloseActivatedCount. Thanks

PR #1527, opened 2020-05-20
Temporarily exclude helix-common module to avoid conflicting domain name

jsjtzyy:
This PR exludes helix-common module and explicitly depends on zookeeper to
avoid conflicting MonitorDomainNames (issue link: https://github.com/apache/helix/issues/1022).
Once apache helix fix is out, we can remove changes in this PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -153,7 +153,12 @@ project(':ambry-utils') {
         compile "org.json:json:$jsonVersion"
         compile "net.sf.jopt-simple:jopt-simple:$joptSimpleVersion"
         compile "io.netty:netty-buffer:$nettyVersion"
-        testCompile "org.apache.helix:helix-core:$helixVersion"
+        testCompile ("org.apache.helix:helix-core:$helixVersion") {
+            // TODO Once new release is out, we can remove this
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -153,7 +153,12 @@ project(':ambry-utils') {
         compile "org.json:json:$jsonVersion"
         compile "net.sf.jopt-simple:jopt-simple:$joptSimpleVersion"
         compile "io.netty:netty-buffer:$nettyVersion"
-        testCompile "org.apache.helix:helix-core:$helixVersion"
+        testCompile ("org.apache.helix:helix-core:$helixVersion") {
+            // TODO Once new release is out, we can remove this
+            exclude group: "org.apache.helix", module: "helix-common"
+        }
+        // TODO remove zookeeper dependency once Helix open source fix is out
+        testCompile "org.apache.zookeeper:zookeeper:3.4.13"
```
pkuwm, line 161: Nit, maybe could set the version number in the property file to make it consistent with other version names: `$zookeeperVersion`?

lightningrob: For a permanent change we would do this.  For a temporary workaround it's okay.

Issue #1526, opened 2020-05-20
Test failures: com.github.ambry.router.NonBlockingRouterTest > testUndeleteBasic[0] FAILED


justinlin-linkedin:
com.github.ambry.router.NonBlockingRouterTest > testUndeleteBasic[0] FAILED
    java.util.concurrent.ExecutionException: com.github.ambry.router.RouterException: Server returned: Blob_Deleted. Chunk ID: AAYQAn4mVd8AAQAAAAAAAAAF43qDQArSTGG0RRK3TBMiiw Error: BlobDeleted
        at com.github.ambry.router.FutureResult.resultOrThrow(FutureResult.java:113)
        at com.github.ambry.router.FutureResult.get(FutureResult.java:99)
        at com.github.ambry.router.NonBlockingRouterTest.testUndeleteBasic(NonBlockingRouterTest.java:395)
        Caused by:
        com.github.ambry.router.RouterException: Server returned: Blob_Deleted. Chunk ID: AAYQAn4mVd8AAQAAAAAAAAAF43qDQArSTGG0RRK3TBMiiw Error: BlobDeleted
    java.lang.AssertionError: HeapMemoryLeak: [allocation|deallocation] before test[40111|38504], after test[40375|38753] expected:<1607> but was:<1622>
        at org.junit.Assert.fail(Assert.java:91)
        at org.junit.Assert.failNotEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:126)
        at org.junit.Assert.assertEquals(Assert.java:470)
        at com.github.ambry.utils.NettyByteBufLeakHelper.afterTest(NettyByteBufLeakHelper.java:96)
        at com.github.ambry.router.NonBlockingRouterTest.after(NonBlockingRouterTest.java:199)

justinlin-linkedin:
it didn't get fixed.

jsjtzyy:
Report the same failure here:
```
com.github.ambry.router.NonBlockingRouterTest > testUndeleteBasic[1] FAILED
    java.util.concurrent.ExecutionException: com.github.ambry.router.RouterException: Server returned: Blob_Deleted. Chunk ID: AAYQAlnYMQgAAQAAAAAAAAAAxZj5xzc1Q4qo5y-TQILmfQ Error: BlobDeleted
        at com.github.ambry.router.FutureResult.resultOrThrow(FutureResult.java:113)
        at com.github.ambry.router.FutureResult.get(FutureResult.java:99)
[...]

PR #1530, opened 2020-05-21
Fix a test failure where delete request arrives after undelete

justinlin-linkedin:
This should fix an ambry-server integration test failure in the RouterTest, where a delete is acknowledged by the notification system, but actually one of the delete requests is handled after undelete.

This is the link to the test failure https://github.com/linkedin/ambry/issues/1526
This is what happened. 
1. Router issues a delete operation, which sends three delete requests to three hosts
2. Two hosts responds and at the same time, third host replicates delete from first two
3. All three hosts has delete so the notification system is acknowledged. 
4. Router issues a undelete operation, which sends undelets request to third host
5. After undelete, somehow the delete request from previous delete operation just arrives at the third host. This would mark blob deleted.

And that's how we get a Blob_Deleted after undelete.

The fix is pretty simple, change the delete parallelism to 2 so we don't have a dangling delete request.

### DIFF
--- a/ambry-server/src/integration-test/java/com/github/ambry/server/RouterServerTestFramework.java
+++ b/ambry-server/src/integration-test/java/com/github/ambry/server/RouterServerTestFramework.java
@@ -127,7 +127,7 @@ class RouterServerTestFramework {
     for (OperationChain opChain : opChains) {
       if (!opChain.latch.await(AWAIT_TIMEOUT, TimeUnit.SECONDS)) {
         Assert.fail("Timeout waiting for operation chain " + opChain.chainId + " to finish and it is stuck at "
-            + opChain.opIndex + "th op:" + opChain.currentOpType + ".");
+            + opChain.opIndex + "the op:" + opChain.currentOpType + " the blob id: " + opChain.blobId);
       }
       synchronized (opChain.testFutures) {
         for (TestFuture testFuture : opChain.testFutures) {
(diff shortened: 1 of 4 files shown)


ambry-server/src/integration-test/java/com/github/ambry/server/RouterServerTestFramework.java
```diff
@@ -141,6 +142,7 @@ void checkOperationChains(List<OperationChain> opChains) throws Exception {
         partitionCount.put(partitionId, count + 1);
       }
     }
+    /*
```
zzmao, line 145: not needed?

justinlin-linkedin: I don't think this part of test is necessary. This test that newly created blobs are not created mostly in one partition. It tests the randomness of partition assignment. There is always a change that all the blobs are created in one partition if we don't somehow force it out.

cgtz:
There was a test error on travis:
```
com.github.ambry.server.RouterServerSSLTest > interleavedOperationsTest[0] FAILED
    java.lang.Exception: Exception occured in operation: undeleteBlob-3 The blob id is AAYQAj3tTbcAAQAAAAAAAAAG5zHC8_CfQgiMnpBNqAHdYA
        at com.github.ambry.server.RouterServerTestFramework$TestFuture.get(RouterServerTestFramework.java:759)
        at com.github.ambry.server.RouterServerTestFramework$11.check(RouterServerTestFramework.java:483)
        at com.github.ambry.server.RouterServerTestFramework.checkOperationChains(RouterServerTestFramework.java:134)
        at com.github.ambry.server.RouterServerSSLTest.interleavedOperationsTest(RouterServerSSLTest.java:184)
        Caused by:
        java.util.concurrent.ExecutionException: com.github.ambry.router.RouterException: Operation timed out Error: OperationTimedOut
            at com.github.ambry.router.FutureResult.resultOrThrow(FutureResult.java:113)
            at com.github.ambry.router.FutureResult.get(FutureResult.java:108)
            at com.github.ambry.server.RouterServerTestFramework$TestFuture.get(RouterServerTestFramework.java:756)
            ... 3 more
            Caused by:
            com.github.ambry.router.RouterException: Operation timed out Error: OperationTimedOut
```

PR #1531, opened 2020-05-22
Integrate Cosmos BulkDelete stored procedure into compaction.

lightningrob:


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
```diff
@@ -136,16 +164,83 @@ void testConnectivity() {
...
+        blobMetadataList.stream().map(metadata -> '"' + metadata.getId() + '"').collect(Collectors.joining(","));
+    String query = String.format(BULK_DELETE_QUERY, quotedBlobIds);
+    String sprocLink = cosmosCollectionLink + BULK_DELETE_SPROC;
+    boolean more = true;
+    int deleteCount = 0;
+    double requestCharge = 0;
+    try {
+      while (more) {
+        StoredProcedureResponse response =
+            asyncDocumentClient.executeStoredProcedure(sprocLink, options, new String[]{query}).toBlocking().single();
```
cgtz, line 227: when continuation is true, we just try again with the same list of blob IDs. Is the stored procedure idempotent?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
```diff
@@ -136,16 +164,83 @@ void testConnectivity() {
...
+        requestCharge += response.getRequestCharge();
+        Document responseDoc = response.getResponseAsDocument();
+        more = responseDoc.getBoolean(PROPERTY_CONTINUATION);
+        deleteCount += responseDoc.getInt(PROPERTY_DELETED);
```
cgtz, line 231: does this count include blobs that were already deleted from previous iterations?

cgtz:
> Changes look good to me. Btw are we planning to checkin the stored procedure too? Maybe in the AmbryLI branch?

+1, at least keeping a link in a javadoc to the github project that contains the stored procedure we are using would be good.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
lightningrob: Yes.  It's idempotent because the sproc first runs the query to get records matching the IN filter, which will simply be smaller if some records have been deleted on prior executions.  Though I was not able to trigger this case during testing, likely because the batch size was small.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
lightningrob: It should not.

lightningrob:
I added the javascript source as a resource and referenced the github link in javadoc.

PR #1534, opened 2020-05-26
Support disabling partition in Helix when there is disk failure

jsjtzyy:
This PR allows server to disable partition when store is shut down due to
disk I/O error. Disabling partition will trigger state transition to mark
this partition in ERROR state. This is helpful for leader based replication
to handle runtime disk failure as Helix controller will automatically re-elect
a new leader and unblock cross-dc replication.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/store/Store.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/Store.java
@@ -174,9 +174,7 @@ public interface Store {
   /**
    * @return {@code true} if the store is disabled on error (due to disk I/O issue).
    */
-  default boolean disabledOnError() {
-    return false;
-  }
+  boolean disabledOnError();
 
   /**
    * Shuts down the store
--- a/ambry-api/src/main/java/com/github/ambry/store/Store.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/Store.java
@@ -172,6 +172,13 @@ public interface Store {
   boolean recoverFromDecommission();
 
   /**
+   * @return {@code true} if the store is disabled on error (due to disk I/O issue).
+   */
+  default boolean disabledOnError() {
+    return false;
+  }
+
+  /**
    * Shuts down the store
    */
   void shutdown() throws StoreException;
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
@@ -25,7 +25,7 @@ import static com.github.ambry.clustermap.ClusterMapUtils.*;
 /**
  * {@link ReplicaId} implementation to use within dynamic cluster managers.
  */
-abstract class AmbryReplica implements ReplicaId {
+public abstract class AmbryReplica implements ReplicaId {
   private final AmbryPartition partition;
   private final long capacityBytes;
   private volatile boolean isSealed;
@@ -40,7 +40,7 @@ abstract class AmbryReplica implements ReplicaId {
    * @param capacityBytes the capacity in bytes for this replica.
    * @param isSealed whether this replica is in sealed state.
    */
-  AmbryReplica(ClusterMapConfig clusterMapConfig, AmbryPartition partition, boolean isReplicaStopped,
+  public AmbryReplica(ClusterMapConfig clusterMapConfig, AmbryPartition partition, boolean isReplicaStopped,
       long capacityBytes, boolean isSealed) throws Exception {
     this.partition = Objects.requireNonNull(partition, "null partition");
     this.capacityBytes = capacityBytes;
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
@@ -40,7 +40,7 @@ public abstract class AmbryReplica implements ReplicaId {
    * @param capacityBytes the capacity in bytes for this replica.
    * @param isSealed whether this replica is in sealed state.
    */
-  public AmbryReplica(ClusterMapConfig clusterMapConfig, AmbryPartition partition, boolean isReplicaStopped,
+  AmbryReplica(ClusterMapConfig clusterMapConfig, AmbryPartition partition, boolean isReplicaStopped,
       long capacityBytes, boolean isSealed) throws Exception {
     this.partition = Objects.requireNonNull(partition, "null partition");
     this.capacityBytes = capacityBytes;
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -42,6 +42,7 @@ import org.apache.helix.task.TaskStateModelFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static com.github.ambry.clustermap.ClusterMapUtils.*;
 import static com.github.ambry.clustermap.StateTransitionException.TransitionErrorCode.*;
 
 
@@ -203,12 +204,7 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
    */
   @Override
   public List<String> getSealedReplicas() {
-    InstanceConfig instanceConfig = helixAdmin.getInstanceConfig(clusterName, instanceName);
-    if (instanceConfig == null) {
-      throw new IllegalStateException(
-          "No instance config found for cluster: \"" + clusterName + "\", instance: \"" + instanceName + "\"");
-    }
-    return ClusterMapUtils.getSealedReplicas(instanceConfig);
+    return ClusterMapUtils.getSealedReplicas(getInstanceConfig());
   }
 
   /**
@@ -216,12 +212,12 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
    */
   @Override
   public List<String> getStoppedReplicas() {
-    InstanceConfig instanceConfig = helixAdmin.getInstanceConfig(clusterName, instanceName);
-    if (instanceConfig == null) {
-      throw new IllegalStateException(
-          "No instance config found for cluster: \"" + clusterName + "\", instance: \"" + instanceName + "\"");
-    }
-    return ClusterMapUtils.getStoppedReplicas(instanceConfig);
+    return ClusterMapUtils.getStoppedReplicas(getInstanceConfig());
+  }
+
+  @Override
+  public List<String> getDisabledReplicas() {
+    return ClusterMapUtils.getDisabledReplicas(getInstanceConfig());
   }
 
   @Override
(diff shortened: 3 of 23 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1033,6 +1064,14 @@ AtomicInteger getErrorCount() {
     return errorCount;
   }
 
+  /**
+   * Exposed for testing.
+   * @return a boolean value that indicates whether current store is disabled due to I/O error.
+   */
+  AtomicBoolean getDisabledOnError() {
```
cgtz, line 1071: for testing i would prefer having a setter instead of getter for the AtomicBoolean object.

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -512,35 +512,45 @@ public void onPartitionBecomeInactiveFromStandby(String partitionName) {
...
-            // we set error code to ReplicaNotFound because that is the only reason why compaction may fail.
-            throw new StateTransitionException("Couldn't disable compaction on replica " + replica.getReplicaPath(),
-                ReplicaNotFound);
-          }
-          logger.info("Compaction is successfully disabled on store {}", partitionName);
+        if (localStore.disabledOnError()) {
+          // if store is disabled due to disk I/O error, we explicitly throw an exception to mark partition as ERROR state
+          throw new StateTransitionException("Store " + partitionName + " is disabled due to I/O error",
+              ReplicaOperationFailure);
         } else {
```
cgtz, line 524: minor: since the added if throws, this code doesn't need to be in an else clause. This can shorten the diff a bit and seems to match the existing style of this method a bit better.

ambry-store/src/test/java/com/github/ambry/store/StorageManagerTest.java
```diff
@@ -357,15 +357,25 @@ public void replicaFromStandbyToInactiveTest() throws Exception {
       assertEquals("Error code doesn't match", StoreNotStarted, e.getErrorCode());
     }
     storageManager.startBlobStore(localReplica.getPartitionId());
-    // 4. success case (verify both replica's state and decommission file)
+    // 4. store is disabled due to disk I/O error
```
cgtz, line 360: is there a test for the `"Store " + partitionName + " is not found on current node"` case?

ambry-store/src/test/java/com/github/ambry/store/StoreTestUtils.java
```diff
@@ -140,6 +144,54 @@ public void setSealedState(boolean isSealed) {
     }
   }
 
+  public static class MockAmbryReplica extends AmbryReplica {
```
cgtz, line 147: Why can't `MockReplicaId` be used here?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -254,6 +245,33 @@ public boolean updateDataNodeInfoInCluster(ReplicaId replicaId, boolean shouldEx
...
+      } else if (disable && !disabledReplicas.contains(partitionName)) {
+        logger.info("Adding the partition {} to disabledReplicas list", partitionName);
+        disabledReplicas.add(partitionName);
+        setDisabledReplicas(disabledReplicas);
+      }
+
+      // invoke Helix native method to enable/disable partition on local node, this will trigger subsequent state transition
+      InstanceConfig instanceConfig = getInstanceConfig();
+      String resourceNameForPartition = getResourceNameOfPartition(helixAdmin, clusterName, partitionName);
+      instanceConfig.setInstanceEnabledForPartition(resourceNameForPartition, partitionName, !disable);
```
cgtz, line 270: it looks like this method just updates the ZnRecord java object. Do you need to call HelixAdmin.setInstanceConfig like was done in `setDisabledReplicas`. Also, I was wondering if we need the additional `disabledReplicas` field or if we could just use `HELIX_DISABLED_PARTITION`? I may be missing something, but it seems like they hold the same set of information and are updated together at the same time.

jsjtzyy: Good question. Yes, you are right, `disabledReplicas` and `HELIX_DISABLED_PARTITION` hold the same information. Let me explain a little more.
1. Why we need an extra disabled list?  `setInstanceEnabledForPartition` is Helix native method and is also used by "move replica" feature. When we decommission a replica, the first step is to call this method to disable replica. It will add `HELIX_DISABLED_PARTITION` map in InstanceConfig. There is a problem that when server restarts, it sees this map and cannot tell if the replica is disabled because of disk failure or because of replica decommission. Hence, I add this `disabledReplicas` to only hold replicas that are disabled due to disk failure and server will attempt to re-enable them.
2. Why we don't call `setInstanceConfig` again? `setInstanceEnabledForPartition` automatically updates InstanceConfig by change MapFields. Looks like it's not necessary to set InstanceConfig again. I also hesitate on calling `setInstanceConfig` as it overrides all fields (ListFields, MapFields and SimpleFields)
Hope this comment makes more sense.

ambry-api/src/main/java/com/github/ambry/store/Store.java
```diff
@@ -171,6 +171,13 @@
    */
   boolean recoverFromDecommission();
 
+  /**
+   * @return {@code true} if the store is disabled on error (due to disk I/O issue).
+   */
+  default boolean disabledOnError() {
```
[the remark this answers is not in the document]
jsjtzyy: No specific reason. I was trying to avoid adding this new method to other implementations. Maybe I should add it to other impl and throw exception there to make it consistent with other methods in this interface.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
```diff
@@ -25,7 +25,7 @@
 /**
  * {@link ReplicaId} implementation to use within dynamic cluster managers.
  */
-abstract class AmbryReplica implements ReplicaId {
+public abstract class AmbryReplica implements ReplicaId {
```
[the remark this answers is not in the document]
jsjtzyy: Making these public for mocking purpose (MockAmbryReplica extends this), I will try to figure out another way to mock and avoiding doing so.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
jsjtzyy: will do

ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixParticipant.java
```diff
@@ -24,20 +24,22 @@
...
   CountDownLatch listenerLatch = null;
   ReplicaState replicaState = ReplicaState.OFFLINE;
   ReplicaId currentReplica = null;
   ReplicaSyncUpManager replicaSyncUpService = null;
   private Set<ReplicaId> sealedReplicas = new HashSet<>();
   private Set<ReplicaId> stoppedReplicas = new HashSet<>();
+  private Set<ReplicaId> disabledReplicas = new HashSet<>();
   private PartitionStateChangeListener mockReplicationManagerListener;
 
   public MockHelixParticipant(ClusterMapConfig clusterMapConfig) {
```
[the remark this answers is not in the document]
jsjtzyy: yep, we can, good point

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -90,6 +91,7 @@
   private FileLock fileLock;
   private volatile ReplicaState currentState;
   private volatile boolean recoverFromDecommission;
+  private AtomicBoolean disabledOnError = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
jsjtzyy: fixed

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
jsjtzyy: Touche!

ambry-store/src/test/java/com/github/ambry/store/StoreTestUtils.java
jsjtzyy: `setReplicaDisabledState` in HelixParticipant checks if passed in replicaId is an instance of AmbryReplica

zzmao:
looks like a rebase needed.

ambry-store/src/test/java/com/github/ambry/store/StorageManagerTest.java
jsjtzyy: test case added

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
cgtz: Thanks for clarifying point 1. 

For 2, I am still a little confused about how `setInstanceEnabledforPartition` ensures that the change is reflected in zookeeper. It looks like all the implementation does is call `_record.setListField(InstanceConfigProperty.HELIX_DISABLED_PARTITION.toString(), list);`

This seems to only change the local state of the java object, but I thought that a call to HelixAdmin.setInstanceConfig was needed to actually call zookeeper client. This is what is done in the implementation of `setSealedReplicas`

jsjtzyy: yeah, you are right. I checked my previous usage of `setInstanceEnabledForPartition` in `HelixBootstrapUpgradeUtil`, it did call `setInstanceConfig` in the end. I will correct this part and probably re-arrange the order: 
 step1: calling `setInstanceEnabledForPartition`, make sure it updates the map fields (for our new version, it won't call `_record.setListField`)
 step2: updating our own disabled replica list
 step3: calling `setInstanceConfig` to make both of them take effect
What do you think?   (changing the order is to avoid `setInstanceEnabledForPartition` overrides the list field, although it should not)

cgtz: Yep, I think that order makes sense

jsjtzyy:
> looks like a rebase needed.

The PR is now ready for review  (cc @Arun-LinkedIn )

ambry-store/src/test/java/com/github/ambry/store/StorageManagerTest.java
```diff
@@ -357,15 +369,26 @@ public void replicaFromStandbyToInactiveTest() throws Exception {
       assertEquals("Error code doesn't match", StoreNotStarted, e.getErrorCode());
     }
     storageManager.startBlobStore(localReplica.getPartitionId());
-    // 4. success case (verify both replica's state and decommission file)
+    // 5. store is disabled due to disk I/O error
+    BlobStore localStore = (BlobStore) storageManager.getStore(localReplica.getPartitionId());
+    AtomicBoolean mockDisabledOnError = new AtomicBoolean(true);
+    localStore.setDisabledOnError(mockDisabledOnError);
```
zzmao, line 375: why pass in a new object instead the one in localStore?

jsjtzyy: This is to override boolean variable `disabledOnError` for testing purpose. We are not allowed to directly modify the `disabledOnError` within blobstore. Hope this makes sense.

PR #1535, opened 2020-05-26
Add lastUpdateTime for container and select container meet retention time during compaction

SophieGuo410:
This PR add a new field variable called lastUpdateTime for containers, so during compaction it will only compact the container meet with retention time at each compaction cycle.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/account/Container.java
+++ b/ambry-api/src/main/java/com/github/ambry/account/Container.java
@@ -61,7 +61,7 @@ public class Container {
   static final String CONTAINER_NAME_KEY = "containerName";
   static final String CONTAINER_ID_KEY = "containerId";
   static final String STATUS_KEY = "status";
-  static final String CONTAINER_LAST_UPDATE_TIME_KEY = "lastUpdateTime";
+  static final String CONTAINER_DELETE_TRIGGER_TIME_KEY = "deleteTriggerTime";
   static final String DESCRIPTION_KEY = "description";
   static final String IS_PRIVATE_KEY = "isPrivate";
   static final String BACKUP_ENABLED_KEY = "backupEnabled";
@@ -79,7 +79,7 @@ public class Container {
   static final boolean PREVIOUSLY_ENCRYPTED_DEFAULT_VALUE = ENCRYPTED_DEFAULT_VALUE;
   static final boolean MEDIA_SCAN_DISABLED_DEFAULT_VALUE = false;
   static final boolean TTL_REQUIRED_DEFAULT_VALUE = true;
-  static final long CONTAINER_LAST_UPDATE_TIME_DEFAULT_VALUE = 0;
+  static final long CONTAINER_DELETE_TRIGGER_TIME_DEFAULT_VALUE = 0;
   static final boolean SECURE_PATH_REQUIRED_DEFAULT_VALUE = false;
   static final boolean CACHEABLE_DEFAULT_VALUE = true;
   static final Set<String> CONTENT_TYPE_WHITELIST_FOR_FILENAMES_ON_DOWNLOAD_DEFAULT_VALUE = Collections.emptySet();
@@ -139,19 +139,19 @@ public class Container {
   public static final ContainerStatus UNKNOWN_CONTAINER_STATUS = ContainerStatus.ACTIVE;
 
   /**
-   * The last update time of {@link #UNKNOWN_CONTAINER}
+   * The delete trigger time of {@link #UNKNOWN_CONTAINER}
    */
-  public static final long UNKNOWN_CONTAINER_LAST_UPDATE_TIME = CONTAINER_LAST_UPDATE_TIME_DEFAULT_VALUE;
+  public static final long UNKNOWN_CONTAINER_DELETE_TRIGGER_TIME = CONTAINER_DELETE_TRIGGER_TIME_DEFAULT_VALUE;
 
   /**
-   * The last update time of {@link #DEFAULT_PUBLIC_CONTAINER}
+   * The delete trigger time of {@link #DEFAULT_PUBLIC_CONTAINER}
    */
-  public static final long DEFAULT_PUBLIC_CONTAINER_LAST_UPDATE_TIME = CONTAINER_LAST_UPDATE_TIME_DEFAULT_VALUE;
+  public static final long DEFAULT_PUBLIC_CONTAINER_DELETE_TRIGGER_TIME = CONTAINER_DELETE_TRIGGER_TIME_DEFAULT_VALUE;
 
   /**
-   * The last update time of {@link #DEFAULT_PRIVATE_CONTAINER}
+   * The delete trigger time of {@link #DEFAULT_PRIVATE_CONTAINER}
    */
-  public static final long DEFAULT_PRIVATE_CONTAINER_LAST_UPDATE_TIME = CONTAINER_LAST_UPDATE_TIME_DEFAULT_VALUE;
+  public static final long DEFAULT_PRIVATE_CONTAINER_DELETE_TRIGGER_TIME = CONTAINER_DELETE_TRIGGER_TIME_DEFAULT_VALUE;
 
   /**
    * The status for the containers to be associated with the blobs that are put without specifying a target container,
@@ -291,7 +291,7 @@ public class Container {
           UNKNOWN_CONTAINER_PREVIOUSLY_ENCRYPTED_SETTING, UNKNOWN_CONTAINER_CACHEABLE_SETTING,
           UNKNOWN_CONTAINER_MEDIA_SCAN_DISABLED_SETTING, null, UNKNOWN_CONTAINER_TTL_REQUIRED_SETTING,
           SECURE_PATH_REQUIRED_DEFAULT_VALUE, CONTENT_TYPE_WHITELIST_FOR_FILENAMES_ON_DOWNLOAD_DEFAULT_VALUE,
-          BACKUP_ENABLED_DEFAULT_VALUE, UNKNOWN_CONTAINER_PARENT_ACCOUNT_ID, UNKNOWN_CONTAINER_LAST_UPDATE_TIME);
+          BACKUP_ENABLED_DEFAULT_VALUE, UNKNOWN_CONTAINER_PARENT_ACCOUNT_ID, UNKNOWN_CONTAINER_DELETE_TRIGGER_TIME);
 
   /**
    * A container defined specifically for the blobs put without specifying target container but isPrivate flag is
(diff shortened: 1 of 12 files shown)


ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
```diff
@@ -29,6 +29,7 @@
   private short id;
   private String name;
   private ContainerStatus status;
+  private long lastUpdateTime;
```
justinlin-linkedin, line 32: we can give lastUpdateTime a default value
```
private long lastUpdateTime = System. currentTimeMillis();
```
This way we don't have to change the constructor's signature.

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
```diff
@@ -78,12 +80,13 @@ public ContainerBuilder(Container origin) {
    * @param description The description of the {@link Container}.
    * @param parentAccountId The id of the parent {@link Account} of the {@link Container} to build.
    */
-  public ContainerBuilder(short id, String name, ContainerStatus status, String description, short parentAccountId) {
+  public ContainerBuilder(short id, String name, ContainerStatus status, String description, short parentAccountId, long lastUpdateTime) {
```
justinlin-linkedin, line 83: The beauty of having builder pattern is to avoid this massive change when adding a new field. I suggest that we make lastUpdateTime as an optional here. We can give lastUpdateTime a default value and only set the lastUpdateTime when it's not the default value.

```
public ContainerBuilder(short id, String name, ContainerStatus status, String description, short parentAccountId) {
    this.id = id;
    this.name = name;
    this.status = status;
    this.description = description;
    this.parentAccountId = parentAccountId;
}

public ContainerBuilder lastUpdateTime(long val) {
    this.lastUpdateTime = val;
    return this;
}
```

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
```diff
@@ -116,6 +119,14 @@ public ContainerBuilder setStatus(ContainerStatus status) {
     return this;
   }
 
+  /**
+   * Sets the last update time of the {@link Container} to build.
+   */
+  public Long setLastUpdateTime(Long lastUpdateTime) {
```
justinlin-linkedin, line 125: this is a bad signature in builder. It should be 
```
public ContainerBuilder lastUpdateTime(long val) {
    this.lastUpdateTime = val;
    return this;
}
```
It's up to you to prefix `set` to the method name. I would not, but to keep it consistent with all the methods in this file, we probably should.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -478,6 +498,7 @@ JSONObject toJson() throws JSONException {
         metadata.put(JSON_VERSION_KEY, JSON_VERSION_1);
         metadata.put(CONTAINER_ID_KEY, id);
         metadata.put(CONTAINER_NAME_KEY, name);
+        metadata.put(CONTAINER_LAST_UPDATE_TIME_KEY, lastUpdateTime);
```
justinlin-linkedin, line 501: please don't change the current version.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -487,6 +508,7 @@ JSONObject toJson() throws JSONException {
         metadata.put(Container.JSON_VERSION_KEY, JSON_VERSION_2);
         metadata.put(CONTAINER_ID_KEY, id);
         metadata.put(CONTAINER_NAME_KEY, name);
+        metadata.put(CONTAINER_LAST_UPDATE_TIME_KEY, lastUpdateTime);
```
justinlin-linkedin, line 511: Please don't change the current version. Please add JSON_VERSON_3 and only in version_3 can we put container_last_update_time_key.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -535,6 +557,12 @@ public ContainerStatus getStatus() {
     return status;
   }
 
+  /**
+   * Gets the last update time of the container.
+   * @return The last update time of the container.
+   */
+  public Long getLastUpdateTime() {return lastUpdateTime;}
```
justinlin-linkedin, line 564: please format this methood.

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
cgtz: Since we have already started using set for the other methods, and this builder is shared with Nuage, I think we should keep prefixing these with `set`. Not sure how this builder ended up having different naming conventions than all the other ones though.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -136,6 +137,21 @@
...
 
+  /**
+   * The last update time of {@link #UNKNOWN_CONTAINER}
+   */
+  public static final Long UNKNOWN_CONTAINER_LAST_UPDATE_TIME = System.currentTimeMillis();
+
+  /**
+   * The last update time of {@link #DEFAULT_PUBLIC_CONTAINER}
+   */
+  public static final Long DEFAULT_PUBLIC_CONTAINER_LAST_UPDATE_TIME = System.currentTimeMillis();
```
cgtz, line 148: This will set the default time to whatever the timestamp was when this classes static initialization completed. Maybe the default time should be 0 for consistency? Also, generally prefer long over Long when there is no need for a primitive to be null and no need to use the long as part of a generic. This avoids unnecessary unboxing penalties.

ambry-api/src/main/java/com/github/ambry/account/Container.java
cgtz: I think we should not create a version 3. version 2 was only created since we removed a field between v1 and v2.

However, you are right that adding a new required field to an existing version will cause issues when deserializing existing records. Generally the best way to make the migration seamless for json documents is to make any new field optional in version 2.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -353,6 +371,7 @@ private Container(JSONObject metadata, short parentAccountId) throws JSONExcepti
         id = (short) metadata.getInt(CONTAINER_ID_KEY);
         name = metadata.getString(CONTAINER_NAME_KEY);
         status = ContainerStatus.valueOf(metadata.getString(STATUS_KEY));
+        lastUpdateTime = metadata.getLong(CONTAINER_LAST_UPDATE_TIME_KEY);
```
cgtz, line 374: any new field added to an existing version should be optional so existing records can be deserialized by the new code.

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
SophieGuo410: Updated.

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
SophieGuo410: Updated to make lastUpdateTime optional.

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
SophieGuo410: Return ContainerBuilder instead of Long type and keep name prefixing with set for consistency.

ambry-api/src/main/java/com/github/ambry/account/Container.java
SophieGuo410: Removed put for version one.

ambry-api/src/main/java/com/github/ambry/account/Container.java
SophieGuo410: Make the new field optional in version2.

ambry-api/src/main/java/com/github/ambry/account/Container.java
SophieGuo410: Done!

ambry-api/src/main/java/com/github/ambry/account/Container.java
SophieGuo410: Updated the type to long and change the default time to 0 for consistency.

ambry-api/src/main/java/com/github/ambry/account/Container.java
SophieGuo410: In version1, I assigned the default value which is 0. In version2, I made it optional for the new field.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -61,6 +61,7 @@
   static final String CONTAINER_NAME_KEY = "containerName";
   static final String CONTAINER_ID_KEY = "containerId";
   static final String STATUS_KEY = "status";
+  static final String CONTAINER_LAST_UPDATE_TIME_KEY = "lastUpdateTime";
```
zzmao, line 64: The PR looks good after addressing people's comments.

I have one concern:  `lastUpdateTime` sounds like a general update time. Any update can refer to it. 
But in this PR, it is used as delete triggered time.

SophieGuo410: I named the field in a general way intentionally cause for container deletion the lastUpdateTime should be when user mark container as DELETE_IN_PROGRESS(since after user mark the status, ACLs would be removed and they can't access this container). But for future we may use this field for different purpose. What do you think? If naming style should be more specific, I can change to deleteTriggerTime.

zzmao: If it's for general use cases, you might need to update this value for all container related update. Not only ACL changes, but all other container metadata changes. I prefer a narrow name:  `deleteTriggerTime `. @justinlin-linkedin @cgtz  any opinion?

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -487,6 +509,7 @@ JSONObject toJson() throws JSONException {
         metadata.put(Container.JSON_VERSION_KEY, JSON_VERSION_2);
         metadata.put(CONTAINER_ID_KEY, id);
         metadata.put(CONTAINER_NAME_KEY, name);
+        metadata.putOpt(CONTAINER_LAST_UPDATE_TIME_KEY, lastUpdateTime);
```
justinlin-linkedin, line 512: nit: we know the key and value won't be null, so we can just use metadata.put.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -198,14 +197,17 @@ void compact(CompactionDetails details, byte[] bundleReadBuffer) throws IOExcept
...
+   * Filters deprecated {@link Container}s for compaction purpose. Deprecated containers include DELETE_IN_PROGRESS
+   * containers met with retention time and all INACTIVE containers.
    * @return the deprecated {@link Container}s' accountId & containerId pairs.
    */
   private void getDeprecatedContainers() {
     deprecatedContainers.clear();
     if (accountService != null) {
       accountService.getContainersByStatus(Container.ContainerStatus.DELETE_IN_PROGRESS).forEach((container) -> {
-        deprecatedContainers.add(new Pair<>(container.getParentAccountId(), container.getId()));
+        if (container.getLastUpdateTime() + TimeUnit.DAYS.toMillis(14) >= System.currentTimeMillis()) {
```
justinlin-linkedin, line 208: nit: we should use a config for this value, maybe we can update it in next pr.

ambry-api/src/main/java/com/github/ambry/account/Container.java
justinlin-linkedin: I am fine with deleteTriggerTime. From my past experience, we actually don't update  container that much, except for adding download content type in the whitelist and delete container. Of course we can update the container in anyway allowed but we just don't. And we don't even have a creation time, which indicates we just don't care about update time at all. So a delete time should be enough for a container.
That being said, if we start to care more about when the container is created, update and deleted, then, we might have to at least have a creation time and update time.

SophieGuo410: Thank you all for the great explanation. Updated to deleteTriggerTime.

ambry-api/src/main/java/com/github/ambry/account/Container.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
SophieGuo410: Updated.

PR #1536, opened 2020-05-27
Fix incorrect offline partition count in HelixParticipant

jsjtzyy:
Offline partition is currently counted based how many times the reset
method is called. However, Helix may call reset method against same
partition multiple times during zk disconnection or graceful shutdown.
This causes incorrect number of offline partitions and triggers false
alarm. To fix that, this PR keeps track of each local partition and its
state, which also supports tracking added/removed replicas on local node.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipantMetrics.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipantMetrics.java
@@ -16,47 +16,65 @@ package com.github.ambry.clustermap;
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
-import java.util.Collections;
+import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.atomic.AtomicInteger;
 
 
 /**
  * Metrics for {@link HelixParticipant} to monitor partition state transitions.
  */
 class HelixParticipantMetrics {
-  final AtomicInteger offlineCount = new AtomicInteger();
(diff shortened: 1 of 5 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipantMetrics.java
```diff
@@ -16,55 +16,47 @@
...
     metricRegistry.register(MetricRegistry.name(HelixParticipant.class, "inactivePartitionCount" + zkSuffix),
         inactivePartitionCount);
-    Gauge<Integer> offlinePartitionCount = offlineCount::get;
+    Gauge<Integer> offlinePartitionCount =
+        () -> Collections.frequency(localPartitionAndState.values(), ReplicaState.OFFLINE);
     metricRegistry.register(MetricRegistry.name(HelixParticipant.class, "offlinePartitionCount" + zkSuffix),
         offlinePartitionCount);
-    Gauge<Integer> errorStatePartitionCount = errorStateCount::get;
+    Gauge<Integer> errorStatePartitionCount =
+        () -> Collections.frequency(localPartitionAndState.values(), ReplicaState.ERROR);
```
lightningrob, line 56: This definitely is cleaner logic wise.
My only concern is that the gauge readings will result in 6 full scans of the map every minute (or whatever frequency), which could get expensive if the number of replicas is large.  A possible optimization would be to do a single scan periodically and update local counters for each state, which the gauges could read as before.  But the performance benefit may be too small to justify this.

jsjtzyy: Fair point, I will try to scan once and cache the result for other 5 metrics.

PR #1539, opened 2020-05-28
Return failure reason in http header for undelete

justinlin-linkedin:
Return failure reason in http response header for undelete. Since undelete is not exposed outside of the linked datacenter and should be an admin operation, we should always return the failure reason in the header.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
@@ -221,7 +221,6 @@ public class RestUtils {
      * The lifeVersion of the blob.
      */
     public final static String LIFE_VERSION = "x-ambry-life-version";
-
   }
 
   public static final class TrackingHeaders {
@@ -297,6 +296,11 @@ public class RestUtils {
      * The key for the {@link RequestPath} that represents the parsed path of an incoming request.
      */
     public static final String REQUEST_PATH = KEY_PREFIX + "request-path";
+
+    /**
+     * To be set to {@code true} if failures reason should be attached to frontend responses.
+     */
+    public static final String SEND_FAILURE_REASON = KEY_PREFIX + "send-failure-reason";
   }
 
   /**
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
@@ -113,10 +113,9 @@ class FrontendRestRequestService implements RestRequestService {
    * @param hostname the hostname for this frontend.
    * @param clusterName the name of the storage cluster that the router communicates with.
    */
-  FrontendRestRequestService(FrontendConfig frontendConfig, FrontendMetrics frontendMetrics,
-      Router router, ClusterMap clusterMap, IdConverterFactory idConverterFactory,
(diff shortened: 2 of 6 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -257,6 +256,10 @@ public void handlePut(RestRequest restRequest, RestResponseChannel restResponseC
       } else if (requestPath.matchesOperation(Operations.UNDELETE) && frontendConfig.enableUndelete) {
         // If the undelete is not enabled, then treat it as unrecognized operation.
         undeleteHandler.handle(restRequest, restResponseChannel, (r, e) -> {
+          if (e instanceof RouterException) {
+            e = new RestServiceException(e.getMessage(),
+                RestServiceErrorCode.getRestServiceErrorCode(((RouterException) e).getErrorCode()), true);
```
cgtz, line 261: sorry, I realize that setting this option whenever a failure reason needs to be sent creates a lot of duplicate code. Maybe a better solution would be to have a request-level internal key that can be set to true when that request should be allowed to receive error info (for any exception type).

THe approach in this PR should work, though!

justinlin-linkedin: This is a good idea, updated the PR to use an internal key.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -297,6 +296,11 @@
      * The key for the {@link RequestPath} that represents the parsed path of an incoming request.
      */
     public static final String REQUEST_PATH = KEY_PREFIX + "request-path";
+
+    /**
+     * To be set to {@code true} if failures reason should be attached to frontend responses.
+     */
+    public static final String SEND_FAILURE_REASON = KEY_PREFIX + "send-failure-reason";
```
cgtz, line 303: thanks for making this change! I can change some of the code inside of AmbryLI to use this instead of the old approach.

ambry-test-utils/src/main/java/com/github/ambry/router/InMemoryRouter.java
```diff
@@ -335,15 +335,13 @@ public void setLifeVersion(short lifeVersion) {
...
-      } else if (undeletedBlobs.contains(blobId)) {
-        exception = new RouterException("Blob already undeleted", RouterErrorCode.BlobUndeleted);
       } else if (!deletedBlobs.contains(blobId)) {
         exception = new RouterException("Blob not deleted", RouterErrorCode.BlobNotDeleted);
       }
       undeletedBlobs.add(blobId);
       deletedBlobs.remove(blobId);
       if (notificationSystem != null) {
-        notificationSystem.onBlobDeleted(blobId, serviceId, null, null);
+        notificationSystem.onBlobUndeleted(blobId, serviceId, null, null);
```
zzmao, line 344: why is this changed to `onBlobUndeleted `?

justinlin-linkedin: This was a mistake, this is actually undelete function, but I copied the function from delete and forgot to change onBlobDeleted to onBlobUndeleted. Now I am changing it.

PR #1540, opened 2020-05-29
Make sure to rebalance helix cluster after updating ideal state in Vcr Update Tool

ankagrawal:


lightningrob:
@jsjtzyy The second run (with --dest only) only needs to happen when we change the configuration like rebalancer, IdealState properties.  Not when adding new partitions.

lightningrob:
I added the DelayedAutoRebalancer on recommendation from Helix team to solve the crazy bouncing we were observing.  We are applying it for the first time now.
I also thought we should add rebalanceDelayTime in config and asked Helix team for recommendation, they seemed to say it wasn't needed.  I agree it seems like a good idea.  What value, 1 minute?

PR #1541, opened 2020-05-29
Replace string concatenation in logger calls

cgtz:
When done inside of hot path logger.trace calls, this can result in
quite a few unnecessary string allocations. For example, there were a
lot of such calls inside of PersistentIndex.

Used intellij inspections and autosuggestions to find and replace across
the entire project.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/BackupFileManager.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/BackupFileManager.java
@@ -125,7 +125,7 @@ class BackupFileManager {
     File[] files = backupDir.listFiles(tempFileFilter);
     if (files != null) {
       for (File file : files) {
-        logger.trace("Delete temp file " + file.getName());
+        logger.trace("Delete temp file {}", file.getName());
         tryDeleteFile(file);
       }
     }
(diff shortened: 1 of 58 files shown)


ssen-li:
Have few minor comments/clarifications.
1. If the configured log <level> is higher than logger.<level>("str1" + "str2") then this string concatenation should not even happen, right?
2. I found few other places where concatenation is used but they are most likely not in hot path. For ex. https://github.com/linkedin/ambry/blob/master/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java#L224

cgtz:
> 1. If the configured log level is higher than logger.level("str1" + "str2") then this string concatenation should not even happen, right?
That is the benefit of using the `"{}"`text replacement syntax. The arguments will be passed to `Logger` but no string concatenation will happen if the log level is disabled. If you write `logger.level("a" + nonConstantThing)` the string concatenation will happen before the call to `logger.level`since java needs to build that argument to call the method with.
> 2. I found few other places where concatenation is used but they are most likely not in hot path. For ex. https://github.com/linkedin/ambry/blob/master/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java#L224
I suppose that intellij did not flag that case since it is a string concatenation of two constants. If all of the things being concatenated are constants, javac will do it at compile time.

PR #1542, opened 2020-05-29
ambry server authorization changes.

ssen-li:
This PR contains following changes.
1. interface definition for ServerSecurityService in ambry-server. It contains 2 APIs. validateConnection API will be used to enforce security check when new HTTP2 connection terminates at server, SSL handshake is done and before any streams are processed by server.
validateRequest API will NOT be used now. It may be used later if we decide to secure individual request seperately from the connection. 
 
2. A default implementation of the above interface(AmbryServerSecurityService) which does not do any certificate based validations. it's allow all. 

3. modification of the Http2 netty pipeline(ServerSecurityChecker) to enforce the security check when a new HTTP2 connections comes in to the server. 

4. UT for new codes. 

5. a configuration to choose the implementation class for ServerSecurityService and new metrics for tracking the feature.

6. Tested through UT and also by running ambry-frontend talking to ambry-server over HTTP2 SSL connection.


ambry-rest/src/main/java/com/github/ambry/rest/StorageServerNettyChannelInitializer.java
```diff
@@ -93,6 +96,7 @@ protected void initChannel(SocketChannel ch) throws Exception {
     int peerPort = peerAddress.getPort();
     SslHandler sslHandler = new SslHandler(sslFactory.createSSLEngine(peerHost, peerPort, SSLFactory.Mode.SERVER));
     pipeline.addLast("SslHandler", sslHandler);
+    pipeline.addLast("securityChecker", this.serverSecurityChecker);
```
zzmao, line 99: `this` is not needed here. Align with other.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+import com.github.ambry.rest.RestRequest;
+import com.github.ambry.rest.RestServiceErrorCode;
+import com.github.ambry.rest.RestServiceException;
+import com.github.ambry.router.Callback;
+import io.netty.channel.ChannelHandlerContext;
+
+/**
+ * Default implementation of {@link ServerSecurityService} for Ambry that doesn't do any validations.
+ */
+public class AmbryServerSecurityService implements ServerSecurityService {
```
zzmao, line 28: How do you think of moving this class to `com.github.ambry.rest` ? so that frontend can also leverage it if needed.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.concurrent.Future;
+
+
+/**
+ * Responsible for performing any security validations on the HTTP2 connection terminating on server. Validations
+ * could involve authentication, authorization, security checks and so on which the implementation can decide.
+ * This could also involve setting headers while responding, based on the request.
+ * Exceptions are returned via {@link Callback}s on any validation failure.
+ */
+public interface ServerSecurityService extends Closeable {
```
zzmao, line 31: Move to `com.github.ambry.rest`? 
This can also help ambry-api get rid of netty dependency.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+   */
+  void validateRequest(RestRequest restRequest, Callback<Void> callback);
+
+  /**
+   * Similar to {@link #validateConnection(ChannelHandlerContext, Callback)} but returns a {@link Future}
+   * instead of requiring a callback.
+   * @param ctx {@link ChannelHandlerContext} upon which validations has to be performed
+   * @return a {@link Future} that is completed when the processing is done.
+   */
+  default Future<Void> validateConnection(ChannelHandlerContext ctx) {
```
zzmao, line 55: You can return Future in validateRequest()

There are some prior code did so:

```
GetBlobOperation
public Future<Long> readInto(AsyncWritableChannel asyncWritableChannel, Callback<Long> callback) {
```

I prefer fewer interface. More interfaces means more maintenance.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Objects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Handler that performs any security validation on the new HTTP2 connection.
+ */
+@ChannelHandler.Sharable
+public class ServerSecurityChecker extends ChannelInboundHandlerAdapter {
```
zzmao, line 31: serverSecurityChecker -> serverSecurityHanlder ? To make it netty style.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
```diff
@@ -0,0 +1,88 @@
...
+        } else {
+          logger.debug("SSL handshake succedded for channel: {}", ctx.channel(), future.cause());
+          try {
+            serverSecurityService.validateConnection(ctx, (r, e) -> {
+              if (e != null) {
+                logger.error("security validation failed for channel: {}", ctx.channel(), e);
+                serverMetrics.serverValidateConnectionFailure.inc();
+                ctx.channel().close();
+              } else {
+                logger.info("security validation succeeded for channel: {}", ctx.channel());
```
zzmao, line 75: debug or trace level, otherwise too many logs.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
```diff
@@ -0,0 +1,88 @@
...
+                serverMetrics.serverValidateConnectionSuccess.inc();
+              }
+            });
+          } catch (Exception e) {
+            logger.error("security validation failed for channel: {}", ctx.channel(), e);
+            serverMetrics.serverValidateConnectionFailure.inc();
+            ctx.channel().close();
```
zzmao, line 82: what If there is stream channel has been establish?

ambry-rest/src/main/java/com/github/ambry/rest/StorageServerNettyChannelInitializer.java
ssen-li: fixed it.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
ssen-li: rename the class as suggested. serverSecurityChecker -> serverSecurityHandler.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
ssen-li: changed it to debug.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityService.java
ssen-li: frontend already has a SecurityService interface/class which does the security validations. This is defined in ambry-api and implmented in ambry-frontend. I mainly followed the same pattern from this frontend class to apply for server.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
cgtz: I disagree on this specific comment. I don't like the pattern of a single method both returning the future and taking a callback since this requires the implementation to deal with both at the same time (and pass them both around internally instead of just one object). I think using default methods is a good way to easily convert a callback based api into a future based one.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
ssen-li: I verified that stream channel is not established at this point. Since this code is getting executed in the same nioEventLoopGroup thread, server will not be processing any other HTTP2 frames for this connection before this handshake callback finishes.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+ */
+public interface ServerSecurityService extends Closeable {
+
+  /**
+   * Performs security validations (if any) before allowing the HTTP2 connection setup to be complete and invokes the
+   * {@code callback} once done.
+   * @param ctx the {@link ChannelHandlerContext} to process.
+   * @param callback the callback to invoke once processing is finished.
+   */
+  void validateConnection(ChannelHandlerContext ctx, Callback<Void> callback);
```
cgtz, line 39: This interface looks a little netty focused and requires a new dependency for ambry-api. Could this method just take in the `SSLSession` object itself?

Additionally, I wanted to learn a bit more about the motivation for this method as opposed to doing all security validation at request time. Is this an optimization so that SSL related checks only have to be checked once since the peer certificate is the same for all requests on a single connection?

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
ssen-li: I kind of followed the same pattern as defined for SecurityService interface defined for frontend, so kept the api similar to those.

ambry-test-utils/src/main/java/com/github/ambry/rest/MockChannelHandlerContext.java
```diff
@@ -0,0 +1,247 @@
...
+import io.netty.util.Attribute;
+import io.netty.util.AttributeKey;
+import io.netty.util.concurrent.EventExecutor;
+import java.net.SocketAddress;
+
+/**
+ * Mock class for ChannelHandlerContext
+ */
+
+public class MockChannelHandlerContext implements ChannelHandlerContext {
```
cgtz, line 36: This may be a good place to use mockito since you only need to mock a very small selection of the many methods.

build.gradle
```diff
@@ -189,6 +189,7 @@ project(':ambry-api') {
         compile "org.codehaus.jackson:jackson-core-asl:$jacksonVersion"
         compile "org.codehaus.jackson:jackson-mapper-asl:$jacksonVersion"
         compile "io.netty:netty-buffer:$nettyVersion"
+        compile "io.netty:netty-all:$nettyVersion"
```
cgtz, line 192: let's try to see if we can avoid adding the netty-all dependency to ambry-api. If we cannot avoid it, we can remove the dependencies on netty-buffer, since netty-all includes it already.

ambry-server/src/test/java/com/github/ambry/server/AmbryServerSecurityServiceTest.java
```diff
@@ -0,0 +1,103 @@
...
+    //success case
+    EmbeddedChannel channel = new EmbeddedChannel();
+    ChannelHandlerContext ctx = new MockChannelHandlerContext(channel);
+    try {
+      serverSecurityService.validateConnection(ctx, (r, e) -> {
+        Assert.assertNull("result not null", r);
+        Assert.assertNull("exception not null", e);
+      });
+    } catch (Exception e) {
+      Assert.fail("unexpected exception happened" + e);
```
cgtz, line 57: I know some ambry test cases may have had this assert.fail pattern, but I think it is generally better to let the exception be thrown in test cases instead of catching it and wrapping it in an assert. Throwing the exception gives us the full stack trace (and still fails the test) which can be helpful in tracing down the issue if the test happens to fail.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
cgtz: Yep, I think I made the changes to SecurityService to use default methods for the future based APIs since I disliked the Future+callback in one method pattern, but there are multiple opinions on this.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
```diff
@@ -0,0 +1,88 @@
...
+  private void validateSslConnection(ChannelHandlerContext ctx) throws Exception {
+    SslHandler sslHandler = ctx.pipeline().get(SslHandler.class);
+    if (sslHandler != null) {
+      sslHandler.handshakeFuture().addListener(future -> {
+        if (!future.isSuccess()) {
+          logger.debug("SSL handshake failed for channel: {}", ctx.channel(), future.cause());
+        } else {
+          logger.debug("SSL handshake succedded for channel: {}", ctx.channel(), future.cause());
+          try {
+            serverSecurityService.validateConnection(ctx, (r, e) -> {
```
lightningrob: securityServiceValidateConnectionTimeInMs should be measured around this call.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+  public void validateConnection(ChannelHandlerContext ctx, Callback<Void> callback) {
+    Exception exception = null;
+    serverMetrics.securityServiceValidateConnectionRate.mark();
+    long startTimeMs = System.currentTimeMillis();
+    if (!isOpen) {
+      exception = new RestServiceException("ServerSecurityService is closed", RestServiceErrorCode.ServiceUnavailable);
+    } else if (ctx == null) {
+      throw new IllegalArgumentException("ctx is null");
+    }
+    serverMetrics.securityServiceValidateConnectionTimeInMs.update(System.currentTimeMillis() - startTimeMs);
```
lightningrob, line 49: The timers should be moved to the calling method.  This is a no-op impl that doesn't need to be timed.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+  public void validateRequest(RestRequest restRequest, Callback<Void> callback) {
+    Exception exception = null;
+    serverMetrics.securityServiceValidateRequestRate.mark();
+    long startTimeMs = System.currentTimeMillis();
+    if (!isOpen) {
+      exception = new RestServiceException("ServerSecurityService is closed", RestServiceErrorCode.ServiceUnavailable);
+    } else if (restRequest == null) {
+      throw new IllegalArgumentException("restRequest is null");
+    }
+    serverMetrics.securityServiceValidateRequestTimeInMs.update(System.currentTimeMillis() - startTimeMs);
```
lightningrob, line 63: Remove timer.

ambry-server/src/test/java/com/github/ambry/server/AmbryServerSecurityServiceTest.java
```diff
@@ -0,0 +1,103 @@
...
+import com.github.ambry.utils.ThrowingConsumer;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.embedded.EmbeddedChannel;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+import org.junit.Assert;
+import org.junit.Test;
+
+
+public class AmbryServerSecurityServiceTest {
```
lightningrob, line 37: Javadocs on class and methods.

ambry-test-utils/src/main/java/com/github/ambry/rest/MockChannelHandlerContext.java
```diff
@@ -0,0 +1,247 @@
+/**
+ * Copyright 2016 LinkedIn Corp. All rights reserved.
```
lightningrob, line 2: 2020

build.gradle
lightningrob, line 192: Agree with Ze it would be best to avoid.  Also, doesn't netty-all include netty-buffer?

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
lightningrob: I can see @cgtz 's point about cumbersome implementation. I'm fine with default.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
ssen-li: 1. Thanks @cgtz for the idea. agree it is netty focused. I have changed the interface to accept SSLSession instead without loosing any information. 
2. Yes, as mentioned in the document, currently I am planning to do the security validation on the certificate provided during the HTTP2 SSL connection setup itself. There is no separate certs/SSL negotiation for individual request stream inside the HTTP2 connection. So There is no need to do the cert validation on per request basis. I have added the validateRequest() interface now as well to do any extra security checks for individual request that we may want to do in future.

ambry-test-utils/src/main/java/com/github/ambry/rest/MockChannelHandlerContext.java
ssen-li: Thanks, this class is actually not needed anymore. I have used mock for the SSLSession object instead.

build.gradle
ssen-li: I have reverted the change in this file as it's no longer needed. I am passing SSLSession instead of ChannelHandlerContext in the validateConnection() interface.

ambry-server/src/test/java/com/github/ambry/server/AmbryServerSecurityServiceTest.java
ssen-li: Thanks for the suggestion. removed the try/catch.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+import com.github.ambry.rest.RestRequest;
+import com.github.ambry.router.Callback;
+import com.github.ambry.router.FutureResult;
+import java.io.Closeable;
+import io.netty.channel.ChannelHandlerContext;
+import java.util.concurrent.Future;
+
+
+/**
+ * Responsible for performing any security validations on the HTTP2 connection terminating on server. Validations
```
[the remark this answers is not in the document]
ssen-li: done.

ambry-api/src/main/java/com/github/ambry/server/ServerSecurityService.java
```diff
@@ -0,0 +1,72 @@
...
+   * {@code callback} once done.
+   * @param ctx the {@link ChannelHandlerContext} to process.
+   * @param callback the callback to invoke once processing is finished.
+   */
+  void validateConnection(ChannelHandlerContext ctx, Callback<Void> callback);
+
+  /**
+   * Performs security validations (if any) on the individual stream {@link RestRequest} asynchronously and invokes the
+   * {@link Callback} when the validation completes.
+   * @param restRequest {@link RestRequest} upon which validations has to be performed
```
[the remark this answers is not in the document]
ssen-li: done.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
```diff
@@ -0,0 +1,88 @@
...
+   * Listen to the handshake future for the SSL termination, then apply the security policies, potentialy closing the
+   * connection in case security check fails.
+   * @param ctx the {@link ChannelHandlerContext}.
+   */
+  private void validateSslConnection(ChannelHandlerContext ctx) throws Exception {
+    SslHandler sslHandler = ctx.pipeline().get(SslHandler.class);
+    if (sslHandler != null) {
+      sslHandler.handshakeFuture().addListener(future -> {
+        if (!future.isSuccess()) {
+          logger.debug("SSL handshake failed for channel: {}", ctx.channel(), future.cause());
```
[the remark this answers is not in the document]
ssen-li: SSL handshake may fails due to many reason like expired certs, not signed by same root ca, etc. None of these are frequent. I have changed the level to error.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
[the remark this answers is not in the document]
ssen-li: This handshakefuture callback is not executed in the same code flow as channelActive. In channelActive() we are just registering the callback for ssl handshake completion future.  Actual handshake happens in separate code flow and once completed this callback will be called.

ambry-rest/src/main/java/com/github/ambry/rest/StorageServerNettyFactory.java
```diff
@@ -61,10 +65,11 @@ public StorageServerNettyFactory(int http2Port, VerifiableProperties verifiableP
     nettyMetrics = new NettyMetrics(metricRegistry);
     Http2ClientConfig http2ClientConfig = new Http2ClientConfig(verifiableProperties);
     ConnectionStatsHandler connectionStatsHandler = new ConnectionStatsHandler(nettyMetrics);
+    this.serverSecurityChecker = new ServerSecurityChecker(serverSecurityService, serverMetrics);
 
     Map<Integer, ChannelInitializer<SocketChannel>> initializers = Collections.singletonMap(http2Port,
-        new StorageServerNettyChannelInitializer(nettyConfig, http2ClientConfig, performanceConfig, nettyMetrics, connectionStatsHandler,
-            requestHandler, sslFactory, metricRegistry));
+        new StorageServerNettyChannelInitializer(nettyConfig, http2ClientConfig, performanceConfig, nettyMetrics,
+            connectionStatsHandler, requestHandler, sslFactory, metricRegistry, this.serverSecurityChecker));
```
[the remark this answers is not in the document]
ssen-li: done.

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
```diff
@@ -232,6 +233,10 @@ public void startup() throws InstantiationException {
         statsManager.start();
       }
 
+      ServerSecurityServiceFactory serverSecurityServiceFactory =
+          Utils.getObj(serverConfig.serverSecurityServiceFactory, properties, this.metrics);
```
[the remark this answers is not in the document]
ssen-li: done.

ambry-rest/src/main/java/com/github/ambry/rest/ServerSecurityChecker.java
ssen-li: done.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityService.java
ssen-li: done.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityService.java
ssen-li: done.

ambry-server/src/test/java/com/github/ambry/server/AmbryServerSecurityServiceTest.java
ssen-li: done.

ambry-test-utils/src/main/java/com/github/ambry/rest/MockChannelHandlerContext.java
ssen-li: i have removed this class completely.

build.gradle
ssen-li: reverted this change as I have removed the dependency on netty-all from ambry-api anymore.

ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
```diff
@@ -422,6 +433,11 @@ public ServerMetrics(MetricRegistry registry, Class<?> requestClass, Class<?> se
       serverShutdownTimeInMs = null;
     }
 
+    securityServiceValidateConnectionTimeInMs =
+        registry.histogram(MetricRegistry.name(requestClass, "securityServiceValidateConnectionTimeInMs"));
```
zzmao, line 437: s->S

ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
```diff
@@ -422,6 +433,11 @@ public ServerMetrics(MetricRegistry registry, Class<?> requestClass, Class<?> se
...
+    securityServiceValidateRequestTimeInMs =
+        registry.histogram(MetricRegistry.name(requestClass, "securityServiceValidateRequestTimeInMs"));
```
zzmao, line 439: s->S

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityServiceFactory.java
```diff
@@ -0,0 +1,42 @@
...
+ * <p/>
+ * Returns a new instance of {@link AmbryServerSecurityService} on {@link #getServerSecurityService()} call.
+ */
+public class AmbryServerSecurityServiceFactory implements ServerSecurityServiceFactory {
+
+  private final ServerConfig serverConfig;
+  private final ServerMetrics serverMetrics;
+
+  public AmbryServerSecurityServiceFactory(VerifiableProperties verifiableProperties, ServerMetrics serverMetrics,
+      ClusterMap clusterMap) {
```
zzmao, line 33: Is ClusterMap needed?

ambry-server/src/main/java/com/github/ambry/server/StorageRestRequestService.java
```diff
@@ -31,9 +31,12 @@
 public class StorageRestRequestService implements RestRequestService {
   private static final Logger logger = LoggerFactory.getLogger(StorageRestRequestService.class);
   NettyServerRequestResponseChannel requestResponseChannel;
+  ServerSecurityService serverSecurityService;
 
-  public StorageRestRequestService(NettyServerRequestResponseChannel requestResponseChannel) {
+  public StorageRestRequestService(NettyServerRequestResponseChannel requestResponseChannel,
```
zzmao, line 36: Why pass in serverSecurityService here? 

BTW, this `StorageRestRequestService` will be deprecated in https://github.com/linkedin/ambry/pull/1554

ssen-li: reverted the changes in this file for now as this will be deprecated #1554 and any future validation on individual stream can be done in the new class

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityServiceFactory.java
ssen-li: this is needed in our ambryli repo to get the metricsregistry.

ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
ssen-li: done.

ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
ssen-li: done.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerSecurityServiceFactory.java
cgtz: Why not just pass MetricRegistry into the factory? That is typically what is done for the other factories in the frontend.

ssen-li: I changed the constructor to pass the MetricRegistry directly. Also @cgtz as discussed 
1. I have fixed the intTest failures by recreating the AmbryServer objects (reinitServer/reinitServers) when a server restart was intended in the test workflow. 
2. I could not change the ClusterAgentsFactory interface to include getMetricRegistry() api as few implementations(RecoveryTestClusterAgentsFactory) does not have their own metricRegistry. Also this change is not needed actually since intTest failures are fixed by properly reinitailizing the server objects.

PR #1543, opened 2020-05-29
Fix store unable to seal replicas in second cluster issue

jsjtzyy:
1. Fixed a issue that store didn't seal replica in second cluster because
in-mem replica state is already sealed (the state comes from first cluster)
2. Assigned a dedicated scheduler to participant consistency checker as there
is contention to get thread from regular thread pool(size = 10).

### DIFF
--- a/ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
@@ -39,20 +39,20 @@ public class ParticipantsConsistencyChecker implements Runnable {
 
   @Override
   public void run() {
-    logger.info("Participant consistency checker is initiated. Participants count = {}", participants.size());
+    logger.debug("Participant consistency checker is initiated. Participants count = {}", participants.size());
     try {
       // when code reaches here, it means there are at least two participants.
       ClusterParticipant clusterParticipant = participants.get(0);
       Set<String> sealedReplicas1 = new HashSet<>(clusterParticipant.getSealedReplicas());
       Set<String> stoppedReplicas1 = new HashSet<>(clusterParticipant.getStoppedReplicas());
       for (int i = 1; i < participants.size(); ++i) {
-        logger.info("Checking sealed replica list");
+        logger.debug("Checking sealed replica list");
         Set<String> sealedReplicas2 = new HashSet<>(participants.get(i).getSealedReplicas());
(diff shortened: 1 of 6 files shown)


ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
```diff
@@ -39,22 +39,29 @@ public ParticipantsConsistencyChecker(List<ClusterParticipant> participants, Ser
...
-            stoppedReplicas2);
-        metrics.stoppedReplicasMismatchCount.inc();
+    logger.info("Participant consistency checker is initiated. Participants count = {}", participants.size());
+    try {
+      // when code reaches here, it means there are at least two participants.
+      ClusterParticipant clusterParticipant = participants.get(0);
+      Set<String> sealedReplicas1 = new HashSet<>(clusterParticipant.getSealedReplicas());
+      Set<String> stoppedReplicas1 = new HashSet<>(clusterParticipant.getStoppedReplicas());
+      for (int i = 1; i < participants.size(); ++i) {
+        logger.info("Checking sealed replica list");
```
justinlin-linkedin, line 49: do we need these info logs?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -383,6 +393,26 @@ private void checkCapacityAndUpdateReplicaStatusDelegate() {
           }
         }
       }
+      // During startup, we also need to reconcile the replica state from both ZK clusters.
+      if (!started && replicaStatusDelegates.size() > 1 && thresholdBytesLow < index.getLogUsedCapacity()
```
justinlin-linkedin, line 397: nit: this can be `else if` following the statement above.

ambry-server/src/main/java/com/github/ambry/server/ParticipantsConsistencyChecker.java
jsjtzyy: I added to debug where the exception was thrown when running this checker. Will change it to DEBUG.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
jsjtzyy: True, but I think maybe it's good to separate it from previous code block. This is a one-off code path which is only called during startup. (In the future, this piece can be removed when migration is complete.)

PR #1546, opened 2020-06-02
Core changes of leader based replication. 

Arun-LinkedIn:
It contains following updates:
1. Filter leader and non-leader replicas in replica threads.
2. Limit cross colo fetching of missing keys to leaders only.
3. Store the missing keys information for standby replicas and track them via intra-dc replication.
4. Update the remote token for standby replicas once all the missing keys are obtained.


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -459,8 +483,11 @@ public void replicate() {
             // Skip stores that were stopped during call to getReplicaMetadataResponse
             if (!remoteReplicaInfo.getLocalStore().isStarted()) {
               exchangeMetadataResponseList.add(new ExchangeMetadataResponse(ServerErrorCode.Temporarily_Disabled));
-              remoteReplicaInfo.setExchangeMetadataResponse(
-                  new ExchangeMetadataResponse(ServerErrorCode.Temporarily_Disabled));
+              if (replicatingFromRemoteColo && replicationConfig.replicationModelType.equals(
+                  ReplicationModelType.LEADER_BASED) && replicationEngine instanceof ReplicationManager) {
```
cgtz, line 487: Are these instanceof checks needed? Maybe the ReplicationEngine implementation can check during construction if LEADER_BASED replication is supported and throw an exception if it is misconfigured.

ambry-replication/src/main/java/com/github/ambry/replication/PartitionInfo.java
```diff
@@ -105,6 +106,24 @@ RemoteReplicaInfo removeReplicaInfoIfPresent(ReplicaId remoteReplica) {
...
 
+  /**
+   * Go through remote replicas of this partition and compare the messages written to store with messages
+   * found missing during the previous replication cycle. This is used during leader-based replication where missing
+   * store messages found in metadata exchange are not fetched for non-leader remote replicas in cross colo
+   * data centers. Instead, we wait for them to come from intra-dc replication.
+   * @param messageInfoList list of messages written to local store.
+   */
+  void updateReplicaInfosOnMessageWrite(List<MessageInfo> messageInfoList) {
+    rwLock.readLock().lock();
```
jsjtzyy, line 117: Why this is a `readLock`? I think it's supposed to be `writeLock`.

jsjtzyy: nvm,  `readLock` is probably ok here.

ambry-replication/src/main/java/com/github/ambry/replication/RemoteReplicaInfo.java
```diff
@@ -222,9 +242,104 @@ public boolean equals(Object obj) {
...
+
+  /**
+   * checks if the metadata response for this replica is empty or if there are no missing store keys present in it.
+   * If this replica is a non-leader in remote colo and leader-based replication is enabled, replica threads will not
+   * send next metadata request for it until all missing keys in this metadata response come via intra-dc replication.
+   * @return true if metadata response is empty (null) or there are no missing store keys in it.
+   */
+  synchronized boolean isExchangeMetadataResponseEmpty() {
+    return exchangeMetadataResponse == null || exchangeMetadataResponse.missingStoreMessages == null
+        || exchangeMetadataResponse.missingStoreMessages.size() == 0;
```
jsjtzyy, line 263: minor: can use `exchangeMetadataResponse.missingStoreMessages.isEmpty()`

ambry-replication/src/main/java/com/github/ambry/replication/PartitionInfo.java
```diff
@@ -105,6 +106,24 @@ RemoteReplicaInfo removeReplicaInfoIfPresent(ReplicaId remoteReplica) {
...
+    try {
+      for(RemoteReplicaInfo remoteReplicaInfo : remoteReplicas){
```
jsjtzyy, line 119: nit: format this file

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -503,4 +507,23 @@ protected void stopPartitionReplication(PartitionId partitionId) {
...
+   * missing messages in metadata exchange, track them through intra-dc replication and update token when all the
+   * missing messages are written to store.
+   * @param partitionId partition ID of the messages written to store
+   * @param messageInfoList list of messages written to store
+   */
+  void onMessageWriteForPartition(PartitionId partitionId, List<MessageInfo> messageInfoList) {
+    rwLock.readLock().lock();
+    try {
+      PartitionInfo partitionInfo = partitionToPartitionInfo.get(partitionId);
+      partitionInfo.updateReplicaInfosOnMessageWrite(messageInfoList);
```
jsjtzyy, line 524: You need to check if `partitionInfo == null` here,  it's possible partition is being concurrently removed from this node

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -122,15 +123,16 @@ public ReplicaThread(String threadName, FindTokenHelper findTokenHelper, Cluster
...
   }
 
   public ReplicaThread(String threadName, FindTokenHelper findTokenHelper, ClusterMap clusterMap,
       AtomicInteger correlationIdGenerator, DataNodeId dataNodeId, ConnectionPool connectionPool,
       ReplicationConfig replicationConfig, ReplicationMetrics replicationMetrics, NotificationSystem notification,
       StoreKeyConverter storeKeyConverter, Transformer transformer, MetricRegistry metricRegistry,
       boolean replicatingOverSsl, String datacenterName, ResponseHandler responseHandler, Time time,
-      ReplicaSyncUpManager replicaSyncUpManager, PartitionLeaderInfo partitionLeaderInfo) {
+      ReplicaSyncUpManager replicaSyncUpManager, PartitionLeaderInfo partitionLeaderInfo,
+      ReplicationEngine replicationEngine) {
```
jsjtzyy, line 135: We probably don't need to pass in `ReplicationEngine`, instead we can pass in something like `leaderBasedReplicationTracker` or `leaderBasedReplicationCoordinator` or `leaderBasedReplicationAdmin` that helps to coordinate/update info across replication threads. The component is instantiated only when model == `ReplicationModelType.LEADER_BASED`. Thus, we can check if this component is null and don't have to check many other things like line350~351

jsjtzyy: If we decide to pass in a component rather than `ReplicationEngine`, we can even move the method `onMessageWriteForPartition()` from `ReplicationEngine` to `ReplicationManager` (for better isolation). The component should be simple and lightweight, which is an inner class like `ClusterMapChangeListenerImpl` and `PartitionStateChangeListenerImpl`, what do you think?

jsjtzyy: cc @cgtz  because you have a similar comment. Feel free to offer your insights here.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -451,4 +487,197 @@ public void onPartitionBecomeDroppedFromOffline(String partitionName) {
...
+     * @return true if remote replica is a leader for a partition (Pre-requisite: the partition itself should be a leader locally).
+     */
+    public boolean isPeerReplicaLeaderForPartition(String partitionName, ReplicaId replicaId) {
+      return partitionLeaderInfo.isPeerReplicaLeaderForPartition(partitionName, replicaId);
+    }
+
+    /**
+     * Maintains the list of leader partitions on local node and their corresponding peer leaders in remote data centers
+     */
+    class PartitionLeaderInfo {
```
cgtz, line 565: Could PartitionLeaderInfo be compressed into LeaderBasedReplicationAdmin unless there is a future use case for PLI to be used separately from LBRA? It seems like most methods of LBRA just call the similarly named method of PLI.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -451,4 +487,197 @@ public void onPartitionBecomeDroppedFromOffline(String partitionName) {
...
+        // Read-write lock avoids contention from threads removing old leader partitions (removePartition()) and
+        // threads updating existing leader partitions (refreshPeerLeadersForAllPartitions())
+        rwLockForLeaderReplicaUpdates.writeLock().lock();
+        try {
+          // 2. Get the peer leader replicas from all data centers for this partition
+          List<? extends ReplicaId> leaderReplicas =
+              localReplica.getPartitionId().getReplicaIdsByState(ReplicaState.LEADER, null);
+
+          // 3. Collect and log the list of peer leader replicas associated with this partition
+          Set<ReplicaId> peerLeaderReplicas = new HashSet<>();
```
cgtz, line 603: It looks like `refreshPeerLeadersForAllPartitions` has a slightly different way of building the set of peerLeaderReplicas. Could both of these methods use the same approach (whichever one you prefer)?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -210,20 +229,27 @@ public boolean removeReplica(ReplicaId replicaId) {
...
+            storeKeyConverterFactory.getStoreKeyConverter());
+
+        replicationMetrics.addMetricsForRemoteReplicaInfo(remoteReplicaInfo, trackPerPartitionLagInMetric);
+        remoteReplicaInfos.add(remoteReplicaInfo);
+      }
+      if (replicationConfig.replicationTrackPerPartitionLagFromRemote) {
+        replicationMetrics.addLagMetricForPartition(partition);
+      }
+    } catch (Exception e) {
+      logger.error("Encountered exception instantiating RemoteReplicaInfos", e);
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure, you are right, may be it is not ok to end up with partial list of remoteReplicaInfos. Removed try/catch block for handling the exception now. Previously, added this to handle exception from instantiation of StoreKeyConverter object being passed (added in this PR) to RemoteReplicaInfo constructor. Made changes to avoid passing StoreKeyConverter to RemoteReplicaInfo and instead directly store the converted keys in RemoteReplicaInfo.exchageMetadataResponse.remoteKeyToLocalKeyMap.

ambry-replication/src/main/java/com/github/ambry/replication/RemoteReplicaInfo.java
```diff
@@ -208,23 +228,130 @@ public boolean equals(Object obj) {
...
+        && replicaThread != null && storeKeyConverter != null) {
+      try {
+        List<MessageInfo> messagesFoundInStore = new ArrayList<>();
+
+        //collect store keys to convert
+        List<StoreKey> storeKeysToConvert = exchangeMetadataResponse.missingStoreMessages.stream()
+            .map(MessageInfo::getStoreKey)
+            .collect(Collectors.toList());
+        storeKeyConverter.dropCache();
+        Map<StoreKey, StoreKey> remoteKeyToLocalKeyMap = storeKeyConverter.convert(storeKeysToConvert);
```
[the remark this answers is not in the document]
Arun-LinkedIn: Yeah, we don't need to execute this code to convert remote keys (of stored missing message infos) to local keys every time as it is duplicate work. Made changes to store the remoteKeyToLocalKeyMap along with missing message information in the ExchangeMetadataResponse itself.

ambry-replication/src/main/java/com/github/ambry/replication/RemoteReplicaInfo.java
```diff
@@ -208,23 +228,130 @@ public boolean equals(Object obj) {
...
+        // Go through the messages that are now found in store and reconcile delete, ttl-update and un-delete states.
+        // After that, delete them from the missingStoreMessages set and move the token forward if all missing messages
+        // are found.
+        for (MessageInfo messageInfo : messagesFoundInStore) {
+          BlobId localStoreKey = (BlobId) remoteKeyToLocalKeyMap.get(messageInfo.getStoreKey());
+
+          // 1. compare blob metadata of newly written message with the cached remote message info and
+          // reconcile delete, ttl-update and un-delete states.
+          if (localStoreKey != null) {
+            replicaThread.applyUpdatesOnLocalStoreKey(messageInfo, this, localStoreKey);
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure, we can skip applying updates (ttlupdate, delete or undelete) to blob in local store if received message info is a PUT.  But, I am just thinking if is safe to leave as it is to have this operation identical to what we do in processReplicaMetadataResponse() when messages received in metadata response are found in local store.

ambry-replication/src/main/java/com/github/ambry/replication/RemoteReplicaInfo.java
```diff
@@ -208,23 +228,130 @@ public boolean equals(Object obj) {
...
+          }
+
+          // 2. remove found message from the missing set
+          exchangeMetadataResponse.missingStoreMessages.remove(messageInfo);
```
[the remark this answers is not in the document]
Arun-LinkedIn: It seems like in a given metadata response, if there are several messages associated with same key in remote node, we get a consolidated/merged message info reflecting its latest value. For example, if we find both PUT and DELETE for same blob id, we only get DELETE. Similarly, if we find PUT and TTLUPDATE, we get one message info with TTL field updated to PUT message. Please correct me in case I am wrong here.

Even if above is not true and we receive several messages associated with same key, we might call 'applyUpdatesToBlobInLocalStore' twice which I think might still be fine as we are applying ttlupdate/delete/undelete based on latest state of local blob. 

Just wanted to note that I just separated code to apply ttlupdate/delete/undelete that we do currently in processReplicaMetadataResponse() when we find a key in local store during metadata exchange to the method 'applyUpdatesToBlobInLocalStore()' in order to reapply the same logic when we find key later in this case.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
Arun-LinkedIn: Sure, compressed PartitionLeaderInfo into LeaderBasedReplicationAdmin.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
Arun-LinkedIn: Yeah, true. Made changes to keep the approach same in both methods.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1161,6 +1320,19 @@ ReplicationMetrics getReplicationMetrics() {
     }
   }
 
+  /**
+   *
+   */
+  public static class RemoteMessageAndLocalStoreKeyInfo {
```
jsjtzyy, line 1359: Could you explain the purpose of this class and where it is used?

ambry-replication/src/main/java/com/github/ambry/replication/RemoteReplicaInfo.java
```diff
@@ -208,23 +230,126 @@ public boolean equals(Object obj) {
...
+          for (MessageInfo messageInfo : exchangeMetadataResponse.missingStoreMessages) {
+            StoreKey convertedKey = exchangeMetadataResponse.remoteKeyToLocalKeyMap.get(messageInfo.getStoreKey());
+            if (convertedKey != null) {
+              remoteMessageToConvertedKeyNonNull.put(messageInfo, convertedKey);
+            }
+          }
+          Set<StoreKey> convertedMissingStoreKeys =
+              localStore.findMissingKeys(new ArrayList<>(remoteMessageToConvertedKeyNonNull.values()));
+          remoteMessageToConvertedKeyNonNull.forEach((messageInfo, convertedKey) -> {
+            if (!convertedMissingStoreKeys.contains(convertedKey)) {
```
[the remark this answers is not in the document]
jsjtzyy: Ignore this comment, the initial logic is correct.

jsjtzyy:
This PR is super large. @justinlin-linkedin you only need to review production code.

Arun-LinkedIn:
> This is my first round of review. I do have a question before I can move on. If we have two fabrics, and each fabric has three replicas( one leader and two standbys). So the standby from fabric one would only exchange metadata with leader from fabric two and no exchange on data, right? And leader from fabric one would exchange both data and metadata with leader from fabric two?

> Yeah, you are right. Minor correction in your statement is that standby from fabric one will only exchange metadata with "all" other replicas in fabric two (not just leaders). This is same behavior as today (metadata exchange happens in all-to-all fashion), i.e. we don't differentiate between leaders and standby replicas while exchanging metadata. However, the actual data exchange for missing blobs is limited to only leaders. In your example, like you mentioned, leader from fabric one will exchange both data and metadata with leader from fabric two.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1129,6 +1267,200 @@ private void applyDelete(MessageInfo messageInfo, RemoteReplicaInfo remoteReplic
...
+    // only PUT record in its replication cycle (DELETE record will come in next cycle) while leader gets both
+    // PUT and DELETE together in its replication cycle. Due to that, leader doesn't fetch
+    // the deleted blob from remote data center and the blob is never replicated from leader to standby.
+    // As a result, the PUT record in standby's missing blobs set is never emptied.
+
+    // Time out period is configurable via replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds. If
+    // replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds == -1, this safety feature is disabled.
+
+    List<RemoteReplicaInfo> remoteReplicasTimedOut = new ArrayList<>();
+    if (replicatingFromRemoteColo && leaderBasedReplicationAdmin != null
```
justinlin-linkedin: nit: the repilcatingFromRemoteColo and leaderBaseReplicationAdmin check seem unnecessary, for this method only used when they are true.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1129,6 +1282,186 @@ private void applyDelete(MessageInfo messageInfo, RemoteReplicaInfo remoteReplic
...
+
+  /**
+   * Checks if the input remote replica is a standby replica and has any messages from its previous metadata
+   * exchange still missing in local store, i.e. they haven't arrived from leader via intra-dc replication.
+   * @param remoteReplicaInfo remote replica information
+   * @return true if missing messages in previous metadata exchange are not yet received
+   */
+  boolean containsMissingKeysFromPreviousMetadataExchange(RemoteReplicaInfo remoteReplicaInfo) {
+    String partitionName = remoteReplicaInfo.getLocalReplicaId().getPartitionId().toPathString();
+    ReplicaId remoteReplica = remoteReplicaInfo.getReplicaId();
```
jsjtzyy, line 1437: minor:  `String partitionName = remoteReplica.getPartitionId().toPathString();`

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -371,44 +390,96 @@ public void replicate() {
...
-          if (connectedChannel != null) {
-            connectionPool.checkInConnection(connectedChannel);
-          }
-          context.stop();
-          portTypeBasedContext.stop();
         }
+      } catch (Throwable e) {
+        if (checkoutConnectionTimeInMs == -1) {
+          // throwable happened in checkout connection phase
+          checkoutConnectionTimeInMs = SystemTime.getInstance().milliseconds() - startTimeInMs;
```
jsjtzyy, line 472: minor: we probably can replace all `SystemTime.getInstance()` with `time` in this file.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1129,6 +1267,200 @@ private void applyDelete(MessageInfo messageInfo, RemoteReplicaInfo remoteReplic
...
+   * missing keys from remote leader replicas. For non-leader replicas, we will wait the missing keys to come from
+   * intra-dc replication via leader in local data center.
+   * @param remoteReplicaInfos list of all remote replicas
+   * @param exchangeMetadataResponseList list of metadata responses received from the remote replicas
+   * @param leaderReplicaInfosOutput output list of leader replicas. It will populated in this method.
+   * @param exchangeMetadataResponseListForLeaderReplicaInfosOutput output list of metadata responses received for the leader
+   *                                                       replicas. It will be populated in this method.
+   * @throws IllegalArgumentException
+   */
+  void getLeaderReplicaList(List<RemoteReplicaInfo> remoteReplicaInfos,
```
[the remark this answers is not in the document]
jsjtzyy: Actually, I didn't see fetches between local standby <-> standby, local standby <-> leader in `replicate()` method.

ambry-replication/src/main/java/com/github/ambry/replication/RemoteReplicaInfo.java
```diff
@@ -208,25 +224,45 @@ public boolean equals(Object obj) {
...
+   * (based on the store key) with messages that are written to store by other replica threads.
+   * @param messagesWrittenToStore list of messages written to local store
+   */
+  synchronized void updateMissingMessagesInMetadataResponse(List<MessageInfo> messagesWrittenToStore) {
+
+    Set<MessageInfo> missingStoreMessages = exchangeMetadataResponse.getMissingStoreMessages();
+    if (missingStoreMessages != null && !missingStoreMessages.isEmpty()) {
+      Set<StoreKey> keysWrittenToStore =
+          messagesWrittenToStore.stream().map(MessageInfo::getStoreKey).collect(Collectors.toSet());
+      Set<MessageInfo> missingMessagesFoundInStore = new HashSet<>();
```
lightningrob, line 256: Minor: can use  stream().filter().collect()

ambry-replication/src/test/java/com/github/ambry/replication/LeaderBasedReplicationTest.java
```diff
@@ -0,0 +1,1177 @@
+package com.github.ambry.replication;
```
jsjtzyy, line 1: @Arun-LinkedIn you need to add license here to pass the travis test. See other file for reference.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -298,6 +297,22 @@ public void addRemoteReplicaInfo(RemoteReplicaInfo remoteReplicaInfo) {
...
+   *       the local store
+   *  During cross-colo replication, depending on the {@link ReplicationModelType}, the second step to fetch missing
+   *  keys happens in either all-to-all fashion (i.e. fetched from all cross colo replicas) or is limited to only leader
+   *  replica pairs (i.e both local and remote replicas should be leaders of their partition).
+   *  Here is a table listing on what is exchanged between local and remote replicas based on their roles (leader or
+   *  standby) when {@link ReplicationModelType is LEADER_BASED}.
+   *
+   *                   |   Local Leader	    |    Local Standby	  |     Remote Leader	  |  Remote Standby
+   *      --------------------------------------------------------------------------------------------------
+   *       Leader: 	   |        ---         |  metadata and data  |   metadata and data	|   metadata only
```
justinlin-linkedin, line 313: the indentation is wonky

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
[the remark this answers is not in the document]
justinlin-linkedin: The method is changed. It used to have replicatingFromRemoteColo if-else statement here within the method. Now it's moved outside. This is fine. It's less confusing.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1129,6 +1282,186 @@ private void applyDelete(MessageInfo messageInfo, RemoteReplicaInfo remoteReplic
...
+            exchangeMetadataResponse.getReceivedStoreMessagesWithUpdatesPending();
+        Set<MessageInfo> receivedMessagesWithUpdatesCompleted = new HashSet<>();
+
+        // 1. Go through messages whose blobs are received now (via other replicas) and compare blob metadata of
+        // remote message info with local blob in store and reconcile delete, ttl_update and undelete states
+        for (MessageInfo messageInfo : receivedStoreMessagesWithUpdatesPending) {
+          BlobId localStoreKey =
+              (BlobId) exchangeMetadataResponse.remoteKeyToLocalKeyMap.get(messageInfo.getStoreKey());
+          if (localStoreKey != null) {
+            applyUpdatesToBlobInLocalStore(messageInfo, remoteReplicaInfo, localStoreKey);
```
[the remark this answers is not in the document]
justinlin-linkedin: Don't worry about this case, since the `applyTtlUpdate` method already catches the `Already_Updated` exception. We will not see the 'Already_Updated` and `ID_Deleted` exception being promoted here.

ambry-replication/src/main/java/com/github/ambry/replication/RemoteReplicaInfo.java
Arun-LinkedIn: Sure, updated code to use stream().filter().collect()

ambry-replication/src/test/java/com/github/ambry/replication/LeaderBasedReplicationTest.java
Arun-LinkedIn: Thanks for letting me know. Added the license text.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -884,14 +1029,25 @@ private GetResponse getMessagesForMissingKeys(ConnectedChannel connectedChannel,
...
+        if (leaderBasedReplicationAdmin != null && replicatingFromRemoteColo) {
+          // If leader-based replication is enabled, we should ideally send cross colo GET requests only between leaders.
+          // However, if standby replicas time out waiting for their keys to come from leader, we send cross colo GETs
+          // for them. Set 'remoteColoGetRequestForStandby' to true so that we track number of such cross colo GETs for standby replicas.
+          ReplicaId localReplica = replicasToReplicatePerNode.get(0).getLocalReplicaId();
+          ReplicaId remoteReplica = replicasToReplicatePerNode.get(0).getReplicaId();
+          if (!leaderBasedReplicationAdmin.isLeaderPair(localReplica, remoteReplica)) {
+            remoteColoGetRequestForStandby = true;
+          }
+        }
```
jsjtzyy, line 1048: I wonder why not put this logic in the code block starting from line 438 and pass in a boolean value to `fixMissingStoreKeys` and `getMessagesForMissingKeys`.  I see two benefits:
1. you don't have to if `leaderBasedReplicationAdmin = null` again;
2. you can skip some RemoteReplicaInfo associated with local leader replica;

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1135,6 +1306,189 @@ private void applyDelete(MessageInfo messageInfo, RemoteReplicaInfo remoteReplic
...
+    // Time out period is configurable via replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds. If
+    // replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds == -1, this safety feature is disabled.
+
+    List<RemoteReplicaInfo> remoteReplicasTimedOut = new ArrayList<>();
+    if (replicationConfig.replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds != -1) {
+      for (RemoteReplicaInfo remoteReplicaInfo : remoteReplicaInfos) {
+        ExchangeMetadataResponse exchangeMetadataResponse = remoteReplicaInfo.getExchangeMetadataResponse();
+        if (exchangeMetadataResponse.hasMissingStoreMessages()
+            && (time.seconds() - exchangeMetadataResponse.metadataReceivedTimeSec)
+            > replicationConfig.replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds) {
```
jsjtzyy, line 1371: This condition can be improved. Instead of using metadata received timestamp. We can use the last timestamp when at least one messageInfo is removed from missing set. This is more reasonable for the method name "no progress", it may make a slow progress and timed out in middle way.

PR #1547, opened 2020-06-02
Skip blobs from deprecated containers in replication

SophieGuo410:
This PR blocks replication for deprecated containers.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -644,18 +644,6 @@ public class ReplicaThread implements Runnable {
   }
 
   /**
-   * Determines if {@link MessageInfo} container in the given status.
-   * @param messageInfo A message info class that contains basic info about a blob.
-   * @param containerStatus Status of the container.
-   * @return {@code true} if the blob belongs to the given status container, {@code false} otherwise.
-   */
-  private boolean isGivenStatus(MessageInfo messageInfo, Container.ContainerStatus containerStatus) {
-    return accountService.getAccountById(messageInfo.getAccountId())
-        .getContainerById(messageInfo.getContainerId())
-        .getStatus() == containerStatus;
-  }
-
-  /**
    * Determines if {@link MessageInfo} container in the status of DELETED_IN_PROGRESS or INACTIVE.
    * @param messageInfo A message info class that contains basic info about a blob
    * @return {@code true} if the blob associates with the deprecated container, {@code false} otherwise.
@@ -663,8 +651,10 @@ public class ReplicaThread implements Runnable {
    */
   private boolean isDeprecatedContainer(MessageInfo messageInfo) {
     if (accountService != null) {
-      return isGivenStatus(messageInfo, Container.ContainerStatus.DELETE_IN_PROGRESS) || isGivenStatus(messageInfo,
-          Container.ContainerStatus.INACTIVE);
+      Container.ContainerStatus status = accountService.getAccountById(messageInfo.getAccountId())
+          .getContainerById(messageInfo.getContainerId())
+          .getStatus();
+      return status == Container.ContainerStatus.DELETE_IN_PROGRESS || status == Container.ContainerStatus.INACTIVE;
     } else {
       return false;
     }
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -112,7 +112,7 @@ public class ReplicaThread implements Runnable {
   private final Condition pauseCondition = lock.newCondition();
   private final ReplicaSyncUpManager replicaSyncUpManager;
   private final int maxReplicaCountPerRequest;
-  private final Predicate predicate;
+  private final Predicate skipPredicate;
   private volatile boolean allDisabled = false;
   private final PartitionLeaderInfo partitionLeaderInfo;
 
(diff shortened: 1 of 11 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -639,14 +644,41 @@ private ReplicaMetadataResponse getReplicaMetadataResponse(List<RemoteReplicaInf
...
+
+  /**
+   * Determines if {@link MessageInfo} container in the status of DELETED_IN_PROGRESS or INACTIVE.
+   * @param messageInfo A message info class that contains basic info about a blob
+   * @return {@code true} if the blob associates with the deprecated container, {@code false} otherwise.
+   * Deprecated containers status include DELETE_IN_PROGRESS and INACTIVE.
+   */
+  private boolean isDeprecatedContainer(MessageInfo messageInfo) {
+    if (accountService != null) {
+      return isGivenStatus(messageInfo, Container.ContainerStatus.DELETE_IN_PROGRESS) || isGivenStatus(messageInfo,
```
zzmao, line 666: is `isGivenStatus ` needed?
```
Container.ContainerStatus status = accountService.getAccountById(messageInfo.getAccountId())
        .getContainerById(messageInfo.getContainerId())
        .getStatus()
return (status == DELETE_IN_PROGRESS || status == INACTIVE)
```

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -639,14 +644,31 @@ private ReplicaMetadataResponse getReplicaMetadataResponse(List<RemoteReplicaInf
   }
 
   /**
-   * Gets the missing store messages by comparing the messages from the remote node
+   * Determines if {@link MessageInfo} container in the status of DELETED_IN_PROGRESS or INACTIVE.
+   * @param messageInfo A message info class that contains basic info about a blob
+   * @return {@code true} if the blob associates with the deprecated container, {@code false} otherwise.
+   * Deprecated containers status include DELETE_IN_PROGRESS and INACTIVE.
+   */
+  private boolean isDeprecatedContainer(MessageInfo messageInfo) {
```
lightningrob, line 652: While the logic is correct, it seems a bit heavy handed to pass AccountService to ReplicaThread and see all this logic here (which probably also is similar in other places).  I would suggest adding a generic interface that lets the replica thread know if the messageInfo should be included or skipped.  The interface would have one method like:
boolean skipFromReplication(MessageInfo messageInfo);
Then pass this interface to ReplicaThread instead of AccountService.  In the future, we may want to extend the logic of what messages to skip and have that be transparent to replication.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -639,14 +644,31 @@ private ReplicaMetadataResponse getReplicaMetadataResponse(List<RemoteReplicaInf
...
+    if (accountService != null) {
+      Container.ContainerStatus status = accountService.getAccountById(messageInfo.getAccountId())
+          .getContainerById(messageInfo.getContainerId())
+          .getStatus();
+      return status == Container.ContainerStatus.DELETE_IN_PROGRESS || status == Container.ContainerStatus.INACTIVE;
```
cgtz, line 657: Should replication be subject to the same retention time after marked as DELETE_IN_PROGRESS that the compactor is?

jsjtzyy:
nit: Could you change the name to something like `Skip blobs from deprecated containers in replication`? "Block replication" sounds not appropriate.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
SophieGuo410: Once the container marked as DELETE_IN_PROGRESS, the ACLs will be removed and user can't access this container. In the mean time, should we skip the associate blobs in replication directly? Unless for the future we want to undo the container deletion within the retention time, in that situation we will need to involve retention time for replication. What do you think?

cgtz: I guess I was thinking that not replicating the blobs is about the same as compacting them away in that it would be very hard to revert this decision. Without erasing the replica tokens, replication can't go back in time to get the missing blobs unless a new event for the blob appears on a peer like a TTLUPDATE or UNDELETE.

What was the motivation for the retention time within compaction? My understanding was that it was so that the container deletion could be reverted if it was done by mistake. Sorry if that was a bad assumption on my part.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
SophieGuo410: Updated.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
SophieGuo410: Thanks for the suggestion. I use Predicate interface and implement a ReplicationSkipPredicate to check what message need to be skipped.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
SophieGuo410: The retention time what I add previously is due to after ACLs has been removed we need some retention time before we start compaction. But you are right, I need to consider about revert the container deletion as well. So I add the retention time same as what we are doing in compaction, so that after within the retention time, the DELETE_IN_PROGRESS container won't be skipped through replication. And it's easier for revert as you mentioned.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -113,7 +112,7 @@
   private final Condition pauseCondition = lock.newCondition();
   private final ReplicaSyncUpManager replicaSyncUpManager;
   private final int maxReplicaCountPerRequest;
-  private final AccountService accountService;
+  private final Predicate predicate;
```
lightningrob, line 115: Could we declare it as Predicate<MessageInfo> and rename it skipPredicate to make it clear we're using it to filter out?

SophieGuo410: Updated.

ambry-replication/src/test/java/com/github/ambry/replication/ReplicationTest.java
```diff
@@ -1328,6 +1335,94 @@ public void replicaThreadTestConverter() throws Exception {
         idsToBeIgnoredByPartition, storeKeyConverter, expectedIndex, expectedIndex, 5);
   }
 
+  /**
+   * Tests if deprecated containers have been blocked during replication.
+   */
+  @Test
+  public void blockDeprecatedContainerReplicationTest() throws Exception {
```
lightningrob, line 1342: The ReplicationSkipPredicate class has low code coverage because the test only runs the positive case where container is skipped.  Please add negative test cases.

lightningrob: See https://codecov.io/gh/linkedin/ambry/pull/1547/diff?src=pr&el=tree#diff-YW1icnktYXBpL3NyYy9tYWluL2phdmEvY29tL2dpdGh1Yi9hbWJyeS9yZXBsaWNhdGlvbi9SZXBsaWNhdGlvblNraXBQcmVkaWNhdGUuamF2YQ==

PR #1548, opened 2020-06-03
Fix a concurrency bug caused by offset comparison

justinlin-linkedin:
We are seeing in EI, some partition has duplicated delete record in log, but not in index. After investigation, this is caused by an offset comparison, which is fixed in this PR.

Brief explanation:
1. There are two concurrent delete requests, one of which from frontend, the other from replication.
2. Both delete requests will call delete method, and both would check the validity of delete record up to line 553, and both would be valid. (There is no delete in index yet)
3. After that, one of delete would enter the mutual exclusive block starting from link 553, and persist delete record to log and index
4. The second delete would figure the `indexEndOffsetBeforeCheck` is not the same as `currentIndexEndOffset`, assuming a delete record is 100 bytes, and `indexEndOffsetBeforeCheck` is 1000. then after first delete record, `currentIndexEndOffset` should be 1100
5. Second delete call gets the the first delete record, and it's offset should be 1000. so here the comparison should be >=, rather than >.

### DIFF
--- a/ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
+++ b/ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
@@ -49,6 +49,7 @@ import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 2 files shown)


ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
```diff
@@ -678,6 +731,59 @@ public void concurrentUndeleteTest() throws Exception {
...
+   * record should be persisted.
+   * @throws Exception
+   */
+  @Test
+  public void concurrentUndeleteTestOnSameBlob() throws Exception {
+    MockId idToUndelete = null;
+    for (MockId id : deletedKeys) {
+      idToUndelete = id;
+      break;
+    }
```
jsjtzyy, line 745: same here

PR #1549, opened 2020-06-04
Optimizations for BlobStoreStats

cgtz:
There seem to be a couple hot spots for stats collection. The main one
is querying the full index for each put index entry in
updateExpiryTimeForAllPuts. I do not have a way of reducing the number
of index lookups here yet, but I am trying to find a couple places where
we can improve things along the way:

In this commit, I get rid of one intermediate list of valid index
entries and instead directly update the map instead of iterating an
extra time and allocating an extra ~40k item list.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -427,17 +427,15 @@ class BlobStoreStats implements StoreStats, Closeable {
             StoreErrorCodes.Store_Shutting_Down);
       }
       long indexSegmentStartProcessTimeMs = time.milliseconds();
-      List<IndexEntry> validIndexEntries =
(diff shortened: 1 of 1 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -1130,4 +1116,16 @@ void cancel() {
       cancelled = true;
     }
   }
+
+  /**
+   * An action to take on a single {@link IndexEntry}
+   */
+  @FunctionalInterface
+  private interface IndexEntryAction {
```
cgtz, line 1124: I added a new interface instead of using `Consumer<IndexEntry>` or `ThrowingConsumer<IndexEntry>` since these some of these methods need to throw `StoreException` specifically unless many method signatures are changed

PR #1553, opened 2020-06-05
Checkpoint Azure compaction progress

lightningrob:
Add configurable limit to number of blobs purged in a partition each compaction round.


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -257,17 +309,61 @@ int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageExce
...
    * @param partitionPath the partition to check.
-   * @param fieldName the field name to use (expiration or deletion time).
-   * @return the {@link CloudBlobMetadata} for the dead blob, or NULL if none was found.
-   * @throws CloudStorageException
+   * @return a {@link Map} containing the progress time for compaction based on deletion and expiration time.
+   * @throws BlobStorageException if the checkpoint blob
+   */
+  Map<String, Long> getCompactionProgress(String partitionPath) throws BlobStorageException {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream(64);
+    boolean hasCheckpoint = azureBlobDataAccessor.downloadFile(CHECKPOINT_CONTAINER, partitionPath, baos, false);
```
ankagrawal, line 319: retries here for retry-able error?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -257,17 +309,61 @@ int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageExce
...
+   * Get the current compaction progress for a partition.
    * @param partitionPath the partition to check.
-   * @param fieldName the field name to use (expiration or deletion time).
-   * @return the {@link CloudBlobMetadata} for the dead blob, or NULL if none was found.
-   * @throws CloudStorageException
+   * @return a {@link Map} containing the progress time for compaction based on deletion and expiration time.
+   * @throws BlobStorageException if the checkpoint blob
+   */
+  Map<String, Long> getCompactionProgress(String partitionPath) throws BlobStorageException {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream(64);
```
ankagrawal, line 318: why 64? Also maybe we can make it private static final

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -257,17 +309,61 @@ int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageExce
...
-  public CloudBlobMetadata getOldestDeadlob(String partitionPath, String fieldName) throws CloudStorageException {
-    // TODO: once we have Cosmos compaction table, can query that.
-    List<CloudBlobMetadata> deadBlobs =
-        requestAgent.doWithRetries(() -> getDeadBlobs(partitionPath, fieldName, 1, System.currentTimeMillis(), 1),
-            "GetDeadBlobs", partitionPath);
-    return deadBlobs.isEmpty() ? null : deadBlobs.get(0);
+  boolean updateCompactionProgress(String partitionPath, String fieldName, long checkpointTime) {
+    try {
+      // load existing progress checkpoint.
+      Map<String, Long> checkpoints = getCompactionProgress(partitionPath);
```
ankagrawal, line 348: We should cache this information instead of doing a get always. Only when we dont see a cache for this partition, should we do `getCompactionInProgress`

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -257,17 +309,61 @@ int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageExce
...
+      // Ensure we don't downgrade progress already recorded.
+      if (checkpointTime <= checkpoints.getOrDefault(fieldName, DEFAULT_TIME)) {
+        logger.info("Skipping update of compaction progress for {} because saved {} is more recent.", partitionPath,
+            fieldName);
+        return false;
+      }
+      checkpoints.put(fieldName, Math.max(checkpointTime, checkpoints.get(fieldName)));
+      String json = objectMapper.writeValueAsString(checkpoints);
+      ByteArrayInputStream bais = new ByteArrayInputStream(json.getBytes());
+      azureBlobDataAccessor.uploadFile(CHECKPOINT_CONTAINER, partitionPath, bais);
```
ankagrawal, line 358: Also maybe retry for retry-able error?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -81,29 +103,39 @@ boolean isShuttingDown() {
...
+          compactPartition(partitionPath, CloudBlobMetadata.FIELD_DELETION_TIME, deletionStartTime, queryEndTime,
               timeToQuit);
       logger.info("Purged {} deleted blobs in partition {} up to {}", numPurged, partitionPath, queryEndDate);
       totalBlobsPurged += numPurged;
-      numPurged = compactPartition(partitionPath, CloudBlobMetadata.FIELD_EXPIRATION_TIME, queryStartTime, queryEndTime,
-          timeToQuit);
+      long expirationStartTime = Math.max(queryStartTime, checkpoints.get(CloudBlobMetadata.FIELD_EXPIRATION_TIME));
+      numPurged =
+          compactPartition(partitionPath, CloudBlobMetadata.FIELD_EXPIRATION_TIME, expirationStartTime, queryEndTime,
+              timeToQuit);
```
ankagrawal, line 138: We don't need to compute `compactionTime` at the end of this method anymore.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -81,29 +103,39 @@ boolean isShuttingDown() {
    * Purge the inactive blobs in the specified partition.
    * @return the total number of blobs purged.
    */
-  public int compactPartition(String partitionPath) {
+  public int compactPartition(String partitionPath) throws CloudStorageException {
```
ankagrawal, line 106: javadocs need update.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -56,11 +74,15 @@ public AzureStorageCompactor(AzureBlobDataAccessor azureBlobDataAccessor, Cosmos
     this.azureMetrics = azureMetrics;
     this.retentionPeriodMs = TimeUnit.DAYS.toMillis(cloudConfig.cloudDeletedBlobRetentionDays);
     this.queryLimit = cloudConfig.cloudBlobCompactionQueryLimit;
+    this.purgeLimit = cloudConfig.cloudCompactionPurgeLimit;
```
ankagrawal, line 77: javadocs of the constructor needs update.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -81,29 +103,39 @@ boolean isShuttingDown() {
...
+    try {
+      checkpoints = getCompactionProgress(partitionPath);
+    } catch (BlobStorageException | UncheckedIOException e) {
+      // If checkpoint couldn't be read, skip and try later.
+      throw new CloudStorageException("Compaction of " + partitionPath + " failed reading checkpoint", e);
+    }
+
     long now = System.currentTimeMillis();
     long compactionStartTime = now;
+    // FIXME: incorrect because time limit is for all partitions
```
ankagrawal, line 122: the `compactionStartTime` variable above is redundant now.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -237,6 +237,22 @@ public boolean downloadFile(String containerName, String fileName, OutputStream
...
   }
 
+  /**
+   * Delete a file from blob storage, if it exists.
+   * @param containerName name of the container containing file to delete.
+   * @param fileName name of the file to delete.
+   * @return true if the file was deleted, otherwise false.
+   * @throws BlobStorageException for any error on ABS side.
+   */
+  boolean deleteFile(String containerName, String fileName) throws BlobStorageException {
```
ssen-li, line 247: Not seeing any caller for this API except UT in the current diff. Will this be used in future PR?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
lightningrob: approximate size of payload, made it static.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
lightningrob: I'm hesitant to cache it since the partition could get reassigned back and forth between runs and be out of date.   I'd probably want to read it anyway to make sure.

lightningrob:
> Have a question about the usage of getCompactionProgress() and updateCompactionProgress() api. The get API gets the progress time for both deletion and expiration time for a partition whereas update api updates them seperately. Will it be any better to update both the checkpoint together or split the checkpointing into seperate container for delete and expire for a partition? This is just a thought, no need to make any change if it does not make sense or improves anything.

At the beginning of compaction (for a partition), we get the latest checkpoint times and use them to compact by deletionTime followed by expirationTime.  UpdateCheckpoint is actually called multiple times, after every time bucket is complete (and as part of updating we also fetch the checkpoint again just to be safe.  The code is probably more paranoid than it really needs to be, but the cost of reading/updating one tiny blob is fairly negligible compared to the mass deletions it performs in between.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
lightningrob: The integration test uses it to wipe out the checkpoint during cleanup.  It could be used, I could actually wrap deleteBlob around it in a future PR.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -56,11 +73,13 @@ public AzureStorageCompactor(AzureBlobDataAccessor azureBlobDataAccessor, Cosmos
     this.azureMetrics = azureMetrics;
     this.retentionPeriodMs = TimeUnit.DAYS.toMillis(cloudConfig.cloudDeletedBlobRetentionDays);
     this.queryLimit = cloudConfig.cloudBlobCompactionQueryLimit;
+    this.purgeLimit = cloudConfig.cloudCompactionPurgeLimit;
     this.queryBucketDays = cloudConfig.cloudCompactionQueryBucketDays;
     this.lookbackDays = cloudConfig.cloudCompactionLookbackDays;
-    // TODO: change this
-    compactionTimeLimitMs = TimeUnit.HOURS.toMillis(cloudConfig.cloudBlobCompactionIntervalHours);
     requestAgent = new CloudRequestAgent(cloudConfig, vcrMetrics);
+    emptyCheckpoints = new HashMap<>();
```
cgtz, line 80: since references to this map get returned via different method calls, consider wrapping it in `Collections.unmodifiableMap` It also seems like it could be static.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -257,17 +289,67 @@ int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageExce
...
-   * @param fieldName the field name to use (expiration or deletion time).
-   * @return the {@link CloudBlobMetadata} for the dead blob, or NULL if none was found.
-   * @throws CloudStorageException
+   * @return a {@link Map} containing the progress time for compaction based on deletion and expiration time.
+   * @throws CloudStorageException if the operation fails.
+   */
+  Map<String, Long> getCompactionProgress(String partitionPath) throws CloudStorageException {
+    ByteArrayOutputStream baos = new ByteArrayOutputStream(CHECKPOINT_BUFFER_SIZE);
+    boolean hasCheckpoint = requestAgent.doWithRetries(
+        () -> azureBlobDataAccessor.downloadFile(CHECKPOINT_CONTAINER, partitionPath, baos, false),
```
cgtz, line 300: Using the same ByteArrayOutputStream for all retries could cause issues if the first try partially wrote to BAOS and then the second try will just start appending to the same stream.
Instead you could do something like this:
```java
    ByteArrayOutputStream baos = requestAgent.doWithRetries(() -> {
      ByteArrayOutputStream os = new ByteArrayOutputStream(CHECKPOINT_BUFFER_SIZE);
      boolean hasCheckpoint = azureBlobDataAccessor.downloadFile(CHECKPOINT_CONTAINER, partitionPath, os, false);
      return hasCheckpoint ? os : null;
    }, "Download compaction checkpoint", partitionPath);
```

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -257,17 +289,67 @@ int purgeBlobs(List<CloudBlobMetadata> blobMetadataList) throws CloudStorageExce
...
+    if (!hasCheckpoint) {
+      return emptyCheckpoints;
+    }
+    try {
+      // Payload format: {"expirationTime" : 12345, "deletionTime" : 67890}
+      ObjectNode jsonNode = (ObjectNode) objectMapper.readTree(baos.toString());
+      Map<String, Long> checkpoints = new HashMap<>();
+      for (String fieldName : new String[]{CloudBlobMetadata.FIELD_DELETION_TIME,
+          CloudBlobMetadata.FIELD_EXPIRATION_TIME}) {
+        checkpoints.put(fieldName, jsonNode.has(fieldName) ? jsonNode.get(fieldName).longValue() : DEFAULT_TIME);
```
cgtz, line 311: For me, it seems cleaner for checkpoints to be a simple object (instead of a map) with getDeletionTimeCheckpoint and getFieldExpirationTime and toBytes method so the logic around serde/defaults can all be in one place.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -238,6 +238,22 @@ public boolean downloadFile(String containerName, String fileName, OutputStream
...
+    BlockBlobClient blobClient = getBlockBlobClient(containerName, fileName, false);
+    if (blobClient.exists()) {
+      blobClient.delete();
```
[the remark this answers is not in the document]
lightningrob: Fair point.  I meant it as deleteIfNeeded.  I can change it in a follow up.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -221,7 +221,7 @@ public short undeleteBlob(BlobId blobId, short lifeVersion, CloudUpdateValidator
     for (List<BlobId> batchOfBlobs : chunkedBlobIdList) {
       metadataList.addAll(getBlobMetadataChunked(batchOfBlobs));
     }
-    return metadataList.stream().collect(Collectors.toMap(CloudBlobMetadata::getId, Function.identity()));
+    return metadataList.stream().collect(Collectors.toMap(CloudBlobMetadata::getId, Function.identity(), (x, y) -> x));
```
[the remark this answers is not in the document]
lightningrob: Yes, I added that for easier test handling, it shouldn't happen in prod flow.  It's not the prettiest but don't think it hurts.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
lightningrob: ok

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
lightningrob: Good idea.  I'll change it.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
lightningrob: Good catch.  Moved the outputstream inside the lambda.

PR #1555, opened 2020-06-08
Introduce PropertyStoreToDataNodeConfigAdapter

cgtz:
Introduce a DataNodeConfigSource implementation that can be used to
listen for changes to configs using HelixPropertyStore.

To avoid blocking the threads used by helix to send property store
change events, this implementation uses a background thread to read
records from the property store and notify any downstream listeners.

To prepare for future participant changes this also adds set and get
methods to DataNodeConfigSource for reading and updating
DataNodeConfigs.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/DataNodeConfig.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/DataNodeConfig.java
@@ -18,6 +18,7 @@ package com.github.ambry.clustermap;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 
@@ -112,6 +113,7 @@ class DataNodeConfig {
   /**
    * @return the xid for this server. After {@link SimpleClusterChangeHandler} is retired, this field will be removed.
    */
+  @Deprecated
   long getXid() {
     return xid;
   }
@@ -152,6 +154,27 @@ class DataNodeConfig {
         + stoppedReplicas + ", disabledReplicas=" + disabledReplicas + ", diskConfigs=" + diskConfigs + '}';
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    DataNodeConfig that = (DataNodeConfig) o;
+    return port == that.port && xid == that.xid && Objects.equals(instanceName, that.instanceName) && Objects.equals(
+        hostName, that.hostName) && Objects.equals(datacenterName, that.datacenterName) && Objects.equals(sslPort,
+        that.sslPort) && Objects.equals(http2Port, that.http2Port) && Objects.equals(rackId, that.rackId)
+        && sealedReplicas.equals(that.sealedReplicas) && stoppedReplicas.equals(that.stoppedReplicas)
+        && disabledReplicas.equals(that.disabledReplicas) && diskConfigs.equals(that.diskConfigs);
+  }
+
+  @Override
+  public int hashCode() {
+    return instanceName.hashCode();
+  }
+
   /**
    * Configuration scoped to a single disk on a server.
    */
@@ -184,7 +207,7 @@ class DataNodeConfig {
     }
 
     /**
-     * @return a map from partition name to {@link ReplicaConfig} for all the replicas on the server.
+     * @return a map from partition id to {@link ReplicaConfig} for all the replicas on the server.
      *         This map is mutable.
      */
     Map<String, ReplicaConfig> getReplicaConfigs() {
@@ -195,6 +218,19 @@ class DataNodeConfig {
     public String toString() {
       return "DiskConfig{" + "state=" + state + ", diskCapacity=" + diskCapacityInBytes + ", replicaConfigs="
           + replicaConfigs + '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      DiskConfig that = (DiskConfig) o;
+      return diskCapacityInBytes == that.diskCapacityInBytes && state == that.state && Objects.equals(replicaConfigs,
+          that.replicaConfigs);
     }
   }
 
@@ -232,6 +268,19 @@ class DataNodeConfig {
     public String toString() {
       return "ReplicaConfig{" + "replicaCapacity=" + replicaCapacityInBytes + ", partitionClass='" + partitionClass
           + '\'' + '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+      if (this == o) {
+        return true;
+      }
+      if (o == null || getClass() != o.getClass()) {
+        return false;
+      }
+      ReplicaConfig that = (ReplicaConfig) o;
+      return replicaCapacityInBytes == that.replicaCapacityInBytes && Objects.equals(partitionClass,
+          that.partitionClass);
     }
   }
 }
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/DataNodeConfig.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/DataNodeConfig.java
@@ -163,9 +163,9 @@ class DataNodeConfig {
       return false;
     }
     DataNodeConfig that = (DataNodeConfig) o;
-    return port == that.port && xid == that.xid && Objects.equals(instanceName, that.instanceName) && Objects.equals(
-        hostName, that.hostName) && Objects.equals(datacenterName, that.datacenterName) && Objects.equals(sslPort,
-        that.sslPort) && Objects.equals(http2Port, that.http2Port) && Objects.equals(rackId, that.rackId)
+    return Objects.equals(instanceName, that.instanceName) && Objects.equals(hostName, that.hostName) && Objects.equals(
+        datacenterName, that.datacenterName) && port == that.port && Objects.equals(sslPort, that.sslPort)
+        && Objects.equals(http2Port, that.http2Port) && Objects.equals(rackId, that.rackId) && xid == that.xid
         && sealedReplicas.equals(that.sealedReplicas) && stoppedReplicas.equals(that.stoppedReplicas)
         && disabledReplicas.equals(that.disabledReplicas) && diskConfigs.equals(that.diskConfigs);
   }
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
@@ -16,7 +16,12 @@
 package com.github.ambry.clustermap;
 
 import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.utils.Utils;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
 import java.util.Objects;
+import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixManager;
 import org.apache.helix.api.listeners.InstanceConfigChangeListener;
 import org.apache.helix.model.InstanceConfig;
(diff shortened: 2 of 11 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
```diff
@@ -33,82 +40,146 @@
...
+          // mount path with no special prefix). There may be extra fields when Helix controller adds partitions in ERROR
+          // state to InstanceConfig.
+          LOGGER.warn("{} field does not contain disk info on {}. Skip it and continue on next one.", mountPath,
+              instanceConfig.getInstanceName());
+        } else {
+          DataNodeConfig.DiskConfig disk = new DataNodeConfig.DiskConfig(
+              diskProps.get(DISK_STATE).equals(AVAILABLE_STR) ? HardwareState.AVAILABLE : HardwareState.UNAVAILABLE,
+              Long.parseLong(diskProps.get(DISK_CAPACITY_STR)));
+          String replicasStr = diskProps.get(REPLICAS_STR);
+          if (!replicasStr.isEmpty()) {
```
SophieGuo410, line 122: Do we need to check if replicasStr is null in this situation?

cgtz: That is a good idea. Updated to do that and added more unit tests.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
```diff
@@ -0,0 +1,276 @@
...
+      }
+
+      DataNodeConfig dataNodeConfig = new DataNodeConfig(record.getId(), record.getSimpleField(HOSTNAME_FIELD),
+          record.getIntField(PORT_FIELD, DataNodeId.UNKNOWN_PORT), dcName, getSslPortStr(record),
+          getHttp2PortStr(record), getRackId(record), DEFAULT_XID);
+      dataNodeConfig.getSealedReplicas().addAll(getSealedReplicas(record));
+      dataNodeConfig.getStoppedReplicas().addAll(getStoppedReplicas(record));
+      dataNodeConfig.getDisabledReplicas().addAll(getDisabledReplicas(record));
+      record.getMapFields().forEach((key, diskProps) -> {
+        if (key.startsWith(MOUNT_PREFIX)) {
```
jsjtzyy, line 215: Looks like we will change the format of mount path in PropertyStore record?  Unlike InstanceConfig (which is a pure mount path), here we add a prefix `mount-` ahead of each mount path. I just wonder why not use the same in `InstanceConfigToDataNodeConfigAdapter` to determine if this is a valid mount path.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
```diff
@@ -0,0 +1,276 @@
...
+        }
+      } catch (Exception e) {
+        LOGGER.error("Exception during DataNodeConfig change", e);
+      } finally {
+        lock.unlock();
+      }
+    }
+
+    /**
+     * If initialization has not yet been completed succesfully, attempt an initialization. This should be called from
```
jsjtzyy, line 154: typo: successfully

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
```diff
@@ -0,0 +1,276 @@
...
+      lock.lock();
+      try {
+        if (initialized) {
+          return false;
+        }
+        List<ZNRecord> records = propertyStore.getChildren(CONFIG_PATH, null, AccessOption.PERSISTENT);
+        listener.onDataNodeConfigChange(lazyIterable(records));
+        initialized = true;
+      } catch (Exception e) {
+        LOGGER.error("Exception during DataNodeConfig initialization", e);
```
jsjtzyy, line 168: Just curious, if exception occurs, how can we block the startup of current node?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
```diff
@@ -33,82 +38,146 @@
...
-        dataNodeConfig.getDiskConfigs().put(mountPath, disk);
+      });
+      return dataNodeConfig;
+    }
+
+    /**
+     * @param dataNodeConfig the {@link DataNodeConfig} to convert into an {@link InstanceConfig}.
+     * @return the {@link InstanceConfig}.
+     */
+    InstanceConfig convert(DataNodeConfig dataNodeConfig) {
```
SophieGuo410, line 142: A quick check, do we have any test cases cover this method?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
cgtz: The way that we had it would work correctly too, but I felt that was a little bit of a hacky approach since it checked the content inside of the map field. My thought around adding a prefix was that it would give us more flexibility to add other types of map fields in the future. 

Since there are a few other changes to the schema, like the removal of datacenter name and xid, I thought this was a good time to make changes. However, if you prefer the shorter name, I could go back to that too.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
cgtz: If you would like, I can make unit tests specifically for the converter. Currently they are tested using the DataNodeConfigSource.set and get methods, which call these functions.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
cgtz: Hmm, that is a good point. So, with InstanceConfig listeners, does the exception thrown in the addListener call bubble up to DatacenterInitializer? If so, I will have to address this. Also, I realize that this method does not currently block until initialization is finished. I may have to make changes to support this.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
SophieGuo410: As long as we have test call these functions, I'm fine with it. Thanks for clarify.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
cgtz: Changed the add listener call to wait for initialization to finish in the background thread to match how the other DataNodeConfigSource works. The code will now rethrow an exception thrown by the listener during initialization.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
jsjtzyy: Make sense, let's keep the prefix there.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
```diff
@@ -0,0 +1,278 @@
...
+    private Iterable<DataNodeConfig> lazyIterable(Collection<ZNRecord> records) {
+      return () -> records.stream().map(converter::convert).filter(Objects::nonNull).iterator();
+    }
+
+    /**
+     * Either perform a full initialization or notify the listener about an incremental update to a path. This should be
+     * called from the event executor thread.
+     * @param changedPath the path that changed.
+     */
+    private synchronized void updateOrInitialize(String changedPath) {
```
jsjtzyy, line 140: This comment might be out of scope but I wonder if node (frontend/server) lost connection to ZK and got reconnected, is there a mechanism to re-read all datanode configs in PropertyStore (and re-populate in-mem clustermap)?
As far as I know, InstanceConfig current has an async way to read all configs again (not ideal) but in next release of Helix, it's changed to a sync way, which guarantees all configs will be read again after reconnection.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
```diff
@@ -33,82 +38,146 @@
...
-      LOGGER.warn("Unknown InstanceConfig schema version {} in {}. Ignoring.", schemaVersion, instanceConfig);
-      return null;
+  @Override
+  public DataNodeConfig get(String instanceName) {
+    InstanceConfig instanceConfig = getHelixAdmin().getInstanceConfig(clusterName, instanceName);
+    return instanceConfig != null ? converter.convert(instanceConfig) : null;
+  }
+
+  void setHelixAdmin(HelixAdmin helixAdmin) {
+    this.helixAdmin = helixAdmin;
```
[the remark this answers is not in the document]
cgtz: Yes, this will be used in the next PR, so I will keep it. I was originally working on these two PRs on the same branch so they weren't fully isolated.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
```diff
@@ -33,82 +38,146 @@
...
+        StringBuilder replicasStrBuilder = new StringBuilder();
+        diskConfig.getReplicaConfigs()
+            .forEach((partitionName, replicaConfig) -> replicasStrBuilder.append(partitionName)
+                .append(REPLICAS_STR_SEPARATOR)
+                .append(replicaConfig.getReplicaCapacityInBytes())
+                .append(REPLICAS_STR_SEPARATOR)
+                .append(replicaConfig.getPartitionClass())
+                .append(REPLICAS_DELIM_STR));
+        diskProps.put(REPLICAS_STR, replicasStrBuilder.toString());
+        instanceConfig.getRecord().setMapField(mountPath, diskProps);
```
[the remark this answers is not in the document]
cgtz: Are you referring to pr #1584 here?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/DataNodeConfig.java
```diff
@@ -152,6 +154,27 @@ public String toString() {
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    DataNodeConfig that = (DataNodeConfig) o;
+    return port == that.port && xid == that.xid && Objects.equals(instanceName, that.instanceName) && Objects.equals(
```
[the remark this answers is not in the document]
cgtz: good idea

ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java
cgtz: That is a good point, since you could potentially miss updates while zk was disconnected. I will follow up with Helix regarding this. If it is not possible for this to happen with the PropertyStore impl, we could consider using a background polling thread that periodically fetches all configs.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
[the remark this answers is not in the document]
jsjtzyy: Correct, but later I found `setMapField` still has potential race condition as whole InstanceConfig will be overridden either by ambry node or by Helix. I added some await mechanism in that PR to fix it.

PR #1556, opened 2020-06-08
Parallelize cloud compaction

ankagrawal:


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -103,47 +106,80 @@ public int compactPartitions() {
...
+        if (isShuttingDown()) {
+          logger.info("Compaction terminated due to shut down.");
+          break;
+        }
+        if (partitionSnapshot.isEmpty()) {
+          break;
+        }
+      }
+      while (compactionInProgress > 0) {
+        totalBlobsPurged += executorCompletionService.take().get();
```
lightningrob, line 142: Is the idea to wait until all the threads are done before assigning any new work?  That won't be optimal since a long running partition can keep the other threads idle for a while.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -103,47 +106,80 @@ public int compactPartitions() {
...
     }
+    List<PartitionId> partitionSnapshot = new ArrayList<>(partitions);
+    long compactionStartTime = System.currentTimeMillis();
+    long timeToQuit = System.currentTimeMillis() + compactionTimeLimitMs;
+    int compactionInProgress = 0;
+    doneLatch.set(new CountDownLatch(1));
+    int totalBlobsPurged = 0;
+    int compactedPartitionCount = 0;
+    try {
+      while (true) {
```
lightningrob, line 117: This could be: while (!partitionSnapshot.isEmpty())

ankagrawal: yes.. will fix this.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
ankagrawal: No.. the idea is not to wait until all threads are done. The idea to wait for any one of the threads to return, and then assign it new work immediately. `executorCompletionService.take()` retrieves and removes the `Future` representing the next completed task, waiting if none are yet present. Calling `.get()` on the future just returns the result.

So basically first we assign work to all threads. Then wait for any single thread to complete its work in the `executorCompletionService.take().get()` call. And then we loop over to assign work to this thread.

ankagrawal: Note that the final while loop `while (compactionInProgress > 0)` does loop through all the threads. This happens when all the partitions have been assigned, and we have come out of the initial `while(true)` loop. Here we are just making sure that all the threads get completed before ending the compaction run.

ankagrawal: Note that the final while loop `while (compactionInProgress > 0)` does loop through all the threads. This happens when all the partitions have been assigned, and we have come out of the initial `while(true)` loop. Here we are just making sure that all the threads get completed before ending the compaction run.

This loop also takes care of the case where we encounter a compaction shutdown. In that case too, we break out of the `while(true)` loop, and wait for all the threads to complete work, before releasing the done latch....

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -103,47 +106,77 @@ public int compactPartitions() {
...
-      try {
-        totalBlobsPurged += cloudDestination.compactPartition(partitionPath);
-      } catch (CloudStorageException ex) {
-        logger.error("Compaction failed for partition {}", partitionPath, ex);
-        vcrMetrics.compactionFailureCount.inc();
-      }
+    String partitionPath = partition.toPathString();
+    if (!partitions.contains(partition)) {
+      // Looks like partition was reassigned since the loop started, so skip it
+      logger.warn("Skipping compaction of Partition {} as the partition was reassgined", partition);
```
lightningrob, line 170: Typo: reassigned

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -103,47 +106,77 @@ public int compactPartitions() {
...
+      while (!partitionSnapshot.isEmpty()) {
+        while (compactionInProgress < numThreads) {
+          if (partitionSnapshot.isEmpty()) {
+            break;
+          }
+          PartitionId partitionId = partitionSnapshot.remove(0);
+          executorCompletionService.submit(() -> compactPartition(partitionId));
+          compactionInProgress++;
+        }
+        totalBlobsPurged += executorCompletionService.take().get();
```
lightningrob, line 126: Could we let compactPartition throw exception on error, and catch it here in the get() call?  Then we can increment compactedPartitionCount only in success case.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
lightningrob: Okay, I see that now.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
ankagrawal: fixed.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
ankagrawal: done.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
```diff
@@ -103,47 +107,84 @@ public int compactPartitions() {
...
-      }
-      if (isShuttingDown()) {
-        logger.info("Compaction terminated due to shut down.");
-        break;
-      }
+    try {
+      return cloudDestination.compactPartition(partitionPath);
+    } catch (CloudStorageException ex) {
+      logger.error("Compaction failed for partition {}", partitionPath, ex);
+      throw ex;
```
jsjtzyy, line 187: I wonder if exception occurs in this method before we call `executorCompletionService.take().get()`, what will happen? Looks like it may not be captured at line 130, line 147. If it is captured by line 153 `catch (Throwable th)`, will the whole batch compaction be terminated? What about some outstanding threads that are already assigned compaction task?

ankagrawal: If exception would happen in `compactPartition()`, it would be thrown only when `take().get()` is called. It would have been ignored silently if there was no call to `take().get`.

ambry-cloud/src/test/java/com/github/ambry/cloud/CloudStorageCompactorTest.java
```diff
@@ -65,27 +64,23 @@ public void testCompactPartitions() throws Exception {
...
+      partitionMap.put(new MockPartitionId(i, defaultClass), null);
+      when(mockDest.compactPartition(eq(Integer.toString(i)))).thenReturn(pageSize);
+    }
 
+    assertEquals(pageSize * 100, compactor.compactPartitions());
 
-    // remove partition2 from map
-    partitionMap.remove(new MockPartitionId(partition2, defaultClass));
-    assertEquals(pageSize, compactor.compactPartitions());
+    // remove a from map
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudStorageCompactor.java
jsjtzyy: I am ok with it as I don't see any issue regarding ignoring exception silently.

ankagrawal: thanks.. one thing I wanted to note is that this code will actually never ignore the exception silently.

The way `ExecutorCompletionService::take`  works is that if any thread throws exception, it keeps the exception in its state without throwing out. But then when we call `take()` it looks at the state of the thread, and if the thread has returned some value, then it gives out the value with `.get()` call or if the thread has thrown exception, then it throws that exception on the `take()` call.

So in our code, we will never miss any exception, as we make sure to drain all the threads before shutdown.

jsjtzyy: Thanks for clarification. You mentioned "_The way ExecutorCompletionService::take works is that if any thread throws exception, it keeps the exception in its state without throwing out._".
I would like to know if this also applies to `executorCompletionService.submit()` and `executor.execute()` inside `ExecutorCompletionService`. If I understand correctly, once the task is submitted, `executor.execute()` may happen before `ExecutorCompletionService::take` and at that moment, is any occurred exception still held internally (without being thrown out)?  Correct me if I am wrong, thanks.

ankagrawal: Your understanding is right. Once the task is submitted, `executor.execute()` may happen anytime. However, if the user task that runs inside the `execute()` throws an exception, `execute` doesn't immediately throw the exception. It catches all the exception (including `RunTimeException`) and keeps it in its internal state. When `take().get()` is called, the `Future::get` method retrieves the outcome of the user task, figures out if the outcome of the task is an exception, and either throws the exception, or returns the result (outcome).
So until the `get` method is called no outcome (either exception or result) of user task is reported.

For more reference, do take a look at the `ThreadPoolExecutor::runWorker` and `FutureTask::report` methods of java executor library.

jsjtzyy: Thanks for thorough explanation, this clarifies everything. I am asking this because I had some issues with `ScheduledExecutorService` due to didn't catch exception correctly.

PR #1557, opened 2020-06-09
A temporary fix of delete's lifeVersionConflict error

justinlin-linkedin:
This is an ad hoc fix for the lifeVersionConflict error. A real fix will come later where findKey will be changed to actually respect the given FileSpan.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -526,7 +526,9 @@ public class BlobStore implements Store {
           // is inserted to index.
           // index.getCurrentEndOffset should create a virtual snapshot for this delete method. Everything happens after
           // this end offset would be considered as happening outside of the snapshot.
-          throw new StoreException("Cannot delete id " + info.getStoreKey() + " since it is not present in the index.",
+          String reason =
+              value == null ? "it is not present in the index" : "PUT record is out of current offset before check";
+          throw new StoreException("Cannot delete id " + info.getStoreKey() + " becuse " + reason,
               StoreErrorCodes.ID_Not_Found);
         }
         if (!info.getStoreKey().isAccountContainerMatch(value.getAccountId(), value.getContainerId())) {
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -42,6 +42,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.Callable;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -96,8 +97,8 @@ public class BlobStore implements Store {
   protected PersistentIndex index;
 
   // THIS IS ONLY FOR TEST.
-  volatile protected Runnable operationBeforeSynchronizationFunc = null;
-  volatile protected Runnable betweenGetEndOffsetAndFindKeyFunc = null;
+  volatile protected Callable<Void> operationBeforeSynchronization = null;
+  volatile protected Callable<Void> inDeleteBetweenGetEndOffsetAndFindKey = null;
 
   /**
    * States representing the different scenarios that can occur when a set of messages are to be written to the store.
@@ -435,7 +436,7 @@ public class BlobStore implements Store {
       Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
       MessageWriteSetStateInStore state = checkWriteSetStateInStore(messageSetToWrite, null);
       if (state == MessageWriteSetStateInStore.ALL_ABSENT) {
-        maybeCallBeforeSynchronizationFunc();
+        maybeCallBeforeSynchronization();
         synchronized (storeWriteLock) {
           // Validate that log end offset was not changed. If changed, check once again for existing
           // keys in store
@@ -514,7 +515,7 @@ public class BlobStore implements Store {
       List<IndexValue> indexValuesToDelete = new ArrayList<>();
       List<Short> lifeVersions = new ArrayList<>();
       Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
-      maybeCallBetweenGetEndOffsetAndFindKeyFunc();
+      maybeCallInDeleteBetweenGetEndOffsetAndFindKey();
       for (MessageInfo info : infosToDelete) {
         IndexValue value = index.findKey(info.getStoreKey());
         // TODO: Passing a FileSpan the findKey and change the findKey implementation to fully restrict the index
@@ -564,7 +565,7 @@ public class BlobStore implements Store {
           lifeVersions.add(info.getLifeVersion());
         }
       }
-      maybeCallBeforeSynchronizationFunc();
+      maybeCallBeforeSynchronization();
       synchronized (storeWriteLock) {
         Offset currentIndexEndOffset = index.getCurrentEndOffset();
         if (!currentIndexEndOffset.equals(indexEndOffsetBeforeCheck)) {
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -835,6 +845,15 @@ private void maybeCallBeforeSynchronizationFunc() {
     }
   }
 
+  /**
+   * Call {@link #betweenGetEndOffsetAndFindKeyFunc} if it's not null. This is for testing only.
+   */
+  private void maybeCallBetweenGetEndOffsetAndFindKeyFunc() {
+    if (betweenGetEndOffsetAndFindKeyFunc != null) {
+      betweenGetEndOffsetAndFindKeyFunc.run();
```
lightningrob, line 853: For this type of conditional execution, AtomicObject is safer than volatile.  II believe it's technically possible for this to get a NPE since volatile doesn't prevent the variable from being overwritten between checking for null and running.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -512,9 +514,17 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
       List<IndexValue> indexValuesToDelete = new ArrayList<>();
       List<Short> lifeVersions = new ArrayList<>();
       Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
+      maybeCallBetweenGetEndOffsetAndFindKeyFunc();
```
jsjtzyy: Since we have this callable here, can we remove `maybeCallBeforeSynchronizationFunc` in this delete method. I think we can reuse `maybeCallBetweenGetEndOffsetAndFindKeyFunc` to complete other concurrency tests.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -94,8 +94,10 @@
   // TODO remove this once ZK migration is complete
   private AtomicBoolean isSealed = new AtomicBoolean(false);
   protected PersistentIndex index;
+
   // THIS IS ONLY FOR TEST.
-  protected Runnable operationBeforeSynchronizationFunc = null;
+  volatile protected Runnable operationBeforeSynchronizationFunc = null;
+  volatile protected Runnable betweenGetEndOffsetAndFindKeyFunc = null;
```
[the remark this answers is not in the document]
justinlin-linkedin: update the variable's name.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -94,8 +94,10 @@
   // TODO remove this once ZK migration is complete
   private AtomicBoolean isSealed = new AtomicBoolean(false);
   protected PersistentIndex index;
+
   // THIS IS ONLY FOR TEST.
-  protected Runnable operationBeforeSynchronizationFunc = null;
+  volatile protected Runnable operationBeforeSynchronizationFunc = null;
```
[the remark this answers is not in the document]
justinlin-linkedin: change the type to callable.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
```diff
@@ -714,6 +714,62 @@ public Void call() throws Exception {
...
+      getEndOffsetLatch.countDown();
+      try {
+        findKeyLatch.await();
+      } catch (Exception e) {
+      }
+    });
+
+    ExecutorService executorService = Executors.newFixedThreadPool(2);
+    try {
+      Future<Void> deleteFuture = executorService.submit(new Callable<Void>() {
```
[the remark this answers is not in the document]
justinlin-linkedin: done.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
```diff
@@ -714,6 +714,62 @@ public Void call() throws Exception {
...
+      // Now make sure delete is on the way
+      getEndOffsetLatch.await();
+      Future<Void> putFuture = executorService.submit(new Callable<Void>() {
+        @Override
+        public Void call() throws Exception {
+          putOneBlobWithId(id, PUT_RECORD_SIZE, Utils.Infinite_Time);
+          return null;
+        }
+      });
+      putFuture.get();
```
jsjtzyy, line 741: I feel like it's more appropriate to put `getEndOffsetLatch.await()` and `findKeyLatch.countDown()` into the `call()` method that performs `putOneBlobWithId `

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: good catch on the NPE. But I think it's fine to use volatile since we are not doing any CAS operation here. I updated to make sure the NPE doesn't happen.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -512,9 +514,17 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
...
         IndexValue value = index.findKey(info.getStoreKey());
-        if (value == null) {
+        // TODO: Passing a FileSpan the findKey and change the findKey implementation to fully restrict the index
+        // searching space to be limited within the given FileSpan.
+        if (value == null || (value.getOffset().compareTo(indexEndOffsetBeforeCheck) >= 0 && value.isPut())) {
+          // A temporary fix to deal with this particular corner case.
+          // Between calling index.getCurrentEndOffset and findKey method, there is a Put record of the same blob
+          // is inserted to index.
+          // index.getCurrentEndOffset should create a virtual snapshot for this delete method. Everything happens after
+          // this end offset would be considered as happening outside of the snapshot.
```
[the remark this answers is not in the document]
justinlin-linkedin: changed the  exception message to carry more information about the error.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
justinlin-linkedin: good idea.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: I think they are both necessary. They are testing different corner cases. And I certainly want to remove the getEndOffsetAndFindKey callback after in next PR.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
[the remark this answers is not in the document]
cgtz: Do ID_Not_Found errors msgs get logged anywhere in AmbryRequests?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -512,10 +515,20 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
...
       List<Short> lifeVersions = new ArrayList<>();
       Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
+      maybeCallInDeleteBetweenGetEndOffsetAndFindKey();
       for (MessageInfo info : infosToDelete) {
         IndexValue value = index.findKey(info.getStoreKey());
-        if (value == null) {
-          throw new StoreException("Cannot delete id " + info.getStoreKey() + " since it is not present in the index.",
+        // TODO: Passing a FileSpan the findKey and change the findKey implementation to fully restrict the index
+        // searching space to be limited within the given FileSpan.
+        if (value == null || (value.getOffset().compareTo(indexEndOffsetBeforeCheck) >= 0 && value.isPut())) {
```
cgtz, line 523: There could be one other edge case here where both a put and a ttl update are present after `indexEndOffsetBeforeCheck`. In this case a put entry will be returned with the "Ttl_Update_Index" flag set, which will make `value.isPut()` return false (See https://github.com/linkedin/ambry/blob/master/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java#L629)

I suppose this may be even lower probability so we may not need to worry about this in this temporary fix. However, at some point its probably worth looking back on the IndexValue.isPut method (i.e. does a put with ttlupdate flag set by the index logic count as a put record?).

justinlin-linkedin: yeah, this is caused by a slight different rules when reading the indexvalue than write the indexvalue. When writing index value, PUT and TTL_UPDATE are different records. While reading indexvalue, TTL_UPDATE will be merged with other records. It becomes just a flag. 

I suppose we don't have to worry about it at this PR, since next PR, I will use a FileSpan to restrict the search space in findKey.

PR #1558, opened 2020-06-10
FindKey method takes FileSpan as the entire search space.

justinlin-linkedin:
Before this PR, findKey in PersistentIndex, uses FileSpan in a unorthodoxy way. It basically uses FileSpan as a way to locate index segments to search.

For instance, if we have three index segments, and their respective start offset and end offset are [0, 1000), [1000, 2000), [2000, 3000), then when we have a FileSpan whose start and end offset are [500, 2500], findKey would know that the all three index segments should be searched up, since the start offset is located within the first index segment and the end offset is located at the third segment.

The problem the actual record's offset is 2700, which is not covered in the FileSpan, but covered in the third segment, this record will be returned by findKey. In this case, FileSpan doesn't serve as a search space for findKey, but rather an pointer to find the index segments, which then would become search space. 

This PR would fix this issue by changing findKey and strictly respect FileSpan as search space. It also has a few subtle changes.

1. FileSpan's start offset is inclusive, but end offset would be exclusive. 
2. Compaction without undelete, would persist a cutoff offset to indicate if the delete is in effect. The cutoffset  used to be the start offset of the index segment, now it will be the end offset.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -532,18 +532,10 @@ public class BlobStore implements Store {
       Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
       maybeCallInDeleteBetweenGetEndOffsetAndFindKey();
       for (MessageInfo info : infosToDelete) {
-        IndexValue value = index.findKey(info.getStoreKey());
-        // TODO: Passing a FileSpan the findKey and change the findKey implementation to fully restrict the index
-        // searching space to be limited within the given FileSpan.
-        if (value == null || (value.getOffset().compareTo(indexEndOffsetBeforeCheck) >= 0 && value.isPut())) {
-          // A temporary fix to deal with this particular corner case.
-          // Between calling index.getCurrentEndOffset and findKey method, there is a Put record of the same blob
-          // is inserted to index.
-          // index.getCurrentEndOffset should create a virtual snapshot for this delete method. Everything happens after
-          // this end offset would be considered as happening outside of the snapshot.
-          String reason =
-              value == null ? "it is not present in the index" : "PUT record is out of current offset before check";
-          throw new StoreException("Cannot delete id " + info.getStoreKey() + " because " + reason,
+        IndexValue value =
+            index.findKey(info.getStoreKey(), new FileSpan(index.getStartOffset(), indexEndOffsetBeforeCheck));
+        if (value == null) {
+          throw new StoreException("Cannot delete id " + info.getStoreKey() + " because it is not present in the index",
               StoreErrorCodes.ID_Not_Found);
         }
         if (!info.getStoreKey().isAccountContainerMatch(value.getAccountId(), value.getContainerId())) {
@@ -592,8 +584,7 @@ public class BlobStore implements Store {
             IndexValue value = index.findKey(info.getStoreKey(), fileSpan,
                 EnumSet.of(PersistentIndex.IndexEntryType.PUT, PersistentIndex.IndexEntryType.DELETE,
                     PersistentIndex.IndexEntryType.UNDELETE));
-            if (value != null && value.getOffset().compareTo(indexEndOffsetBeforeCheck) >= 0) {
-              // Make sure the value is actually after the indexEndOffsetBeforeCheck
+            if (value != null) {
               if (value.isDelete() && value.getLifeVersion() == lifeVersions.get(i)) {
                 throw new StoreException(
                     "Cannot delete id " + info.getStoreKey() + " since it is already deleted in the index.",
@@ -667,7 +658,8 @@ public class BlobStore implements Store {
           throw new StoreException("BlobStore only supports removing the expiration time",
               StoreErrorCodes.Update_Not_Allowed);
         }
-        IndexValue value = index.findKey(info.getStoreKey());
+        IndexValue value =
+            index.findKey(info.getStoreKey(), new FileSpan(index.getStartOffset(), indexEndOffsetBeforeCheck));
         if (value == null) {
           throw new StoreException("Cannot update TTL of " + info.getStoreKey() + " since it's not in the index",
               StoreErrorCodes.ID_Not_Found);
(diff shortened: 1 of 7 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -587,7 +603,7 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
         logger.trace("Searching for {} in index with filespan ranging from {} to {}", key, fileSpan.getStartOffset(),
             fileSpan.getEndOffset());
         segmentsMapToSearch = indexSegments.subMap(indexSegments.floorKey(fileSpan.getStartOffset()), true,
-            indexSegments.floorKey(fileSpan.getEndOffset()), true).descendingMap();
+            indexSegments.lowerKey(fileSpan.getEndOffset()), true).descendingMap();
```
justinlin-linkedin, line 606: End offset in FileSpan is exclusive so we don't have to include the end offset. Assuming we have several index segments and the map of indexSegments look like this (it's offset to index segment).
{ 
    0       :index segment1
    1000 :index segment2
    2000 :index segment3
    3000 :index segment4
}
When the filespan is [500, 2000), then we should only search on index segment1 and index segment2. index segment 3's start offset is 2000, but 2000 in filespan is exclusive, we don't need it.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -575,6 +588,9 @@ IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryType> type
    */
   private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryType> types,
```
zzmao, line 589: I looked at the callers of the this method, looks like they are still using [start, end] assumption. Any reason not change caller to [start, end+1)?

justinlin-linkedin: I don't think the callers ever care about the inclusiveness of the end offset. For example, in PersistentIndex.getDeletedBlobReadOptions method, we call findKey with a FileSpan, whose start offset is the beginning of the index, and end offset if the offset of the DELETE record. In this case, it obviously doesn't want to include the end offset, the DELETE record.

ambry-store/src/main/java/com/github/ambry/store/FileSpan.java
```diff
@@ -53,7 +54,14 @@ Offset getEndOffset() {
    * @return {@code true} if {@code offset} is in this {@link FileSpan} (start and end offsets are considered inclusive)
    */
   boolean inSpan(Offset offset) {
-    return offset.compareTo(startOffset) >= 0 && offset.compareTo(endOffset) <= 0;
+    return offset.compareTo(startOffset) >= 0 && offset.compareTo(endOffset) < 0;
```
cgtz, line 57: I feel like there shouldn't be issues here, but could you double check the queries that `PersistentIndex.findEntriesSince()` uses to make sure that the exclusive end offset is okay? I think those methods use `getCurrentEndOffset` which uses the end offset of the last index segment. Is IndexSegment.getEndOffset() inclusive or exclusive?

justinlin-linkedin: I have to double check with the code. But I want to point out something here, that end offset of index segment or log segment are kinda different than the end offset in the filespan. 

Say if we have a 10 puts in a log segment and each put is 1000 bytes, then this start offset of this log segment is 18 (with header), and the end offset is 1000 * 10 + 18 = 10018, which is expected.  Of course, this log segment would have index segment(s). Let's say, now we want to search within [18, 10018], both inclusive, we can have a filespan with start offset = 18 and end offset = 10018. Wait, the end offset in filespan is not inclusive, would this be a problem? Actually no. The 10 Puts, their offsets are 18, 1018, 2018, 3018... 9018 respectively. Even if we don't include 10018, we effectively covers all 10 puts here.

As you can see, the end offset of a log segment is greater than the offset of *LAST RECORD* in the segment, so an exclusive end offset in filespan can still cover all the records. (No records would be a zero-byte record)

The is also quite convenient for boundaries. There are two examples I want to share here.
1. Say we have a DELETE index value, and we have to find the corresponding PUT for this key. Then we can call `index.findKey(key, new FileSpan(index.getStartOffset(), deleteValue.getOffset(), EnumSet.of(PersistentIndex.IndexEntryType.PUT))`. The filespan's start offset is the start offset of the index, and the end offset is the DELETE's offset. Since the end offset is not included, so the DELETE record will be excluded from the search.
2. Say in blobstore, we have `indexEndOffsetBeforeCheck` and we want to do a findKey up to that offset. If we take a log segment above as an example, the `indexEndOffsetBeforeCheck` would be 10018 now. If before we call findKey, there is another put being inserted, then this last put's offset would be 10018, and in findKey, it will be excluded.

jsjtzyy:
> The problem the actual record's offset is 2700, which is covered in the FileSpan, but covered in the third segment, this record will be returned by findKey. In this case, FileSpan doesn't serve as a search space for findKey, but rather an pointer to find the index segments, which then would become search space.

minor: `which is not covered in the FileSpan ...`

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -801,8 +803,7 @@ public short undelete(MessageInfo info) throws StoreException {
           FileSpan fileSpan = new FileSpan(indexEndOffsetBeforeCheck, currentIndexEndOffset);
           IndexValue value = index.findKey(info.getStoreKey(), fileSpan,
               EnumSet.of(PersistentIndex.IndexEntryType.DELETE, PersistentIndex.IndexEntryType.UNDELETE));
-          if (value != null && value.getOffset().compareTo(indexEndOffsetBeforeCheck) >= 0) {
-            // Make sure the value is actually after the indexEndOffsetBeforeCheck
+          if (value != null) {
             if (value.isUndelete() && value.getLifeVersion() == revisedLifeVersion) {
```
jsjtzyy, line 807: I need a little clarification of this if-else block. The `if` branch makes sense to me but for the `else` branch, what if the `value.isDelete() == true`,  it seems to throw exception directly.  Actually this might be a valid case. 
(Correct me if I am wrong and let me know anything I missed here)

justinlin-linkedin: It's not a valid delete. If we are here, then the previous record should be DELETE. For example, if the index values for this blob is PUT, DELETEv0. Then this UNDELETE's lifeVersion should be 1. so in `if` statement, we see a UNDELETEv1 out of the previous scope, this UNDELETEv1 might come from replication, this is valid. But if it's not UNDELETEv1, then all the alternatives should be invalid. For instance, UNDELETEv2 is invalid since when the undelete method is invoked, the DELETE is at version 0. DELETEv2 is also invalid since we don't really know the order of DELETEv2 and UNDELETEv1 anymore. Imagine that this host some how freezes for a long time, and within this period, client send a DELETE, UNDELETE and DELETE. So in other host, the final state is DELETE, if we append UNDELETE after this DELETE, then we change the final state. This is not right.

Basically what I am doing here is to make sure that only same UNDELETE from replication can be valid.

PR #1559, opened 2020-06-11
Support Hybrid Compaction Policy

SophieGuo410:
StatsBasedCompactionPolicy which might not include all the log segments that contain data from the target container. This pr support a hybrid compaction policy so the StatsBasedCompactionPolicy one running more frequently, and a full scan compactAllPolicy running less frequently.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -82,6 +82,13 @@ public class StoreConfig {
   public final int storeDeletedMessageRetentionDays;
 
   /**
+   * How often the HybridCompactionPolicy swith from StatsBasedCompactionPolicy to CompactAllPolicy.
+   */
+  @Config("store.compaction.policy.switch.period")
+  @Default("7")
+  public final int storeCompactionPolicySwitchPeriod;
+
+  /**
    * How long (in days) a container must be in DELETE_IN_PROGRESS state before it's been deleted during compaction.
    */
   @Config("store.container.deletion.retention.days")
@@ -347,6 +354,7 @@ public class StoreConfig {
     storeMaxNumberOfEntriesToReturnFromJournal =
         verifiableProperties.getIntInRange("store.max.number.of.entries.to.return.from.journal", 5000, 1, 10000);
     storeDeletedMessageRetentionDays = verifiableProperties.getInt("store.deleted.message.retention.days", 7);
+    storeCompactionPolicySwitchPeriod = verifiableProperties.getInt("store.compaction.policy.switch.period", 7);
     storeContainerDeletionRetentionDays = verifiableProperties.getInt("store.container.deletion.retention.days", 14);
     storeHardDeleteOperationsBytesPerSec =
         verifiableProperties.getIntInRange("store.hard.delete.operations.bytes.per.sec", 100 * 1024, 1,
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -82,7 +82,7 @@ public class StoreConfig {
   public final int storeDeletedMessageRetentionDays;
 
   /**
-   * How often the HybridCompactionPolicy swith from StatsBasedCompactionPolicy to CompactAllPolicy.
+   * How often the HybridCompactionPolicy switch from StatsBasedCompactionPolicy to CompactAllPolicy.
    */
   @Config("store.compaction.policy.switch.period")
   @Default("7")
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -82,11 +82,18 @@ public class StoreConfig {
   public final int storeDeletedMessageRetentionDays;
 
   /**
-   * How often the HybridCompactionPolicy switch from StatsBasedCompactionPolicy to CompactAllPolicy.
+   * How often the HybridCompactionPolicy switch from StatsBasedCompactionPolicy to CompactAllPolicy based on timestamp.
    */
-  @Config("store.compaction.policy.switch.period")
+  @Config("store.compaction.policy.switch.timestamp.days")
   @Default("7")
-  public final int storeCompactionPolicySwitchPeriod;
+  public final int storeCompactionPolicySwitchTimestampDays;
+
+  /**
+   * How often the HybridCompactionPolicy switch from StatsBasedCompactionPolicy to CompactAllPolicy based on counter value.
+   */
+  @Config("store.compaction.policy.switch.counter.days")
+  @Default("7")
+  public final int storeCompactionPolicySwitchCounterDays;
 
   /**
    * How long (in days) a container must be in DELETE_IN_PROGRESS state before it's been deleted during compaction.
@@ -354,7 +361,8 @@ public class StoreConfig {
     storeMaxNumberOfEntriesToReturnFromJournal =
         verifiableProperties.getIntInRange("store.max.number.of.entries.to.return.from.journal", 5000, 1, 10000);
     storeDeletedMessageRetentionDays = verifiableProperties.getInt("store.deleted.message.retention.days", 7);
-    storeCompactionPolicySwitchPeriod = verifiableProperties.getIntInRange("store.compaction.policy.switch.period", 7, 1, 10);
+    storeCompactionPolicySwitchTimestampDays = verifiableProperties.getIntInRange("store.compaction.policy.switch.timestamp.days", 7, 1, 14);
+    storeCompactionPolicySwitchCounterDays = verifiableProperties.getIntInRange("store.compaction.policy.switch.counter.days", 7, 1, 14);
     storeContainerDeletionRetentionDays = verifiableProperties.getInt("store.container.deletion.retention.days", 14);
     storeHardDeleteOperationsBytesPerSec =
         verifiableProperties.getIntInRange("store.hard.delete.operations.bytes.per.sec", 100 * 1024, 1,
@@ -417,4 +425,3 @@ public class StoreConfig {
         verifiableProperties.getBoolean("store.set.local.partition.state.enabled", false);
   }
 }
-
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -354,7 +354,7 @@ public class StoreConfig {
     storeMaxNumberOfEntriesToReturnFromJournal =
         verifiableProperties.getIntInRange("store.max.number.of.entries.to.return.from.journal", 5000, 1, 10000);
     storeDeletedMessageRetentionDays = verifiableProperties.getInt("store.deleted.message.retention.days", 7);
-    storeCompactionPolicySwitchPeriod = verifiableProperties.getInt("store.compaction.policy.switch.period", 7);
+    storeCompactionPolicySwitchPeriod = verifiableProperties.getIntInRange("store.compaction.policy.switch.period", 7, 1, 10);
     storeContainerDeletionRetentionDays = verifiableProperties.getInt("store.container.deletion.retention.days", 14);
     storeHardDeleteOperationsBytesPerSec =
         verifiableProperties.getIntInRange("store.hard.delete.operations.bytes.per.sec", 100 * 1024, 1,
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -1081,13 +1081,6 @@ public class BlobStore implements Store {
     return replicaStatusDelegates;
   }
 
-  /**
-   * @return a {@link ReplicaId} associated with this store
-   */
-  public ReplicaId getReplicaId() {
-    return replicaId;
-  }
-
   @Override
   public void shutdown() throws StoreException {
     shutdown(false);
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -1026,9 +1026,9 @@ public class BlobStore implements Store {
    * {@code null} if there isn't anything to compact
    * @throws StoreException on any issues while reading entries from index
    */
-  CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy, CompactionPolicyCounter compactionPolicyCounter) throws StoreException {
+  CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy) throws StoreException {
     return compactionPolicy.getCompactionDetails(capacityInBytes, index.getLogUsedCapacity(), log.getSegmentCapacity(),
-        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats, compactionPolicyCounter);
+        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats);
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -1028,7 +1028,7 @@ public class BlobStore implements Store {
    */
   CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy) throws StoreException {
     return compactionPolicy.getCompactionDetails(capacityInBytes, index.getLogUsedCapacity(), log.getSegmentCapacity(),
-        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats);
+        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats, dataDir);
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -1026,9 +1026,9 @@ public class BlobStore implements Store {
    * {@code null} if there isn't anything to compact
    * @throws StoreException on any issues while reading entries from index
    */
-  CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy) throws StoreException {
+  CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy, SafeCounterWithoutLock safeCounterWithoutLock) throws StoreException {
     return compactionPolicy.getCompactionDetails(capacityInBytes, index.getLogUsedCapacity(), log.getSegmentCapacity(),
-        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats);
+        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats, safeCounterWithoutLock);
   }
 
   /**
@@ -1079,6 +1079,13 @@ public class BlobStore implements Store {
    */
   public List<ReplicaStatusDelegate> getReplicaStatusDelegates() {
     return replicaStatusDelegates;
+  }
+
+  /**
+   * @return a {@link ReplicaId} associated with this store
+   */
+  public ReplicaId getReplicaId() {
+    return replicaId;
   }
 
   @Override
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -1026,9 +1026,9 @@ public class BlobStore implements Store {
    * {@code null} if there isn't anything to compact
    * @throws StoreException on any issues while reading entries from index
    */
-  CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy, SafeCounterWithoutLock safeCounterWithoutLock) throws StoreException {
+  CompactionDetails getCompactionDetails(CompactionPolicy compactionPolicy, CompactionPolicyCounter compactionPolicyCounter) throws StoreException {
     return compactionPolicy.getCompactionDetails(capacityInBytes, index.getLogUsedCapacity(), log.getSegmentCapacity(),
-        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats, safeCounterWithoutLock);
+        LogSegment.HEADER_SIZE, index.getLogSegmentsNotInJournal(), blobStoreStats, compactionPolicyCounter);
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -210,6 +210,7 @@ class BlobStoreCompactor {
           deprecatedContainers.add(new Pair<>(container.getParentAccountId(), container.getId()));
         }
       });
+      //TODO: Filter out the INACTIVE containers from deprecatedContainers set if it's already been compacted.
       accountService.getContainersByStatus(Container.ContainerStatus.INACTIVE).forEach((container) -> {
         deprecatedContainers.add(new Pair<>(container.getParentAccountId(), container.getId()));
       });
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -1118,10 +1118,17 @@ class BlobStoreStats implements StoreStats, Closeable {
   }
 
   /**
-   * @return the storeId for this blob.
+   * @return the storeId for this {@link BlobStore}.
    */
   String getStoreId() {
     return this.storeId;
+  }
+
+  /**
+   * @return the {@link StoreMetrics} for this {@link BlobStore}.
+   */
+  StoreMetrics getMetrics() {
+    return this.metrics;
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -1118,6 +1118,13 @@ class BlobStoreStats implements StoreStats, Closeable {
   }
 
   /**
+   * @return the storeId for this blob.
+   */
+  String getStoreId() {
+    return this.storeId;
+  }
+
+  /**
    * An action to take on a single {@link IndexEntry}
    */
   @FunctionalInterface
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
@@ -179,14 +179,6 @@ class CompactionManager {
   }
 
   /**
-   * Get the current replicaToCounterMap policy.
-   * @return {@link this.replicaToCounterMap}
-   */
-  Map<ReplicaId, CompactionPolicyCounter> getReplicaToCounterMap(){
-    return this.replicaToCounterMap;
-  }
-
-  /**
    * Get compaction details for a given {@link BlobStore} if any
    * @param blobStore the {@link BlobStore} for which compaction details are requested
    * @return the {@link CompactionDetails} containing the details about log segments that needs to be compacted.
@@ -194,14 +186,7 @@ class CompactionManager {
    * @throws StoreException when {@link BlobStore} is not started
    */
   CompactionDetails getCompactionDetails(BlobStore blobStore) throws StoreException {
-    ReplicaId replicaId = blobStore.getReplicaId();
-    if (compactionPolicyFactory != null && compactionPolicyFactory instanceof HybridCompactionPolicyFactory) {
-      CompactionPolicyCounter compactionPolicyCounter =
-          replicaToCounterMap.getOrDefault(replicaId, new CompactionPolicyCounter(storeConfig));
-      compactionPolicyCounter.increment();
-      replicaToCounterMap.put(replicaId, compactionPolicyCounter);
-    }
-    return blobStore.getCompactionDetails(compactionPolicy, replicaToCounterMap.get(replicaId));
+    return blobStore.getCompactionDetails(compactionPolicy);
   }
 
   /**
@@ -223,6 +208,13 @@ class CompactionManager {
    */
   Set<BlobStore> getAllStores() {
     return stores;
+  }
+
+  /**
+   * @return {@link CompactionPolicy} in compaction manager.
+   */
+  CompactionPolicy getCompactionPolicy() {
+    return compactionPolicy;
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
@@ -196,7 +196,8 @@ class CompactionManager {
   CompactionDetails getCompactionDetails(BlobStore blobStore) throws StoreException {
     ReplicaId replicaId = blobStore.getReplicaId();
     if (compactionPolicyFactory != null && compactionPolicyFactory instanceof HybridCompactionPolicyFactory) {
-      SafeCounterWithoutLock safeCounterWithoutLock = replicaToCounterMap.getOrDefault(replicaId, new SafeCounterWithoutLock(storeConfig));
+      SafeCounterWithoutLock safeCounterWithoutLock =
+          replicaToCounterMap.getOrDefault(replicaId, new SafeCounterWithoutLock(storeConfig));
       safeCounterWithoutLock.incrementAndGet();
       replicaToCounterMap.put(replicaId, safeCounterWithoutLock);
     }
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
@@ -50,7 +50,7 @@ class CompactionManager {
   private final StorageManagerMetrics metrics;
   private final CompactionPolicy compactionPolicy;
   private static final Logger logger = LoggerFactory.getLogger(CompactionManager.class);
-  private final Map<ReplicaId, SafeCounterWithoutLock> replicaToCounterMap;
+  private final Map<ReplicaId, CompactionPolicyCounter> replicaToCounterMap;
   private final CompactionPolicyFactory compactionPolicyFactory;
   private Thread compactionThread;
 
@@ -182,7 +182,7 @@ class CompactionManager {
    * Get the current replicaToCounterMap policy.
    * @return {@link this.replicaToCounterMap}
    */
-  Map<ReplicaId, SafeCounterWithoutLock> getReplicaToCounterMap(){
+  Map<ReplicaId, CompactionPolicyCounter> getReplicaToCounterMap(){
     return this.replicaToCounterMap;
   }
 
@@ -196,10 +196,10 @@ class CompactionManager {
   CompactionDetails getCompactionDetails(BlobStore blobStore) throws StoreException {
     ReplicaId replicaId = blobStore.getReplicaId();
     if (compactionPolicyFactory != null && compactionPolicyFactory instanceof HybridCompactionPolicyFactory) {
-      SafeCounterWithoutLock safeCounterWithoutLock =
-          replicaToCounterMap.getOrDefault(replicaId, new SafeCounterWithoutLock(storeConfig));
-      safeCounterWithoutLock.incrementAndGet();
-      replicaToCounterMap.put(replicaId, safeCounterWithoutLock);
+      CompactionPolicyCounter compactionPolicyCounter =
+          replicaToCounterMap.getOrDefault(replicaId, new CompactionPolicyCounter(storeConfig));
+      compactionPolicyCounter.increment();
+      replicaToCounterMap.put(replicaId, compactionPolicyCounter);
     }
     return blobStore.getCompactionDetails(compactionPolicy, replicaToCounterMap.get(replicaId));
   }
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
@@ -13,15 +13,12 @@
  */
 package com.github.ambry.store;
 
-import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.config.StoreConfig;
 import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.util.Collection;
 import java.util.EnumSet;
-import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.LinkedBlockingDeque;
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
@@ -13,12 +13,15 @@
  */
 package com.github.ambry.store;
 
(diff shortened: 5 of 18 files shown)


ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+/**
+ * HybridCompactionPolicy will run compaction with StatsBasedCompactionPolicy more frequently and with CompactAllPolicy
+ * one time out of storeCompactionPolicySwitchPeriod.
+ */
+public class HybridCompactionPolicy implements CompactionPolicy{
+  private final Time time;
+  private final StoreConfig storeConfig;
+  private final Logger logger = LoggerFactory.getLogger(getClass());
```
justinlin-linkedin, line 31: we can make this static.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,80 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * HybridCompactionPolicy will run compaction with StatsBasedCompactionPolicy more frequently and with CompactAllPolicy
+ * one time out of storeCompactionPolicySwitchPeriod.
+ */
+public class HybridCompactionPolicy implements CompactionPolicy{
```
justinlin-linkedin, line 28: nit: need to format this file.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Selects which compaction policy needs to be used for current compaction cycle.
+   * @return CompactAllPolicy if the round number of compaction reach to storeConfig.storeCompactionPolicySwitchPeriod.
+   * Othewise @return StatsBasedCompactionPolicy.
+   */
+  CompactionPolicy selectCompactionPolicy(){
+    if (counter.incrementAndGet() == 0) {
+      logger.info("Return CompactAllPolicy this round");
```
justinlin-linkedin, line 55: info logs seem to verbose. Can you change it to trace?

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapacity, long segmentCapacity,
+      long segmentHeaderSize, List<String> logSegmentsNotInJournal, BlobStoreStats blobStoreStats)
+      throws StoreException {
+    return selectCompactionPolicy().getCompactionDetails(totalCapacity, usedCapacity, segmentCapacity,
+        segmentHeaderSize, logSegmentsNotInJournal, blobStoreStats);
+  }
+
+  /**
+   * Selects which compaction policy needs to be used for current compaction cycle.
```
justinlin-linkedin, line 49: nit: Select which compaction policy to use ...

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
justinlin-linkedin, line 28: I don't feel good about the way we count the compaction round. All blobStore(partition) uses the same CompactionPolicy instance, they would interface each other and in some cases causes unexpected result. 

Assume we have 7 partitions in one host, and compaction manager trigger compaction on each one of them in the same order alway, p1, p2, p3...p7, p1, p2, p3... p7. In this case, P7 would always use CompactionAllPolicy and the other partitions would always use statsbased compaction policy.

I think what we need is a per partition HybridCompactionPolicy. It can't be shared by partitions, because it's no longer a stateless object.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Thanks for the advise. Implemented with first approach that maintains a map which key is replicaId and value is the counter.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
```diff
@@ -173,15 +178,30 @@ boolean removeBlobStore(BlobStore store) {
...
-  private CompactionDetails getCompactionDetails(BlobStore blobStore) throws StoreException {
-    return blobStore.getCompactionDetails(compactionPolicy);
+  CompactionDetails getCompactionDetails(BlobStore blobStore) throws StoreException {
+    ReplicaId replicaId = blobStore.getReplicaId();
+    if (compactionPolicyFactory != null && compactionPolicyFactory instanceof HybridCompactionPolicyFactory) {
+      CompactionPolicyCounter compactionPolicyCounter =
+          replicaToCounterMap.getOrDefault(replicaId, new CompactionPolicyCounter(storeConfig));
+      compactionPolicyCounter.increment();
+      replicaToCounterMap.put(replicaId, compactionPolicyCounter);
+    }
```
justinlin-linkedin, line 203: we actually don't have to differentiate the compaction policy factory when creating and incrementing the couter. We can just increment it for every policy.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -339,6 +346,7 @@ public StoreConfig(VerifiableProperties verifiableProperties) {
     storeMaxNumberOfEntriesToReturnFromJournal =
         verifiableProperties.getIntInRange("store.max.number.of.entries.to.return.from.journal", 5000, 1, 10000);
     storeDeletedMessageRetentionDays = verifiableProperties.getInt("store.deleted.message.retention.days", 7);
+    storeCompactionPolicySwitchPeriod = verifiableProperties.getInt("store.compaction.policy.switch.period", 7);
```
justinlin-linkedin, line 349: let's set a limit here by using getInt(name, default, min, max). This would make sure that period would be 0 or negative number.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicyCounter.java
```diff
@@ -0,0 +1,39 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.store;
+
+import com.github.ambry.config.StoreConfig;
+import java.util.concurrent.atomic.AtomicInteger;
```
justinlin-linkedin, line 17: not used.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicy.java
```diff
@@ -37,6 +37,6 @@
    * @throws StoreException
    */
   CompactionDetails getCompactionDetails(long totalCapacity, long usedCapacity, long segmentCapacity,
-      long segmentHeaderSize, List<String> logSegmentsNotInJournal, BlobStoreStats blobStoreStats)
+      long segmentHeaderSize, List<String> logSegmentsNotInJournal, BlobStoreStats blobStoreStats, CompactionPolicyCounter compactionPolicyCounter)
```
justinlin-linkedin, line 40: I was wonder if you would consider this way. Using BlobStoreStats to pass store id, which is essentially the same as replica id number, to CompactionPolicy. And in CompactionPolicy, we keep a map from store id to a counter? In this way, we don't have to change interface here.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicyCounter.java
```diff
@@ -0,0 +1,39 @@
...
+
+
+/**
+ * A counter used to switch {@link CompactAllPolicy}.
+ */
+public class CompactionPolicyCounter {
```
justinlin-linkedin, line 23: nit: drop public and make this class package-private.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
SophieGuo410: Move map inside Hybrid compaction policy.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicyCounter.java
SophieGuo410: Removed.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicyCounter.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,192 @@
...
+  @Override
+  public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapacity, long segmentCapacity,
+      long segmentHeaderSize, List<String> logSegmentsNotInJournal, BlobStoreStats blobStoreStats, String dataDir)
+      throws StoreException {
+    String storeId = blobStoreStats.getStoreId();
+    CompactionPolicySwitchInfo compactionPolicySwitchInfo = blobToCompactionPolicySwitchInfoMap.getOrDefault(storeId,
+        new CompactionPolicySwitchInfo(new CompactionPolicyCounter(storeConfig), INIT_COMPACT_ALL_TIME));
+    blobToCompactionPolicySwitchInfoMap.put(storeId, compactionPolicySwitchInfo);
+    CompactionPolicy selectCompactionPolicy =
+        selectCompactionPolicyAndUpdateCompactionPolicySwitchInfo(dataDir, storeId, compactionPolicySwitchInfo);
```
justinlin-linkedin, line 77: we can make logic a bit better without checking file every time selecting a policy. Something like
```
if (!blobToCompactionPolicySwitchInfos.contain(storeId)) {
    File file = new File(Paths.get(dataDir, COMPACT_POLICY_INFO_PATH))
    if (file.exists()) {
        ObjectMapper objectMapper = new ObjectMapper();
        CompactionPolicySwitchInfo switchInfo = objectMapper.readValue(file, CompactionPolicySwithInfo.class);
        blobToCompactionPolicySwtichInfos.put(storeId, switchInfo);
    } else {
        blobToCompactionPolicySwtichInfos.put(storeId, new CompactionPolicySwitchInfo(new CompactionPolicyCounter(storeConfig), INIT_COMPACT_ALL_TIME));
    }
}
```

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,192 @@
...
+public class HybridCompactionPolicy implements CompactionPolicy {
+  private final Time time;
+  private final StoreConfig storeConfig;
+  private static final Logger logger = LoggerFactory.getLogger(HybridCompactionPolicy.class);
+  private final Map<String, CompactionPolicySwitchInfo> blobToCompactionPolicySwitchInfoMap;
+  private final ObjectMapper mapper = new ObjectMapper();
+  private static final String COMPACTION_POLICY_COUNTER = "compactionPolicyCounter";
+  private static final String VALUE = "value";
+  private static final String LAST_COMPACT_ALL_TIME = "lastCompactAllTime";
+  private static final String COMPACT_POLICY_INFO_PATH = "/compactionPolicyInfo.";
```
justinlin-linkedin, line 45: you can omit leading slash for two reasons
1. It's less cross-platform compatible, we can always use File.separator. Not that it matters here, just a nice pattern.
2. Use Paths.get is probably a better way to concatenate different parts of the file path.

Also, dataDir already has the partition id/replica id/store id as the last part of the filepath, we don't need to suffix storeId in the filename here. But we can add .json as extension, just to indicate this is a json formatted file.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,192 @@
...
+    CompactionPolicySwitchInfo compactionPolicySwitchInfo = blobToCompactionPolicySwitchInfoMap.getOrDefault(storeId,
+        new CompactionPolicySwitchInfo(new CompactionPolicyCounter(storeConfig), INIT_COMPACT_ALL_TIME));
+    blobToCompactionPolicySwitchInfoMap.put(storeId, compactionPolicySwitchInfo);
+    CompactionPolicy selectCompactionPolicy =
+        selectCompactionPolicyAndUpdateCompactionPolicySwitchInfo(dataDir, storeId, compactionPolicySwitchInfo);
+    logger.info("Current compaction policy  is : {}", selectCompactionPolicy);
+    backUpCompactionPolicyInfo(dataDir, storeId, compactionPolicySwitchInfo);
+    compactionPolicySwitchInfo.getCompactionPolicyCounter().increment();
+    return selectCompactionPolicy.getCompactionDetails(totalCapacity, usedCapacity, segmentCapacity, segmentHeaderSize,
+        logSegmentsNotInJournal, blobStoreStats, null);
```
justinlin-linkedin, line 82: please pass dataDir here.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,192 @@
...
+   * {
+   *   "compactionPolicyCounter" : {
+   *     "value" : 1
+   *   },
+   *   "lastCompactionTime" : 1593128052284
+   * }
+   * @param compactionPolicySwitchInfo the info to determine which {@link CompactionPolicy} to use this round.
+   * @param file the backup file stores {@link CompactionPolicySwitchInfo}
+   */
+  private void recoverCompactionPolicySwitchInfo(CompactionPolicySwitchInfo compactionPolicySwitchInfo, File file) {
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,192 @@
...
+            compactionPolicyCounter.has(VALUE) ? compactionPolicyCounter.getInt(VALUE) : INIT_COUNTER_VALUE;
+      } else {
+        logger.trace("CompactionPolicyCounter is null");
+      }
+      long lastCompactAllTime =
+          jsonObject.has(LAST_COMPACT_ALL_TIME) ? jsonObject.getLong(LAST_COMPACT_ALL_TIME) : INIT_COMPACT_ALL_TIME;
+      compactionPolicySwitchInfo.setLastCompactAllTime(lastCompactAllTime);
+      compactionPolicySwitchInfo.getCompactionPolicyCounter().setValue(compactionPolicyCounterValue);
+    } catch (IOException e) {
+      logger.trace("tempFile : {} is not exist", file);
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -81,6 +81,13 @@
   @Default("7")
   public final int storeDeletedMessageRetentionDays;
 
+  /**
+   * How often the HybridCompactionPolicy swith from StatsBasedCompactionPolicy to CompactAllPolicy.
```
zzmao, line 85: typo: swith-> switch

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1081,6 +1081,13 @@ public long getEndPositionOfLastPut() throws StoreException {
     return replicaStatusDelegates;
   }
 
+  /**
+   * @return a {@link ReplicaId} associated with this store
+   */
+  public ReplicaId getReplicaId() {
```
zzmao, line 1087: Looks like this method is not used.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
```diff
@@ -13,12 +13,15 @@
  */
 package com.github.ambry.store;
 
+import com.github.ambry.clustermap.ReplicaId;
```
zzmao, line 16: not used.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
```diff
@@ -13,12 +13,15 @@
...
 import com.github.ambry.config.StoreConfig;
 import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.util.Collection;
 import java.util.EnumSet;
+import java.util.HashMap;
```
zzmao, line 22: not used.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
```diff
@@ -13,12 +13,15 @@
...
 import java.util.HashSet;
+import java.util.Map;
```
zzmao, line 24: not used.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicy.java
```diff
@@ -33,10 +33,11 @@
    *                                {@link Journal}
    * @param blobStoreStats {@link BlobStoreStats} pertaining to the {@link BlobStore} for which
    * {@link CompactionDetails} are requested
+   * @param dataDir
```
zzmao, line 36: java doc

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,192 @@
...
+    return blobToCompactionPolicySwitchInfoMap.get(storeId);
+  }
+
+  /**
+   * Recover the {@link CompactionPolicySwitchInfo} from backup file.
+   * {
+   *   "compactionPolicyCounter" : {
+   *     "value" : 1
+   *   },
+   *   "lastCompactionTime" : 1593128052284
```
zzmao, line 104: add a field called "lastCompactionAllTime"?

SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicy.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
SophieGuo410: Removed.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
SophieGuo410: Removed.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+   */
+  CompactionPolicy selectCompactionPolicyAndUpdateCompactionPolicySwitchInfo(
+      CompactionPolicySwitchInfo compactionPolicySwitchInfo) {
+    if (compactionPolicySwitchInfo == null) {
+      logger.trace("CompactionPolicySwitchInfo is null");
+      return new StatsBasedCompactionPolicy(storeConfig, time);
+    }
+    if (readyToTriggerCompactionAllPolicy(compactionPolicySwitchInfo)) {
+      logger.trace("Return CompactAllPolicy this round");
+      updateCompactionInfo(compactionPolicySwitchInfo);
```
lightningrob, line 133: Don't you want to do the update either way?

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+      return new CompactionPolicySwitchInfo(new CompactionPolicyCounter(storeConfig.storeCompactionPolicySwitchPeriod),
+          INIT_COMPACT_ALL_TIME);
+    }
+  }
+
+  /**
+   * Selects which compaction policy to use for current compaction cycle.
+   * @return {@link CompactionPolicy} to use for current compaction cycle(Selects between {@link StatsBasedCompactionPolicy} and {@link CompactAllPolicy}).
+   */
+  CompactionPolicy selectCompactionPolicyAndUpdateCompactionPolicySwitchInfo(
```
lightningrob, line 125: This method name is a mouthful.  Is there a downside to the caller first calling updateCompactionInfo() and then this method which could just be selectCompactionPolicy()?

ambry-store/src/test/java/com/github/ambry/store/CompactionPolicyTest.java
```diff
@@ -82,6 +94,135 @@ public CompactionPolicyTest(String compactionPolicyFactoryStr) throws Exception
     mockBlobStoreStats = blobStore.getBlobStoreStats();
     CompactionPolicyFactory compactionPolicyFactory = Utils.getObj(compactionPolicyFactoryStr, config, time);
     compactionPolicy = compactionPolicyFactory.getCompactionPolicy();
+    dataDirPath = Paths.get(MOUNT_PATH).toAbsolutePath();
+  }
+
+  private void cleanupBackupFiles() throws IOException {
```
lightningrob, line 100: Minor: please move private utility methods to bottom.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -81,6 +81,13 @@
   @Default("7")
   public final int storeDeletedMessageRetentionDays;
 
+  /**
+   * How often the HybridCompactionPolicy switch from StatsBasedCompactionPolicy to CompactAllPolicy.
+   */
+  @Config("store.compaction.policy.switch.period")
+  @Default("7")
+  public final int storeCompactionPolicySwitchPeriod;
```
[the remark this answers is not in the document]
SophieGuo410: Updated. And I separate the storeCompactionPolicySwitchTimestampDays and storeCompactionPolicySwitchCounterDays to make the tuning more flexible.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -1117,6 +1117,13 @@ void cancel() {
     }
   }
 
+  /**
+   * @return the storeId for this blob.
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicy.java
```diff
@@ -33,10 +33,11 @@
    *                                {@link Journal}
    * @param blobStoreStats {@link BlobStoreStats} pertaining to the {@link BlobStore} for which
    * {@link CompactionDetails} are requested
+   * @param dataDir The directory to use to store compactionPolicyInfo
```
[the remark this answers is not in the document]
SophieGuo410: Updated with CompactionPolicySwitchInfo. Sorry for the confusion.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicyCounter.java
```diff
@@ -0,0 +1,48 @@
...
+
+  CompactionPolicyCounter(int storeCompactionPolicySwitchPeriod) {
+    this.storeCompactionPolicySwitchPeriod = storeCompactionPolicySwitchPeriod;
+  }
+
+  //make sure objectMapper can work correctly
+  CompactionPolicyCounter() {
+  }
+
+  public int getValue() {
```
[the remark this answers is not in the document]
SophieGuo410: Change to counter to be more clear and added the javadocs.

ambry-store/src/main/java/com/github/ambry/store/CompactionPolicySwitchInfo.java
```diff
@@ -0,0 +1,49 @@
...
+  CompactionPolicySwitchInfo(CompactionPolicyCounter compactionPolicyCounter, long lastCompactAllTime) {
+    this.compactionPolicyCounter = compactionPolicyCounter;
+    this.lastCompactAllTime = lastCompactAllTime;
+  }
+
+  //make sure objectMapper can work correctly
+  CompactionPolicySwitchInfo() {
+  }
+
+  CompactionPolicyCounter getCompactionPolicyCounter() {
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+        logSegmentsNotInJournal, blobStoreStats, dataDir);
+  }
+
+  /**
+   * Get and Recover the CompactionPolicySwitchInfo from file.
+   * @param dataDir The directory to store the file.
+   * @param storeId id of the BlobStore
+   * @return {@link CompactionPolicySwitchInfo} gets from map or recover from file if needed.
+   */
+  private CompactionPolicySwitchInfo getAndRecoverCompactionPolicySwitchInfo(String storeId, String dataDir) {
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+    return blobToCompactionPolicySwitchInfoMap.get(storeId);
+  }
+
+  /**
+   * Recover the {@link CompactionPolicySwitchInfo} from backup file.
+   * {
+   *   "compactionPolicyCounter" : {
+   *     "storeCompactionPolicySwitchPeriod" : 3,
+   *     "counter" : 1,
+   *     "value" : 1
```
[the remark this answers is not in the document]
SophieGuo410: Updated to have counter only.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+   *   },
+   *   "lastCompactAllTime" : 1593463435900
+   * }
+   * @param file the backup file stores {@link CompactionPolicySwitchInfo}
+   */
+  private CompactionPolicySwitchInfo recoverCompactionPolicySwitchInfo(File file) {
+    try {
+      ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
SophieGuo410: Done.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+      return objectMapper.readValue(file, CompactionPolicySwitchInfo.class);
+    } catch (IOException e) {
+      logger.error("Could not deserialize file : {} into {} Object", file, CompactionPolicySwitchInfo.class.getName());
+      return new CompactionPolicySwitchInfo(new CompactionPolicyCounter(storeConfig.storeCompactionPolicySwitchPeriod),
```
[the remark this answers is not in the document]
SophieGuo410: Add the metrics.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+   */
+  private boolean readyToTriggerCompactionAllPolicy(CompactionPolicySwitchInfo compactionPolicySwitchInfo) {
+    return compactionPolicySwitchInfo.getCompactionPolicyCounter() != null
+        && compactionPolicySwitchInfo.getCompactionPolicyCounter().getValue() == 0 ||
+        compactionPolicySwitchInfo.getLastCompactAllTime() + TimeUnit.DAYS.toMillis(
+            storeConfig.storeCompactionPolicySwitchPeriod) <= System.currentTimeMillis();
+  }
+
+  /**
+   * Update the {@link CompactionPolicySwitchInfo} before the start of {@link CompactAllPolicy}
```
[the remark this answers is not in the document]
SophieGuo410: In this situation, it will recover the counter value in file which is 0 since we trigger compactAllPolicy before host crashes,  so which means after restart it will run compactAllPolicy one more time.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+        tempFile.createNewFile();
+        mapper.defaultPrettyPrintingWriter().writeValue(tempFile, compactionPolicySwitchInfo);
+      } catch (IOException e) {
+        logger.error("Exception while store compaction policy info for local report. Output file path - {}",
+            tempFile.getAbsolutePath(), e);
+      }
+    }
+  }
+}
+
```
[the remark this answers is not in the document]
SophieGuo410: Removed.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: I updated the method name to updateCompactionInfoWhenCompactAll to avoid confusion. This method is used to keep record of the last time when we run CompactAllPolicy, so if we passed "storeCompactionPolicySwitchTimestampDays" days since we run compactAllPolicy, we will re-run it again.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: I will only update the compactionPolicySwitchInfo when we trigger the compactAllPolicy. Already re-named the method name to updateCompactionInfoWhenCompactAll() instead of updateCompactionInfo().

ambry-store/src/test/java/com/github/ambry/store/CompactionPolicyTest.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -209,6 +209,7 @@ private void getDeprecatedContainers() {
           deprecatedContainers.add(new Pair<>(container.getParentAccountId(), container.getId()));
         }
       });
+      //TODO: Filter out the INACTIVE containers from deprecatedContainers set if it's already been compacted.
```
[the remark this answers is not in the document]
SophieGuo410: Yes. That's an optimize that I plan to implement in a separate pr.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,191 @@
...
+  public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapacity, long segmentCapacity,
+      long segmentHeaderSize, List<String> logSegmentsNotInJournal, BlobStoreStats blobStoreStats, String dataDir)
+      throws StoreException {
+    String storeId = blobStoreStats.getStoreId();
+    CompactionPolicySwitchInfo compactionPolicySwitchInfo = getCompactionPolicySwitchInfo(storeId, dataDir, blobStoreStats);
+    CompactionPolicy selectCompactionPolicy =
+        selectCompactionPolicyAndUpdateCompactionPolicySwitchInfo(compactionPolicySwitchInfo);
+    logger.info("Current compaction policy  is : {}", selectCompactionPolicy);
+    backUpCompactionPolicyInfo(dataDir, compactionPolicySwitchInfo);
+    compactionPolicySwitchInfo.getCompactionPolicyCounter().increment();
```
justinlin-linkedin, line 72: shouldn't we increment the counter before select which policy to use?

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,191 @@
...
+   */
+  @Override
+  public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapacity, long segmentCapacity,
+      long segmentHeaderSize, List<String> logSegmentsNotInJournal, BlobStoreStats blobStoreStats, String dataDir)
+      throws StoreException {
+    String storeId = blobStoreStats.getStoreId();
+    CompactionPolicySwitchInfo compactionPolicySwitchInfo = getCompactionPolicySwitchInfo(storeId, dataDir, blobStoreStats);
+    CompactionPolicy selectCompactionPolicy =
+        selectCompactionPolicyAndUpdateCompactionPolicySwitchInfo(compactionPolicySwitchInfo);
+    logger.info("Current compaction policy  is : {}", selectCompactionPolicy);
```
justinlin-linkedin, line 70: please add datadir and store id in the info log.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Selects which compaction policy to use for current compaction cycle.
+   * @return {@link CompactionPolicy} to use for current compaction cycle(Selects between {@link StatsBasedCompactionPolicy} and {@link CompactAllPolicy}).
+   */
+  CompactionPolicy selectCompactionPolicyAndUpdateCompactionPolicySwitchInfo(
+      CompactionPolicySwitchInfo compactionPolicySwitchInfo) {
+    if (compactionPolicySwitchInfo == null) {
+      logger.trace("CompactionPolicySwitchInfo is null");
```
justinlin-linkedin, line 127: nit: please add datadir or store id in the log.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * Back up {@link CompactionPolicySwitchInfo} in Json format for certain {@link BlobStore}
+   * @param dataDir The directory to store the file.
+   * @param compactionPolicySwitchInfo the info to determine which {@link CompactionPolicy} to use this round.
+   */
+  private void backUpCompactionPolicyInfo(String dataDir, CompactionPolicySwitchInfo compactionPolicySwitchInfo) {
+    if (dataDir != null && !dataDir.isEmpty()) {
+      File tempFile = new File(Paths.get(dataDir, COMPACT_POLICY_INFO_PATH).toString());
```
justinlin-linkedin, line 181: nit: I encourage to create and write data to a temporary file and then rename the temporary file to COMPACT_POLICY_INFO_PATH file
Something like 
```
File tempFile = new File(dataDir, COMPACT_POLICY_INFO_PATH + ".temp");
if (!tempFile.exists()) {
    tempFile.createNewFile();
}
objectMapper.defaultPrettyPrintingWriter().writeValue(tempFile, compactionPolicySwitchInfo);
tempFile.renameTo(new File(dataDir, COMPACT_POLICY_INFO_PATH);
```
The reason to use a temporary file and rename it after is that POSIX guarantees file rename is an atomic operation, but not the file write. The json file is pretty small so that file writing should be atomic, but it's just a good pattern to follow.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: That's the tricky part. I'm planning to increment after the backup with two reasons:
1. when counter equals to 0, it will run compactAllPolicy. which means if we don't have any back up files yet, we will run compact all policy.
2. if we are running compactAll for a long time and host restart, the file backs up the current status and re-run compactAll instead of statsBased.
Let me know if you have any other concern.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Done.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Done.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
SophieGuo410: Updated to use temp file and rename after write. I didn't add the file exist check since createNewFile() is able to handle it.

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -0,0 +1,194 @@
...
+
+  /**
+   * @return blobToCompactionPolicySwitchInfoMap which key is storeId and value is {@link CompactionPolicySwitchInfo}
+   */
+  Map<String, CompactionPolicySwitchInfo> getBlobToCompactionPolicySwitchInfoMap() {
+    return this.blobToCompactionPolicySwitchInfoMap;
+  }
+
+  /**
+   * Back up {@link CompactionPolicySwitchInfo} in Json format for each {@link BlobStore}
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

PR #1561, opened 2020-06-12
Make loggers static final when possible

cgtz:
In almost all cases, loggers should be static. When run with log4j2,
there is more overhead to get a Logger instance, whereas with log4j1,
usually only the first request to get a logger for a given name would be
expensive. This means that instantiating loggers for classes with many
instances such as BlobReadOptions, which is instantiated for every blob
served, have a visible impact when analyzed with a sampling profiler.

This commit is a mass find and replace of non-static loggers.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/CompositeClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/CompositeClusterManager.java
@@ -35,7 +35,7 @@ import org.slf4j.LoggerFactory;
  * and reports inconsistencies in the views from the two underlying cluster managers.
  */
 class CompositeClusterManager implements ClusterMap {
-  private static final Logger LOGGER = LoggerFactory.getLogger(CompositeClusterManager.class);
+  private static final Logger logger = LoggerFactory.getLogger(CompositeClusterManager.class);
   final StaticClusterManager staticClusterManager;
   final HelixClusterManager helixClusterManager;
   final HelixClusterManagerMetrics helixClusterManagerMetrics;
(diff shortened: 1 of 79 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/CompositeClusterManager.java
```diff
@@ -39,7 +39,6 @@
   final StaticClusterManager staticClusterManager;
   final HelixClusterManager helixClusterManager;
   final HelixClusterManagerMetrics helixClusterManagerMetrics;
-  private final Logger logger = LoggerFactory.getLogger(CompositeClusterManager.class);
```
lightningrob, line 42: Can you rename LOGGER to logger?

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogger.java
```diff
@@ -22,11 +22,11 @@
  */
 public class PublicAccessLogger {
 
-  private Logger publicAccessLogger = LoggerFactory.getLogger("PublicAccessLogger");
+  private static final Logger publicAccessLogger = LoggerFactory.getLogger("PublicAccessLogger");
 
   private final String[] requestHeaders;
   private final String[] responseHeaders;
-  private final Logger logger = LoggerFactory.getLogger(getClass());
+  private static final Logger logger = LoggerFactory.getLogger(PublicAccessLogger.class);
```
lightningrob, line 29: We can probably remove this second logger, it's only used in one trace message.

cgtz: Good idea, that was not a very useful log message anyways.

PR #1564, opened 2020-06-15
Check the size of vcr node list before finding one to assign as cloud peer.

ankagrawal:


ambry-replication/src/main/java/com/github/ambry/replication/CloudToStoreReplicationManager.java
```diff
@@ -240,8 +240,13 @@ private void removeCloudReplica(String partitionName) {
...
    * @return randomly selected {@link DataNodeId} object.
+   * @throws ReplicationException if there are no vcr nodes.
    */
-  private DataNodeId getCloudDataNode() {
+  private DataNodeId getCloudDataNode() throws ReplicationException {
+    if (vcrNodes.get().size() == 0) {
+      logger.error("No vcr node found to replicate partition from cloud.");
+      throw new ReplicationException("No vcr node found to replicate partition from cloud.");
+    }
     return vcrNodes.get().toArray(new CloudDataNode[0])[Utils.getRandomShort(new Random()) % vcrNodes.get().size()];
```
lightningrob, line 250: I realize this isn't new code, but it seems inefficient for a getter method to have to copy a collection to an array every time it needs to get an element from it.  It makes me wonder if the concurrent set is really the optimal data structure for this.

lightningrob: It's fine to address last comment in a future PR.

ambry-replication/src/main/java/com/github/ambry/replication/CloudToStoreReplicationManager.java
```diff
@@ -240,8 +240,13 @@ private void removeCloudReplica(String partitionName) {
   /**
    * Randomly select a {@link DataNodeId} from list of {@code vcrNodes}.
    * @return randomly selected {@link DataNodeId} object.
+   * @throws ReplicationException if there are no vcr nodes.
    */
-  private DataNodeId getCloudDataNode() {
+  private DataNodeId getCloudDataNode() throws ReplicationException {
+    if (vcrNodes.get().size() == 0) {
```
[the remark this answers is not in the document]
ankagrawal: done.

ambry-replication/src/main/java/com/github/ambry/replication/CloudToStoreReplicationManager.java
```diff
@@ -240,8 +240,13 @@ private void removeCloudReplica(String partitionName) {
...
+      logger.error("No vcr node found to replicate partition from cloud.");
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-replication/src/main/java/com/github/ambry/replication/CloudToStoreReplicationManager.java
ankagrawal: You are right. It makes sense to make it a list. We modify it only in the `CloudToStoreReplicationManager::handleChangeInVcrNodes()` method and access it only in the `CloudToStoreReplicationManager::getCloudDataNode()`. And it makes sense to modify the atomic reference directly.  Have made the change.

ambry-replication/src/main/java/com/github/ambry/replication/CloudToStoreReplicationManager.java
```diff
@@ -240,9 +240,14 @@ private void removeCloudReplica(String partitionName) {
...
+   * @throws ReplicationException if there are no vcr nodes.
    */
-  private DataNodeId getCloudDataNode() {
-    return vcrNodes.get().toArray(new CloudDataNode[0])[Utils.getRandomShort(new Random()) % vcrNodes.get().size()];
+  private DataNodeId getCloudDataNode() throws ReplicationException {
+    List<CloudDataNode> nodes = vcrNodes.get();
+    if (nodes.isEmpty()) {
+      throw new ReplicationException("No VCR node found to replicate partition from cloud.");
+    }
+    return nodes.get(Utils.getRandomShort(new Random()) % nodes.size());
```
lightningrob, line 250: I didn't notice it before, but I think initializing a new Random() each time will result in the same value each time.  At least that's how it used to behave, not sure if it's still the case.  Either way, I would initialize the Random statically and reuse.  I recommend adding a test case that calls this method consecutive times and verifies different results are returned.  And one to test the no-nodes exception case.

ankagrawal: Sure, I will change the `Random` to static, and add the tests.
Note that ideally, in a perfect random setting, I assume it could be possible for a few calls to `getNextInt()` to return same value (Maybe this will happen rarely, but this is a possibility). So in my test I am just checking that multiple calls return more than one value.

As for the behavior of `Random`, from the constructor of `Random` it seems like every instantiation gets a unique seed. Below code snippet is from `java.util.Random` class.

```
    /**
     * Creates a new random number generator. This constructor sets
     * the seed of the random number generator to a value very likely
     * to be distinct from any other invocation of this constructor.
     */
    public Random() {
        this(seedUniquifier() ^ System.nanoTime());
    }
```


Also I wrote an independent test that sort of proves this.

```
public class RandomTest {
	public static void main(String[] args) {
		for(int i = 0; i < 10; i++) {
			System.out.println(new Random().nextInt()%17);
		}
	}
}
```

lightningrob: My information was from at least 10 years ago, so they likely changed it since then.

lightningrob:
There are a lot of failing router tests that seem unrelated to your change.

PR #1568, opened 2020-06-16
Don't rely on ABS blob properties metadata to contain blob id

ankagrawal:
This fixes the Null Pointer Exception in `CloudBlobStore::validateAccountAndContainer`

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -551,7 +551,7 @@ class CloudBlobStore implements Store {
       // This is a delete request from frontend
       if (metadata.isDeleted()) {
         throw new StoreException(
-            "Cannot delete id " + metadata.getId() + " since it is already marked as deleted in cloud.",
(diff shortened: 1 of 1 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -538,7 +538,7 @@ private void validateAccountAndContainer(Map<String, CloudBlobMetadata> cloudBlo
    */
   private boolean preDeleteValidation(CloudBlobMetadata metadata, StoreKey key, Map<String, Object> updateFields)
       throws StoreException {
-    validateAccountAndContainer(Collections.singletonMap(metadata.getId(), metadata), Collections.singletonList(key));
+    validateAccountAndContainer(Collections.singletonMap(key.getID(), metadata), Collections.singletonList(key));
```
jsjtzyy, line 541: Do we also need to change the `metadata.getId()` at line 554?

ankagrawal: fixed.

PR #1570, opened 2020-06-18
Change sanity check logic for undelete so recovery can work

justinlin-linkedin:
There are several changes to this PR
1. Undelete will no longer check if the blob is expired for replication
2. Undelete will now take a MessageInfo as argument in recovery, and will only use info to construct an index value, without any sanity checking.
3. copyRecords in compaction will not use markAsUndelete to check sanity, instead it will have it's own sanity check.

This would fix certain recovery and replication corner cases.
Case 1. We have P, D, U, T for a blob. The PUT has a short ttl, so we need the last TTL_UPDATE to make it permanent. When the whole index files are gone and we are reconstructing index from log segments, we would fail on U since the P is expired in recovery. This PR would fix that.

Case 2. We have P, D, U, D for a blob. The compaction compacts Put and first DELETE, (UNDELETE and second DELETE are in a different segment, which is not under compaction). Then we only have U and D after compaction. When the whole index files are gone and we are reconstructing index from log, we would fail on U since there is nothing prior to it. This PR would fix that.

Case 3. We have P, D in local, and the remote has P, D, U, T. Somehow local stops for a while until P is expired. When restarting local process, it starts replicating from remote. And it will fail on applying U, since P is expired. This PR would fix that.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -45,6 +45,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+
 /**
  * Component that removes the "dead" data from the set of provided log segments and reclaims space.
  * <p/>
@@ -85,7 +86,7 @@ class BlobStoreCompactor {
   private final AccountService accountService;
   private volatile boolean isActive = false;
   private PersistentIndex srcIndex;
-  private final Set<Pair<Short,Short>> deprecatedContainers;
+  private final Set<Pair<Short, Short>> deprecatedContainers;
 
   private Log tgtLog;
   private PersistentIndex tgtIndex;
@@ -116,8 +117,7 @@ class BlobStoreCompactor {
   BlobStoreCompactor(String dataDir, String storeId, StoreKeyFactory storeKeyFactory, StoreConfig config,
       StoreMetrics srcMetrics, StoreMetrics tgtMetrics, DiskIOScheduler diskIOScheduler,
       DiskSpaceAllocator diskSpaceAllocator, Log srcLog, Time time, UUID sessionId, UUID incarnationId,
-      AccountService accountService)
-      throws IOException, StoreException {
+      AccountService accountService) throws IOException, StoreException {
     this.dataDir = new File(dataDir);
     this.storeId = storeId;
     this.storeKeyFactory = storeKeyFactory;
@@ -205,7 +205,8 @@ class BlobStoreCompactor {
     deprecatedContainers.clear();
     if (accountService != null) {
       accountService.getContainersByStatus(Container.ContainerStatus.DELETE_IN_PROGRESS).forEach((container) -> {
-        if (container.getDeleteTriggerTime() + TimeUnit.DAYS.toMillis(config.storeContainerDeletionRetentionDays) <= System.currentTimeMillis()) {
+        if (container.getDeleteTriggerTime() + TimeUnit.DAYS.toMillis(config.storeContainerDeletionRetentionDays)
+            <= System.currentTimeMillis()) {
           deprecatedContainers.add(new Pair<>(container.getParentAccountId(), container.getId()));
         }
       });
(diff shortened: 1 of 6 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -980,35 +964,54 @@ IndexValue markAsPermanent(StoreKey id, FileSpan fileSpan, MessageInfo info, lon
...
-        new IndexValue(value.getSize(), value.getOffset(), value.getFlags(), value.getExpiresAtMs(), operationTimeMs,
-            value.getAccountId(), value.getContainerId(), lifeVersion);
-    newValue.setNewOffset(fileSpan.getStartOffset());
-    newValue.setNewSize(size);
+    IndexValue newValue;
+    if (info != null) {
+      // This is from recovery. In recovery, we don't need to do any sanity check because
+      // 1. we already know the IndexValue has it's source in the log.
+      // 2. some sanity check will fail.
+      //    assume we have P, D, U, D in the log, then a compaction cycle compacted P and first D,
```
zzmao, line 993: In this case, we still recovery Undelete index, but P can't be accessed, right?

justinlin-linkedin: That's right. We used to think that when there is a U, there must be a P before U. This is wrong. After compaction, anything could happen. After compaction, in this case, we are left with only U and D, without any P. And recovery need to work for this case as well.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -722,15 +720,22 @@ private IndexValue findKey(StoreKey key, FileSpan fileSpan, EnumSet<IndexEntryTy
...
-    // less than the given lifeVersion. We don't care about the first value's type, it can be a put, ttl_update or delete, it
-    // can even be an undelete.
+    // This is from replication. For replication, undelete should be permitted only when
+    // 1. The oldest record is PUT
+    // 2. the latest record's lifeVersion is less then undelete's lifeVersion.
+    // 3. Replication doesn't care if the latest record is delete or not. In local, we can have a PUT record when
+    // the remote has PUT, DELETE, UNDELETE. When replicating from remote, local has to insert UNDELETE, where
+    // there is no prior DELETE.
+    // 4. Replication doesn't care if the latest record is expired or not. In local, we can have a PUT record when
+    // the remove has PUT, DELETE, UNDELETE TTL_UPDATE. When replicating from remote, PUT might already have expired.
```
jsjtzyy, line 736: remove -> remote

ambry-store/src/main/java/com/github/ambry/store/IndexValue.java
```diff
@@ -362,6 +362,14 @@ void setNewOffset(Offset newOffset) {
     setOriginalMessageOffset(oldOffset);
   }
 
+  /**
+   * Updates the {@link #lifeVersion} of this {@link IndexValue}.
+   * @param lifeVersion the new lifeVersion to set.
+   */
+  void setNewLifeVersion(short lifeVersion) {
```
jsjtzyy, line 369: This is only used for testing, right?  (Also format the file please)

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -767,16 +768,27 @@ private boolean copyRecords(LogSegment logSegmentToCopy, List<IndexEntry> srcInd
...
+                  throw new StoreException(
+                      "Id " + srcIndexEntry.getKey() + " already has undelete " + valueFromTgtIdx + " in index "
+                          + dataDir, StoreErrorCodes.ID_Undeleted);
+                }
+                // Undelete would increase the life version
+                if (valueFromTgtIdx.getLifeVersion() >= srcValue.getLifeVersion()) {
+                  throw new StoreException(
+                      "Id " + srcIndexEntry.getKey() + " has bad lifeversion " + valueFromTgtIdx + " in index "
+                          + dataDir, StoreErrorCodes.Life_Version_Conflict);
+                }
```
jsjtzyy, line 784: Looks like if `valueFromTgtIdx != null`, we only check version and undelete state, and we still add a new index entry to `tgtIndex`?

justinlin-linkedin: We are throwing an error when the `valueFromTgtIndex` is not what we expect. The idea is that, in target index, if there are already index value existed for this key, then it can't be an undelete. Compaction would remove all previous undelete in the same segment. And it's lifeVersion can't be equal or greater than the undelete's lifeVersion. It' pretty easy to understand why it can't be greater than undelete, for example if we have PUT in the target index, and it's lifeVersion has to be smaller than the undelete. Also it can't be equal to undelete, because undelete always increase the lifeVersion.

PR #1571, opened 2020-06-21
Force empty store to stay in BOOTSTRAP before catchup completes

jsjtzyy:
When disk crashed due to I/O error, we lost all partitions on this disk. After a new disk
is mounted, Ambry is able to recreate the stores on new disk during reboot. These
recreated stores are empty and probably way behind their peer replicas, so we need
to put them in BOOTSTRAP state until they have caught up with peers. Considering
Leader-based Replication will be introduced, this PR guarantees those empty stores
won't be selected LEADER because they are kept in BOOTSTRAP state. (Otherwise, some stores
may be selected as LEADER but they have to spend a couple of hours/days completing catchup.
Meanwhile new data in remote dc(s) cannot be replicated to current dc.)

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -411,6 +411,18 @@ public class StorageManager implements StoreManager {
   }
 
   /**
+   * Create a bootstrap file in given replica directory if the file is not present.
+   * @param replica the {@link ReplicaId} whose directory should contain the bootstrap file.
+   * @throws IOException
+   */
+  static void createBootstrapFileIfAbsent(ReplicaId replica) throws IOException {
+    File bootstrapFile = new File(replica.getReplicaPath(), BlobStore.BOOTSTRAP_FILE_NAME);
+    if (!bootstrapFile.exists()) {
+      bootstrapFile.createNewFile();
+    }
+  }
+
+  /**
    * Maybe delete the residual directory associated with removed replica.
    * @param partitionName name of replica that is already removed
    */
@@ -440,6 +452,7 @@ public class StorageManager implements StoreManager {
     public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
       // check if partition exists on current node
       ReplicaId replica = partitionNameToReplicaId.get(partitionName);
+      Store store = null;
       if (replica == null) {
         // there can be two scenarios:
         // 1. this is the first time to add new replica onto current node;
@@ -473,10 +486,13 @@ public class StorageManager implements StoreManager {
                 StateTransitionException.TransitionErrorCode.HelixUpdateFailure);
           }
         }
+        // if addBlobStore succeeds, it is guaranteed that store is started and thus getStore result is not null.
+        store = getStore(replicaToAdd.getPartitionId(), false);
+
         // note that partitionNameToReplicaId should be updated if addBlobStore succeeds, so replicationManager should be
         // able to get new replica from storageManager without querying Helix
       } else {
-        // if the replica is already on current node, there are 3 cases need to discuss:
+        // if the replica is already on current node, there are 4 cases need to discuss:
         // 1. replica was initially present in clustermap and this is a regular reboot.
         // 2. replica was dynamically added to this node but may fail during BOOTSTRAP -> STANDBY transition.
         // 3. replica is on current node but its disk is offline. The replica is not able to start.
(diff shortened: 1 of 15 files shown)


ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -476,16 +477,40 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
...
+            if (!bootstrapFile.exists()) {
+              bootstrapFile.createNewFile();
+            }
+          } catch (IOException e) {
+            logger.error("Failed to create bootstrap file for store {}", partitionName);
+            throw new StateTransitionException("Failed to create bootstrap file for " + partitionName,
+                ReplicaOperationFailure);
+          }
+        }
+        store.setCurrentState(ReplicaState.BOOTSTRAP);
```
SophieGuo410, line 513: Sorry I didn't quite follow with this part. From the comments, the store will be forced in BOOTSTRAP state if capacity is less or equal to header size, but looks like no matter what the value of the storeUsedCapacity is, the store will be forced in BOOTSTRAP state. Is this intended?

justinlin-linkedin, line 513: creating a bootstrap_in_progress file and set setting currentstate to bootstrap has to go toe to toe, I suppose we can make a method out of them.

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -476,16 +477,40 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
         // able to get new replica from storageManager without querying Helix
       } else {
         // if the replica is already on current node, there are 3 cases need to discuss:
```
justinlin-linkedin, line 479: there are 4 cases need to discuss:

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -476,16 +477,40 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
...
+        Store store = getStore(replica.getPartitionId(), false);
+        if (store == null) {
           throw new StateTransitionException(
               "Store " + partitionName + " didn't start correctly, replica should be set to ERROR state",
               StoreNotStarted);
         }
+        // if store's used capacity is less than or equal to header size, we create a bootstrap_in_progress file and force
+        // it to stay in BOOTSTRAP state when catching up with peers.
+        long storeUsedCapacity = store.getSizeInBytes();
+        if (storeUsedCapacity <= HEADER_SIZE) {
```
justinlin-linkedin, line 498: I wonder if this is really necessary. If we create a bootstrap_in_progress file no matter the used capacity and let the remaining transition remove it later. In this way I think we can address another corner case where only half of the data in this partition gets wiped out.

Imagine that we have two log segment files in this partition, somehow we remove the second log segment file and then we restart the process. Here since the used capacity is no less then HEADER_SIZE, so the bootstrap_in_progress file will not be created, then the replicationmanager's callback would just mark it as standby on the onPartitionBecomeStandByFromBootstrap.

But if we create the bootstrap without checking the HEADER_SIZE, then we can avoid this issue.

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
jsjtzyy: Yes, it's intended to update store current state to BOOTSTRAP.  The difference is, if storeUsedCapacity is <= 18 (header size which means it's empty), we create a **bootstrap_in_progress** file in that store. Within `onPartitionBecomeStandbyFromBootstrap`, the replication manager will force stores that have **bootstrap_in_progress** file to stay in BOOTSTRAP before catchup completes. No actions will be taken on other stores (non-empty), they will quickly complete BOOTSTRAP -> STANDBY transition.

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
jsjtzyy: We can extend this to some other interesting edge cases, I will talk with you offline. For now, let's start with empty store first, that's the most common case when we wipe out whole partition or replace the disk.

jsjtzyy:
@justinlin-linkedin @SophieGuo410 @Arun-LinkedIn gentle reminder to review.

PR #1574, opened 2020-06-23
Mute ERROR message caused by replication of deprecated container and change log level

SophieGuo410:


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -741,6 +741,10 @@ private void processReplicaMetadataResponse(Set<MessageInfo> missingRemoteStoreM
       } else {
         // the key is present in the local store. Mark it for deletion if it is deleted in the remote store and not
         // deleted yet locally
+        // if the blob is from deprecated container, then nothing needs to be done.
+        if (skipPredicate != null && skipPredicate.test(messageInfo)) {
+          continue;
+        }
```
jsjtzyy, line 747: Is there a unit test for this?

SophieGuo410: Added more test cases to cover this.

ambry-replication/src/test/java/com/github/ambry/replication/ReplicationTest.java
```diff
@@ -108,6 +108,7 @@
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
+import org.mockito.Mock;
```
jsjtzyy, line 111: can be removed

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -741,6 +741,10 @@ private void processReplicaMetadataResponse(Set<MessageInfo> missingRemoteStoreM
       } else {
         // the key is present in the local store. Mark it for deletion if it is deleted in the remote store and not
         // deleted yet locally
+        // if the blob is from deprecated container, then nothing needs to be done.
+        if (skipPredicate != null && skipPredicate.test(messageInfo)) {
```
justinlin-linkedin, line 745: I still can't wrap my head around the exception findKey throws. How does it happen?
If we are here at line 745, then the local key is not missing from the local store. But if we move on to findKey, it throws an exception about not able to find it? Does this mean the key is compacted between?

ambry-replication/src/test/java/com/github/ambry/replication/ReplicationTest.java
```diff
@@ -1417,6 +1418,9 @@ public void blockDeprecatedContainerReplicationTest() throws Exception {
       }
       Set<MessageInfo> remoteMissingStoreKeys = replicaThread.getMissingStoreMessages(replicaMetadataResponseInfo, remoteNode, remoteReplicaInfo);
       assertEquals("All DELETE_IN_PROGRESS blobs qualified with retention time should be skipped during replication", 0, remoteMissingStoreKeys.size());
+      Map<StoreKey, StoreKey> remoteKeyToLocalKeyMap = replicaThread.batchConvertReplicaMetadataResponseKeys(response);
+      replicaThread.processReplicaMetadataResponse(remoteMissingStoreKeys, replicaMetadataResponseInfo, remoteReplicaInfo,
+          remoteNode, remoteKeyToLocalKeyMap);
```
jsjtzyy, line 1423: minor:  `checkSet` is never used, consider removing it?  (Also, format this file please)

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
SophieGuo410: I think the messageInfo is coming from replicaMetadataResponseInfo which stands for "the replica metadata response from the remote store". And during replication the deprecated containers won't be added in the missingRemoteStoreMessages list, so it will jump into line 741 no matter the local store has the key or not. And let's say the local store key has been compacted already, and for it's peer replica still have this key, then it will move on to findKey and throw exception. so in this situation, I use the skipPredicate to bypass the below procedure.

justinlin-linkedin: that makes sense.

PR #1577, opened 2020-06-24
Use DataNodeConfigSource in participant

cgtz:
Use DataNodeConfigSource as an abstraction around HelixAdmin and
InstanceConfigs in HelixParticipant to allow for future configurable
implementations to be plugged into this class.

For now, this only supports InstanceConfigToDataNodeConfigAdapter, but
will be modified to allow configuration in future commits.

This commit switches HelixParticipant to use a shared spectator
HelixManager instance for any config manipulation. This
avoids the need to construct a temporary HelixAdmin before the
"official" admin is ready.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
@@ -13,16 +13,23 @@
  */
 package com.github.ambry.clustermap;
 
-import org.apache.helix.HelixAdmin;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import org.apache.helix.HelixManager;
 import org.apache.helix.HelixManagerFactory;
 import org.apache.helix.InstanceType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * A factory class to construct and get a reference to a {@link HelixManager}
  */
 public class HelixFactory {
+  private static final Logger LOGGER = LoggerFactory.getLogger(HelixFactory.class);
+  Map<ManagerKey, HelixManager> helixManagers = new ConcurrentHashMap<>();
+
   /**
    * Get a reference to a {@link HelixManager}
    * @param clusterName the name of the cluster for which the manager is to be gotten.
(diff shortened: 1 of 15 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -279,135 +278,81 @@ public void setReplicaDisabledState(ReplicaId replicaId, boolean disable) {
...
+      dataNodeConfigUpdated = diskConfig.getReplicaConfigs().remove(partitionName) != null;
     }
-    if (instanceConfigUpdated) {
-      logger.info("Updating config: {} in Helix by removing partition {}", instanceConfig, partitionName);
-      removalResult = helixAdmin.setInstanceConfig(clusterName, instanceName, instanceConfig);
+    if (dataNodeConfigUpdated) {
+      logger.info("Updating config: {} in Helix by removing partition {}", dataNodeConfig, partitionName);
+      removalResult = dataNodeConfigSource.set(dataNodeConfig);
     } else {
       logger.warn("Partition {} is not found on instance {}, skipping removing it from InstanceConfig in Helix.",
```
jsjtzyy, line 357: minor: `skipping removing it from config in Helix."`

ambry-clustermap/src/test/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapterTest.java
```diff
@@ -109,10 +108,5 @@ public void testSetGetListener() throws Exception {
       assertEquals("get() call returned incorrect result", config, source.get(config.getInstanceName()));
     }
     assertNull("Should not receive non-existent instance", source.get("abc"));
-
-    source.setHelixAdmin(null);
-    TestUtils.assertException(NullPointerException.class, () -> source.set(createConfig(1, 1)), null);
-    String firstInstance = allConfigs.iterator().next().getInstanceName();
-    TestUtils.assertException(NullPointerException.class, () -> source.get(firstInstance), null);
   }
 }
```
jsjtzyy, line 112: minor: new line

ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixAdmin.java
```diff
@@ -398,6 +400,13 @@ long getTotalDiskCapacity() {
     return totalDiskCapacity;
   }
 
+  /**
+   * @return the number of calls to the {@link #setInstanceConfig} method.
+   */
+  public int getSetInstanceConfigCallCount() {
```
jsjtzyy, line 406: Package private? (If this method will be called in int-test, just ignore this comment)

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -33,15 +40,73 @@
...
+    HelixManager manager = getZKHelixManager(clusterName, instanceName, instanceType, zkAddr);
+    synchronized (manager) {
+      if (!manager.isConnected()) {
+        LOGGER.info("Connecting to HelixManager at {}", zkAddr);
+        manager.connect();
+        LOGGER.info("Established connection to HelixManager at {}", zkAddr);
+      } else {
+        LOGGER.info("HelixManager at {} already connected", zkAddr);
+      }
+    }
```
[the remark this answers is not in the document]
cgtz: There should not be such a case in production since we would never start `HelixClusterManager` and `HelixFactory` in parallel. I put this here since there is nothing that strictly guarantees that will always be the case. If you feel that this is unneeded for now I could take it out.

We could move this into a Map.compute lambda, but I felt that this approach allowed for finer grained locking as we don't need the construction and connection to be atomic, we just need to make sure that after construction and adding to the map, only one thread does the check connected, connect sequence at a time.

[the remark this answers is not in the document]
jsjtzyy: I am okay with this approach after more clarification. Let's keep it as it is.

PR #1579, opened 2020-06-25
Change java ByteBuffer to netty ByteBuf in request and response.

justinlin-linkedin:
Starting use netty ByteBuf to replace java.nio.ByteBuffer in the network layer.
There are some issues with GetResponse, since it supports non-prefetched data from disk. Something need to be done in order to actually support ByteBufHolder for GetResponse.

After this PR, Request and Response would have a method "content", if the returned value from method content is not null, then caller can use the returned ByteBuf as serialized bytes for the request and response.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/network/Send.java
+++ b/ambry-api/src/main/java/com/github/ambry/network/Send.java
@@ -15,6 +15,8 @@ package com.github.ambry.network;
 
 import com.github.ambry.router.AsyncWritableChannel;
 import com.github.ambry.router.Callback;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufHolder;
 import java.io.IOException;
 import java.nio.channels.WritableByteChannel;
 
@@ -23,7 +25,7 @@ import java.nio.channels.WritableByteChannel;
  * Any data that needs to be sent over the network can implement
  * this interface
  */
-public interface Send {
+public interface Send extends ByteBufHolder {
   /**
    * Writes content into the provided channel
    * @param channel The channel into which data needs to be written to
@@ -52,7 +54,9 @@ public interface Send {
   long sizeInBytes();
 
   /**
-   * Release all the resource this object holds. Make this a default method so subclasses don't have to override it.
+   * Return the data which is held by this {@link ByteBufHolder}. The result could be null, if it's null,
+   * please use writeTo methods to write the content to remote peer.
    */
-  default void release() {}
+  @Override
+  ByteBuf content();
 }
--- a/ambry-network/src/main/java/com/github/ambry/network/BlockingChannel.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/BlockingChannel.java
@@ -165,11 +165,15 @@ public class BlockingChannel implements ConnectedChannel {
 
   @Override
   public void send(Send request) throws IOException {
-    if (!connected) {
-      throw new ClosedChannelException();
-    }
-    while (!request.isSendComplete()) {
-      request.writeTo(writeChannel);
+    try {
+      if (!connected) {
+        throw new ClosedChannelException();
+      }
+      while (!request.isSendComplete()) {
+        request.writeTo(writeChannel);
+      }
+    } finally {
+      request.release();
     }
   }
 
--- a/ambry-network/src/main/java/com/github/ambry/network/SocketNetworkClient.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/SocketNetworkClient.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.ListIterator;
@@ -128,9 +129,6 @@ public class SocketNetworkClient implements NetworkClient {
     } catch (Exception e) {
       logger.error("Received an unexpected error during sendAndPoll(): ", e);
       networkMetrics.networkClientException.inc();
-      if (sends != null) {
-        sends.forEach(send -> send.getPayload().release());
-      }
     } finally {
       numPendingRequests.set(pendingRequests.size());
       networkMetrics.networkClientSendAndPollTime.update(time.milliseconds() - startTime, TimeUnit.MILLISECONDS);
@@ -158,13 +156,19 @@ public class SocketNetworkClient implements NetworkClient {
       if (time.milliseconds() - requestMetadata.requestQueuedAtMs > checkoutTimeoutMs) {
         responseInfoList.add(
             new ResponseInfo(requestMetadata.requestInfo, NetworkClientErrorCode.ConnectionUnavailable, null));
+        requestMetadata.requestInfo.getRequest().release();
         logger.trace("Failing request to host {} port {} due to connection unavailability",
             requestMetadata.requestInfo.getHost(), requestMetadata.requestInfo.getPort());
         iter.remove();
-        requestMetadata.requestInfo.getRequest().release();
         if (requestMetadata.pendingConnectionId != null) {
           pendingConnectionsToAssociatedRequests.remove(requestMetadata.pendingConnectionId);
           requestMetadata.pendingConnectionId = null;
+        }
+        String connId =
+            correlationIdInFlightToConnectionId.get(requestMetadata.requestInfo.getRequest().getCorrelationId());
+        if (connId != null) {
+          connectionTracker.checkInConnection(connId);
+          connectionIdToRequestInFlight.remove(connId);
         }
         networkMetrics.connectionCheckoutTimeoutError.inc();
       } else {
@@ -186,9 +190,9 @@ public class SocketNetworkClient implements NetworkClient {
           throw new IllegalStateException("ReplicaId in request is null.");
         }
         if (requestsToDrop.contains(requestMetadata.requestInfo.getRequest().getCorrelationId())) {
-          requestMetadata.requestInfo.getRequest().release();
           responseInfoList.add(
               new ResponseInfo(requestMetadata.requestInfo, NetworkClientErrorCode.ConnectionUnavailable, null));
+          requestMetadata.requestInfo.getRequest().release();
           if (requestMetadata.pendingConnectionId != null) {
             pendingConnectionsToAssociatedRequests.remove(requestMetadata.pendingConnectionId);
             requestMetadata.pendingConnectionId = null;
@@ -223,7 +227,6 @@ public class SocketNetworkClient implements NetworkClient {
           }
         }
       } catch (IOException e) {
-        requestMetadata.requestInfo.getRequest().release();
         networkMetrics.networkClientIOError.inc();
         logger.error("Received exception while checking out a connection", e);
       }
@@ -358,6 +361,8 @@ public class SocketNetworkClient implements NetworkClient {
       networkMetrics.connectionDisconnected.inc();
     }
 
+    // We don't need to release the completed sends' resources here, since they are already released by selector.
+
     for (NetworkReceive recv : selector.completedReceives()) {
       String connId = recv.getConnectionId();
       logger.trace("Receive completed for connectionId {} and checking in the connection back to connection tracker",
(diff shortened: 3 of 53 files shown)


ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
```diff
@@ -175,93 +179,95 @@ private int sizeExcludingBlobAndCrcSize() {
...
+   * Override release even if the {@link #content()}'s result is not null. When the {@link #prepareBuffer()} is
+   * not invoked, there will be {@link #blob} and {@link #crcByteBuf} created but not returned by {@link #content()},
+   * and we have to release them.
+   * @return
+   */
   @Override
-  public void release() {
+  public boolean release() {
     if (blob != null) {
       blob.release();
```
zzmao, line 260: Http2NetworkClient throws an exception

```
release
[2020-06-26 14:16:03,876] WARN An exception was thrown by com.github.ambry.network.http2.Http2NetworkClient$1.operationComplete() (io.netty.util.concurrent.DefaultPromise)
io.netty.util.IllegalReferenceCountException: refCnt: 0, decrement: 1
	at io.netty.util.internal.ReferenceCountUpdater.toLiveRealRefCnt(ReferenceCountUpdater.java:74)
	at io.netty.util.internal.ReferenceCountUpdater.release(ReferenceCountUpdater.java:138)
	at io.netty.buffer.AbstractReferenceCountedByteBuf.release(AbstractReferenceCountedByteBuf.java:100)
	at com.github.ambry.protocol.PutRequest.release(PutRequest.java:269)
```

Netty write decrease the ref cnt, but it doesn't set the variable to null. So we hit above exception.

justinlin-linkedin: I purposely didn't set the content to null so we don't release it multiple times. Where does it get released before? Can we figure that out?

ambry-protocol/src/main/java/com/github/ambry/protocol/DeleteResponse.java
```diff
@@ -15,6 +15,7 @@
 
 import com.github.ambry.server.ServerErrorCode;
 import com.github.ambry.utils.Utils;
+import io.netty.buffer.ByteBuf;
```
cgtz, line 18: Remove this import?

ambry-network/src/main/java/com/github/ambry/network/SocketNetworkClient.java
```diff
@@ -158,14 +156,19 @@ public SocketNetworkClient(Selector selector, NetworkConfig networkConfig, Netwo
...
         iter.remove();
-        requestMetadata.requestInfo.getRequest().release();
         if (requestMetadata.pendingConnectionId != null) {
           pendingConnectionsToAssociatedRequests.remove(requestMetadata.pendingConnectionId);
           requestMetadata.pendingConnectionId = null;
         }
+        String connId =
+            correlationIdInFlightToConnectionId.get(requestMetadata.requestInfo.getRequest().getCorrelationId());
+        if (connId != null) {
+          connectionIdToRequestInFlight.remove(connId);
```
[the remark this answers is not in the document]
justinlin-linkedin: I suppose we should checkin the connection here. There is a chance that when a request is associated with a connection and somehow it takes so long to send it out that this request just times out. In this case, we are releasing the request's bytebuf here in this line 159, so we have to make sure the request is removed from all bookkeeping maps and sets.

ambry-api/src/main/java/com/github/ambry/network/Send.java
```diff
@@ -52,7 +54,9 @@ default void writeTo(AsyncWritableChannel channel, Callback<Long> callback) {
   long sizeInBytes();
 
   /**
-   * Release all the resource this object holds. Make this a default method so subclasses don't have to override it.
+   * Return the data which is held by this {@link ByteBufHolder}. The result could be null, if it's null,
```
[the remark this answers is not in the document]
justinlin-linkedin: Eventually we will support it. All the MessageFormatSend's bytes are ready in memory, I don't see a reason why not to support it in the future.

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
```diff
@@ -175,93 +179,95 @@ private int sizeExcludingBlobAndCrcSize() {
...
       crc.update(bb);
       // change it back to 0 since we are going to write it to the channel later.
       bb.position(0);
     }
-    crcBuf.putLong(crc.getValue());
-    crcBuf.flip();
-    bufferToSend.flip();
+    crcByteBuf.writeLong(crc.getValue());
+
+    CompositeByteBuf compositeByteBuf = bufferToSend.alloc().compositeHeapBuffer(2 + blob.nioBufferCount());
```
zzmao, line 206: Why explicitly use `compositeHeapBuffer`? I thought blob are direct buffers, no?

zzmao: `bufferToSend.alloc()` -> `PooledByteBufAllocator.DEFAULT`? 
Feel we can use PooledByteBufAllocator.DEFAULT almost everywhere in the first few versions.

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
```diff
@@ -175,93 +179,95 @@ private int sizeExcludingBlobAndCrcSize() {
...
+      while (iter.hasNext()) {
+        compositeByteBuf.addComponent(true, iter.next());
+      }
+    } else {
+      compositeByteBuf.addComponent(true, blob);
+    }
+    compositeByteBuf.addComponent(true, crcByteBuf);
+    blob = null;
+    crcByteBuf = null;
+    bufferToSend = compositeByteBuf;
```
zzmao, line 219: The implementation overuses bufferToSend and caused some confusion when I looked at the code. 
How about just set 
bufferToSend = PooledByteBufAllocator.DEFAULT.compositBuffer() in the beginning, 
then allocate temp buffer for headers and write blob to bufferToSend,
finally return bufferToSend?

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
zzmao: In Http2NetworkClient::sendAndPoll()
```
                  // release related bytebuf
                  requestInfo.getRequest().release();
```
This is called in the success case, but with your PR, this no longer needed.

ambry-protocol/src/main/java/com/github/ambry/protocol/GetResponse.java
```diff
@@ -114,46 +122,111 @@ public static GetResponse readFrom(DataInputStream stream, ClusterMap map) throw
    * {@link GetResponse#writeTo(AsyncWritableChannel, Callback)}.
    * This method allocate bufferToSend and write metadata to it if bufferToSend is null.
    */
-  private void prepareBufferToSend() {
-    if (bufferToSend == null) {
-      bufferToSend = ByteBuffer.allocate(
+  @Override
+  protected void prepareBuffer() {
+    if (toSend != null && toSend.content() != null) {
+      bufferToSend = PooledByteBufAllocator.DEFAULT.heapBuffer((int) sizeInBytes());
```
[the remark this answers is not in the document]
justinlin-linkedin: My thought on the heap buffer is that we started to manage the memory and it shouldn't be too bad to use heap memory for everything. I suppose we can start using direct memory for it.

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
justinlin-linkedin: sure

ambry-protocol/src/main/java/com/github/ambry/protocol/GetResponse.java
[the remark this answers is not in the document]
zzmao: I remember direct memory is friendly for network data.

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
justinlin-linkedin: here we at least have three separate bytebuf, first one is header bytebuf, second is blob content bytebuf, the third is the crc bytebuf. We don't want to create a new bytebuf and copy all three bytebufs to the newly created one. So here we create a compositebytebuf and just add the existing bytebufs to the the compositebytebuf to avoid extra copying.

ambry-network/src/main/java/com/github/ambry/network/BlockingChannel.java
```diff
@@ -171,6 +171,7 @@ public void send(Send request) throws IOException {
     while (!request.isSendComplete()) {
       request.writeTo(writeChannel);
     }
+    request.release();
```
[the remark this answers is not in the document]
justinlin-linkedin: you are right, request should be released in all cases.

ambry-protocol/src/main/java/com/github/ambry/protocol/GetResponse.java
```diff
@@ -114,46 +122,111 @@ public static GetResponse readFrom(DataInputStream stream, ClusterMap map) throw
...
-  private void prepareBufferToSend() {
-    if (bufferToSend == null) {
-      bufferToSend = ByteBuffer.allocate(
+  @Override
+  protected void prepareBuffer() {
+    if (toSend != null && toSend.content() != null) {
+      bufferToSend = PooledByteBufAllocator.DEFAULT.ioBuffer((int) sizeInBytes());
+    } else {
+      bufferToSend = PooledByteBufAllocator.DEFAULT.ioBuffer(
           (int) super.sizeInBytes() + (Partition_Response_Info_List_Size + partitionResponseInfoSize));
```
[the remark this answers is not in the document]
justinlin-linkedin: I made a mistake, bufferToSend should never allocate `sizeInBytes()` bytes.
GetResponse is made of two parts, the metadata part and the data part.
`bufferToSend` should only has metadata part, since the data is carried by `toSend`.
if `toSend` doesn't support ByteBufHolder, then when writing GetResponse to network, we write `bufferToSend` first, and then `toSend` later. 
if `toSend` supports ByteBufHolder, then bufferToSend will be `CompositeByteBuf` later. This `CompositeByteBuf` contains metadata and data.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -107,6 +107,7 @@ public void operationComplete(ChannelFuture future) throws Exception {
                     http2ClientMetrics.http2StreamWriteAndFlushTime.update(
                         System.currentTimeMillis() - streamAcquiredTime);
                     requestInfo.setStreamSendTime(System.currentTimeMillis());
+                    // When success, handler would release the request.
```
zzmao, line 110: Please help add a counter metric(channelNotWriteableCounter) before ```streamChannel.writeAndFlush()```  .

if (!streamChannel.isWriteable() || !streamChannel.parentChannel().isWriteable() ) {
     channelNotWriteableCounter.inc()
}

zzmao: also a warn message.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -98,6 +98,10 @@ public Http2NetworkClient(Http2ClientMetrics http2ClientMetrics, Http2ClientConf
               long streamAcquiredTime = System.currentTimeMillis();
               Channel streamChannel = future.getNow();
               streamChannel.attr(REQUEST_INFO).set(requestInfo);
+              if (!streamChannel.isWritable() || !streamChannel.parent().isWritable()) {
+                http2ClientMetrics.http2StreamNotWritableCount.inc();
+                logger.warn("Stream {} {} not writable", streamChannel.hashCode(), streamChannel);
```
zzmao, line 103:
```
logger.warn("Stream {} {} not writable. BytesBeforeWritable {} {}", streamChannel.hashCode(), streamChannel, streamChannel().bytesBeforeWritable(), streamChannel.parent().bytesBeforeWritable());

```

ambry-network/src/main/java/com/github/ambry/network/SocketNetworkClient.java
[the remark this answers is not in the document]
cgtz: I guess this piece of code doesn't hurt. But I still don't see how `correlationIdInFlightToConnectionId.get` will ever return non-null at this point. This part of the code handles requests that do not yet have a connection to make the request over, and that map is only added to after a connection is checked out. 

One confusing part of `SocketNetworkClient` is how the pending connection ID works. The pending connection ID is not actually checked out yet but is an indicator to show that a new connection is in the process of being established and will eventually be added to the pool but weirdly it doesn't guarantee that that specific connection is actually used for the request it was associated with.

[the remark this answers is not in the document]
justinlin-linkedin: I think we have a disagreement on whether a request will timeout after it gets associated with a connection id. I certainly think it's possible that even after a request has checked out a connection, it can still time out, due to slow outbound link, gc or some other reasons. 

In MockSelector of unit tests, we do have a case where the MockSelector would throw IOException when sending out a request without disconnecting the connection. SocketNetworkClient won't remove this request from pending requests when seeing an IOException. So I take that we will retry this request, on the same connection, until it times out.

PR #1584, opened 2020-07-08
Avoid concurrent modification on InstanceConfig when disabling partitions

jsjtzyy:
Since both Ambry and Helix can update InstanceConfig, there is a race condition where ambry server is removing replica entry from InstanceConfig(move replica case) and meanwhile Helix is adding the disabled partition into same InstanceConfig. It's possible that ambry server updates InstanceConfig with an old version view, which may override the latest change
from Helix. This usually happens when we use HelixBootstrapTool to remove multiple replicas and it may take some time to complete. Before it's done, some replicas have finished decommission and start to update InstanceConfig, which causes concurrent modification.

This PR ensures HelixBbootstrapTool creates a ZNode for specific server only when disabling partitions
on that node has fully completed. The decommission process of replicas will be blocked until
the ZNode is generated. Thus, concurrent modification can be avoided.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
@@ -108,6 +108,9 @@ public class HelixBootstrapUpgradeUtil {
   static final int DEFAULT_MAX_PARTITIONS_PER_RESOURCE = 100;
   static final String HELIX_DISABLED_PARTITION_STR =
       InstanceConfig.InstanceConfigProperty.HELIX_DISABLED_PARTITION.name();
+  // blockRemovingNodeLatch is for testing purpose
+  static CountDownLatch blockRemovingNodeLatch = null;
+  static CountDownLatch disablePartitionLatch = null;
   private final StaticClusterManager staticClusterMap;
   private final String zkLayoutPath;
   private final String stateModelDef;
@@ -687,6 +690,7 @@ public class HelixBootstrapUpgradeUtil {
       if (!helixPropertyStore.set(adminConfigZNodePath, znRecord, AccessOption.PERSISTENT)) {
         logger.error("Failed to upload {} infos for datacenter {}", clusterAdminType, entry.getKey());
       }
+      helixPropertyStore.close();
     }
   }
 
@@ -706,6 +710,7 @@ public class HelixBootstrapUpgradeUtil {
       if (!helixPropertyStore.remove(adminConfigZNodePath, AccessOption.PERSISTENT)) {
         logger.error("Failed to remove {} infos from datacenter {}", clusterAdminType, entry.getKey());
       }
+      helixPropertyStore.close();
     }
   }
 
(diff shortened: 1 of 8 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -1072,6 +1074,24 @@ private void addUpdateResources(String dcName, Map<String, Set<String>> partitio
...
     }
+    // if there are some partitions are disabled, mark whole disabling process complete in Helix PropertyStore
+    if (helixAdminOperation == HelixAdminOperation.DisablePartition && !instancesWithDisabledPartition.isEmpty()) {
+      // create znode under admin configs
+      Properties storeProps = new Properties();
+      storeProps.setProperty("helix.property.store.root.path", "/" + clusterName + "/" + PROPERTYSTORE_STR);
+      HelixPropertyStoreConfig propertyStoreConfig = new HelixPropertyStoreConfig(new VerifiableProperties(storeProps));
+      String zkConnectStr = dataCenterToZkAddress.get(dcName).getZkConnectStrs().get(0);
+      HelixPropertyStore<ZNRecord> helixPropertyStore =
+          CommonUtils.createHelixPropertyStore(zkConnectStr, propertyStoreConfig, null);
```
cgtz, line 1085: Do we have to call the helixPropertyStore.stop() and close() methods after we are done using the property store? Same comment for its usage in HelixParticipant.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -415,6 +437,27 @@ private boolean removeOldReplicaInfo(ReplicaId replicaId, InstanceConfig instanc
...
+   */
+  private void awaitDisablingPartition() throws InterruptedException {
+    Properties properties = new Properties();
+    properties.setProperty("helix.property.store.root.path", "/" + clusterName + "/" + PROPERTYSTORE_STR);
+    HelixPropertyStoreConfig propertyStoreConfig = new HelixPropertyStoreConfig(new VerifiableProperties(properties));
+    HelixPropertyStore<ZNRecord> helixPropertyStore =
+        CommonUtils.createHelixPropertyStore(zkConnectStr, propertyStoreConfig, null);
+    String path = PARTITION_DISABLED_ZNODE_PATH + instanceName;
+    int count = 1;
+    while (!helixPropertyStore.exists(path, AccessOption.PERSISTENT)) {
```
cgtz, line 454: Do these paths have to be manually removed using a zookeeper client after the decommission is complete and all replicas have called removeOldReplicaInfo()?

What if, instead, `HelixBootStrapUpgradeTool` created a znode before it begins updating using `HelixAdmin` and then deletes it after it is finished? Than, this loop could wait for the znode to stop existing.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
jsjtzyy: Good point. Since this is one-off change, we should close it after updating the property store in Helix.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -451,11 +451,12 @@ private void awaitDisablingPartition() throws InterruptedException {
         CommonUtils.createHelixPropertyStore(zkConnectStr, propertyStoreConfig, null);
     String path = PARTITION_DISABLED_ZNODE_PATH + instanceName;
     int count = 1;
-    while (!helixPropertyStore.exists(path, AccessOption.PERSISTENT)) {
+    while (helixPropertyStore.exists(path, AccessOption.PERSISTENT)) {
       // Thread.sleep() pauses the current thread but does not release any locks
       Thread.sleep(clusterMapConfig.clustermapRetryDisablePartitionCompletionBackoffMs);
       logger.info("{} th attempt on checking the completion of disabling partition.", ++count);
     }
+    helixPropertyStore.close();
```
cgtz, line 459: one weird thing I noticed was that the stop() method in ZkCacheBaseDataAccessor also stops the background event handling thread in addition to shutting down the zkClient like in the close() method. We may want to call stop() instead

jsjtzyy: sure, let me update the method.

PR #1585, opened 2020-07-11
Remove hardcoded MAX_DISK_CAPACITY_IN_BYTES and make it configurable.

ankagrawal:
With max disk capacity being hardcoded to 10TB currently, it's not possible to use disks with size greater than 10TB. This PR makes the max disk size configurable to allow for large disks.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
@@ -267,6 +267,9 @@ public class ClusterMapConfig {
   @Default("{}")
   public final JSONObject clustermapRecoveryTestHardwareLayout;
 
+  /**
+   * Max capacity of a disk in bytes.
+   */
   @Config(CLUSTERMAP_MAX_DISK_CAPACITY_IN_BYTES)
   public final long clustermapMaxDiskCapacityInBytes;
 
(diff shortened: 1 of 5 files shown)


ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
```diff
@@ -266,6 +267,9 @@
   @Default("{}")
   public final JSONObject clustermapRecoveryTestHardwareLayout;
 
+  @Config(CLUSTERMAP_MAX_DISK_CAPACITY_IN_BYTES)
```
jsjtzyy, line 270: minor: java doc

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
```diff
@@ -320,5 +324,7 @@ public ClusterMapConfig(VerifiableProperties verifiableProperties) {
         new JSONObject(verifiableProperties.getString("clustermap.recovery.test.hardware.layout", "{}"));
     clustermapRecoveryTestPartitionLayout =
         new JSONObject(verifiableProperties.getString("clustermap.recovery.test.partition.layout", "{}"));
+    clustermapMaxDiskCapacityInBytes =
+        verifiableProperties.getLong(CLUSTERMAP_MAX_DISK_CAPACITY_IN_BYTES, 10L * 1024 * 1024 * 1024 * 1024);
```
jsjtzyy, line 328: Could you change this to `getLongInRange`?  The parameters are:  min = 0, max = Long.MAX_VALUE, default = 20L * 1024 * 1024 * 1024 * 1024 (let's use 20TB as default value)

ankagrawal: done.

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
ankagrawal: done.

PR #1586, opened 2020-07-13
Optimize number of calls to CloudBlobStore::findKey()

ankagrawal:
`CloudBlobStore::findKey()` is called multiple times in `ReplicaThread::processReplicaMetadataResponse` inside the `ReplicaThread::replicate()` loop. Since `CloudBlobStore::findKey()` internally calls cosmos db, which is an expensive call, this change batches all the calls to `findKey()` at one place. This will minimize the calls to cosmos db, while keeping the number of `findKey()` calls in case of `BlobStore` same as before.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/store/Store.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/Store.java
@@ -96,14 +96,6 @@ public interface Store {
   Set<StoreKey> findMissingKeys(List<StoreKey> keys) throws StoreException;
 
   /**
-   * Return {@link MessageInfo} of given key. This method will only be used in replication thread.
-   * @param key The key of which blob to return {@link MessageInfo}.
-   * @return The {@link MessageInfo}.
-   * @throws StoreException
-   */
-  MessageInfo findKey(StoreKey key) throws StoreException;
-
-  /**
    * Return a {@link Map} of {@code storeKeys} to {@link MessageInfo} of the keys.
    * This method will only be used in replication thread, and the map returned will contain message infos for only those
    * keys that are found in the local store.
--- a/ambry-api/src/main/java/com/github/ambry/store/Store.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/Store.java
@@ -17,6 +17,7 @@ import com.github.ambry.clustermap.ReplicaState;
 import com.github.ambry.replication.FindToken;
 import java.util.EnumSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 
@@ -101,6 +102,16 @@ public interface Store {
    * @throws StoreException
    */
   MessageInfo findKey(StoreKey key) throws StoreException;
+
+  /**
+   * Return a {@link Map} of {@code storeKeys} to {@link MessageInfo} of the keys.
+   * This method will only be used in replication thread, and the map returned will contain message infos for only those
+   * keys that are found in the local store.
+   * @param storeKeys The list of keys for which to return {@link MessageInfo}s.
+   * @return The {@link Map} of key to {@link MessageInfo}.
+   * @throws StoreException
+   */
+  Map<StoreKey, MessageInfo> findKeys(List<? extends StoreKey> storeKeys) throws StoreException;
 
   /**
    * Get the corresponding {@link StoreStats} instance for this store.
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -890,6 +890,7 @@ public class ReplicaThread implements Runnable {
           // all the keys being processed are missing in local store.
           localMessageInfoMap = remoteReplicaInfo.getLocalStore()
               .findKeys(messageInfoList.stream()
+                  .filter(msgInfo -> !missingRemoteStoreMessages.contains(msgInfo))
                   .map(msgInfo -> (BlobId) remoteKeyToLocalKeyMap.get(msgInfo.getStoreKey()))
                   .filter(key -> key != null)
                   .collect(Collectors.toList()));
@@ -917,7 +918,8 @@ public class ReplicaThread implements Runnable {
   public void applyUpdatesToBlobInLocalStore(MessageInfo messageInfo, RemoteReplicaInfo remoteReplicaInfo,
       BlobId localKey, MessageInfo localMessageInfo) throws StoreException {
     if (localMessageInfo == null) {
-      throw new StoreException("Key " + localKey + " not found in store.", StoreErrorCodes.ID_Not_Found);
+      throw new StoreException(String.format("Key %s not found in store.", localKey.getID()),
+          StoreErrorCodes.ID_Not_Found);
     }
     boolean deletedLocally = localMessageInfo.isDeleted();
     boolean ttlUpdatedLocally = localMessageInfo.isTtlUpdated();
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -885,7 +885,7 @@ public class ReplicaThread implements Runnable {
             messageInfo)) {
           continue;
         }
-        if(localMessageInfoMap == null) {
+        if (localMessageInfoMap == null) {
           // This makes sure that the potentially expensive findKeys() method isn't called for a replication loop, if
           // all the keys being processed are missing in local store.
           localMessageInfoMap = remoteReplicaInfo.getLocalStore()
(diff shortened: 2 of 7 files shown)


ambry-api/src/main/java/com/github/ambry/store/Store.java
```diff
@@ -102,6 +103,16 @@
...
 
+  /**
+   * Return a {@link Map} of {@code storeKeys} to {@link MessageInfo} of the keys.
+   * This method will only be used in replication thread, and the map returned will contain message infos for only those
+   * keys that are found in the local store.
+   * @param storeKeys The list of keys for which to return {@link MessageInfo}s.
+   * @return The {@link Map} of key to {@link MessageInfo}.
+   * @throws StoreException
+   */
+  Map<StoreKey, MessageInfo> findKeys(List<? extends StoreKey> storeKeys) throws StoreException;
```
ankagrawal, line 114: If this method and approach is acceptable, we no longer need `findKey()` to be a public method, and hence it can be removed from the interface.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -934,6 +934,19 @@ public MessageInfo findKey(StoreKey key) throws StoreException {
     }
   }
 
+  @Override
+  public Map<StoreKey, MessageInfo> findKeys(List<? extends StoreKey> storeKeys) {
+    Map<StoreKey, MessageInfo> messageInfoMap = new HashMap<>();
```
ankagrawal, line 939: TODO: Need to look into the possibility if its more efficient to not call `findKey()` multiple times in a loop, and instead search the index once for all the blobs in the list. Will do so in a separate ticket. For now,  calling `findKey()` multiple times in a loop is no worse than what was happening before for the case of `BlobStore`.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -880,10 +881,20 @@ void processReplicaMetadataResponse(Set<MessageInfo> missingRemoteStoreMessages,
...
+            messageInfo)) {
           continue;
         }
-        applyUpdatesToBlobInLocalStore(messageInfo, remoteReplicaInfo, localKey);
+        if(localMessageInfoMap == null) {
+          // This makes sure that the potentially expensive findKeys() method isn't called for a replication loop, if
+          // all the keys being processed are missing in local store.
+          localMessageInfoMap = remoteReplicaInfo.getLocalStore()
+              .findKeys(messageInfoList.stream()
+                  .map(msgInfo -> (BlobId) remoteKeyToLocalKeyMap.get(msgInfo.getStoreKey()))
```
lightningrob, line 893: Formatting looks off.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -900,11 +911,14 @@ void processReplicaMetadataResponse(Set<MessageInfo> missingRemoteStoreMessages,
...
    * @param remoteReplicaInfo remote replica information
    * @param localKey local blob information
+   * @param localMessageInfo message information of the blob from the local replica
    * @throws StoreException
    */
   public void applyUpdatesToBlobInLocalStore(MessageInfo messageInfo, RemoteReplicaInfo remoteReplicaInfo,
-      BlobId localKey) throws StoreException {
-    MessageInfo localMessageInfo = remoteReplicaInfo.getLocalStore().findKey(localKey);
+      BlobId localKey, MessageInfo localMessageInfo) throws StoreException {
+    if (localMessageInfo == null) {
```
lightningrob, line 919: I would move this check to the calling method, and just do Objects.assertNonNull() here.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -880,10 +881,20 @@ void processReplicaMetadataResponse(Set<MessageInfo> missingRemoteStoreMessages,
         // it is deleted in the remote store and not deleted yet locally.
 
         // if the blob is from deprecated container, then nothing needs to be done.
-        if (replicationConfig.replicationContainerDeletionEnabled && skipPredicate != null && skipPredicate.test(messageInfo)) {
+        if (replicationConfig.replicationContainerDeletionEnabled && skipPredicate != null && skipPredicate.test(
+            messageInfo)) {
           continue;
         }
-        applyUpdatesToBlobInLocalStore(messageInfo, remoteReplicaInfo, localKey);
+        if(localMessageInfoMap == null) {
```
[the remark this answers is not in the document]
lightningrob: Also, is there a reason to call this inside the for loop rather than up front?

ambry-api/src/main/java/com/github/ambry/store/Store.java
lightningrob: Yes, I would remove it.

ankagrawal: done.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -934,6 +939,21 @@ public MessageInfo findKey(StoreKey key) throws StoreException {
...
+    for (StoreKey storeKey : storeKeys) {
+      try {
+        // TODO: Need to look into the possibility if its more efficient to not call findKey() multiple times in a loop,
+        // and instead search the index once for all the blobs in the list.
+        messageInfoMap.put(storeKey, findKey(storeKey));
+      } catch (StoreException stEx) {
+        logger.info("Key %s not found in findKey", storeKey.getID());
```
justinlin-linkedin, line 951: not all exception means not_found from findKey, please print out the exception as well.
Also, we should use error here.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -880,10 +881,20 @@ void processReplicaMetadataResponse(Set<MessageInfo> missingRemoteStoreMessages,
...
+            messageInfo)) {
           continue;
         }
-        applyUpdatesToBlobInLocalStore(messageInfo, remoteReplicaInfo, localKey);
+        if (localMessageInfoMap == null) {
+          // This makes sure that the potentially expensive findKeys() method isn't called for a replication loop, if
+          // all the keys being processed are missing in local store.
+          localMessageInfoMap = remoteReplicaInfo.getLocalStore()
+              .findKeys(messageInfoList.stream()
+                  .map(msgInfo -> (BlobId) remoteKeyToLocalKeyMap.get(msgInfo.getStoreKey()))
```
justinlin-linkedin, line 893: we can filter out those keys we already know that are now now in local store.
```
messageInfoList.stream().filter(messageInfo -> !missingRemoteStoreMessage.contains(messageInfo)).
```

ankagrawal: done.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
ankagrawal: True. I have changed the code to throw the errors from `findKey()` as is, after logging.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -940,15 +940,15 @@ public MessageInfo findKey(StoreKey key) throws StoreException {
   }
 
   @Override
-  public Map<StoreKey, MessageInfo> findKeys(List<? extends StoreKey> storeKeys) {
+  public Map<StoreKey, MessageInfo> findKeys(List<? extends StoreKey> storeKeys) throws StoreException {
```
justinlin-linkedin, line 943: This method doesn't throw out StoreException.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -949,6 +949,7 @@ public MessageInfo findKey(StoreKey key) throws StoreException {
         messageInfoMap.put(storeKey, findKey(storeKey));
       } catch (StoreException e) {
         logger.error("findKey failed for key {} with reason {}", storeKey.getID(), e.getErrorCode().toString());
+        throw e;
```
justinlin-linkedin, line 952: if we are throwing exception here, then we don't need the try-catch statement.

ankagrawal: The reason for try-catch here is because I want to log a meaningful message with the key that resulted in exception. The exception shouldn't be thrown often here, hence I thought it might be a good idea to log it here to help with quick debugging, and it will not result in any performance issues.

justinlin-linkedin: I am good with that.

PR #1587, opened 2020-07-13
Use ByteBuf in decrypt job

justinlin-linkedin:
Before this PR, in GetBlobOperation, we use java ByteBuffer as the byte container to store decrypted blob content.
This PR switches to netty ByteBuf.
It also switches to use ioBuffer (direct memory) for encryption and decryption.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GCMCryptoService.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GCMCryptoService.java
@@ -124,7 +124,6 @@ public class GCMCryptoService implements CryptoService<SecretKeySpec> {
       encrypter.init(Cipher.ENCRYPT_MODE, key, new IvParameterSpec(iv));
       int outputSize = encrypter.getOutputSize(toEncrypt.readableBytes());
 
-      // stick with heap memory for now so to compare with the java.nio.ByteBuffer.
       encryptedContent = PooledByteBufAllocator.DEFAULT.ioBuffer(IVRecord_Format_V1.getIVRecordSize(iv) + outputSize);
       IVRecord_Format_V1.serializeIVRecord(encryptedContent, iv);
 
--- a/ambry-router/src/main/java/com/github/ambry/router/GCMCryptoService.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GCMCryptoService.java
@@ -19,8 +19,8 @@ import com.github.ambry.messageformat.MessageFormatException;
 import com.github.ambry.utils.ByteBufferInputStream;
 import com.github.ambry.utils.Utils;
 import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufAllocator;
 import io.netty.buffer.ByteBufInputStream;
+import io.netty.buffer.PooledByteBufAllocator;
 import java.io.DataInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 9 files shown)


ambry-router/src/main/java/com/github/ambry/router/GCMCryptoService.java
```diff
@@ -123,33 +125,32 @@ public ByteBuf encrypt(ByteBuf toEncrypt, SecretKeySpec key, byte[] iv) throws G
       int outputSize = encrypter.getOutputSize(toEncrypt.readableBytes());
 
       // stick with heap memory for now so to compare with the java.nio.ByteBuffer.
```
zzmao, line 127: No longer valid.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1353,12 +1356,19 @@ private void handleMetadataBlob(BlobData blobData, byte[] userMetadata, ByteBuff
...
+                      result.getDecryptedBlobContent().release();
+                    }
+                  } else {
+                    decryptJobMetricsTracker.onJobCallbackProcessingStart();
+                    logger.trace("Handling decrypt job call back for Metadata chunk {} to set decrypt callback results",
+                        blobId);
+                    decryptCallbackResultInfo.setResultAndException(result, exception);
+                    routerCallback.onPollReady();
+                    decryptJobMetricsTracker.onJobCallbackProcessingComplete();
+                    processCallbacks();
```
cgtz, line 1370: i'm a little worried about calling processCallbacks from a different thread in a callback. I don't think it was designed to be called from non event loop threads. Will take a deeper look to see if there's any alternative (or if it is already thread safe)

justinlin-linkedin: you are right, the progresstracker and someother object is not thread safe and should probably only be used in event loop thread. I will update it.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1354,12 +1357,19 @@ private void handleMetadataBlob(BlobData blobData, byte[] userMetadata, ByteBuff
...
```
cgtz, line 1371: Apologies if you already briefed me on this, but what was the original reason for moving the process callback logic to the callback? Does this prevent a specific leak situation?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1139,49 +1142,49 @@ MessageFormatFlags getOperationFlag() {
...
+            decryptCallbackResultInfo.exception);
+        setOperationException(
+            new RouterException("Exception thrown on decrypting content for " + blobType + " blob " + blobId,
+                decryptCallbackResultInfo.exception, RouterErrorCode.UnexpectedInternalError));
+        progressTracker.setCryptoJobFailed();
+      } else {
+        // in case of Metadata blob, only user-metadata needs decryption if the blob is encrypted
+        if (blobType == BlobType.MetadataBlob) {
+          logger.trace("Processing stored decryption callback result for Metadata blob {}", blobId);
+          initializeDataChunks();
```
cgtz, line 1159: For 100% read-after-write across thread guarantee, the variables set in `initializeDataChunks()`, `blobInfo`, etc. should probably be volatile (possible I am misinterpreting the documentation). However, I suspect that we have a few other cases where this was not done which have not caused (discoverable) issues in practice :P 

https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.4.4

justinlin-linkedin:
Hi @cgtz  I revert the thread change for decryption callback, for I found another way to avoid a corner case bytebuf leak. Now the decryption callback is in the same thread as poll and handleResponse.

cgtz (approved):

PR #1589, opened 2020-07-13
Drop http2 request in two cases

zzmao:
Drop request on router requests to drop a request.
Drop request on writeAndFlush exceeds threshold.
Adjusted Netty High and Low WaterMark for isWriteable() tracking. 
There was an assumption that a stream channel can be released twice. This is wrong because MultiplexedChannelRecord maintains a counter for streams on it. This PR fixs it by checking if RESPONSE_INFO attr null or not.

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
@@ -17,26 +17,33 @@ package com.github.ambry.network.http2;
 import com.github.ambry.commons.SSLFactory;
 import com.github.ambry.config.Http2ClientConfig;
 import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandler;
+import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelPipeline;
 import io.netty.channel.pool.AbstractChannelPoolHandler;
 import io.netty.channel.pool.ChannelPoolHandler;
+import io.netty.handler.codec.http2.Http2Exception;
 import io.netty.handler.codec.http2.Http2FrameCodecBuilder;
 import io.netty.handler.codec.http2.Http2FrameLogger;
 import io.netty.handler.codec.http2.Http2MultiplexHandler;
 import io.netty.handler.codec.http2.Http2Settings;
 import io.netty.handler.logging.LogLevel;
 import io.netty.handler.ssl.SslHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * A {@link ChannelPoolHandler} to be used with {@link Http2ChannelPoolMap}.
  */
 public class Http2ChannelPoolHandler extends AbstractChannelPoolHandler {
+  private static final Logger logger = LoggerFactory.getLogger(Http2ChannelPoolHandler.class);
   private final SSLFactory sslFactory;
   private final String host;
   private final int port;
   private final Http2ClientConfig http2ClientConfig;
+  private final ConnectionInboundExceptionHandler connectionInboundExceptionHandler;
 
   /**
    * Construct a {@link Http2ChannelPoolHandler}.
@@ -49,6 +56,7 @@ public class Http2ChannelPoolHandler extends AbstractChannelPoolHandler {
     this.host = host;
     this.port = port;
     this.http2ClientConfig = http2ClientConfig;
+    this.connectionInboundExceptionHandler = new ConnectionInboundExceptionHandler();
   }
 
   @Override
@@ -63,6 +71,20 @@ public class Http2ChannelPoolHandler extends AbstractChannelPoolHandler {
         .frameLogger(new Http2FrameLogger(LogLevel.DEBUG, "client"))
         .build());
     pipeline.addLast(new Http2MultiplexHandler(new ChannelInboundHandlerAdapter()));
+  }
+
+  @ChannelHandler.Sharable
+  private static class ConnectionInboundExceptionHandler extends ChannelInboundHandlerAdapter {
+
+    @Override
+    public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
+      if (cause instanceof Http2Exception.StreamException) {
+        // This usually happens when server returns response which from a dropped request.
+        logger.info(cause.getMessage());
+      } else {
+        logger.warn("Connection inbound exception:", cause);
+      }
+    }
   }
 }
 
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
@@ -71,6 +71,7 @@ public class Http2ChannelPoolHandler extends AbstractChannelPoolHandler {
         .frameLogger(new Http2FrameLogger(LogLevel.DEBUG, "client"))
         .build());
     pipeline.addLast(new Http2MultiplexHandler(new ChannelInboundHandlerAdapter()));
+    pipeline.addLast(connectionInboundExceptionHandler);
   }
 
   @ChannelHandler.Sharable
@@ -79,8 +80,8 @@ public class Http2ChannelPoolHandler extends AbstractChannelPoolHandler {
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
       if (cause instanceof Http2Exception.StreamException) {
-        // This usually happens when server returns response which from a dropped request.
-        logger.info(cause.getMessage());
+        // This usually happens when server returns response for a request which was already dropped.
+        logger.info("StreamException: {}", cause.getMessage());
       } else {
         logger.warn("Connection inbound exception:", cause);
       }
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
@@ -90,9 +90,24 @@ public class Http2NetworkClient implements NetworkClient {
   @Override
   public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Integer> requestsToDrop,
       int pollTimeoutMs) {
-    long startTime = System.currentTimeMillis();
 
     List<ResponseInfo> readyResponseInfos = new ArrayList<>();
+    if (requestsToDrop.size() != 0) {
+      logger.warn("Number of requestsToDrop: {}", requestsToDrop.size());
+      http2ClientMetrics.http2RequestsToDropCount.inc(requestsToDrop.size());
+      for (int correlationId : requestsToDrop) {
+        Channel streamChannel = correlationIdInFlightToChannelMap.get(correlationId);
+        if (streamChannel != null) {
+          logger.warn("Drop request on streamChannel: {}", streamChannel);
+          RequestInfo requestInfo = releaseAndCloseStreamChannel(streamChannel);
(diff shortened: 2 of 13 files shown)


ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -79,34 +83,61 @@ public Http2NetworkClient(Http2ClientMetrics http2ClientMetrics, Http2ClientConf
...
+                    long writeAndFlushUsedTime = System.currentTimeMillis() - streamAcquiredTime;
+                    http2ClientMetrics.http2StreamWriteAndFlushTime.update(writeAndFlushUsedTime);
                     requestInfo.setStreamSendTime(System.currentTimeMillis());
+                    if (writeAndFlushUsedTime > http2ClientConfig.http2WriteAndFlushTimeoutMs) {
+                      // This usually happens if remote can't accept data in time.
+                      logger.warn(
+                          "WriteAndFlush exceeds http2RequestTimeoutMs {}ms, used time: {}ms, stream channel {}",
+                          http2ClientConfig.http2WriteAndFlushTimeoutMs, writeAndFlushUsedTime, streamChannel);
+                      if (http2ClientConfig.http2DropRequestOnWriteAndFlushTimeout) {
+                        releaseAndCloseStreamChannel(streamChannel);
```
cgtz, line 133: Is there a reason for dropping the connection at this point as opposed to just waiting until the router tells the NetworkClient to drop the request after routerRequestTimeoutMs? At this point, the request has already been flushed successfully so the request memory would have already been freed.

zzmao: I agree with you . Changed http2DropRequestOnWriteAndFlushTimeout to false by default.

zzmao: Want to discuss with your guys @justinlin-linkedin @cgtz : 

If a request is flushed successfully(here), its refCnt is decreased, but there is a reference chain that ```AttributeKey<RequestInfo> -> PutRequest -> Bytebuf```, so even refCnt is 0, data is still in memory and can't be recycled. Is this something we can improve?

justinlin-linkedin: I am worried about this. Comparing to the small object ByteBuf, the byte content held by ByteBuf is way larger. As long as the byte content can be reclaimed by the memory pool, we can ignore the small ByteBuf object.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -144,6 +174,12 @@ public void operationComplete(ChannelFuture future) throws Exception {
...
       http2ClientMetrics.http2RequestsToDropCount.inc(requestsToDrop.size());
+      for (int correlationId : requestsToDrop) {
+        Channel streamChannel = correlationIdInFlightToChannelMap.remove(correlationId);
+        if (streamChannel != null) {
+          releaseAndCloseStreamChannel(streamChannel);
+        }
+      }
     }
 
     http2ClientResponseHandler.getResponseInfoQueue().poll(readyResponseInfos, pollTimeoutMs);
```
justinlin-linkedin, line 185: is there a possibility that redyResponseInfos contain responses for the dropped request? if so, do we care?

justinlin-linkedin, line 185: Also, should probably remove the correlationIds for these responses from the inflight map.

zzmao: Good catch. Re-organized the map logic:
add: when a stream is acquired
remove: after `http2ClientResponseHandler.getResponseInfoQueue().poll()`

zzmao: It should include responses from dropped request. That's how router knows it was drooped.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
zzmao: Not fully understand your comment.
If it's PutRquest and it holds big chunk of data, we should break the link between `PutRequst -> ByteBuf`, for GC to recycle, right?

ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
```diff
@@ -64,5 +72,19 @@ public void channelCreated(Channel ch) throws Exception {
         .build());
     pipeline.addLast(new Http2MultiplexHandler(new ChannelInboundHandlerAdapter()));
   }
+
+  @ChannelHandler.Sharable
+  private static class ConnectionInboundExceptionHandler extends ChannelInboundHandlerAdapter {
```
justinlin-linkedin, line 77: what is this? This handler is not used anywhere, and it seems like we are trying to print out some info when it's a stream exception, which makes this a stream handler, doesn't stream handler belong to Http2NetworkClient.StreamChannelInitializer?

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -79,24 +84,47 @@ public Http2NetworkClient(Http2ClientMetrics http2ClientMetrics, Http2ClientConf
...
   public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Integer> requestsToDrop,
       int pollTimeoutMs) {
-    long startTime = System.currentTimeMillis();
+
     List<ResponseInfo> readyResponseInfos = new ArrayList<>();
+    if (requestsToDrop.size() != 0) {
+      logger.warn("Number of requestsToDrop: {}", requestsToDrop.size());
+      http2ClientMetrics.http2RequestsToDropCount.inc(requestsToDrop.size());
+      for (int correlationId : requestsToDrop) {
+        Channel streamChannel = correlationIdInFlightToChannelMap.get(correlationId);
```
justinlin-linkedin, line 99: `correlationIdInFlightToChannelMap.remove(correlationId)` would be better.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
justinlin-linkedin: Yes and no. When using Unpooled bytebuf, big chunk of data will rely on gc to release the memory. When using Pooled ByteBuf, big chunk of data will rely on calling `release` method to return the chunk back to the pool. GC doesn't reclaim the data chunk, only the small heap memory of ByteBuf fields.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
zzmao: If a request is drop, its streamChannel will be closed, but if a response is already inflight, and then arrives client. 
Http2MultiplexHandler will report unknown stream id error and print a long stack. 
The handler here is to avoid such useless stack in our log.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
```diff
@@ -49,6 +56,7 @@ public Http2ChannelPoolHandler(SSLFactory sslFactory, String host, int port, Htt
     this.host = host;
     this.port = port;
     this.http2ClientConfig = http2ClientConfig;
+    this.connectionInboundExceptionHandler = new ConnectionInboundExceptionHandler();
```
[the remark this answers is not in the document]
zzmao: yes, added to connection pipeline. Stream pipeline is not impacted.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -124,8 +164,11 @@ public void operationComplete(ChannelFuture future) throws Exception {
                     releaseAndCloseStreamChannel(streamChannel);
                     http2ClientResponseHandler.getResponseInfoQueue()
                         .put(new ResponseInfo(requestInfo, NetworkClientErrorCode.NetworkError, null));
-                    // release related bytebuf
-                    requestInfo.getRequest().release();
+                    if (!(future.cause() instanceof ClosedChannelException)) {
+                      // If it's ClosedChannelException caused by drop request, it's probably refCnt has been decreased.
```
cgtz, line 168: Why doesn't the ref count get decremented in other exception cases? Are we assuming that other exception cases come from ChannelHandlers written by Ambry that don't decrement the ref count?

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
cgtz: Does this mean that a request cannot be dropped if stream acquisition takes a long time? I guess that is okay since that may be a rare occurrence and would require some complicated logic in the acquire listener to determine if the request can be sent or not.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2ChannelPoolHandler.java
cgtz: Are we sure that all resources on child channels will be released if the exception is caught and swallowed here?

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
zzmao: You are right, if a stream is not acquired, it's won't be in inflight map, so it can't be dropped. It's rare, we don't expect stream acquisition takes a long time. 

And since the inflight map's value is streamChannel(to close it in case of drop request), we can't add <correlationId, streamChannel> to inflight map before stream acquisition done. 

An extra time check can be made right after stream acquired if it's really need.

build.gradle
```diff
@@ -248,7 +248,8 @@ project(':ambry-network') {
         compile project(':ambry-api'),
                 project(':ambry-utils'),
                 project(':ambry-commons'),
-                project(':ambry-clustermap')
+                project(':ambry-clustermap'),
+                project(':ambry-protocol')
```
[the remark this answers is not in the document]
zzmao: It's used for a metric:

```
      RequestOrResponse request = (RequestOrResponse) (requestInfo.getRequest());
      long waitingTime = streamInitiateTime - request.requestCreateTime;
      http2ClientMetrics.requestToNetworkClientLatencyMs.update(waitingTime);
```

I can remove this metric if it's not clean to import `ambry-protocol`.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
zzmao: It's because I am not clear when exactly refCnt is decreased. Let me do more reading.

zzmao: @cgtz @justinlin-linkedin   I need suggestions from you.

I just did an experiment which added  an excpetion throw in AmbrySendToHttp2Adaptor.write() (after `ctx.write(dataFrame, promise)`). In this case, writeAndFlush listener see refCnt 1. 

But in the other case we saw, channel closed, writeAndFlush listener is also triggered. refCnt is 0. (because if we do one more refCnt decrase, netty reports error.)

build.gradle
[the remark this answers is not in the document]
cgtz: It is probably okay to import since that dependency is probably needed by anyone that uses the network client anyways. One other alternative would be to add a `getRequestCreationTime()` method to the `SendWithCorrelationId` interface. I don't have a strong preference for either direction though.

PR #1590, opened 2020-07-15
Add support for undelete in blobstorestats

justinlin-linkedin:
Adding undelete support for blobstore stats in calculating valid entry size for log segment and container.
Realtime updating valid size is not supported in this PR.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -531,14 +531,14 @@ class BlobStoreStats implements StoreStats, Closeable {
       } else if (indexValue.isUndelete()) {
         if (keyFinalStates.containsKey(key)) {
           IndexFinalState state = keyFinalStates.get(key);
-          if (state.isTtlUpdate()) {
-            indexValue.setExpiresAtMs(Utils.Infinite_Time);
-          }
           if (state.isDelete() || (state.getLifeVersion() != indexValue.getLifeVersion())) {
             // This UNDELETE is not valid, when the final state of this storeKey is
             // 1. DELETE, or
             // 2. the current lifeVersion is not the same
             continue;
+          }
+          if (state.isTtlUpdate()) {
+            indexValue.setExpiresAtMs(Utils.Infinite_Time);
           }
         } else {
           long operationTimeInMs =
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -383,10 +383,9 @@ class BlobStoreStats implements StoreStats, Closeable {
   /**
    * Function that handles new UNDELETE after a scan to keep the current {@link ScanResults} relevant.
    * @param undeleteValue the {@link IndexValue} of the new UNDELETE
-   * @param originalPutValue the {@link IndexValue} of the original PUT that is getting deleted
    */
-  void handleNewUndeleteEntry(IndexValue undeleteValue, IndexValue originalPutValue) {
-    enqueueNewValue(undeleteValue, originalPutValue);
+  void handleNewUndeleteEntry(IndexValue undeleteValue) {
+    enqueueNewValue(undeleteValue, null);
   }
 
   /**
@@ -496,6 +495,9 @@ class BlobStoreStats implements StoreStats, Closeable {
    * @param indexSegment the {@link IndexSegment} where the entries came from
    * @param referenceTimeInMs the reference time in ms until which deletes and expiration are relevant
    * @param keyFinalStates a {@link Map} of key to {@link IndexFinalState}.
+   * @param removeFromStates if {@code True}, then remove the {@link IndexFinalState} from the given map {@code keyFinalStates}
+   *                         when encountering PUT IndexValue. This method iterates through IndexValues from most recent one to
+   *                         earliest one, so PUT IndexValue is the last IndexValue for the same key.
    * @param validIndexEntryAction the action to take on each valid {@link IndexEntry} found.
    * @throws StoreException if there are problems reading the index.
    */
@@ -515,7 +517,7 @@ class BlobStoreStats implements StoreStats, Closeable {
             // This DELETE is not valid, when the final state of this storeKey is
             // 1. UNDELETE, or
             // 2. DELETE, but the current lifeVersion is not the same, or
-            // 2. TTL_UPDATE
+            // 3. TTL_UPDATE
             continue;
           }
         } else {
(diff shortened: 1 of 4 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -488,40 +495,73 @@ public void close() {
...
+            // 2. DELETE, but the current lifeVersion is not the same, or
+            // 2. TTL_UPDATE
+            continue;
+          }
+        } else {
+          long operationTimeInMs =
+              indexValue.getOperationTimeInMs() == Utils.Infinite_Time ? indexSegment.getLastModifiedTimeMs()
+                  : indexValue.getOperationTimeInMs();
+          keyFinalStates.put(indexEntry.getKey(),
+              new IndexFinalState(indexValue.getFlags(), operationTimeInMs, indexValue.getLifeVersion()));
```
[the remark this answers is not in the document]
justinlin-linkedin: that's right, we iterate through the index values in descending order, where we iterate from that last index segment to the first one. And within each index segment, we iterate from the latest IndexValue to the earliest one. So long as there is a value in the map, this value would be the final state.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -488,40 +494,76 @@ public void close() {
...
+        }
         validIndexEntryAction.accept(indexEntry);
-        if (!isExpired(indexValue.getExpiresAtMs(), referenceTimeInMs)) {
+      } else if (indexValue.isUndelete()) {
+        if (keyFinalStates.containsKey(key)) {
+          IndexFinalState state = keyFinalStates.get(key);
+          if (state.isTtlUpdate()) {
+            indexValue.setExpiresAtMs(Utils.Infinite_Time);
+          }
```
jsjtzyy, line 536: nit:  put this `if` after line542

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -488,40 +494,76 @@ public void close() {
...
+              new IndexFinalState(indexValue.getFlags(), operationTimeInMs, indexValue.getLifeVersion()));
+        }
+        if (!isExpired(indexValue.getExpiresAtMs(), referenceTimeInMs)) {
+          validIndexEntryAction.accept(indexEntry);
         }
       } else if (indexValue.isTtlUpdate()) {
-        if (isTtlUpdateEntryValid(indexEntry.getKey(), indexValue, referenceTimeInMs, deletedKeys)) {
+        if (isTtlUpdateEntryValid(key, indexValue, referenceTimeInMs, keyFinalStates)) {
           validIndexEntryAction.accept(indexEntry);
         }
```
jsjtzyy, line 556: Maybe I overlooked something but I feel like it doesn't check if final state is delete case when validating the current TtlUpdate, does it? Please point me to the logic if I missed anything.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -488,40 +494,76 @@ public void close() {
...
-        // This is a put record
-        if (isExpired(indexValue.getExpiresAtMs(), referenceTimeInMs)) {
-          // try to update the expiration time for this put if there is an ttl update
-          IndexValue newValue = index.findKey(indexEntry.getKey());
-          if (newValue != null) {
-            indexValue.setExpiresAtMs(newValue.getExpiresAtMs());
-          }
+        IndexFinalState finalState = removeFromStates ? keyFinalStates.remove(key) : keyFinalStates.get(key);
+        if (finalState != null && finalState.isDelete() && finalState.getOperationTime() < referenceTimeInMs) {
+          // Put is deleted before reference time, it's not valid.
```
jsjtzyy, line 560: I don't understand why we need check `inalState.getOperationTime() < referenceTimeInMs`.  I think once final state is delete, this put is no longer valid.

justinlin-linkedin: it's not just that. when the final state is delete, but it's still within the retention duration, say it's deleted no more than 7 days ago, then the put is still a valid record. It's only after 7 days, the put would be invalid.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: it's in line 595. It checks if there is a put in the same log segment with ttl update, if so, check if this put is valid or not. if it's not, then the ttl update is not valid either.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -550,26 +592,20 @@ private boolean isTtlUpdateEntryValid(StoreKey key, IndexValue ttlUpdateValue, l
       valid = false;
     } else if (putValue.getOffset().getName().equals(ttlUpdateValue.getOffset().getName())) {
       // can be invalid if it is in the same log segment and the put is invalid
-      valid = isPutEntryValid(key, putValue, referenceTimeInMs, deletedKeys);
+      IndexFinalState finalState = keyFinalStates.get(key);
+      if (finalState != null && finalState.isDelete() && finalState.getOperationTime() < referenceTimeInMs) {
```
jsjtzyy, line 596: why this is not `finalState.getOperationTime() > referenceTimeInMs`?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
jsjtzyy: I see, what if put is not in the same log segment, how do we handle that?

justinlin-linkedin: There is a comments regarding this case. As long as the PUT exists and not in the same log segment, TTL_UPDATE is considered as valid. I suppose this is because the stats is used to trigger compaction, and it's possible that in compaction we only compact log segment with TTL_UPDATE and not PUT. For instance, if PUT is in 0_0_log and TTL_UPDATE is in 1_0_log and in compaction, we only compaction 1_0_log, in this case, TTL_UPDATE would be considered as valid and will be kept after compaction. I suppose the logic here is trying to be as close to the compaction logic as possible.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: finalState is delete so the operation time is when this blob is deleted. when deletedAt is less than reference time, it means this PUT value is not valid. For example. the reference is now - 7days and the deletedAt is now - 10 days, it means the blob is deleted 10 days ago and we only have 7 days as retention duration. now -10 days < now - 7 days, the PUT is invalid, so as the TTL_UPDATE.

jsjtzyy: I see, make sense.

PR #1593, opened 2020-07-21
Convert (and cache) remote keys in replication GET response for timed out standby replicas in leader-based replication

Arun-LinkedIn:
During replication, as of part of legacy GDPR requirements, remote keys are converted (and cached) to local keys using StoreKeyConverter during metadata exchange. These converted/cached keys are used for validation and transformation after the missing blobs are fetched for them. In case of leader-based replication, when we are fetching missing blobs for standby replicas (timed out waiting for their keys to arrive from leaders), StoreKeyConverter would have cleared these keys from its cache as it is replicating with other replicas before time out happened. This fix is to re-convert remote keys and fill the cache when we fetch missing blobs for timed out standby replicas.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -436,7 +436,7 @@ public class ReplicaThread implements Runnable {
         }
 
         if (replicatingFromRemoteColo && leaderBasedReplicationAdmin != null) {
-          // Get a list of inactive standby replicas whose missing keys haven't arrived for long time.
+          // Get a list of blocked standby replicas whose missing keys haven't arrived for long time.
           // Use case: In leader-based cross colo replication, standby replicas don't send GET requests for missing keys
           // found in metadata exchange and expect them to come via leader <-> leader replication.
           // This is a safety condition to ensure that standby replicas are not stuck waiting for the keys to come from leader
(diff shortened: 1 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -708,6 +708,13 @@ void fixMissingStoreKeys(ConnectedChannel connectedChannel, List<RemoteReplicaIn
       GetResponse getResponse =
           getMessagesForMissingKeys(connectedChannel, exchangeMetadataResponseList, replicasToReplicatePerNode,
               remoteNode, remoteColoGetRequestForStandby);
+      if (remoteColoGetRequestForStandby) {
+        // If we are fetching missing keys for standby replicas (timed out waiting for their keys to arrive from leaders)
+        // during leader-based replication, convert (and cache) received keys in the GET response. This is needed as the
+        // StoreKeyConverter would have cleared these keys from its cache as it is replicating with other replicas before
+        // time out happened for these standby replicas.
+        batchConvertGetResponseKeys(getResponse);
```
jsjtzyy, line 716: If we move this to handling no progress replicas code block (line 438 ~470), we might be able to re-use `batchConvertReplicaMetadataResponseKeys` without creating a new method.
It seems the one minor change would suffice: change argument of `batchConvertReplicaMetadataResponseKeys` from `ReplicaMetadataResponse` to `List<ExchangeMetadataResponse>`

PR #1594, opened 2020-07-29
Introduce tool for migrating instance configs to property store

cgtz:
This HelixBootstrapUpgradeTool operation will fetch all instance configs
for a cluster, convert them into the new format, and then persist them
in the property store.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
@@ -694,21 +694,18 @@ public class HelixBootstrapUpgradeUtil {
    */
   private void uploadClusterAdminInfos(Map<String, Map<String, Map<String, String>>> adminInfosByDc,
       String clusterAdminType, String adminConfigZNodePath) {
-    Properties storeProps = new Properties();
-    storeProps.setProperty("helix.property.store.root.path", "/" + clusterName + "/" + PROPERTYSTORE_STR);
-    HelixPropertyStoreConfig propertyStoreConfig = new HelixPropertyStoreConfig(new VerifiableProperties(storeProps));
-    for (Map.Entry<String, ClusterMapUtils.DcZkInfo> entry : dataCenterToZkAddress.entrySet()) {
-      info("Uploading {} infos for datacenter {}.", clusterAdminType, entry.getKey());
-      List<String> zkConnectStrs = entry.getValue().getZkConnectStrs();
-      // The number of zk endpoints has been validated in the ctor of HelixBootstrapUpgradeUtil, no need to check it again
-      HelixPropertyStore<ZNRecord> helixPropertyStore =
-          CommonUtils.createHelixPropertyStore(zkConnectStrs.get(0), propertyStoreConfig, null);
-      ZNRecord znRecord = new ZNRecord(clusterAdminType);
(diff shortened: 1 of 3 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -700,20 +718,47 @@ private void uploadClusterAdminInfos(Map<String, Map<String, Map<String, String>
...
       }
       helixPropertyStore.stop();
     }
   }
 
+  /**
+   * Convert instance configs to the new DataNodeConfig format and persist them in the property store.
+   */
+  private void migrateToPropertyStore() {
+    adminForDc.forEach((dcName, helixAdmin) -> {
```
jsjtzyy, line 735: Optional: considering making this multi-threaded (one thread  per colo). 

(Never mind, it should be fine, as it's one-off operation)

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -700,20 +718,47 @@ private void uploadClusterAdminInfos(Map<String, Map<String, Map<String, String>
    * @param adminConfigZNodePath the znode path of admin config.
    */
   private void deleteClusterAdminInfos(String clusterAdminType, String adminConfigZNodePath) {
-    Properties storeProps = new Properties();
-    storeProps.setProperty("helix.property.store.root.path", "/" + clusterName + "/" + PROPERTYSTORE_STR);
-    HelixPropertyStoreConfig propertyStoreConfig = new HelixPropertyStoreConfig(new VerifiableProperties(storeProps));
```
jsjtzyy, line 705: Thanks for simplifying this. Could you also make the change for `uploadClusterAdminInfos` method as well (line 701 ~ 705)?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
cgtz: Parallelism is a good idea anyways. I made this multi-threaded since it doesn't add much complexity to the code.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
cgtz: done

cgtz:
@jsjtzyy,
That is a good point that the servers will still be mutating the instance configs while the tool is running. In fact, further inconsistencies can arise even after the tool finishes running if the servers are not using `CompositeDataNodeConfigSource`.

Since blocking the servers from updating the sealed state is hard, I think the best approach will be to coordinate usage of this tool with usage of `CompositeDataNodeConfigSource` and fix discrepancies as they arise by rerunning the tool (annoying but doable while the property store is not the source of truth for the servers):
1. Run migration once to make a best effort complete cluster view
2. Roll out `CompositeDataNodeConfigSource` to servers.
3. If discrepancies are detected, either through the built-in listener monitoring or manual zk inspection, run the migration tool again.
4. Repeat 3, until convergence has been achieved. Given that the number of servers sealed tends to change in batches roughly once per hour, we hopefully shouldn't have to repeat this step many times.
5. Start switching source of truth for `CompositeDataNodeConfigSource`

PR #1598, opened 2020-08-04
Log more about router operation time out error

justinlin-linkedin:
Print out more log info about the router operation time out error.
This is for https://jira01.corp.linkedin.com:8443/browse/EXC-191015

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
@@ -177,8 +177,8 @@ class DeleteOperation {
       logger.trace("DeleteRequest with response correlationId {} timed out for replica {} ",
           deleteRequest.getCorrelationId(), replica.getDataNodeId());
       onErrorResponse(replica, new RouterException(
-          "Operation timed out because of " + responseInfo.getError() + " at DataNode " + responseInfo.getDataNode(),
-          RouterErrorCode.OperationTimedOut));
+          "Operation to delete " + blobId + " timed out because of " + responseInfo.getError() + " at DataNode "
+              + responseInfo.getDataNode(), RouterErrorCode.OperationTimedOut));
     } else {
       if (deleteResponse == null) {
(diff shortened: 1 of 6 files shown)


ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -176,7 +176,9 @@ void handleResponse(ResponseInfo responseInfo, DeleteResponse deleteResponse) {
     if (responseInfo.getError() != null) {
       logger.trace("DeleteRequest with response correlationId {} timed out for replica {} ",
           deleteRequest.getCorrelationId(), replica.getDataNodeId());
-      onErrorResponse(replica, new RouterException("Operation timed out", RouterErrorCode.OperationTimedOut));
+      onErrorResponse(replica, new RouterException(
+          "Operation timed out because of " + responseInfo.getError() + " at DataNode " + responseInfo.getDataNode(),
+          RouterErrorCode.OperationTimedOut));
```
jsjtzyy, line 181: `deleteRequest` should be able to get blob id, this may be worthwhile to log.
Same for replica id.

justinlin-linkedin: you mean log out the blob id and replica id?

jsjtzyy: yes

justinlin-linkedin: replica id is the same ad datanode here so I just log out blob id.

jsjtzyy: actually replicaId contains datanode info but datanode has no info about replica

justinlin-linkedin: I mean we only want to know which host failed, datanode can provide that info.

PR #1599, opened 2020-08-04
Fix direct memory leak in GetResponse

zzmao:
Fix direct memory leak in GetResponse


ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -79,6 +79,7 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
       try {
         send.writeTo(byteBufChannel);
         dataContent = byteBufChannel.getBuf();
+        send.release();
```
justinlin-linkedin, line 82: let's use finally here.

PR #1600, opened 2020-08-05
Mark container status inactive in Zk

SophieGuo410:
Mark status inactive for all the containers in inactiveContainerCandidateSet. The set is generated based on aggregation stats report, all the delete_in_progress containers which have been fully compacted will be included in the set.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
@@ -390,7 +390,7 @@ public class HelixAccountService extends AbstractAccountService implements Accou
     if (inactiveContainerCandidateSet != null) {
       boolean successful = false;
       int retry = 0;
-      while (!successful && retry < config.retryCount) {
+      while (!successful && retry < config.maxRetryCountOnUpdateFailure) {
         Map<Short, Account> accountToUpdateMap = new HashMap<>();
         inactiveContainerCandidateSet.forEach(container -> {
           // start by getting account, and then get container from account to make sure that we are editing the most
@@ -406,11 +406,12 @@ public class HelixAccountService extends AbstractAccountService implements Accou
         successful = updateAccounts(accountToUpdateMap.values());
         if (!successful) {
           retry++;
-          Thread.sleep(config.retryDelay);
+          Thread.sleep(config.retryDelayMs);
         }
       }
       if (!successful) {
-        logger.error("Container failed to be marked as INACTIVE in inactiveContainerCandidateSet : {}", inactiveContainerCandidateSet);
+        logger.error("Failed to mark containers INACTIVE in set : {}  after {} retries", inactiveContainerCandidateSet, config.maxRetryCountOnUpdateFailure);
+        accountServiceMetrics.accountUpdatesToZkErrorCount.inc();
       }
     }
   }
--- a/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
@@ -375,15 +375,44 @@ public class HelixAccountService extends AbstractAccountService implements Accou
    */
   public void selectInactiveContainersAndMarkInZK(StatsSnapshot statsSnapshot) {
     Set<Container> inactiveContainerCandidateSet = selectInactiveContainerCandidates(statsSnapshot);
-    markContainerInactiveOnZk(inactiveContainerCandidateSet);
+    try {
+      markContainerInactiveOnZk(inactiveContainerCandidateSet);
+    } catch (InterruptedException e) {
+      logger.error("Mark inactive container in zookeeper is interrupted", e);
+    }
   }
 
   /**
    * Mark the given {@link Container}s status to INACTIVE in zookeeper.
    * @param inactiveContainerCandidateSet DELETE_IN_PROGRESS {@link Container} set which has been deleted successfully during compaction.
    */
-  private void markContainerInactiveOnZk(Set<Container> inactiveContainerCandidateSet) {
-    // TODO: mark the given containers status to INACTIVE in zookeeper.
+  void markContainerInactiveOnZk(Set<Container> inactiveContainerCandidateSet) throws InterruptedException {
+    if (inactiveContainerCandidateSet != null) {
+      boolean successful = false;
+      int retry = 0;
+      while (!successful && retry < config.retryCount) {
+        Map<Short, Account> accountToUpdateMap = new HashMap<>();
+        inactiveContainerCandidateSet.forEach(container -> {
+          // start by getting account, and then get container from account to make sure that we are editing the most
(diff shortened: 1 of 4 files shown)


ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -382,8 +382,17 @@ public void selectInactiveContainersAndMarkInZK(StatsSnapshot statsSnapshot) {
...
+  void markContainerInactiveOnZk(Set<Container> inactiveContainerCandidateSet) {
+    if (inactiveContainerCandidateSet != null) {
+      Map<Short, Account> accountToUpdateMap = new HashMap<>();
+      inactiveContainerCandidateSet.forEach(container -> {
+        container = new ContainerBuilder(container).setStatus(Container.ContainerStatus.INACTIVE).build();
+        short accountId = container.getParentAccountId();
+        Account account = new AccountBuilder(accountToUpdateMap.getOrDefault(accountId, getAccountById(accountId))).addOrUpdateContainer(container).build();
+        accountToUpdateMap.put(accountId, account);
+      });
+      updateAccounts(accountToUpdateMap.values());
```
cgtz, line 394: There can be some situations where this update will fail to persist if someone is making other account updates in Nuage concurrently. This is because the update will fail if the account's "snapshotVersion" field does not match. This field is used to make sure that the base account that was modified matches the current account stored in zookeeper so that one update does not erase someone else's change unexpectedly.

I would suggest using a compareAndSet style loop (checking the return value of `updateAccounts` to see if the update went through):
```java
    if (inactiveContainerCandidateSet != null) {
      boolean successful = false;
      while (!successful) {
        Map<Short, Account> accountToUpdateMap = new HashMap<>();
        inactiveContainerCandidateSet.forEach(container -> {
          // start by getting account, and then get container from account to make sure that we are editing the most
          // recent snapshot
          short accountId = container.getParentAccountId();
          Account accountToEdit = accountToUpdateMap.computeIfAbsent(accountId, this::getAccountById);
          Container containerToEdit = accountToEdit.getContainerById(container.getId());
          Container editedContainer =
              new ContainerBuilder(containerToEdit).setStatus(Container.ContainerStatus.INACTIVE).build();
          accountToUpdateMap.put(accountId,
              new AccountBuilder(accountToEdit).addOrUpdateContainer(editedContainer).build());
        });
        // Note that this assumes that unsuccessful updates are always retriable (like snapshot version violations).
        // If there is something unrecoverable wrong, this method will get stuck in this. Max retries may be needed.
        successful = updateAccounts(accountToUpdateMap.values());
      }
    }
```

ambry-api/src/main/java/com/github/ambry/config/HelixAccountServiceConfig.java
```diff
@@ -99,6 +100,20 @@
   @Default("100")
   public final int totalNumberOfVersionToKeep;
 
+  /**
+   * The number of retry times when the update accounts fails by marking delete_in_progress container status to inactive;
+   */
+  @Config(RETRY_COUNT)
+  @Default("10")
+  public final int retryCount;
```
jsjtzyy, line 108: nit: can we call this maxRetryCountOnUpdateFailure ?

ambry-api/src/main/java/com/github/ambry/config/HelixAccountServiceConfig.java
```diff
@@ -99,6 +100,20 @@
...
+
+  /**
+   * The sleep time between each retry action when the update accounts fails by marking delete_in_progress container status to inactive;
+   */
+  @Config(RETRY_DELAY)
+  @Default("1000")
+  public final long retryDelay;
```
jsjtzyy, line 115: Can we add time unit into the name? Like retryDelaySeconds or retryDelayMs ?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -375,15 +375,44 @@ private void onAccountChangeMessage(String topic, String message) {
...
+                  new AccountBuilder(accountToEdit).addOrUpdateContainer(editedContainer).build());
+        });
+        successful = updateAccounts(accountToUpdateMap.values());
+        if (!successful) {
+          retry++;
+          Thread.sleep(config.retryDelay);
+        }
+      }
+      if (!successful) {
+        logger.error("Container failed to be marked as INACTIVE in inactiveContainerCandidateSet : {}", inactiveContainerCandidateSet);
```
jsjtzyy, line 413:
```
logger.error("Failed to mark containers INACTIVE in set : {}  after {} retries", inactiveContainerCandidateSet, config.retryCount);
```

jsjtzyy: Also I am considering if we need a metric here for alerting purpose.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
SophieGuo410: Thanks for the reminding. I've updated to resolve this concurrency issue.

ambry-api/src/main/java/com/github/ambry/config/HelixAccountServiceConfig.java
SophieGuo410: Updated.

ambry-api/src/main/java/com/github/ambry/config/HelixAccountServiceConfig.java
SophieGuo410: Sure. Updated.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
SophieGuo410: Updated and add the metric.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -375,15 +375,45 @@ private void onAccountChangeMessage(String topic, String message) {
...
 
   /**
    * Mark the given {@link Container}s status to INACTIVE in zookeeper.
    * @param inactiveContainerCandidateSet DELETE_IN_PROGRESS {@link Container} set which has been deleted successfully during compaction.
    */
-  private void markContainerInactiveOnZk(Set<Container> inactiveContainerCandidateSet) {
-    // TODO: mark the given containers status to INACTIVE in zookeeper.
+  void markContainerInactiveOnZk(Set<Container> inactiveContainerCandidateSet) throws InterruptedException {
+    if (inactiveContainerCandidateSet != null) {
+      boolean successful = false;
```
zzmao, line 391: minor:  rename to `success` ?

SophieGuo410: Updated.

PR #1603, opened 2020-08-08
Add a test suite for AmbrySendToHttp2Adaptor

justinlin-linkedin:
Adding a test suite for AmbrySendToHttp2Adaptor
Make sure there is no memory leak.

Bonus testing: remove send.release at AmbrySendToHttp2Adaptor.java#L86 would leak memory.


justinlin-linkedin:
> Thanks for making the change!
> 
> Just want to make sure: can this test reproduce the memory leak without #1599 ?

Yes, I have tried. It can reproduce the memory leak.

PR #1604, opened 2020-08-10
Code changes to switch vcr helix cluster to DelayedAutoRebalancer.

ankagrawal:
Switch vcr helix cluster to DelayedAutoRebalancer strategy with OfflineOnline state machine and number of replicas 1.


ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -48,11 +49,13 @@
...
   private static final String SEPARATOR = "/";
   static List<String> ignoreResourceKeyWords = Arrays.asList("aggregation", "trigger", "stats");
-  private static final int REPLICA_NUMBER = 3;
+  private static final int REPLICA_NUMBER = 1;
   private static final ZNRecordSerializer ZN_RECORD_SERIALIZER = new ZNRecordSerializer();
   // TODO: get from properties file
   private static final int MAX_OFFLINE_INSTANCES_ALLOWED = 4;
   private static final int NUM_OFFLINE_INSTANCES_FOR_AUTO_EXIT = 2;
+  private static final int MIN_ACTIVE_REPLICAS = 0;
+  private static final long REBALANCE_DELAY = TimeUnit.MINUTES.toMillis(20);
```
jsjtzyy, line 58: same here

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -214,11 +217,13 @@ static void updateResourceIdealState(String destZkString, String destClusterName
    */
   static IdealState buildIdealState(String resource, Set<String> partitionSet) {
     FullAutoModeISBuilder builder = new FullAutoModeISBuilder(resource);
-    builder.setStateModel(LeaderStandbySMD.name);
+    builder.setStateModel(OnlineOfflineSMD.name);
     for (String partition : partitionSet) {
       builder.add(partition);
     }
-    builder.setRebalanceStrategy(CrushEdRebalanceStrategy.class.getName());
```
jsjtzyy, line 221: I suggest keeping this unless we have another of piece of code setting this for new resource, which I didn't see here.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -214,11 +217,13 @@ static void updateResourceIdealState(String destZkString, String destClusterName
...
+    builder.setMinActiveReplica(MIN_ACTIVE_REPLICAS);
+    builder.setRebalanceDelay((int) REBALANCE_DELAY);
+    builder.setRebalancerClass(DelayedAutoRebalancer.class.getName());
```
lightningrob, line 226: Would it possible to store the ideal state as a json doc in config property and have the populate tool take it as an argument? It seems clunky to have to modify the code every time we want to tweak our Helix config.

I realize the HelixVcrStateModel does need to know the state names.

ambry-cloud/src/test/java/com/github/ambry/cloud/VcrTestUtil.java
```diff
@@ -46,7 +47,9 @@
  */
 public class VcrTestUtil {
 
-  public static String helixResource = "resource1";
+  private static final int MIN_ACTIVE_REPLICAS = 0;
+  private static final long REBALANCE_DELAY = TimeUnit.SECONDS.toMillis(2);
```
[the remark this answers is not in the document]
ankagrawal: done.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
ankagrawal: done.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
ankagrawal: fixed. Interestingly the default rebalance strategy (which is Waged strategy i guess), isnt able to finalize a partition assignment in some cases. This was the reason the tests were failing too.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrCluster.java
```diff
@@ -167,4 +169,25 @@ public void close() {
     manager.disconnect();
     helixAdmin.close();
   }
+
+  /**
+   * Register the state model factory corresponding to {@code vcrHelixStateModelName}.
+   * @param vcrHelixStateModelName
+   */
+  private void registerStateModelFactory(String vcrHelixStateModelName) {
```
lightningrob, line 177: It's cool that you found a way to parametrize the state model.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrCluster.java
```diff
@@ -167,4 +169,25 @@ public void close() {
...
+        break;
+      case LeaderStandbySMD.name:
+        manager.getStateMachineEngine()
+            .registerStateModelFactory(vcrHelixStateModelName, new LeaderStandbyHelixVcrStateModelFactory(this));
+        break;
+      default:
+        throw new IllegalArgumentException(String.format(
+            "Illegal vcr helix state model name: " + vcrHelixStateModelName + ". Should be one of"
+                + OnlineOfflineSMD.name + ", " + LeaderStandbySMD.name));
+    }
```
jsjtzyy, line 191: I suggest adding state model validation into CloudConfig, thus it throws exception at the very beginning (when generating the config) rather than keep initializing other components until participation.
You can refer to `clustermapStateModelDefinition` in ClusterMapConfig.

ambry-cloud/src/main/java/com/github/ambry/cloud/OnlineOfflineHelixVcrStateModelFactory.java
```diff
@@ -0,0 +1,41 @@
...
+package com.github.ambry.cloud;
+
+import org.apache.helix.participant.statemachine.StateModel;
+import org.apache.helix.participant.statemachine.StateModelFactory;
+
+
+/**
+ * A factory for creating {@link OnlineOfflineHelixVcrStateModel}
+ */
+public class OnlineOfflineHelixVcrStateModelFactory extends StateModelFactory<StateModel> {
```
jsjtzyy, line 23: May I ask why we need two state model factories?  The code logic looks same with LeaderStandbyHelixVcrStateModelFactory. I think one factory should suffice.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -212,13 +222,19 @@ static void updateResourceIdealState(String destZkString, String destClusterName
    * @param partitionSet the set of partitions managed by the resource.
    * @return the {@link IdealState}.
    */
-  static IdealState buildIdealState(String resource, Set<String> partitionSet) {
+  static IdealState buildIdealState(String resource, Set<String> partitionSet,
+      IdealStateConfigFields idealStateConfigFields) {
```
jsjtzyy, line 226: nit: add java doc for new argument

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -13,64 +13,65 @@
...
     OptionSpec controlResourceOpt = parser.accepts("controlResource",
-        "Enable/Disable a resource. --controlResource --dest destZkEndpoint/destClusterName --resource resource --enable true");
+        "Enable/Disable a resource. --controlResource --dest destZkEndpoint/destClusterName"
+            + " --resource resource --enable true");
     ArgumentAcceptingOptionSpec<String> resourceOpt =
         parser.accepts("resource").withRequiredArg().describedAs("resource name").ofType(String.class);
 
     ArgumentAcceptingOptionSpec<Boolean> maintenanceOpt = parser.accepts("maintainCluster",
-        "Enter/Exit helix maintenance mode. --maintainCluster --dest destZkEndpoint/destClusterName --enable true")
+        "Enter/Exit helix maintenance mode. --maintainCluster --dest destZkEndpoint/destClusterName" + " --enable true")
```
jsjtzyy, line 74: nit: can a single string without concatenation.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -328,5 +344,120 @@ static void errorAndExit(String error) {
...
+      return allowAutoJoin;
+    }
+  }
+
+  /**
+   * Class for configs passed to the {@link HelixVcrPopulateTool}
+   */
+  static class Config {
+    private IdealStateConfigFields idealStateConfigFields;
+    private ClusterConfigFields clusterConfigFields;
```
jsjtzyy, line 446: I suppose there is ordering requirement (i.e. idealStateConfigFields comes first) when preparing the json file. In that case, it's a bit of error-prone for user to use this tool.

ambry-tools/src/integration-test/resources/mocktio-extensions/DelayedAutoRebalancerHelixConfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "minActiveReplicas": 0,
+    "rebalancerClassName": "org.apache.helix.controller.rebalancer.DelayedAutoRebalancer",
+    "rebalanceDelayInMins": 20
+  },
+  "clusterConfigFields": {
+    "maxOfflineInstancesAllowed": 4,
+    "numOfflineInstancesForAutoExit": 2,
+    "allowAutoJoin": true
+  }
+}
```
jsjtzyy, line 15: You can move this to `config` directory. For testing, you can just instantiate a `Config` object (as you defined HelixVcrPopulateTool) and write it into string via jackson's `writeValueAsString` method.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
ankagrawal: done.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
```diff
@@ -335,5 +344,120 @@ static void errorAndExit(String error) {
...
+   * @return true if rebalancer is {@link org.apache.helix.controller.rebalancer.DelayedAutoRebalancer}, false otherwise.
+   */
+  private static boolean isDelayedRebalanceEnabled(String rebalancerClassName) {
+    return (rebalancerClassName != null) && rebalancerClassName.endsWith(DELAYED_REBALANCER_CLASS_NAME);
+  }
+
+  /**
+   * Class for ideal state configs.
+   */
+  static class IdealStateConfigFields {
```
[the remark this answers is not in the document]
ankagrawal: If you mean how does the deserialization happen? I guess, the jackson library matches the json keys with the getters and assigns values to the fields using reflection.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
ankagrawal: fixed.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
ankagrawal: done.

ambry-tools/src/main/java/com/github/ambry/clustermap/HelixVcrPopulateTool.java
ankagrawal: There is no ordering requirement. The ordering of the keys in the json file doesn't matter, as long as the general structure is correct.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrCluster.java
ankagrawal: fixed.

ambry-tools/src/integration-test/resources/mocktio-extensions/DelayedAutoRebalancerHelixConfig.json
ankagrawal: done

ambry-cloud/src/main/java/com/github/ambry/cloud/OnlineOfflineHelixVcrStateModelFactory.java
ankagrawal: Thanks for pointing this out. I have changed the logic to make factory as the argument.

While the code is similar, there are still a few reasons to maintain separate factories:

1. We need two state model factory classes because the `createStateModel` method that is used by helix to instantiate the state model takes only `resourceName` and `partitionName` as arguments. We need the `createStateModel` method to create the state model class based on the config.
2. When we call `HelixStateMachineEngine.registerStateModelFactory`, the code expects to have a separate factory for each statemodel. Check `org.apache.helix.participant.HelixStateMachineEngine::registerStateModelFactory`. We can of course create and register just one factory and still make it work. But creating separate factory seems cleaner based on the helix implementation.
3. Usually maintaining different factories keeps the test code cleaner too. In this case we are able to get away because helix provides us with a way to do integration testing, but in case any test requires us to pass a custom or mock state model, it can just pass it via config without any test related abstraction leaking into the production code.
4. The helix code examples use the same model too. Check `org.apache.helix.examples.BrokerResourceOnlineOfflineStateModelFactory.java` and `org.apache.helix.examples.BootstrapHandler.java`

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrCluster.java
ankagrawal: I have changed to code to pass factory class instead. Let me know what do you think.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrCluster.java
```diff
@@ -167,4 +171,22 @@ public void close() {
...
+    if (stateModelFactoryClass.equals(OnlineOfflineHelixVcrStateModelFactory.class.getName())) {
+      return OnlineOfflineSMD.name;
+    }
+    if (stateModelFactoryClass.equals(LeaderStandbyHelixVcrStateModelFactory.class.getName())) {
+      return LeaderStandbySMD.name;
+    }
+    throw new IllegalArgumentException(String.format(
+        "Illegal vcr helix state model factory class: " + stateModelFactoryClass + ". Should be one of"
+            + OnlineOfflineHelixVcrStateModelFactory.class.getName() + ", "
+            + LeaderStandbyHelixVcrStateModelFactory.class.getName()));
```
jsjtzyy, line 190: [An alternative but totally optional]
Since we decide to use two factories, we can make a abstract factory `VcrStateModelFactory` which extends `StateModelFactory` and make two state model factories extend it. Then this method can be replaced by adding state model name to each StateModelFactory. Thus, line149~151 can be changed to:
```
VcrStateModelFactory stateModelFactory = Utils.getObj(cloudConfig.vcrHelixStateModelFactoryClass, this);
manager.getStateMachineEngine()
        .registerStateModelFactory(stateModelFactory.modelName, stateModelFactory);
```
And this method can be removed.

PR #1605, opened 2020-08-17
Remove helix-core dependency in ambry-api module

cgtz:
Having ambry-api depend on helix can cause some transitive dependency
issues when ambry-api is depended on as a library in non-ambry applications.
Also, since helix is more of an implementation-specific component, the
api module should ideally not depend on it.

To remove the dependency, the following changes were made:
1. Move CommonUtils and ClusterParticipant to ambry-commons, and have
   ambry-commons depend on helix-core.
2. Make ambry-clustermap depend on ambry-commons instead of the other
   way around. This required getting rid of an instanceof check in
   ResponseHandler and delegating that logic to ClusterMap impls.
3. Move the UNKNOWN_DATACENTER_ID constant to the API package. This
   was the only reason ambry-commons depended on ambry-clustermap before.
4. Remove InstanceType from the VirtualReplicatorCluster interface. In
   light of the creation of ClusterSpectator, there is no need to run
   a VirtualReplicatorCluster as a spectator.
5. Replace instanceof check in store module with new feature check
   method in ClusterParticipant interface.


jsjtzyy:
Could you present a little more context of `ambry-api is depended on as a library in non-ambry applications`?

cgtz:
> Could you present a little more context of `ambry-api is depended on as a library in non-ambry applications`?

@jsjtzyy . Within LinkedIn, ambry-client depends on ambry-api. ambry-client is depended by about ~150 multiproducts. One complication that can arise is around commonly used transitive dependencies like guava and jackson. For example, some play apps have checks that very specific versions (2.x to 2.y) of jackson are used.

Long term, I would like to factor out the parts of ambry-api that are truly needed by ambry-client into a separate subproject -- something like ambry-public-api. Most of ambry-api, other than the rest API header utilities and some json request serdes, is for "internal" apis used within ambry implementations and is not needed by ambry-client. However, even before doing this, I think it is a best practice to not have ambry-api depend on helix-core since it is relevant to specific implementations of AccountService and ClusterMap, but not a core part of the generic interface.

jsjtzyy:
> > Could you present a little more context of `ambry-api is depended on as a library in non-ambry applications`?
> 
> @jsjtzyy . Within LinkedIn, ambry-client depends on ambry-api. ambry-client is depended by about ~150 multiproducts. One complication that can arise is around commonly used transitive dependencies like guava and jackson. For example, some play apps have checks that very specific versions (2.x to 2.y) of jackson are used.
> 
> Long term, I would like to factor out the parts of ambry-api that are truly needed by ambry-client into a separate subproject -- something like ambry-public-api. Most of ambry-api, other than the rest API header utilities and some json request serdes, is for "internal" apis used within ambry implementations and is not needed by ambry-client. However, even before doing this, I think it is a best practice to not have ambry-api depend on helix-core since it is relevant to specific implementations of AccountService and ClusterMap, but not a core part of the generic interface.

Got it and make sense. Other than ambry-api, are there any other packages that are depended by non-ambry apps?

PR #1607, opened 2020-08-18
Add configuration to limit the number of concurrent delete operations in background deleter

justinlin-linkedin:
Blobs in delivery cluster are usually pretty huge. When router is trying to delete a blob, router would create delete operations for all the chunks of this blob. The number of chunks which belong to this blob can be huge, and background deleter would flood the network client with delete operations. This might cause connection unavailable error when the servers are not responding to the frontends fast enough. 

Since they are background deletes, we can de-prioritize them by setting a limit on how many concurrent delete operations a router can have. By doing so, we would mitigate the connection unavailable error.

However, this is not an ideal solution. To solve this issue, we should
1. Set a different time out for checking connection from pool for background deletes.
2. De-prioritize background delete requests in the network layer (and maybe retry?) to support better QOS.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -103,6 +103,8 @@ public class RouterConfig {
   public static final String ROUTER_ENABLE_HTTP2_NETWORK_CLIENT = "router.enable.http2.network.client";
   public static final String ROUTER_CROSS_COLO_REQUEST_TO_DC_WITH_MOST_REPLICAS =
       "router.cross.colo.request.to.dc.with.most.replicas";
+  public static final String ROUTER_BACKGROUND_DELETER_MAX_CONCURRENT_OPERATIONS =
+      "router.background.deleter.max.concurrent.operations";
 
   /**
    * Number of independent scaling units for the router.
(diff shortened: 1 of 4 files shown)


ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -500,6 +502,13 @@
   @Default("false")
   public final boolean routerCrossColoRequestToDcWithMostReplicas;
 
+  /**
+   * The maximum number of outgoing delete operations in background deleter. 0 means no limit.
+   */
+  @Config(ROUTER_BACKGROUND_DELETER_MAX_CONCURRENT_OPERATIONS)
+  @Default("0")
```
zzmao, line 509: Prefer to use -1. To reduce some `<=` confusion in below.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -1085,12 +1104,49 @@ protected void updateBlobTtl(String blobIdStr, final String serviceId, long expi
       completeUpdateBlobTtlOperation(routerException, futureResult, callback);
     }
 
+    /**
+     * Requests for a blob to be deleted asynchronously and invokes the {@link Callback} when the request completes.
```
zzmao, line 1108: Update Java Doc.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
justinlin-linkedin: I prefer 0, because of the getIntegerInRange method. If I set the range from -1 to MAX_INT, then 0 would be another special value besides -1. Setting range from 0 to MAX_INT solve this issue.

zzmao: got you. Sounds good.

PR #1608, opened 2020-08-20
Use copy of ExchangeMetadataResponse while fetching missing keys for timed out standby replicas

Arun-LinkedIn:
Issues:
 1. IllegalStateException thrown in writeMessagesToLocalStoreAndAdvanceTokens() due to mismatch in partition ID in partitionResponseInfo with remoteReplicaInfo
 2. IllegalArgumentException thrown in getReplicaMetadataResponse() when trying to read from connectionChannel inputStream (size of the stream from first 8 bytes is incorrect)

Explanation: We are using references of ExchangeMetadataResponse stored in RemoteReplicaInfo while fetching the missing keys for standby(blocked) replicas. Due to that, it is possible that while an
inter-colo thread is fetching missing store messages present in ExchangeMetadaResponse::missingStoreMessages set, other threads receive these messages and "modify/empty the same set reference".

This will lead to mismatch in state of missingStoreMessages tracked for a partition before and after fetching the keys. That is, before fetch, we would find missingStoreMessages is non-empty and send PartitionRequestInfo in
GET request but when we receive the response,  we will skip/ignore the PartitionResponseInfo since missingStoreMessages is empty were emptied. This could lead to two kinds of issues during iteration of received responses in
writeMessagesToLocalStoreAndAdvanceTokens:
 1. If we skip/ignore such partitions in middle of iteration, indexes of replica list and partition response list mismatch. This causes IllegalStateException.
 2. If we skip/ignore such partitions at end of iteration, the inputStream is not read completely resulting in left over bytes for next replication cycle. This causes IllegalArgumentException.

Fix: Create a copy of ExchangeMetadataResponse (and its missingStoreMessages set) while fetching missing keys.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -465,7 +465,8 @@ public class ReplicaThread implements Runnable {
 
             List<ExchangeMetadataResponse> exchangeMetadataResponseListForBlockedReplicas =
                 standbyReplicasTimedOutOnNoProgress.stream()
-                    .map(RemoteReplicaInfo::getExchangeMetadataResponse)
+                    .map(remoteReplicaInfo -> new ExchangeMetadataResponse(
+                        remoteReplicaInfo.getExchangeMetadataResponse()))
                     .collect(Collectors.toList());
 
             // Convert (and cache) the remote keys that are being fetched as the StoreKeyConverter would have cleared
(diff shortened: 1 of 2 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1551,6 +1547,28 @@ ReplicationMetrics getReplicationMetrics() {
       this.receivedStoreMessagesWithUpdatesPending = null;
     }
 
+    /**
+     * Shallow copy Constructor for {@link ExchangeMetadataResponse}, which copies all field references with exception
+     * of 'missingStoreMessages' for which it creates new Set object.
+     * @param old old {@link ExchangeMetadataResponse} object.
+     */
+    ExchangeMetadataResponse(ExchangeMetadataResponse old) {
```
justinlin-linkedin, line 1555: nit: please use a different name like "other" instead of "old".

PR #1611, opened 2020-08-26
Use one EventLoopGroup in all http2NetworkClients

zzmao:
Use one EventLoopGroup in all http2NetworkClients. This is suggested in Netty best practice: The change is based on a best practice of netty: http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#25.0

### DIFF
--- a/ambry-server/src/integration-test/java/com/github/ambry/server/Http2NetworkClientTest.java
+++ b/ambry-server/src/integration-test/java/com/github/ambry/server/Http2NetworkClientTest.java
@@ -51,6 +51,10 @@ import com.github.ambry.utils.SystemTime;
 import com.github.ambry.utils.TestUtils;
 import com.github.ambry.utils.Utils;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.EventLoopGroup;
+import io.netty.channel.epoll.Epoll;
+import io.netty.channel.epoll.EpollEventLoopGroup;
+import io.netty.channel.nio.NioEventLoopGroup;
 import java.io.DataInputStream;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


justinlin-linkedin:
I don't really understand this PR. One OperationController in router would create one HTTP2NetworkClient, which would have its own eventloop. But with this change, we are forcing all the Http2NetworkClient in all OperationController to use the same eventloop?

justinlin-linkedin:
And did you see any performance boost with this change? And can you explain the reason why if there is any? I would be surprised if there is any performance benefit coming from this change.

zzmao:
@justinlin-linkedin  The change is based on a best practice of netty: http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#25.0

It's explainable, because a set of eventloopgroup has # of CPU threads. If more operation controllers, it would be multiple sets of threads. More threads cause more context switch. I think that's why the doc suggest to reuse eventloopgroup.

justinlin-linkedin:
> @justinlin-linkedin The change is based on a best practice of netty: http://normanmaurer.me/presentations/2014-facebook-eng-netty/slides.html#25.0
> 
> It's explainable, because a set of eventloopgroup has # of CPU threads. If more operation controllers, it would be multiple sets of threads. More threads cause more context switch. I think that's why the doc suggest to reuse eventloopgroup.

Can you add this to description so we will have a better commit message for this PR.

ambry-server/src/integration-test/java/com/github/ambry/server/Http2NetworkClientTest.java
```diff
@@ -97,6 +102,8 @@ public static void initializeTests() throws Exception {
         "http2-client");
     TestSSLUtils.addHttp2Properties(clientSSLProps, SSLFactory.Mode.CLIENT, false);
     clientSSLConfig = new SSLConfig(new VerifiableProperties(clientSSLProps));
+    eventLoopGroup = new NioEventLoopGroup();
```
justinlin-linkedin, line 105: nit: this line is no use at all.

justinlin-linkedin (approved):

PR #1612, opened 2020-08-28
Fix bugs in helix populate tool.

ankagrawal:


jsjtzyy:
Is the bug about `DelayedRebalancer`?  I see you changed it to `DelayedAutoRebalancer`.

ankagrawal:
> Is the bug about `DelayedRebalancer`? I see you changed it to `DelayedAutoRebalancer`.

Yes there is one about using the correct string for DelayedAutoRebalancer. There is another one in HelixVcrPopulateTool::72. `.withRequiredArg().ofType(Boolean.class);` should be used with the `enableOpt`.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrCluster.java
```diff
@@ -171,22 +168,4 @@ public void close() {
...
     helixAdmin.close();
   }
-
-  /**
-   * Get the name of the state model for the factory class.
-   * @param stateModelFactoryClass name of the {@link org.apache.helix.participant.statemachine.StateModelFactory} class.
-   * @return state model name corresponding to the factor class.
-   */
-  private String getStateModelNameFromFactory(String stateModelFactoryClass) {
-    if (stateModelFactoryClass.equals(OnlineOfflineHelixVcrStateModelFactory.class.getName())) {
```
lightningrob, line 181: Minor: could use equalsIgnoreCase (optional)

ankagrawal: This code has been removed. Each `VcrStateModelFactory` class now maintains its own statemodel name, so there is no longer any need for this comparison anymore.

PR #1613, opened 2020-08-31
Fix bucket reports in BlobStoreStats to support TTL_UPDATE and UNDELETE

justinlin-linkedin:
This PR
1. Fixes bugs that fail test cases
2. Add two delta buckets for log segment to deal with delete and expiration separately
3. Add support for TTL_UPDATE and UNDELETE
4. Add more test cases to test DELETE, TTL_UPDATE and UNDELETE.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -557,8 +557,6 @@ public class BlobStore implements Store {
                 "Cannot delete id " + info.getStoreKey() + " since it is already deleted in the index.",
                 StoreErrorCodes.ID_Deleted);
           }
-          indexValuesPriorToDelete.add(value);
-          lifeVersions.add(value.getLifeVersion());
         } else {
           // This is a delete request from replication
           if (value.isDelete() && value.getLifeVersion() == info.getLifeVersion()) {
@@ -571,9 +569,9 @@ public class BlobStore implements Store {
                 "Cannot delete id " + info.getStoreKey() + " since it has a higher lifeVersion than the message info: "
                     + value.getLifeVersion() + ">" + info.getLifeVersion(), StoreErrorCodes.Life_Version_Conflict);
           }
-          indexValuesPriorToDelete.add(value);
-          lifeVersions.add(info.getLifeVersion());
         }
+        indexValuesPriorToDelete.add(value);
+        lifeVersions.add(info.getLifeVersion());
         if (!value.isDelete() && !value.isUndelete()) {
           originalPuts.add(value);
         } else {
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -550,6 +550,7 @@ public class BlobStore implements Store {
             metrics.deleteAuthorizationFailureCount.inc();
           }
         }
+        short revisedLifeVersion = info.getLifeVersion();
         if (info.getLifeVersion() == MessageInfo.LIFE_VERSION_FROM_FRONTEND) {
           // This is a delete request from frontend
           if (value.isDelete()) {
@@ -557,6 +558,7 @@ public class BlobStore implements Store {
                 "Cannot delete id " + info.getStoreKey() + " since it is already deleted in the index.",
                 StoreErrorCodes.ID_Deleted);
           }
+          revisedLifeVersion = value.getLifeVersion();
         } else {
           // This is a delete request from replication
           if (value.isDelete() && value.getLifeVersion() == info.getLifeVersion()) {
@@ -571,7 +573,7 @@ public class BlobStore implements Store {
           }
         }
         indexValuesPriorToDelete.add(value);
-        lifeVersions.add(info.getLifeVersion());
+        lifeVersions.add(revisedLifeVersion);
         if (!value.isDelete() && !value.isUndelete()) {
           originalPuts.add(value);
         } else {
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -251,8 +251,9 @@ public class BlobStore implements Store {
             TimeUnit.MINUTES.toMillis(config.storeStatsRecentEntryProcessingIntervalInMinutes);
         blobStoreStats =
             new BlobStoreStats(storeId, index, config.storeStatsBucketCount, bucketSpanInMs, logSegmentForecastOffsetMs,
-                queueProcessingPeriodInMs, config.storeStatsWaitTimeoutInSecs, time, longLivedTaskScheduler,
-                taskScheduler, diskIOScheduler, metrics);
+                queueProcessingPeriodInMs, config.storeStatsWaitTimeoutInSecs,
+                config.storeEnableBucketForLogSegmentReports, time, longLivedTaskScheduler, taskScheduler,
+                diskIOScheduler, metrics);
         checkCapacityAndUpdateReplicaStatusDelegate();
         logger.trace("The store {} is successfully started", storeId);
         onSuccess();
@@ -526,7 +527,8 @@ public class BlobStore implements Store {
     checkDuplicates(infosToDelete);
     final Timer.Context context = metrics.deleteResponse.time();
     try {
-      List<IndexValue> indexValuesToDelete = new ArrayList<>();
+      List<IndexValue> indexValuesPriorToDelete = new ArrayList<>();
+      List<IndexValue> originalPuts = new ArrayList<>();
       List<Short> lifeVersions = new ArrayList<>();
       Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
       maybeCallInDeleteBetweenGetEndOffsetAndFindKey();
@@ -555,7 +557,7 @@ public class BlobStore implements Store {
                 "Cannot delete id " + info.getStoreKey() + " since it is already deleted in the index.",
                 StoreErrorCodes.ID_Deleted);
           }
-          indexValuesToDelete.add(value);
+          indexValuesPriorToDelete.add(value);
           lifeVersions.add(value.getLifeVersion());
         } else {
           // This is a delete request from replication
@@ -569,8 +571,14 @@ public class BlobStore implements Store {
                 "Cannot delete id " + info.getStoreKey() + " since it has a higher lifeVersion than the message info: "
                     + value.getLifeVersion() + ">" + info.getLifeVersion(), StoreErrorCodes.Life_Version_Conflict);
           }
-          indexValuesToDelete.add(value);
+          indexValuesPriorToDelete.add(value);
           lifeVersions.add(info.getLifeVersion());
+        }
+        if (!value.isDelete() && !value.isUndelete()) {
+          originalPuts.add(value);
+        } else {
+          originalPuts.add(index.findKey(info.getStoreKey(), new FileSpan(index.getStartOffset(), value.getOffset()),
+              EnumSet.of(PersistentIndex.IndexEntryType.PUT)));
         }
       }
       maybeCallBeforeSynchronization();
@@ -605,6 +613,7 @@ public class BlobStore implements Store {
                     StoreErrorCodes.Life_Version_Conflict);
               }
             }
+            indexValuesPriorToDelete.set(i, value);
             i++;
           }
         }
@@ -634,7 +643,9 @@ public class BlobStore implements Store {
           IndexValue deleteIndexValue =
               index.markAsDeleted(info.getStoreKey(), fileSpan, null, info.getOperationTimeMs(), info.getLifeVersion());
           endOffsetOfLastMessage = fileSpan.getEndOffset();
-          blobStoreStats.handleNewDeleteEntry(deleteIndexValue, indexValuesToDelete.get(correspondingPutIndex++));
+          blobStoreStats.handleNewDeleteEntry(info.getStoreKey(), deleteIndexValue,
+              originalPuts.get(correspondingPutIndex), indexValuesPriorToDelete.get(correspondingPutIndex));
+          correspondingPutIndex++;
         }
         logger.trace("Store : {} delete has been marked in the index ", dataDir);
       }
@@ -785,6 +796,7 @@ public class BlobStore implements Store {
       // Check if the undelete record is valid.
       index.validateSanityForUndelete(id, values, lifeVersionFromMessageInfo);
       IndexValue latestValue = values.get(0);
+      IndexValue originalPut = values.get(values.size() - 1);
       if (!IndexValue.hasLifeVersion(revisedLifeVersion)) {
         revisedLifeVersion = (short) (latestValue.getLifeVersion() + 1);
       }
@@ -838,7 +850,7 @@ public class BlobStore implements Store {
         // we still use lifeVersion from message info here so that we can re-verify the sanity of undelete request in persistent index.
         IndexValue newUndelete = index.markAsUndeleted(info.getStoreKey(), fileSpan, null, info.getOperationTimeMs(),
             lifeVersionFromMessageInfo);
-        blobStoreStats.handleNewUndeleteEntry(newUndelete);
+        blobStoreStats.handleNewUndeleteEntry(info.getStoreKey(), newUndelete, originalPut, latestValue);
       }
       onSuccess();
       return revisedLifeVersion;
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -810,7 +810,7 @@ class BlobStoreStats implements StoreStats, Closeable {
       processDeleteUpdateLogSegmentHelper(results, key, previousValue, originalPutValue, SUBTRACT);
     } else if (previousValue.isUndelete()) {
       // Previous IndexValue is an UNDELETE, it should already
-      // 1. Added it's size to the first bucket of the scan result
+      // 1. Added it's size to the base bucket of the scan result
       // we need to recover from this operation
       results.updateLogSegmentBaseBucket(previousValue.getOffset().getName(), SUBTRACT * previousValue.getSize());
     }
@@ -890,16 +890,13 @@ class BlobStoreStats implements StoreStats, Closeable {
     // Current UNDELETE is always valid
     results.updateLogSegmentBaseBucket(undeleteValue.getOffset().getName(), undeleteValue.getSize());
     if (previousValue.isDelete()) {
-      // Previous IndexValue is a DELETE, it should already
-      // 1. Added it's size to the first bucket in the scan result
-      // 2. Subtracted original PUT's size from the previousValue.getOperationTimeInMs() bucket.
-      // We need to recover from these two operations, so
-      // 1. Subtract previous DELETE's size from first bucket in the scan result.
-      // 2. Add original PUT's size to the previousValue.getOperationTimeInMs() bucket.
+      // Previous IndexValue is a DELETE, it will be invalidated by current DELETE. So we have to revert the previous
+      // DELETE here as if it doesn't exist. After reverting changes made by previous DELETE, we will process the current
+      // DELETE IndexValue.
       processDeleteUpdateLogSegmentHelper(results, key, previousValue, originalPutValue, SUBTRACT);
     } else if (previousValue.isUndelete()) {
       // Previous IndexValue is an UNDELETE, it should already
-      // 1. Added it's size to the first bucket of the scan result
+      // 1. Added it's size to the base bucket of the scan result
       // we need to recover from this operation
       results.updateLogSegmentBaseBucket(previousValue.getOffset().getName(), SUBTRACT * previousValue.getSize());
     }
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -983,11 +983,12 @@ class BlobStoreStats implements StoreStats, Closeable {
     } else if (!indexValue.isDelete() && !indexValue.isUndelete() && indexValue.isTtlUpdate()) {
       IndexFinalState finalState = keyFinalStates.get(indexEntry.getKey());
       if (finalState != null && finalState.isDelete()) {
+        Offset beginningOfThisLogSegment = new Offset(indexValue.getOffset().getName(), LogSegment.HEADER_SIZE);
         IndexValue putValue =
-            index.findKey(indexEntry.getKey(), new FileSpan(index.getStartOffset(), indexValue.getOffset()),
+            index.findKey(indexEntry.getKey(), new FileSpan(beginningOfThisLogSegment, indexValue.getOffset()),
                 EnumSet.of(PersistentIndex.IndexEntryType.PUT));
-        // We know the putValue exists, otherwise, TTL_UPDATE will not be valid
-        if (putValue.getOffset().getName().equals(indexValue.getOffset().getName())) {
+        if (putValue != null) {
+          // When the Put IndexValue is in the same log segment as the TTL_UPDATE value, then TTL_UPDATE value is invalid.
           handleLogSegmentDeletedBucketUpdate(results, indexValue, finalState.getOperationTime(), SUBTRACT);
         }
       }
@@ -1029,7 +1030,7 @@ class BlobStoreStats implements StoreStats, Closeable {
           entryCount = queueEntryCount.get();
         }
         long processStartTimeMs = time.milliseconds();
-        for (int i = 0; i < entryCount && !cancelled; i++) {
+        for (int i = 0; i < entryCount && !cancelled && !isScanning; i++) {
           EntryContext entryContext = recentEntryQueue.poll();
           if (entryContext == null) {
             throw new IllegalStateException(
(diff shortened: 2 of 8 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -572,6 +574,12 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
           indexValuesToDelete.add(value);
           lifeVersions.add(info.getLifeVersion());
         }
+        if (!value.isDelete() && !value.isUndelete()) {
```
jsjtzyy, line 577: Why TtlUpdate is also counted here?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -605,6 +613,7 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
                     StoreErrorCodes.Life_Version_Conflict);
               }
             }
+            indexValuesToDelete.set(i, value);
```
jsjtzyy, line 616: minor: can we call this `indexValuePriorToNewDelete` or the more descriptive name?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: Because we use findKey to get the value, and in findKey, we are not returning TTL_UPDATE record.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -818,6 +830,7 @@ public short undelete(MessageInfo info) throws StoreException {
           IndexValue value = index.findKey(info.getStoreKey(), fileSpan,
               EnumSet.of(PersistentIndex.IndexEntryType.DELETE, PersistentIndex.IndexEntryType.UNDELETE));
           if (value != null) {
+            latestValue = value;
```
jsjtzyy, line 833: Where is `latestValue` used? In this `if` block, no matter what the `value` is, the StoreException will be thrown and `latestValue` is never used.

ambry-store/src/main/java/com/github/ambry/store/ScanResults.java
```diff
@@ -26,46 +26,124 @@
...
+  // This is how delta bucket works. We can pick any point of time and add base validSize and all the delta value in those
+  // buckets prior to this point of time and get the answer we need.
 
+  // Container buckets are very close to the example above, except that it doesn't aggregate all containers' valid data,
+  // rather it keeps them in a map. The key of this map is the AccountId, and the value of this map is yet another map,
+  // whose key is the ContainerId and the value is the valid size of this container.
+  // So {@code containerBaseBucket} is base value for all containers and the {@code containerBuckets} is the delta values
+  // on each time bucket.
+  private final Map<String, Map<String, Long>> containerBaseBucket = new HashMap<>();
+  private final NavigableMap<Long, Map<String, Map<String, Long>>> containerBuckets = new TreeMap<>();
```
jsjtzyy, line 82: Let's call this `containerDeltaBuckets`.  Btw, what does `bucketCount` comprise? Is bucketCount = 1 baseBucket + number of delta buckets?

ambry-store/src/main/java/com/github/ambry/store/ScanResults.java
```diff
@@ -26,46 +26,124 @@
...
+    logSegmentForecastStartTimeMsForExpired = startTimeInMs;
+    long bucketTimeMs = logSegmentForecastStartTimeMsForExpired + bucketSpanInMs;
+    for (int i = 1; i < bucketCount; i++) {
+      logSegmentBucketsDeltaForExpired.put(bucketTimeMs, new TreeMap<>(LogSegmentNameHelper.COMPARATOR));
+      bucketTimeMs += bucketSpanInMs;
+    }
+    logSegmentLastBucketTimeMsForExpired =
+        logSegmentBucketsDeltaForExpired.isEmpty() ? logSegmentForecastStartTimeMsForExpired
+            : logSegmentBucketsDeltaForExpired.lastKey();
+    logSegmentForecastEndTimeMsForExpired = logSegmentLastBucketTimeMsForExpired + bucketSpanInMs;
```
jsjtzyy, line 135: `containerForecastEndTimeMs`,  `logSegmentForecastEndTimeMsForExpired` and  `logSegmentForecastEndTimeMsForDeleted` all seem to be extending another `bucketSpanInMs` from last bucket time. Can you explain why? We can talk this offline if needed.

ambry-store/src/main/java/com/github/ambry/store/ScanResults.java
```diff
@@ -26,46 +26,124 @@
...
+  // 1. LogSegment has two delta buckets. one for deleted blobs, the other for expired blobs.
+  // 2. Since we have a retention period for deleted blob, deleted delta's first bucket would start earlier than expired
+  //    delta.
+  // The reason to have two delta buckets is because deleted blobs and expired blobs are invalidated at different time.
+  // When a blob is deleted, we have to wait for a retention duration(7 days in prod) to invalidate it. But when a blob
+  // is expired, it's invalidated right aways. So when calculating valid size for log segment, we have to pass two time
+  // stamps, a delete reference time, usually is now - 7 days, and a expiry time, usually is now.
+  private final NavigableMap<String, Long> logSegmentBaseBucket = new TreeMap<>();
+  private final NavigableMap<Long, NavigableMap<String, Long>> logSegmentBucketsDeltaForDeleted = new TreeMap<>();
+  private final NavigableMap<Long, NavigableMap<String, Long>> logSegmentBucketsDeltaForExpired = new TreeMap<>();
```
jsjtzyy, line 101: I think the maps in this class are better to concurrent map as IndexScanner and QueueProcessor may concurrently update the map.

jsjtzyy (commented):
Continuing to review.

ambry-store/src/main/java/com/github/ambry/store/ScanResults.java
justinlin-linkedin: QueueProcessor shouldn't be running at the same time with IndexScanner. IndexScanner would change `isScanning` to true while scanning the index files. And when the `isScanning` is true, QueueProcessor won't do anything.

ambry-store/src/main/java/com/github/ambry/store/ScanResults.java
justinlin-linkedin: sure.

So when bucketCount is 1, we only have baseBucket.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: you are right, will update.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -493,17 +526,18 @@ private void enqueueNewValue(IndexValue newValue, IndexValue originalPutValue) {
...
    *                         when encountering PUT IndexValue. This method iterates through IndexValues from most recent one to
    *                         earliest one, so PUT IndexValue is the last IndexValue for the same key.
    * @param validIndexEntryAction the action to take on each valid {@link IndexEntry} found.
    * @throws StoreException if there are problems reading the index.
    */
-  private void forEachValidIndexEntry(IndexSegment indexSegment, long referenceTimeInMs,
-      Map<StoreKey, IndexFinalState> keyFinalStates, boolean removeFromStates, IndexEntryAction validIndexEntryAction)
-      throws StoreException {
+  private void forEachValidIndexEntry(IndexSegment indexSegment, long deleteReferenceTimeInMs,
+      long expiryReferenceTimeInMs, Map<StoreKey, IndexFinalState> keyFinalStates, boolean removeFromStates,
```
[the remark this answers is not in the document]
justinlin-linkedin: will update

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -139,6 +140,7 @@ static StatsSnapshot convertStoreUsageToContainerStatsSnapshot(Map<String, Map<S
     this.logSegmentForecastOffsetMs = logSegmentForecastOffsetMs;
     this.waitTimeoutInSecs = waitTimeoutInSecs;
     this.metrics = metrics;
+    this.enableBucketForLogSegmentReports = enableBucketForLogSegmentReports;
 
     if (bucketCount > 0) {
```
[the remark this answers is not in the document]
justinlin-linkedin: I am thinking about 24 buckets. One bucket for 10 minutes and 4 hours. I am willing to extend to 8 or 10 hours. So 24, 48 or 60 are all good options for me.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: because then the TTL_UPDATE flag is true, value.isPut will be false. findKey has this weird logic where it will populate TTL_UPDATE flag even if we are not trying to find TTL_UPDATE record. 

So when we have PUT, TTL_UPDATE and DELETE for this blob and we call findKey(blobId, [INDEX_TYPE.PUT]),  we will get a PUT value with TTL_UPDATE flag being true. And if we do value.isPut, it will return false.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -569,9 +571,15 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
                 "Cannot delete id " + info.getStoreKey() + " since it has a higher lifeVersion than the message info: "
                     + value.getLifeVersion() + ">" + info.getLifeVersion(), StoreErrorCodes.Life_Version_Conflict);
           }
-          indexValuesToDelete.add(value);
+          indexValuesPriorToDelete.add(value);
           lifeVersions.add(info.getLifeVersion());
```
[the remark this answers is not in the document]
justinlin-linkedin: will update

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -240,53 +242,74 @@ long getMaxBlobSize() {
...
-        scanLock.lock();
-        try {
-          if (isScanning) {
-            if (waitCondition.await(waitTimeoutInSecs, TimeUnit.SECONDS)) {
+    if (enableBucketForLogSegmentReports) {
+      if (referenceTimeInMs != REF_TIME_OUT_OF_BOUNDS) {
+        retValue = currentScanResults.getValidSizePerLogSegment(referenceTimeInMs, expiryReferenceTime);
+      } else {
+        if (isScanning && getLogSegmentRefTimeMs(indexScanner.newScanResults, timeRange) != REF_TIME_OUT_OF_BOUNDS) {
+          scanLock.lock();
```
jsjtzyy, line 274: I am actually confused about this `scanLock`. If IndexScanner is running (which means `isScanning == true`), current thread is able to acquire and hold `scanLock` (because it can be released by IndexScanner at line 1089). After that, current thread can be blocked by `waitCondition` (without releasing lock), however, IndexScanner has to re-acquire the  lock before signaling the `waitCondition` (see line 1141 - 1146).  It seems like a deadlock except for `waitCondition` can time out eventually. Is this intended?

jsjtzyy (commented):
a few more comments

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: waitCondition is a condition on scanLock, when calling waitCondition.await, it will release the scan lock. This is pretty much how the condition and lock work.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -1054,18 +1235,19 @@ private void processIndexSegmentEntriesBackward(IndexSegment indexSegment, Predi
           indexSegment.getFile().getName(), storeId);
 
       // valid index entries wrt log segment reference time
-      forEachValidIndexEntry(indexSegment, newScanResults.logSegmentForecastStartTimeMs, keyFinalStates, false,
-          entry -> {
+      forEachValidIndexEntry(indexSegment, newScanResults.logSegmentForecastStartTimeMsForDeleted,
+          newScanResults.logSegmentForecastStartTimeMsForExpired, keyFinalStates, false, entry -> {
```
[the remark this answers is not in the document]
justinlin-linkedin: `removeFinalStateOnPut` will remove the final state of a storeKey when encountering PUT record, since PUT record is the first record for the storeKey. This is for reducing memory usage. We don't want to keep final states of all store key in the map throughout the iteration of all index segment. 

However, in this method, we will process same index segment twice, first time for calculating valid size of log segment, second time for calculating valid size for container. So when calculating for log segment, we don't remove the final state  since later we would need the final state when calculating for containers.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -616,11 +657,11 @@ private boolean isTtlUpdateEntryValid(StoreKey key, IndexValue ttlUpdateValue, l
    * is returned
    */
   private long getLogSegmentRefTimeMs(ScanResults results, TimeRange timeRange) {
```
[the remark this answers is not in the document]
justinlin-linkedin: updated.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -665,11 +706,12 @@ private void updateMapHelper(Map<String, Long> map, String key, Long value) {
    * Helper function for container buckets for blob expiration/deletion related updates.
    * @param results the {@link ScanResults} to be updated
    * @param indexValue the PUT {@link IndexValue} of the expiring/deleting blob
-   * @param expOrDelTimeInMs either the expiration or deletion time of the blob in ms
+   * @param expOrDelTimeInMs deletion time of the blob in ms
```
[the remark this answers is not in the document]
justinlin-linkedin: that was a mistake. updated.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -706,46 +765,132 @@ private void processNewPut(ScanResults results, IndexValue putValue) {
...
+    }
+    // Now deal with log segment bucket
+    // Current DELETE is always valid
+    if (previousValue.isDelete()) {
+      // Previous IndexValue is a DELETE, it should already
+      // 1. Added it's size to the first bucket in the scan result
+      // 2. Subtracted original PUT's size from the previousValue.getOperationTimeInMs() bucket.
+      // We need to recover from these two operations, so
+      // 1. Subtract previous DELETE's size from bucket in the scan result.
+      // 2. Add original PUT's size to the previousValue.getOperationTimeInMs() bucket.
```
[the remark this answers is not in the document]
justinlin-linkedin: if we are under this if statement, then we know the current IndexValue is a DELETE and the previous IndexValue is also a DELETE, the IndexValue history can go like this, PUT, DELETE(v0), DELETE(v1).

When we see the DELETE(v0), we will 1) add size of DELETE to base bucket, then 2)subtract size of PUT int the DELETE's operation time bucket. 
When we see the DELETE(v1), we know it will invalidate DELETE(v0). So one way to do this is to revert what DELETE(v0) did before, so we 1)subtract size of DELETE from base bucket, then 3)add size of PUT in the DELETE(v0)'s operation time bucket. After doing this, we can treat DELETE(v0) just like it never happened. Then we move on to deal with DELETE(v1).

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
jsjtzyy: Got it, I didn't notice `waitCondition` will release the lock.

zzmao:
BlobStore changes looks good to me.

After spending some time to read the blobstats code, but I feel I am not qualified to review blobstats code 😢 . @SophieGuo410  probably a good reviewer for blobstats change.

justinlin-linkedin:
@SophieGuo410 adding you as third reviewer if you are interested.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
jsjtzyy: I see. Thanks for explanation. I think it probably justifies a comment here.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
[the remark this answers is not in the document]
jsjtzyy: Correct, this is what I thought about. My suggestion is to make the comments more accurate:
`Added it's size to the first bucket in the scan result` -> `Added it's size to the base bucket`;
`Subtract previous DELETE's size from bucket in the scan result` -> `Subtract previous DELETE's size from base bucket`;
Also add one sentence in the end something like:
```
After recovering changes made by previous DELETE, we update both base bucket and operation time bucket for new DELETE.
```

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -706,46 +766,132 @@ private void processNewPut(ScanResults results, IndexValue putValue) {
...
+      // This previous value is not DELETE, current UNDELETE would not change container bucket.
+      // There are several possibility when the previous value is not DELETE.
+      // 1. PUT UNDELETE: UNDELETE doesn't change any bucket
+      // 2. PUT TTL_UPDATE UNDELETE: UNDELETE doesn't change any bucket
+      // 3. PUT DELETE UNDELETE UNDELETE: the last UNDELETE doesn't change any bucket
+      // UNDELETE would resurrect a deleted IndexValue, if the previous value is not DELETE, then the IndexValue doesn't
+      // need to be resurrected.
+    } else {
+      handleContainerBucketUpdate(results, originalPutValue, undeleteValue.getOperationTimeInMs(), ADD);
+      handleContainerBucketUpdate(results, originalPutValue, originalPutValue.getExpiresAtMs(), SUBTRACT);
```
jsjtzyy, line 886: I understand this part and just want to confirm again about UNDELETE. The UNDELETE operation always checks if original PUT has expired. So, if we reach here, it means original PUT hasn't expired and should be subtracted, correct?

I am asking this because I am thinking about a case where original PUT expires before DELETE and there is an UNDELETE in the end. In this case, original PUT is subtracted twice but I think this case would never happen.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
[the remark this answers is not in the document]
justinlin-linkedin: I reword the comments, please take another look, see if it's good for you.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -706,46 +766,132 @@ private void processNewPut(ScanResults results, IndexValue putValue) {
...
+    // Current UNDELETE is always valid
+    results.updateLogSegmentBaseBucket(undeleteValue.getOffset().getName(), undeleteValue.getSize());
+    if (previousValue.isDelete()) {
+      // Previous IndexValue is a DELETE, it should already
+      // 1. Added it's size to the first bucket in the scan result
+      // 2. Subtracted original PUT's size from the previousValue.getOperationTimeInMs() bucket.
+      // We need to recover from these two operations, so
+      // 1. Subtract previous DELETE's size from first bucket in the scan result.
+      // 2. Add original PUT's size to the previousValue.getOperationTimeInMs() bucket.
+      processDeleteUpdateLogSegmentHelper(results, key, previousValue, originalPutValue, SUBTRACT);
```
jsjtzyy, line 899: Let's consider a case: PUT, DELETE, UNDELETE and PUT expires after UNDELETE op time. Based on my understanding, it will go through following steps:
```
 Operation  | Base bucket | Delete Op bucket | Undelete Op bucket | Expired Bucket
    DELETE   | + D size        |  - P size                 |                                     |  + P size
 UNDELETE | - D + U size  |  + P size                |                                     |  + P size
```
Looks like the expired bucket adds PUT record size twice, did I misunderstand anything?

jsjtzyy (commented):
Two more comments, other code looks good.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: if the current IndexValue is UNDELETE and the previous IndexValue is DELETE, then we will first revert the change made by DELETE, so for UNDELETE, at Expired Bucket, it will do -P size, not, +P size.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: Your first statement is correct, and there is no chance when the PUT is expired before DELETE and UNDELETE would succeeds. If we have PUT and DELETE for the blob, but PUT is already expired, then UNDELETE will fail since it can't undelete an expired PUT.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -799,17 +963,33 @@ private void processEntryForLogSegmentBucket(ScanResults results, IndexEntry ind
...
       }
-      if (!indexValue.isTtlUpdate() || !isTtlUpdateEntryValid(indexEntry.getKey(), indexValue,
-          expOrDelTimeInMs == Utils.Infinite_Time ? expOrDelTimeInMs : expOrDelTimeInMs + 1, keyFinalStates)) {
-        handleLogSegmentBucketUpdate(results, indexValue, expOrDelTimeInMs, SUBTRACT);
+    } else if (!indexValue.isDelete() && !indexValue.isUndelete() && indexValue.isTtlUpdate()) {
+      IndexFinalState finalState = keyFinalStates.get(indexEntry.getKey());
+      if (finalState != null && finalState.isDelete()) {
+        IndexValue putValue =
+            index.findKey(indexEntry.getKey(), new FileSpan(index.getStartOffset(), indexValue.getOffset()),
+                EnumSet.of(PersistentIndex.IndexEntryType.PUT));
```
jsjtzyy, line 988: I understand the piece of code is to check whether TtlUpdate is valid or not. If PUT and TtlUpdate are in same log segment and final state is delete, then TtlUpdate is not valid. 
I wonder if we can search the log segment of TtlUpdate only instead of findKey from the very beginning of whole PersistentIndex.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -799,17 +963,33 @@ private void processEntryForLogSegmentBucket(ScanResults results, IndexEntry ind
...
+        if (expiresAtMs < deleteAtMs) {
+          handleLogSegmentExpiredBucketUpdate(results, indexValue, expiresAtMs, SUBTRACT);
+        } else {
+          handleLogSegmentDeletedBucketUpdate(results, indexValue, deleteAtMs, SUBTRACT);
+        }
       }
-      if (!indexValue.isTtlUpdate() || !isTtlUpdateEntryValid(indexEntry.getKey(), indexValue,
-          expOrDelTimeInMs == Utils.Infinite_Time ? expOrDelTimeInMs : expOrDelTimeInMs + 1, keyFinalStates)) {
-        handleLogSegmentBucketUpdate(results, indexValue, expOrDelTimeInMs, SUBTRACT);
+    } else if (!indexValue.isDelete() && !indexValue.isUndelete() && indexValue.isTtlUpdate()) {
```
jsjtzyy, line 983: Since the `indexValue` is not from findKey() method, then I believe this can be simplified to `indexValue.isTtlUpdate()`.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -850,26 +1030,29 @@ public void run() {
         }
         long processStartTimeMs = time.milliseconds();
         for (int i = 0; i < entryCount && !cancelled; i++) {
```
jsjtzyy, line 1032: I wonder if it's worthwhile to add an additional check in this "for" statement. That is, checking if there is IndexScanner is ongoing. 
```
for (int i = 0; i < entryCount && !cancelled &&  !isScanning; i++)
```
It's possible `QueueProcessor` starts before `IndexScanner`. When `IndexScanner` may begin while `QueueProcessor` is still ongoing.

jsjtzyy (approved):
LGTM after remaining comments are addressed.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: that's a good idea, updated.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: A DELETE or UNDELETE IndexValue can still return true when calling isTtlUpdate(), so there we have to do this kinda check.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
justinlin-linkedin: that's right, will put a check just in case.

zzmao (approved):
LGTM

PR #1614, opened 2020-08-31
[RANGE_REQUEST] add option for always resolving ranges on empty blobs

cgtz:
Some clients make preemptive range requests against blobs. Since the
client does not know the blob size without making at least one request,
they could be making a range request against an empty (0-byte) blob. All
range types with a start offset will fail with a 416 since byte 0 is
already past the end of an empty blob.

To keep this behavior for most users (since changing range request
behavior can cause CDN interaction challenges), this PR introduces an
optional request header that allows the client to tell ambry-frontend to
that it wants a success response when any range request is made against
an empty blob.


ankagrawal (approved):
LGTM

zzmao (approved):
LGTM

PR #1615, opened 2020-09-03
[LeaderBasedReplication] Notify response handler on connection check out failure in all places.

Arun-LinkedIn:
Issue: While fetching missing keys for timed out standby replicas, we have following issues:
1. Response handler is not notified on connection checkout failure. Due to that remote node might not be marked as down. 
2. All remote hosts are being iterated (including ones which are marked as down) instead of ‘Active replicas’.

Due to this, we see frequent retries for connection checkout with remote hosts which are down.

This PR addresses above two issues.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -357,6 +357,7 @@ public class ReplicaThread implements Runnable {
 
       // Get a list of active replicas that needs be included for this replication cycle
       List<RemoteReplicaInfo> activeReplicasPerNode = new ArrayList<>();
+      List<RemoteReplicaInfo> standbyReplicasWithNoProgress = new ArrayList<>();
       for (RemoteReplicaInfo remoteReplicaInfo : replicasToReplicatePerNode) {
         ReplicaId replicaId = remoteReplicaInfo.getReplicaId();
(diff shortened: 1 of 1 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -452,11 +446,12 @@ public void replicate() {
           // TODO: If the result to fetch a blob from local dc is Blob_Not_Found, then we can fetch it from replicas in remote datacenter.
           // This will involve co-ordination between replica threads containing replicas of same partition.
           List<RemoteReplicaInfo> standbyReplicasTimedOutOnNoProgress =
-              getRemoteStandbyReplicasTimedOutOnNoProgress(replicasToReplicatePerNode);
+              getRemoteStandbyReplicasTimedOutOnNoProgress(standbyReplicasWithNoProgress);
           if (standbyReplicasTimedOutOnNoProgress.size() > 0) {
             allCaughtUp = false;
             currentReplicaList = standbyReplicasTimedOutOnNoProgress;
             if (connectedChannel == null) {
+              checkoutConnectionTimeInMs = -1;
```
jsjtzyy, line 454: minor: line 458 `startTimeInMs` is not updated before checking out the channel

jsjtzyy (approved):
LGTM after a minor comment is addressed.

cgtz (approved):

PR #1622, opened 2020-09-10
Implementation of storage node throttling interfaces

ankagrawal:


ambry-commons/src/main/java/com/github/ambry/commons/Throttler.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.commons;
+
+public abstract class Throttler {
```
zzmao, line 16: We have another Throttler in com.github.ambry.utils. Is it possible to reuse or enhance that one?

zzmao:
Per out discussion before, my impression was to do throttling on router side. Storage node just needs to reports load status.

If we do throttling on server side, frontend spends time to wait response but get 429, then it needs to request  from another node. Latency will be worse in this case.

ankagrawal:
Going to create a new PR on a new branch for this.

PR #1624, opened 2020-09-12
[Container Scalability] Basic implementation of MySqlAccountService 

Arun-LinkedIn:
Functionalities implemented:
a) Fetch all accounts and container metadata from DB and initialize cache on boot-up,
b) Implement AccountService interface update APIs to write changes to mysql database and getter APIs to read changes from cache
c) Scheduler thread to sync changes from DB
d) Add MySqlAccountServiceFactory and MySqlAccountServiceConfig


ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -0,0 +1,315 @@
...
+      logger.error("Duplicate account id or name exist in the accounts to update");
+      //accountServiceMetrics.updateAccountErrorCount.inc();
+      return false;
+    }
+
+    // Make a pre check for conflict between the accounts to update and the accounts in the local cache. Will fail this
+    // update operation for all the accounts if any conflict exists. For existing accounts, there is a chance that the account to update
+    // conflicts with the accounts in the local cache, but does not conflict with those in the MySql database. This
+    // will happen if some accounts are updated but the local cache is not yet refreshed.
+    // TODO: Once we have APIs (and versioning) for updating containers, we will need to check conflicts for containers as well.
```
lightningrob, line 194: I think we can check container name/id conflicts the same as accounts, even without versioning.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -0,0 +1,315 @@
...
+    rwLock.readLock().lock();
+    try {
+      if (accountInfoMap.hasConflictingAccount(accounts)) {
+        logger.error("Accounts={} conflict with the accounts in local cache. Cancel the update operation.", accounts);
+        //accountServiceMetrics.updateAccountErrorCount.inc();
+        return false;
```
lightningrob, line 200: I really wish the interface method threw an exception.  The caller has no way to tell if it failed due to bad input or system error.  We should revisit this later.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -0,0 +1,315 @@
...
+      return false;
+    }
+
+    if (mySqlAccountStore == null) {
+      logger.info("MySql Account DB store is not accessible");
+      return false;
+    }
+
+    if (config.updateDisabled) {
+      logger.info("Updates has been disabled");
```
lightningrob, line 180: Minor: have been.

lightningrob (changes requested):
Looks good overall.  Will wait for the tests.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountStore.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+  }
+
+  /**
+   * Gets all {@link Account}s that have been created or modified since the specified time.
+   * @param updatedSince the last modified time used to filter.
+   * @return a list of {@link Account}s
+   * @throws SQLException
+   */
+  public List<Account> getNewAccounts(long updatedSince) throws SQLException {
```
jsjtzyy, line 90: Can we change the return type from List to Collection in this file?

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -0,0 +1,315 @@
...
+  /**
+   * Initialize in-memory cache by fetching all the {@link Account}s and {@link Container}s metadata records.
+   * It consists of 2 steps:
+   * 1. Check local disk for back up copy and load metadata and last modified time of Accounts/Containers into cache.
+   * 2. Fetch added/modified accounts and containers from mysql database since the last modified time (found in step 1)
+   *    and load into cache.
+   */
+  private void initialFetchAndSchedule() {
+
+    // TODO: Check local disk for back up copy and load metadata and last modified time into cache.
```
jsjtzyy, line 97: Just a reminder that, for hadoop frontend, we are using rain which doesn't allow us to back up copy on disk. This can be solved by disabling backup based on`BACKUP_DIRECTORY_KEY`.  So, when backup is unavailable and MySQL connection failed, I think we should block the startup.

jsjtzyy (commented):
Initial comments.

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  @Config(UPDATE_DISABLED)
+  @Default("false")
+  public final boolean updateDisabled;
+
+  public MySqlAccountServiceConfig(VerifiableProperties verifiableProperties) {
+    dbUrl = verifiableProperties.getString(DB_URL, "");
+    dbUser = verifiableProperties.getString(DB_USER, "");
+    dbPassword = verifiableProperties.getString(DB_PASSWORD, "");
+    updaterPollingIntervalMs =
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure, true. Updated default to 1 minute and changed granularity to seconds.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
lightningrob: @jsjtzyy Does hadoop-frontend fail to start up if it can't retrieve the account metadata from ZK?
It sounds like it will have to query MySql for the entire data set on every restart.

cgtz: I thought that the issue was just that the directory paths that the container had access to were different in that environment but reading/writing to disk was still possible?

jsjtzyy: Oh my bad, I recall it now. We are actually using a different directory path. Sorry for misleading comment, please ignore it.

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
```diff
@@ -0,0 +1,94 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.config;
+
+/**
+ * Config for {@link MySqlAccountServiceConfig}
```
jsjtzyy, line 17: minor: `Configs for {@link MySqlAccountService}`

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
```diff
@@ -0,0 +1,94 @@
...
+  public final boolean updateDisabled;
+
+  public MySqlAccountServiceConfig(VerifiableProperties verifiableProperties) {
+    dbUrl = verifiableProperties.getString(DB_URL);
+    dbUser = verifiableProperties.getString(DB_USER);
+    dbPassword = verifiableProperties.getString(DB_PASSWORD);
+    updaterPollingIntervalSeconds =
+        verifiableProperties.getIntInRange(UPDATER_POLLING_INTERVAL_SECONDS, 60, 0, Integer.MAX_VALUE);
+    updaterShutDownTimeoutSeconds =
+        verifiableProperties.getIntInRange(UPDATER_SHUT_DOWN_TIMEOUT_SECONDS, 5, 1, Integer.MAX_VALUE);
```
jsjtzyy, line 90: Is 5 seconds sufficient to shutdown the scheduler in MySqlAccountService if there is an ongoing task performing "fetchAndUpdate" with Mysql?

jsjtzyy (approved):
LGTM with two minor comments.

ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceTest.java
```diff
@@ -0,0 +1,336 @@
...
+public class MySqlAccountServiceTest {
+
+  AccountService mySqlAccountService;
+  Properties mySqlConfigProps = new Properties();
+
+  public MySqlAccountServiceTest() {
+    resetConfig();
+  }
+
+  private void resetConfig() {
```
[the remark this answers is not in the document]
lightningrob: Also, instead of calling this at the end of each method, you could add the @Before tag to have it called before each test automatically (probably need to make it public).

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
```diff
@@ -72,13 +81,13 @@
...
-    dbPassword = verifiableProperties.getString(DB_PASSWORD, "");
-    updaterPollingIntervalMs =
-        verifiableProperties.getIntInRange(UPDATER_POLLING_INTERVAL_MS_KEY, 2 * 1000, 0, Integer.MAX_VALUE);
-    updaterShutDownTimeoutMs =
-        verifiableProperties.getIntInRange(UPDATER_SHUT_DOWN_TIMEOUT_MS_KEY, 5 * 1000, 1, Integer.MAX_VALUE);
+    dbUrl = verifiableProperties.getString(DB_URL);
+    dbUser = verifiableProperties.getString(DB_USER);
+    dbPassword = verifiableProperties.getString(DB_PASSWORD);
+    updaterPollingIntervalSeconds =
+        verifiableProperties.getIntInRange(UPDATER_POLLING_INTERVAL_SECONDS, 60, 0, Integer.MAX_VALUE);
```
lightningrob, line 88: What would value of 0 mean?  Update continuously?

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -187,22 +172,30 @@ public boolean updateAccounts(Collection<Account> accounts) {
...
         logger.error("Accounts={} conflict with the accounts in local cache. Cancel the update operation.", accounts);
         //accountServiceMetrics.updateAccountErrorCount.inc();
         return false;
       }
+      for (Account account : accounts) {
+        if (accountInfoMap.hasConflictingContainer(account.getAllContainers(), account.getId())) {
+          logger.error(
+              "Containers={} under Account={} conflict with the containers in local cache. Cancel the update operation.",
+              account.getAllContainers(), account.getId());
+          //accountServiceMetrics.updateAccountErrorCount.inc();
```
lightningrob, line 189: Can we add the error metrics and check they get incremented in the tests?

lightningrob (approved):
Looks good.  Approved after minor changes.

lightningrob:
One general comment on the tests is that there is some duplicate logic between the Helix and MySql account service tests.  I realize there are setup differences, but it might be worth refactoring this in the near future to use a base test class with common test logic, plus hooks to setup/mock the dependencies that would be overridden.  Since AccountService is an interface, it would be useful to have a black-box test that verifies that a given impl behaves as expected.  We probably don't do this in general.

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
Arun-LinkedIn: Used value to 0 to actually disable periodic polling. We won't start scheduler thread if config.updaterPollingIntervalSeconds is 0.

jsjtzyy (approved):

lightningrob (approved):

PR #1626, opened 2020-09-16
[HTTP2] Return ByteBuf in GetResponse content() method

justinlin-linkedin:


jsjtzyy (approved):
Looks good with minor comments.

ambry-store/src/main/java/com/github/ambry/store/StoreMessageReadSet.java
```diff
@@ -159,12 +161,12 @@ public void close() {
    */
   void doPrefetch(long relativeOffset, long size) throws IOException {
     long sizeToRead = Math.min(size, getMessageInfo().getSize() - relativeOffset);
-    prefetchedData = ByteBuffer.allocate((int) sizeToRead);
-    getChannel().read(prefetchedData, offset.getOffset() + relativeOffset);
+    prefetchedData = PooledByteBufAllocator.DEFAULT.ioBuffer((int) sizeToRead);
```
zzmao, line 164: Just want to double check:

We prefer direct buffer here. I looked at ioBuffer() function, it returns direct buffer if `isDirectBufferPooled`. How to know value of `isDirectBufferPooled`?

ambry-protocol/src/main/java/com/github/ambry/protocol/GetResponse.java
```diff
@@ -139,14 +139,7 @@ protected void prepareBuffer() {
         int maxNumComponent = 1 + toSend.content().nioBufferCount();
         CompositeByteBuf compositeByteBuf = bufferToSend.alloc().compositeHeapBuffer(maxNumComponent);
         compositeByteBuf.addComponent(true, bufferToSend);
```
zzmao, line 141: Can we create a test to verify the content of the GetResponse?

zzmao (approved):
LGTM. 3 minors comments.

zzmao:
One more comment:  in AmbrySendToHttp2Adaptor.write(), there is check for dataContent == null or not. With this PR, this check can be removed.

ambry-store/src/main/java/com/github/ambry/store/StoreMessageReadSet.java
justinlin-linkedin: netty has a jvm property to control that, but by default, in linux box, it's enabled.

zzmao:
> One more comment: in AmbrySendToHttp2Adaptor.write(), there is check for dataContent == null or not. With this PR, this check can be removed.

How is this one? Do you plan to change it?

justinlin-linkedin:
@ankagrawal  I edited some code in CloudMessageReadSet. please take a look when you got time.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -840,11 +840,6 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
       // from concurrent operations.
       if (!successfullyDeserialized) {
         BlobData blobData = MessageFormatRecord.deserializeBlob(payload);
-        // Note that for segment GET there is only one chunk to get, so we can be sure that its this one.
```
justinlin-linkedin, line 843: Removing this because of three reasons
1. This method only deal with data chunk, and changing blob size in property is already done in metadata chunk.
2. There is a race condition when the frontend is writing header back to client but the blob size is changed at the same time.
3. Here the size is undecrypted size, we have to run decryption to get the real size. 
@ankagrawal

justinlin-linkedin:
> > One more comment: in AmbrySendToHttp2Adaptor.write(), there is check for dataContent == null or not. With this PR, this check can be removed.
> 
> How is this one? Do you plan to change it?

updated.

zzmao (approved):

zzmao:
@justinlin-linkedin  can you also add a metric for me? this metric is missed by http2.

```diff --git a/ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java b/ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
index 2920a87c6..721cc9add 100644
--- a/ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
@@ -15,6 +15,7 @@ package com.github.ambry.network;
 
 import com.github.ambry.network.http2.Http2ServerMetrics;
 import com.github.ambry.server.EmptyRequest;
+import com.github.ambry.utils.SystemTime;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
@@ -54,12 +55,14 @@ public class NettyServerRequestResponseChannel implements RequestResponseChannel
     if (!(originalRequest instanceof NettyServerRequest)) {
       throw new IllegalArgumentException("NetworkRequest should be NettyRequest");
     }
+    long startTime = SystemTime.getInstance().milliseconds();
     ChannelHandlerContext ctx = ((NettyServerRequest) originalRequest).getCtx();
     http2ServerMetrics.requestTotalProcessingTime.update(
         System.currentTimeMillis() - originalRequest.getStartTimeInMs());
     ChannelFutureListener channelFutureListener = new ChannelFutureListener() {
       @Override
       public void operationComplete(ChannelFuture future) throws Exception {
+        metrics.updateSendTime(SystemTime.getInstance().milliseconds() - startTime);
         if (!future.isSuccess()) {
           if (!(future.cause() instanceof ClosedChannelException)) {
             // TODO: a round solution is needed. For now, use same logic as Http2NetworkClient in frontend.
```

ankagrawal (approved):
lgtm!

PR #1628, opened 2020-09-22
[Container API] Introducing create/update container APIs in AccountService

jsjtzyy:
When number of containers in single account keeps growing, it becomes expensive to transmit whole account
to clients, edit on client side and send it back. This PR introduces some container management APIs (prototype)
to help create/update container within frontend to avoid unnecessary transmission overhead. Also, this PR
incorporates the logic to generate container id.
The addContainer implementation in HelixAccountService will be called by PostAccountsHandler in future PR.


ambry-frontend/src/main/java/com/github/ambry/frontend/PostAccountsHandler.java
```diff
@@ -17,6 +17,7 @@
 import com.github.ambry.account.Account;
 import com.github.ambry.account.AccountCollectionSerde;
 import com.github.ambry.account.AccountService;
+import com.github.ambry.commons.Callback;
```
lightningrob, line 20: Probably can omit this file from the PR.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -476,6 +480,24 @@ public static void setCurrentJsonVersion(short currentJsonVersion) {
     Container.currentJsonVersion = currentJsonVersion;
   }
 
+  /**
+   * Check if two containers' editable fields are same.
+   * @param container1 the {@link Container} to compare.
+   * @param container2 the {@link Container} to compare.
+   * @return {@code true} if two containers are equivalent in terms of editable fields. {@code false} otherwise.
+   */
+  public static boolean containerEditableFieldsAreSame(Container container1, Container container2) {
```
lightningrob, line 489: This is tricky since "editable" concept is not well defined in OS ambry.
Also, should make this instance method taking only one Container argument to compare.

ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -84,6 +84,7 @@
   static final boolean CACHEABLE_DEFAULT_VALUE = true;
   static final Set<String> CONTENT_TYPE_WHITELIST_FOR_FILENAMES_ON_DOWNLOAD_DEFAULT_VALUE = Collections.emptySet();
 
+  public static final short CONTAINER_ID_START_RANGE = 8;
```
lightningrob, line 87: This can be confusing to someone who is only aware of OS ambry.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -246,6 +250,69 @@ public boolean updateAccounts(Collection<Account> accounts) {
     return updateAccountsWithAccountMetadataStore(accounts, accountMetadataStore);
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  @Override
+  public Pair<String, String> addContainer(String accountName, Container newContainer) {
```
lightningrob, line 257: This entire method is generic except for the line that updates the underlying store.  Can we move it to the base class?  Maybe AccountService can also have a method like getMetadataStore().

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -107,6 +108,36 @@
...
+  }
+
+  /**
+   * Get an existing container from a given account.
+   * @param accountName the name of account which container belongs to.
+   * @param containerId the id of container to get.
+   * @return the requested {@link Container} object or null if not present.
+   */
+  default Container getContainer(String accountName, String containerId) {
+    throw new UnsupportedOperationException("This method is not supported");
```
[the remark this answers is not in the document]
lightningrob: It seems strange to have accountName and containerId in same signature.

lightningrob (changes requested):

ambry-frontend/src/main/java/com/github/ambry/frontend/PostAccountsHandler.java
jsjtzyy: Omitted

ambry-api/src/main/java/com/github/ambry/account/Container.java
jsjtzyy: Yeah, I changed it to instance method and renamed it to "isSameContainer"

ambry-api/src/main/java/com/github/ambry/account/Container.java
jsjtzyy: True, make this as a config

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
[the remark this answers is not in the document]
jsjtzyy: My bad, I was trying to use `containerName` but somehow it became containerId.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
jsjtzyy: Good point. The only blocker here is how to pass in a config to specify "container id start number". Or can we hard code it somewhere? (I am thinking if container id can start from 0)

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -246,6 +250,69 @@ public boolean updateAccounts(Collection<Account> accounts) {
...
+      throw new IllegalArgumentException("Account " + accountName + " is not found");
+    }
+    String accountId = String.valueOf(account.getId());
+    for (Container container : account.getAllContainers()) {
+      // make sure there is no conflict container (conflict means a container with same name but different attributes already exists).
+      if (container.getName().equals(newContainer.getName())) {
+        if (containerEditableFieldsAreSame(container, newContainer)) {
+          // If an exactly same container already exists, we directly return its id. Adding same container multiple times
+          // should be no-op.
+          return new Pair<>(accountId, String.valueOf(container.getId()));
```
cgtz, line 276: Why are the numerical IDs returned as a pair of strings instead of `Pair<Short, Short>`?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -246,6 +250,69 @@ public boolean updateAccounts(Collection<Account> accounts) {
...
+    // if code reaches here, it means no duplicate container in this account
+    // generate container id for new container
+    short nextContainerId = account.getAllContainers()
+        .stream()
+        .map(Container::getId)
+        .max(Short::compareTo)
+        .map(maxId -> (short) (maxId + 1))
+        .orElse(CONTAINER_ID_START_RANGE);
+    // construct a container based on input container and next containerId
+    Container containerToAdd = new ContainerBuilder(nextContainerId, newContainer.getName(), newContainer.getStatus(),
```
cgtz, line 292: `ContainerBuilder` has a copy constructor, could we use that here?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
jsjtzyy: I find containerId/accountId are `short` type, however, they are `int` in closed source code. Considering we may change the type to int in the future, so I want to use `String` here no matter what type of accountId/containerId are.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
jsjtzyy: Yep, we can. Thanks for the suggestion.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
lightningrob: Yes, we will need to change containerId to int.  AccountId can stay short; we may want to also make it int for consistency, if we don't run out of bits in blobId.

lightningrob: Since you added it to config, can you make the method generic now?

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -107,6 +108,36 @@
    */
   public boolean removeAccountUpdateConsumer(Consumer<Collection<Account>> accountUpdateConsumer);
 
+  /**
+   * Add a new container to an existing account.
+   * @param accountName the name of account in which new container will be added.
+   * @param newContainer the new container to add.
+   * @return a pair of account id and container id strings associated with existing account and new container.
+   */
+  default Pair<String, String> addContainer(String accountName, Container newContainer) {
```
lightningrob, line 117: The PostAccountsHandler will have method updateContainers that takes a collection of (new or updated) containers in the payload.  Would it make more sense to mirror that in this interface?  Otherwise the handler will need to get the existing containers, compare with the input and call add() and update() as appropriate.  Which may be okay.  But the MySqlAccountService should ideally make all the updates in a single transaction (in case multiple containers are added/updated).

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
lightningrob: Since you added it to config, can you make the method generic now?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
cgtz: why not change this to `Pair<Integer, Integer>` then? That would fit both shorts and ints and allow for future flexibility.

I would like to point out that changing container ID to an int isn't the simplest change. There are implications in the blob ID, blob properties, and index value binary formats. A new version of each one will likely be required along with a staged rollout.

jsjtzyy: I see the complexity of changing container id to `int`. For now, sure, I can change the return type to `Integer`.

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -107,6 +108,36 @@
...
+    throw new UnsupportedOperationException("This method is not supported");
+  }
+
+  /**
+   * Update attributes of an existing container.
+   * @param accountName the name of account which container belongs to.
+   * @param updatedContainer the updated {@link Container} used to replace current one.
+   * @return {@code true} if container is successfully updated. {@code false} otherwise.
+   */
+  default boolean updateContainer(String accountName, Container updatedContainer) {
```
lightningrob, line 127: How about calling this updateContainers (plural) and accepting a collection of containers?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
jsjtzyy: Sure, will do

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
jsjtzyy: Yeah, I actually renamed it already, will push out the change to github after rest changes are done.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -253,56 +253,62 @@ public boolean updateAccounts(Collection<Account> accounts) {
...
-          // should be no-op.
-          return new Pair<>(accountId, String.valueOf(container.getId()));
-        } else {
-          logger.error("A container: {} already exists with different attributes", container.getName());
-          throw new IllegalArgumentException("There is a conflicting container in account " + accountName);
+    Set<Container> existingContainers = new HashSet<>();
+    Set<Container> newContainers = new HashSet<>(containers);
+    for (Container existingContainer : account.getAllContainers()) {
+      for (Container newContainer : containers) {
+        // make sure there is no conflict container (conflict means a container with same name but different attributes already exists).
```
lightningrob, line 271: Minor: conflicting container

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -253,56 +253,62 @@ public boolean updateAccounts(Collection<Account> accounts) {
...
-          // If an exactly same container already exists, we directly return its id. Adding same container multiple times
-          // should be no-op.
-          return new Pair<>(accountId, String.valueOf(container.getId()));
-        } else {
-          logger.error("A container: {} already exists with different attributes", container.getName());
-          throw new IllegalArgumentException("There is a conflicting container in account " + accountName);
+    Set<Container> existingContainers = new HashSet<>();
+    Set<Container> newContainers = new HashSet<>(containers);
+    for (Container existingContainer : account.getAllContainers()) {
+      for (Container newContainer : containers) {
```
Arun-LinkedIn, line 270: May be we can make it O(n) by storing existing container names in hash set?

jsjtzyy: Good point, thanks for the suggestion. Will make changes based on hashmap.

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -109,22 +108,22 @@
   public boolean removeAccountUpdateConsumer(Consumer<Collection<Account>> accountUpdateConsumer);
 
   /**
-   * Add a new container to an existing account.
+   * Add a collection of containers to an existing account.
    * @param accountName the name of account in which new container will be added.
-   * @param newContainer the new container to add.
-   * @return a pair of account id and container id strings associated with existing account and new container.
+   * @param containers a collection of containers to add.
+   * @return a collection of containers that are successfully added.
```
lightningrob, line 114: The Helix impl returns all containers (added or already exist).  Javadoc should mention that existing containers are ignored.

lightningrob (approved):
Looks good.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -253,56 +253,62 @@ public boolean updateAccounts(Collection<Account> accounts) {
...
+    // revert changes in original account when there is a failure.
+    Account accountCopy = new AccountBuilder(account).build();
+    accountCopy.updateContainerMap(createdContainers);
+    boolean hasSucceeded = updateAccounts(Collections.singletonList(accountCopy));
     if (!hasSucceeded) {
       throw new IllegalStateException("Account update failed for " + accountName);
     }
-    return new Pair<>(accountId, String.valueOf(nextContainerId));
+    // after metadata store is successfully updated, we safely update original account
+    account.updateContainerMap(createdContainers);
```
Arun-LinkedIn, line 308: Even if we don't update in-memory account here, may be it might eventually get updated later when fetchAndUpdateCache() is called by background updater thread or on ZK notification (similar to how accounts in cache get updated today after updateAccounts() is called from PostAccountsHandler).

Arun-LinkedIn (approved):
Looks good. Approved with couple of comments.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
jsjtzyy: Yes, that's true. However, I think updating the in-mem account immediately doesn't do harm to the correctness. It ensures the in-mem account immediately has latest view (in case ZK notification missed and there is subsequent request against new container immediately after the creation).

lightningrob:
Can you rerun the travis build?  One GetManager test is failing.  I'll merge once it passes.

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -107,6 +107,37 @@
...
   public boolean removeAccountUpdateConsumer(Consumer<Collection<Account>> accountUpdateConsumer);
 
+  /**
+   * Add a collection of containers to an existing account.
+   * @param accountName the name of account in which new container will be added.
+   * @param containers a collection of containers to add.
+   * @return a collection of containers that are successfully added. If original input containers includes existing one,
+   *         the existing containers will be ignored and not present in the result.
+   */
+  default Collection<Container> addContainers(String accountName, Collection<Container> containers) {
```
lightningrob, line 117: I think these methods should throw exceptions (might need to create one).  They can't return false, and having the impl throw IllegalStateException on any failure doesn't feel right.

lightningrob (changes requested):

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
jsjtzyy: Yes, I was trying to use RestException but felt not appropriate. I can introduce new type exception, for example, `AccountServiceException`. What do you think?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -305,7 +307,8 @@ public boolean updateAccounts(Collection<Account> accounts) {
       accountCopy.updateContainerMap(createdContainers);
       boolean hasSucceeded = updateAccounts(Collections.singletonList(accountCopy));
```
lightningrob, line 308: updateAccounts can also throw AccountServiceException (future PR)

lightningrob (approved):

PR #1631, opened 2020-09-25
[DataNodeConfigSource] use dedicated HPS instance

cgtz:
The property store provided by the HelixManager.getPropertyStore()
method does not support caching/subscribed paths. As a solution, we can
move the ownership/instantiation of the PropertyStore into the
PropertyStoreToDataNodeConfigAdapter class.

To keep views of the property store consistent between HelixParticipant
and HelixClusterManager, this commit also adds a shared pool of
DataNodeConfigSources so that the same instance (and ZkClient) can be
used by both classes.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
@@ -13,6 +13,8 @@
  */
 package com.github.ambry.clustermap;
 
+import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.utils.SystemTime;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
@@ -28,7 +30,9 @@ import org.slf4j.LoggerFactory;
  */
 public class HelixFactory {
   private static final Logger LOGGER = LoggerFactory.getLogger(HelixFactory.class);
-  Map<ManagerKey, HelixManager> helixManagers = new ConcurrentHashMap<>();
+  // exposed for use in testing
+  final Map<ManagerKey, HelixManager> helixManagers = new ConcurrentHashMap<>();
+  private final Map<String, DataNodeConfigSource> dataNodeConfigSources = new ConcurrentHashMap<>();
 
   /**
    * Get a reference to a {@link HelixManager}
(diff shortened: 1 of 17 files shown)


jsjtzyy:
@cgtz Could you explain a little more about `The property store provided by the HelixManager.getPropertyStore()
method does not support caching/subscribed paths.` ?  I see `ZkHelixPropertyStore` also inherits from `HelixPropertyStore` indirectly. Just wonder why it doesn't support subscribed paths?

cgtz:
> @cgtz Could you explain a little more about `The property store provided by the HelixManager.getPropertyStore() method does not support caching/subscribed paths.` ? I see `ZkHelixPropertyStore` also inherits from `HelixPropertyStore` indirectly. Just wonder why it doesn't support subscribed paths?

To use the `subscribe(HelixPropertyListener)` method the `ZkHelixPropertyStore` instance must be constructed with either that path or a parent of that path, otherwise it won't receive any notifications. 

The [`HelixManager.getHelixPropertyStore()`](https://github.com/apache/helix/blob/master/helix-core/src/main/java/org/apache/helix/manager/zk/ZKHelixManager.java#L1014) method has a couple of issues related to this:
1. It constructs an `AutoFallbackPropertyStore`. This class overrides the `start()` method (and doesn't call `super.start()`) which leads to an NPE when I call the `subscribe` method because.
2. Even if this NPE was fixed, `AutoFallbackPropertyStore` passes `null` into the cached paths argument of the [`ZkHelixPropertyStore` constructor](https://github.com/apache/helix/blob/master/helix-core/src/main/java/org/apache/helix/store/zk/AutoFallbackPropertyStore.java#L46). This means that any call to `subscribe` would return 0 updates since paths can only be registered with `ZkCallbackCache` during [construction](https://github.com/apache/helix/blob/master/helix-core/src/main/java/org/apache/helix/manager/zk/ZkCacheBaseDataAccessor.java#L777) and not at the moment of calling `subscribe`.

One alternative would be to have used the lower level `subscribeChildChanges` and `subscribeDataChanges`, but the logic around listening would be more complicated IMO.

jsjtzyy:
> 2\. construction

I see. `AutoFallbackPropertyStore` doesn't pass in the cachedPath in ctor and even if `start()` is called, the `eventThread` won't be instantiated and started. Will Helix fix this in open source code?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -69,10 +73,65 @@ public HelixManager getZkHelixManagerAndConnect(String clusterName, String insta
...
+      throw new RuntimeException("Exception while instantiating DataNodeConfigSource", e);
+    }
+  }
+
   /**
    * Hashable key used to cache instances of {@link HelixManager} that match desired parameters.
+   * Exposed for use in testing.
    */
-  private static class ManagerKey {
+  static class ManagerKey {
```
jsjtzyy, line 134: nit: in the `equals` method, suggest comparing `zkAddr` and `instanceName` first.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -28,7 +30,9 @@
  */
 public class HelixFactory {
   private static final Logger LOGGER = LoggerFactory.getLogger(HelixFactory.class);
-  Map<ManagerKey, HelixManager> helixManagers = new ConcurrentHashMap<>();
+  // exposed for use in testing
+  final Map<ManagerKey, HelixManager> helixManagers = new ConcurrentHashMap<>();
```
jsjtzyy, line 34: I see this map is only used by `HelixClusterManagerTest.java`. If so, can we move this map to `MockHelixManagerFactory` in that file?

jsjtzyy (approved):
LGTM with minor comments.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
cgtz: Compared zkAddr and instanceType first since those are most likely to be different in a single process.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
cgtz: It is actually accessed in the getZkHelixManager method, but I was able to make it private and add a method that can be overridden for the test impl, which seems to be cleaner.

jsjtzyy (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
```diff
@@ -75,6 +75,10 @@ public DataNodeConfig get(String instanceName) {
     return instanceConfig != null ? converter.convert(instanceConfig) : null;
   }
 
+  @Override
+  public void close() {
+  }
```
SophieGuo410, line 80: what's this close method for? Does the helixManager needs to be disconnected here?

SophieGuo410 (commented):
Overall LGTM.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -69,10 +73,78 @@ public HelixManager getZkHelixManagerAndConnect(String clusterName, String insta
...
+      if (clusterMapConfig.clusterMapDataNodeConfigSourceType.isPropertyStoreAware()) {
+        propertyStoreSource = new PropertyStoreToDataNodeConfigAdapter(zkAddr, clusterMapConfig);
+      }
+
+      switch (clusterMapConfig.clusterMapDataNodeConfigSourceType) {
+        case INSTANCE_CONFIG:
+          return instanceConfigSource;
+        case PROPERTY_STORE:
+          return propertyStoreSource;
+        case COMPOSITE_INSTANCE_CONFIG_PRIMARY:
```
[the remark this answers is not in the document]
cgtz: Sure, I should publish a short doc on the migration steps actually.

These two both can be used as migration tools so that both property store and instance configs get updated.
The first step would be to switch to `COMPOSITE_INSTANCE_CONFIG_PRIMARY`, which would keep logic in ambry-server/ambry-frontend based on what the instance configs (legacy source) present, but also check for discrepancies and keep the property store up to date on updates to things like sealed partitions.

After we are comfortable that we aren't seeing discrepancies, we can switch to `COMPOSITE_PROPERTY_STORE_PRIMARY`. This will use the new source for application logic/listeners but still keep the instanceconfigs up to date in case we need to do a rollback.

Finally, one can switch to `PROPERTY_STORE` to complete the migration.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/InstanceConfigToDataNodeConfigAdapter.java
cgtz: This close method is just here since I added close() to the interface, but this specific impl shouldn't have to do anything. Unlike PropertyStore...Adapter, this one doesn't own the `HelixManager`'s lifecycle, since it was passed into the constructor. HelixClusterManager will disconnect the `HelixManager`.

SophieGuo410 (approved):

PR #1633, opened 2020-09-28
Fix warmupConnection concurrent issue

zzmao:
responseInfoList need to be access by two threads(netty event loop threadand router thread). A synchronizedList is needed to make it thread safe.
Fix decryptTime metric name.
Add routerPutReqeustRate and routerGetRequestRate.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -1366,6 +1366,7 @@ class PutOperation {
               replicaId.getDataNodeId());
         }
         routerMetrics.getDataNodeBasedMetrics(replicaId.getDataNodeId()).putRequestRate.mark();
+        routerMetrics.routerPutRequestRate.mark();
       }
     }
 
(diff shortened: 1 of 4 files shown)


cgtz (approved):

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1366,6 +1366,7 @@ private void fetchRequests(RequestRegistrationCallback<PutOperation> requestRegi
               replicaId.getDataNodeId());
         }
         routerMetrics.getDataNodeBasedMetrics(replicaId.getDataNodeId()).putRequestRate.mark();
+        routerMetrics.routerPutRequestRate.mark();
```
jsjtzyy, line 1369: We have node-level `putRequestRate` and `getRequestRate`, can we aggregate them without introducing new metrics?

zzmao: Talked with @jsjtzyy  offline. We will add this metric this time because aggregate is a little bit complex.

ankagrawal (approved):
lgtm

PR #1641, opened 2020-10-02
Fix a NPE in BlobStoreStats

justinlin-linkedin:
This should fix NPE in BlobStoreStats
```
2020/10/02 16:41:16.343 ERROR [BlobStoreStats] [ambry-scheduler-1] [ambry-server] [] Unexpected exception while running QueueProcessor in store Partition[7813]
java.lang.NullPointerException: null
```

The data flow that causes this NPE goes like this:
1. For given StoreKey `key`, delete method calls findKey to find the latest IndexValue in the persistent index.
2. At the same time, there is a new IndexValue being added to the persistent index. This new IndexValue has nothing to do with `key`.
3 After after synchronized block, delete method finds the `indexEndOffset` is changed because of the IndexValue from step 2, so it calls findKey again and this time findKey method returns null.
4 Before this PR, delete method would use this null value and set it in`IndexValuesPriorToDelete` list. That's the reason why we have null value int he blobstorestats.
5 After this PR, this will be fixed.

### DIFF
--- a/ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
+++ b/ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
@@ -1044,6 +1044,81 @@ public class BlobStoreTest {
     }
   }
 
+  @Test
+  public void testBlobStoreStats() throws StoreException, InterruptedException, IOException, ExecutionException {
+    store.shutdown();
+    ReplicaId replicaId = getMockReplicaId(tempDirStr);
+    StoreConfig config = new StoreConfig(new VerifiableProperties(properties));
+    MetricRegistry registry = new MetricRegistry();
+    StoreMetrics metrics = new StoreMetrics(registry);
+    MockBlobStoreStats mockBlobStoreStats = new MockBlobStoreStats(time);
+    store = new MockBlobStore(replicaId, config, null, metrics, mockBlobStoreStats);
+    store.start();
+
+    MockId id = put(1, PUT_RECORD_SIZE, Utils.Infinite_Time).get(0);
+    assertNotNull(mockBlobStoreStats.currentValue);
+    assertNull(mockBlobStoreStats.originalPutValue);
+    assertNull(mockBlobStoreStats.previousValue);
+    assertTrue(mockBlobStoreStats.currentValue.isPut());
+
+    updateTtl(id);
+    assertNotNull(mockBlobStoreStats.currentValue);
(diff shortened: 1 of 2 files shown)


jsjtzyy:
@justinlin-linkedin Instead of pasting the error log, could you explain how it happened and what is the root cause? (To make commit message more descriptive.)

justinlin-linkedin:
> @justinlin-linkedin Instead of pasting the error log, could you explain how it happened and what is the root cause? (To make commit message more descriptive.)

Updated.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
```diff
@@ -1044,6 +1044,81 @@ public void concurrentAllTest() throws Exception {
     }
   }
 
+  @Test
+  public void testBlobStoreStats() throws StoreException, InterruptedException, IOException, ExecutionException {
```
jsjtzyy, line 1048: nit: IOException is not thrown in this method.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
```diff
@@ -1044,6 +1044,81 @@ public void concurrentAllTest() throws Exception {
...
+
+      // MockBlobStoreStats should capture delete operation
+      assertNotNull(mockBlobStoreStats.currentValue);
+      assertNotNull(mockBlobStoreStats.originalPutValue);
+      assertNotNull(mockBlobStoreStats.previousValue);
+      assertTrue(mockBlobStoreStats.currentValue.isDelete());
+      assertTrue(mockBlobStoreStats.previousValue.isPut());
+      assertTrue(mockBlobStoreStats.originalPutValue.isPut());
+    } finally {
+      ((MockBlobStore) store).setInDeleteBetweenGetEndOffsetAndFindKey(null);
```
jsjtzyy, line 1117: I think you were trying to reset this in finally block:
```
((MockBlobStore) store).setOperationBeforeSynchronization(null);
```

ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
jsjtzyy, line 1048: The test name is too general, maybe rename it ? (Also add some java doc)

jsjtzyy (approved):
Thanks for fixing this NPE. This PR looks good after minor comments are addressed.

cgtz (approved):
lgtm

PR #1642, opened 2020-10-02
Return copy of Set references in ExchangeMetadataResponse get methods to avoid concurrentModificationException

Arun-LinkedIn:
Due to multiple replica threads referencing same references of ExchangeMetadataResponse Set collection, modification of set by one thread while iterating over it in another thread is resulting in concurrentModificationException. Return copy of set references in get() methods to avoid it.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -1451,7 +1451,7 @@ public class ReplicaThread implements Runnable {
     ExchangeMetadataResponse exchangeMetadataResponse = remoteReplicaInfo.getExchangeMetadataResponse();
     Set<MessageInfo> missingStoreMessages = exchangeMetadataResponse.getMissingStoreMessages();
 
-    if (missingStoreMessages != null && !missingStoreMessages.isEmpty()) {
+    if (!missingStoreMessages.isEmpty()) {
       Set<MessageInfo> missingStoreMessagesFoundInStore = new HashSet<>();
       try {
         // construct map of message info -> converted non-null local key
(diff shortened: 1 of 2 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1629,7 +1630,7 @@ synchronized boolean isEmpty() {
     }
 
     /**
-     *  get the keys corresponding to the missing messages in local store
+     *  Get the keys corresponding to the missing messages in local store
      */
     Set<StoreKey> getMissingStoreKeys() {
       return missingStoreMessages == null ? Collections.emptySet()
```
jsjtzyy, line 1636: I was under the impression the stream way to collect to set is deep copy, so this should be ok (better to confirm it)

jsjtzyy (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1629,7 +1630,7 @@ synchronized boolean isEmpty() {
     }
 
     /**
-     *  get the keys corresponding to the missing messages in local store
+     *  Get the keys corresponding to the missing messages in local store
      */
     Set<StoreKey> getMissingStoreKeys() {
```
cgtz, line 1635: Should this method be synchronized too? It seems like missingStoreMessages can be mutated and using a stream is just like iterating through it and subject to a ConcurrentModificationException.

cgtz (commented):
This mostly LGTM if we are sticking with the copy approach.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
jsjtzyy: + 1, good point.

cgtz (approved):

PR #1644, opened 2020-10-03
Install mysql 8.0 on travis and upgrade to Ubuntu xenial with openjdk11

lightningrob:


.travis.yml
```diff
@@ -13,12 +13,25 @@ cache:
...
+  - mysql --version
+  - wget https://repo.mysql.com//mysql-apt-config_0.8.10-1_all.deb
+  - sudo dpkg -i mysql-apt-config_0.8.10-1_all.deb
+  - sudo apt-get update -q
+  - sudo apt-get install -q -y --allow-unauthenticated -o Dpkg::Options::=--force-confnew mysql-server
+  - sudo systemctl restart mysql
+  - sudo mysql_upgrade
+  - mysql --version
+  - mysql -e 'CREATE DATABASE AccountMetadata;'
+  - mysql -e 'USE AccountMetadata; SOURCE ./ambry-account/src/main/resources/AccountSchema.ddl;'
```
lightningrob, line 30: One optimization I just thought of, now that we are splitting unit and int tests into separate jobs, is that only the int test job needs mysql.  Can this step be done on per-job basis?  (Can be follow up PR.)

ambry-account/src/main/resources/AccountSchema.ddl
```diff
@@ -26,7 +26,7 @@ CREATE TABLE IF NOT EXISTS Accounts
     INDEX lmtIndex (lastModifiedTime),
     INDEX statusIndex (status)
 )
-CHARACTER SET utf8 COLLATE utf8-bin;
```
lightningrob, line 29: @Arun-LinkedIn I ran into collation errors when I ran the script on local Mysql db.  I couldn't find much info on utf8-bin,   Can we go with default collation for the char set?

ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceTest.java
```diff
@@ -192,8 +189,9 @@ public void testBackgroundUpdater() throws InterruptedException, SQLException, I
 
     // force shutdown background updater thread. As the default timeout value is 1 minute, it is possible that thread is
     // present after close() due to actively executing task.
-    ((MySqlAccountService) mySqlAccountService).getScheduler().shutdownNow();
+    mySqlAccountService.getScheduler().shutdownNow();
 
+    Thread.sleep(100);
```
lightningrob, line 194: @Arun-LinkedIn Added the sleep after the assertion failed.  Thread removal may not be immediate.

ambry-router/src/test/java/com/github/ambry/router/CryptoJobHandlerTest.java
```diff
@@ -73,6 +73,11 @@ public void before() {
 
   @After
   public void after() {
+    // Add delay in case buffers are released in background
+    try {
+      Thread.sleep(500);
```
lightningrob, line 78: @justinlin-linkedin The test passes consistently (so far) with this sleep added. This is a temp fix.  A better one is to enhance the afterTest() to check the metrics repeatedly up to a timeout.

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -349,6 +349,14 @@ public boolean addBlobStore(ReplicaId replica) {
     return true;
   }
 
+  /**
+   * Helper utility to add replicaId to partitionToDiskManager.
+   * @param replicaId the {@link ReplicaId} whose partition and disk to use.
+   */
+  protected void updatePartitionToDiskManager(ReplicaId replicaId) {
+    partitionToDiskManager.put(replicaId.getPartitionId(), diskToDiskManager.get(replicaId.getDiskId()));
```
lightningrob, line 357: Note: added because modified line in MockStorageManager hits a compile error in JDK11 trying to access package private class Disk.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreStatsTest.java
```diff
@@ -42,6 +42,7 @@
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.After;
+import org.junit.Ignore;
```
lightningrob, line 45: Oops, will remove.

ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceTest.java
cgtz: For cases where a poll with timeout is needed, we have a test utility that can be used:

```
TestUtils.checkAndSleep(0, () -> numThreadsByThisName(MySqlAccountService.MYSQL_ACCOUNT_UPDATER_PREFIX), 200);
```

.travis.yml
cgtz: I think that may be possible, although I can't find documentation that explicitly says that you can change the services. However, I did find instructions for changing other parameters per each job: https://docs.travis-ci.com/user/build-matrix/#using-different-programming-languages-per-job

So, maybe moving `services` and `before_install` to the int test job will work:
```
jobs:
  include:
    - stage: test
      name: "Unit Tests"
      script: ./travis-unit-test.sh
    - stage: test
      name: "Integration Tests"
      script: ./travis-int-test.sh
      services:
        - mysql
      before_install:
        ...
    - stage: publish
      name: "Publish To Bintray"
      script: ./travis-publish.sh
```

ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
```diff
@@ -0,0 +1,333 @@
...
+import static com.github.ambry.config.MySqlAccountServiceConfig.*;
+import static com.github.ambry.utils.TestUtils.*;
+import static org.junit.Assert.*;
+import static org.mockito.Mockito.*;
+
+
+/**
+ * Integration tests for {@link MySqlAccountService}.
+ */
+public class MySqlAccountServiceIntegrationTest {
```
cgtz, line 48: should this be in the integration-test directory instead of the test directory?

ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceTest.java
lightningrob: Good point.  I'll change that in a follow up PR.

ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
lightningrob: Definitely.  My next PR will add more tests here so I'll also move the class.

ambry-account/src/main/resources/AccountSchema.ddl
Arun-LinkedIn: Seems like default collation for char set uf8 (utf8_general_ci) is case-insensitive and hence complains of unique key violation for case insensitive duplicate account or container names. This is differing from our code logic which seems to do case-sensitive comparison. May be we can update our code logic if there are no existing case insensitive duplicates in PROD? EI has one such case - two accounts with names 'MayFlower' and 'mayflower'.

jsjtzyy: Since EI has such case, we probably need to maintain the case-sensitive comparison.

lightningrob: Running the script on local mysql instance throws the following error:
ERROR 1273 (HY000): Unknown collation: 'utf8'

I'm not sure why it works on EI but we need an option that works consistently.

lightningrob: Resolved by using utf8_bin.

ambry-router/src/test/java/com/github/ambry/router/CryptoJobHandlerTest.java
lightningrob: Added the changes to NettyByteBufLeakHelper.afterTest() in this PR.

ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
```diff
@@ -0,0 +1,333 @@
...
+        new ContainerBuilder(mySqlAccountService.getAccountById((short) 1).getContainerById((short) 3)).setName("c2")
+            .build();
+    accountToUpdate = new AccountBuilder(accountToUpdate).addOrUpdateContainer(containerToUpdate).build();
+    assertFalse("Account update should fail due to name conflict in containers",
+        mySqlAccountService.updateAccounts(Collections.singletonList(accountToUpdate)));
+    assertEquals("UpdateAccountErrorCount in metrics should be 2", 2,
+        accountServiceMetrics.updateAccountErrorCount.getCount());
+  }
+
+  // TODO: add updateContainers tests simiilar to Helix test
```
jsjtzyy, line 325: nit: typo "similar"

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
jsjtzyy: nit: [optional]  probably move this method to somewhere after public methods.

jsjtzyy (approved):
LGTM with two minor comments.

PR #1647, opened 2020-10-06
[Container API] Attach account id to response header for container update

jsjtzyy:
Previously, container update returns a collecion of updated result in the format of JSON.
To deserialize containers from JSON on client side, account id is needed bub not available
in response. This PR attaches account id to response header which will be returned together
with response body (updated containers).


ambry-frontend/src/test/java/com/github/ambry/frontend/FrontendIntegrationTest.java
```diff
@@ -124,30 +124,22 @@
   private static final String HOST_NAME = "localhost";
   private static final String CLUSTER_NAME = "Cluster-name";
   private static boolean enableUndeleteTested = false;
-
```
jsjtzyy, line 127: Most of changes in this file are made by auto-reformat in Intellij, I'll try to revert them.

cgtz (approved):
lgtm

lightningrob (approved):

PR #1653, opened 2020-10-09
[Container API] Support getting single container from given account in ambry-frontend

jsjtzyy:
This PR makes changes in GetAccountsHandler to support getting single container if the
request path matches "/accounts/containers". This can avoid transmitting whole account
info when number of containers in this account is super large. The account id is also attached
to response header to help deserialize container.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/GetAccountsHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/GetAccountsHandler.java
@@ -17,14 +17,17 @@ package com.github.ambry.frontend;
 import com.github.ambry.account.Account;
 import com.github.ambry.account.AccountCollectionSerde;
 import com.github.ambry.account.AccountService;
+import com.github.ambry.account.AccountServiceException;
+import com.github.ambry.account.Container;
+import com.github.ambry.commons.Callback;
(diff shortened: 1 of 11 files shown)


SophieGuo410 (approved):
LGTM!

ambry-frontend/src/main/java/com/github/ambry/frontend/GetAccountsHandler.java
```diff
@@ -152,5 +165,30 @@ private void start() {
       }
       return accounts;
     }
+
+    /**
+     * @return a pair of account and the container to get.
+     * @throws RestServiceException
+     */
+    private Pair<Account, Container> getContainer() throws RestServiceException {
```
lightningrob, line 173: Why does this method return account along with container?  I saw you are setting the account id in the response, but can't you get that from container.getParentAccountId()?

lightningrob (approved):
Approved with one comment.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetAccountsHandler.java
jsjtzyy: oh, that's true. Let me remove the account.

PR #1654, opened 2020-10-09
Small changes for test resilience and exception logging

lightningrob:


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1097,6 +1097,12 @@ private void writeMessagesToLocalStoreAndAdvanceTokens(List<ExchangeMetadataResp
     for (int i = 0; i < exchangeMetadataResponseList.size(); i++) {
       ExchangeMetadataResponse exchangeMetadataResponse = exchangeMetadataResponseList.get(i);
       RemoteReplicaInfo remoteReplicaInfo = replicasToReplicatePerNode.get(i);
+      if (!remoteReplicaInfo.getLocalStore().isStarted()) {
+        logger.debug("Store for {} is stopped, dropping metadata response.",
+            remoteReplicaInfo.getReplicaId().getPartitionId().toPathString());
+        continue;
+      }
```
jsjtzyy, line 1104: There might be an edge where local store was started when exchanging metadata and sending get request but it became unavailable on receiving blobs.  If we look at line 1106 ~ line 1111, we probably get an incorrect `partitionResponseInfoIndex` in subsequent iteration because we skip directly here without updating `partitionResponseInfoIndex`. What do you think?

justinlin-linkedin (approved):
checkAndSleep part looks good to me

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
lightningrob: It's possible.  Is there a better way to avoid the store exception without messing up the index? If you think this is risky I'll revert the change.

jsjtzyy: Throwing a store exception is probably fine for now as it doesn't prevent other store from being written. Could you revert the change and add a "TODO" here that we can check store state before writing blobs into store?  I can make relevant changes in future PR.

lightningrob: Done

jsjtzyy (approved):

PR #1657, opened 2020-10-13
Mysql error handling and retries

lightningrob:
* When container update fails with conflict error, refresh cache and retry operation.
* AccountService.updateAccounts() now throws AccountServiceException.
* Fail creation of MySqlAccountService if bad DB credentials are supplied.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
@@ -252,27 +252,25 @@ public class HelixAccountService extends AbstractAccountService implements Accou
    * </p>
    */
   @Override
-  public boolean updateAccounts(Collection<Account> accounts) throws AccountServiceException {
-    return updateAccountsWithAccountMetadataStore(accounts, accountMetadataStore);
+  public void updateAccounts(Collection<Account> accounts) throws AccountServiceException {
+    updateAccountsWithAccountMetadataStore(accounts, accountMetadataStore);
   }
 
   /**
    * Helper function to update {@link Account} metadata.
    * @param accounts The {@link Account} metadata to update.
    * @param accountMetadataStore The {@link AccountMetadataStore}.
-   * @return True when the update operation succeeds.
+   * @throws AccountServiceException when the update operation fails.
    */
-  boolean updateAccountsWithAccountMetadataStore(Collection<Account> accounts,
+  void updateAccountsWithAccountMetadataStore(Collection<Account> accounts,
       AccountMetadataStore accountMetadataStore) throws AccountServiceException {
     checkOpen();
     Objects.requireNonNull(accounts, "accounts cannot be null");
     if (accounts.isEmpty()) {
-      logger.debug("Empty account collection to update.");
-      return false;
+      throw new IllegalArgumentException("Empty account collection to update.");
     }
     if (config.updateDisabled) {
-      logger.info("Updates has been disabled");
-      return false;
+      throw new AccountServiceException("Updates have been disabled", AccountServiceErrorCode.UpdateDisabled);
     }
     if (hasDuplicateAccountIdOrName(accounts)) {
       accountServiceMetrics.updateAccountErrorCount.inc();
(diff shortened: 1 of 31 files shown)


ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -152,6 +156,8 @@ void fetchAndUpdateCache() {
         } finally {
           infoMapLock.writeLock().unlock();
         }
+        // At this point we can safely say cash is refreshed
```
Arun-LinkedIn, line 159: nit: typo for word cache

Arun-LinkedIn (approved):

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -263,7 +263,7 @@ public boolean updateAccounts(Collection<Account> accounts) {
    * @return True when the update operation succeeds.
    */
   boolean updateAccountsWithAccountMetadataStore(Collection<Account> accounts,
-      AccountMetadataStore accountMetadataStore) {
+      AccountMetadataStore accountMetadataStore) throws AccountServiceException {
```
jsjtzyy, line 266: It seems this method uses both `return false` and AccountServiceException to handle failure cases. Can we throw AccountServiceException everywhere to make it consistent? Also we can consider removing boolean return value, if the method goes through without exception, then it should be a success.

jsjtzyy: Feel free to add new type of error code to `AccountServiceErrorCode`

ambry-test-utils/src/main/java/com/github/ambry/account/InMemAccountService.java
```diff
@@ -167,7 +167,11 @@ public synchronized void clear() {
    */
   public synchronized Account createAndAddRandomAccount() {
     Account account = generateRandomAccount();
-    updateAccounts(Collections.singletonList(account));
+    try {
+      updateAccounts(Collections.singletonList(account));
+    } catch (AccountServiceException ase) {
+      throw new IllegalStateException(ase);
```
jsjtzyy, line 173: Is this because  throwing AccountServiceException will bring lots of changes in too many places?

jsjtzyy (approved):
LGTM (a couple of minor comments)

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
lightningrob: I didn't know what to do about the case where config.updateDisabled = true.  What exception should that be?  Do we ever actually turn off account updates?

ambry-test-utils/src/main/java/com/github/ambry/account/InMemAccountService.java
lightningrob: Pretty much.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
jsjtzyy: Maybe this case justifies `TemporaryDisabled` and from customer point's of view, it's 503 (Unavailable) error. 
This config was introduced to migrate ZK based metadata store to router based metadata store. It gives us an option to temporarily turn off any account update during the migration (to avoid potential race condition). 
We can remove this or if you think it's useful for account service migration, then we can keep it.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -199,20 +199,17 @@ public Account getAccountByName(String accountName) {
   }
 
   @Override
-  public boolean updateAccounts(Collection<Account> accounts) throws AccountServiceException {
+  public void updateAccounts(Collection<Account> accounts) throws AccountServiceException {
     Objects.requireNonNull(accounts, "accounts cannot be null");
     if (accounts.isEmpty()) {
-      logger.debug("Empty account collection to update.");
-      return false;
+      throw new IllegalArgumentException("Empty account collection to update.");
```
jsjtzyy, line 205: nit:  This can be `BadRequest` AccountServiceErrorCode. (We can do this in future PR)

jsjtzyy (approved):

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
lightningrob: ok

PR #1660, opened 2020-10-14
[BlobStoreCompaction] Always enable target index duplicate checking to ignore duplicate PUTs

justinlin-linkedin:
We experienced a compaction error due to duplicate PUTs in log. This is a weird error since there shouldn't be any duplicate PUTs in the log. However, it takes too long to figure out why it happened. Before we fully understand the root cause, this is a bandage to temporarily fix the issue.

We can always check the duplicates in target persistent index while we are check if a IndexValue is a valid one or not. If it's a duplicate of certain IndexValue in target persistent index, then it's not a valid one. This logic would ignore the duplicate PUT and mark it as invalid.

### DIFF
--- a/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
+++ b/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
@@ -39,6 +39,8 @@ import java.util.Map;
 import java.util.NavigableMap;
 import java.util.Properties;
 import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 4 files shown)


ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
```diff
@@ -2393,6 +2445,11 @@ private BlobStoreCompactor getCompactor(Log log, DiskIOScheduler ioScheduler) th
     if (withUndelete) {
       state.properties.put("store.compaction.filter", "IndexSegmentValidEntryWithUndelete");
     }
+    if (alwaysEnableTargetIndexDuplicateChecking) {
+      state.properties.put(StoreConfig.storeAlwaysEnableTargetIndexDuplicateCheckingName, "true");
+    } else {
+      state.properties.put(StoreConfig.storeAlwaysEnableTargetIndexDuplicateCheckingName, "false");
+    }
```
jsjtzyy, line 2452: Can we simplify this if-else as follows?
```
state.properties.put(StoreConfig.storeAlwaysEnableTargetIndexDuplicateCheckingName, Boolean.toString(alwaysEnableTargetIndexDuplicateChecking));
```

jsjtzyy (approved):
Looks good after one minor comment is addressed.

cgtz (approved):

PR #1661, opened 2020-10-15
[HTTP2] use ReferenceCountUtil.safeRelease to avoid setting refCnt to negative number

justinlin-linkedin:
We've seen some refCnt being set to negative number errors. This PR should fix that.


zzmao:
Just two minor questions:
1. Why you think the original code caused negative refCnt?
2. How did you find this safeRelease() ?

zzmao (approved):

cgtz (approved):

PR #1656, opened 2020-10-12
Fix test failures in BlobStoreStatsTest and GetBlobOperationTest

justinlin-linkedin:
This PR should fix test failures  in master branch.
https://github.com/linkedin/ambry/issues/1648
https://github.com/linkedin/ambry/issues/1645

Issue #1636, opened 2020-10-01
Test failure: com.github.ambry.server.RouterServerSSLTest > interleavedOperationsTest[1] FAILED


justinlin-linkedin:
Stacktrace
```
com.github.ambry.server.RouterServerSSLTest > interleavedOperationsTest[1] FAILED
    java.lang.AssertionError: Timeout waiting for operation chain 7 to finish and it is stuck at 9the op:GET_DELETED_SUCCESS the blob id: AAYUAmbdB3sAAQAAAAAAAAAAAawLqSY_RC6XzGJcO2mhcA
        at org.junit.Assert.fail(Assert.java:91)
        at com.github.ambry.server.RouterServerTestFramework.checkOperationChains(RouterServerTestFramework.java:128)
        at com.github.ambry.server.RouterServerSSLTest.interleavedOperationsTest(RouterServerSSLTest.java:184)
```

Gradle scan
https://gradle.com/s/zu7sn4dyu7ips

PR #1662, opened 2020-10-15
Fix CompositeByteBuf memory leak and test failure in RouterServerSSLTest

justinlin-linkedin:
This PR
1. Fix a the compositeByteBuf leak in GetResponse.
2. Increase the timeout to 50 second and try to fix the test failure (at least mitigate) in https://github.com/linkedin/ambry/issues/1636
3. Add code change ignored by https://github.com/linkedin/ambry/pull/1656

### DIFF
--- a/ambry-protocol/src/main/java/com/github/ambry/protocol/CompositeSend.java
+++ b/ambry-protocol/src/main/java/com/github/ambry/protocol/CompositeSend.java
@@ -49,7 +49,6 @@ public class CompositeSend extends AbstractByteBufHolder<CompositeSend> implemen
       if (compositeSendList.size() == 1) {
         compositeSendContent = compositeSendList.get(0).content();
       } else {
-        int numComponents = 0;
         boolean allContentPresent = true;
         List<ByteBuf> byteBufs = new ArrayList<>(compositeSendList.size());
         for (Send send : compositeSendList) {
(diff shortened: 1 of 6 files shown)


ambry-protocol/src/main/java/com/github/ambry/protocol/CompositeSend.java
```diff
@@ -71,8 +71,9 @@ public CompositeSend(List<Send> compositeSendList) {
             if (content instanceof CompositeByteBuf) {
               Iterator<ByteBuf> iterator = ((CompositeByteBuf) content).iterator();
               while (iterator.hasNext()) {
-                ((CompositeByteBuf) compositeSendContent).addComponent(true, iterator.next());
+                ((CompositeByteBuf) compositeSendContent).addComponent(true, iterator.next().retain());
               }
+              content.release();
             } else {
               ((CompositeByteBuf) compositeSendContent).addComponent(true, content);
```
cgtz, line 78: why to CompositeByteBufs have to be treated differently here as opposed to just calling `addComponent(true, content)`?

Is there a reason not to nest CompositeByteBufs?

cgtz (approved):
LGTM, just had a question about why we need to unnest composite buffers

ambry-protocol/src/main/java/com/github/ambry/protocol/CompositeSend.java
justinlin-linkedin: I guess my original ideal is not to nest compositebytebuf, I just don't like nesting things. Anyway, it shouldn't be any difference here. I will nest compositebytebuf.

zzmao (approved):

PR #1666, opened 2020-10-19
[named_blobs] Add NamedBlobDB interface/MySQL impl

cgtz:
Add an interface and a MySQL implementation of NamedBlobDB.
This interface will enable interaction with a mutable mapping layer for
named blobs.

Add a mysql based implementation that supports the put/get/list/delete
APIs. For the first iteration, this does not include retry and failover to
other active databases. This implementation uses Hikari for connection
pooling and built-in mysql connector options for prepared statement
caching.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ before_install:
   - mysql -e 'USE AccountMetadata; SOURCE ./ambry-account/src/main/resources/AccountSchema.ddl;'
   - mysql -e 'CREATE DATABASE ambry_container_storage_stats;'
   - mysql -e 'USE ambry_container_storage_stats; SOURCE ./ambry-server/src/main/resources/AmbryContainerStorageStats.ddl;'
+  - mysql -e 'CREATE DATABASE AmbryNamedBlobs;'
+  - mysql -e 'USE AmbryNamedBlobs; SOURCE ./ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl;'
 branches:
   except: # don't build tags
     - /^v\d/
--- a/ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
+++ b/ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ */
+
+package com.github.ambry.named;
+
+import java.util.concurrent.CompletableFuture;
+
+
+/**
+ * A layer for interacting with a metadata store that holds mappings between blob names and blob IDs.
+ */
+public interface NamedBlobDb {
+
+  /**
+   * Look up a {@link NamedBlobRecord} by name.
+   * @param accountName the name of the account.
+   * @param containerName the name of the container.
+   * @param blobName the name of the blob.
+   * @return a {@link CompletableFuture} that will eventually contain either the {@link NamedBlobRecord} for the named
+   *         blob or an exception if an error occurred.
+   */
+  CompletableFuture<NamedBlobRecord> get(String accountName, String containerName, String blobName);
+
+  /**
+   * List blobs that start with a provided prefix in a container. This returns paginated results. If there are
+   * additional pages to read, {@link Page#getContinuationToken()} will be non null.
+   * @param accountName the name of the account.
+   * @param containerName the name of the container.
+   * @param blobNamePrefix the name prefix to search for.
+   * @param continuationToken if {@code null}, return the first page of {@link NamedBlobRecord}s that start with
+   *                          {@code blobNamePrefix}. If set, use this as a token to resume reading additional pages
+   *                          of records that start with the prefix.
+   * @return a {@link CompletableFuture} that will eventually contain a {@link Page} of {@link NamedBlobRecord}s
+   *         starting with the specified prefix or an exception if an error occurred.
+   */
+  CompletableFuture<Page<NamedBlobRecord>> list(String accountName, String containerName, String blobNamePrefix,
+      String continuationToken);
+
+  /**
+   * Persist a {@link NamedBlobRecord} in the database.
+   * @param record the {@link NamedBlobRecord}
+   * @return a {@link CompletableFuture} that will eventually contain a {@link PutResult} or an exception if an error
+   *         occurred.
+   */
+  CompletableFuture<PutResult> put(NamedBlobRecord record);
+
+  /**
+   * Delete a record for a blob in the database.
+   * @param accountName the name of the account.
+   * @param containerName the name of the container.
+   * @param blobName the name of the blob.
+   * @return a {@link CompletableFuture} that will eventually contain a {@link DeleteResult} or an exception if an error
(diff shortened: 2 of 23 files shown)


.travis.yml
```diff
@@ -28,6 +28,8 @@ before_install:
   - mysql --version
   - mysql -e 'CREATE DATABASE AccountMetadata;'
   - mysql -e 'USE AccountMetadata; SOURCE ./ambry-account/src/main/resources/AccountSchema.ddl;'
+  - mysql -e 'CREATE DATABASE AmbryBlobs;'
```
lightningrob, line 31: I would have preferred AmbryNamedBlobs, but you may have already provisioned it with that name.

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.github.ambry.named;
+
+import com.github.ambry.account.Container;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionStage;
+
+
+public interface NamedBlobDb {
```
lightningrob, line 23: Javadocs

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
```diff
@@ -0,0 +1,31 @@
...
+  CompletableFuture<NamedBlobRecord> get(String accountName, String containerName, String blobName);
```
lightningrob, line 24: Are the methods going to throw only unchecked exceptions on errors?

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
```diff
@@ -0,0 +1,31 @@
...
+
+  CompletableFuture<Page<NamedBlobRecord>> list(String accountName, String containerName, String blobNamePrefix, String continuationToken);
+
+  CompletableFuture<Void> put(NamedBlobRecord record);
```
lightningrob, line 28: Optional: might be useful to return the NamedBlobRecord with blobId filled in.

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
```diff
@@ -0,0 +1,31 @@
...
+
+  CompletableFuture<Void> delete(String accountName, String containerName, String blobName);
```
lightningrob, line 30: Will the caller ever need to know if the deletion happened?  Or will this throw an exception on blob not found?

ambry-api/src/main/java/com/github/ambry/named/NamedBlobRecord.java
```diff
@@ -0,0 +1,73 @@
...
+  public NamedBlobRecord(String accountName, String containerName, String blobName, String blobId,
+      long expirationTimeMs) {
+    this.accountName = accountName;
+    this.containerName = containerName;
+    this.blobName = blobName;
+    this.blobId = blobId;
+    this.expirationTimeMs = expirationTimeMs;
+  }
+
+  public String getAccountName() {
```
lightningrob, line 40: Minor: javadocs

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -0,0 +1,290 @@
...
+/**
+ * An implementation of {@link NamedBlobDb} that uses an active-active MySQL deployment as the source of truth for
+ * named blob mappings.
+ *
+ * It uses the Hikari library for connection pooling, which is a widely used and performant JDBC connection pool
+ * implementation.
+ */
+public class MySqlNamedBlobDb implements NamedBlobDb {
+  // query building blocks
+  private static final String IS_DELETED = "(deleted_ts IS NOT NULL AND deleted_ts <= CURRENT_TIMESTAMP(6))";
```
lightningrob, line 52: Can you use string constants for table and column names?

lightningrob (changes requested):
Looks good.  A few minor comments.

.travis.yml
cgtz: I can change this. This is really just the name for the integration test DB and does need to match the naming patterns in other places.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDbFactory.java
```diff
@@ -0,0 +1,66 @@
...
+import com.zaxxer.hikari.HikariDataSource;
+
+
+public class MySqlNamedBlobDbFactory implements NamedBlobDbFactory {
+  private final MySqlNamedBlobDbConfig config;
+  private final String localDatacenter;
+  private final MetricRegistry metricRegistry;
+  private final AccountService accountService;
+
+  public MySqlNamedBlobDbFactory(VerifiableProperties verifiableProperties, MetricRegistry metricRegistry,
```
SophieGuo410, line 34: May I know which module will call this factory to get the db? Are you planning to call it in IdConverter?

SophieGuo410 (commented):
LGTM! Added a few questions want to get confirmed.

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
cgtz: Checked exceptions will be communicated asynchronously using the `CompletableFuture` to simplify the error handling path. So, the synchronous part of this will only throw unchecked exceptions (on coding bugs), or complete the future with a checked exception (on system/user errors).

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
cgtz: The `NamedBlobRecord` provided already contains the blob ID (since this method is called after Router.putBlob). I do agree that it may be a good practice to have some sort of response object where we can put extra metadata if needed on put/delete requests.

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
cgtz: After, my most recent update, If this returns a success, the deletion will have taken place, or we found that the blob does not exist or was already deleted. I used to have it throw an exception on not found, but i don't think that is really needed since the deletion of the ambry blob would already tell us that. If there is a reason to know that info in the future, I can look at ways of changing this query. i.e. saving info to a variable and then selecting the variable afterwards.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -0,0 +1,290 @@
...
+            if (resultIndex++ == config.listMaxResults) {
+              nextContinuationToken = resultSet.getString(1);
+              break;
+            }
+            String blobName = resultSet.getString(1);
+            String blobId = Base64.encodeBase64URLSafeString(resultSet.getBytes(2));
+            Timestamp expirationTime = resultSet.getTimestamp(3);
+            Timestamp deletionTime = resultSet.getTimestamp(4);
+            long currentTime = System.currentTimeMillis();
+            if (compareTimestamp(expirationTime, currentTime) > 0 && compareTimestamp(deletionTime, currentTime) > 0) {
```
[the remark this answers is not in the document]
cgtz: I think for now, it is good to exclude them (although I added some trace logging for this case). If there is a use case for knowing which records are expired but still have some elements stored (i.e. include_all flag), we can add support for that in the future.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDbFactory.java
cgtz: Yep, let me add the code to add this into IdConverterFactory.

SophieGuo410 (approved):

lightningrob (approved):

PR #1668, opened 2020-10-22
Upgrade metrics to 4.1.2, allow custom configuration for JmxReporter

cgtz:
- Upgrade dropwizard metrics to 4.1.2
- Upgrading to 4.1.2 brings some changes in MBean naming in JmxReporter,
  which can effect users that query these MBeans based on naming patterns.
- To address this, add an optional argument to each "server" class to
  allow users to pass in a factory that produces a JmxReporter with
  non-default configs.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
@@ -50,6 +50,7 @@ import java.util.ArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -74,6 +75,7 @@ public class VcrServer {
   private JmxReporter reporter = null;
   private ConnectionPool connectionPool = null;
   private final NotificationSystem notificationSystem;
+  private final Function<MetricRegistry, JmxReporter> reporterFactory;
   private CloudDestinationFactory cloudDestinationFactory;
   private VcrRequests requests;
   private RequestHandlerPool requestHandlerPool;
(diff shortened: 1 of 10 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
```diff
@@ -98,10 +103,12 @@ public VcrServer(VerifiableProperties properties, ClusterAgentsFactory clusterAg
    * @param clusterAgentsFactory the {@link ClusterAgentsFactory} to use.
    * @param notificationSystem the {@link NotificationSystem} to use.
    * @param cloudDestinationFactory the {@link CloudDestinationFactory} to use.
+   * @param reporterFactory
```
jsjtzyy, line 106: nit, complete the comment

jsjtzyy (approved):
LGTM

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
```diff
@@ -74,6 +75,7 @@
   private JmxReporter reporter = null;
   private ConnectionPool connectionPool = null;
   private final NotificationSystem notificationSystem;
+  private final Function<MetricRegistry, JmxReporter> reporterFactory;
```
lightningrob, line 78: I like this Function usage.  It's inconsistent with the other factory interfaces we use, but it seems like a sensible shortcut.

ambry-rest/src/main/java/com/github/ambry/rest/RestServer.java
```diff
@@ -171,17 +172,20 @@ public RestServer(VerifiableProperties verifiableProperties, ClusterMap clusterM
...
-      throws Exception {
+      NotificationSystem notificationSystem, SSLFactory sslFactory, List<ChannelHandler> addedChannelHandlers,
+      Function<MetricRegistry, JmxReporter> reporterFactory) throws Exception {
     if (verifiableProperties == null || clusterMap == null || notificationSystem == null) {
       throw new IllegalArgumentException("Null arg(s) received during instantiation of RestServer");
     }
     MetricRegistry metricRegistry = clusterMap.getMetricRegistry();
     RestServerConfig restServerConfig = new RestServerConfig(verifiableProperties);
-    reporter = JmxReporter.forRegistry(metricRegistry).build();
+    reporter = reporterFactory != null ? reporterFactory.apply(metricRegistry)
```
lightningrob, line 187: Ideally we should have a test case that passes a non-null reporterFactory.

lightningrob (approved):
Code looks good.  Test case recommended.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
cgtz: I was going to add a Factory instance but putting it in a place where all 3 entry points (server, frontend, vcr) could share it without adding a dependency on metrics-jmx to a package became a problem (metrics-jmx is where the JmxReporter class is defined)

ambry-rest/src/main/java/com/github/ambry/rest/RestServer.java
cgtz: added cases for each server class

jsjtzyy (approved):

lightningrob (approved):

PR #1671, opened 2020-10-24
Encode user metadata values in response header if needed.

lightningrob:


cgtz (approved):
LGTM, thanks for making this change!

PR #1680, opened 2020-11-01
Check data integrity before loading sealed index segment to memory

jsjtzyy:
Previously, sealed index segment is mapped to memory map/buffer without a full read. This may ignore potential corrupted data. Recently, we saw corrupted index file being loaded to memory on production host and kept throwing exception on invalid data. This PR adds an explicit full read ahead of mapping process to ensure each index entry and CRC value are correct.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
@@ -15,6 +15,7 @@ package com.github.ambry.store;
 
 import com.github.ambry.config.StoreConfig;
 import com.github.ambry.utils.ByteBufferInputStream;
+import com.github.ambry.utils.Crc32;
 import com.github.ambry.utils.CrcInputStream;
 import com.github.ambry.utils.CrcOutputStream;
 import com.github.ambry.utils.FilterFactory;
@@ -770,14 +771,10 @@ class IndexSegment implements Iterable<IndexEntry> {
    * @throws StoreException if there are problems with the index
    */
   void seal() throws StoreException {
-    try {
-      sealed.set(true);
-      map();
-      // we should be fine reading bloom filter here without synchronization as the index is read only
-      persistBloomFilter();
-    } catch (FileNotFoundException e) {
-      throw new StoreException("File not found while sealing index segment.", StoreErrorCodes.File_Not_Found);
-    }
+    sealed.set(true);
+    map();
+    // we should be fine reading bloom filter here without synchronization as the index is read only
+    persistBloomFilter();
   }
 
   /**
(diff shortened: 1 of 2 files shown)


justinlin-linkedin (changes requested):
We can leverage the crc at the end of each index segment file to check data integrity, instead of iterating through IndexValues one by one.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -823,8 +824,9 @@ IndexValue getIndexValueOfLastPut() throws StoreException {
    * Maps the segment of index either as a memory map or a in memory buffer depending on config.
    * @throws StoreException if there are problems with the index
    */
-  private void map() throws StoreException {
+  private void map() throws StoreException, FileNotFoundException {
     rwLock.writeLock().lock();
+    checkDataIntegrity(indexFile);
```
justinlin-linkedin, line 829: I would recommend to directly use `serEntries` initialized below to create a CrcInputStream for data integrity validation. This is good for two reasons: 1. We don't have to read the file twice(although the second read would be just a memory copy) 2. We can deal with FileNotFoundException in the same manner.

jsjtzyy: Sure,  I use `serEntries` to check data integrity in latest commit. For non-heap buffer, I don't find a good way to avoid bytes transfer (as you can see I still transfer bytes to a heap bytes array for crc calculation)

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -1010,11 +1009,37 @@ private void readFromFile(File fileToRead, Journal journal) throws StoreExceptio
...
+      // this is for heap buffer only
+      byteArray = serEntries.array();
+    } else {
+      // this is for non-heap (direct) buffer
+      serEntries.position(0);
+      byteArray = new byte[serEntries.capacity()];
+      serEntries.get(byteArray, 0, serEntries.capacity());
+    }
+    Crc32 crc = new Crc32();
+    crc.update(byteArray, 0, byteArray.length - CRC_FIELD_LENGTH);
```
justinlin-linkedin, line 1029: actually you don't have to copy bytebuffer to an byte array, Crc32 has a method to update CRC with ByteBuffer directly and it doesn't matter if it's heap bytebuffer or direct bytebuffer.

jsjtzyy: True, there is a method to update CRC with ByteBuffer. However, we should exclude the last 8 bytes in buffer to calculate CRC, right?  How can we do that with that method?

cgtz: You can do this by setting the limit on the ByteBuffer before calling the CRC method.

jsjtzyy: Thank you @justinlin-linkedin and @cgtz , the code should be more concise now.

justinlin-linkedin (approved):
LGTM

cgtz (approved):

PR #1689, opened 2020-11-10
Support UTF-8 char set in buildUserMetadata method

jsjtzyy:
Replace ASCII char set with UTF-8 to support more characters (i.e Chinese characters) in the user metadata.


ambry-router/src/main/java/com/github/ambry/router/GCMCryptoService.java
```diff
@@ -366,18 +366,17 @@ private static void serializeKeyRecord(ByteBuffer outputBuffer, byte[] key, Stri
       outputBuffer.putShort(KEY_RECORD_VERSION_V_1);
       outputBuffer.putInt(key.length);
       outputBuffer.put(key);
-      Utils.serializeString(outputBuffer, keyGenAlgo, StandardCharsets.US_ASCII);
+      Utils.serializeString(outputBuffer, keyGenAlgo, StandardCharsets.UTF_8);
```
jsjtzyy, line 369: I changed this because ` String keyAlgo = Utils.readIntString(dataStream)` in `deserializeKeyRecord()` is actually using UTF-8. (line 383)

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -443,12 +443,12 @@ public static BlobProperties buildBlobProperties(Map<String, Object> args) throw
           // key size
           sizeToAllocate += 4;
           String keyToStore = key.substring(Headers.USER_META_DATA_HEADER_PREFIX.length());
-          sizeToAllocate += keyToStore.getBytes(StandardCharsets.US_ASCII).length;
+          sizeToAllocate += keyToStore.getBytes(StandardCharsets.UTF_8).length;
```
cgtz, line 446: I thought of one slight deployment issue with changing the charset without making a new user metadata version.

If you deploy this change to one frontend, a user could write a non-ascii value into user metadata, and the binary format will now store its real utf-8 value instead of a `?` character.
If this blob's user metadata is fetched on a frontend without this change, that frontend will attempt to decode as ascii and a deserialization error will occur.

However, this may be a negligible number of users affected and any problems will be resolved by completing the deployment, so maybe continuing with this approach and opting for a fast deployment is the easiest solution.

jsjtzyy: Yeah, this is indeed a deployment issue as frontends can't be deployed simultaneously. The edge case only affects China team for now. I guess the impact should be very minimum.  (I agree introducing metadata v2 is theoretically the safest way to migrate from ascii to utf-8 smoothly, however we can tolerate the transient error as only one team is really building metadata with non-ascii character, we can coordinate with them when rolling out this change).

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -443,12 +443,12 @@ public static BlobProperties buildBlobProperties(Map<String, Object> args) throw
...
           String value = getHeader(args, key, true);
           userMetadataMap.put(keyToStore, value);
           // value size
           sizeToAllocate += 4;
-          sizeToAllocate += value.getBytes(StandardCharsets.US_ASCII).length;
+          sizeToAllocate += value.getBytes(StandardCharsets.UTF_8).length;
```
lightningrob, line 451: Can you make the Charset a class level constant?

lightningrob (changes requested):

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
jsjtzyy: Sure, will do.

lightningrob (approved):

cgtz (approved):

PR #1694, opened 2020-11-12
[MySQL] Moving common mysql utility to ambry-commons module

justinlin-linkedin:
Moving several MySql classes from accounts to commons module.

### DIFF
--- a/ambry-commons/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
+++ b/ambry-commons/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
@@ -11,10 +11,8 @@
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  */
-package com.github.ambry.account.mysql;
+package com.github.ambry.mysql;
 
-import com.github.ambry.account.AccountServiceErrorCode;
-import com.github.ambry.account.AccountServiceException;
 import com.github.ambry.utils.Pair;
 import com.mysql.cj.exceptions.MysqlErrorNumbers;
 import java.sql.Connection;
@@ -22,7 +20,6 @@ import java.sql.Driver;
 import java.sql.DriverManager;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
-import java.sql.SQLIntegrityConstraintViolationException;
 import java.sql.SQLTransientConnectionException;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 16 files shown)


cgtz (changes requested):
Changes look good but there seems to be a compilation error on travis

I was initially wondering if it would be good to create a separate ambry-mysql subproject to separate out the mysql dependencies, but the mysql connector is not a huge dependency and is probably needed by most of the subprojects anyways, so commons should be fine.

justinlin-linkedin:
> Changes look good but there seems to be a compilation error on travis
> 
> I was initially wondering if it would be good to create a separate ambry-mysql subproject to separate out the mysql dependencies, but the mysql connector is not a huge dependency and is probably needed by most of the subprojects anyways, so commons should be fine.

The compilation error is already fixed.

cgtz (approved):
lgtm

build.gradle
```diff
@@ -230,6 +230,7 @@ project(':ambry-commons') {
         compile ("org.apache.helix:helix-core:$helixVersion") {
             exclude group: "log4j", module: "log4j"
         }
+        compile "mysql:mysql-connector-java:8.0.21"
```
lightningrob, line 233: Could you add $mysqlVersion and use it here.  Also, can we remove it from account project?

ambry-commons/src/main/java/com/github/ambry/mysql/MySqlMetrics.java
```diff
@@ -11,17 +11,17 @@
...
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 
 
 /**
- * Expose metrics related to writes and reads to/from MySql DB used in {@link com.github.ambry.account.MySqlAccountService}
+ * Expose metrics related to writes and reads to/from MySql DB.
  */
-public class MySqlAccountStoreMetrics {
+public class MySqlMetrics {
```
lightningrob, line 24: Good idea, this is independent of accounts.

ambry-commons/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
```diff
@@ -317,7 +286,7 @@ private boolean isBetterEndpoint(DbEndpoint first, DbEndpoint second) {
    * Close a resource without throwing exception.
    * @param resource the resource to close.
    */
-  private static void closeQuietly(AutoCloseable resource) {
+  private void closeQuietly(AutoCloseable resource) {
```
lightningrob, line 289: Why not static?  (This can probably move to MySqlUtils anyway.)

lightningrob (approved):
LGTM.  A few minor comments.

Arun-LinkedIn (approved):
Looks good to me.

build.gradle
justinlin-linkedin: updated

ambry-commons/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
justinlin-linkedin: I can revert the change.

PR #1697, opened 2020-11-18
Ignore version mismatch temporarily in MySqlAccountService while updating Accounts and Containers

Arun-LinkedIn:
Since HelixAccountService and MySqlAccountService handles versioning of Accounts and Containers differently, this causes mismatch between Accounts when compared in CompositeAccountService. This PR temporarily ignores version mismatch during Account/Container updates in MySqlAccountService.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
@@ -131,6 +131,7 @@ abstract class AbstractAccountService implements AccountService {
         }
       } else {
         // existing container
+        //TODO: Include condition to check version as well after moving to MySql only.
         Container existingContainer = existingContainersInAccount.get(container.getName());
         if (existingContainer == null) {
           throw new AccountServiceException(
@@ -140,11 +141,6 @@ abstract class AbstractAccountService implements AccountService {
           throw new AccountServiceException(
               "In account " + accountName + ", container " + container.getName() + " has containerId "
                   + existingContainer.getId() + " (" + container.getId() + " was supplied)",
-              AccountServiceErrorCode.ResourceConflict);
-        } else if (existingContainer.getSnapshotVersion() != container.getSnapshotVersion()) {
-          throw new AccountServiceException(
-              "In account " + accountName + ", container " + container.getName() + " has version "
-                  + existingContainer.getSnapshotVersion() + " (" + container.getSnapshotVersion() + " was supplied)",
               AccountServiceErrorCode.ResourceConflict);
         } else {
           resolvedContainers.add(container);
--- a/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
@@ -131,7 +131,6 @@ abstract class AbstractAccountService implements AccountService {
         }
       } else {
         // existing container
-        //TODO: Include condition to check version as well after moving to MySql only.
         Container existingContainer = existingContainersInAccount.get(container.getName());
         if (existingContainer == null) {
           throw new AccountServiceException(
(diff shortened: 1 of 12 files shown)


ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
```diff
@@ -141,11 +142,6 @@ public Account getAccountById(short id) {
               "In account " + accountName + ", container " + container.getName() + " has containerId "
                   + existingContainer.getId() + " (" + container.getId() + " was supplied)",
               AccountServiceErrorCode.ResourceConflict);
-        } else if (existingContainer.getSnapshotVersion() != container.getSnapshotVersion()) {
```
lightningrob, line 144: Does this need to be changed?  I thought only account version was out of sync.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -249,14 +249,16 @@ public void updateAccounts(Collection<Account> accounts) throws AccountServiceEx
     infoMapLock.readLock().lock();
     try {
       AccountInfoMap accountInfoMap = accountInfoMapRef.get();
-      if (accountInfoMap.hasConflictingAccount(accounts)) {
+      //TODO: Temporarily disabling version check with CompositeAccountService enabled. Remove it after moving to MySql only.
+      if (accountInfoMap.hasConflictingAccount(accounts, true)) {
```
lightningrob, line 253: Can we make ignoreAccountVersion a mysql config property?  That way we won't need to change the code later.  Default can be false and override in config when composite AS is used.

lightningrob (changes requested):
Overall good, couple of comments.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
jsjtzyy: Right, config seems to be better. This is one-off migration and we can keep the default value false after it's complete.

jsjtzyy (approved):

ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
lightningrob: Discussed offline.  This does need to be adjusted.  Please use the mentioned config property to toggle the behavior.

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
```diff
@@ -97,6 +98,13 @@
   @Default("false")
   public final boolean updateDisabled;
 
+  /**
+   * If true, MySqlAccountService would ignore version mismatch while updating Accounts and Containers.
+   */
+  @Config(IGNORE_VERSION_MISMATCH)
+  @Default("false")
+  public final boolean ignoreVersionMismatch;
```
lightningrob, line 106: I would move this to base AccountServiceConfig class so AbstractAccountService can also use it.

lightningrob (changes requested):

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -340,6 +341,23 @@ ExecutorService getScheduler() {
...
       throws AccountServiceException {
+
+    // Check for name/id/version conflicts for containers being updated with those in local cache.
+    infoMapLock.readLock().lock();
+    try {
+      if (accountInfoMapRef.get()
+          .hasConflictingContainer(resolvedContainers, account.getId(), config.ignoreVersionMismatch)) {
+        logger.error(
+            "Containers={} under Account={} conflict with the containers in local cache. Cancel the update operation.",
+            account.getAllContainers(), account.getId());
```
jsjtzyy, line 352: minor: can you print out  account name instead of account id?  
Also `account.getAllContainers()` can be replaced with `resolvedContainers`

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -289,7 +289,7 @@ void updateAccountsWithAccountMetadataStore(Collection<Account> accounts, Accoun
     // update operation for all the accounts if any conflict exists. There is a slight chance that the account to update
     // conflicts with the accounts in the local cache, but does not conflict with those in the helixPropertyStore. This
     // will happen if some accounts are updated but the local cache is not refreshed.
-    if (accountInfoMapRef.get().hasConflictingAccount(accounts)) {
+    if (accountInfoMapRef.get().hasConflictingAccount(accounts, false)) {
```
lightningrob, line 292: I'm wondering if this will be a problem when we switch to mysql primary and Helix secondary.

lightningrob (approved):
LGTM

PR #1698, opened 2020-11-18
Refactor RestUtils.setUserMetadataHeaders to take byte array or map as input

lightningrob:


ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -860,38 +860,50 @@ public static void accountAndContainerNamePreconditionCheck(RestRequest restRequ
...
+    }
+  }
+
+  /**
+   * Sets the user metadata in the headers of the response.
+   * @param userMetadataMap map of user metadata that needs to be sent.
+   * @param restResponseChannel the {@link RestResponseChannel} that is used for sending the response.
+   * @throws RestServiceException if there are any problems setting the header.
+   */
+  public static void setUserMetadataHeaders(Map<String, String> userMetadataMap,
```
cgtz, line 884: There seems to be a difference between the map provided in open source and the closed source place where this is intended to be used.

`RestUtils.buildUserMetadata` returns a map with keys that already have the "x-ambry-um-" prefix.
The decoded map in closed source does not (without writing some code that adds them).

Would it be possible to make `buildUserMetadata` not prepend the prefix and have `setUserMetadataHeaders` do it instead?

jsjtzyy: Agree. Also I am trying to understand the motivation to have this new method. Are you planning to replace the piece of similar logic after `codec.decode` in closed source to avoid duplicate code?

jsjtzyy (approved):
Looks good after Casey's comment is addressed.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
jsjtzyy: Nvm, you have explained offline.

lightningrob: I updated the new method to add the prefix if not already present.

cgtz (approved):

PR #1700, opened 2020-11-20
Fix illegal character in Azure container name when using container layout strategy

lightningrob:
Note: I have tested AzureIntegrationTest locally with container layout to verify the fix, and will make a separate PR that runs a few test cases with both layout options.


cgtz (approved):

ankagrawal (commented):
lgtm!

PR #1702, opened 2020-11-23
[StorageQuota] Add MySql aggregation task

justinlin-linkedin:
Add mysql implementation of aggregation task and update AccountStatsStore to interface with mysql.


ambry-api/src/main/java/com/github/ambry/clustermap/ClusterParticipant.java
```diff
@@ -32,10 +33,12 @@
   /**
    * Initiate the participation of cluster participant.
    * @param ambryHealthReports {@link List} of {@link AmbryHealthReport} to be registered to the participant.
+   * @param accountStatsStore
```
cgtz, line 36: javadoc

ambry-api/src/main/java/com/github/ambry/server/AccountStatsStore.java
```diff
@@ -0,0 +1,97 @@
...
+   * @return {@link StatsWrapper} of given {@code hostname} in the {@code clusterName}.
+   * @throws Exception
+   */
+  StatsWrapper queryStatsOf(String clusterName, String hostname) throws Exception;
+
+  /**
+   * Return the aggregated stats for the given {@code clusterName}. This is the stats stored by method {@link #storeAggregatedStats}.
+   * Since the {@link StatsSnapshot} passed to {@link #storeAggregatedStats} only have account and container stats, this
+   * method returns a map back to the caller. The key of the outer map is the account id in string format, and the key of
+   * the inner map is the container id in string format, the value of the inner map is the storage usage for this container.
```
cgtz, line 52: why are the account/container IDs in string format instead of Integer? Also, I remember their being another format where the stats report is grouped by partition class. Is that handled somewhere else?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterAggregator.java
```diff
@@ -53,33 +54,59 @@
...
+  Pair<StatsSnapshot, StatsSnapshot> doWorkOnStatsWrapperMap(Map<String, StatsWrapper> statsWrappers,
+      StatsReportType type, boolean removeExceptionOnType) throws IOException {
     StatsSnapshot partitionSnapshot = new StatsSnapshot(0L, new HashMap<>());
     Map<String, Long> partitionTimestampMap = new HashMap<>();
     StatsSnapshot rawPartitionSnapshot = new StatsSnapshot(0L, new HashMap<>());
-    exceptionOccurredInstances.remove(type);
-    for (Map.Entry<String, String> statsWrapperJSON : statsWrappersJSON.entrySet()) {
-      if (statsWrapperJSON != null && statsWrapperJSON.getValue() != null) {
+    if (removeExceptionOnType) {
+      exceptionOccurredInstances.remove(type);
```
cgtz, line 85: What was the purpose of this piece of code resetting exceptionOccuredInstances for a given type again? Sorry, I'm not up to date on this.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -0,0 +1,134 @@
...
+    } finally {
+      if (clusterMapConfig.clustermapEnableContainerDeletionAggregation && callback != null && results != null
+          && statsReportType.equals(StatsReportType.ACCOUNT_REPORT)) {
+        callback.onCompletion(results.getFirst(), exception);
+      }
+    }
+  }
+
+  private String stripPortNumber(String instanceName) {
+    return instanceName.replace("_" + clusterMapConfig.clusterMapPort, "");
```
cgtz, line 127: this may not be 100% accurate for a local deployment where multiple processes share the same host name but different ports.

ambry-api/src/main/java/com/github/ambry/server/AccountStatsStore.java
jsjtzyy: I think it's because  the `subMap` in `StatsSnapshot` uses String type key (account/container and partition Id are converted to String when doing aggregation.)

Talked with Justin offline, in the interest of time, partition class stats report won't be migrated this quarter. We probably will take this up next year.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -123,14 +125,14 @@ public void setInitialLocalPartitions(Collection<String> localPartitions) {
    * @throws IOException if there is an error connecting to the Helix cluster.
    */
   @Override
-  public void participate(List<AmbryHealthReport> ambryHealthReports, Callback<StatsSnapshot> callback)
-      throws IOException {
+  public void participate(List<AmbryHealthReport> ambryHealthReports, AccountStatsStore accountStatsStore,
```
jsjtzyy, line 128: minor: add java doc for `accountStatsStore`

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -439,7 +441,7 @@ private void awaitDisablingPartition() throws InterruptedException {
    * @param callback a callback which will be invoked when the aggregation report has been generated successfully.
    */
   private void registerHealthReportTasks(StateMachineEngine engine, List<AmbryHealthReport> healthReports,
-      Callback<StatsSnapshot> callback) {
+      AccountStatsStore accountStatsStore, Callback<StatsSnapshot> callback) {
```
jsjtzyy, line 444: same here. (Also, we should consider removing the callback, this can be done in future PR)

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param callback a callback which will be invoked when the aggregation report has been generated successfully.
+   * @param clusterMapConfig the {@link ClusterMapConfig} associated with helix participant.
+   */
+  MySqlReportAggregatorTask(HelixManager manager, long relevantTimePeriodInMs, StatsReportType statsReportType,
+      AccountStatsStore accountStatsStore, Callback<StatsSnapshot> callback, ClusterMapConfig clusterMapConfig) {
+    this(manager, relevantTimePeriodInMs, statsReportType, accountStatsStore, callback, clusterMapConfig,
+        SystemTime.getInstance());
+  }
+
+  MySqlReportAggregatorTask(HelixManager manager, long relevantTimePeriodInMs, StatsReportType statsReportType,
```
jsjtzyy, line 72: Why we need to constructors here?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -0,0 +1,134 @@
...
+        String recordedMonthValue = accountStatsStore.queryRecordedMonth(clusterMapConfig.clusterMapClusterName);
+        if (recordedMonthValue == null || recordedMonthValue.isEmpty() || !currentMonthValue.equals(
+            recordedMonthValue)) {
+          accountStatsStore.takeSnapshotOfAggregatedStatsSetMonth(clusterMapConfig.clusterMapClusterName,
+              currentMonthValue);
+        }
+      }
+      return new TaskResult(TaskResult.Status.COMPLETED, "Aggregation success");
+    } catch (Exception e) {
+      logger.error("Exception thrown while aggregating stats from health reports across all nodes ", e);
```
jsjtzyy, line 115: let's update this error message to reflect aggregation on mysql based stats report.

ambry-server/src/main/java/com/github/ambry/server/mysql/AccountStatsMySqlStore.java
```diff
@@ -125,21 +137,98 @@ public void publish(StatsWrapper statsWrapper) {
...
+   * value of the inner map is the storage usage of the container.</p>
+   * @param clusterName the clusterName.
+   * @return A map that represents container storage usage.
+   * @throws Exception
+   */
+  @Override
+  public Map<String, Map<String, Long>> queryAggregatedStats(String clusterName) throws Exception {
+    Map<String, Map<String, Long>> result = new HashMap<>();
+    aggregatedaccountReportsDao.queryContainerUsageForCluster(clusterName, (accountId, containerId, storageUsage) -> {
+      result.computeIfAbsent(String.valueOf(accountId), k -> new HashMap<>())
```
cgtz, line 199: Do we want to use the latest updated storageUsage value or just the last one in order?

ambry-api/src/main/java/com/github/ambry/server/AccountStatsStore.java
justinlin-linkedin: Second Yingyi's comments. The partition class stats will be handled next quarter.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterAggregator.java
justinlin-linkedin: Trying to not change anywhere unnecessary to this PR.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
justinlin-linkedin: This method is just trying to get the hostname. I will append the port number later in the database method. So in helix the  instance name looks something like this `ltx1-app12345.prod.linkedin.com_15999`, but in the database, the hostname looks like `ltx1-app12345.prod_15999`. So I have to get the hostname first, remove the domain name and then append the port number. Even if we have several processes running at the same host, as long as their port numbers are different, it will be fine for database.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterAggregator.java
```diff
@@ -53,33 +54,59 @@
...
+  Pair<StatsSnapshot, StatsSnapshot> doWorkOnStatsWrapperMap(Map<String, StatsWrapper> statsWrappers,
+      StatsReportType type) throws IOException {
+    return doWorkOnStatsWrapperMap(statsWrappers, type, true);
+  }
+
+  Pair<StatsSnapshot, StatsSnapshot> doWorkOnStatsWrapperMap(Map<String, StatsWrapper> statsWrappers,
+      StatsReportType type, boolean removeExceptionOnType) throws IOException {
     StatsSnapshot partitionSnapshot = new StatsSnapshot(0L, new HashMap<>());
     Map<String, Long> partitionTimestampMap = new HashMap<>();
     StatsSnapshot rawPartitionSnapshot = new StatsSnapshot(0L, new HashMap<>());
```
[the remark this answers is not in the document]
justinlin-linkedin: Trying to avoid unnecessary change as much as possible. We can do this later.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
justinlin-linkedin: This constructor is for testing, but i realize the tests are redundant, removed this constructor.

ambry-server/src/main/java/com/github/ambry/server/mysql/AccountStatsMySqlStore.java
justinlin-linkedin: For aggregated stats, there is only one storage usage. When there is a new value, it will override the old ones in database.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
cgtz: This seems like this would have problems removing the port number if different hosts in the cluster have different ports. For example, let's say the cluster has `host1_1234` and `host2_2345` and `clusterMapConfig.clusterMapPort` is set to 1234. Then, this method would work correctly for host1, but return `host2_2345` for host2.
Maybe it is better to split by underscore and take the first part of it.

justinlin-linkedin: But in different ambry server process, the port number in clustermap should be the port number that this process registered with helix cluster. If the cluster has `host1_1234` and `host2_2345`, then the clustermap port number in `host1` would be `1234` and the clustermap port number in `host2` would be `2345`.

cgtz: Maybe I am misreading this, but it seems like `stripPortNumber` is called for all hosts in a cluster (line 86), but `clusterMapConfig.clusterMapPort` is static and only represents this host's port number.

ambry-server/src/main/java/com/github/ambry/server/mysql/AggregatedAccountReportsDao.java
```diff
@@ -0,0 +1,237 @@
...
+  public static final String MONTH_COLUMN = "month";
+
+  private static final Logger logger = LoggerFactory.getLogger(AccountReportsDao.class);
+  private static final String insertSql =
+      String.format("INSERT INTO %s (%s, %s, %s, %s, %s) VALUES (?, ?, ?, ?, NOW())", AGGREGATED_ACCOUNT_REPORTS_TABLE,
+          CLUSTER_NAME_COLUMN, ACCOUNT_ID_COLUMN, CONTAINER_ID_COLUMN, STORAGE_USAGE_COLUMN, UPDATED_AT_COLUMN);
+  private static final String queryUsageSqlForCluster =
+      String.format("SELECT %s, %s, %s FROM %s WHERE %s = ?", ACCOUNT_ID_COLUMN, CONTAINER_ID_COLUMN,
+          STORAGE_USAGE_COLUMN, AGGREGATED_ACCOUNT_REPORTS_TABLE, CLUSTER_NAME_COLUMN);
+  private static final String queryMontlyUsageSqlForCluster =
```
cgtz, line 60: nit: montly -> monthly

cgtz (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
justinlin-linkedin: oh, you are absolutely right, I was just being stupid, will update.

jsjtzyy (approved):

PR #1707, opened 2020-11-30
[named_blobs] Integrate NamedBlobDb and support deletes

cgtz:
- Use NamedBlobDb in AmbryIdConverter
- Add support for getting the blob ID deleted after a delete call
- Call delete from inside of AmbryIdConverter.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AccountAndContainerInjector.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AccountAndContainerInjector.java
@@ -42,7 +42,6 @@ public class AccountAndContainerInjector {
   private static final Set<String> requiredAmbryHeadersForPutWithAccountAndContainerName = Collections.unmodifiableSet(
       new HashSet<>(Arrays.asList(Headers.TARGET_ACCOUNT_NAME, Headers.TARGET_CONTAINER_NAME)));
   private static final Logger logger = LoggerFactory.getLogger(AccountAndContainerInjector.class);
-  private static final String NAMED_BLOB_PREFIX = "/named";
 
   private final AccountService accountService;
   private final FrontendMetrics frontendMetrics;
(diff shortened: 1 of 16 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/AccountAndContainerInjector.java
```diff
@@ -240,7 +239,7 @@ private void injectAccountAndContainerUsingAccountAndContainerHeaders(RestReques
    */
   private void injectAccountAndContainerUsingAccountAndContainerUri(RestRequest restRequest,
       RestRequestMetricsGroup metricsGroup) throws RestServiceException {
-    NamedBlobPath namedBlobPath = RestUtils.parseInput(RestUtils.getRequestPath(restRequest).getOperationOrBlobId(false));
+    NamedBlobPath namedBlobPath = parseNamedBlobPath(RestUtils.getRequestPath(restRequest).getOperationOrBlobId(false));
```
SophieGuo410, line 242: minor: Although it's the original code, but I guess maybe we can remove the RestUtils for getRequestPath as well for consistency right?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -274,7 +289,22 @@
...
       try (Connection connection = dcToDataSource.get(localDatacenter).getConnection()) {
-        future.complete(transaction.run(account.getId(), container.getId(), connection));
+        T result;
+        if (autoCommit) {
+          result = transaction.run(account.getId(), container.getId(), connection);
+        } else {
+          connection.setAutoCommit(false);
+          try {
+            result = transaction.run(account.getId(), container.getId(), connection);
+            connection.commit();
```
SophieGuo410, line 299: Can you give me a brief introduction about what's the connection.commit() and connection.rollback() are doing? a little bit confuse about it. Thank you!

SophieGuo410 (commented):
Thanks for integrate NamedBlobDb and support delete, learned a lot while reviewing this pr.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
cgtz: I will update this. The way I am using variables is a little wonky and soon to be deprecated behavior in mysql. I think I will update the delete query to do a 2 step transaction:
1. select the current row to give us the blob ID
2. if the current row is present and not already deleted, update (set deletion time)
Since this is two operations that we want to be an atomic unit we have to manually commit the transaction like so.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -65,6 +73,11 @@ public void close() {
       isOpen = false;
     }
 
+    @Override
+    public Future<String> convert(RestRequest restRequest, String input, Callback<String> callback) {
+      return convert(restRequest, input, null, callback);
```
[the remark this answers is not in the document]
cgtz: This method is used for get/put/delete/ttlUpdate operations where there is no BlobInfo to provide. The NamedBlobPutHandler uses the second convert method that takes in BlobInfo.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -73,29 +86,30 @@ public void close() {
...
       final CompletableFuture<String> future = new CompletableFuture<>();
       String convertedId = null;
       Exception exception = null;
       frontendMetrics.idConverterRequestRate.mark();
       long startTimeInMs = System.currentTimeMillis();
       try {
         if (!isOpen) {
-        frontendMetrics.idConverterProcessingTimeInMs.update(System.currentTimeMillis() - startTimeInMs);
-        exception = new RestServiceException("IdConverter is closed", RestServiceErrorCode.ServiceUnavailable);
+          frontendMetrics.idConverterProcessingTimeInMs.update(System.currentTimeMillis() - startTimeInMs);
```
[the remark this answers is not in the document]
cgtz: Good catch, this was redundant

zzmao:
Just went through the code. Will take another round of review tomorrow. 

2 basic questions:

The `NamedBlobPath` is constructed by `String accountName, String containerName, String blobName`. 

It looks like `/` is not allowed in blobName to indicate a user defined path. What's intention of not supporting user defined path? 

Is URL encoding/decoding needed for blobName part? (Can people may use space or UTF8 chars in blobName? )

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -245,26 +245,29 @@ private PutBlobOptions getPutBlobOptionsFromRequest() throws RestServiceExceptio
     }
 
     /**
-     * After reading the body of the stitch request
-     * @param channel
-     * @param blobInfo
-     * @return
+     * After reading the body of the stitch request, parse the request body,
+     * and make a call to {@link Router#stitchBlob}.
+     * @param channel the
```
zzmao, line 250: Incomplete java doc

zzmao:
> @zzmao
> 
> > The `NamedBlobPath` is constructed by `String accountName, String containerName, String blobName`.
> > It looks like `/` is not allowed in blobName to indicate a user defined path. What's intention of not supporting user defined path?
> 
> With this change, slashes will be allowed in the user defined path. By using `path.split("/", 4)`, the split string will include up to 4 parts and any extra `/` remaining in the path will go into the fourth part.
> 
> So `named/acct-a/cont-a/a/b/c` will split to `[acct-a, cont-a, a/b/c]`.
> 
> > Is URL encoding/decoding needed for blobName part? (Can people may use space or UTF8 chars in blobName? )
> 
> I think that it is a good practice to use URL encoding when possible. Spaces and international characters may be used in file names so it is useful to support them here as well. You bring up a good point though that we have to think about safe limitations on the blob name. In a future PR I may introduce additional validation for "dangerous" characters. Another alternative is to only support URL safe characters for now and then expand the scope as needed. What do you think?

I get confused here actually. If the browser send url encoded request, does our netty layer decode it? 
I am good with a future PR to validate "dangerous" characters.
My understanding is that we are only supporting URL safe characters to this PR.

cgtz:
> I get confused here actually. If the browser send url encoded request, does our netty layer decode it?
> I am good with a future PR to validate "dangerous" characters.
> My understanding is that we are only supporting URL safe characters to this PR.

You are right, I made the wrong assumption about what NettyRequest does. I thought it would leave URL encoded characters in the path as is, but that is not true since we use the `QueryStringDecoder::path` method instead of `QueryStringDecoder::rawPath`. This means that I should not be URL decoding a second time when parsing a path.
```java
    QueryStringDecoder query = new QueryStringDecoder("/a/b%20/c?key=val%20ue");
    System.out.println(query.path());
    System.out.println(query.rawPath());
    System.out.println(query.parameters());
```
returns
```
/a/b /c
/a/b%20/c
{key=[val ue]}
```

I will update the PR.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
cgtz: Fixed

SophieGuo410 (approved):
LGTM!

zzmao (approved):
LGTM

PR #1712, opened 2020-12-02
[StorageQuota] Add a storage usage refresher on top of mysql

justinlin-linkedin:
Add mysql implementation of storage usage refresher.

### DIFF
--- a/ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
@@ -55,7 +55,7 @@ import org.slf4j.LoggerFactory;
  *
  * At the beginning of each month, aggregation task will take a snapshot of container storage usage and save this snapshot
  * on a mysql table. Aggregation task will not change this table until next month. The container storage usage in this
- * table becomes a monthly base for this month's container usage. And storage usage exposed by this refresher *represents
+ * table becomes a monthly base for this month's container usage. And storage usage exposed by this refresher represents
  * how many bytes containers used in this month. So it's monthly usage, not total usage.
  *
  * For example, in the beginning of Dec 2020, container A[100]_C[1]'s total container usage is 100GB, then this usage will
--- a/ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
@@ -167,9 +167,8 @@ public class MySqlStorageUsageRefresher implements StorageUsageRefresher {
     try {
       String monthValue = accountStatsMySqlStore.queryRecordedMonth(clusterMapConfig.clusterMapClusterName);
       if (monthValue.equals(currentMonth)) {
-        Map<String, Map<String, Long>> base =
+        containerStorageUsageMonthlyBase =
             accountStatsMySqlStore.queryMonthlyAggregatedStats(clusterMapConfig.clusterMapClusterName);
-        containerStorageUsageMonthlyBase = base;
       } else {
         shouldRetry = true;
       }
@@ -179,17 +178,18 @@ public class MySqlStorageUsageRefresher implements StorageUsageRefresher {
     }
 
     if (shouldRetry) {
-      if (retries >= config.mysqlStoreRetryMax) {
+      if (retries >= config.mysqlStoreRetryMaxCount) {
         logger.error("Failed to refresh monthly storage usage after " + retries + " retries, will skip for month: "
             + currentMonth);
         retries = 0;
+        scheduleStorageUsageMonthlyBaseFetcher();
+        return;
       }
       retries++;
       if (config.mysqlStoreRetryBackoffMs != 0) {
         scheduler.schedule(this::fetchMonthlyStorageUsageAndMaybeRetry, config.mysqlStoreRetryBackoffMs,
             TimeUnit.MILLISECONDS);
       }
-      return;
     } else {
       retries = 0;
       tryPersistMonthlyUsage();
--- a/ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
@@ -1,0 +1,470 @@
+/*
+ * Copyright 2020 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
```diff
@@ -0,0 +1,470 @@
...
+      if (retries >= config.mysqlStoreRetryMax) {
+        logger.error("Failed to refresh monthly storage usage after " + retries + " retries, will skip for month: "
+            + currentMonth);
+        retries = 0;
+      }
+      retries++;
+      if (config.mysqlStoreRetryBackoffMs != 0) {
+        scheduler.schedule(this::fetchMonthlyStorageUsageAndMaybeRetry, config.mysqlStoreRetryBackoffMs,
+            TimeUnit.MILLISECONDS);
+      }
```
jsjtzyy, line 191: Could you explain the logic here?  Looks like if `retries` exceeds max count, it resets `retries` , increments its value and  schedule another round of retry ?

ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
```diff
@@ -0,0 +1,470 @@
...
+      return;
```
jsjtzyy, line 192: `return` can be removed

ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
```diff
@@ -0,0 +1,470 @@
...
+   * retry for several times before giving up.
+   */
+  private void fetchMonthlyStorageUsageAndMaybeRetry() {
+    boolean shouldRetry = false;
+    try {
+      String monthValue = accountStatsMySqlStore.queryRecordedMonth(clusterMapConfig.clusterMapClusterName);
+      if (monthValue.equals(currentMonth)) {
+        Map<String, Map<String, Long>> base =
+            accountStatsMySqlStore.queryMonthlyAggregatedStats(clusterMapConfig.clusterMapClusterName);
+        containerStorageUsageMonthlyBase = base;
```
jsjtzyy, line 172: Why we need local variable `base`?

ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
```diff
@@ -0,0 +1,470 @@
...
+    return backupFileManager;
+  }
+
+  /**
+   * Fetch the storage usage and schedule a task to periodically refresh the storage usage.
+   */
+  private void initialFetchAndSchedule() {
+    Runnable updater = () -> {
+      try {
+        Map<String, Map<String, Long>> base = containerStorageUsageMonthlyBase;
```
jsjtzyy, line 239: In bootstrap case or `containerStorageUsageMonthlyBase` cannot be fetched from Mysql or local backup, shouldn't we give it an empty map at line 85?

justinlin-linkedin: It's kinda dangerous to do so. If initialing monthly base failed and we give it an empty map, and somehow fetching total storage usage succeeded, then the container usage for this month would highly likely to surpass the quota. For instance, if a container's total usage is 100GB and the monthly base is 95 GB and the quota is 10 GB. If everything works fine, then for this month, it already used 100 - 95 = 5GB, the quota is 10GB, it still has 5GB to use. But if we give monthly base an empty map, then for this month, we will consider this container using 100GB already.

This is a tricky one, I suppose we don't want to stop the process just because we can't get the monthly base for quota service, but we still want to have a correct logic going forward. That's why I come to this compromise. Monthly base would be highly likely fetched from backup files but we still fail the process if we can't initialize it from database. However, we don't fail the process if we can't get it later.

ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
justinlin-linkedin: I have made a mistake. I should schedule for next fetch after resetting the retries back to 0 and return.

The logic goes like this, assume max retry count is 3. Then after starting to fetching the monthly base, we can try it for 3 times. If for all this 3 times, it failed, then we would just schedule for next fetch.

ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
```diff
@@ -0,0 +1,470 @@
...
+      logger.error("Failed to refresh monthly storage usage", e);
+      shouldRetry = true;
+    }
+
+    if (shouldRetry) {
+      if (retries >= config.mysqlStoreRetryMaxCount) {
+        logger.error("Failed to refresh monthly storage usage after " + retries + " retries, will skip for month: "
+            + currentMonth);
+        retries = 0;
+        scheduleStorageUsageMonthlyBaseFetcher();
```
ankagrawal, line 185: So in this case is the base value the last month's value? If yes, then can customers be penalized for our failure to get monthly base. 
I am worried about the following scenario:
example accountId: A, containerId: B, container quota: 12G
current date: Dec 2
Base Value Nov = 85G
Base Value Dec = 90G (this could not be fetched due to some reason), so containerStorageUsageMonthlyBase["A"]["B"] = 85G
Current Usage Value for container = 100G
Whats our response in this case?

build.gradle
```diff
@@ -448,6 +448,7 @@ project(':ambry-named-mysql') {
 project(':ambry-quota') {
     dependencies {
         compile project(':ambry-api'),
+            project(':ambry-server'),
```
ankagrawal, line 451: Why does ambry-quota need ambry-server? It should be the other way round.

ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
```diff
@@ -0,0 +1,470 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A {@link StorageUsageRefresher} implementation on top of mysql database.
+ *
+ * At the beginning of each month, aggregation task will take a snapshot of container storage usage and save this snapshot
+ * on a mysql table. Aggregation task will not change this table until next month. The container storage usage in this
+ * table becomes a monthly base for this month's container usage. And storage usage exposed by this refresher *represents
```
ankagrawal, line 58: Nit: extra "*" in comment.

ankagrawal (commented):
Overall looks good to me. Very well written code and very well documented!!
Just one concern in the comment.

ankagrawal (approved):
LGTM!

ambry-quota/src/main/java/com/github/ambry/quota/MySqlStorageUsageRefresher.java
jsjtzyy: I see your point, but in the edge case aforementioned, is it possible that `containerStorageUsageMonthlyBase` is not initialized?  I have a minor concern the "base" might be null and there is no check before passing it to `subtract` method where NPE may occur.

jsjtzyy (approved):

PR #1714, opened 2020-12-03
[Container API] Check status of existing container before creating same container

jsjtzyy:
This PR fixes an edge case where user may attempt to create same container even though the container is already deprecated.
There are two posssible states to handle:
1. if existing container is INACTIVE, Ambry throws ResourceHasGone exception indicating the container is unrecoverable.
2. if existing container is DELETE_IN_PROGRESS, Ambry throws MethodNotAllowed exception. User is able to cancel deprecation to restore the whole container.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
@@ -115,12 +115,12 @@ abstract class AbstractAccountService implements AccountService {
         if (existingContainer != null) {
           switch (existingContainer.getStatus()) {
             case INACTIVE:
-              throw new AccountServiceException("The container has gone and cannot be restored",
+              throw new AccountServiceException(
+                  "The container " + container.getName() + " has gone and cannot be restored",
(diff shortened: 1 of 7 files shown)


ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
```diff
@@ -111,18 +111,28 @@ public Account getAccountById(short id) {
...
-            // But include it in the list returned to caller to provide the containerId.
-            logger.info("Request to create container with existing name and properties: {}",
-                existingContainer.toJson().toString());
-            existingUnchangedContainers.add(existingContainer);
-          } else {
-            throw new AccountServiceException("There is a conflicting container in account " + accountName,
-                AccountServiceErrorCode.ResourceConflict);
+          switch (existingContainer.getStatus()) {
+            case INACTIVE:
+              throw new AccountServiceException("The container has gone and cannot be restored",
```
lightningrob, line 118: Include container name in error messages.

lightningrob (approved):

Arun-LinkedIn (approved):

PR #1715, opened 2020-12-04
[ContentEncoding] Add two fields in blob property

SophieGuo410:
This PR includes two main updates:
1. Adding two new fields in blobProperty.
2. Introduce support in the deserializer.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
+++ b/ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
@@ -201,21 +201,9 @@ public class BlobProperties {
     StringBuilder sb = new StringBuilder();
     sb.append("BlobProperties[");
     sb.append("BlobSize=").append(getBlobSize());
-    if (getContentType() != null) {
-      sb.append(", ").append("ContentType=").append(getContentType());
-    } else {
-      sb.append(", ").append("ContentType=Null");
-    }
-    if (getOwnerId() != null) {
-      sb.append(", ").append("OwnerId=").append(getOwnerId());
-    } else {
-      sb.append(", ").append("OwnerId=Null");
-    }
-    if (getServiceId() != null) {
-      sb.append(", ").append("ServiceId=").append(getServiceId());
-    } else {
-      sb.append(", ").append("ServiceId=Null");
-    }
+    sb.append(", ").append("ContentType=").append(getContentType());
+    sb.append(", ").append("OwnerId=").append(getOwnerId());
+    sb.append(", ").append("ServiceId=").append(getServiceId());
     sb.append(", ").append("IsPrivate=").append(isPrivate());
     sb.append(", ").append("CreationTimeInMs=").append(getCreationTimeInMs());
     if (getTimeToLiveInSeconds() != Utils.Infinite_Time) {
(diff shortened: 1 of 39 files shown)


ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
```diff
@@ -209,6 +227,16 @@ public String toString() {
     sb.append(", ").append("ContainerId=").append(getContainerId());
     sb.append(", ").append("IsEncrypted=").append(isEncrypted());
     sb.append(", ").append("externalAssetTag=").append(getExternalAssetTag());
+    if (getContentEncoding() != null) {
+      sb.append(", ").append("ContentEncoding=").append(getContentEncoding());
+    } else {
+      sb.append(", ").append("ContentEncoding=Null");
```
cgtz, line 233: I think this null handling case is not actually required, although I see that this pattern was used before in this method.

`sb.append("ContentEncoding=").append(null)` will work and return "ContentEncoding=null"

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -314,6 +320,12 @@ private void setGetBlobResponseHeaders(BlobInfo blobInfo, GetBlobOptions options
         restResponseChannel.setHeader("Content-Disposition", "attachment");
       }
     }
+    if (blobProperties.getContentEncoding() != null) {
+      restResponseChannel.setHeader(Headers.AMBRY_CONTENT_ENCODING, blobProperties.getContentEncoding());
```
cgtz, line 324: For a get blob response, this should be set as the HTTP standard `Content-Encoding` header instead of `x-ambry-content-encoding`

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -314,6 +320,12 @@ private void setGetBlobResponseHeaders(BlobInfo blobInfo, GetBlobOptions options
...
+    }
+    if (blobProperties.getFilename() != null) {
+      restResponseChannel.setHeader(Headers.AMBRY_FILENAME, blobProperties.getFilename());
```
cgtz, line 327: In our LinkedIn SecurityService, we use the filename to set the value of the Content-Disposition. Not sure if we want to port that behavior here.

ambry-replication/src/test/java/com/github/ambry/replication/BlobIdTransformerTest.java
```diff
@@ -527,7 +527,7 @@ private Message buildMessage(String blobIdString, Class clazz, boolean hasEncryp
       MessageInfo messageInfo;
       BlobProperties blobProperties =
           new BlobProperties(blobPropertiesSize, "serviceId", "ownerId", "contentType", false, 0, 0,
-              blobId.getAccountId(), blobId.getContainerId(), hasEncryption, null);
+              blobId.getAccountId(), blobId.getContainerId(), hasEncryption, null, "gzip", "filename");
```
jsjtzyy, line 530: minor: is there a unit test to verify these two properties in BlobIdTransformerTest?

jsjtzyy (approved):
Lgtm after Casey's comments are addressed.

ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
SophieGuo410: Removed all the redundant part.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
SophieGuo410: Sure. Thanks for catching that, fixed it.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
SophieGuo410: Removed.

ambry-replication/src/test/java/com/github/ambry/replication/BlobIdTransformerTest.java
SophieGuo410: Will add the test cases once the serialize part is done.

cgtz (approved):
lgtm!

PR #1734, opened 2020-12-19
Encode user metadata value if it contains non-ascii character

jsjtzyy:
User metadata now supports UTF-8 charset. However, when setting header that includes non-ascii characters, underlying netty replaces these characters with question mark without throwing any exception. This PR explicitly checks if header value contains non-ascii characters. If yes, it encodes the string and associate it with a special header, which will be identified and decoded by client.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
@@ -15,8 +15,6 @@ package com.github.ambry.rest;
 
 import com.github.ambry.account.Account;
 import com.github.ambry.account.Container;
-import com.github.ambry.frontend.NamedBlobPath;
-import com.github.ambry.frontend.Operations;
 import com.github.ambry.messageformat.BlobProperties;
 import com.github.ambry.protocol.GetOption;
 import com.github.ambry.router.ByteRange;
@@ -910,21 +908,16 @@ public class RestUtils {
       if (!headerName.startsWith(Headers.USER_META_DATA_HEADER_PREFIX)) {
         headerName = Headers.USER_META_DATA_HEADER_PREFIX + headerName;
       }
-      try {
-        restResponseChannel.setHeader(headerName, headerValue);
-      } catch (IllegalArgumentException iae) {
+      if (StandardCharsets.US_ASCII.newEncoder().canEncode(headerValue)) {
         try {
-          // Value may require encoding to set in response header.
-          // Set in special header designated for encoded values.  The client receiving the response
-          // will need to check these headers and perform the decode.
-          headerName = headerName.replaceFirst(Headers.USER_META_DATA_HEADER_PREFIX,
-              Headers.USER_META_DATA_ENCODED_HEADER_PREFIX);
-          headerValue = URLEncoder.encode(headerValue, CHARSET.name());
           restResponseChannel.setHeader(headerName, headerValue);
-          logger.debug("Set encoded value in response header {}", headerName);
-        } catch (Exception e) {
-          logger.error("Unable to set response header {} to value {}", headerName, headerValue, e);
+        } catch (IllegalArgumentException iae) {
+          // if responseChannel fails to set header due to special characters like "\n\r", we encode it and add to a special header
+          encodeMetadataValueAndSetHeader(headerName, headerValue, restResponseChannel);
         }
+      } else {
+        // if the header value contains non-ascii character, we explicitly encode the value and set in response header
+        encodeMetadataValueAndSetHeader(headerName, headerValue, restResponseChannel);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


ambry-protocol/src/test/java/com/github/ambry/protocol/RequestResponseTest.java
```diff
@@ -369,14 +364,12 @@ public void putRequestResponseTest() throws IOException {
         blob, blobSize, blobKey);
 
     // Response test
-    for (boolean useByteBufContent : new boolean[]{true, false}) {
```
jsjtzyy, line 372: Changes in this file are some cleanup and not related to user metadata change. (there was a travis failure, although this doesn't really fix the intermittent test error, it removes redundant code)

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -910,21 +908,16 @@ public static void setUserMetadataHeaders(Map<String, String> userMetadataMap,
       if (!headerName.startsWith(Headers.USER_META_DATA_HEADER_PREFIX)) {
         headerName = Headers.USER_META_DATA_HEADER_PREFIX + headerName;
       }
-      try {
-        restResponseChannel.setHeader(headerName, headerValue);
-      } catch (IllegalArgumentException iae) {
+      if (StandardCharsets.US_ASCII.newEncoder().canEncode(headerValue)) {
```
lightningrob, line 911: What happens if the value contains non ascii characters and you pass it to setHeader() as before?  Does it not throw IllegalArgumentException?

jsjtzyy: It's hard to say, I would prefer merging @cgtz 's ambry-client change and verify. If it doesn't work (not able to display foreign language characters), we can use this piece of code to explicitly and proactively encode the head value.

cgtz: Before, setHeader() would convert the value into valid ASCII by replacing any non-ascii characters with "?" symbols. If , after this step, this ascii encoded string did not contain any characters prohibited in headers but still valid ascii (like newline), the setHeader() call will succeed. However, if it did contain something like a newline, it would throw an exception.

cgtz: Here is a test case showing how Netty behaves (== how `NettyResponseChannel` behaves):
```java

  @Test
  public void headerBehavior() {
    HttpResponse responseMetadata = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);

    String nonASCIIStr = "å∫ \uD83D\uDE1D\uD83D\uDE31abcd";
    responseMetadata.headers().set("non-ascii", nonASCIIStr);

    try {
      responseMetadata.headers().set("ascii-with-non-header-chars", "word word \n\r next line");
    } catch (Exception exception) {
      exception.printStackTrace();
    }

    System.out.println(responseMetadata.headers());
  }
```

results in 
```
java.lang.IllegalArgumentException: only ' ' and '\t' are allowed after '\n': word word 
 next line
DefaultHttpHeaders[non-ascii: å∫ 😝😱abcd]
```

cgtz: Here is the actual behavior. The conversion from unicode to question mark happens in the HttpResponseEncoder:
```java
  @Test
  public void headerBehavior() {
    HttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK);

    String nonASCIIStr = "å∫ \uD83D\uDE1D\uD83D\uDE31abcd";
    response.headers().set("non-ascii", nonASCIIStr);

    try {
      response.headers().set("ascii-with-non-header-chars", "word word \n\r next line");
    } catch (Exception exception) {
      System.out.println("Got exception while setting \n\r in httpheaders:  " + exception);
    }

    System.out.println("HttpHeaders object" + response.headers());

    EmbeddedChannel channel = new EmbeddedChannel(new HttpResponseEncoder());
    response.headers().set(HttpHeaderNames.TRANSFER_ENCODING, HttpHeaderValues.CHUNKED);
    assertTrue(channel.writeOutbound(response));
    channel.flushOutbound();
    System.out.println("Actual HTTP response buffer:");
    ByteBuf buffer = channel.readOutbound();
    System.out.println(buffer.toString(Charset.defaultCharset()));
  }
```

```
Got exception while setting 
 in httpheaders:  java.lang.IllegalArgumentException: only ' ' and '\t' are allowed after '\n': word word 
 next line
HttpHeaders objectDefaultHttpHeaders[non-ascii: å∫ 😝😱abcd]
Actual HTTP response buffer:
HTTP/1.1 200 OK
non-ascii: �? ????abcd
transfer-encoding: chunked
```

jsjtzyy: Good find! Thanks for writing a unit test to narrow down the root cause. So, we can conclude that if head value has non-ascii but no new line character, `setHeader()` doesn't throw exception and non-ascii characters will be silently replaced with question mark in `HttpResponseEncoder`.

lightningrob (approved):

cgtz (approved):

PR #1735, opened 2020-12-21
[named_blobs] Avoid bad affected row count assumptions

cgtz:
Previously, the put code used an "insert on duplicate key" expression
with conditional assignments. However, this required using the affected
rows to count how many rows actually changed. If the DB is set up to use
triggers that assign other metadata columns, the row may change even if
the specific fields set in the update call did not.

As an alternative, we can switch to the default "found rows" jdbc
behavior and use an optimistic insert followed by a conditional update.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
@@ -146,6 +146,100 @@ public class NamedBlobPutHandler {
     }
 
     /**
+     * After {@link SecurityService#processRequest} finishes, call {@link SecurityService#postProcessRequest} to perform
+     * request time security checks that rely on the request being fully parsed and any additional arguments set.
+     * @param blobInfo the {@link BlobInfo} to carry to future stages.
+     * @return a {@link Callback} to be used with {@link SecurityService#processRequest}.
+     */
+    private Callback<Void> securityProcessRequestCallback(BlobInfo blobInfo) {
+      return buildCallback(frontendMetrics.putSecurityProcessRequestMetrics,
+          securityCheckResult -> securityService.postProcessRequest(restRequest,
+              securityPostProcessRequestCallback(blobInfo)), uri, LOGGER, finalCallback);
+    }
+
+    /**
+     * After {@link SecurityService#postProcessRequest} finishes, call {@link Router#putBlob} to persist the blob in the
+     * storage layer.
+     * @param blobInfo the {@link BlobInfo} to make the router call with.
+     * @return a {@link Callback} to be used with {@link SecurityService#postProcessRequest}.
+     */
+    private Callback<Void> securityPostProcessRequestCallback(BlobInfo blobInfo) {
+      return buildCallback(frontendMetrics.putSecurityPostProcessRequestMetrics, securityCheckResult -> {
+        if (STITCH.equals(
+            RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.UPLOAD_NAMED_BLOB_MODE, false))) {
+          RetainingAsyncWritableChannel channel =
+              new RetainingAsyncWritableChannel(frontendConfig.maxJsonRequestSizeBytes);
+          restRequest.readInto(channel, fetchStitchRequestBodyCallback(channel, blobInfo));
+        } else {
+          PutBlobOptions options = getPutBlobOptionsFromRequest();
+          router.putBlob(blobInfo.getBlobProperties(), blobInfo.getUserMetadata(), restRequest, options,
+              routerPutBlobCallback(blobInfo));
+        }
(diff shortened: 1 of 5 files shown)


SophieGuo410 (approved):
LGTM!

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -145,48 +145,6 @@ private void start() {
       }
     }
 
-    /**
```
lightningrob, line 148: Is there any logic change in this class or only refactoring?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -84,27 +83,32 @@
...
-   * Eventually the put operation will support updates to the mapping in the third case above, but that requires other
-   * work around concurrency control.
+   * Attempt to insert a new mapping into the database.
    */
-  private static final String PUT_QUERY = String.format(
-      "INSERT INTO %1$s (%2$s, %3$s, %4$s, %5$s, %6$s) VALUES (?, ?, ?, ?, ?) ON DUPLICATE KEY UPDATE "
-          + "%5$s = IF(%8$s, VALUES(%5$s), %5$s), %6$s = IF(%8$s, VALUES(%6$s), %6$s), %7$s = IF(%8$s, null, %7$s)",
-      NAMED_BLOBS, ACCOUNT_ID, CONTAINER_ID, BLOB_NAME, BLOB_ID, EXPIRES_TS, DELETED_TS, IS_DELETED_OR_EXPIRED);
+  private static final String INSERT_QUERY =
+      String.format("INSERT INTO %s (%s, %s, %4$s, %5$s, %6$s) VALUES (?, ?, ?, ?, ?)", NAMED_BLOBS, ACCOUNT_ID,
```
lightningrob, line 89: Still need the indexes (4,5,6)?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -84,27 +83,32 @@
...
+          CONTAINER_ID, BLOB_NAME, BLOB_ID, EXPIRES_TS);
 
-  private static final String SELECT_FOR_DELETE_QUERY =
+  /**
+   * If a record already exists for a named blob, attempt an update if the record in the DB represents an expired or
+   * deleted blob.
+   */
+  private static final String UPDATE_IF_DELETED_OR_EXPIRED_QUERY =
+      String.format("UPDATE %s SET %s = ?, %s = ?, %s = null WHERE %s AND %s", NAMED_BLOBS, BLOB_ID, EXPIRES_TS,
```
lightningrob, line 97: Just curious, why are you updating the blobId column here?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -250,8 +276,8 @@
       }
       // only need to issue an update statement if the row was not already marked as deleted.
```
lightningrob, line 277: What happens to the locked record if the update is not made?  In either case, when does it get unlocked (not sure if this whole thing is inside a transaction)?

lightningrob (approved):
The logic looks good.  In addition to my inline question about record lock, is there the potential for lock escalation that could block other named operations?

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
cgtz: Just refactoring

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
cgtz: You are right that these aren't required. I will make sure to fix this in my next PR

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
cgtz: This would allow one to replace a soft deleted record with a new blob.
A put to the same key when a value was previously deleted is an allowed operation.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
cgtz: Yep this whole thing is inside a transaction. In this case, the row is unlocked when the transaction is finished (`executeTransactionAsync` calls transaction.commit() or rollback())

PR #1736, opened 2020-12-22
[StorageQuota] Fix a mysql query error when parsing month from empty set

justinlin-linkedin:
When there is no data in the table, AccountStatsMysqlStore should return empty string for month query. This PR fixes that.


cgtz (approved):
lgtm!

lightningrob (approved):

PR #1740, opened 2021-01-04
[memory] fix leak on Selector.write error

cgtz:
Switch where the networkSend variable is assigned so that it is assigned
prior to the place where the IOException can occur, allowing the catch
block to release it.

Fix NPE in AmbryRequests.


zzmao (approved):
LGTM

justinlin-linkedin (approved):

PR #1743, opened 2021-01-08
[StorageQuota] Use batch to update account stats tables

justinlin-linkedin:
This PR:
1) Add a general BatchUpdater class to perform batch update for a given sql command
2) Add batch support for updating container storage usage and aggregated container storage usage.
3) Add support for rewriteBatchedStatements for mysql database for better batch performance.

### DIFF
--- a/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
@@ -48,6 +48,7 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
   private StatsWrapper previousStats;
   private final Metrics storeMetrics;
   private final AccountStatsMySqlConfig config;
+  private final boolean batchEnabled;
 
   /**
    * Metrics for {@link AccountStatsMySqlStore}.
@@ -121,6 +122,7 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
     aggregatedaccountReportsDao = new AggregatedAccountReportsDao(dataAccessor, clusterName);
     this.hostnameHelper = hostnameHelper;
     storeMetrics = new AccountStatsMySqlStore.Metrics(registry);
+    this.batchEnabled = this.config.updateBatchSize >= 0;
     if (!Strings.isNullOrEmpty(localBackupFilePath)) {
       // load backup file and this backup is the previous stats
       ObjectMapper objectMapper = new ObjectMapper();
@@ -143,12 +145,12 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
         previousStats == null ? new StatsSnapshot((long) -1, new HashMap<>()) : previousStats.getSnapshot();
     ContainerUsageFunction function = this::tryUpdateStorageUsage;
     ContainerStorageFunctionBatchUpdater batchUpdater = null;
-    if (config.updateBatchSize >= 0) {
+    if (batchEnabled) {
       batchUpdater = new ContainerStorageFunctionBatchUpdater();
       function = batchUpdater;
     }
     applyFunctionToContainerUsageInDifferentStatsSnapshots(statsWrapper.getSnapshot(), prevSnapshot, function);
-    if (config.updateBatchSize >= 0) {
+    if (batchEnabled) {
       batchUpdater.commit();
     }
     previousStats = statsWrapper;
@@ -195,7 +197,7 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
     int batchSize = 0;
     long startTimeMs = System.currentTimeMillis();
     AggregatedAccountReportsDao.AggregatedStorageBatchUpdater batch = null;
-    if (config.updateBatchSize >= 0) {
+    if (batchEnabled) {
       batch = aggregatedaccountReportsDao.new AggregatedStorageBatchUpdater(config.updateBatchSize);
     }
     for (Map.Entry<String, StatsSnapshot> accountMapEntry : snapshot.getSubMap().entrySet()) {
@@ -214,7 +216,7 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
         batchSize++;
       }
     }
-    if (config.updateBatchSize >= 0) {
+    if (batchEnabled) {
       batch.commit();
     }
     storeMetrics.aggregatedPublishTimeMs.update(System.currentTimeMillis() - startTimeMs);
--- a/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
@@ -142,12 +142,14 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
     StatsSnapshot prevSnapshot =
         previousStats == null ? new StatsSnapshot((long) -1, new HashMap<>()) : previousStats.getSnapshot();
     ContainerUsageFunction function = this::tryUpdateStorageUsage;
+    ContainerStorageFunctionBatchUpdater batchUpdater = null;
     if (config.updateBatchSize >= 0) {
-      function = new ContainerStorageFucntionBatchUpdater();
+      batchUpdater = new ContainerStorageFunctionBatchUpdater();
+      function = batchUpdater;
     }
     applyFunctionToContainerUsageInDifferentStatsSnapshots(statsWrapper.getSnapshot(), prevSnapshot, function);
-    if (function instanceof ContainerStorageFucntionBatchUpdater) {
-      ((ContainerStorageFucntionBatchUpdater) function).flush();
+    if (config.updateBatchSize >= 0) {
+      batchUpdater.commit();
     }
     previousStats = statsWrapper;
   }
@@ -357,11 +359,17 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
     }
   }
 
-  class ContainerStorageFucntionBatchUpdater implements ContainerUsageFunction {
+  /**
+   * A wrapper to the {@link AccountReportsDao.StorageBatchUpdater} and an implementation of {@link ContainerUsageFunction}.
+   */
+  class ContainerStorageFunctionBatchUpdater implements ContainerUsageFunction {
     private AccountReportsDao.StorageBatchUpdater batch = null;
     private boolean exceptionCaught = false;
 
-    public ContainerStorageFucntionBatchUpdater() {
+    /**
+     * Constructor to instantiate {@link ContainerStorageFunctionBatchUpdater}.
+     */
+    public ContainerStorageFunctionBatchUpdater() {
       try {
         batch = accountReportsDao.new StorageBatchUpdater(config.updateBatchSize);
       } catch (SQLException e) {
@@ -383,7 +391,10 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
       }
     }
 
-    public void flush() {
+    /**
+     * Wrapper function to {@link AccountReportsDao.StorageBatchUpdater#commit}.
+     */
+    public void commit() {
       if (exceptionCaught) {
         return;
       }
--- a/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
@@ -15,6 +15,7 @@ package com.github.ambry.accountstats;
 
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
+import com.github.ambry.config.AccountStatsMySqlConfig;
 import com.github.ambry.mysql.MySqlDataAccessor;
 import com.github.ambry.mysql.MySqlMetrics;
 import com.github.ambry.mysql.MySqlUtils;
@@ -46,6 +47,7 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
   private final HostnameHelper hostnameHelper;
   private StatsWrapper previousStats;
   private final Metrics storeMetrics;
+  private final AccountStatsMySqlConfig config;
 
   /**
    * Metrics for {@link AccountStatsMySqlStore}.
@@ -84,6 +86,7 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
 
   /**
    * Constructor to create {@link AccountStatsMySqlStore}.
+   * @param config The {@link AccountStatsMySqlConfig}.
    * @param dbEndpoints MySql DB end points.
    * @param localDatacenter The local datacenter name. Endpoints from local datacenter are preferred when creating connection to MySql DB.
    * @param clusterName  The name of the cluster, like Ambry-test.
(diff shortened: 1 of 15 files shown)


ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Add statement to the batch. Since the sql statement provided to the constructor is a prepared statement,
+   * this method only takes a lambda to supply values to the {@link PreparedStatement}.
+   * @param valueSupplier The consumer to supply the values. The {@link PreparedStatement} is the parameter to this
+   *                      consumer. It can also throws an {@link SQLException}.
+   * @throws SQLException
+   */
+  protected void addUpdateToBatch(GenericThrowableConsumer<PreparedStatement, SQLException> valueSupplier)
```
lightningrob, line 94: Recommend making this synchronized.

ambry-mysql/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
```diff
@@ -134,6 +134,20 @@ public synchronized void setAutoCommit(boolean enable) throws SQLException {
...
+   * object.
+   *
+   * @return the current state of this <code>Connection</code> object's
+   *         auto-commit mode
+   * @exception SQLException if a database access error occurs
+   * or this method is called on a closed connection
+   * @see #setAutoCommit
+   */
+  public synchronized boolean getAutoCommmit() throws SQLException {
+    return activeConnection.getAutoCommit();
```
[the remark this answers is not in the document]
lightningrob: Could you update this and the existing methods (setAutoCommit, commit, rollback) to throw IllegalStateException if activeConnection is null?  That would be preferable to NPE if we ever hit it.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -136,8 +141,16 @@ public AccountStatsMySqlStore(List<MySqlUtils.DbEndpoint> dbEndpoints, String lo
   public void storeStats(StatsWrapper statsWrapper) {
     StatsSnapshot prevSnapshot =
         previousStats == null ? new StatsSnapshot((long) -1, new HashMap<>()) : previousStats.getSnapshot();
-    applyFunctionToContainerUsageInDifferentStatsSnapshots(statsWrapper.getSnapshot(), prevSnapshot,
-        this::tryUpdateStorageUsage);
+    ContainerUsageFunction function = this::tryUpdateStorageUsage;
+    ContainerStorageFunctionBatchUpdater batchUpdater = null;
+    if (config.updateBatchSize >= 0) {
```
lightningrob, line 146: I would add class variable boolean batchEnabled, set
`this.batchEnabled = config.updateBatchSize >= 0`
here and reference that throughout.

lightningrob (changes requested):

ambry-mysql/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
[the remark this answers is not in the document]
justinlin-linkedin: updated

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -0,0 +1,138 @@
...
+      logger.error("Failed to add batch on {}", tableName, e);
+      throw e;
+    }
+  }
+
+  /**
+   * Commit all the sql commands added to this batch.
+   * @throws SQLException
+   */
+  public void commit() throws SQLException {
```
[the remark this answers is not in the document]
justinlin-linkedin: rolling back in the catch statement.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -0,0 +1,138 @@
...
+    try {
+      executeBatch();
+      dataAccessor.commit();
+      dataAccessor.setAutoCommit(autoCommit);
```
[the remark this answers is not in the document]
justinlin-linkedin: updated

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -189,10 +208,17 @@ public void storeAggregatedStats(StatsSnapshot snapshot) throws SQLException {
         String containerIdKey = currContainerMapEntry.getKey();
         short containerId = Short.valueOf(containerIdKey.substring(2, containerIdKey.length() - 1));
         long currStorageUsage = currContainerMapEntry.getValue().getValue();
-        aggregatedaccountReportsDao.updateStorageUsage(accountId, containerId, currStorageUsage);
+        if (config.updateBatchSize >= 0) {
```
Arun-LinkedIn, line 211: We can probably use 'batchEnabled' boolean variable instead of config.updateBatchSize.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -0,0 +1,153 @@
...
+   * @throws SQLException
+   */
+  protected synchronized void addUpdateToBatch(GenericThrowableConsumer<PreparedStatement, SQLException> valueSupplier)
+      throws SQLException {
+    try {
+      if (startTime == 0) {
+        startTime = System.currentTimeMillis();
+      }
+      if (maxBatchSize != 0 && currentBatchSize >= maxBatchSize) {
+        executeBatch();
```
Arun-LinkedIn, line 104: Just thinking if it is possible to commit after each batch execution instead of committing one large transaction in the end. Would it require additional logic to avoid rewriting successful batch updates during partial failures.

lightningrob: That's a good point.  You can run into issues if the transaction is too large.  Have you tested it with the number of records you expect in this use case?

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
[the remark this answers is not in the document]
lightningrob: But what if the consumer of this gets an exception midway through and needs to roll back?  They won't ever call commit.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -189,10 +208,17 @@ public void storeAggregatedStats(StatsSnapshot snapshot) throws SQLException {
...
+          batch.addUpdateToBatch(accountId, containerId, currStorageUsage);
```
lightningrob, line 212: If this throws SqlException, you want to catch it and call batch.rollback() since commit will not be called.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -334,4 +360,52 @@ private void tryUpdateStorageUsage(short partitionId, short accountId, short con
...
+     */
+    public void commit() {
+      if (exceptionCaught) {
+        return;
+      }
+      try {
+        batch.commit();
+      } catch (Exception e) {
+        // the exception is already logged
+        exceptionCaught = true;
```
lightningrob, line 407: Minor: not sure this is needed.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -136,8 +143,16 @@ public AccountStatsMySqlStore(List<MySqlUtils.DbEndpoint> dbEndpoints, String lo
   public void storeStats(StatsWrapper statsWrapper) {
     StatsSnapshot prevSnapshot =
         previousStats == null ? new StatsSnapshot((long) -1, new HashMap<>()) : previousStats.getSnapshot();
-    applyFunctionToContainerUsageInDifferentStatsSnapshots(statsWrapper.getSnapshot(), prevSnapshot,
-        this::tryUpdateStorageUsage);
+    ContainerUsageFunction function = this::tryUpdateStorageUsage;
+    ContainerStorageFunctionBatchUpdater batchUpdater = null;
+    if (batchEnabled) {
+      batchUpdater = new ContainerStorageFunctionBatchUpdater();
```
lightningrob, line 149: Would it be better to catch SQLException here (or declare in method signature) than to eat exceptions and march on?

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -334,4 +360,52 @@ private void tryUpdateStorageUsage(short partitionId, short accountId, short con
...
+
+    /**
+     * Constructor to instantiate {@link ContainerStorageFunctionBatchUpdater}.
+     */
+    public ContainerStorageFunctionBatchUpdater() {
+      try {
+        batch = accountReportsDao.new StorageBatchUpdater(config.updateBatchSize);
+      } catch (SQLException e) {
+        // the exception is already logged
+        exceptionCaught = true;
```
lightningrob, line 379: I don't like this approach of eating/ignoring exceptions everywhere.  I realize you are calling it from a method that doesn't have checked exceptions (and might be difficult to add it), but this code does not look clean.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -0,0 +1,153 @@
...
+      if (startTime != 0) {
+        dataAccessor.onSuccess(Write, System.currentTimeMillis() - startTime);
+      }
+    } catch (SQLException e) {
+      dataAccessor.onException(e, Write);
+      logger.error("Failed to commit batch on {}, rolling back, batch size {}", tableName, totalBatchSize, e);
+      throw e;
+    } finally {
+      try {
+        dataAccessor.rollback();
```
lightningrob, line 134: Huh?  Why are you calling rollback in the finally block?  This should only happen in the exception path, no?

justinlin-linkedin: typo here, updated.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
justinlin-linkedin: restructure code a bit, it should be good now.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
justinlin-linkedin: updated.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
justinlin-linkedin: updated

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
justinlin-linkedin: I have different thoughts on this one. The way we structure the single write update is that 1) we have MySqlDataAccessor cache  the prepared statement 2) we have XXXXXDao class to supply values to prepared statement and then execute it, at the same time, catch the exception and rethrow it 3) we have higher functions to just call XXXXDao method to update mysql database. 
AccountStatsMySqlStore is a class at layer 3, it shouldn't deal with the exception. That's why I move the exception handing logic to BatchUpdater. In BatchUpdater, I will roll back when failing to add the parameter to batch.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
justinlin-linkedin: code removed.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -0,0 +1,158 @@
...
+        executeBatchAndCommit();
+      }
+      valueSupplier.accept(statement);
+      statement.addBatch();
+      currentBatchSize++;
+    } catch (SQLException e) {
+      failed = true;
+      dataAccessor.onException(e, BatchUpdate);
+      logger.error("Failed to add batch on {}, already committed {}, rolling back", tableName, committedBatchSize, e);
+      dataAccessor.rollback();
```
Arun-LinkedIn, line 114: Minor: Just thinking if it is safer to call dataAccessor.rollback() before calling dataAccessor.onException() since the latter method has code to close the db connection in case the SQL Exception is a transient one.

Arun-LinkedIn (approved):

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -96,58 +96,63 @@ public BatchUpdater(MySqlDataAccessor dataAccessor, String sql, String tableName
...
-      dataAccessor.onException(e, Write);
-      logger.error("Failed to add batch on {}", tableName, e);
+      failed = true;
+      dataAccessor.onException(e, BatchUpdate);
+      logger.error("Failed to add batch on {}, already committed {}, rolling back", tableName, committedBatchSize, e);
+      dataAccessor.rollback();
       throw e;
+    } finally {
+      if (failed) {
+        dataAccessor.setAutoCommit(autoCommit);
```
lightningrob, line 118: Why only if failed?

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
lightningrob: I don't think this is minor.  The rollback method will throw IllegalStateException if connection has been closed.  Do we have a unit test to simulate this scenario?

lightningrob: I checked the codecov, and the exception clauses are red (not covered).  I recommend adding a BatchUpdater test class or adding test cases to the stats store test to trigger the exceptions.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
justinlin-linkedin: Moving setAutoCommit to rollback method. This way the code looks a bit cleaner.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
justinlin-linkedin: Unit test added.

lightningrob (approved):

PR #1745, opened 2021-01-19
[QUOTA] Move storage quota to a separate package as preparation for quota interfaces

ankagrawal:
This PR just moves the storage quota classes and tests to their own separate package as preparation to the upcoming PR containing the changes for quota interfaces.


justinlin-linkedin (approved):
lgtm

PR #1747, opened 2021-01-20
Collect stats for delete tombstones and token reset events

jsjtzyy:
1. added logic in blobstore stats to count both expired and permanent delete tombstones left by compaction;
2. added metrics to track token reset events in PersistentIndex

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -53,7 +53,7 @@ import org.slf4j.LoggerFactory;
  * a forecast boundary. Stats related requests are either served via these buckets or a separate scan that will walk
  * through the entire index if the request is outside of the forecast boundary.
  */
-public class BlobStoreStats implements StoreStats, Closeable {
+class BlobStoreStats implements StoreStats, Closeable {
   static final String IO_SCHEDULER_JOB_TYPE = "BlobStoreStats";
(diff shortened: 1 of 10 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -521,6 +535,8 @@ private void enqueueNewValue(StoreKey key, IndexValue newValue, IndexValue origi
             indexSegment.getFile().getName(), storeId);
       }
     }
+    // The remaining index entries in keyFinalStates are DELETE tombstones left by compaction (whose associated PUT is not found)
+    updateDeleteTombstoneStats(keyFinalStates.values());
```
justinlin-linkedin, line 539: I don't think we should update delete tombstone for the log segment collection here. So this function is only used for statsBasedCompaction policy to decide which log segments should be compacted. We can solely rely on the periodical collection of account stats from StatsManager's StatsAggregator thread.

justinlin-linkedin (commented):
Some minor comments.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
jsjtzyy: I see your point. Updating delete tombstone stats in this method doesn't bring too much overhead (the stats is already in IndexFinalState). I made this change here with hope that the delete tombstone stats is more likely to be updated.

justinlin-linkedin (approved):

Arun-LinkedIn (approved):

PR #1749, opened 2021-01-21
Reduce stats report size, remove all 0-value snapshots, use v to replace value

justinlin-linkedin:
There are a couple low hanging fruits to reduce the size of the stats report to helix
1) remove all the 0-value snapshots
2) use "v" to replace "value" so the key would be shorter.

Hopefully this would make report a bit smaller.

### DIFF
--- a/ambry-api/src/test/java/com/github/ambry/server/StatsSnapshotTest.java
+++ b/ambry-api/src/test/java/com/github/ambry/server/StatsSnapshotTest.java
@@ -80,7 +80,7 @@ public class StatsSnapshotTest {
 
   @Test
   public void testCopyConstructor() {
-    // Anonymous class with static block initialization, not a good way to create map, it's only for map.
+    // Anonymous class with static block initialization, not a good way to create map, it's only for testing.
     StatsSnapshot original = new StatsSnapshot(60L, new HashMap<String, StatsSnapshot>() {
       {
         put("L1_K1", new StatsSnapshot(30L, new HashMap<String, StatsSnapshot>() {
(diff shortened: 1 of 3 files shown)


lightningrob (approved):
LGTM

ambry-api/src/test/java/com/github/ambry/server/StatsSnapshotTest.java
```diff
@@ -66,5 +66,147 @@ public void deserializeStatsSnapshotTest() throws IOException {
...
+        }));
+        put("L1_K2", new StatsSnapshot(0L, new HashMap<String, StatsSnapshot>() {
+          {
+            put("L2_K3", new StatsSnapshot(70L, null));
+            put("L2_K4", new StatsSnapshot(0L, null));
+          }
+        }));
+        put("L1_K3", new StatsSnapshot(0L, new HashMap<String, StatsSnapshot>() {
+          {
+            put("L2_K5", new StatsSnapshot(20L, new HashMap<String, StatsSnapshot>() {
```
jsjtzyy, line 166: In practice, `L2_K5` initial value `20L` shouldn't happen, right?

jsjtzyy (approved):
LGTM

ambry-api/src/test/java/com/github/ambry/server/StatsSnapshotTest.java
justinlin-linkedin: that's right, I was just being lazy to calculate the values for these snapshots so I call `updateValue` before `removeZeroValueSnapshots`. Anyway, I will put the right values in the constructor.

jsjtzyy (approved):

PR #1750, opened 2021-01-23
Add a configuration to exclude accounts from stats report

justinlin-linkedin:
Add this configuration so we can specify a list of account names to be excluded from health report, to make sure a big account doesn't blow up the health report.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/server/StatsSnapshot.java
+++ b/ambry-api/src/main/java/com/github/ambry/server/StatsSnapshot.java
@@ -154,7 +154,7 @@ public class StatsSnapshot {
           .stream()
           .filter(ent -> ent.getValue().getValue() != 0)
           .peek(ent -> ent.getValue().removeZeroValueSnapshots())
-          .collect(Collectors.toMap(ent -> ent.getKey(), ent -> ent.getValue()));
+          .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
     }
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/server/StatsSnapshot.java
+++ b/ambry-api/src/main/java/com/github/ambry/server/StatsSnapshot.java
@@ -19,10 +19,9 @@ import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonPropertyOrder;
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 
 /**
@@ -151,15 +150,11 @@ public class StatsSnapshot {
     // if we are here, then value is not 0, we can examine all the StatsSnapshots in the subMap and remove the 0-value
     // StatsSnapshot.
     if (subMap != null) {
-      List<String> keysToRemove = new ArrayList<>();
-      for (Map.Entry<String, StatsSnapshot> ent : subMap.entrySet()) {
-        if (ent.getValue().getValue() == 0) {
-          keysToRemove.add(ent.getKey());
-        } else {
-          ent.getValue().removeZeroValueSnapshots();
-        }
-      }
-      keysToRemove.forEach(k -> subMap.remove(k));
+      subMap = subMap.entrySet()
+          .stream()
+          .filter(ent -> ent.getValue().getValue() != 0)
+          .peek(ent -> ent.getValue().removeZeroValueSnapshots())
+          .collect(Collectors.toMap(ent -> ent.getKey(), ent -> ent.getValue()));
     }
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/store/StoreStats.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/StoreStats.java
@@ -17,6 +17,7 @@ package com.github.ambry.store;
 import com.github.ambry.server.StatsReportType;
 import com.github.ambry.server.StatsSnapshot;
 import com.github.ambry.utils.Pair;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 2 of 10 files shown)


ambry-api/src/main/java/com/github/ambry/server/StatsSnapshot.java
```diff
@@ -151,15 +150,11 @@ public void removeZeroValueSnapshots() {
...
-        if (ent.getValue().getValue() == 0) {
-          keysToRemove.add(ent.getKey());
-        } else {
-          ent.getValue().removeZeroValueSnapshots();
-        }
-      }
-      keysToRemove.forEach(k -> subMap.remove(k));
+      subMap = subMap.entrySet()
+          .stream()
+          .filter(ent -> ent.getValue().getValue() != 0)
```
lightningrob, line 155: Another idea to consider (for future PR) is to exclude containers from stats report if value is below a configurable threshold (like 100MB) rather than only zero.  That way, for accounts with many containers, only those with storage above the threshold will be counted (and ones with very little data yet will be excluded).

ambry-api/src/main/java/com/github/ambry/store/StoreStats.java
```diff
@@ -50,11 +51,12 @@
...
    * @param statsReportTypes the specified {@link StatsReportType} to fetch
    * @param referenceTimeInMs the reference time in ms until which deletes and expiration are relevant
+   * @param accountIdToExclude the list of account id to exclude from the {@link StatsSnapshot}.
    * @return a map of {@link StatsReportType} to {@link StatsSnapshot}
    * @throws StoreException
    */
-  Map<StatsReportType, StatsSnapshot> getStatsSnapshots(Set<StatsReportType> statsReportTypes, long referenceTimeInMs)
-      throws StoreException;
+  Map<StatsReportType, StatsSnapshot> getStatsSnapshots(Set<StatsReportType> statsReportTypes, long referenceTimeInMs,
+      List<Short> accountIdToExclude) throws StoreException;
```
lightningrob, line 59: Might be best to avoid new uses of short for account/container ids since they will change soon.  Perhaps Numeric, or round up to Integer?

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -108,6 +112,13 @@
       logger.info("Stats Manager's state change listener registered!");
     }
     this.accountStatsMySqlStore = accountStatsMySqlStore;
+    Function<List<String>, List<Short>> convertAccountNamesToIds = names -> names.stream()
+        .map(name -> accountService.getAccountByName(name))
```
lightningrob, line 116: Just verifying that account service is already initialized (cache populated) before this constructor.  (Probably is.)

lightningrob (changes requested):
Looks good but prefer avoiding use of short for id.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: yep, the caching part should happen in the AccountService's constructor.

ambry-api/src/main/java/com/github/ambry/store/StoreStats.java
justinlin-linkedin: Just talked to Ze, since the account and container ids conversion is not requested by our customers, we are not going to do it any more. I am think that we should probably keep use shorts for account and container ids, so there is no confusion.

ambry-api/src/main/java/com/github/ambry/server/StatsSnapshot.java
justinlin-linkedin: no, sorry about this one. The reason to remove the zero value snapshots is because in bucket reporting, lots of containers might have 0 values and still be included in the reports. Before bucket reporting, if a container has 0 byte usage, then it will not present in the map. But in bucket report, since we keep the previous container usage values and apply the delta values in each bucket, there are chances that containers previously have non-0 values but now have 0 value would be included. This is to fix that. For instance, assume each bucket span is 10 minutes and container A has only one blob(4mb) that would expire in 10 minutes. So if we fetch usages now, we will see container A's value is 4m. Since this blob will be expired in next bucket, when fetching usages after 10 minutes, we will apply the delta in next bucket (-4m) and then get a 0 value. In this case, we don't need to include this container.

Your idea seems quite reasonable. We are giving each container a minimal 1GB as its quota. If a container' usage is less than some threshold,(maybe 500MB, we can do some calculations) we don't have to include it. This will work just fine for storage quota.

lightningrob (approved):

ambry-server/src/test/java/com/github/ambry/server/StatsManagerTest.java
```diff
@@ -136,7 +154,11 @@ public StatsManagerTest() throws Exception {
     ClusterMapConfig clusterMapConfig = new ClusterMapConfig(new VerifiableProperties(properties));
     storeMap = new HashMap<>();
     partitionToSnapshot = new HashMap<>();
-    preAggregatedSnapshot = generateRandomSnapshot().get(StatsReportType.ACCOUNT_REPORT);
+    List<Short> accountIdsToExlude = new ArrayList<>();
+    if (enableAccountExclusion) {
+      accountIdsToExlude.add((short) 0);
+    }
+    preAggregatedSnapshot = generateRandomSnapshot(accountIdsToExlude).get(StatsReportType.ACCOUNT_REPORT);
```
jsjtzyy, line 161: Honestly, I feel like we probably don't need to preclude  account ids in `generateRandomSnapshot`. The "exclude accounts" logic is in BlobStoreStats and is already covered by `BlobStoreStatsTest`.  In the scope of StatsManager, I feel like verify the two lists of excluding accounts in constructor would suffice.

jsjtzyy (approved):
Mostly looks good. A few minor comments.

ambry-server/src/test/java/com/github/ambry/server/StatsManagerTest.java
justinlin-linkedin: OK, then I will remove the parameter and create a new test cases for account exclusion.

jsjtzyy (approved):

PR #1752, opened 2021-01-25
Add LogSegmentName class

cgtz:
Parsing the log segment name to find the position and generation becomes
expensive when used as a comparator in frequently used index data
structures. This commit adds a class that represents a parsed log
segment name with a lighter weight compareTo method.

It is a significant refactoring to replace all usages of log segment
name strings with this class, so I only adopted the new class in Offset,
which should help in the frequently used
PersistentIndex.validIndexSegments and Journal data structures.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/LogSegmentName.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/LogSegmentName.java
@@ -85,9 +85,16 @@ class LogSegmentName implements Comparable<LogSegmentName> {
    * @param name the name string to parse.
    */
   private LogSegmentName(String name) {
-    int separatorIndex = name.indexOf(BlobStore.SEPARATOR);
-    this.position = Long.parseLong(name.substring(0, separatorIndex));
-    this.generation = Long.parseLong(name.substring(name.indexOf(BlobStore.SEPARATOR) + 1));
+    try {
+      int separatorIndex = name.indexOf(BlobStore.SEPARATOR);
+      if (separatorIndex == -1) {
+        throw new IllegalArgumentException("Separator char not found");
+      }
+      this.position = Long.parseLong(name.substring(0, separatorIndex));
+      this.generation = Long.parseLong(name.substring(name.indexOf(BlobStore.SEPARATOR) + 1));
+    } catch (RuntimeException e) {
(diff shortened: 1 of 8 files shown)


ambry-store/src/main/java/com/github/ambry/store/LogSegmentName.java
```diff
@@ -0,0 +1,174 @@
...
+   */
+  private LogSegmentName(long position, long generation) {
+    this.position = position;
+    this.generation = generation;
+  }
+
+  /**
+   * @param name the name string to parse.
+   */
+  private LogSegmentName(String name) {
```
justinlin-linkedin, line 87: nit: use a regexp to verify that name is a valid log segment name?

ambry-store/src/main/java/com/github/ambry/store/LogSegmentName.java
```diff
@@ -0,0 +1,174 @@
...
+    return new LogSegmentName(name);
+  }
+
+  /**
+   * Parse the string form of a log segment name.
+   * @param name the name string to parse.
+   * @return the {@link LogSegmentName}.
+   */
+  static LogSegmentName fromString(String name) {
+    if (name.isEmpty()) {
```
justinlin-linkedin, line 60: what if name is "log_current", shouldn't we return SINGLE_SEGMENT_LOG_NAME as well?

cgtz: An odd thing I found is that there are 2 name forms, the filename and this string name.

In `LogSegmentNameHelper` a filename of "log_current" would be converted to a string name of "".

So, the only valid string names are "" or "{pos}_{gen}".

ambry-store/src/main/java/com/github/ambry/store/LogSegmentName.java
cgtz: Since this constructor is currently called pretty often (whenever new `Offset` constructed), I wanted to make this do the least amount of string parsing possible.

To make the error message better, i wrapped it in a try-catch that will report when the name is invalid.

justinlin-linkedin (approved):
LGTM

zzmao (approved):
LGTM

PR #1753, opened 2021-01-26
[StorageQuota] Add metrics and apply StorageQuotaService to SecurityService

justinlin-linkedin:
Before this PR, we only instantiate a StorageQuotaService and start it, but never apply it to the frontend operations.
This PR starts applying StoargeQuotaService's shouldThrottle method to frontend Post operations.

This PR is doing it by adding this method to AmbrySecurityService's postProcessRequest method. The reason is apply StorageQuotaService here is 
1. PostProcessRequest is invoked before calling router.putBlob method. If an account/container should be throttled, then we have to throttle it before calling router methods.
2. Information about account and container has to be extracted from request before applying StorageQuotaService. The extraction is done before postProcessRequest.
3. HostLevelThrottler applies its logic in the same method.


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -135,6 +140,17 @@ public void postProcessRequest(RestRequest restRequest, Callback<Void> callback)
           exception = e;
         }
       }
+    } else if (restRequest.getRestMethod() == RestMethod.POST && storageQuotaService != null) {
```
ankagrawal, line 143: I think `AsyncRequestResponseHandler::handleRequest` might be a better place to handle the storage quota instead of AmbrySecurityService. The reason is primarily because this is comes earlier than `AmbrySecurityService::processRequest` in the request handling call chain. Also by the time code hits `handleRequest` the `RestRequest` object is already populated so account and container info should be already populated by then.
The reason I am nitpicking this is that if we have enough information to throttle request in  `AsyncRequestResponseHandler::handleRequest` (in form of `RestRequest` object passed to it), then there is no point of carrying request processing forward and adding it to the `AsyncRequestWorker` queue.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -135,6 +140,18 @@ public void postProcessRequest(RestRequest restRequest, Callback<Void> callback)
           exception = e;
         }
       }
+    } else if (restRequest.getRestMethod() == RestMethod.POST && storageQuotaService != null
+        && restRequest.getSize() > 0) {
+      try {
+        Account account = getAccountFromArgs(restRequest.getArgs());
+        Container container = getContainerFromArgs(restRequest.getArgs());
+        if (storageQuotaService.shouldThrottle(account.getId(), container.getId(), QuotaOperation.Post,
+            restRequest.getSize())) {
```
ankagrawal, line 149: Nit: add a comment here that we are optimistically accounting for the request size beforehand and substracting it from quota. This might have edge cases if the request actually fails later.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
ankagrawal: This is ok since we need to get account and container id from the args, which is not available till PostBlobHandler.

ambry-quota/src/main/java/com/github/ambry/quota/storage/AmbryStorageQuotaEnforcer.java
```diff
@@ -69,9 +74,18 @@ public boolean shouldThrottle(short accountId, short containerId, QuotaOperation
             return v;
           }
         });
+    logger.debug("Account id {} container id {} quota {}, existing usage {} new size {}, new usage {}", accountId,
+        containerId, quota == Long.MAX_VALUE ? "MAX_VALUE" : String.valueOf(quota), existingUsage.get(), size,
+        storageUsage.get(String.valueOf(accountId)).get(String.valueOf(containerId)));
+    if (exceedQuota.get()) {
+      metrics.quotaExceededCount.inc();
```
zzmao, line 81: Besides metric, you can also log warning level messages with acct id, container id, blob id, etc for easy debug.

zzmao (approved):
One minor comment

ambry-quota/src/main/java/com/github/ambry/quota/storage/AmbryStorageQuotaEnforcer.java
justinlin-linkedin: I am planning on adding a callback method when the quota exceeded. In ambryli callback method, I would probably send a kafka event or somehow send out an alert. But in ambry callback event, I would just log it out. Some what like NotificationSystem.

ankagrawal (approved):
LGTM

PR #1756, opened 2021-01-28
Replace usages of raw string names with LogSegmentName

cgtz:
Used type migration to replace usages of strings as log segment names
with the dedicated LogSegmentName class throughout the ambry-store
package and tests.

This allows for the usage of the more efficient comparator in many
data structures.


justinlin-linkedin (approved):
Big diff, I didn't review all of them, but I don't see this could go wrong.

zzmao (approved):
no objection

PR #1758, opened 2021-01-30
[StorageQuota] Refactor StorageQuotaService

justinlin-linkedin:
I found a bug in ambryli wrt the newly added logic on storage quota service. Here let's separate the logic to a method so we can reuse it.


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -114,50 +116,40 @@ public void processRequest(RestRequest restRequest, Callback<Void> callback) {
...
+        if (restRequest.getRestMethod() == RestMethod.DELETE || restRequest.getRestMethod() == RestMethod.PUT) {
+          accountAndContainerNamePreconditionCheck(restRequest);
+        } else if (restRequest.getRestMethod() == RestMethod.GET) {
+          RequestPath requestPath = getRequestPath(restRequest);
+          String operationOrBlobId = requestPath.getOperationOrBlobId(true);
+          // ensure that secure path validation is only performed when getting blobs rather than other operations.
+          if (!operationOrBlobId.isEmpty() && !OPERATIONS.contains(operationOrBlobId)) {
+            try {
+              validateSecurePathIfRequired(restRequest, requestPath.getPrefix(), frontendConfig.securePathPrefix);
+            } catch (Exception e) {
```
ankagrawal, line 141: There is no need of catch here. Outside try block can handle this.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -114,50 +116,40 @@ public void processRequest(RestRequest restRequest, Callback<Void> callback) {
 
   @Override
   public void postProcessRequest(RestRequest restRequest, Callback<Void> callback) {
+    if (restRequest == null || callback == null) {
```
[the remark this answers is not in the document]
justinlin-linkedin: no, but the restrequest and callback shouldn't be null. Even in ambryli, we don't check if the restRequest and callback is null.

ambry-api/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
```diff
@@ -65,6 +69,15 @@
    */
   boolean shouldThrottle(short accountId, short containerId, QuotaOperation op, long size);
```
ankagrawal, line 70: Do we need both the `shouldThrottle` methods in the interface. Maybe we can say this shouldThrottle is an implementation detail, and keep only one?

ambry-api/src/main/java/com/github/ambry/quota/storage/StorageQuotaService.java
```diff
@@ -49,6 +50,13 @@
    */
   boolean shouldThrottle(short accountId, short containerId, QuotaOperation op, long size);
```
ankagrawal, line 51: Same comment as for `shouldThrottle` method above.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestServiceFactory.java
```diff
@@ -90,9 +93,12 @@ public RestRequestService getRestRequestService() {
           new AccountAndContainerInjector(accountService, frontendMetrics, frontendConfig);
       StorageQuotaService storageQuotaService = null;
       if (frontendConfig.enableStorageQuotaService) {
+        // TODO: after enabling StatsReport API, add mysql store to frontend rest request service.
```
ankagrawal, line 96: I do not understand this comment. Where are we enabling StatsReport API in FrontendRestRequestServiceFactory?

ambry-api/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
justinlin-linkedin: updated.

ambry-api/src/main/java/com/github/ambry/quota/storage/StorageQuotaService.java
justinlin-linkedin: updated.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestServiceFactory.java
justinlin-linkedin: We are going to expose another HTTP API, /statsReport in frontend, to send back stats report to client. Stats reports are stored in mysql database. AccountStatsStore is the object to interact with mysql database to fetch the stats reports. This API will be handled in FrontendRestRequestService just like any other APIs.That why we need to pass the AccountStatsStore object into the factory. However, StorageQuotaService also need to interact with mysql database, so we have to share the object. That's why I need to create this object outside of storage quota service.

ambry-api/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
zzmao: I see this is changed to `boolean shouldThrottle(RestRequest restRequest);`. This means ,on backend, we need to construct `RestRequest` somewhere?

zzmao:
In @ankagrawal 's PR #1751 , there is a `QuotaEnforcer` interface, here we have a `StorageQuotaEnforcer` interface. Is it possible to unify them?

ankagrawal:
> In @ankagrawal 's PR #1751 , there is a `QuotaEnforcer` interface, here we have a `StorageQuotaEnforcer` interface. Is it possible to unify them?

Right now StorageQuota is working independently. Once the PR#1751 for quota interfaces is merged, I will create a PR to unify StorageQuota with the quota interfaces. Since quota interfaces is still in review we cannot unify in this PR.

ambry-api/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
justinlin-linkedin: we don't need to, the storage quota service is only working in frontend.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -51,7 +53,7 @@
  * sets the respective headers on response.
  */
 class AmbrySecurityService implements SecurityService {
-
+  private static final Logger logger = LoggerFactory.getLogger(AmbrySecurityService.class);
```
zzmao, line 56: logger is not used. Did you want to log anything?

zzmao (approved):
LGTM

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
justinlin-linkedin: no, let me remove it.

ankagrawal (approved):
LGTM

PR #1759, opened 2021-02-01
Schedule azure ABS ad auth token refresh before expiry.

ankagrawal:
Currently ABS token refresh happens only on expiry. If there are some requests that are in flight when the token expires, the azure libraries throw `UnexpectedLengthException`. This fix is an attempt to ensure that token refresh happens before expiry so that inflight requests don't see exceptions.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ADAuthBasedStorageClient.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ADAuthBasedStorageClient.java
@@ -24,6 +24,7 @@ import com.azure.storage.blob.batch.BlobBatchClient;
 import com.azure.storage.blob.models.BlobStorageException;
 import com.azure.storage.common.policy.RequestRetryOptions;
 import com.github.ambry.config.CloudConfig;
+import com.github.ambry.utils.Utils;
 import com.microsoft.aad.msal4j.ClientCredentialFactory;
 import com.microsoft.aad.msal4j.ClientCredentialParameters;
 import com.microsoft.aad.msal4j.ConfidentialClientApplication;
@@ -32,6 +33,9 @@ import java.net.MalformedURLException;
 import java.time.OffsetDateTime;
 import java.util.Collections;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.http.HttpStatus;
 import reactor.core.publisher.Mono;
@@ -41,6 +45,10 @@ import reactor.core.publisher.Mono;
  * {@link StorageClient} implementation for AD based authentication.
  */
 public class ADAuthBasedStorageClient extends StorageClient {
+  private static final String AD_AUTH_TOKEN_REFRESHER_PREFIX = "AdAuthTokenRefresher";
+  private final ScheduledExecutorService tokenRefreshScheduler =
+      Utils.newScheduler(1, AD_AUTH_TOKEN_REFRESHER_PREFIX, false);
+  private final AtomicReference<ScheduledFuture<?>> scheduledFutureRef = new AtomicReference<>(null);
   private AtomicReference<AccessToken> accessTokenRef;
 
   /**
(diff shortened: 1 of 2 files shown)


zzmao:
Is it possible to track the expiration time and do token refresh ahead of the expiration. I am thinking if we can avoid scheduled taks.

```
if (accessToken.getExpiresAt().toEpochSecond() - OffsetDateTime.now().toEpochSecond() < threshold) {
     refreshStorageClient();
}
```

ankagrawal:
> Is it possible to track the expiration time and do token refresh ahead of the expiration. I am thinking if we can avoid scheduled taks.
> 
> ```
> if (accessToken.getExpiresAt().toEpochSecond() - OffsetDateTime.now().toEpochSecond() < threshold) {
>      refreshStorageClient();
> }
> ```

The problem with that is that there are multiple threads using the token at the same time. We will have to proactively check the token expiry every time we make a request to ABS. In case the token is within expiry threshold (meaning we need to expire token because expiry remaining time < threshold), we will have to check if some other thread is already doing the token refresh, and then wait for that thread.
Instead of doing that, with the current logic, all threads will just use the existing the token and make requests to ABS without any checks. The token refresh will happen in the background, and atomically updated so that new requests take the new token automatically. I wanted to avoid the problem of tracking multiple threads and some other multithreading nuances that come with it. Hence this solution.

cgtz:
> > Is it possible to track the expiration time and do token refresh ahead of the expiration. I am thinking if we can avoid scheduled taks.
> > ```
> > if (accessToken.getExpiresAt().toEpochSecond() - OffsetDateTime.now().toEpochSecond() < threshold) {
> >      refreshStorageClient();
> > }
> > ```
> 
> The problem with that is that there are multiple threads using the token at the same time. We will have to proactively check the token expiry every time we make a request to ABS. In case the token is within expiry threshold (meaning we need to expire token because expiry remaining time < threshold), we will have to check if some other thread is already doing the token refresh, and then wait for that thread.
> Instead of doing that, with the current logic, all threads will just use the existing the token and make requests to ABS without any checks. The token refresh will happen in the background, and atomically updated so that new requests take the new token automatically. I wanted to avoid the problem of tracking multiple threads and some other multithreading nuances that come with it. Hence this solution.

You could potentially use an atomic swap to guarantee only one thread does the operation:
```java
AtomicLong lastRefreshTime;

long lastRefreshSnapshot = lastRefreshTime.get();
long currentTime
if (eligibleForNewUpdate(lastRefreshSnapshot, currentTime)) {
  if (lastRefreshTime.compareAndSet(lastRefreshSnapshot, currentTime) {
    doRefresh()
  } else {
    // we know someone else already did it
  }
}
```
```

File: ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ADAuthBasedStorageClient.java [hunk 1]
```diff
@@ -135,14 +158,26 @@ protected boolean handleExceptionAndHintRetry(BlobStorageException blobStorageEx
...
     return false;
   }
+
+  /**
+   * Refreshes AD authentication token and creates a new ABS client with the new token.
+   */
+  private synchronized void refreshStorageClient() {
+    // if a task is scheduled to refresh token then remove it.
+    if (scheduledFutureRef.get() != null) {
+      scheduledFutureRef.get().cancel(false);
```
lightningrob, line 175: Is it necessary to cancel and reschedule the refresh task every time?  buildBlobServiceClient can create it only if necessary.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ADAuthBasedStorageClient.java
```diff
@@ -67,6 +75,17 @@ public ADAuthBasedStorageClient(BlobServiceClient blobServiceClient, BlobBatchCl
     super(blobServiceClient, blobBatchClient, azureMetrics, blobLayoutStrategy);
   }
 
+  /**
+   * Note that this method is not thread safe. Its need to be called from a thread safe context.
```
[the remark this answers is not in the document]
ankagrawal: This method is always called from a thread safe context. The reason for the comment is to avoid thread unsafe usage in future. The only place `buildBlobServiceClient` is called from is `StorageClient::createBlobStorageClient`.

`createBlobStorageClient` itself is called either from `StorageClient`'s constructor or from `ADAuthBasedStorageClient::refreshStorageToken`, both of them are thread safe contexts.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ADAuthBasedStorageClient.java
ankagrawal: I have added the cancel task just to avoid any case where a token might get expired earlier than expected. In that case we don't want token refreshes to happen in quick succession.

zzmao (approved):
LGTM

lightningrob (approved):
lgtm

cgtz (approved):

PR #1760, opened 2021-02-02
Introduce new properties in Account/Container to prepare for account-level ACL

jsjtzyy:
1. Introduced AccountBlobsResource as new type of ACL resource dedicated for account-level ACL.
2. Added "aclInheritedByContainer" property to Account class. If it's true, containers within account will inherit account's ACL.
3. Added "overrideAccountAcl" property to Container class. If it's true, container can override account's ACL even though account-level ACL inheritance is enabled.


lightningrob:
Minor correction to the description: it's AccountBlobsResource, not AmbryBlobsResource.

lightningrob (approved):
LGTM, thanks for adding this.

jsjtzyy:
> Minor correction to the description: it's AccountBlobsResource, not AmbryBlobsResource.

Yeah, thanks for pointing it out. I have updated the description.

Arun-LinkedIn (approved):

PR #1763, opened 2021-02-04
Integrate request quota with frontend.

ankagrawal:
This PR applies the request quota to frontend request flow. Note that in order to complete the integration with frontend in open source code two more components are required which will come as future PRs in the interest of keeping this PR short. The two components are

1. Returning quota response headers containing quota usage and request costs, to response sent from frontend to client.
2. Integrating storage quota with the quota interfaces.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
@@ -36,7 +36,7 @@ public interface QuotaManager {
    * @param restRequest {@link RestRequest} object.
    * @return ThrottlingRecommendation object that captures the overall recommendation.
    */
-    ThrottlingRecommendation getThrottleRecommendation(RestRequest restRequest);
+  ThrottlingRecommendation getThrottleRecommendation(RestRequest restRequest);
 
   /**
    * Computes the overall boolean recommendation to throttle a request or not for all the types of request quotas supported.
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
@@ -36,7 +36,7 @@ public interface QuotaManager {
    * @param restRequest {@link RestRequest} object.
    * @return ThrottlingRecommendation object that captures the overall recommendation.
    */
-  ThrottlingRecommendation shouldThrottle(RestRequest restRequest);
+    ThrottlingRecommendation getThrottleRecommendation(RestRequest restRequest);
 
   /**
    * Computes the overall boolean recommendation to throttle a request or not for all the types of request quotas supported.
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
@@ -128,7 +128,7 @@ class AmbrySecurityService implements SecurityService {
         exception = new RestServiceException("Too many requests", RestServiceErrorCode.TooManyRequests);
       } else if (storageQuotaService != null && storageQuotaService.shouldThrottle(restRequest)) {
         exception = new RestServiceException("StorageQuotaExceeded", RestServiceErrorCode.TooManyRequests);
-      } else if (quotaManager.getQuotaConfig().requestThrottlingEnabled && quotaManager.shouldThrottle(restRequest)
+      } else if (quotaManager.getQuotaConfig().requestThrottlingEnabled && quotaManager.getThrottleRecommendation(restRequest)
           .shouldThrottle()) {
         // TODO Use ThrottlingRecommendation to set response headers about quota usage.
         exception = new RestServiceException("User Quota Exceeded", RestServiceErrorCode.TooManyRequests);
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
@@ -134,7 +134,7 @@ class AmbrySecurityService implements SecurityService {
           ThrottlingRecommendation throttlingRecommendation = quotaManager.getThrottleRecommendation(restRequest);
           if (throttlingRecommendation != null && throttlingRecommendation.shouldThrottle()) {
             // TODO Use ThrottlingRecommendation to set response headers about quota usage.
-            exception = new RestServiceException("User Quota Exceeded", RestServiceErrorCode.TooManyRequests);
+            throw new RestServiceException("User Quota Exceeded", RestServiceErrorCode.TooManyRequests);
           }
         }
         if (restRequest.getRestMethod() == RestMethod.DELETE || restRequest.getRestMethod() == RestMethod.PUT) {
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
@@ -22,6 +22,7 @@ import com.github.ambry.messageformat.BlobInfo;
 import com.github.ambry.messageformat.BlobProperties;
 import com.github.ambry.protocol.GetOption;
 import com.github.ambry.quota.QuotaManager;
+import com.github.ambry.quota.ThrottlingRecommendation;
 import com.github.ambry.quota.storage.StorageQuotaService;
 import com.github.ambry.rest.RequestPath;
 import com.github.ambry.rest.ResponseStatus;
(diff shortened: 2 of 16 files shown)


ambry-rest/src/main/java/com/github/ambry/rest/RestServerMain.java
```diff
@@ -50,8 +54,14 @@ public static void main(String[] args) {
...
       SSLFactory sslFactory = getSSLFactoryIfRequired(verifiableProperties);
+
+      // Quota
+      QuotaConfig quotaConfig = new QuotaConfig(verifiableProperties);
+      QuotaManagerFactory quotaManagerFactory =
+          Utils.getObj(quotaConfig.quotaManagerFactory, quotaConfig, Collections.emptyList(), new MaxThrottlePolicy());
       logger.info("Bootstrapping RestServer");
-      restServer = new RestServer(verifiableProperties, clusterMap, new LoggingNotificationSystem(), sslFactory);
+      restServer = new RestServer(verifiableProperties, clusterMap, new LoggingNotificationSystem(), sslFactory,
+          quotaManagerFactory);
```
justinlin-linkedin, line 64: I wouldn't create QuotaManagerFactory here in RestServerMain. because then we have to do it again in AmbryLi. 
I suggest that we create a QuotaManagerFactory in either RestServer's constructor, or even in FrontendRestRequestServiceFactory's getRestRequestService method.
If we create a QuotaManager in RestServer, then RestServer has to call init() in its start method.

justinlin-linkedin (changes requested):

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -125,6 +128,10 @@ public void postProcessRequest(RestRequest restRequest, Callback<Void> callback)
         exception = new RestServiceException("Too many requests", RestServiceErrorCode.TooManyRequests);
       } else if (storageQuotaService != null && storageQuotaService.shouldThrottle(restRequest)) {
         exception = new RestServiceException("StorageQuotaExceeded", RestServiceErrorCode.TooManyRequests);
+      } else if (quotaManager.getQuotaConfig().requestThrottlingEnabled && quotaManager.shouldThrottle(restRequest)
```
lightningrob, line 131: The double call to shouldThrottle() looks really weird.  I think QuotaManager.shouldThrottle should be renamed if it doesn't actually tell you if you should throttle.  I would also expect the shouldThrottle method to include the config check internally.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityServiceFactory.java
```diff
@@ -34,20 +35,23 @@
   private final FrontendMetrics frontendMetrics;
   private final UrlSigningService urlSigningService;
   private final StorageQuotaService storageQuotaService;
+  private final QuotaManager quotaManager;
 
   public AmbrySecurityServiceFactory(VerifiableProperties verifiableProperties, ClusterMap clusterMap,
       AccountService accountService, UrlSigningService urlSigningService, IdSigningService idSigningService,
-      AccountAndContainerInjector accountAndContainerInjector, StorageQuotaService storageQuotaService) {
+      AccountAndContainerInjector accountAndContainerInjector, StorageQuotaService storageQuotaService,
+      QuotaManager quotaManager) {
```
lightningrob, line 43: Does StorageQuotaService not have its own reference to the QuotaManager?  It seems odd that they both need to be supplied.

ambry-test-utils/src/main/java/com/github/ambry/quota/QuotaTestUtils.java
```diff
@@ -15,6 +15,7 @@
 
 import com.github.ambry.config.QuotaConfig;
 import com.github.ambry.config.VerifiableProperties;
+import java.util.Collections;
```
lightningrob, line 18: Doesn't seem needed.

lightningrob (changes requested):

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityServiceFactory.java
ankagrawal: Yes in the complete implementation, StorageQuota will be part of the QuotaManager. But as noted in the PR comment, I haven't merged StorageQuota with the QuotaManager yet, as I don't want to interfere with Justin's testing, until this thing is ready and is working end to end in closed source.
I will do that in a followup PR.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
ankagrawal: If not for config check, we would have to add a check for `quotaManager != null` , which actually shouldn't be null if `requestThrottlingEnabled` config is set to true. Hence I felt this check here is better than null check. But I can change this if thats the preferred approach.
Will rename `QuotaManager`'s `shouldThrottle`.

ambry-test-utils/src/main/java/com/github/ambry/quota/QuotaTestUtils.java
ankagrawal: removed.

ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
```diff
@@ -36,7 +36,7 @@
    * @param restRequest {@link RestRequest} object.
    * @return ThrottlingRecommendation object that captures the overall recommendation.
    */
-  ThrottlingRecommendation shouldThrottle(RestRequest restRequest);
+    ThrottlingRecommendation getThrottleRecommendation(RestRequest restRequest);
```
lightningrob, line 39: Minor: looks like format issue introduced.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
lightningrob: I didn't really follow that.  If quotaManager is null, quotaManager.getQuotaConfig() will throw a NPE.

lightningrob (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -125,6 +129,12 @@ public void postProcessRequest(RestRequest restRequest, Callback<Void> callback)
         exception = new RestServiceException("Too many requests", RestServiceErrorCode.TooManyRequests);
       } else if (storageQuotaService != null && storageQuotaService.shouldThrottle(restRequest)) {
         exception = new RestServiceException("StorageQuotaExceeded", RestServiceErrorCode.TooManyRequests);
+      } else if (quotaManager != null && quotaManager.getQuotaConfig().requestThrottlingEnabled) {
```
justinlin-linkedin, line 132: This is not right, this `else if` block should be moved to `else` block down below. If `quotaManager is not null and request throttling is enabled`, then the execution flow would go inside this`else if` block and never go to the `else` block down below. This is not what we want to, we should change it like this
```
} else {
    if (quotaManager != null && quotaManager.getQuotaConfig().requestThrottlingEnabled) {
                ThrottlingRecommendation throttlingRecommendation = quotaManager.getThrottleRecommendation(restRequest);
        if (throttlingRecommendation != null && throttlingRecommendation.shouldThrottle()) {
          // TODO Use ThrottlingRecommendation to set response headers about quota usage.
          throw new RestServiceException("User Quota Exceeded", RestServiceErrorCode.TooManyRequests);
        }
    }

    if (restRequest.getRestMethod() == RestMethod.DELETE || ....)
}
```

ambry-rest/src/main/java/com/github/ambry/rest/RestServer.java
```diff
@@ -212,10 +216,15 @@ public RestServer(VerifiableProperties verifiableProperties, ClusterMap clusterM
...
+    // setup Quota
+    QuotaConfig quotaConfig = new QuotaConfig(verifiableProperties);
+    QuotaManagerFactory quotaManagerFactory =
+        Utils.getObj(quotaConfig.quotaManagerFactory, quotaConfig, Collections.emptyList(), new MaxThrottlePolicy());
+
     // setup restRequestService
     RestRequestServiceFactory restRequestServiceFactory =
         Utils.getObj(restServerConfig.restServerRestRequestServiceFactory, verifiableProperties, clusterMap, router,
-            accountService);
+            accountService, quotaManagerFactory.getQuotaManager());
```
justinlin-linkedin, line 227: Who is going to call quotaManager.init()?

justinlin-linkedin (approved):

justinlin-linkedin (changes requested):
I think the ambry security service logic is wrong.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -126,6 +130,13 @@ public void postProcessRequest(RestRequest restRequest, Callback<Void> callback)
       } else if (storageQuotaService != null && storageQuotaService.shouldThrottle(restRequest)) {
         exception = new RestServiceException("StorageQuotaExceeded", RestServiceErrorCode.TooManyRequests);
       } else {
+        if (quotaManager != null) {
+          ThrottlingRecommendation throttlingRecommendation = quotaManager.getThrottleRecommendation(restRequest);
+          if (throttlingRecommendation != null && throttlingRecommendation.shouldThrottle()) {
+            // TODO Use ThrottlingRecommendation to set response headers about quota usage.
+            exception = new RestServiceException("User Quota Exceeded", RestServiceErrorCode.TooManyRequests);
```
justinlin-linkedin, line 137: should throw this exception, if you only set the exception, then code in the rest of this block would still get executed.

justinlin-linkedin (approved):

PR #1765, opened 2021-02-06
Changes to fetch missing Containers from MySql db on demand during blob operations

Arun-LinkedIn:
Changes to:
1. Fetch container on-demand from mysql db during blob operations if it is not found in in-memory cache.
2. Add a local LRA cache for storing container names not found during getContainer() operations to avoid repeated queries to mysql db.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/AccountServiceMetrics.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AccountServiceMetrics.java
@@ -50,6 +50,7 @@ public class AccountServiceMetrics {
   public final Counter accountFetchFromAmbryServerErrorCount;
   public final Counter accountUpdatesToStoreErrorCount;
   public final Counter getAccountInconsistencyCount;
+  public final Counter onDemandContainerFetchCount;
 
   // Gauge
   Gauge<Integer> accountDataInconsistencyCount;
@@ -102,6 +103,8 @@ public class AccountServiceMetrics {
         metricRegistry.counter(MetricRegistry.name(HelixAccountService.class, "AccountFetchFromAmbryServerErrorCount"));
     getAccountInconsistencyCount =
         metricRegistry.counter(MetricRegistry.name(CompositeAccountService.class, "GetAccountInconsistencyCount"));
+    onDemandContainerFetchCount =
+        metricRegistry.counter(MetricRegistry.name(MySqlAccountService.class, "onDemandContainerFetchCount"));
   }
 
   /**
@@ -130,7 +133,6 @@ public class AccountServiceMetrics {
    */
   void trackContainerCount(Gauge<Integer> gauge) {
     containerCountGauge = gauge;
-    metricRegistry.register(MetricRegistry.name(MySqlAccountService.class, "ContainerCount"),
-        containerCountGauge);
+    metricRegistry.register(MetricRegistry.name(MySqlAccountService.class, "ContainerCount"), containerCountGauge);
   }
 }
--- a/ambry-account/src/main/java/com/github/ambry/account/AccountServiceMetrics.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AccountServiceMetrics.java
@@ -25,6 +25,35 @@ import com.codahale.metrics.MetricRegistry;
  * Exports metrics that are triggered by the {@link AccountService} to the provided {@link MetricRegistry}.
  */
 public class AccountServiceMetrics {
+
+  public static final String STARTUP_TIME_MSEC = "StartupTimeInMs";
+  public static final String UPDATE_ACCOUNT_TIME_MSEC = "UpdateAccountTimeInMs";
+  public static final String FETCH_REMOTE_ACCOUNT_TIME_MSEC = "FetchRemoteAccountTimeInMs";
+  public static final String ACCOUNT_UPDATE_CONSUMER_TIME_MSEC = "AccountUpdateConsumerTimeInMs";
+  public static final String ACCOUNT_UPDATE_TO_AMBRY_TIME_MSEC = "AccountUpdateToAmbryTimeInMs";
+  public static final String ACCOUNT_FETCH_FROM_AMBRY_TIME_MSEC = "AccountFetchFromAmbryTimeInMs";
+  public static final String BACKUP_WRITE_TIME_MSEC = "BackupWriteTimeInMs";
+  public static final String BACKUP_READ_TIME_MSEC = "BackupReadTimeInMs";
+  public static final String UNRECOGNIZED_MESSAGE_ERROR_COUNT = "UnrecognizedMessageErrorCount";
+  public static final String NOTIFY_ACCOUNT_DATA_CHANGE_ERROR_COUNT = "NotifyAccountDataChangeErrorCount";
+  public static final String UPDATE_ACCOUNT_ERROR_COUNT = "UpdateAccountErrorCount";
+  public static final String CONFLICT_RETRY_COUNT = "ConflictRetryCount";
+  public static final String ACCOUNT_UPDATES_TO_STORE_ERROR_COUNT = "AccountUpdatesToStoreErrorCount";
+  public static final String FETCH_REMOTE_ACCOUNT_ERROR_COUNT = "FetchRemoteAccountErrorCount";
+  public static final String REMOTE_DATA_CORRUPTION_ERROR_COUNT = "RemoteDataCorruptionErrorCount";
+  public static final String BACKUP_ERROR_COUNT = "BackupErrorCount";
+  public static final String NULL_NOTIFIER_COUNT = "NullNotifierCount";
+  public static final String ACCOUNT_UPDATES_CAPTURED_BY_SCHEDULED_UPDATER_COUNT =
+      "AccountUpdatesCapturedByScheduledUpdaterCount";
+  public static final String ACCOUNT_UPDATES_TO_AMBRY_SERVER_ERROR_COUNT = "AccountUpdatesToAmbryServerErrorCount";
+  public static final String ACCOUNT_DELETES_TO_AMBRY_SERVER_ERROR_COUNT = "AccountDeletesToAmbryServerErrorCount";
+  public static final String ACCOUNT_FETCH_FROM_AMBRY_SERVER_ERROR_COUNT = "AccountFetchFromAmbryServerErrorCount";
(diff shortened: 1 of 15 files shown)


ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -130,6 +130,18 @@ default Container getContainer(String accountName, String containerName) throws
...
   }
 
+  /**
+   * Get an existing container from a given account.
+   * @param accountName the name of account which container belongs to.
+   * @param containerId the id of container to get.
+   * @return the requested {@link Container} object or null if not present.
+   * @throws AccountServiceException if exception occurs when getting container.
+   */
+  default Container getContainer(String accountName, Short containerId) throws AccountServiceException {
```
jsjtzyy, line 140: How about calling this `getContainerById`? And change line 128 to `getContainerByName`?

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -378,6 +378,44 @@ protected void updateResolvedContainers(Account account, Collection<Container> r
...
+  public Container getContainer(String accountName, String containerName) throws AccountServiceException {
+    Account account = getAccountByName(accountName);
+    if (account == null) {
+      return null;
+    }
+    Container container = account.getContainerByName(containerName);
+    if (container == null) {
+      // If container is not present in the account, refresh cache and retry
+      try {
+        fetchAndUpdateCache();
```
jsjtzyy, line 391: add log or metric here to track its behavior

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -378,6 +378,44 @@ protected void updateResolvedContainers(Account account, Collection<Container> r
...
+  public Container getContainer(String accountName, Short containerId) throws AccountServiceException {
+    Account account = getAccountByName(accountName);
+    if (account == null) {
+      return null;
+    }
+    Container container = account.getContainerById(containerId);
+    if (container == null) {
+      // If container is not present in the account, refresh cache and retry
+      try {
+        fetchAndUpdateCache();
```
jsjtzyy, line 410: same here

lightningrob, line 410: I was thinking we could query the individual container record.  A bit lighter weight, and also a guard against any issues that might crop up with the sync-latest query (isolation, corrupted timestamp).

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
lightningrob, line 140: I think we can use accountId instead of name here.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -378,6 +378,44 @@ protected void updateResolvedContainers(Account account, Collection<Container> r
...
+    Account account = getAccountByName(accountName);
+    if (account == null) {
+      return null;
+    }
+    Container container = account.getContainerByName(containerName);
+    if (container == null) {
+      // If container is not present in the account, refresh cache and retry
+      try {
+        fetchAndUpdateCache();
+        container = getAccountByName(accountName).getContainerByName(containerName);
```
lightningrob, line 392: If container is still null here, I would cache the account-container combo in a LRU cache, and check membership in that cache before querying the DB.

lightningrob (changes requested):

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -372,12 +385,87 @@ protected void updateResolvedContainers(Account account, Collection<Container> r
...
+   * @param accountName the name of account which container belongs to.
+   * @param containerName the name of container to get.
+   * @return {@link Container} if found in cache or mysql db. Else, returns {@code null}.
+   * @throws AccountServiceException
+   */
+  @Override
+  public Container getContainerByName(String accountName, String containerName) throws AccountServiceException {
+    Container container = null;
+
+    if (recentNotFoundContainersCache.contains(accountName + SEPARATOR + containerName)) {
```
jsjtzyy, line 403: Imagine a case where customers attempt to write into a non-existent container and get an error. They immediately create a new container (with same account/container name). If the request hits the same frontend, do we still return null?

jsjtzyy (approved):
LGTM after a minor question is answered.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
lightningrob: That's a good point.  I think every time we add a new container to the cache (create request of DB sync), we need to remove it (if needed) from the not-found cache.

ambry-account/src/main/java/com/github/ambry/account/AccountServiceMetrics.java
```diff
@@ -102,6 +103,8 @@ public AccountServiceMetrics(MetricRegistry metricRegistry) {
         metricRegistry.counter(MetricRegistry.name(HelixAccountService.class, "AccountFetchFromAmbryServerErrorCount"));
     getAccountInconsistencyCount =
         metricRegistry.counter(MetricRegistry.name(CompositeAccountService.class, "GetAccountInconsistencyCount"));
+    onDemandContainerFetchCount =
+        metricRegistry.counter(MetricRegistry.name(MySqlAccountService.class, "onDemandContainerFetchCount"));
```
lightningrob, line 107: Start name upper case for consistency.
Would you mind creating string constants for these names (I know most were already there).  It makes it safer for closed source to reference and avoid typos.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -53,6 +54,10 @@
 public class MySqlAccountService extends AbstractAccountService {
 
   private static final Logger logger = LoggerFactory.getLogger(MySqlAccountService.class);
+  private static final String SEPARATOR = "_";
+  private static final int cacheInitialCapacity = 100;
```
lightningrob, line 58: Minor: add comment that these are for recentNotFoundContainersCache, since we now have multiple caches.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -481,6 +569,39 @@ private void updateContainersWithMySqlStore(Account account, Collection<Containe
...
+    infoMapLock.writeLock().lock();
+    try {
+      accountInfoMapRef.get().addOrUpdateContainers(containers);
+    } finally {
+      infoMapLock.writeLock().unlock();
+    }
+
+    // Remove containers from not-found LRU cache.
+    for (Container container : containers) {
+      recentNotFoundContainersCache.remove(
```
lightningrob, line 600: This negates my earlier comment in response to Yingyi's question.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -481,6 +569,39 @@ private void updateContainersWithMySqlStore(Account account, Collection<Containe
...
+    } finally {
+      infoMapLock.writeLock().unlock();
+    }
+  }
+
+  /**
+   * Updates {@link Container}s to in-memory cache.
+   * @param containers added or modified {@link Container}s
+   */
+  private void updateContainersInCache(Collection<Container> containers) {
```
lightningrob, line 589: I like this refactoring.  Too much duplicate code before.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -372,12 +385,87 @@ protected void updateResolvedContainers(Account account, Collection<Container> r
...
+  public Container getContainerByName(String accountName, String containerName) throws AccountServiceException {
+    Container container = null;
+
+    if (recentNotFoundContainersCache.contains(accountName + SEPARATOR + containerName)) {
+      // If container was not found in recent get attempts, avoid another query to db and return null.
+      return null;
+    }
+
+    Account account = getAccountByName(accountName);
+    if (account != null) {
```
lightningrob, line 409: Minor: we can bail out here (return null) if account is null.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -372,12 +385,87 @@ protected void updateResolvedContainers(Account account, Collection<Container> r
...
+   * @param accountId the name of account which container belongs to.
+   * @param containerId the id of container to get.
+   * @return {@link Container} if found in cache or mysql db. Else, returns {@code null}.
+   * @throws AccountServiceException
+   */
+  @Override
+  public Container getContainerById(short accountId, Short containerId) throws AccountServiceException {
+    Container container = null;
+    Account account = getAccountById(accountId);
+    if (account != null) {
```
lightningrob, line 446: Here too (return)

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -53,6 +54,10 @@
 public class MySqlAccountService extends AbstractAccountService {
 
   private static final Logger logger = LoggerFactory.getLogger(MySqlAccountService.class);
+  private static final String SEPARATOR = "_";
```
lightningrob, line 57: Can we change this to ":" ?  That char is less commonly (or never) used in account or container names

lightningrob (commented):
Logic looks good.  Approved with minor changes.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
jsjtzyy: I guess I was trying to say in the case I mentioned if container name is in `recentNotFoundContainersCache` it requires periodic refresh to update cache (remove this container name out of "NotFound" cache). On-demand fetch in `getContainerByName` won't be performed. 
Anyway, I think it's fine to rely on periodic refresh (and probably it's the only way we can do).

lightningrob (approved):

PR #1766, opened 2021-02-06
Close mysql connection after every account service sync

lightningrob:
Allow only read-only account service endpoints


Arun-LinkedIn (approved):

justinlin-linkedin (approved):
LGTM

PR #1768, opened 2021-02-08
Send multiple frames if content size greater than http2 max frame size

zzmao:
It was assumed we can send out all content in one frame(5MB), but replication data size maybe 100MB.
This change is to allow multiple frames if content size greater than http2 max frame size.

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
@@ -37,14 +37,16 @@ import org.slf4j.LoggerFactory;
 public class AmbrySendToHttp2Adaptor extends ChannelOutboundHandlerAdapter {
   private static final Logger logger = LoggerFactory.getLogger(AmbrySendToHttp2Adaptor.class);
   private final boolean forServer;
-  private final int maxFrameSize = 5000;
+  private final int maxFrameSize;
 
   /**
    * @param forServer if true, the handler is used as server side outbound handler. Otherwise, it's use as client side
    *                  outbound handler.
+   * @param maxFrameSize
    */
-  public AmbrySendToHttp2Adaptor(boolean forServer) {
+  public AmbrySendToHttp2Adaptor(boolean forServer, int maxFrameSize) {
     this.forServer = forServer;
+    this.maxFrameSize = maxFrameSize;
   }
 
   /**
@@ -73,8 +75,6 @@ public class AmbrySendToHttp2Adaptor extends ChannelOutboundHandlerAdapter {
     DefaultHttp2HeadersFrame headersFrame = new DefaultHttp2HeadersFrame(http2Headers, false);
     ctx.write(headersFrame);
 
-    int frameNumber =
-        send.content().readableBytes() / maxFrameSize + (send.content().readableBytes() % maxFrameSize == 0 ? 0 : 1);
     // Referencing counting for derived {@link ByteBuf}: https://netty.io/wiki/reference-counted-objects.html#derived-buffers
     try {
       while (send.content().readableBytes() > maxFrameSize) {
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
@@ -37,6 +37,7 @@ import org.slf4j.LoggerFactory;
 public class AmbrySendToHttp2Adaptor extends ChannelOutboundHandlerAdapter {
   private static final Logger logger = LoggerFactory.getLogger(AmbrySendToHttp2Adaptor.class);
   private final boolean forServer;
+  private final int maxFrameSize = 5000;
 
   /**
    * @param forServer if true, the handler is used as server side outbound handler. Otherwise, it's use as client side
@@ -71,9 +72,26 @@ public class AmbrySendToHttp2Adaptor extends ChannelOutboundHandlerAdapter {
     }
     DefaultHttp2HeadersFrame headersFrame = new DefaultHttp2HeadersFrame(http2Headers, false);
     ctx.write(headersFrame);
-    ByteBuf dataContent = send.content();
-    DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(dataContent, true);
-    // Caller should call writeAndFlush().
-    ctx.write(dataFrame, promise);
+
+    int frameNumber =
+        send.content().readableBytes() / maxFrameSize + (send.content().readableBytes() % maxFrameSize == 0 ? 0 : 1);
+    // Referencing counting for derived {@link ByteBuf}: https://netty.io/wiki/reference-counted-objects.html#derived-buffers
+    try {
+      while (send.content().readableBytes() > maxFrameSize) {
+        ByteBuf slice = send.content().readSlice(maxFrameSize);
+        slice.retain();
+        DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(slice, false);
+        ctx.write(dataFrame, promise);
+      }
+      // The last slice
+      ByteBuf slice = send.content().readSlice(send.content().readableBytes());
+      slice.retain();
+      DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(slice, true);
+      ctx.writeAndFlush(dataFrame, promise);
+    } catch (Exception e) {
+      logger.error("Error while processing frames. Channel: {}", ctx.channel(), e);
+    } finally {
+      send.content().release();
+    }
   }
 }
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
@@ -77,11 +77,11 @@ public class AmbrySendToHttp2Adaptor extends ChannelOutboundHandlerAdapter {
 
     // Referencing counting for derived {@link ByteBuf}: https://netty.io/wiki/reference-counted-objects.html#derived-buffers
     try {
-      while (send.content().readableBytes() > maxFrameSize) {
+      while (send.content().isReadable(maxFrameSize)) {
         ByteBuf slice = send.content().readSlice(maxFrameSize);
         slice.retain();
         DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(slice, false);
-        ctx.write(dataFrame, promise);
+        ctx.write(dataFrame);
       }
       // The last slice
       ByteBuf slice = send.content().readSlice(send.content().readableBytes());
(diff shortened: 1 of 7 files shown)


ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -71,9 +74,24 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
-    // Caller should call writeAndFlush().
-    ctx.write(dataFrame, promise);
+
+    // Referencing counting for derived {@link ByteBuf}: https://netty.io/wiki/reference-counted-objects.html#derived-buffers
+    try {
+      while (send.content().readableBytes() > maxFrameSize) {
+        ByteBuf slice = send.content().readSlice(maxFrameSize);
+        slice.retain();
+        DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(slice, false);
+        ctx.write(dataFrame, promise);
```
justinlin-linkedin, line 84: so the sliced bytebuf shares the same refCnt as send.content() and ctx.write would decrease the refCnt? Is that the reason why you have to retain the sliced bytebuf?

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -71,9 +74,24 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+      }
+      // The last slice
+      ByteBuf slice = send.content().readSlice(send.content().readableBytes());
+      slice.retain();
+      DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(slice, true);
+      ctx.writeAndFlush(dataFrame, promise);
+    } catch (Exception e) {
+      logger.error("Error while processing frames. Channel: {}", ctx.channel(), e);
+    } finally {
+      send.content().release();
```
justinlin-linkedin, line 94: And here since send.content() is never passed to ctx.write method, send.content()'s refCnt won't be decreased, is that the reason why you need to call release here?

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -71,9 +74,24 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+        ByteBuf slice = send.content().readSlice(maxFrameSize);
+        slice.retain();
+        DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(slice, false);
+        ctx.write(dataFrame, promise);
+      }
+      // The last slice
+      ByteBuf slice = send.content().readSlice(send.content().readableBytes());
+      slice.retain();
+      DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(slice, true);
+      ctx.writeAndFlush(dataFrame, promise);
```
justinlin-linkedin, line 90: I would not call writeAndFlush here, since we already call writeAndFlush outside of this handler. Before this PR, we only call ctx.write.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2BlockingChannelStreamChannelInitializer.java
```diff
@@ -25,22 +26,25 @@
...
-  private static final AmbrySendToHttp2Adaptor ambrySendToHttp2Adaptor = new AmbrySendToHttp2Adaptor(false);
+  private static AmbrySendToHttp2Adaptor ambrySendToHttp2Adaptor = null;
 
-  Http2BlockingChannelStreamChannelInitializer(int http2MaxContentLength) {
-    this.http2MaxContentLength = http2MaxContentLength;
+  Http2BlockingChannelStreamChannelInitializer(Http2ClientConfig http2ClientConfig) {
+    this.http2ClientConfig = http2ClientConfig;
+    if (ambrySendToHttp2Adaptor == null) {
+      ambrySendToHttp2Adaptor = new AmbrySendToHttp2Adaptor(false, http2ClientConfig.http2FrameMaxSize);
+    }
```
justinlin-linkedin, line 40: is this `if` statement necessary? i think we can just create `AmbrySendToHttp2Adaptor` without this if statement.

ambry-network/src/test/java/com/github/ambry/network/http2/AmbrySendToHttp2AdaptorTest.java
```diff
@@ -54,31 +55,44 @@ public void after() {
...
   public void testServerWrite() throws Exception {
-    EmbeddedChannel channel = newChannel();
-    // Test server writing the Send
-    Send send = new SendWithContent();
+    int maxFrameSize = 2001;
+    EmbeddedChannel channel = new EmbeddedChannel(new AmbrySendToHttp2Adaptor(true, maxFrameSize));
+    int contentSize = 7000;
+    byte[] byteArray = new byte[contentSize];
+    new Random().nextBytes(byteArray);
+    ByteBuf content = UnpooledByteBufAllocator.DEFAULT.heapBuffer(contentSize).writeBytes(byteArray);
```
justinlin-linkedin, line 63: using a pooled bytebuf seems more convincing wrt the reference counter here.

ambry-network/src/test/java/com/github/ambry/network/http2/AmbrySendToHttp2AdaptorTest.java
```diff
@@ -54,31 +55,44 @@ public void after() {
...
+    data.content().readBytes(resultArray, i * maxFrameSize, data.content().readableBytes());
     data.content().release();
-  }
-
-  /**
-   * @return a new {@link EmbeddedChannel} with {@link AmbrySendToHttp2Adaptor} added to the pipeline.
-   */
-  static EmbeddedChannel newChannel() {
-    return new EmbeddedChannel(new AmbrySendToHttp2Adaptor(true));
+    Assert.assertArrayEquals(byteArray, resultArray);
```
justinlin-linkedin, line 84: shouldn't you call `content.release()` after you finish comparison given that you already call content.retain() before?

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
zzmao: That's correct. See examples from netty doc:  https://netty.io/wiki/reference-counted-objects.html#derived-buffers

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
zzmao: That's correct. See examples from netty doc: https://netty.io/wiki/reference-counted-objects.html#derived-buffers

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
zzmao: changed it back to ctx.write()

ambry-network/src/main/java/com/github/ambry/network/http2/Http2BlockingChannelStreamChannelInitializer.java
zzmao: without the if, we need to new an AmbrySendToHttp2Adaptor object, but AmbrySendToHttp2Adaptor is stateless. It can shared.

ambry-network/src/test/java/com/github/ambry/network/http2/AmbrySendToHttp2AdaptorTest.java
zzmao: Didn't get you. Data was first read into the resultArray, then content is released immediately.

zzmao: Fixed. (Looks like only Pool ByteBuf needs release. )

zzmao:
@cgtz  this PR is ready.

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
```diff
@@ -71,9 +74,24 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
     DefaultHttp2HeadersFrame headersFrame = new DefaultHttp2HeadersFrame(http2Headers, false);
     ctx.write(headersFrame);
-    ByteBuf dataContent = send.content();
-    DefaultHttp2DataFrame dataFrame = new DefaultHttp2DataFrame(dataContent, true);
-    // Caller should call writeAndFlush().
-    ctx.write(dataFrame, promise);
+
+    // Referencing counting for derived {@link ByteBuf}: https://netty.io/wiki/reference-counted-objects.html#derived-buffers
+    try {
+      while (send.content().readableBytes() > maxFrameSize) {
```
cgtz, line 80: From the page linked above, we can do `while (send.content().isReadable(maxFrameSize))`

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
cgtz, line 84: it seems like we are sharing the same promise for every frame. Can that cause issues at other parts of the pipeline if called before all chunks have been written?

ambry-network/src/main/java/com/github/ambry/network/http2/Http2BlockingChannelStreamChannelInitializer.java
cgtz: Can we instead pass in the single `AmbrySendToHttp2Adaptor` to the constructor?

cgtz: It seems like we only have one instance of `Http2BlockingChannelStreamChannelInitializer` per `Http2BlockingChannelPool` anyways, so maybe you could keep this as a member variable instead of making it static?

zzmao: you are right. A member can solve this cleanly.

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
zzmao: good catch. removed it here. Only the last ctx.write() uses promise.

ambry-network/src/main/java/com/github/ambry/network/http2/AmbrySendToHttp2Adaptor.java
zzmao: fixed.

cgtz (approved):

justinlin-linkedin (approved):

PR #1769, opened 2021-02-08
Fix null pointer issue in vcr due to uninitialized values in AD storage client.

ankagrawal:


jsjtzyy (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ADAuthBasedStorageClient.java
```diff
@@ -46,9 +46,9 @@
  */
 public class ADAuthBasedStorageClient extends StorageClient {
   private static final String AD_AUTH_TOKEN_REFRESHER_PREFIX = "AdAuthTokenRefresher";
-  private final ScheduledExecutorService tokenRefreshScheduler =
+  private ScheduledExecutorService tokenRefreshScheduler =
       Utils.newScheduler(1, AD_AUTH_TOKEN_REFRESHER_PREFIX, false);
```
lightningrob, line 50: Are you sure you want to schedule the refresher here as well as in buildBlobServiceClient?  Won't you end up with two of them?

lightningrob (changes requested):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ADAuthBasedStorageClient.java
ankagrawal: fixed it.

lightningrob (approved):

PR #1774, opened 2021-02-10
Return request cost for each ambry request in response headers.

ankagrawal:


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -252,8 +252,10 @@ public void processResponse(RestRequest restRequest, RestResponseChannel respons
       }
     }
     if (quotaManager != null) {
-      quotaManager.charge(restRequest, blobInfo);
-      // TODO Use ThrottlingRecommendation returned from charge() to set appropriate response headers.
+      ThrottlingRecommendation throttlingRecommendation = quotaManager.charge(restRequest, blobInfo);
+      if (throttlingRecommendation != null) {
```
ankagrawal, line 256: @lightningrob As a followup from our discussion in previous PR, I feel that checking if requestQuotaThrottling is enabled is cleaner than checking for null throttlingRecommendation. Checking for null throttlingRecommendation makes us depend on the implementation of interface which is not clean. Let me know if you are ok to change
`if(quotaManager!=null)` to `if(quotaManager!=null && quotaManger.getQuotaConfig().isRequestThrottlingEnabled())`

lightningrob: Seems fine.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -281,6 +284,14 @@
     }
   }
 
+  public static final class QuotaHeaders {
+
+    /**
+     * Response header indicating cost incurred by the request against capacity unit and storage quotas.
+     */
+    public static final String REQUEST_COST = "x-ambry-request-cost";
```
lightningrob, line 292: It's not a separate cost for read and write?  I created them as separate in the avro schema based on that understanding.

ambry-api/src/test/java/com/github/ambry/rest/RestUtilsTest.java
```diff
@@ -768,30 +804,66 @@ public void setUserMetadataHeadersMapTest() throws UnsupportedEncodingException
...
-  // general.
+  /**
+   * Tests for {@link RestUtils#setRequestCostHeader} method.
+   */
+  @Test
+  public void setRequestCostHeaderTest() {
+    Map<QuotaName, Double> costMap = new HashMap<>();
+
+    // test for valid cost map
+    costMap.put(QuotaName.READ_CAPACITY_UNIT, 1.9);
```
lightningrob, line 815: Here the capacity unit is for read.  Why is the request cost not that way?

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
ankagrawal: The value of the header will contain the separate costs for read, write and storage. Something like below:
"READ_CAPACITY_UNIT=1.9; STORAGE_IN_GB=3.0; "

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -1011,6 +1033,20 @@ private static void encodeMetadataValueAndSetHeader(String headerName, String he
...
+  /**
+   * Encode a map of {@link String}s as http header value.
+   * @param map {@link Map} of {@link String}s to encode.
+   * @return encoded http header value.
+   */
+  static String encodeKVHeaderValue(Map<String, String> map) {
+    Objects.requireNonNull(map);
+    String value = "";
+    for(Map.Entry<String, String> entry : map.entrySet()) {
+      value += entry.getKey() + "=" + entry.getValue() + "; ";
```
lightningrob, line 1045: Semicolon is non-standard for key-value string format,  please use comma (and remove trailing one).
Eg.`"READ_CAPACITY_UNIT=1.9, STORAGE_IN_GB=3.0" `
I think there's a utility to print a map out this way but need to look it up.
That way, the header value can be parsed as a Property map.

lightningrob: Would you mind adding a public method to reverse parse the header value back to a Map<String,Double>?  That way I can just call it and not worry about the format?

lightningrob (changes requested):

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
ankagrawal: done.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -1022,4 +1046,53 @@ public static String stripSlashAndExtensionFromId(String blobIdWithExtension) {
...
+      Objects.requireNonNull(value);
+      Map<String, String> valueMap = new HashMap<>();
+      if (value.isEmpty()) {
+        return valueMap;
+      }
+      if (!value.contains(DELIM)) {
+        throw new IllegalArgumentException("Invalid kv header value: " + value);
+      }
+      for (String kvStr : value.split(DELIM)) {
+        if (!kvStr.contains(KV_SEPERATOR)) {
```
lightningrob, line 1086: If the string ends in DELIM, won't split-ing it result in an extra empty string at the end of the array?

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -1022,4 +1046,53 @@ public static String stripSlashAndExtensionFromId(String blobIdWithExtension) {
...
+      }
+      return value;
+    }
+
+    /**
+     * Decode a kv header value.
+     * @param value {@link String} containing the encoded kv values.
+     * @return decoded kv {@link Map}.
+     */
+    static Map<String, String> decodeKVHeaderValue(String value) {
```
lightningrob, line 1076: Can you make this class and method public so it can be called outside the package?

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -933,6 +944,19 @@ public static void setUserMetadataHeaders(Map<String, String> userMetadataMap,
     }
   }
 
+  /**
+   * Sets the request cost header in response.
+   * @param costMap {@link Map} of cost for each quota.
+   * @param restResponseChannel the {@link RestResponseChannel} that is used for sending the response.
+   */
+  public static void setRequestCostHeader(Map<QuotaName, Double> costMap, RestResponseChannel restResponseChannel) {
+    Objects.requireNonNull(costMap, "cost map cannot contain be null");
```
justinlin-linkedin, line 953: typo: cost map can't be null.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -1022,4 +1046,53 @@ public static String stripSlashAndExtensionFromId(String blobIdWithExtension) {
...
+     * Encode a map of {@link String}s as http header value.
+     * @param map {@link Map} of {@link String}s to encode.
+     * @return encoded http header value.
+     */
+    public static String encodeKVHeaderValue(Map<String, String> map) {
+      Objects.requireNonNull(map);
+      String value = "";
+      for (Map.Entry<String, String> entry : map.entrySet()) {
+        value += entry.getKey() + KV_SEPERATOR + entry.getValue() + DELIM;
+      }
```
justinlin-linkedin, line 1067: we could just use join method from collectors to do this
```
String  value = map.entrySet().stream().map(e -> e.getKey() + KV_SEPERATOR + e.getValue()).collect(Collectors.joining(DELIM));
```

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -1022,4 +1046,53 @@ public static String stripSlashAndExtensionFromId(String blobIdWithExtension) {
...
+     * @param value {@link String} containing the encoded kv values.
+     * @return decoded kv {@link Map}.
+     */
+    public static Map<String, String> decodeKVHeaderValue(String value) {
+      Objects.requireNonNull(value);
+      Map<String, String> valueMap = new HashMap<>();
+      if (value.isEmpty()) {
+        return valueMap;
+      }
+      if (!value.contains(DELIM)) {
```
justinlin-linkedin, line 1082: All in stream ?
```
      return Arrays.stream(value.split(DELIM)).map(kv -> kv.split(KV_SEPERATOR)).peek(kv -> {
        if (kv.length != 2) {
          throw new IllegalArgumentException("Invalue kv header value: " + kv);
        }
      }).collect(Collectors.toMap(kv -> kv[0], kv -> kv[1]));```

File: ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java (hunk 8 above)
Comment (lightningrob) [reply ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java:1067]: +1

File: ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java (hunk 9 above)
Comment (lightningrob) [reply ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java:1082]: I find that hard to read, even if it works.

File: ambry-api/src/test/java/com/github/ambry/rest/RestUtilsTest.java [hunk 10]
```diff
@@ -768,30 +804,100 @@ public void setUserMetadataHeadersMapTest() throws UnsupportedEncodingException
...
+      fail("Null value should throw error");
+    } catch (NullPointerException npEx) {
+    }
+
+    // test for valid string
+    String validEncodedHeader = "READ_CAPACITY_UNIT=1.9; STORAGE_IN_GB=3.0; ";
+    Map<String, String> decodedMap = new HashMap<>();
+    decodedMap.put("READ_CAPACITY_UNIT", "1.9");
+    decodedMap.put("STORAGE_IN_GB", "3.0");
+    assertTrue("Invalid encoded value",
```
lightningrob, line 884: Minor: assertEquals

lightningrob (approved):
Approved after addressing Justin's comments (except one)

lightningrob:
@ankagrawal Can you address the remaining comments so we can merge this PR?

PR #1777, opened 2021-02-17
Hide Stats Key prefix in Utils class

justinlin-linkedin:
In stats reports, we use string as the key for partition id, account id and container id, with some prefix. Eg. for account id 100, the account key is A[100]. The prefix is spread all over the codebase. This PR reuse the common methods in Utils class to replace those statements that directly implement the logic of converting ids to string keys and vice versa.

### DIFF
--- a/ambry-server/src/test/java/com/github/ambry/server/StatsManagerTest.java
+++ b/ambry-server/src/test/java/com/github/ambry/server/StatsManagerTest.java
@@ -659,7 +659,7 @@ public class StatsManagerTest {
       for (int j = 0; j < random.nextInt(MAX_CONTAINER_COUNT - MIN_CONTAINER_COUNT + 1) + MIN_CONTAINER_COUNT; j++) {
         long validSize = random.nextInt(2501) + 500;
         subTotalSize += validSize;
-        containerMap.put(Utils.statsContainerKey(j), new StatsSnapshot(validSize, null));
+        containerMap.put(Utils.statsContainerKey((short) j), new StatsSnapshot(validSize, null));
         accountContainerPairMap.put(Utils.partitionClassStatsAccountContainerKey((short) i, (short) j),
             new StatsSnapshot(validSize, null));
       }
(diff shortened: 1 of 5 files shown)


ambry-server/src/test/java/com/github/ambry/server/StatsManagerTest.java
```diff
@@ -654,19 +654,17 @@ public void testReplicaFromOfflineToDropped() throws Exception {
...
-      String accountIdStr = "A[".concat(String.valueOf(i)).concat("]");
       Map<String, StatsSnapshot> containerMap = new HashMap<>();
       long subTotalSize = 0;
       for (int j = 0; j < random.nextInt(MAX_CONTAINER_COUNT - MIN_CONTAINER_COUNT + 1) + MIN_CONTAINER_COUNT; j++) {
-        String containerIdStr = "C[".concat(String.valueOf(j)).concat("]");
         long validSize = random.nextInt(2501) + 500;
         subTotalSize += validSize;
-        containerMap.put(containerIdStr, new StatsSnapshot(validSize, null));
-        accountContainerPairMap.put(accountIdStr + Utils.ACCOUNT_CONTAINER_SEPARATOR + containerIdStr,
+        containerMap.put(Utils.statsContainerKey((short) j), new StatsSnapshot(validSize, null));
```
jsjtzyy, line 662: Thanks for making this change. One minor suggestion: I see there are lots of places where `int`  is casted to `short`. Can we add same method something like this:
```
  public static String statsContainerKey(int containerId) {
    return statsContainerKey((short) containerId);
  }
```

ambry-server/src/test/java/com/github/ambry/server/StatsManagerTest.java
```diff
@@ -654,19 +654,17 @@ public void testReplicaFromOfflineToDropped() throws Exception {
...
+        accountContainerPairMap.put(Utils.partitionClassStatsAccountContainerKey((short) i, (short) j),
             new StatsSnapshot(validSize, null));
       }
       totalSize += subTotalSize;
-      accountMap.put(accountIdStr, new StatsSnapshot(subTotalSize, containerMap));
+      accountMap.put(Utils.statsAccountKey((short) i), new StatsSnapshot(subTotalSize, containerMap));
```
jsjtzyy, line 667: same here

jsjtzyy (approved):

ambry-server/src/test/java/com/github/ambry/server/StatsManagerTest.java
justinlin-linkedin: I think it's better not to do this, container id and account id should stay short. This casting is because in test cases, we use int to iterate from 0 to some number.

cgtz (approved):

PR #1778, opened 2021-02-17
[StorageQuota] Fix NPE in storeAccountStats

justinlin-linkedin:
There are cases that partition stats snapshot has empty subMap because all of the account and containers under this partition has 0 storage usage values. In such case, storeAccountStats method would throw a NPE. This PR fixes that.

### DIFF
--- a/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
@@ -193,7 +193,7 @@ public class AccountStatsMySqlStore implements AccountStatsStore {
     // 1. If a container storage usage exists in both StatsSnapshot, and the values are different.
     // 2. If a container storage usage only exists in first StatsSnapshot.
     // If a container storage usage only exists in the second StatsSnapshot, then it will not be applied to the given function.
-    // TODO: should delete rows in datbase when the previous statsSnapshot has more data than current one.
+    // TODO: should delete rows in database when the previous statsSnapshot has more data than current one.
     Map<String, StatsSnapshot> currPartitionMap = statsWrapper.getSnapshot().getSubMap();
     Map<String, StatsSnapshot> prevPartitionMap = prevSnapshot.getSubMap();
     for (Map.Entry<String, StatsSnapshot> currPartitionMapEntry : currPartitionMap.entrySet()) {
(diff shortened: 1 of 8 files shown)


ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -200,16 +202,22 @@ public synchronized void storeAccountStats(StatsWrapper statsWrapper) throws SQL
       StatsSnapshot prevAccountStatsSnapshot =
           prevPartitionMap.getOrDefault(partitionIdKey, new StatsSnapshot((long) 0, new HashMap<>()));
       short partitionId = Utils.partitionIdFromStatsPartitionKey(partitionIdKey);
```
jsjtzyy, line 204: minor: partition id is actually `long` type, see `AmbryPartition`, maybe make them consistent ? (I see both `statsPartitionKey` and `partitionIdFromStatsPartitionKey` are using short type)

jsjtzyy (approved):
LGTM after minor comments are addressed

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -192,6 +193,7 @@ public synchronized void storeAccountStats(StatsWrapper statsWrapper) throws SQL
     // 1. If a container storage usage exists in both StatsSnapshot, and the values are different.
     // 2. If a container storage usage only exists in first StatsSnapshot.
     // If a container storage usage only exists in the second StatsSnapshot, then it will not be applied to the given function.
+    // TODO: should delete rows in datbase when the previous statsSnapshot has more data than current one.
```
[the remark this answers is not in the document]
justinlin-linkedin: it's container

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
justinlin-linkedin: Change it to int.

jsjtzyy (approved):

lightningrob (approved):
Was changing short to int related to the NPE somehow?  It seems fine but I'm not sure why you did it.

justinlin-linkedin:
> Was changing short to int related to the NPE somehow? It seems fine but I'm not sure why you did it.

The type changing for partition id has nothing to do with NPE. Just that Yingyi pointed out that partition id in AmbryPartition class is a long type, not a short type. However, we already set the partition id in mysql database schema as "INT", that's why in this PR I changed it to int.

PR #1779, opened 2021-02-18
Fix account property didn't get updated in AccountInfoMap

jsjtzyy:
We introduced new property "aclInheritedByContainer" to Account but didn't correctly update it within AccountInfoMap. This caused in-mem cache always has initial default value for this property and never gets a chance to update. This PR should fix issue in account update method.


lightningrob (approved):
Good catch, thanks for fixing.

PR #1780, opened 2021-02-19
[StatsMigration] Create PartitionClassStatsPublisher in StatsManager

justinlin-linkedin:
Adding PartitionClassStatsPublisher in StatsManager.
Adding integration test for StatsManager to publish stats to mysql database.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
@@ -29,7 +29,6 @@ public class StatsManagerConfig {
   public static final String STATS_ENABLE_MYSQL_REPORT = "stats.enable.mysql.report";
   public static final String STATS_HEALTH_REPORT_EXCLUDE_ACCOUNT_NAMES = "stats.health.report.exclude.account.names";
   public static final String STATS_PUBLISH_EXCLUDE_ACCOUNT_NAMES = "stats.publish.exclude.account.names";
-  public static final String STATS_ENABLE_PARTITION_CLASS_MYSQL_REPORT = "stats.enable.partition.class.mysql.report";
   public static final String STATS_PUBLISH_PARTITION_CLASS_REPORT_PERIOD_IN_SECS =
       "stats.publish.partition.class.report.period.in.secs";
 
@@ -78,18 +77,11 @@ public class StatsManagerConfig {
   public final List<String> publishExcludeAccountNames;
 
   /**
-   * True to enable publishing partition class report stats to mysql. You have to enable mysql report before you enble
-   * this configuration.
-   */
-  @Config(STATS_ENABLE_PARTITION_CLASS_MYSQL_REPORT)
-  @Default("false")
-  public final boolean enablePartitionClassMysqlReport;
-
-  /**
-   * The time period in seconds that configures how often partition class stats are published to mysql.
+   * The time period in seconds that configures how often partition class stats are published to mysql. Set it to 0 to
+   * disable publishing partition class reports to mysql database.
    */
   @Config(STATS_PUBLISH_PARTITION_CLASS_REPORT_PERIOD_IN_SECS)
-  @Default("518400")
+  @Default("0")
   public final long publishPartitionClassReportPeriodInSecs;
 
   public StatsManagerConfig(VerifiableProperties verifiableProperties) {
@@ -104,13 +96,11 @@ public class StatsManagerConfig {
     excludeNames = verifiableProperties.getString(STATS_PUBLISH_EXCLUDE_ACCOUNT_NAMES, "").trim();
     publishExcludeAccountNames =
         excludeNames.isEmpty() ? Collections.EMPTY_LIST : Arrays.asList(excludeNames.split(","));
-    enablePartitionClassMysqlReport = verifiableProperties.getBoolean(STATS_ENABLE_PARTITION_CLASS_MYSQL_REPORT, false);
     publishPartitionClassReportPeriodInSecs =
-        verifiableProperties.getLongInRange(STATS_PUBLISH_PARTITION_CLASS_REPORT_PERIOD_IN_SECS, 60 * 60 * 24 * 6, 0,
-            Long.MAX_VALUE);
-    if (enablePartitionClassMysqlReport && !enableMysqlReport) {
+        verifiableProperties.getLongInRange(STATS_PUBLISH_PARTITION_CLASS_REPORT_PERIOD_IN_SECS, 0, 0, Long.MAX_VALUE);
+    if (publishPartitionClassReportPeriodInSecs != 0 && !enableMysqlReport) {
       throw new IllegalStateException(
-          "Bad configuration, you have to enableMysqlReport if you enablePartitionClassMysqlReport");
+          "Bad configuration, you have to enableMysqlReport if you set a non-zero value for publishPartitionClassReportPeriodInSecs");
     }
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
@@ -29,6 +29,9 @@ public class StatsManagerConfig {
   public static final String STATS_ENABLE_MYSQL_REPORT = "stats.enable.mysql.report";
   public static final String STATS_HEALTH_REPORT_EXCLUDE_ACCOUNT_NAMES = "stats.health.report.exclude.account.names";
   public static final String STATS_PUBLISH_EXCLUDE_ACCOUNT_NAMES = "stats.publish.exclude.account.names";
+  public static final String STATS_ENABLE_PARTITION_CLASS_MYSQL_REPORT = "stats.enable.partition.class.mysql.report";
+  public static final String STATS_PUBLISH_PARTITION_CLASS_REPORT_PERIOD_IN_SECS =
+      "stats.publish.partition.class.report.period.in.secs";
 
   /**
    * The file path (including filename) to be used for publishing the stats.
@@ -74,6 +77,21 @@ public class StatsManagerConfig {
   @Default("")
   public final List<String> publishExcludeAccountNames;
 
+  /**
+   * True to enable publishing partition class report stats to mysql. You have to enable mysql report before you enble
+   * this configuration.
+   */
+  @Config(STATS_ENABLE_PARTITION_CLASS_MYSQL_REPORT)
+  @Default("false")
+  public final boolean enablePartitionClassMysqlReport;
+
+  /**
+   * The time period in seconds that configures how often partition class stats are published to mysql.
+   */
+  @Config(STATS_PUBLISH_PARTITION_CLASS_REPORT_PERIOD_IN_SECS)
+  @Default("518400")
+  public final long publishPartitionClassReportPeriodInSecs;
+
   public StatsManagerConfig(VerifiableProperties verifiableProperties) {
     outputFilePath = verifiableProperties.getString(STATS_OUTPUT_FILE_PATH, "/tmp/stats_output.json");
     publishPeriodInSecs = verifiableProperties.getLongInRange(STATS_PUBLISH_PERIOD_IN_SECS, 7200, 0, Long.MAX_VALUE);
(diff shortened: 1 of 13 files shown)


ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -474,16 +521,67 @@ public void run() {
...
+      }
+    }
+
+    void cancel() {
+      cancelled = true;
+    }
+  }
+
+  /**
+   * Runnable calss that collects, publishes partition class stats to mysql database.
```
jsjtzyy, line 544: typo: `class`

jsjtzyy (approved):
lgtm after minor comments are addressed.

ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
```diff
@@ -74,6 +77,21 @@
...
+   * this configuration.
+   */
+  @Config(STATS_ENABLE_PARTITION_CLASS_MYSQL_REPORT)
+  @Default("false")
+  public final boolean enablePartitionClassMysqlReport;
+
+  /**
+   * The time period in seconds that configures how often partition class stats are published to mysql.
+   */
+  @Config(STATS_PUBLISH_PARTITION_CLASS_REPORT_PERIOD_IN_SECS)
```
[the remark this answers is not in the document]
justinlin-linkedin: sure.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -129,18 +131,31 @@
...
-    scheduler.scheduleAtFixedRate(statsAggregator, actualDelay, publishPeriodInSecs, TimeUnit.SECONDS);
+    accountsStatsPublisher = new AccountStatsPublisher(accountStatsMySqlStore);
+    int actualDelay = config.initialDelayUpperBoundInSecs > 0 ? ThreadLocalRandom.current()
+        .nextInt(config.initialDelayUpperBoundInSecs) : 0;
+    logger.info("Scheduling account stats publishing job with an initial delay of {} secs", actualDelay);
+    scheduler.scheduleAtFixedRate(accountsStatsPublisher, actualDelay, config.publishPeriodInSecs, TimeUnit.SECONDS);
+
+    if (config.enablePartitionClassMysqlReport) {
+      partitionClassStatsPublisher = new PartitionClassStatsPublisher(accountStatsMySqlStore);
+      long initalDelay = ThreadLocalRandom.current().nextLong(config.publishPartitionClassReportPeriodInSecs / 2)
```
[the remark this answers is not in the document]
justinlin-linkedin: the logic is that we always wait for at least half of the interval. If the interval is 1 hour, then at least we will wait for 30 minutes.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -209,6 +224,37 @@ void collectAndAggregate(StatsSnapshot aggregatedSnapshot, PartitionId partition
...
+    Store store = storageManager.getStore(partitionId, false);
+    if (store == null) {
+      unreachablePartitions.add(partitionId);
+    } else {
+      try {
+        long fetchAndAggregatePerStoreStartTimeMs = time.milliseconds();
+        StoreStats storeStats = store.getStoreStats();
+        Map<StatsReportType, StatsSnapshot> snapshotsByType =
+            storeStats.getStatsSnapshots(EnumSet.of(StatsReportType.PARTITION_CLASS_REPORT), time.milliseconds(),
+                publishExcludeAccountIds);
```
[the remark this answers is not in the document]
justinlin-linkedin: we can do that with current dataset. In aggregated account reports, we have account level data as well.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -435,53 +478,108 @@ long getPermanentDeleteTombstoneSize() {
...
-          StatsWrapper statsWrapper = new StatsWrapper(statsHeader, aggregatedSnapshot);
+      Object lock = accountStatsMySqlStore != null ? accountStatsMySqlStore : this;
+      synchronized (lock) {
+        logger.info("Aggregating account stats for local report");
+        try {
+          resetDeleteTombstoneStats();
+          long totalFetchAndAggregateStartTimeMs = time.milliseconds();
+          StatsSnapshot aggregatedSnapshot = new StatsSnapshot(0L, new HashMap<>());
+          List<PartitionId> unreachablePartitions = new ArrayList<>();
+          Iterator<PartitionId> iterator = (new HashSet<>(partitionToReplicaMap.keySet())).iterator();
```
lightningrob, line 501: If partitionToReplicaMap is never modified after the constructor, why do you need to copy the key set before iterating?

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -435,53 +478,108 @@ long getPermanentDeleteTombstoneSize() {
...
-        updateAggregatedDeleteTombstoneStats();
-        metrics.initDeleteStatsGaugesIfNeeded();
-        if (!cancelled) {
-          metrics.totalFetchAndAggregateTimeMs.update(time.milliseconds() - totalFetchAndAggregateStartTimeMs);
-          StatsHeader statsHeader = new StatsHeader(StatsHeader.StatsDescription.STORED_DATA_SIZE, time.milliseconds(),
-              partitionToReplicaMap.keySet().size(), partitionToReplicaMap.keySet().size() - unreachableStores.size(),
-              unreachableStores);
-          // First write to account stats
-          StatsWrapper statsWrapper = new StatsWrapper(statsHeader, aggregatedSnapshot);
+      Object lock = accountStatsMySqlStore != null ? accountStatsMySqlStore : this;
```
lightningrob, line 493: Probably simpler to use synchronized(this) in both cases.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -435,53 +478,108 @@ long getPermanentDeleteTombstoneSize() {
...
+      synchronized (lock) {
+        logger.info("Aggregating account stats for local report");
+        try {
+          resetDeleteTombstoneStats();
```
lightningrob, line 497: Could you add a few more inline comments explaining what each of the steps does?  It's a lot of code with relatively few comments.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -435,53 +478,108 @@ long getPermanentDeleteTombstoneSize() {
...
+    private volatile boolean cancelled = false;
+    private final AccountStatsMySqlStore accountStatsMySqlStore;
+
+    PartitionClassStatsPublisher(AccountStatsMySqlStore accountStatsMySqlStore) {
+      this.accountStatsMySqlStore = accountStatsMySqlStore;
+    }
+
+    @Override
+    public void run() {
+      synchronized (accountStatsMySqlStore) {
```
lightningrob, line 555: Why can we assume that accountStatsMySqlStore is non-null here but not in the other task?
Also, do the actions of PartitionClassStatsPublisher and AccountStatsPublisher need to be serialized with each other (using same lock)?

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -435,53 +478,108 @@ long getPermanentDeleteTombstoneSize() {
...
+        logger.info("Aggregating partition class stats for local report");
+        try {
+          long totalFetchAndAggregateStartTimeMs = time.milliseconds();
+          StatsSnapshot aggregatedSnapshot = new StatsSnapshot(0L, new HashMap<>());
+          List<PartitionId> unreachablePartitions = new ArrayList<>();
+          Iterator<PartitionId> iterator = (new HashSet<>(partitionToReplicaMap.keySet())).iterator();
```
lightningrob, line 561: Same question about partitionToReplicaMap.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -435,53 +478,108 @@ long getPermanentDeleteTombstoneSize() {
...
+            metrics.totalFetchAndAggregateTimeMs.update(time.milliseconds() - totalFetchAndAggregateStartTimeMs);
+            StatsHeader statsHeader =
+                new StatsHeader(StatsHeader.StatsDescription.STORED_DATA_SIZE, time.milliseconds(),
+                    partitionToReplicaMap.keySet().size(),
+                    partitionToReplicaMap.keySet().size() - unreachableStores.size(), unreachableStores);
+            StatsWrapper statsWrapper = new StatsWrapper(statsHeader, aggregatedSnapshot);
+            accountStatsMySqlStore.storePartitionClassStats(statsWrapper);
+            logger.info("Local partition class stats snapshot published to {}", statsOutputFile.getAbsolutePath());
+          }
+        } catch (Exception | Error e) {
```
lightningrob, line 579: Throwable is equivalent.  But if we get an Error like OOM, do we really want to eat it?

ambry-server/src/integration-test/java/com/github/ambry/server/StatsManagerIntegrationTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+  /**
+   * Test partition class stats publisher
+   * @throws Exception
+   */
+  @Test
+  public void testPartitionClassStatsPublisher() throws Exception {
+    // Before publishing partition class stats, we have to publish account stats since they use the same data
+    StatsManager.AccountStatsPublisher accountStatsPublisher =
+        statsManager.new AccountStatsPublisher(accountStatsMySqlStore);
```
lightningrob, line 137: Should also test this will null store since that is an option.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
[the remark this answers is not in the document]
lightningrob: Please avoid mention of LI things in github reviews.

lightningrob (commented):
Logic looks good.  A few comments mainly about concurrency.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: updated.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: I want to synchronize on accountStatsMysqlStore since accountStatsMysqlStore only has one connection and multiple threads accessing this connection is not safe. That's why every time I want to use accountStatsMysqlStore, I would have to lock this object.
There are other threads trying to use accountStatsMysqlStore, one of them is MysqlReportAggregatorTask. This is a aggregation task that read all the stats reports and aggregate them and then save the aggregated result back to mysql database. By synchronizing on accountStatsMysqlStore,  we can be sure that there is no concurrency issue with accessing the single connection in accountStatsMysqlStore.
However, the accountStatsMysqlStore might be null, it's not yet enabled everywhere.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: updated.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: This is guaranteed, since we have to enable mysql report before we can enable the partition class reports. We have a check in StatsManagerConfig.java.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: updated.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: Will ignore Error.

ambry-server/src/integration-test/java/com/github/ambry/server/StatsManagerIntegrationTest.java
justinlin-linkedin: This integration test is mainly for testing mysql database. In unit test, we consider the mysql database doesn't exist by default. So we have to do the test for mysql database here. The case when accountStatsMysqlStore is null, has already been tested in unit tests.

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
lightningrob: Okay, makes sense.

lightningrob (approved):
LGTM

Issue #1742, opened 2021-01-07
Cleanup ReplicationEngine:createThreadPool to avoid hardcoding Vcr


lightningrob:
The method that creates replication threads hardcodes the prefix "Vcr" in the thread name when the startThread argument has value of true.  While it currently happens to be the case that the parameter is only set for Vcr, it is dangerous to make that assumption and makes the code non-intuitive to read.
 
`  private List<ReplicaThread> createThreadPool(String datacenter, int numberOfThreads, boolean startThread) {
      ...
      String threadIdentity =
          (startThread ? "Vcr" : "") + "ReplicaThread-" + (dataNodeId.getDatacenterName().equals(datacenter) ? "Intra-"
              : "Inter-") + i + "-" + datacenter;
`

PR #1781, opened 2021-02-20
Cleanup ReplicationEngine:createThreadPool to avoid hardcoding Vcr

parasgithub:
Sending a short pull request to make sure change is OK, will add a test afterwards (in same branch) if needed.

Issue #1742

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
@@ -35,6 +35,7 @@ import com.github.ambry.store.Store;
 import com.github.ambry.store.StoreKeyConverterFactory;
 import com.github.ambry.store.StoreKeyFactory;
 import com.github.ambry.utils.SystemTime;
+
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -269,6 +270,11 @@ public class VcrReplicationManager extends ReplicationEngine {
   public long getRemoteReplicaLagFromLocalInBytes(PartitionId partitionId, String hostName, String replicaPath) {
     // TODO get replica lag from cosmos?
     return -1;
+  }
+
+  @Override
+  protected String getReplicaThreadName(String datacenterToReplicateFrom, int threadIndexWithinPool) {
+    return "Vcr" + super.getReplicaThreadName(datacenterToReplicateFrom, threadIndexWithinPool);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -370,6 +368,17 @@ private int getReplicaThreadIndexToUse(String datacenter) {
     return replicaThreads;
   }
 
+  /**
+   * Chooses a name for a new replica thread.
+   * @param datacenterToReplicateFrom The datacenter that we replicate from in this thread.
+   * @param threadIndexWithinPool The index of the thread within the thread pool.
+   * @return The name of the thread.
+   */
+  protected String getReplicaThreadName(String datacenterToReplicateFrom, int threadIndexWithinPool) {
```
lightningrob, line 377: If the only difference is the prefix, the method could be getReplicaThreadPrefix() with no arguments.

lightningrob (commented):
Thanks for contributing.  See  inline comment.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
parasgithub: Yep, agreed. Though I named it like this on purpose to
1) prevent the creation of very specialized functions
2) allow the subclass needs to add more details to the thread name if needed
What do you think? I'm also okay with changing it to your suggestion.

ankagrawal: This is a good change. Thanks @parasgithub for making this change. I don't disagree with your argument to allow the subclasses to add more details to the thread name if needed. 
However, I have a nitpick in that case. I think other than prefix, the rest of the logic is repeated code. Can you separate them out, so that it is ensured that thread naming convention doesn't change inadvertently in subclasses, unless explicitly desired.

parasgithub: Done, please take a look

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -271,6 +267,11 @@ public long getRemoteReplicaLagFromLocalInBytes(PartitionId partitionId, String
     return -1;
   }
 
+  @Override
+  protected String getReplicaThreadName(String datacenterToReplicateFrom, int threadIndexWithinPool) {
+    return "Vcr" + ReplicaThread.getCanonicalReplicaThreadName(dataNodeId.getDatacenterName(), datacenterToReplicateFrom, threadIndexWithinPool);
```
lightningrob, line 272: This can be simply:
`return "Vcr" + super. getReplicaThreadName(datacenterToReplicateFrom, threadIndexWithinPool);`
And then I think the canonical logic can be collapsed back into the superclass method.

parasgithub: Sorry about the delay. Made the change.

lightningrob (approved):
Looks good.   Will merge once the build passes.

lightningrob:
@parasgithub Do you want to take a stab at fixing the failing test?

parasgithub:
> @parasgithub Do you want to take a stab at fixing the failing test?

Done

PR #1782, opened 2021-02-22
Add compaction metrics and storeCompactionThrottlerCheckIntervalMs

zzmao:
Small PR to add compaction metrics and storeCompactionThrottlerCheckIntervalMs

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -117,6 +117,13 @@ public class StoreConfig {
   public final int storeCompactionOperationsBytesPerSec;
 
   /**
+   * The check interval used in compaction rate throttler.
+   */
+  @Config("store.compaction.throttler.check.interval.ms")
+  @Default("-1")
+  public final int storeCompactionThrottlerCheckIntervalMs;
+
+  /**
    * Whether direct IO are to be enable or not for compaction.
    * This is only supported on > Linux 2.6
    */
@@ -393,6 +400,8 @@ public class StoreConfig {
     storeCompactionOperationsBytesPerSec =
         verifiableProperties.getIntInRange("store.compaction.operations.bytes.per.sec", 1 * 1024 * 1024, 1,
             Integer.MAX_VALUE);
+    storeCompactionThrottlerCheckIntervalMs =
+        verifiableProperties.getIntInRange("store.compaction.throttler.check.interval.ms", -1, 1, Integer.MAX_VALUE);
     storeCompactionEnableDirectIO = verifiableProperties.getBoolean("store.compaction.enable.direct.io", false);
     storeCompactionPurgeExpiredDeleteTombstone =
         verifiableProperties.getBoolean("store.compaction.purge.expired.delete.tombstone", false);
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -117,7 +117,7 @@ public class StoreConfig {
   public final int storeCompactionOperationsBytesPerSec;
 
   /**
-   * The check interval used in compaction rate throttler.
+   * The check interval used in compaction rate throttler. -1 means we check rate in every call.
    */
   @Config("store.compaction.throttler.check.interval.ms")
   @Default("-1")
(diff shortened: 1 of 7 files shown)


ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -116,6 +116,13 @@
   @Default("1*1024*1024")
   public final int storeCompactionOperationsBytesPerSec;
 
+  /**
+   * The check interval used in compaction rate throttler.
+   */
+  @Config("store.compaction.throttler.check.interval.ms")
+  @Default("-1")
```
jsjtzyy, line 123: Please add some comment describing what does -1 mean. Looks like if it's <0,  rate is checked on every call.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1212,7 +1212,7 @@ void compact(CompactionDetails details, byte[] bundleReadBuffer) throws IOExcept
     checkStarted();
     compactor.compact(details, bundleReadBuffer);
     checkCapacityAndUpdateReplicaStatusDelegate();
-    logger.trace("One cycle of compaction is completed on the store {}", storeId);
+    logger.info("One cycle of compaction is completed on the store {}", storeId);
```
jsjtzyy, line 1215: Maybe we can use `DEBUG` level?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -551,17 +552,23 @@ private boolean isIndexSegmentUnderCopy(Offset indexSegmentStartOffset) {
   private boolean copyDataByLogSegment(LogSegment logSegmentToCopy, FileSpan duplicateSearchSpan)
       throws IOException, StoreException {
     logger.info("Copying data from {}", logSegmentToCopy);
+    long logSegmentStartTime = SystemTime.getInstance().milliseconds();
```
jsjtzyy, line 555: can we use `time.milliseconds()` ?

jsjtzyy (approved):
lgtm after minor comments are addressed

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: we should probably use System.currentMillis(); this is for metrics.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -494,7 +494,9 @@ boolean removeBlobStore(PartitionId id) {
   private Map<String, Throttler> getThrottlers(StoreConfig config, Time time) {
     Map<String, Throttler> throttlers = new HashMap<>();
     // compaction ops
-    Throttler compactionOpsThrottler = new Throttler(config.storeCompactionOperationsBytesPerSec, -1, true, time);
+    Throttler compactionOpsThrottler =
+        new Throttler(config.storeCompactionOperationsBytesPerSec, config.storeCompactionThrottlerCheckIntervalMs, true,
```
justinlin-linkedin, line 498: could you elaborate why adding a positive checkIntervalMs would change behavior of Throttler?

justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
jsjtzyy: The `time.miliseconds()` actually calls `System.currentMillis()` (`MockTime` isn't used in production code), so I think either is fine.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
zzmao: yes. You are right. Added in the comment.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
zzmao: This is changed to `info` deliberately. When I look the log, there is log recording the start of compaction cycle, but no log for the end of the compaction cycle. 
In addition, this is not a repeated printed log. It should be ok to be in info level.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
zzmao: changed to time.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
zzmao: with -1, the rated is check every call. 
We also have a similar metric, `compactionCopyRateInBytes`, which shows stats per 60 seconds. 
Currently, if we set `storeCompactionOperationsBytesPerSec` to 5MB, the `compactionCopyRateInBytes` may go up to 15MB/s.
I suspect the difference come from the `-1`.

PR #1786, opened 2021-02-26
Add life version to reset key in store find token

jsjtzyy:
Previously we introduced reset key and its type to help find key in current log (after compaction). As UNDELETE may change the version of record, we need life version to help uniquely identify the reset key in the log. This PR fixes this potential issue in find key logic and adds life version into token deserialization code. 

(Note: token serialization change will be in future PR.)


ankagrawal (approved):
lgtm!

justinlin-linkedin (approved):

PR #1787, opened 2021-03-01
Introduce PersistentIndex version 4 to incorporate reset key version in index segment

jsjtzyy:
Previous index segment only contains reset key and its type without version info. This PR ensures index segment also tracks reset key life version to help generate new store find token. To achieve that, Persistent Index version 4 is introduced to support serializing/deserializing new index segment format.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
@@ -112,7 +112,7 @@ class IndexSegment implements Iterable<IndexEntry> {
   private Offset prevSafeEndPoint = null;
   // reset key refers to the first StoreKey that is added to the index segment
   private Pair<StoreKey, PersistentIndex.IndexEntryType> resetKey = null;
-  private short resetKeyVersion = UNINITIALIZED_RESET_KEY_VERSION;
+  private short resetKeyLifeVersion = UNINITIALIZED_RESET_KEY_VERSION;
   private NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> index = null;
 
   /**
@@ -337,8 +337,13 @@ class IndexSegment implements Iterable<IndexEntry> {
   /**
    * @return life version of reset key.
    */
-  short getResetKeyVersion() {
-    return resetKeyVersion;
+  short getResetKeyLifeVersion() throws StoreException {
+    if (resetKey == null || resetKeyLifeVersion != UNINITIALIZED_RESET_KEY_VERSION) {
+      return resetKeyLifeVersion;
+    }
+    NavigableSet<IndexValue> indexValues = find(resetKey.getFirst());
+    // reset key should be present in current index segment, hence indexValues is guaranteed to be non-empty
+    return indexValues.first().getLifeVersion();
   }
 
   /**
@@ -567,7 +572,7 @@ class IndexSegment implements Iterable<IndexEntry> {
       }
       if (resetKey == null) {
         resetKey = new Pair<>(entry.getKey(), entry.getValue().getIndexValueType());
-        resetKeyVersion = entry.getValue().getLifeVersion();
+        resetKeyLifeVersion = entry.getValue().getLifeVersion();
       }
       numberOfItems.incrementAndGet();
       sizeWritten.addAndGet(entry.getKey().sizeInBytes() + entry.getValue().getBytes().capacity());
@@ -753,7 +758,7 @@ class IndexSegment implements Iterable<IndexEntry> {
         }
         if (getVersion() >= PersistentIndex.VERSION_4) {
           // write reset key life version
-          writer.writeShort(resetKeyVersion);
+          writer.writeShort(resetKeyLifeVersion);
         }
 
         byte[] maxPaddingBytes = null;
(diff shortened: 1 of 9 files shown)


ambry-store/src/main/java/com/github/ambry/store/IndexValue.java
```diff
@@ -77,7 +77,7 @@
           + ORIGINAL_MESSAGE_OFFSET_SIZE_IN_BYTES + OPERATION_TIME_SECS_SIZE_IN_BYTES + ACCOUNT_ID_SIZE_IN_BYTES
           + CONTAINER_ID_SIZE_IN_BYTES;
 
-  final static int INDEX_VALUE_SIZE_IN_BYTES_V3 = INDEX_VALUE_SIZE_IN_BYTES_V1_V2 + LIFE_VERSION_SIZE_IN_BYTES;
+  final static int INDEX_VALUE_SIZE_IN_BYTES_V3_V4 = INDEX_VALUE_SIZE_IN_BYTES_V1_V2 + LIFE_VERSION_SIZE_IN_BYTES;
```
cgtz, line 80: Does V3_V4 mean this size applies to both?

cgtz (commented):
How will resetKey life version be used?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -330,6 +334,13 @@ void setVersion(short version) {
     return resetKey;
   }
 
+  /**
+   * @return life version of reset key.
+   */
+  short getResetKeyVersion() {
```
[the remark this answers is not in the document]
jsjtzyy: sure

jsjtzyy:
> How will resetKey life version be used?

Reset key life version will be used when rebuilding find token (due to compaction). As new find token tracks `reset key`, `reset key type` and `reset key life version`, we need to feed in such info when constructing a new token.
Reset key life version will help uniquely identify an index entry (assuming there are several entries with same key and same type, i.e. multiple DELETES on same key).
Reset key life version is supposed to be added when we introduced UNDELETE and life version in V3  PersistentIndex but I guess we didn't really use reset key at that time.

ambry-store/src/main/java/com/github/ambry/store/IndexValue.java
jsjtzyy: Yes, it applies to both. Version 4 doesn't change `IndexValue` format but `IndexSegment` format only. It adds reset key life version at the beginning of index segment (together with reset key and reset key type) without touching index entries.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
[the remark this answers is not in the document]
jsjtzyy: I am actually planning to do  a lookup within this method if resetKeyLifeVersion is `UNINITIALIZED_RESET_KEY_VERSION`.  This is to allow PersistentIndex with V1, V2, V3 to rebuild token with reset key as well.  We have `resetKey` since V1, and we should be able to find the `IndexValue` associated with that key and therefore get the life version. 
V4 PersistentIndex can skip this lookup as it explicitly maintains reset key life version in the index segment.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -109,6 +112,7 @@
   private Offset prevSafeEndPoint = null;
   // reset key refers to the first StoreKey that is added to the index segment
   private Pair<StoreKey, PersistentIndex.IndexEntryType> resetKey = null;
+  private short resetKeyLifeVersion = UNINITIALIZED_RESET_KEY_VERSION;
```
justinlin-linkedin, line 115: I think it's better to create a class to represent the reset key info, Something like
```
public class ResetKeyInfo {
    private final StoreKey storeKey;
    private final IndexEntryType keyType;
    private final short lifeVersion;
}
```
The resetKey isn't used anywhere, we can assume that all three pieces of info will be required by the caller all together.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -330,6 +334,18 @@ void setVersion(short version) {
...
+   * @return life version of reset key.
+   */
+  short getResetKeyLifeVersion() throws StoreException {
+    if (resetKey == null || resetKeyLifeVersion != UNINITIALIZED_RESET_KEY_VERSION) {
+      return resetKeyLifeVersion;
+    }
+    NavigableSet<IndexValue> indexValues = find(resetKey.getFirst());
+    // reset key should be present in current index segment, hence indexValues is guaranteed to be non-empty
+    return indexValues.first().getLifeVersion();
+  }
```
justinlin-linkedin, line 347: With a new class, we only need to have one method to return the resetKey info.

justinlin-linkedin (commented):
just minor comments.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
jsjtzyy: yeah, it's a good way to wrap them up and would be more extensible if we add new fields in the future.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
jsjtzyy: I am leaning towards to return them separately (3 methods). It has some additional logic in those methods to 1. check whether resetKeyInfo == null, 2. find the correct life version of reset key if not set in V1~V3 index. Thus `ResetKeyInfo` is pure internal data structure, caller doesn't have to know it.

justinlin-linkedin (approved):

cgtz (approved):

Issue #1708, opened 2020-12-01
Wrong URL in README


liotev:
In the README file, under section "Research", the URL of the "Paper introducing Ambry"  (http://dprg.cs.uiuc.edu/docs/SIGMOD2016-a/ambry.pdf) is not correct: clicking on it shows a "Not Found" page.

jsjtzyy:
Hi @liotev  try this: https://assured-cloud-computing.illinois.edu/files/2014/03/Ambry-LinkedIns-Scalable-GeoDistributed-Object-Store.pdf

djKooks:
@jsjtzyy @liotev seems already known issue but haven't been updated for several month, so created a PR 🙏

PR #1788, opened 2021-03-01
Fix paper link in README file

djKooks:
Hello,
Seems link of paper in README file are not exists anymore, so updated.
It is related with #1708 .

Please check 🙏


jsjtzyy (approved):
Thank you for fixing this!

PR #1789, opened 2021-03-01
[StatsMigration] Add partition class stats aggregation task

justinlin-linkedin:
This PR
1. Add partition class aggregation tasks
2. Change query stats interface to include a port number so same hostname with different port number will be supported.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
@@ -199,19 +199,18 @@ public class MySqlReportAggregatorTask extends UserContentStore implements Task 
   }
 
   private Pair<String, Integer> getHostNameAndPort(String instanceName) {
+    String hostname = instanceName;
     int port = clusterMapConfig.clusterMapPort;
     int ind = instanceName.lastIndexOf("_");
-    if (ind == -1) {
-      return new Pair<>(instanceName, port);
-    } else {
+    if (ind != -1) {
       try {
         port = Short.valueOf(instanceName.substring(ind + 1));
+        hostname = instanceName.substring(0, ind);
       } catch (NumberFormatException e) {
-        // string after "_" is not a port number, then return instance name.
-        return new Pair<>(instanceName, port);
+        // String after "_" is not a port number, then the hostname should be the instanceName
       }
-      return new Pair<>(instanceName.substring(0, ind), port);
     }
+    return new Pair<>(hostname, port);
   }
 
   @Override
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
@@ -13,6 +13,8 @@
  */
 package com.github.ambry.clustermap;
 
+import com.codahale.metrics.Histogram;
+import com.codahale.metrics.MetricRegistry;
 import com.github.ambry.commons.Callback;
 import com.github.ambry.config.ClusterMapConfig;
 import com.github.ambry.server.AccountStatsStore;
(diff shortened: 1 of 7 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -119,18 +161,56 @@ public TaskResult run() {
...
+      statsWrappers.put(instanceName,
+          accountStatsStore.queryPartitionClassStatsByHost(pair.getFirst(), pair.getSecond(), partitionNameAndIds));
+    }
+    return statsWrappers;
+  }
+
+  private Pair<String, Integer> getHostNameAndPort(String instanceName) {
+    int port = clusterMapConfig.clusterMapPort;
     int ind = instanceName.lastIndexOf("_");
     if (ind == -1) {
```
cgtz, line 204: Just a style thing, but you could reduce the number of returns like this:
```
int port, ind
String hostName = instanceName;
if (ind != -1) {
  try {
    port = Short.parseShort(...);
    hostName = substring(...);
  } catch (NumberFormatException e) {
    // ignore
  }
}
return new Pair<>(hostName, port);
```

cgtz (approved):
Looks good!

ambry-mysql/src/integration-test/java/com/github/ambry/accountstats/AccountStatsMySqlStoreIntegrationTest.java
```diff
@@ -60,10 +60,13 @@
 public class AccountStatsMySqlStoreIntegrationTest {
   private static final String clusterName1 = "Ambry-test";
   private static final String clusterName2 = "Ambry-random";
+  // hostname1 and hostname2 are the same, but with different port numbers
   private static final String hostname1 = "ambry1.test.github.com";
-  private static final String hostname2 = "ambry2.test.github.com";
+  private static final String hostname2 = "ambry1.test.github.com";
```
jsjtzyy, line 65: you can remove `hostname2` then

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -62,34 +91,46 @@
...
-      AccountStatsStore accountStatsStore, Callback<StatsSnapshot> callback, ClusterMapConfig clusterMapConfig) {
+      AccountStatsStore accountStatsStore, Callback<StatsSnapshot> callback, ClusterMapConfig clusterMapConfig,
+      MetricRegistry registry) {
     this.manager = manager;
     clusterAggregator = new HelixClusterAggregator(relevantTimePeriodInMs);
     this.statsReportType = statsReportType;
     this.accountStatsStore = accountStatsStore;
     this.callback = callback;
     this.clusterMapConfig = clusterMapConfig;
+    this.metrics = new Metrics(registry);
```
jsjtzyy, line 105: Wouldn't this cause any problem? We have `AccountStats` and `PartitionClassStats` to collect and two aggregation tasks will be created. For each task, we register all 4 metrics (line 74 ~ 81). Would second aggregation task throw exception something like:  `***metric has been registered already` ?

jsjtzyy (approved):
LGTM after comments are addressed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
justinlin-linkedin: Yes, you are right about the two tasks. But the creating a histogram is "getOrAdd". So if it's also being added to the registry, then it will not be added again.

ambry-mysql/src/integration-test/java/com/github/ambry/accountstats/AccountStatsMySqlStoreIntegrationTest.java
justinlin-linkedin: I like to keep it so that we know the difference between these two instances.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
jsjtzyy: Got it, I have no concern now.

PR #1791, opened 2021-03-01
[named blobs] add container option, fix small bugs

cgtz:
- Add an option to designate whether the named blob API can be used for
      a container
- Fix how pages are serialized. Otherwise, all fields in the element
      type will be serialized instead of the limited serializer being used.
 - Feed the unconverted ID that the router understands to the TTL update
      call.
 - Send location header on put responses so that blob ID can be used as a
      fallback for clients that are migrating.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ build/
 logs/
 null/
 .DS_Store
+
+.remote*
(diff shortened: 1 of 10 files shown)


supermariogo (approved):
LGTM

zzmao (commented):
LGTM. 

(@supermariogo  is my account. Used wrong account. )

zzmao (approved):

SophieGuo410 (approved):
LGTM!

PR #1794, opened 2021-03-03
Convert raw calculation to TimeUnit

kination:
Hello,
This PR is to follow-up issue https://github.com/linkedin/ambry/issues/597.
(but its pretty old one, so not sure if it still needs to be updated)

Please review 🙏


ambry-utils/src/main/java/com/github/ambry/utils/Time.java
```diff
@@ -24,12 +25,10 @@
   /**
    * Some common constants
    */
-  public static final int NsPerUs = 1000;
-  public static final int UsPerMs = 1000;
-  public static final int MsPerSec = 1000;
-  public static final int NsPerMs = NsPerUs * UsPerMs;
-  public static final int NsPerSec = NsPerMs * MsPerSec;
-  public static final int SecsPerMin = 60;
+  public static final int MsPerSec = Math.toIntExact(TimeUnit.SECONDS.toMicros(1));
```
cgtz, line 28: Hi @djKooks, for this issue, my thought was that these constants could be removed from our codebase entirely and replaced with usages of `TimeUnit` inline. Doing so will help to avoid integer overflow issues and typos from having to chain a lot of unit conversions.

Would you be open to exploring the codebase and finding places where these constants are used?

djKooks: @cgtz oh, I see. 
1. So you mean to remove variables such as `MsPerSec`, `SecsPerMin` in entire project?
2. Will it be okay to convert all values to `Long` type?
3

cgtz: 1. Yep, that's what I mean.
2. I think in most cases, time should be a `long`. If there are edge cases where time is represented as an `int`, feel free to ask about it on this thread and I can comment about what would be best in those situations.

On another note, thank you for contributing to Ambry! Do you personally deploy ambry? We are trying to engage more with our open source community.

djKooks: @cgtz thanks for comment. I'll try to update it.

Not personally deploying it, but maybe could in future. I'm pretty interested in it. If there's any future plan or update, please let me know 🙏

djKooks:
@cgtz please check when you're free 🙏 
Also should remove `Time`, `SystemTime` class if it's fine.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -188,7 +188,7 @@ void addReplica(PartitionId partitionId) throws ReplicationException {
             tokenHelper.getFindTokenFactoryFromReplicaType(peerReplica.getReplicaType());
         RemoteReplicaInfo remoteReplicaInfo =
             new RemoteReplicaInfo(peerReplica, cloudReplica, store, findTokenFactory.getNewFindToken(),
-                storeConfig.storeDataFlushIntervalSeconds * SystemTime.MsPerSec * Replication_Delay_Multiplier,
+                storeConfig.storeDataFlushIntervalSeconds * TimeUnit.SECONDS.toMicros(1) * Replication_Delay_Multiplier,
```
cgtz, line 191: Generally it's preferred to use this pattern: `TimeUnit.SECONDS.toMillis(frontendConfig.cacheValiditySeconds)` instead of multiplying times one unit.

This will handle some overflow cases and makes the code more concise.

Also, I noticed that microsecond (10E-6) conversions are used where milliseconds (10E-3) (`TimeUnit::toMillis`) is needed. Please double check the places where `toMicros` is used.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -571,8 +571,8 @@ void addEntry(IndexEntry entry, Offset fileEndOffset) throws StoreException {
       long operationTimeInMs = entry.getValue().getOperationTimeInMs();
       if (operationTimeInMs == Utils.Infinite_Time) {
         lastModifiedTimeSec.set(time.seconds());
-      } else if ((operationTimeInMs / Time.MsPerSec) > lastModifiedTimeSec.get()) {
-        lastModifiedTimeSec.set(operationTimeInMs / Time.MsPerSec);
+      } else if ((operationTimeInMs / TimeUnit.SECONDS.toMicros(1)) > lastModifiedTimeSec.get()) {
```
cgtz, line 574: Instead of a division, we can do, `TimeUnit.MILLISECONDS.toSeconds(operationTimeInMs)`. Same comment for other files where division is used to go from a smaller to larger unit.

cgtz (changes requested):
Thank you for working on this issue! Sorry for the delayed review.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
djKooks: @cgtz thanks for review.
> Generally it's preferred to use this pattern: TimeUnit.SECONDS.toMillis(frontendConfig.cacheValiditySeconds) instead of multiplying times one unit.
- Previously `SystemTime.MsPerSec` was equivalent to `TimeUnit.SECONDS.toMicros(1)`, so I've changed in this way. Is `frontendConfig.cacheValiditySeconds` a new feature?

> I noticed that microsecond (10E-6) conversions are used where milliseconds (10E-3) (TimeUnit::toMillis) is needed. Please double check the places where toMicros is used.
- Parameter for `storeConfig.storeDataFlushIntervalSeconds * TimeUnit.SECONDS.toMicros(1)` is used as `tokenPersistIntervalInMs` in `RemoteReplicaInfo`, and seems it should be `milli-second` as you commented. But just bit worried that this change can effect other logics. Is it okay to change on?(because I couldn't find document/comment about 'RemoteReplicaInfo' so not sure of this...)

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
djKooks: @cgtz thanks for comment. I've made few changes.
But I'm bit confused that which one is micro, and which one is milli...is there some documentation about this_/

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -238,8 +238,10 @@ public int compare(IndexEntry e1, IndexEntry e2) {
       if (scheduler != null) {
         // start scheduler thread to persist index in the background
         persistorTask = scheduler.scheduleAtFixedRate(persistor,
-            config.storeDataFlushDelaySeconds + new Random().nextInt(Time.SecsPerMin),
-            config.storeDataFlushIntervalSeconds, TimeUnit.SECONDS);
+                config.storeDataFlushDelaySeconds +
+                        new Random().nextInt(Math.toIntExact(TimeUnit.MINUTES.toSeconds(1))),
```
kieraCurtis, line 242: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [DMI: Random object created and used only once (DMI_RANDOM_USED_ONLY_ONCE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#dmi-random-object-created-and-used-only-once-dmi-random-used-only-once).
This code creates a java.util.Random object, uses it to generate one random number, and then discards the Random object. This produces mediocre quality random numbers and is inefficient. If possible, rewrite the code so that the Random object is created once and saved, and each time a new random number is required invoke a method on the existing Random object to obtain it.
If it is important that the generated Random numbers not be guessable, you *must* not create a new Random for each random number; the values are too easily guessable. You should strongly consider using a java.security.SecureRandom instead (and avoid allocating a new SecureRandom for each random number needed).

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
cgtz: Generally in our codebase, whenever there is an abbreviation `ms`, it stands for milliseconds (1/1000 of a second). Typically microseconds (1/1000000 seconds) is abbreviated as `us` or `μs`. I don't believe there are many (or any) places in the ambry project that use microseconds as opposed to nanoseconds or milliseconds.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
djKooks: @kieraCurtis seems reasonable, but it is related with all of Random object used inside the project. Maybe you could create separate PR for this feature 👍

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
djKooks: @cgtz thanks for the feedback. Updated~

ambry-utils/src/main/java/com/github/ambry/utils/Time.java
```diff
@@ -13,6 +13,7 @@
  */
 package com.github.ambry.utils;
 
+import java.util.concurrent.TimeUnit;
```
cgtz, line 16: unused import

ambry-tools/src/main/java/com/github/ambry/tools/perf/ServerReadPerformance.java
```diff
@@ -172,7 +173,7 @@ public static void main(String args[]) {
 
       ToolUtils.ensureOrExit(listOpt, options, parser);
 
-      long measurementIntervalNs = options.valueOf(measurementIntervalOpt) * SystemTime.NsPerSec;
+      long measurementIntervalNs = options.valueOf(measurementIntervalOpt) * TimeUnit.SECONDS.toNanos(1);
```
cgtz, line 176: i guess you could supply the value of the option directly to the timeunit converter instead of multiplying

cgtz:
You may need to rebase or merge and fix conflicts in BlobStoreCompactorTest for the CI to run.

djKooks:
@cgtz thanks for review 🙏 
Updated the comments.

cgtz:
Hi @djKooks, there seem to be some tests timing out or failing after this change. Do you want to take a try at determining the issue?

cgtz:
Hi @djKooks, are you interested in continuing to work on this PR? Do you need help to investigate the tests that were timing out?

djKooks:
@cgtz oh, I've missed your comment 🙇 let me keep looking on.

ankagrawal:
> @cgtz oh, I've missed your comment 🙇 let me keep looking on.

Hi @djKooks.. do let us know if you need any support to close out this PR

kination:
@ankagrawal thanks for notice.

Because I'm in bit busy situation, so couldn't do it directly. Also because it has been far different with main branch, I'll close this and create new PR when prepared.

PR #1799, opened 2021-03-10
Dont include get in StatsReport frontend operation

justinlin-linkedin:
Remove "get" from stats report frontend operation literal. “GET" is already reflected in http method.


cgtz (approved):
lgtm

PR #1800, opened 2021-03-11
[Stats Report Migration]Add getAggregatedStats in AccountService

SophieGuo410:


ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -118,6 +119,17 @@
     throw new UnsupportedOperationException("This method is not supported");
   }
 
+  /**
+   * Get aggregated account stats from given cluster in specific fabric group.
```
jsjtzyy, line 123: nit: remove `in specific fabric group`

jsjtzyy (commented):
Looks good to me. I understand the reason to add such method but I feel like it's probably not in the scope of account service. Will discuss this offline.

PR #1801, opened 2021-03-11
[StorageQuota] Unify interface with QuotaManager

justinlin-linkedin:
Integrate storage quota into unified quota manager interface.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
@@ -29,7 +29,7 @@ public interface QuotaManager {
   /**
    * Method to initialize the {@link QuotaManager}.
    */
-  void init();
+  void init() throws InstantiationException;
 
   /**
    * Computes the overall boolean recommendation to throttle a request or not for all the types of request quotas supported.
--- a/ambry-api/src/test/java/com/github/ambry/rest/RestUtilsTest.java
+++ b/ambry-api/src/test/java/com/github/ambry/rest/RestUtilsTest.java
@@ -16,6 +16,7 @@ package com.github.ambry.rest;
 import com.github.ambry.account.Account;
 import com.github.ambry.account.Container;
 import com.github.ambry.account.InMemAccountService;
+import com.github.ambry.frontend.Operations;
 import com.github.ambry.messageformat.BlobProperties;
 import com.github.ambry.protocol.GetOption;
 import com.github.ambry.quota.QuotaName;
@@ -27,21 +28,26 @@ import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.TestUtils;
 import com.github.ambry.utils.Utils;
 import java.io.UnsupportedEncodingException;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Random;
 import java.util.Set;
 import java.util.TimeZone;
+import java.util.stream.Collectors;
 import org.json.JSONException;
 import org.json.JSONObject;
 import org.junit.Test;
@@ -904,6 +910,36 @@ public class RestUtilsTest {
         fail("Decode of invalid encoded header " + invalidEncodedHeader + " should fail.");
       } catch (IllegalArgumentException iaEx) {
       }
+    }
+  }
+
+  /**
+   * Test request is upload request.
+   * @throws Exception
+   */
+  @Test
+  public void testIsUploadRequest() throws Exception {
+    List<String> operations = Utils.getStaticFieldValuesAsStrings(Operations.class).collect(Collectors.toList());
+    RestMethod[] methods = RestMethod.values();
+
+    for (String operation : operations) {
+      for (RestMethod method : methods) {
+        JSONObject header = new JSONObject();
+        header.put(RestUtils.InternalKeys.REQUEST_PATH,
+            RequestPath.parse("/" + operation, Collections.emptyMap(), Collections.emptyList(), "ambry-test"));
+        RestRequest request = createRestRequest(method, "/" + operation, header);
+        boolean isUpload = RestUtils.isUploadRequest(request);
+        assertEquals(operation.equals(Operations.NAMED_BLOB) && method == RestMethod.PUT, isUpload);
+      }
+    }
+
+    for (RestMethod method : methods) {
+      JSONObject header = new JSONObject();
+      header.put(RestUtils.InternalKeys.REQUEST_PATH,
+          RequestPath.parse("/", Collections.emptyMap(), Collections.emptyList(), "ambry-test"));
+      RestRequest request = createRestRequest(method, "/", header);
+      boolean isUpload = RestUtils.isUploadRequest(request);
+      assertEquals(method == RestMethod.POST, isUpload);
     }
   }
 
--- a/ambry-api/src/test/java/com/github/ambry/rest/RestUtilsTest.java
+++ b/ambry-api/src/test/java/com/github/ambry/rest/RestUtilsTest.java
@@ -28,14 +28,11 @@ import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.TestUtils;
 import com.github.ambry.utils.Utils;
 import java.io.UnsupportedEncodingException;
-import java.lang.reflect.Field;
-import java.lang.reflect.Modifier;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.text.SimpleDateFormat;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Date;
@@ -931,6 +928,16 @@ public class RestUtilsTest {
         boolean isUpload = RestUtils.isUploadRequest(request);
         assertEquals(operation.equals(Operations.NAMED_BLOB) && method == RestMethod.PUT, isUpload);
       }
+    }
+    // One exception for PUT named blob
+    {
+      JSONObject header = new JSONObject();
+      header.put(RestUtils.InternalKeys.REQUEST_PATH,
+          RequestPath.parse("/" + Operations.NAMED_BLOB, Collections.emptyMap(), Collections.emptyList(),
+              "ambry-test"));
+      header.put(RestUtils.Headers.UPLOAD_NAMED_BLOB_MODE, "STITCH");
+      RestRequest request = createRestRequest(RestMethod.PUT, "/" + Operations.NAMED_BLOB, header);
+      assertFalse(RestUtils.isUploadRequest(request));
     }
 
     for (RestMethod method : methods) {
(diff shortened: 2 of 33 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestServiceFactory.java
```diff
@@ -94,23 +93,21 @@ public RestRequestService getRestRequestService() {
...
       AccountStatsStore accountStatsStore =
           Utils.<AccountStatsStoreFactory>getObj(frontendConfig.accountStatsStoreFactory, verifiableProperties,
               clusterMapConfig, new StatsManagerConfig(verifiableProperties),
               clusterMap.getMetricRegistry()).getAccountStatsStore();
-      if (frontendConfig.enableStorageQuotaService) {
-        storageQuotaService =
-            Utils.<StorageQuotaServiceFactory>getObj(frontendConfig.storageQuotaServiceFactory, verifiableProperties,
-                accountStatsStore, clusterMap.getMetricRegistry()).getStorageQuotaService();
-      }
+      QuotaConfig quotaConfig = new QuotaConfig(verifiableProperties);
```
ankagrawal, line 100: My concern here is that we are passing quotaManager object to SecurityService without calling quotaManager.init. For now it could be ok, but in general this is an unsafe practice, because there is nothing that prevents SecurityService from using uninitialized QuotaManager before the quotaManager.init is called in FrontendRestRequestService.
Why can't we call storageService.start() here?

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
```diff
@@ -66,28 +66,11 @@ private String restMethodToCostMetric(RestMethod restMethod) {
...
-        }
-        return contentSize / BYTES_IN_GB;
-      case DELETE:
-      case PUT:
-        return CU_COST_UNIT / BYTES_IN_GB; // Assuming 1 chunk worth of storage cost for deletes and ttl updates.
-      default:
-        return 0;
-    }
+  private double calculateStorageCost(RestRequest restRequest) {
+    return RestUtils.isUploadRequest(restRequest) ? restRequest.getBlobBytesReceived() / BYTES_IN_GB : 0;
```
ankagrawal, line 73: I think this cost calculation logic needs more thought for stitched requests. Also we are charging customers for ttl updates and deletes as well.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
```diff
@@ -66,28 +65,11 @@ private String restMethodToCostMetric(RestMethod restMethod) {
...
-        }
-        return contentSize / BYTES_IN_GB;
-      case DELETE:
-      case PUT:
-        return CU_COST_UNIT / BYTES_IN_GB; // Assuming 1 chunk worth of storage cost for deletes and ttl updates.
-      default:
-        return 0;
-    }
+  private double calculateStorageCost(RestRequest restRequest) {
+    return RestUtils.isUploadRequest(restRequest) ? restRequest.getBlobBytesReceived() : 0;
```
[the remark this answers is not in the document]
justinlin-linkedin: My thinking is that storage cost for now, should work only for uploads. This is because when we calculate the storage used for certain container, we only accumulate the PUT, not the TTL_UPDATE, not the DELETE either. (of course, PUT is the chunk content with blob metadata).
And for stitch upload, since the chunk content is the list of blob ids, it's very hard to know the size of the metadata chunk in this write path. I figured we can probably ignore it in the frontend, and let the server figure it out.

ambry-frontend/src/test/java/com/github/ambry/frontend/AmbrySecurityServiceTest.java
```diff
@@ -269,14 +270,15 @@ public void postProcessQuotaManagerTest() throws Exception {
    */
   @Test
   public void postProcessStorageQuotaServiceTest() throws Exception {
+    Assume.assumeTrue(false);
```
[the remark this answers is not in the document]
justinlin-linkedin: I already removed this test, it seems like the quota manager test is already done in FrontendQuotaIntegrationTest.

ambry-frontend/src/test/java/com/github/ambry/frontend/UserQuotaRequestCostPolicyTest.java
```diff
@@ -72,15 +77,15 @@ public void testCalculateRequestCost() throws UnsupportedEncodingException, URIS
         DEFAULT_ENCRYPTED_FLAG), null, DEFAULT_LIFE_VERSION);
     restRequest = createMockRequestWithMethod(RestMethod.POST, 8 * 1024 * 1024);
     costMap = quotaRequestCostPolicy.calculateRequestCost(restRequest, restResponseChannel, blobInfo);
-    verifyWriteCost(costMap, 2, .0078125);
+    verifyWriteCost(costMap, 2, 8 * 1024 * 1024);
```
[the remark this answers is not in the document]
justinlin-linkedin: Updated to GB already, please take another look,  thanks.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
justinlin-linkedin: same above.

ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
```diff
@@ -29,7 +29,7 @@
   /**
    * Method to initialize the {@link QuotaManager}.
    */
-  void init();
+  void init() throws InstantiationException;
```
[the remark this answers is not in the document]
justinlin-linkedin: I agree, however, the chain of having InstantiationException in start/init method is too long from RestServer all the way down to quotaManager. I would have to change all of them. Instead, I will add a log message here when there is an exception.

ambry-api/src/test/java/com/github/ambry/rest/RestUtilsTest.java
```diff
@@ -907,6 +913,44 @@ public void decodeKVHeaderValueTest() {
     }
   }
 
+  /**
+   * Test request is upload request.
+   * @throws Exception
+   */
+  @Test
+  public void testIsUploadRequest() throws Exception {
+    List<Field> fields = Arrays.stream(Operations.class.getDeclaredFields())
```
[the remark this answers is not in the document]
justinlin-linkedin: that's great, will update.

ambry-api/src/main/java/com/github/ambry/quota/QuotaManager.java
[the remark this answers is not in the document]
cgtz: Ah, I didn't see that the chain of callers also dealt with InstantiationException. You could also consider using unchecked exceptions.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -141,4 +154,8 @@ public QuotaConfig getQuotaConfig() {
     }
     return quotaSourceObjectMap;
   }
+
+  public final static class AmbryQuotaManagerComponents {
+    public static AccountStatsStore accountStatsStore;
```
ankagrawal, line 159: I don't think there is really any need to create a separate AmbryQuotaManagerComponents class. Just make accountStatsStore private final and add a getter inside the AmbryQuotaManager itself.
Pass the accountStatsStore in AmbryQuotaManager's constructor.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManagerFactory.java
```diff
@@ -25,10 +26,12 @@
...
   /**
    * @param quotaConfig {@link QuotaConfig} object.
+   * @param accountStatsStore
    * @throws ReflectiveOperationException
    */
-  public AmbryQuotaManagerFactory(QuotaConfig quotaConfig, ThrottlePolicy throttlePolicy, AccountService accountService)
-      throws ReflectiveOperationException {
+  public AmbryQuotaManagerFactory(QuotaConfig quotaConfig, ThrottlePolicy throttlePolicy, AccountService accountService,
+      AccountStatsStore accountStatsStore) throws ReflectiveOperationException {
+    AmbryQuotaManager.AmbryQuotaManagerComponents.accountStatsStore = accountStatsStore;
```
ankagrawal, line 34: See comments for accountStatsStore above.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
ankagrawal: Right now the contract is that any request that comes to Ambry storage node, needs to get charged with a min storage cost which is same as if for a 4MB chunk. The same logic is also used by ResourceIntelligence when we send them the cost values. If we want to change this behavior we should talk with Rob if he has any concerns about this.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestServiceFactory.java
justinlin-linkedin: I am adding all the initialization work in the start method so the constructor method would be a light-weight method with only variable assignment statements. 
If we follow this rule, then the security service would not do anything in the its constructor.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
justinlin-linkedin: the problem is that when constructing a StorageQuotaAdaptorEnforcer, AmbryQuotaManager has not yet constructed. So we need a static accessor to set and get accountStatsStore.
One way to do this is to add static field in AmbryQuotaManager and set this field in the AmbryQuotaManagerFactory so we can get this field out in the StorageQuotaAdaptorEnforcer. But this way has no difference from having a static inner class.
Another possible solution is to pass accountStatsStore to all the enforcer factories' constructors, but I figure all the other enforcers don't need it, only storage enforcer need it, then it seems intrusive.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
justinlin-linkedin: I don't want to miss use this info here, if this storage cost is shared for RI, then I should probably ignore this cost in my enforcer implementation. I can get the blob size from restRequest when charging storage quota. i don't need cost map passed to my enforcer.

ankagrawal (approved):
lgtm

cgtz (changes requested):

cgtz:
LGTM, I just had a few comments/questions above ^

ambry-quota/src/main/java/com/github/ambry/quota/storage/StorageQuotaAdaptorQuotaEnforcerFactory.java
```diff
@@ -0,0 +1,45 @@
...
+public class StorageQuotaAdaptorQuotaEnforcerFactory implements QuotaEnforcerFactory {
+  private final StorageQuotaAdaptorQuotaEnforcer quotaEnforcer;
+
+  /**
+   * Constructor to instantiate a factory.
+   * @param quotaConfig the {@link QuotaConfig}.
+   * @param quotaSource the {@link QuotaSource}.
+   * @throws Exception
+   */
+  public StorageQuotaAdaptorQuotaEnforcerFactory(QuotaConfig quotaConfig, QuotaSource quotaSource) throws Exception {
```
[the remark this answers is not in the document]
justinlin-linkedin: The reason to have a static getter and setter method is because only this factory requires `accountStatsStore` and other EnforcerFactories don't need it. After all only storage quota enforcer would fetch data from mysql database.

I suppose we can have it in the constructor, it looks more elegant that way.

cgtz (approved):

PR #1808, opened 2021-03-23
Handle BlobInfo GET requests properly in cost calculation

lightningrob:


ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
```diff
@@ -112,7 +114,11 @@ private double calculateCapacityUnitCost(RestRequest restRequest, RestResponseCh
         double cost = Math.ceil(contentSize / CU_COST_UNIT);
         return (cost > 0) ? cost : 1;
       case GET:
-        if (restRequest.getUri().contains(Operations.GET_SIGNED_URL)) {
+        String uri = restRequest.getUri();
+        if (uri.contains(Operations.GET_SIGNED_URL)) {
+          return INDEX_ONLY_COST;
+        } else if (uri.contains(SubResource.BlobInfo.name()) || uri.contains(SubResource.UserMetadata.name())) {
```
lightningrob, line 120: I originally used:
```
RestUtils.SubResource subResource = getRequestPath(restRequest).getSubResource();
if (subResource == SubResource.BlobInfo || subResource == SubResource.UserMetadata) { ... }
```
But this threw NPE in unit test that uses mock request with no request path.  Does it make more sense to change the tests?

cgtz: Yep, I think it makes sense to change the test to set the `RequestPath` in the mock request. Many of the frontend package tests do this (or the argument is set automatically by the production code in FrontendRestRequestService). It may be good to add a test utility to do this.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
```diff
@@ -112,7 +114,11 @@ private double calculateCapacityUnitCost(RestRequest restRequest, RestResponseCh
         double cost = Math.ceil(contentSize / CU_COST_UNIT);
         return (cost > 0) ? cost : 1;
       case GET:
-        if (restRequest.getUri().contains(Operations.GET_SIGNED_URL)) {
+        String uri = restRequest.getUri();
+        if (uri.contains(Operations.GET_SIGNED_URL)) {
+          return INDEX_ONLY_COST;
```
cgtz, line 119: getSignedUrl calls do not even check for blob existence so could they be considered near zero cost?

Also, once `RequestPath` is used, you could do `getRequestPath(restRequest).matchesOperation(GET_SIGNED_URL)`

cgtz (changes requested):
Looks good after changes made to use RequestPath.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
cgtz, line 120: One extra thing I noticed is that any HEAD request should also fall into this condition regardless of if it has the BlobInfo subresource. This is because HEAD will only check blob metadata since it only needs to return response headers.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
```diff
@@ -112,7 +114,11 @@ private double calculateCapacityUnitCost(RestRequest restRequest, RestResponseCh
         double cost = Math.ceil(contentSize / CU_COST_UNIT);
```
cgtz, line 114: Please also use `RequestPath::matchesOperation` for line 107

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
cgtz: Sorry ignore this, I didn't see that HEAD requests will fall through to the default block.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
ankagrawal: For now we are going with the assumption that any call to Ambry incurs a min INDEX_ONLY_COST. The min cost is same as cost to fetch a 4MB chunk, and Ambry CAPACITY_UNIT is defined in terms of 4MB reads/writes.

Do you think we should change the granularity of that? Is it overcharging to say any request to Ambry incurs cost of 1 CAPACITY_UNIT with the definition above?

lightningrob: We could make INDEX_ONLY_COST configurable, and set it to a fractional value (like 0.25) if we wanted to.

cgtz: For now, maybe we can just keep it as whole units. If we see that heavy users of signed URLs appear to be charged too much, maybe we can consider reducing the cost of the operations that don't even hit the storage nodes.

cgtz (approved):

PR #1809, opened 2021-03-24
Ignore conflicts on unmodified containers

lightningrob:
Add config option to delay account cache update until next DB sync.


Arun-LinkedIn (approved):
Looks good to me.

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
```diff
@@ -107,6 +108,14 @@
   @Default("50")
   public final int dbExecuteBatchSize;
 
+  /**
+   * If true, account service will write changes to the cache immediately after updating the persistent store.
+   * If false, the cache will be updated on the subsequent sync call.
+   * If written immediately, updated accounts and containers will have out of date snapshot version.
```
jsjtzyy, line 114: I have some trouble understanding this. Why updated accounts/containers will have out-of-date version if written immediately?

jsjtzyy (approved):
looks good to me

ambry-api/src/main/java/com/github/ambry/config/MySqlAccountServiceConfig.java
lightningrob: Because the snapshot version is incremented inside the AccountDao methods during DB update, and this change is not reflected in the immutable account/container objects that the MySqlAccountService is holding when it updates the cache.  And propagating that info backwards is fairly cumbersome.

jsjtzyy: Ah, got it, thanks for clarification.

PR #1814, opened 2021-03-28
Add a ContainerUpdate consumer to listen for updates to a Container.

ankagrawal:
Add a listener to update a container's quota when a container is created or updated in AccountService.


justinlin-linkedin:
We already have account updater consumers in AccountService, which exposes the same information and invoke the callback at the same time. i don't see a reason to add container updater consumers here.

cgtz:
> We already have account updater consumers in AccountService, which exposes the same information and invoke the callback at the same time. i don't see a reason to add container updater consumers here.

Also, if there is ever an account-level quota feature in the future, we may want to react to account property changes in addition to container property changes.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -90,6 +90,7 @@
  *   back to zookeeper.
  * </p>
  */
+// TODO ANKUR why does it both extend AbstractAccountService and implement AccountService??
```
lightningrob, line 93: I probably just forgot to remove the implement clause.  Feel free to.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -326,6 +328,8 @@ public void updateAccounts(Collection<Account> accounts) throws AccountServiceEx
     }
     // Tell the world
     publishChangeNotice();
+    containerUpdateConsumers.forEach(containerUpdateConsumer -> containerUpdateConsumer.accept(
```
lightningrob, line 331: These calls should be made in HelixAccountService too.  Consider adding `AbstractAccountService.notifyAccountUpdateConsumers` and calling from both impls.

PR #1822, opened 2021-04-14
[StatsMigration] Remove database rows for missing partition/account/container in AccountReports 

justinlin-linkedin:
When a new StatsWrapper is generated and persisted to database, we only override the database rows where the storage value is different. Since we have a copy of previous StatsWrapper, we can easily know which database rows to update.
However, we don't remove those database rows when any partition or account or container no longer exists in the host, we just silently ignore it.

This PR would take the absence of partition/account/container into consideration and remove corresponding database rows.
We are not using batch operations for these DELETEs, since we know the delete should be a rare operation.

### DIFF
--- a/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountReportsDao.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountReportsDao.java
@@ -132,6 +132,13 @@ public class AccountReportsDao {
     }
   }
 
+  /**
+   * Delete container storage usage rows for given {@code clusterName}, {@code hostname} and {@code partitionId}.
+   * @param clusterName The clusterName
+   * @param hostname The hostname
+   * @param partitionId The partitionId
+   * @throws SQLException
+   */
   void deleteStorageUsageForPartition(String clusterName, String hostname, int partitionId) throws SQLException {
     try {
       long startTimeMs = System.currentTimeMillis();
@@ -148,6 +155,14 @@ public class AccountReportsDao {
     }
   }
 
+  /**
+   * Delete container storage usage rows for given {@code clusterName}, {@code hostname}, {@code partitionId} and {@code accountId}.
+   * @param clusterName The clusterName
+   * @param hostname The hostname
+   * @param partitionId The partitionId
+   * @param accountId The accountId
+   * @throws SQLException
+   */
   void deleteStorageUsageForAccount(String clusterName, String hostname, int partitionId, int accountId)
       throws SQLException {
     try {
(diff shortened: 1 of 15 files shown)


ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -233,8 +241,57 @@ public void storeAccountStats(StatsWrapper statsWrapper) throws SQLException {
...
+    int deleteStatementCounter = 0;
+    // Now delete the rows that appear in previousStats but not in the currentStats
+    for (Map.Entry<String, StatsSnapshot> prevPartitionMapEntry : prevPartitionMap.entrySet()) {
+      String partitionIdKey = prevPartitionMapEntry.getKey();
+      StatsSnapshot prevAccountStatsSnapshot = prevPartitionMapEntry.getValue();
+      StatsSnapshot currAccountStatsSnapshot = currPartitionMap.get(partitionIdKey);
+      int partitionId = Utils.partitionIdFromStatsPartitionKey(partitionIdKey);
+      if (prevAccountStatsSnapshot != null && prevAccountStatsSnapshot.getSubMap() != null && (
+          currAccountStatsSnapshot == null || currAccountStatsSnapshot.getSubMap() == null)) {
+        accountReportsDao.deleteStorageUsageForPartition(clusterName, hostname, partitionId);
```
jsjtzyy, line 257: I feel like deleting partition only happens when replica is decommissioned on current node. No need to delete its rows if partition is empty.

jsjtzyy: On second thought, empty partition should be very rare, so I think it's fine to delete it in mysql.

jsjtzyy (approved):
lgtm

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountReportsDao.java
```diff
@@ -123,6 +132,61 @@ void queryStorageUsageForHost(String clusterName, String hostname, ContainerUsag
     }
   }
 
+  void deleteStorageUsageForPartition(String clusterName, String hostname, int partitionId) throws SQLException {
```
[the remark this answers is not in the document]
justinlin-linkedin: It will not, I am not add these methods to the callback function of partition state transition because we already can do it by just comparing the previous stats and current stats.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
justinlin-linkedin: empty partition means all the blobs in this partition are deleted (still in retention maybe, but deleted)

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -233,8 +241,57 @@ public void storeAccountStats(StatsWrapper statsWrapper) throws SQLException {
       }
     }
     batch.flush();
-    storeMetrics.publishTimeMs.update(System.currentTimeMillis() - startTimeMs);
     storeMetrics.batchSize.update(batchSize);
+    storeMetrics.insertAccountStatsTimeMs.update(System.currentTimeMillis() - startTimeMs);
```
ankagrawal, line 245: nit: either now or later, this method needs to be divided into smaller methods. This method is way too long, and seems to be doing multiple things, making it an ideal candidate for refactoring.

ankagrawal (approved):
A minor comment. Looks good to me otherwise.

PR #1828, opened 2021-04-20
Fix NPE in NamedBlobPutHandler

SophieGuo410:
Bug fix for routerStitchBlobCallback.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
@@ -32,6 +32,7 @@ import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Utils;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Objects;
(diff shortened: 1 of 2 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -44,6 +47,7 @@
   private final IdSigningService idSigningService;
   private final NamedBlobDb namedBlobDb;
   private final FrontendMetrics frontendMetrics;
+  private static final Logger logger = LoggerFactory.getLogger(AmbryIdConverterFactory.class);
```
cgtz, line 50: it seems like this logger is not used.

cgtz (approved):
LGTM thanks for making this fix!

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
SophieGuo410: Removed the redundant logic.

PR #1829, opened 2021-04-20
Close notifications system during RestServer shutdown

jsjtzyy:
This PR explicitly closes notification system when shutting down rest server. It will tell notification system when to publish remaining events in the queue (if any).


justinlin-linkedin (approved):
LGTM

PR #1831, opened 2021-04-22
Add compaction related metric

zzmao:
Add compaction related metric

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactionDetails.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactionDetails.java
@@ -33,16 +33,19 @@ class CompactionDetails {
 
   private final long referenceTimeMs;
   private final List<LogSegmentName> logSegmentsUnderCompaction;
+  private final CostBenefitInfo costBenefitInfo;
 
   /**
    * Construct a representation of all the details required for a compaction cycle.
    * @param referenceTimeMs the epoch time to use to get the same results from the {@link BlobStore} as when these
    *                        details were created.
    * @param logSegmentsUnderCompaction the names of the {@link LogSegment} under compaction.
+   * @param costBenefitInfo
    * @throws IllegalArgumentException if {@code referenceTimeMs} < 0 or if {@code logSegmentsUnderCompaction} has no
    * elements.
    */
-  CompactionDetails(long referenceTimeMs, List<LogSegmentName> logSegmentsUnderCompaction) {
+  CompactionDetails(long referenceTimeMs, List<LogSegmentName> logSegmentsUnderCompaction,
+      CostBenefitInfo costBenefitInfo) {
     if (referenceTimeMs < 0 || logSegmentsUnderCompaction.size() == 0) {
       throw new IllegalArgumentException(
           "Illegal arguments provided. Ref time: [" + referenceTimeMs + "]. " + "Segments under compaction size: ["
@@ -50,6 +53,7 @@ class CompactionDetails {
     }
     this.referenceTimeMs = referenceTimeMs;
     this.logSegmentsUnderCompaction = logSegmentsUnderCompaction;
+    this.costBenefitInfo = costBenefitInfo;
   }
 
   /**
@@ -69,7 +73,7 @@ class CompactionDetails {
         for (int i = 0; i < segmentCount; i++) {
           logSegmentsUnderCompaction.add(LogSegmentName.fromString(Utils.readIntString(stream)));
         }
-        details = new CompactionDetails(referenceTime, logSegmentsUnderCompaction);
+        details = new CompactionDetails(referenceTime, logSegmentsUnderCompaction, null);
         break;
       default:
         throw new IllegalArgumentException("Unrecognized version: " + version);
@@ -129,6 +133,10 @@ class CompactionDetails {
       bufWrap.put(segmentNameBytes);
     }
     return buf;
+  }
+
+  CostBenefitInfo getCostBenefitInfo() {
+    return costBenefitInfo;
   }
 
   @Override
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactionDetails.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactionDetails.java
@@ -40,7 +40,7 @@ class CompactionDetails {
    * @param referenceTimeMs the epoch time to use to get the same results from the {@link BlobStore} as when these
    *                        details were created.
    * @param logSegmentsUnderCompaction the names of the {@link LogSegment} under compaction.
-   * @param costBenefitInfo
+   * @param costBenefitInfo the {@link CostBenefitInfo} based on current Compaction Details
    * @throws IllegalArgumentException if {@code referenceTimeMs} < 0 or if {@code logSegmentsUnderCompaction} has no
    * elements.
    */
--- a/ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
@@ -64,7 +64,7 @@ class StatsBasedCompactionPolicy implements CompactionPolicy {
                 blobStoreStats.getMaxBlobSize());
         if (bestCandidateToCompact != null) {
           details = new CompactionDetails(validDataSizeByLogSegment.getFirst(),
-              bestCandidateToCompact.getSegmentsToCompact());
+              bestCandidateToCompact.getSegmentsToCompact(), bestCandidateToCompact);
           logger.info("Best candidate to compact {}", bestCandidateToCompact);
         } else {
           logger.trace("No best candidate found");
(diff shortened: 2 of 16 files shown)


ambry-store/src/main/java/com/github/ambry/store/CompactionDetails.java
```diff
@@ -33,23 +33,27 @@
...
   private final long referenceTimeMs;
   private final List<LogSegmentName> logSegmentsUnderCompaction;
+  private final CostBenefitInfo costBenefitInfo;
 
   /**
    * Construct a representation of all the details required for a compaction cycle.
    * @param referenceTimeMs the epoch time to use to get the same results from the {@link BlobStore} as when these
    *                        details were created.
    * @param logSegmentsUnderCompaction the names of the {@link LogSegment} under compaction.
+   * @param costBenefitInfo
```
jsjtzyy, line 43: nit: complete java doc please

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -64,7 +64,7 @@ public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapac
                 blobStoreStats.getMaxBlobSize());
         if (bestCandidateToCompact != null) {
           details = new CompactionDetails(validDataSizeByLogSegment.getFirst(),
-              bestCandidateToCompact.getSegmentsToCompact());
+              bestCandidateToCompact.getSegmentsToCompact(), bestCandidateToCompact);
```
jsjtzyy, line 67: The 2nd parameter seems to be redundant if we already pass `bestCandidateToCompact` in.

jsjtzyy (approved):
Overall looks good to me.  A couple of minor comments.

ambry-store/src/main/java/com/github/ambry/store/CompactionDetails.java
zzmao: done

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
zzmao: In some tests, the 2nd parameters is not related to 3rd parameter, so we can't remove the 2nd one.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -231,14 +232,25 @@ public StoreMetrics(String prefix, MetricRegistry registry) {
...
     registry.register(MetricRegistry.name(Log.class, prefix + "PercentageUsedCapacity"), percentageUsedCapacity);
     Gauge<Long> currentSegmentCount = index::getLogSegmentCount;
     registry.register(MetricRegistry.name(Log.class, prefix + "CurrentSegmentCount"), currentSegmentCount);
+    Gauge<Long> currentInvalidDataSize = () -> {
+      long invalidDataSize = -1;
+      try {
+        invalidDataSize = index.getLogUsedCapacity() - blobStoreStats.getValidSize(
+            new TimeRange(System.currentTimeMillis(), blobStoreStats.getBucketSpanTimeInMs())).getSecond();
+      } catch (StoreException e) {
+      }
```
justinlin-linkedin, line 250: please log out some error message here.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -231,14 +232,25 @@ public StoreMetrics(String prefix, MetricRegistry registry) {
...
     Gauge<Long> currentCapacityUsed = index::getLogUsedCapacity;
     registry.register(MetricRegistry.name(Log.class, prefix + "CurrentCapacityUsed"), currentCapacityUsed);
     Gauge<Double> percentageUsedCapacity = () -> ((double) index.getLogUsedCapacity() / capacityInBytes) * 100;
     registry.register(MetricRegistry.name(Log.class, prefix + "PercentageUsedCapacity"), percentageUsedCapacity);
     Gauge<Long> currentSegmentCount = index::getLogSegmentCount;
     registry.register(MetricRegistry.name(Log.class, prefix + "CurrentSegmentCount"), currentSegmentCount);
+    Gauge<Long> currentInvalidDataSize = () -> {
+      long invalidDataSize = -1;
+      try {
+        invalidDataSize = index.getLogUsedCapacity() - blobStoreStats.getValidSize(
```
justinlin-linkedin, line 247: what is the frequency of calling this method? `getValidSize` should be a small memory operation, but we never test it on real prod cluster before, maybe we want to also add a metric for getValidSize as well.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -250,6 +262,7 @@ private void deregisterIndexGauges(String storeId) {
     registry.remove(MetricRegistry.name(Log.class, prefix + "CurrentCapacityUsed"));
     registry.remove(MetricRegistry.name(Log.class, prefix + "PercentageUsedCapacity"));
     registry.remove(MetricRegistry.name(Log.class, prefix + "CurrentSegmentCount"));
+    registry.remove(MetricRegistry.name(Log.class, prefix + "CurrentInvalidDataSize"));
```
justinlin-linkedin, line 265: just being curious, why are we removing these metrics here? It's just that I've never seen we remove gauge metrics  anywhere else.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -285,11 +298,30 @@ private void deregisterHardDeleteMetric(String storeId) {
...
   }
 
-  void initializeCompactorGauges(String storeId, final AtomicBoolean compactionInProgress) {
+  void initializeCompactorGauges(String storeId, final AtomicBoolean compactionInProgress,
+      AtomicReference<CompactionDetails> compactionDetailsAtomicReference) {
     String prefix = storeId + SEPARATOR;
     Gauge<Long> compactionInProgressGauge = () -> compactionInProgress.get() ? 1L : 0L;
     registry.register(MetricRegistry.name(BlobStoreCompactor.class, prefix + "CompactionInProgress"),
         compactionInProgressGauge);
+    Gauge<Long> compactionCost = () -> {
```
justinlin-linkedin, line 307: I have a big concern about this,  we might have several compactions running at various disks and since they all using the same StoreMetric object, they will override each other.
And since `MetricRegistry.register` is not an idempotent method, if the first compaction registered these metric, the others would throw an exception. we should probably use `MetricRegitry.gauge` method to register this.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
zzmao: it's for test purpose. metric can't be registered twice. In test, store may start twice, so we remove it. `CurrentSegmentCount ` is an existing example.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
zzmao: The metric's name has a prefix, which is storeId, which should be good. No?

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
zzmao: with Invalid data, we can infer the valid data. Since we care about invalid data more, so invalid data is emitted.

zzmao: The frequency might be high. This is a valid concern. Let me think about this.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
justinlin-linkedin: yeah, this is good.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
justinlin-linkedin: it's weird to do such thing, to register a gauge, we can call `MetricRegistry.gauge`, this way, we can register same metric multiple times ( this method would check if it's already registered or not). `MetricRegistry.register` should be avoided. This way, we don't have to remove metrics from registry.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
justinlin-linkedin: it shouldn't take long to finish this method since it's all on memory. I think adding a metric in StoreMetrics and record it every time we call it would just do the work. If the latency is too high, we can think of something else.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
zzmao: added.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
zzmao: You made a good point. 

But for this change, we probably need to use `MetricRegistry.register` because BlobStore may be shutdown due to replica move, so we need to de-register some metrics.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
zzmao: OK. Let's try. This is not in critical path, it should be OK.

justinlin-linkedin (approved):

PR #1833, opened 2021-04-23
Notify quota source on creation of new account or container.

ankagrawal:
When new containers are added to Ambry's account service, the quota source should be notified about it so that it can take required action for quotas of the new container.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
@@ -13,10 +13,15 @@
  */
 package com.github.ambry.quota;
 
+import com.github.ambry.account.Account;
+import com.github.ambry.account.Container;
+
+
 /**
  * Resource for which quota is specified for enforced.
  */
 public class QuotaResource {
+  private final static String DELIM = "_";
   private final String resourceId; // unique identifier for Ambry account, container or host.
   private final QuotaResourceType quotaResourceType;
 
@@ -28,6 +33,26 @@ public class QuotaResource {
   public QuotaResource(String resourceId, QuotaResourceType quotaResourceType) {
     this.resourceId = resourceId;
     this.quotaResourceType = quotaResourceType;
+  }
+
+  /**
+   * Create {@link QuotaResource} from {@link Container}.
+   * @param container {@link Container} object.
+   * @return QuotaResource object.
+   */
+  public static QuotaResource fromContainer(Container container) {
+    return new QuotaResource(
+        String.join(DELIM, String.valueOf(container.getId()), String.valueOf(container.getParentAccountId())),
+        QuotaResourceType.CONTAINER);
+  }
+
+  /**
+   * Create {@link QuotaResource} from {@link Account}.
+   * @param account {@link Account} object.
+   * @return QuotaResource object.
+   */
+  public static QuotaResource fromAccount(Account account) {
+    return new QuotaResource(String.valueOf(account.getId()), QuotaResourceType.CONTAINER);
   }
 
   /**
@@ -48,6 +73,6 @@ public class QuotaResource {
    * Type of Ambry resource for which quota can be applied.
    */
   enum QuotaResourceType {
-    AMBRY_ACCOUNT, AMBRY_CONTAINER
+    ACCOUNT, CONTAINER
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
@@ -80,7 +80,7 @@ public class QuotaResource {
   /**
    * Type of Ambry resource for which quota can be applied.
    */
-  enum QuotaResourceType {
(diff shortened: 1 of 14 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -70,6 +72,10 @@ public AmbryQuotaManager(QuotaConfig quotaConfig, ThrottlePolicy throttlePolicy,
     this.throttlePolicy = throttlePolicy;
     this.quotaConfig = quotaConfig;
     this.quotaMetrics = new QuotaMetrics(metricRegistry);
+    Consumer<Collection<Account>> accountUpdateQuotaConsumer = (accounts) -> {
+      onAccountUpdateNotification(accounts);
+    };
+    accountService.addAccountUpdateConsumer(accountUpdateQuotaConsumer);
```
justinlin-linkedin, line 78: you can just do `accountService.addAccountUpdateConsumer(this::onAccountUpdateNotification);`.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -147,6 +153,19 @@ public QuotaConfig getQuotaConfig() {
     return quotaConfig;
   }
 
+  /**
+   * Notify {@link QuotaSource}s about creation of new Ambry {@link Account} or {@link com.github.ambry.account.Container}.
+   * @param updatedAccounts {@link Collection} of {@link Account}s updated.
+   */
+  protected void onAccountUpdateNotification(Collection<Account> updatedAccounts) {
```
justinlin-linkedin, line 160: `accountUpdateConsumer` callback will be invoked not only when the new account/container is created, but also when the existing account/container is updated, like encryption flag, cache flag, etc..
If we only want `QuotaSource` to deal with newly created account/container, we should probably keep a list of account/container in `AmbryQuotaManager` so we can filter the updates out.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -147,6 +153,19 @@ public QuotaConfig getQuotaConfig() {
...
+    requestQuotaEnforcers.stream().map(QuotaEnforcer::getQuotaSource).forEach(quotaSource -> {
```
justinlin-linkedin, line 161: there are a couple things that can be done to improve this stream operation.
1. move `updatedAccounts.stream()` out so we only have to do it once.
2. filter out the null quota source since there is no guarantee `QuotaEnforcer` will return a non-null `QuotaSource`.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/UnlimitedQuotaSource.java
```diff
@@ -17,15 +17,31 @@
...
+  private final List<QuotaResource> quotaResourceList = new ArrayList<>();
 
   @Override
   public Quota getQuota(QuotaResource quotaResource, QuotaName quotaName) {
     return new Quota<>(quotaName, Long.MAX_VALUE, quotaResource);
   }
+
+  @Override
+  public void updateNewQuotaResources(Collection<QuotaResource> quotaResources) {
+    quotaResourceList.addAll(quotaResources);
```
justinlin-linkedin, line 38: given the fact the "updated" account (not newly created) will also be passed to this method, we might see same quota resource multiple times in the list. One way to fix it is to change the list to set.

justinlin-linkedin (changes requested):

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
justinlin-linkedin: Then please update the comments for this method.

justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
```diff
@@ -13,10 +13,14 @@
  */
 package com.github.ambry.quota;
 
+import com.github.ambry.account.Container;
+
+
 /**
  * Resource for which quota is specified for enforced.
  */
 public class QuotaResource {
```
Arun-LinkedIn, line 22: Just wondering if equals() and hashCode() methods of Object class should be overriden for HashSet to filter duplicates correctly.

ankagrawal: done.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
ankagrawal: done.

Arun-LinkedIn (approved):

ankagrawal (commented):
.

PR #1834, opened 2021-04-29
[StatsMigration] Remove invalid aggregated stats in mysql aggregation task

justinlin-linkedin:
When an account/container is deleted, the host account stats and partition class stats would not generate any data for it so that the aggregated account and partition class stats in aggregation task don't include any data for this account/container. At this time, the aggregated account and partition class stats for this account/container is useless and can be removed.

This PR adds some logic to remove those data and a configuration to turn it on and off.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/accountstats/AccountStatsStore.java
+++ b/ambry-api/src/main/java/com/github/ambry/accountstats/AccountStatsStore.java
@@ -21,7 +21,6 @@ import java.util.Set;
 
 /**
  * The interface that stores and fetches account stats, aggregated account stats.
- * TODO: add method to deal with replica reassignment.
  */
 public interface AccountStatsStore {
   /**
@@ -38,6 +37,14 @@ public interface AccountStatsStore {
    * @throws Exception
    */
   void storeAggregatedAccountStats(StatsSnapshot snapshot) throws Exception;
+
+  /**
+   * Delete aggregated account stats for the given {@code accountId} and {@code containerId}.
+   * @param accountId The account id
+   * @param containerId The container id
+   * @throws Exception
+   */
+  void deleteAggregatedAccountStatsForContainer(short accountId, short containerId) throws Exception;
 
   /**
    * Return individual ambry server's stats for the given {@code hostname}. This is the stats stored by method {@link #storeAccountStats}.
@@ -103,6 +110,12 @@ public interface AccountStatsStore {
    * @throws Exception
    */
   void takeSnapshotOfAggregatedAccountStatsAndUpdateMonth(String monthValue) throws Exception;
+
+  /**
+   * Delete the snapshot of latest aggregated account stats. This is usually used to prepare for the new snapshot.
+   * @throws Exception
+   */
+  void deleteSnapshotOfAggregatedAccountStats() throws Exception;
 
   /**
    * Return all the partition class names and their corresponding partition ids in a set.
@@ -180,6 +193,17 @@ public interface AccountStatsStore {
    * @throws Exception
    */
   void storeAggregatedPartitionClassStats(StatsSnapshot statsSnapshot) throws Exception;
+
+  /**
+   * Delete aggregated partition class stats for the given {@code partitionClassName} and the {@code accountContainerKey}.
+   * The {@code accountContainerKey} is the key in the PartitionClass StatsSnapshot. It includes account id and container id.
+   * The account id and container id can be extracted from this key.
+   * @param partitionClassName The partition class name.
+   * @param accountContainerKey The key that contains account id and container id.
+   * @throws Exception
+   */
+  void deleteAggregatedPartitionClassStatsForAccountContainer(String partitionClassName, String accountContainerKey)
+      throws Exception;
 
   /**
    * Return the aggregated partition class stats for given. The returned StatsSnapshot is constructed in the same way
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
@@ -172,7 +172,7 @@ public class MySqlReportAggregatorTask extends UserContentStore implements Task 
     }
   }
 
-  private boolean removeInvalidAggregatedAccountAndContainerStats(StatsSnapshot currentStats) throws Exception {
+  private void removeInvalidAggregatedAccountAndContainerStats(StatsSnapshot currentStats) throws Exception {
     Map<String, Map<String, Long>> existingStats = accountStatsStore.queryAggregatedAccountStats();
     List<Pair<Short, Short>> toBeDeletedAccountAndContainers = new ArrayList<>();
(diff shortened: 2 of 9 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -161,6 +172,69 @@ public TaskResult run() {
...
+          toBeDeletedAccountAndContainers.add(new Pair<>(accountId, containerId));
+        }
+      }
+    }
+    // delete the account/container stats that are no longer valid
+    for (Pair<Short, Short> accountContainer : toBeDeletedAccountAndContainers) {
+      accountStatsStore.deleteAggregatedAccountStatsForContainer(accountContainer.getFirst(),
+          accountContainer.getSecond());
+    }
+    return !toBeDeletedAccountAndContainers.isEmpty();
```
jsjtzyy, line 192: Looks like the return value is never used. Is it for future use?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -161,6 +172,69 @@ public TaskResult run() {
     }
   }
 
+  private boolean removeInvalidAggregatedAccountAndContainerStats(StatsSnapshot currentStats) throws Exception {
+    Map<String, Map<String, Long>> existingStats = accountStatsStore.queryAggregatedAccountStats();
+    List<Pair<Short, Short>> toBeDeletedAccountAndContainers = new ArrayList<>();
+    for (Map.Entry<String, Map<String, Long>> accountEntry : existingStats.entrySet()) {
+      short accountId = Short.valueOf(accountEntry.getKey());
+      for (String containerIdStr : accountEntry.getValue().keySet()) {
+        short containerId = Short.valueOf(containerIdStr);
```
jsjtzyy, line 181: Suggest using `Short.parseShort()`

ambry-api/src/main/java/com/github/ambry/accountstats/AccountStatsStore.java
```diff
@@ -104,6 +111,12 @@
    */
   void takeSnapshotOfAggregatedAccountStatsAndUpdateMonth(String monthValue) throws Exception;
 
+  /**
+   * Delete the snapshot of latest aggregated account stats. This is usually used to prepare for the new snapshot.
+   * @throws Exception
+   */
+  void deleteSnapshotOfAggregatedAccountStats() throws Exception;
```
jsjtzyy, line 118: Rename it to `deleteMonthlySnapshotOfAggregatedAccountStats` ? Feel like this is more consistent with what we do at `aggregatedAccountReportsDao`

ambry-mysql/src/main/java/com/github/ambry/accountstats/PartitionClassReportsDao.java
```diff
@@ -98,6 +98,18 @@
...
 
+  /**
+   * eg : DELETE FROM AggregatedPartitionClassReports
+   *      WHERE
+   *      PartitionClassId IN (SELECT id FROM PartitionClassNames WHERE clusterName = "ambry-test" AND name = "default")
+   *      AND accountId = 101
+   *      AND containerId = 201;
+   */
+  private static final String deleteAggregatedSql =
+      String.format("DELETE FROM %s WHERE %s IN (SELECT %s FROM %s WHERE %s=? AND %s=?) AND %s=? AND %s=?",
```
jsjtzyy, line 109: Just to confirm, in `PartitionClassNames` table, if two partition classes have same name but in different clusters, they will have two different ids, correct?

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -509,7 +509,7 @@ public void run() {
           Iterator<PartitionId> iterator = partitionToReplicaMap.keySet().iterator();
           while (!cancelled && iterator.hasNext()) {
             PartitionId partitionId = iterator.next();
-            logger.info("Aggregating account stats for local report started for store {}", partitionId);
+            logger.trace("Aggregating account stats for local report started for store {}", partitionId);
```
jsjtzyy, line 512: good change

jsjtzyy (commented):
Mostly looks good. Will approve after comments are addressed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -161,6 +172,69 @@ public TaskResult run() {
...
+        if (!partitionClassNameAndAccountContainerExistsInStatsSnapshot(currentStats, partitionClassName,
+            accountContainerKey)) {
+          toBeDeletedPartitionClassNameAndAccountContainer.add(new Pair<>(partitionClassName, accountContainerKey));
+        }
+      }
+    }
+    for (Pair<String, String> pair : toBeDeletedPartitionClassNameAndAccountContainer) {
+      accountStatsStore.deleteAggregatedPartitionClassStatsForAccountContainer(pair.getFirst(), pair.getSecond());
+    }
+    return !toBeDeletedPartitionClassNameAndAccountContainer.isEmpty();
```
jsjtzyy, line 210: same here

ambry-mysql/src/main/java/com/github/ambry/accountstats/PartitionClassReportsDao.java
justinlin-linkedin: that's right, the `PartitionClassNames` has a auto incremented id and (clusterName, partitionClassName) form an unique key in the table.

ambry-api/src/main/java/com/github/ambry/accountstats/AccountStatsStore.java
justinlin-linkedin: I think stick to `deleteSnapshotOfAggregatedAccountStats` since the "write" method is `takeSnapshotOfAggregatedAccountStatsAndUpdateMonth`. If we ignore the implementation, then this interface means that we can somehow take a snapshot of current aggregated account stats and attach a "month" tag with this snapshot. We only have one snapshot, if we move to next month, the snapshot will be replaced with new data and new "month" tag. So here, we just remove the snapshot.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
justinlin-linkedin: why do we prefer `parseShort` over `valueOf`?

jsjtzyy: `parseShort` returns a primitive `short` while `valueOf` returns `Short` object which has an extra unboxing internally.

jsjtzyy (approved):

ankagrawal (approved):
LGTM!

PR #1837, opened 2021-05-06
[build] Enable gradle caching and use ubuntu 20.04

cgtz:
This commit will enable gradle caching. Enabling caching means that
tests only run for a module if there are changes in that module or a
dependency. This can speed up builds for many small changes and makes
reruns because of a single test failure much faster, but does not give
us as much insight into flaky tests or potentially hidden dependencies.

Also, change the codeCoverageReport to always run since this task does
not explicitly declare its dependencies (it creates a report for
whatever test cases were run in the build but not necessarily all test
cases).

Finally, this switches our travis build to ubuntu 20.04 (focal). This
means that mysql 8 is in the default repos and we no longer have to
install it from an external repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: java
-dist: xenial
-sudo: required
+dist: focal
 jdk:
   - openjdk11
 before_cache:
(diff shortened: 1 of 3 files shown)


Arun-LinkedIn (approved):

ankagrawal (approved):
LGTM

PR #1840, opened 2021-05-07
Sample workflow to use github-actions-ci 

Arun-LinkedIn:
Starter workflow to build and run tests on Github-Actions CI.


cgtz:
Nice! It looks like we can get all of the tests to pass on GH actions. Thank you for figuring out the mysql setup. I will also check if GH actions was able to upload coverage reports in the most recent build.

cgtz (approved):

PR #1841, opened 2021-05-07
Add API to decode signed url parameters

justinlin-linkedin:
Add a new API to return the decoded signed url parameters.


jsjtzyy:
Have we decided whether to add this interface to frontend or use a script to decode signed url?  (I took a cursory look and feel like adding decodeSignedUrl handler is a bit of heavy).

justinlin-linkedin:
Going to create script to cli tool for this task.

PR #1844, opened 2021-05-12
Auto close last logsegment during compaction

SophieGuo410:
This pr support close last log segment before compact all policy, journal need to be clean up right after it.


ambry-store/src/main/java/com/github/ambry/store/Journal.java
```diff
@@ -117,6 +117,24 @@ void addEntry(Offset offset, StoreKey key, Long crc) {
     }
   }
 
+  /**
+   * This method should only be called when replica gets sealed. Remove all entries belongs to the auto closed log segment.
+   * It will delay the addEntry method for sealed replica, so the replication performance will be hold while running this method.
+   * @param activeLogSegmentName journal should remove all entries not belongs to active log segment.
+   */
+  synchronized void cleanUpJournal(LogSegmentName activeLogSegmentName) {
+    while (!inBootstrapMode) {
```
zzmao, line 126: If `inBootstrapMode` is False, the loop will eat a lot of CPU. Can we do sleep wait?

SophieGuo410: Actually I have other condition to break the loop. To be more clear, I move that condition inside the while loop check, also, I'm not sure for the performance which way is better. 1. each time when we remove the entry, we release the lock.
2. release the lock wait until the cleanup journal finished. Since it's for sealed replica only, I'm planning to use the config to switch between the two approach test the performance.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1227,6 +1245,14 @@ void compact(CompactionDetails details, byte[] bundleReadBuffer) throws IOExcept
     logger.info("One cycle of compaction is completed on the store {}", storeId);
   }
 
+  /**
+   * Closes the last log segment periodically if replica is in sealed status.
```
jsjtzyy, line 1249: Could you add comment describing how this period is determined?

jsjtzyy (commented):
some initial comments, will continue to review rest of files

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -325,6 +325,30 @@
...
+  @Default("false")
+  public final boolean storeAutoCloseLastLogSegmentEnabled;
+  public static final String storeAutoCloseLastLogSegmentEnabledName = "store.auto.close.last.log.segment.enabled";
+
+  /**
+   * Specifies the minimum value for maxLagForPartition which can unseal replica.
+   */
+  @Config(storeUnsealReplicaMinimumLagName)
+  @Default("10485760")
+  public final long storeUnsealReplicaMinimumLag;
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -469,6 +493,9 @@ public StoreConfig(VerifiableProperties verifiableProperties) {
         verifiableProperties.getIntInRange("store.io.error.count.to.trigger.shutdown", Integer.MAX_VALUE, 1,
             Integer.MAX_VALUE);
     storeSetFilePermissionEnabled = verifiableProperties.getBoolean("store.set.file.permission.enabled", false);
+    storeAutoCloseLastLogSegmentEnabled = verifiableProperties.getBoolean("store.auto.close.last.log.segment.enabled", false);
```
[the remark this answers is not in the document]
SophieGuo410: Done.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -454,9 +459,10 @@ public void addLagMetricForPartition(PartitionId partitionId) {
     if (!partitionLags.containsKey(partitionId)) {
       partitionLags.put(partitionId, new HashMap<>());
       // Set up metrics if and only if no mapping for this partition before.
-      Gauge<Long> replicaLag = () -> getMaxLagForPartition(partitionId);
-      registry.register(MetricRegistry.name(ReplicaThread.class,
-          String.format(MAX_LAG_FROM_PEERS_IN_BYTE_METRIC_NAME_TEMPLATE, partitionId.toPathString())), replicaLag);
+      if (replicationConfig.replicationTrackPerPartitionLagFromRemote) {
+        Gauge<Long> replicaLag = () -> getMaxLagForPartition(partitionId);
+        registry.register(MetricRegistry.name(ReplicaThread.class, String.format(MAX_LAG_FROM_PEERS_IN_BYTE_METRIC_NAME_TEMPLATE, partitionId.toPathString())), replicaLag);
+      }
```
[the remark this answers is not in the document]
SophieGuo410: Update.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -87,7 +87,7 @@
   private final RemoteTokenTracker remoteTokenTracker;
   private final AtomicInteger errorCount;
   private final AccountService accountService;
-
+  private long maxLagForPartition;
```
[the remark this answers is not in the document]
SophieGuo410: Renamed.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
SophieGuo410: Added the comments. And later I'm planning to update this part since currently we are leveraging counter and timestamp to decided when should we running compactAllPolicy for hybridCompactionPolicy. Later we might speed up the frequency of compaction for main and video cluster, so counter frequency needs to be tuned. So the best way is only leveraging timestamp for the hybridCompactionPolicy.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -335,6 +335,18 @@ public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> stor
...
 
+  ReplicaId getReplicaId() {
+    return this.replicaId;
+  }
+
+  /**
+   * Set max lag for blob store.
+   * @param maxLagForPartition the partition's maximum lag between local and its remote replicas.
+   */
+  public void setMaxLagForPartition(long maxLagForPartition) {
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -325,6 +325,30 @@
...
+  @Config(storeAutoCloseLastLogSegmentEnabledName)
+  @Default("false")
+  public final boolean storeAutoCloseLastLogSegmentEnabled;
+  public static final String storeAutoCloseLastLogSegmentEnabledName = "store.auto.close.last.log.segment.enabled";
+
+  /**
+   * Specifies the minimum value for maxLagForPartition which can unseal replica.
+   */
+  @Config(storeUnsealReplicaMinimumLagName)
+  @Default("10485760")
```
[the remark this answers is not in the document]
SophieGuo410: Actually after a second thought, I think maybe it's better to set to 0. Since I checked the metric and most of time the lag is zero after it fully catch up with it's peer replica after token gets reset.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -200,7 +200,25 @@ public int compare(IndexEntry e1, IndexEntry e2) {
...
+            LogSegmentName logSegmentName =
+                IndexSegment.getIndexSegmentStartOffset(indexFiles.get(i).getName()).getName();
+            LogSegmentName lastLogSegmentName = log.getLastSegment().getName();
+            if (logSegmentName.compareTo(lastLogSegmentName) < 0) {
+              sealed = true;
+            } else {
+              sealed = false;
+            }
+          } else {
+            sealed = i < indexFiles.size() - 1;
```
zzmao, line 217: `i < indexFiles.size() - 1` is always True in this case, no?

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -325,6 +325,30 @@
...
+  @Config(storeUnsealReplicaMinimumLagBytesName)
+  @Default("0")
+  public final long storeUnsealReplicaMinimumLagBytes;
+  public static final String storeUnsealReplicaMinimumLagBytesName = "store.unseal.replica.minimum.lag.bytes";
+
+
+  /**
+   * Whether to enable add synchronizer inside while loop for journal clean up.
+   */
+  @Config("store.synchronizer.inside.while.loop.enabled")
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -241,9 +241,7 @@ public boolean removeReplica(ReplicaId replicaId) {
       replicationMetrics.addMetricsForRemoteReplicaInfo(remoteReplicaInfo, trackPerPartitionLagInMetric);
       remoteReplicaInfos.add(remoteReplicaInfo);
     }
-    if (replicationConfig.replicationTrackPerPartitionLagFromRemote) {
-      replicationMetrics.addLagMetricForPartition(partition);
-    }
+    replicationMetrics.addLagMetricForPartition(partition);
```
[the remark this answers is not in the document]
SophieGuo410: Actually the config is inside the addLagMetricForPartition method, cause I need to get the lag info, but no need to emitting to ingraph.

ambry-store/src/main/java/com/github/ambry/store/Journal.java
```diff
@@ -117,6 +117,39 @@ void addEntry(Offset offset, StoreKey key, Long crc) {
...
+        && earliestEntry.getKey().getName().compareTo(activeLogSegmentName) < 0) {
+      journal.remove(earliestEntry.getKey());
+      recentCrcs.remove(earliestEntry.getValue());
+      currentNumberOfEntries.decrementAndGet();
+      earliestEntry = journal.firstEntry();
+    }
+  }
+
+  /**
+   * This method is used to test the impact of performance between release lock after journal finish clean up and release lock each time entry gets removed.
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Yeah, Updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
```diff
@@ -288,6 +288,9 @@ public void run() {
                     metrics.markCompactionStart(true);
                     compactionStarted = true;
                     store.compact(details, bundleReadBuffer);
+                    if (store.getReplicaId().isSealed()) {
```
zzmao, line 291: why not add config `storeAutoCloseLastLogSegmentEnabled` check in this level(top level) to control auto close?

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -439,14 +450,72 @@ private void free(LogSegment logSegment, boolean isSwapSegment) throws StoreExce
...
+  void autoCloseLastLogSegmentIfQualified(Journal journal) throws StoreException {
+    if (compactionPolicySwitchInfoCounterValueReached()) {
+      //ensure the capacity to open the new log segment and allocate new log segment.
+      //if not able to close the last log segment, continue running the compaction.
+      try {
+        ensureCapacity(0);
+        setActiveSegment();
+        //refresh journal.
+        if (config.storeSynchronizerInsideWhileLoopEnabled) {
+          journal.cleanUpJournalWithSynchronizeInside(activeSegment.getName());
```
zzmao, line 471: Remember to remove one the methods once you figure out which one is efficient. 
(Just want to avoid repeating code.)

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -636,6 +635,11 @@ public void replicate() {
 
                 replicationMetrics.updateLagMetricForRemoteReplica(remoteReplicaInfo,
                     exchangeMetadataResponse.localLagFromRemoteInBytes);
+                long maxLag =
+                    replicationMetrics.getMaxLagForPartition(remoteReplicaInfo.getReplicaId().getPartitionId());
```
jsjtzyy, line 639: `maxLag` can be -1 if the `perDataNodeLag` hasn't been populated in ReplicationMetrics, is this OK?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -87,7 +87,8 @@
   private final RemoteTokenTracker remoteTokenTracker;
   private final AtomicInteger errorCount;
   private final AccountService accountService;
-
+  //The max replication lag indicating how far local store is behind peers.
+  private long localStoreMaxLagFromPeer;
```
jsjtzyy, line 91: make this `volatile`, as it will be updated and accessed by different threads

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -325,6 +325,31 @@
...
+  /**
+   * Whether to enable auto close last log segment during compaction.
+   */
+  @Config(storeAutoCloseLastLogSegmentEnabledName)
+  @Default("false")
+  public final boolean storeAutoCloseLastLogSegmentEnabled;
+  public static final String storeAutoCloseLastLogSegmentEnabledName = "store.auto.close.last.log.segment.enabled";
+
+  /**
+   * Specifies the minimum value of Bytes for maxLagForPartition which can unseal replica.
```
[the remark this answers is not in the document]
SophieGuo410: Updated.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -325,6 +325,31 @@
...
+   */
+  @Config(storeUnsealReplicaMinimumLagBytesName)
+  @Default("0")
+  public final long storeUnsealReplicaMinimumLagBytes;
+  public static final String storeUnsealReplicaMinimumLagBytesName = "store.unseal.replica.minimum.lag.bytes";
+
+
+  /**
+   * Whether to enable add synchronizer inside while loop for journal clean up.
```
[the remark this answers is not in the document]
SophieGuo410: Updated the description.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
SophieGuo410: Thx for catching up, updated.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
```diff
@@ -288,6 +288,9 @@ public void run() {
                     metrics.markCompactionStart(true);
                     compactionStarted = true;
                     store.compact(details, bundleReadBuffer);
+                    if (storeConfig.storeAutoCloseLastLogSegmentEnabled && store.getReplicaId().isSealed()) {
+                      store.closeLastLogSegmentIfQualified();
+                    }
```
jsjtzyy, line 293: This comes after `store.compact()` because it's intended for next round of compaction? 
I am trying to understand if the ordering matters.

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -439,14 +451,75 @@ private void free(LogSegment logSegment, boolean isSwapSegment) throws StoreExce
...
+        } else {
+          journal.cleanUpJournal(activeSegment.getName());
+        }
+        logger.debug("Time to clean up journal size for store : {} in dataDir: {} is {} ms", storeId, dataDir,
+            SystemTime.getInstance().milliseconds() - startTime);
+      } catch (IllegalStateException e) {
+        //if there is no more capacity left in data dir to create new log segment, just bypass this logic.
+        //no-op
```
jsjtzyy, line 481: add a log message here for debugging (can be `INFO` level)

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -439,14 +451,75 @@ private void free(LogSegment logSegment, boolean isSwapSegment) throws StoreExce
...
+      //if not able to close the last log segment, continue running the compaction.
+      try {
+        ensureCapacity(0);
+        setActiveSegment();
+        //refresh journal.
+        long startTime = SystemTime.getInstance().milliseconds();
+        if (config.storeSynchronizerInsideWhileLoopEnabled) {
+          journal.cleanUpJournalWithSynchronizeInside(activeSegment.getName());
+        } else {
+          journal.cleanUpJournal(activeSegment.getName());
```
jsjtzyy, line 475: Any specific reason that the journal is cleaned up in `Log.java` ?  I feel like it can be done in `PersistentIndex` and thus we don't have to pass the journal variable into this method.

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -439,14 +451,75 @@ private void free(LogSegment logSegment, boolean isSwapSegment) throws StoreExce
...
+  }
+
+  /**
+   * Set the last log segment as active segment and init buffer for append.
+   * @throws StoreException
+   */
+  private void setActiveSegment() throws StoreException {
+    // this cannot be null since capacity has either been ensured or has thrown.
+    LogSegment nextActiveSegment = segmentsByName.higherEntry(activeSegment.getName()).getValue();
+    logger.info("Auto close last log segment: {} and create new active segment : {}", activeSegment.getName(),
```
jsjtzyy, line 493: This method could be called by a regular `rollOverIfRequired()` as well, the info message here will be confusing in that case.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -385,7 +394,7 @@ private void checkSanity(CompactionDetails details) {
       }
       // should be outside the range of the journal
       Offset segmentEndOffset = new Offset(segment.getName(), segment.getEndOffset());
-      if (segmentEndOffset.compareTo(srcIndex.journal.getFirstOffset()) >= 0) {
+      if (srcIndex.journal.getFirstOffset() != null && segmentEndOffset.compareTo(srcIndex.journal.getFirstOffset()) >= 0) {
```
jsjtzyy, line 397: good check

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -439,14 +451,75 @@ private void free(LogSegment logSegment, boolean isSwapSegment) throws StoreExce
...
+      }
+      if (config.storeCompactionPolicySwitchCounterDays - 1 == (compactionPolicySwitchInfo.getCompactionPolicyCounter()
+          .getCounter())) {
+        logger.trace("The compaction policy switch counter value is qualified for closing last log segment.");
+        return true;
+      } else {
+        return false;
+      }
+    } else {
+      return false;
```
jsjtzyy, line 522: add a log message here

ambry-store/src/main/java/com/github/ambry/store/Journal.java
```diff
@@ -117,6 +117,40 @@ void addEntry(Offset offset, StoreKey key, Long crc) {
...
+        && earliestEntry.getKey().getName().compareTo(activeLogSegmentName) < 0) {
+      synchronized (this) {
+        journal.remove(earliestEntry.getKey());
+        recentCrcs.remove(earliestEntry.getValue());
+        currentNumberOfEntries.decrementAndGet();
+        earliestEntry = journal.firstEntry();
+      }
+    }
+  }
+
```
jsjtzyy, line 153: For either `cleanUpJournal` or `cleanUpJournalWithSynchronizeInside`, how can we guarantee the correctness when `getEntriesSince()` is concurrently called by a metadata exchange request?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -260,6 +267,28 @@ public int compare(IndexEntry e1, IndexEntry e2) {
...
 
+  /**
+   *
+   * @param index the index of the list of indexFiles.
+   * @param indexFiles a list of file with index entry info.
+   * @return {@code true} if the index file needs to be set to seal status.
+   */
+  private boolean defineSealStatusForIndexFiles(int index, List<File> indexFiles) {
+    return index < indexFiles.size() - 1 || (config.storeAutoCloseLastLogSegmentEnabled
+        && indexFileNotInLastLogSegment(index, indexFiles));
```
jsjtzyy, line 278: I wonder when will the code reach the second condition (`(config.storeAutoCloseLastLogSegmentEnabled
        && indexFileNotInLastLogSegment(index, indexFiles)`)?

jsjtzyy (commented):
a few more comments

ambry-store/src/main/java/com/github/ambry/store/Log.java
SophieGuo410: Updated.

ambry-store/src/main/java/com/github/ambry/store/Log.java
SophieGuo410: Updated to move it to BlobStore.

ambry-store/src/main/java/com/github/ambry/store/Log.java
SophieGuo410: Thx for catching that. Added both logs.

ambry-store/src/main/java/com/github/ambry/store/Log.java
SophieGuo410: done.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: when the last log segment get auto closed and the last index file is not allocated in last log segment, the index file needs to be sealed.

ambry-store/src/main/java/com/github/ambry/store/CompactionManager.java
SophieGuo410: done.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
SophieGuo410: I think it's ok. We only need to get max lag from peer replica when local replica in the sealed status during compaction. The lag will be updated in partitionLags when exchange metadata. So I think when the local replica is in sealed status, it should finish exchange metadata for several rounds. And the edge case you mentioned when some peer replica gets closed, it won't affect the local replica since the max lag(from its other peer replica) will determine if it's fully replicated.

ambry-store/src/main/java/com/github/ambry/store/Journal.java
SophieGuo410: Checked the logic. The getEntriesSince has some protection logic to prevent synchronizing the addEntry method. And I think it also works for cleanUpJournal. if the offset in input is not in journal, it will return null. Which is similar like when Journal size is cross the boundary and the entry mapping to the offset has been removed. So the edge case would be what if the journal size is zero. Inside the getEntriesSince there are logic check if the journal contains the offset at the beginning, so I think it should be good as well.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
```diff
@@ -2837,14 +2876,71 @@ private long reduceValidDataSizeInLogSegments(List<LogSegmentName> logSegmentsTo
...
+      LogSegmentName activeSegmentName = state.index.log.getLastSegment().getName();
+      state.index.journal.cleanUpJournal(activeSegmentName);
+    }
+    compactor.closeLastLogSegmentIfQualified();
+    if (!changeExpected) {
+      assertEquals("Journal size should be cleaned up after last log segment closed",0, state.index.journal.getCurrentNumberOfEntries());
+    }
+
+    long logSegmentSizeAfterCompaction = getSumOfLogSegmentEndOffsets();
+    long logSegmentCountAfterCompaction = state.index.getLogSegmentCount();
```
zzmao, line 2913: Check logSegmentCountAfterCompaction and indexSegmentCountAfterCompaction?

ambry-store/src/main/java/com/github/ambry/store/Journal.java
```diff
@@ -117,6 +117,22 @@ void addEntry(Offset offset, StoreKey key, Long crc) {
...
+   */
+  void cleanUpJournal(LogSegmentName activeLogSegmentName) {
+    Map.Entry<Offset, StoreKey> earliestEntry = journal.firstEntry();
+    while (!inBootstrapMode && earliestEntry != null
+        && earliestEntry.getKey().getName().compareTo(activeLogSegmentName) < 0) {
+      journal.remove(earliestEntry.getKey());
+      recentCrcs.remove(earliestEntry.getValue());
+      currentNumberOfEntries.decrementAndGet();
+      earliestEntry = journal.firstEntry();
+    }
```
jsjtzyy, line 133: If we use `storeWriteLock` to ensure deletes/updates are blocked while we close the last log segment, do we still need this "while" loop?  Can we directly call `journal.clear()` and `recentCrcs.clear()` and  set `currentNumberOfEntries` to 0?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -99,7 +99,7 @@ public int compare(IndexEntry e1, IndexEntry e2) {
   final HardDeleter hardDeleter;
   final Thread hardDeleteThread;
 
-  private final Log log;
+  protected final Log log;
```
jsjtzyy, line 102: May I ask why this is `protected`?

SophieGuo410: Because previous in BlobStoreCompactorTest I need to call state.index.log. But since you suggest for clean up journal we can directly clear the Hashmap, so there's no need to bring in the active log segment name. So after I tested, I will change it back.

ambry-store/src/main/java/com/github/ambry/store/Journal.java
SophieGuo410: You are right, updated the corresponding logic.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
SophieGuo410: Added the check for logSegmentCountAfterCompaction, and I removed indexSegmentCountAfterCompaction check since there's no need to check it.

zzmao (approved):
LGTM

jsjtzyy (approved):

PR #1845, opened 2021-05-13
[Compaction] Fix blob store stats to return 0 value for the invalid log segment

justinlin-linkedin:
This PR fixes one crucial bug for compaction.
When using StatsBasedCompactionStrategy, we would load all the log segments stats from BlobStoreStats, and figure out the cost benefit of compact various log segment. However, if a log segment doesn't have any valid index values, it would not be included in the map returned to StatsBasedCompactionStrategy, thus we might ignore this log segment, and decide to compact the log segments before and after it. This is a huge problem, since the log segments under compaction have to be consecutive.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -76,7 +76,6 @@ class BlobStoreStats implements StoreStats, Closeable {
   private final long logSegmentForecastOffsetMs;
   private final long waitTimeoutInSecs;
   private final boolean enableBucketForLogSegmentReports;
-  private final boolean fillupLogSegmentNames;
   private final StoreMetrics metrics;
   private final ReentrantLock scanLock = new ReentrantLock();
   private final Condition waitCondition = scanLock.newCondition();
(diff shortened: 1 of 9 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -141,6 +142,16 @@ static StatsSnapshot convertStoreUsageToContainerStatsSnapshot(Map<Short, Map<Sh
       long logSegmentForecastOffsetMs, long queueProcessingPeriodInMs, long waitTimeoutInSecs,
       boolean enableBucketForLogSegmentReports, Time time, ScheduledExecutorService longLiveTaskScheduler,
       ScheduledExecutorService shortLiveTaskScheduler, DiskIOScheduler diskIOScheduler, StoreMetrics metrics) {
+    this(storeId, index, bucketCount, bucketSpanTimeInMs, logSegmentForecastOffsetMs, queueProcessingPeriodInMs,
+        waitTimeoutInSecs, enableBucketForLogSegmentReports, true, time, longLiveTaskScheduler, shortLiveTaskScheduler,
+        diskIOScheduler, metrics);
+  }
+
+  BlobStoreStats(String storeId, PersistentIndex index, int bucketCount, long bucketSpanTimeInMs,
```
zzmao, line 150: Looks `fillupLogSegmentNames` = false is only used in test? 

If test fails without your fix, and test passes with your fix. That's good enough. I don' think we need to test a wrong case by adding extra parameters.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -335,6 +347,13 @@ long getMaxBlobSize() {
       retValue =
           new Pair<>(referenceTimeInMs, collectValidDataSizeByLogSegment(referenceTimeInMs, expiryReferenceTime));
     }
+    if (fillupLogSegmentNames) {
+      // Before return the value, make sure all the log segments are in the final map
+      for (IndexSegment indexSegment : index.getIndexSegments().descendingMap().values()) {
```
jsjtzyy, line 352: `index` has access to `log`,  and `log` has a map `segmentsByName`. Maybe we can expose some API in PersistentIndex to get all current log segment names. Thus, we don't have to go through all index segments (total count > log segment count).

jsjtzyy (commented):
Mostly LGTM, will approve after comments are addressed.

jsjtzyy (approved):

zzmao (approved):
LGTM

PR #1846, opened 2021-05-14
Add ResponseBytesSent to PublicAccessLogHandler

jsjtzyy:
Add ResponseBytesSent to PublicAccessLogHandler to track how many bytes have been sent out so far.

### DIFF
--- a/ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
+++ b/ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
@@ -42,6 +42,7 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
   private long requestArrivalTimeInMs;
   private long requestLastChunkArrivalTimeInMs;
   private long responseFirstChunkStartTimeInMs;
+  private long responseBytesSent;
   private StringBuilder logMessage;
   private HttpRequest request;
   private StringBuilder sslLogMessage;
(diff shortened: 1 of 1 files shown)


ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -158,6 +163,7 @@ private void logHeaders(String tag, HttpMessage message, String[] headers) {
    */
   private void logDurations() {
     long nowMs = System.currentTimeMillis();
+    logMessage.append("bytes sent=").append(responseBytesSent).append(", ");
```
cgtz, line 166: for some reason, this section does not separate using commas, only spaces, so we can remove the comma to keep it consistent.

In the long run, we should probably take another look at the log message format to make it easier to parse programatically.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -107,6 +108,10 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
         logger.error(
             "Sending response that is not of type HttpResponse or HttpContent. Sending response to {}. Request is of type {}. No action being taken other than logging this unexpected state.",
             ctx.channel().remoteAddress(), msg.getClass());
+      } else {
```
cgtz, line 111: this may not catch the (rarer) case of use `FullHttpResponse` which implements both `HttpResponse` and `HttpContent`.

What if we changed it to this:
```
boolean isHttpContent = msg instanceof HttpContent;
if (msg instanceof HttpResponse) {
  ...
} else if (!isHttpContent) {
 log
}
if (isHttpContent) {
  countBytes
}
```

cgtz (approved):
LGTM other than these comments

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
jsjtzyy: sure

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
jsjtzyy: Yeah, good point. Will take your suggestion.

PR #1849, opened 2021-05-19
Add log to show chunk retrieved and written when GetBlobOperationError

zzmao:
Add log to show chunk retrieved and written when GetBlobOperationError happens


ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -512,6 +516,10 @@ void completeRead() {
         if (e == null) {
           updateChunkingAndSizeMetricsOnSuccessfulGet();
         } else {
+          logger.warn(
```
cgtz, line 519: Do we need to log `e` as well? The blob ID may also be useful to log in order to correlate with the public access log.

zzmao:
Addressed @cgtz comment. Ready for review @jsjtzyy

jsjtzyy (approved):
LGTM with one minor comment

cgtz (approved):

PR #1852, opened 2021-05-26
Support async get valid data size

SophieGuo410:
While debugging the dotted metric issue, we found that one piece of code using gauge takes a really long time(5 to 10 s) to get the size of valid data at a particular point in time for all log segments. And the reason this issue only happens in the main cluster is we have many small blobs and many partitions in the main cluster are almost full. This pr support to calculate the valid data size in a background thread that won't block metrics collection.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -241,13 +241,6 @@ public class StoreConfig {
   public final long storeStatsRecentEntryProcessingIntervalInMinutes;
 
   /**
-   * Enable asynchronously get valid data size for metric emitting.
-   */
-  @Config("store.async.get.valid.size.enable")
-  @Default("false")
-  public final boolean storeAsyncGetValidSizeEnable;
-
-  /**
    * Period in minutes to specify how frequent the validDataSizeCollector executed.
    */
   @Config("store.get.valid.size.interval.in.minutes")
@@ -526,7 +519,6 @@ public class StoreConfig {
         verifiableProperties.getLongInRange("store.stats.bucket.span.in.minutes", 60, 1, 10000);
     storeStatsRecentEntryProcessingIntervalInMinutes =
         verifiableProperties.getLongInRange("store.stats.recent.entry.processing.interval.in.minutes", 2, 1, 60);
-    storeAsyncGetValidSizeEnable = verifiableProperties.getBoolean("store.async.get.valid.size.enable", false);
     storeGetValidSizeIntervalInMinutes =
         verifiableProperties.getLongInRange("store.get.valid.size.interval.in.minutes", 2, 1, 60);
     storeStatsWaitTimeoutInSecs =
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -241,6 +241,20 @@ public class StoreConfig {
   public final long storeStatsRecentEntryProcessingIntervalInMinutes;
 
   /**
+   * Enable asynchronously get valid data size for metric emitting.
+   */
+  @Config("store.async.get.valid.size.enable")
+  @Default("false")
+  public final boolean storeAsyncGetValidSizeEnable;
+
+  /**
+   * Period in minutes to specify how frequent the validDataSizeCollector executed.
+   */
+  @Config("store.get.valid.size.interval.in.minutes")
+  @Default("2")
+  public final long storeGetValidSizeIntervalInMinutes;
+
+  /**
    * The upper limit in seconds for requests to wait for a ongoing construction of buckets (that contains the answer)
    * to complete.
    */
@@ -512,6 +526,9 @@ public class StoreConfig {
         verifiableProperties.getLongInRange("store.stats.bucket.span.in.minutes", 60, 1, 10000);
     storeStatsRecentEntryProcessingIntervalInMinutes =
         verifiableProperties.getLongInRange("store.stats.recent.entry.processing.interval.in.minutes", 2, 1, 60);
+    storeAsyncGetValidSizeEnable = verifiableProperties.getBoolean("store.async.get.valid.size.enable", false);
+    storeGetValidSizeIntervalInMinutes =
+        verifiableProperties.getLongInRange("store.get.valid.size.interval.in.minutes", 2, 1, 60);
     storeStatsWaitTimeoutInSecs =
         verifiableProperties.getLongInRange("store.stats.wait.timeout.in.secs", 2 * 60, 0, 30 * 60);
     storeStatsIndexEntriesPerSecond =
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -241,11 +241,11 @@ public class StoreConfig {
   public final long storeStatsRecentEntryProcessingIntervalInMinutes;
 
   /**
-   * Period in minutes to specify how frequent the validDataSizeCollector executed.
+   * Period in seconds to specify how frequent the validDataSizeCollector executed.
    */
-  @Config("store.get.valid.size.interval.in.minutes")
-  @Default("2")
-  public final long storeGetValidSizeIntervalInMinutes;
+  @Config("store.get.valid.size.interval.in.secs")
+  @Default("2*60")
+  public final long storeGetValidSizeIntervalInSecs;
 
   /**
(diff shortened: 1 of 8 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -160,6 +164,12 @@ static StatsSnapshot convertStoreUsageToContainerStatsSnapshot(Map<Short, Map<Sh
       queueProcessor = new QueueProcessor();
       shortLiveTaskScheduler.scheduleAtFixedRate(queueProcessor, 0, queueProcessingPeriodInMs, TimeUnit.MILLISECONDS);
     }
+    if (enableAsyncGetValidSize) {
+      ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor();
```
cgtz, line 168: Here we are using a single thread per each blob store. This could be a lot of extra threads that will be mostly idle. Could we reuse the shared `longLiveTaskScheduler` or `shortLiveTaskScheduler`? (not exactly sure when a task is considered to be short vs long, but we can look into that.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -160,6 +164,12 @@ static StatsSnapshot convertStoreUsageToContainerStatsSnapshot(Map<Short, Map<Sh
...
+      TimeRange timeRange = new TimeRange(System.currentTimeMillis(), getBucketSpanTimeInMs());
```
cgtz, line 169: Should this time range be a constant or should it be refreshed to the current time whenever ValidDataSizeCollector.run is called?

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -253,8 +253,12 @@ void initializeIndexGauges(String storeId, final PersistentIndex index, final lo
       Gauge<Long> currentInvalidDataSize = () -> {
         long invalidDataSize = -1;
         try {
-          invalidDataSize = index.getLogUsedCapacity() - blobStoreStats.getValidSize(
-              new TimeRange(System.currentTimeMillis(), blobStoreStats.getBucketSpanTimeInMs())).getSecond();
+          if (storeAsyncGetValidSizeEnable) {
+            invalidDataSize = index.getLogUsedCapacity() - blobStoreStats.getValidSizeAsync().getSecond();
+          } else {
```
zzmao, line 258: The else case can be fully removed since it's not workable. 

We can also remove `storeAsyncGetValidSizeEnable`. Once `enableStoreCurrentInvalidSize` is enabled, your solution is the only solution.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -175,6 +185,13 @@ static StatsSnapshot convertStoreUsageToContainerStatsSnapshot(Map<Short, Map<Sh
     return new Pair<>(logSegmentValidSizeResult.getFirst(), totalValidSize);
   }
 
+  /**
+   * Get valid data size asynchronously.
+   */
+  Pair<Long, Long> getValidSizeAsync() {
```
zzmao, line 191: This is not async. Name it to `getCachedValidSize`?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
SophieGuo410: Thanks for catching that, fixed it.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
SophieGuo410: Currently the non-main clusters are using the previous way, so what I'm planning to do is test on main cluster first, so the logic won't affect other clusters. And after verification, I'll remove the redundant logic. What do you think?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
SophieGuo410: Renamed.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
SophieGuo410: I used the shortLiveTaskScheduler for my case. longLiveTaskScheduler only have 1 thread in thread pool it's running less frequently.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
zzmao: Though video and delivery cluster are working, I don't think previous solution is efficient. We will eventually use your solution in this PR. So I suggest to remove the `else` case in this PR directly.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -148,14 +150,13 @@ static StatsSnapshot convertStoreUsageToContainerStatsSnapshot(Map<Short, Map<Sh
         TimeUnit.MINUTES.toMillis(config.storeStatsRecentEntryProcessingIntervalInMinutes),
         config.storeStatsWaitTimeoutInSecs, config.storeEnableBucketForLogSegmentReports,
         config.storeCompactionPurgeDeleteTombstone, time, longLiveTaskScheduler, shortLiveTaskScheduler,
-        diskIOScheduler, metrics);
+        diskIOScheduler, metrics, TimeUnit.MINUTES.toMillis(config.storeGetValidSizeIntervalInMinutes));
```
zzmao, line 153: Formatting. This line is too long.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -240,6 +240,13 @@
   @Default("2")
   public final long storeStatsRecentEntryProcessingIntervalInMinutes;
 
+  /**
+   * Period in minutes to specify how frequent the validDataSizeCollector executed.
+   */
+  @Config("store.get.valid.size.interval.in.minutes")
+  @Default("2")
+  public final long storeGetValidSizeIntervalInMinutes;
```
zzmao, line 248: Change its time unit to second in case we want to have a interval less than 1 minute?

zzmao (approved):
Minor comments. LGTM

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
SophieGuo410: This line is not exceed the right margin. Do you mean the constructer after this one? I've formatted that one.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
SophieGuo410: Done

ambry-store/src/test/java/com/github/ambry/store/BlobStoreStatsTest.java
```diff
@@ -90,7 +90,7 @@ public BlobStoreStatsTest(boolean isBucketingEnabled) throws IOException, StoreE
   private BlobStoreStats setupBlobStoreStats(int bucketCount, long logSegmentForecastOffsetMs) {
     return new BlobStoreStats("", state.index, bucketCount, BUCKET_SPAN_IN_MS, logSegmentForecastOffsetMs,
         QUEUE_PROCESSOR_PERIOD_IN_Ms, DEFAULT_WAIT_TIMEOUT_SECS, true, true, state.time, indexScannerScheduler,
-        queueProcessorScheduler, diskIOScheduler, METRICS);
+        queueProcessorScheduler, diskIOScheduler, METRICS, 1, false);
```
cgtz, line 93: Could we leave a TODO saying that currently the valid data size background job is not tested and needs changes to the logic to check the number of throttling events?

I think this is just a test issue that needs to be fixed. I think we equate calls to throttle as a sign that the indexscanner is running and has passed over a certain number of entries. But since getValidSize also iterates over entries it also adds to the throttle count. For now, we can ignore it and leave the TODO. The test may need many changes since IndexScanner is not the only thing scanning the index now.

cgtz (approved):

ambry-store/src/test/java/com/github/ambry/store/BlobStoreStatsTest.java
SophieGuo410: Sure, added the TODO on top of setupBlobStoreStats since many tests need to be fixed later.

PR #1853, opened 2021-05-26
Return failure reason back to client when data chunk is missing

justinlin-linkedin:
When data chunk is missing, return RestServiceException from getBlob and include the failure reason in the response header.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -38,6 +38,8 @@ import com.github.ambry.protocol.GetOption;
 import com.github.ambry.protocol.GetRequest;
 import com.github.ambry.protocol.GetResponse;
 import com.github.ambry.protocol.PartitionResponseInfo;
+import com.github.ambry.rest.RestServiceErrorCode;
+import com.github.ambry.rest.RestServiceException;
 import com.github.ambry.server.ServerErrorCode;
 import com.github.ambry.store.MessageInfo;
 import com.github.ambry.store.StoreKey;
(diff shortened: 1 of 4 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -505,6 +507,14 @@ private void maybeWriteToChannel() {
     void completeRead() {
       if (readIntoCallbackCalled.compareAndSet(false, true)) {
         Exception e = operationException.get();
+        if (e instanceof RouterException) {
+          RouterException routerException = (RouterException) e;
+          if (routerException.getErrorCode() == RouterErrorCode.UnexpectedInternalError) {
```
cgtz, line 512: Is there a specific reason to convert to the RestServiceException here instead of inside of FrontendRestRequestService.submitResponse (line 407-410)?

cgtz (commented):
Logic looks good, I just have a question about whether it's best to do the check here or outside of the router.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
justinlin-linkedin: No particular reason, it's just that we are dealing with RouterException here, so it's better to be dealt in router.

cgtz (approved):

build.gradle
```diff
@@ -320,6 +320,7 @@ project(':ambry-replication') {
         compile "io.dropwizard.metrics:metrics-core:$metricsVersion"
         testCompile project(':ambry-store')
         testCompile project(':ambry-test-utils')
+        testCompile project(':ambry-clustermap')
```
jsjtzyy, line 323: I wonder why we need this dependency?

jsjtzyy (approved):
LGTM  with one minor comment

build.gradle
justinlin-linkedin: removed

jsjtzyy (approved):

PR #1854, opened 2021-05-27
Support returning a snapshot of replicas in different states in ClusterChangeHandler

jsjtzyy:
Previously, ClusterChangeHandler only allows getting replicas in one state at a time. To get replicas in multiple states, it has to call getReplicaByState() multiple times in sequential manner. The replica state may change during this process which causes race condition where a replica is counted twice or a replica is omitted. This PR introduces a new method to support get a snapshot of replicas in different states. It's populated based on same view of cluster which avoids aforementioned race condition induced by state transition.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/clustermap/PartitionId.java
+++ b/ambry-api/src/main/java/com/github/ambry/clustermap/PartitionId.java
@@ -14,6 +14,8 @@
 package com.github.ambry.clustermap;
 
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 
 
 /**
@@ -41,6 +43,14 @@ public interface PartitionId extends Resource, Comparable<PartitionId> {
    * @return list of Replicas that satisfy requirement.
    */
   List<? extends ReplicaId> getReplicaIdsByState(ReplicaState state, String dcName);
+
+  /**
+   * Get replicas (grouped by required state) from specified datacenter.
+   * @param states a set of {@link ReplicaState} that replicas should match.
+   * @param dcName the name of datacenter from which the replica should come.
+   * @return a map whose key is {@link ReplicaState} and value is a list of replicas in that state.
+   */
+  Map<ReplicaState, ? extends List<? extends ReplicaId>> getReplicaIdsByStates(Set<ReplicaState> states, String dcName);
 
   /**
    * Gets the state of this PartitionId.
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
@@ -18,6 +18,8 @@ import java.io.DataInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import org.json.JSONArray;
@@ -71,6 +73,11 @@ public class AmbryPartition implements PartitionId {
   @Override
   public List<AmbryReplica> getReplicaIdsByState(ReplicaState state, String dcName) {
     return clusterManagerCallback.getReplicaIdsByState(this, state, dcName);
+  }
+
+  @Override
+  public Map<ReplicaState, List<AmbryReplica>> getReplicaIdsByStates(Set<ReplicaState> states, String dcName){
+    return clusterManagerCallback.getReplicaIdsByStates(this,  states, dcName);
   }
 
   @Override
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
@@ -17,6 +17,7 @@ import com.github.ambry.utils.Utils;
 import java.io.DataInputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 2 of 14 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterChangeHandler.java
```diff
@@ -54,4 +59,22 @@
         .map(instanceConfig -> getDataNode(instanceConfig.getInstanceName()))
         .map(dataNode -> getReplicaId(dataNode, partition.toPathString())).filter(Objects::nonNull);
   }
+
+  @Override
+  default Map<ReplicaState, List<AmbryReplica>> getSnapshotOfReplicaStates(AmbryPartition partition) {
+    String resourceName = getPartitionToResourceMap().get(partition.toPathString());
+    RoutingTableSnapshot snapshot = getRoutingTableSnapshot();
+    Map<ReplicaState, List<AmbryReplica>> replicasByState = new HashMap<>();
+    for (ReplicaState state : EnumSet.of(ReplicaState.OFFLINE, ReplicaState.BOOTSTRAP, ReplicaState.STANDBY,
```
cgtz, line 68: Just wondering why ERROR and DROPPED states aren't included here

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -702,6 +702,26 @@ public String getResourceNameForPartition(AmbryPartition partition) {
...
 
+    @Override
+    public Map<ReplicaState, List<AmbryReplica>> getReplicaIdsByStates(AmbryPartition partition,
+        Set<ReplicaState> states, String dcName) {
+      Map<ReplicaState, List<AmbryReplica>> replicasByState = new HashMap<>();
+      for (DcInfo dcInfo : dcToDcInfo.values()) {
+        String dc = dcInfo.dcName;
+        if (dcName == null || dcName.equals(dc)) {
+          Map<ReplicaState, List<AmbryReplica>> replicaStateMap =
+              dcInfo.clusterChangeHandler.getSnapshotOfReplicaStates(partition);
```
cgtz, line 713: Could potentially have this method take in a map and allowed states as an argument and update the map directly to avoid having to iterate/copy again.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/Partition.java
```diff
@@ -110,6 +112,21 @@ public Partition(long id, String partitionClass, PartitionState partitionState,
     return Collections.unmodifiableList(result);
   }
 
+  @Override
+  public Map<ReplicaState, List<ReplicaId>> getReplicaIdsByStates(Set<ReplicaState> states, String dcName) {
+    Map<ReplicaState, List<ReplicaId>> replicaByState = new HashMap<>();
+    for (ReplicaState state : states) {
+      List<ReplicaId> replicaIds = new ArrayList<>();
+      if (state == ReplicaState.STANDBY) {
```
cgtz, line 120: Why are all replicas added when the state STANDBY is requested?

ambry-api/src/main/java/com/github/ambry/clustermap/PartitionId.java
```diff
@@ -42,6 +44,14 @@
    */
   List<? extends ReplicaId> getReplicaIdsByState(ReplicaState state, String dcName);
 
+  /**
+   * Get replicas (grouped by required state) from specified datacenter.
+   * @param states a set of {@link ReplicaState} that replicas should match.
+   * @param dcName the name of datacenter from which the replica should come.
+   * @return a map whose key is {@link ReplicaState} and value is a list of replicas in that state.
+   */
+  Map<ReplicaState, ? extends List<? extends ReplicaId>> getReplicaIdsByStates(Set<ReplicaState> states, String dcName);
```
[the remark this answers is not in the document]
jsjtzyy: I actually tried to remove `? extends List` but Intellij reported below error:
```
'getReplicaIdsByStates(Set<ReplicaState>, String)' in 'com.github.ambry.clustermap.AmbryPartition' clashes with 'getReplicaIdsByStates(Set<ReplicaState>, String)' in 'com.github.ambry.clustermap.PartitionId'; attempting to use incompatible return type
```
The nested generics are a bit of tricky, we might add an ArrayList<ReplicaId> to a map, which was supposed to hold a List<ReplicaId> only. That will certainly give us trouble at runtime. That is why it's not allowed, and compiler prevents this at compile time only. More context is available in this post: https://stackoverflow.com/questions/19219635/java-hashmap-nested-generics-with-wildcards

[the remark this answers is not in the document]
cgtz: Ah, I see, this makes sense. Thanks for the explanation and the helpful link!

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
```diff
@@ -73,6 +75,11 @@
     return clusterManagerCallback.getReplicaIdsByState(this, state, dcName);
   }
 
+  @Override
+  public Map<ReplicaState, List<AmbryReplica>> getReplicaIdsByStates(Set<ReplicaState> states, String dcName){
+    return clusterManagerCallback.getReplicaIdsByStates(this,  states, dcName);
```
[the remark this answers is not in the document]
jsjtzyy: fixed

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterChangeHandler.java
jsjtzyy: DROPPED replica means the replica is fully decommissioned and data files are deleted. Hence, we don't have to include it. 
ERROR replica is put into ERROR state for some reason (i.e. disk failure or data corruption etc), the data may not be trustable and we'd better exclude it.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/Partition.java
jsjtzyy: Oh, I should add a comment there. This is because `Partition` is used by static clustermap where all replicas are assumed to be `STANDBY` and we leverage native failure detector to check their status (and backoff if needed).

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
jsjtzyy: yeah, fair point. I'll make a change

cgtz (approved):

PR #1861, opened 2021-06-09
Remove counter from Hybrid Compaction Policy

SophieGuo410:
In the future, we may speed up the frequency for running compaction, if we still leverage the counter to determine the policy switch, the counter value need to be updated correspondingly. So this pr removed the counter in Hybrid Compaction Policy and only leverage timestamp.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
@@ -18,7 +18,6 @@ import com.github.ambry.config.StoreConfig;
 import com.github.ambry.utils.Time;
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


justinlin-linkedin (approved):
LGTM

ambry-store/src/main/java/com/github/ambry/store/HybridCompactionPolicy.java
```diff
@@ -80,16 +84,23 @@ public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapac
    * @param storeId id of the BlobStore
    * @return {@link CompactionPolicySwitchInfo} gets from map or recover from file if needed.
    */
-  private CompactionPolicySwitchInfo getCompactionPolicySwitchInfo(String storeId, String dataDir, BlobStoreStats blobStoreStats) {
+  private CompactionPolicySwitchInfo getCompactionPolicySwitchInfo(String storeId, String dataDir,
+      BlobStoreStats blobStoreStats) {
     if (!blobToCompactionPolicySwitchInfoMap.containsKey(storeId)) {
-      File file = new File(Paths.get(dataDir, COMPACT_POLICY_INFO_PATH).toString());
+      File file = new File(Paths.get(dataDir, COMPACT_POLICY_INFO_PATH_V2).toString());
+      File oldFile = new File(Paths.get(dataDir, COMPACT_POLICY_INFO_PATH).toString());
```
zzmao, line 91: You can create File without Paths:
```
new File(dataDir, COMPACT_POLICY_INFO_PATH_V2)
```

Also,
```
  private static final String COMPACT_POLICY_INFO_PATH_V2 = File.separator + "compactionPolicyInfoV2.json";
```
I feel `File.separator` is not necessary.

SophieGuo410: When we print dataDir, it won't have the slash at the end right? So without the File.separator, it will be like "/mnt/u001/ambrydata/230compactionPolicyInfo.json"?

SophieGuo410: Updated.

zzmao (approved):
LGTM

PR #1862, opened 2021-06-09
Fix adaptive compaction metric

zzmao:
Fix adaptive compaction metric

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -76,7 +76,7 @@ class BlobStoreCompactor {
   private static final Logger logger = LoggerFactory.getLogger(BlobStoreCompactor.class);
   private final File dataDir;
   private final String storeId;
-  private final String diskMountPath;
+  private final DiskMetrics diskMetrics;
   private final StoreKeyFactory storeKeyFactory;
   private final StoreConfig config;
   private final StoreMetrics srcMetrics;
@@ -119,15 +119,17 @@ class BlobStoreCompactor {
    * @param incarnationId the incarnation ID of the store.
    * @param accountService the {@link AccountService} instance to use.
    * @param remoteTokenTracker the {@link RemoteTokenTracker} that tracks tokens from all peer replicas.
+   * @param diskMetrics the {@link DiskMetrics}
    * @throws IOException if the {@link CompactionLog} could not be created or if commit/cleanup failed during recovery.
    * @throws StoreException if the commit failed during recovery.
    */
   BlobStoreCompactor(String dataDir, String storeId, StoreKeyFactory storeKeyFactory, StoreConfig config,
       StoreMetrics srcMetrics, StoreMetrics tgtMetrics, DiskIOScheduler diskIOScheduler,
       DiskSpaceAllocator diskSpaceAllocator, Log srcLog, Time time, UUID sessionId, UUID incarnationId,
-      AccountService accountService, RemoteTokenTracker remoteTokenTracker) throws IOException, StoreException {
+      AccountService accountService, RemoteTokenTracker remoteTokenTracker, DiskMetrics diskMetrics)
+      throws IOException, StoreException {
     this.dataDir = new File(dataDir);
-    this.diskMountPath = dataDir.substring(0, dataDir.lastIndexOf(File.separator));
+    this.diskMetrics = diskMetrics;
     this.storeId = storeId;
     this.storeKeyFactory = storeKeyFactory;
     this.config = config;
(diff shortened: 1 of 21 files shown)


ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -83,6 +84,7 @@
           StoreErrorCodes.File_Not_Found);
     }
     this.file = file;
+    this.diskMountPath = parseDiskMountPath(file);
```
jsjtzyy, line 87: Can we get this `diskMountPath` from `BlobStore`?

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -87,9 +89,9 @@
   public final Counter duplicateKeysInBatch;
   public final Counter storeIoErrorTriggeredShutdownCount;
   public final Counter blobStoreRecoverCompactionPolicySwitchInfoErrorCount;
-  public Histogram diskReadTimePerMbInMs;
-  public Histogram diskWriteTimePerMbInMs;
-  public Meter diskCompactionCopyRateInBytes;
+  public Map<String, Histogram> diskReadTimePerMbInMs = new HashMap<>();
+  public Map<String, Histogram> diskWriteTimePerMbInMs = new HashMap<>();
+  public Map<String, Meter> diskCompactionCopyRateInBytes = new HashMap<>();
```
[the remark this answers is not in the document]
jsjtzyy: please ignore this message, misunderstood this part.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
zzmao: If we want to get diskMountPath from BlobStore, we need to get it from `dataDir`.
This is not necessary, because in `Log.java`, we can directly get `diskMountPath` from parsing `dataDir`  and then pass `diskMountPath` to LogSegment.java. 
However, complexity wise, I didn't feel it's better than current solution. 

@cgtz  do you have any suggestions?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -821,19 +824,29 @@ private boolean copyRecords(LogSegment logSegmentToCopy, List<IndexEntry> srcInd
           long usedCapacity = tgtIndex.getLogUsedCapacity();
           if (isActive && (tgtLog.getCapacityInBytes() - usedCapacity >= srcValue.getSize())) {
             Offset endOffsetOfLastMessage = tgtLog.getEndOffset();
-            if (config.storeCompactionMinOperationsBytesPerSec < config.storeCompactionOperationsBytesPerSec) {
+            if (config.storeCompactionMinOperationsBytesPerSec < config.storeCompactionOperationsBytesPerSec
+                && srcMetrics.diskReadTimePerMbInMs.containsKey(diskMountPath)
```
cgtz, line 828: Could the disk metrics be saved as class members in the constructor of `BlobStoreCompactor`? `diskMountPath` never changes so maybe we can avoid the hashmap lookup each entry

ambry-store/src/main/java/com/github/ambry/store/StoreMessageReadSet.java
```diff
@@ -165,8 +166,11 @@ void doPrefetch(long relativeOffset, long size) throws IOException {
     prefetchedData = PooledByteBufAllocator.DEFAULT.ioBuffer((int) sizeToRead);
     long fetchStartTime = SystemTime.getInstance().milliseconds();
     prefetchedData.writeBytes(getChannel(), offset.getOffset() + relativeOffset, (int) sizeToRead);
-    segment.getMetrics().diskReadTimePerMbInMs.update(
-        ((SystemTime.getInstance().milliseconds() - fetchStartTime) << 20) / sizeToRead);
+    String diskMountPath = segment.getDiskMountPath();
```
cgtz, line 169: Similar comment here. The LogSegment could have a member variable with it's metrics and could expose `updateDiskReadTime` and `updateDiskWriteTime` methods instead of looking up in the hashmap every time

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
cgtz: What if we had the `BlobStore` pass down the metrics themselves. You could even make a small class to hold the three metrics together:
```
class DiskMetrics {
  Histogram diskRead, diskWrite;
  Meter copyRate
}
```
This would avoid any need to use maps, I feel, and is cleaner IMO, at the expense of more boilerplate constructor changes.

1. DiskManager constructs a DiskMetrics for the disk that it manages. It already knows it represents a single disk.
2. BlobStore takes in DiskMetrics in its constructor
3. LogSegment and BlobStoreCompactor take in DiskMetrics objects. They no longer need to worry about the mount path or hashmap lookups.

cgtz (changes requested):

cgtz (approved):
Looks like there are compilation issues because of usage of the old method in the tests:
` metrics.initializeDiskMetrics("/disk/mount/path", 200);`

Otherwise this looks good!

PR #1863, opened 2021-06-09
Use concurrent hash map for inner value-Map to avoid concurrent modification exception.

Arun-LinkedIn:
This PR fixes the exception due to concurrent modification on a collection used as a value inside concurrent hash map.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
@@ -453,7 +453,8 @@ public class ReplicationMetrics {
    */
   public void addLagMetricForPartition(PartitionId partitionId, boolean enableEmmitMetricForReplicaLag) {
     if (!partitionLags.containsKey(partitionId)) {
-      partitionLags.put(partitionId, new HashMap<>());
+      // Use concurrent hash map for the inner value map as well to avoid concurrent modification exception.
+      partitionLags.put(partitionId, new ConcurrentHashMap<>());
       // Set up metrics if and only if no mapping for this partition before.
       if (enableEmmitMetricForReplicaLag) {
         Gauge<Long> replicaLag = () -> getMaxLagForPartition(partitionId);
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -612,11 +612,6 @@ public class ReplicaThread implements Runnable {
                   }
                 }
 
-                // add exchangeMetadataResponse to list after replicaSyncUpManager(if not null) has completed update. The
-                // reason is replicaSyncUpManager may also throw exception and add one more exchangeMetadataResponse
-                // associated with same RemoteReplicaInfo.
-                exchangeMetadataResponseList.add(exchangeMetadataResponse);
-
                 // If remote token has not moved forward, wait for back off time before resending next metadata request
                 if (remoteReplicaInfo.getToken().equals(exchangeMetadataResponse.remoteToken)) {
                   remoteReplicaInfo.setReEnableReplicationTime(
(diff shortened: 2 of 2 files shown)


SophieGuo410 (approved):
Thanks for fixing this concurrent issue. LGTM!

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -453,7 +453,8 @@ public void incrementInvalidMessageError(PartitionId partitionId) {
    */
   public void addLagMetricForPartition(PartitionId partitionId, boolean enableEmmitMetricForReplicaLag) {
     if (!partitionLags.containsKey(partitionId)) {
-      partitionLags.put(partitionId, new HashMap<>());
+      // Use concurrent hash map for the inner value map as well to avoid concurrent modification exception.
+      partitionLags.put(partitionId, new ConcurrentHashMap<>());
```
jsjtzyy, line 457: I am not entirely sure if ConcurrentHashMap can block modification while iterating the entries within it.  Better to use a snapshot in `getMaxLagForPartition()` ?

jsjtzyy: I see this in ConcurrentHashMap:
```
    public Set<Map.Entry<K,V>> entrySet() {
        EntrySetView<K,V> es;
        return (es = entrySet) != null ? es : (entrySet = new EntrySetView<K,V>(this));
    }
```
It seems to just get a view of map and the comment says:
```
Returns a Set view of the mappings contained in this map. The set is backed by the map, so changes to the map are reflected in the set, and vice-versa.  ...
```
I wonder how does it protect iteration by blocking any concurrent modification on this map?

jsjtzyy (commented):
Mostly looks good. One minor comment.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
cgtz: The implementation of `EntryIterator` is written such that it will never throw `ConcurrentModificationException`.

It extends the `Traverser` inner class which iterates over the bins in a way that may a potentially older view but does not fail if changes since the time the iterator was created are detected. It should not block concurrent modification, but your iterator may or may not return values changed after the iterator was created. If there is a reason for iteration and modification of the structure to be mutually exclusive, you may need a lock. If you are okay with slightly older values showing up in the iterator as long as the iterator doesn't throw, i think this approach should be okay.
```
     * Encapsulates traversal for methods such as containsValue; also
     * serves as a base class for other iterators and spliterators.
     *
     * Method advance visits once each still-valid node that was
     * reachable upon iterator construction. It might miss some that
     * were added to a bin after the bin was visited, which is OK wrt
     * consistency guarantees. Maintaining this property in the face
     * of possible ongoing resizes requires a fair amount of
     * bookkeeping state that is difficult to optimize away amidst
     * volatile accesses.  Even so, traversal maintains reasonable
     * throughput.
     *
     * Normally, iteration proceeds bin-by-bin traversing lists.
     * However, if the table has been resized, then all future steps
     * must traverse both the bin at the current index as well as at
     * (index + baseSize); and so on for further resizings. To
     * paranoically cope with potential sharing by users of iterators
     * across threads, iteration terminates if a bounds checks fails
     * for a table read.
```

jsjtzyy: Thanks for more detailed clarification @cgtz .  I am now interested in how ConcurrentModificationException was triggered,  @Arun-LinkedIn could you provide a little more context?

jsjtzyy: After talking with @Arun-LinkedIn offline, we don't seem to have the up-to-date view for this metric. ConcurrentHashMap should be fine.

jsjtzyy (approved):

PR #1865, opened 2021-06-10
Add a gauge metric to record the number of log segment compacted in compaction

justinlin-linkedin:
Adding a metric to record how many log segments we can save from each round of compaction. 
We are expecting at least 1 log segment saved in StatsBasedCompactionStrategy.

This PR introduced a complex way to count how many segments are compacted because this PR also handles the case where process failed in the middle of compaction. The easy way is to have a variable to store the number of log segment before compaction and then after compaction we just subtract this number. But this won't work when the process fails in the middle of the compaction. That's why we need this complex method.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -43,6 +43,7 @@ import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1125,6 +1127,29 @@ private void endCompaction() throws StoreException {
     if (compactionLog != null) {
       if (compactionLog.getCompactionPhase().equals(CompactionLog.Phase.DONE)) {
         logger.info("Compaction of {} finished", storeId);
+        if (srcIndex != null && !srcIndex.isEmpty() && !compactionLog.cycleLogs.isEmpty()) {
+          // The log segment positions after compaction
+          Set<Long> compactedLogSegments = srcIndex.getLogSegments()
```
zzmao, line 1132: Still think the size of `compactedLogSegments` is good enough to give us a good reference. 

Also, suggest rename it to `logSegmentsAfterCompaction`

zzmao (approved):
LGTM

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -102,6 +102,8 @@ public boolean accept(File dir, String name) {
   private volatile CountDownLatch runningLatch = new CountDownLatch(0);
   private byte[] bundleReadBuffer;
   private final AtomicReference<CompactionDetails> currentCompactionDetails = new AtomicReference();
+  private final AtomicLong compactedLogCount = new AtomicLong(0L);
+  private final AtomicLong logSegmentCount = new AtomicLong(0L);
```
jsjtzyy, line 106: (optional) I feel like `AtomicInteger` should suffice.

jsjtzyy (approved):

PR #1866, opened 2021-06-11
Fix NPE in findkey method when index is updated due to compaction

jsjtzyy:
Assume that fileSpan's start offset is (0_0, 18) and there is a compaction happening parallel to this function. Also assume that compaction is compacting log segment [0_0, 1_0] to [0_1]. Then before compaction, the indexSegments.floorKey( (0_0, 18)) would return (0_0, 18) itself, since this is the smallest offset in the index segments map. However, if compaction finishes before this method call, then the fileSpan's start offset is still (0_0, 18) but the smallest index segment offset would become (0_1, 18). In this case, indexSegments.floorKey(fileSpan.getStartOffset()) would return null, thus we would see NullPointerException.

This PR makes index use current first log segment if we cannot find any segment less than or equal to start offset in old fileSpan.

### DIFF
--- a/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
+++ b/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
@@ -401,11 +401,36 @@ public class BlobStoreCompactorTest {
   }
 
   /**
+   * Test finding key with old file span. The log segment associated with start offset in file span has been compacted.
+   * @throws Exception
+   */
+  @Test
+  public void findKeyWithOldFileSpan() throws Exception{
+    assumeTrue(purgeDeleteTombstone);
+    refreshState(false, true, false);
+    FileSpan oldFileSpan = new FileSpan(state.index.getStartOffset(), state.index.getCurrentEndOffset());
+    MockId tombstone1 = state.permanentDeleteTombstones.get(0);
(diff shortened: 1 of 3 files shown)


ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
```diff
@@ -400,12 +400,37 @@ public void deleteTombstoneCleanupTest() throws Exception {
...
+    List<LogSegmentName> segmentsUnderCompaction = getLogSegments(0, 2);
+    CompactionDetails details = new CompactionDetails(state.time.milliseconds(), segmentsUnderCompaction, null);
+    compactor = getCompactor(state.log, DISK_IO_SCHEDULER, null, false);
+    compactor.initialize(state.index);
+    try {
+      compactor.compact(details, bundleReadBuffer);
+    } finally {
+      compactor.close(0);
+    }
+    // find the key by using old file span (the 1_0 log segment has been compacted to 1_1). The index should use current
```
justinlin-linkedin, line 422: minor issue: the old filespan's start offset is 0_0, not 1_0.

justinlin-linkedin (approved):

Arun-LinkedIn (approved):
lgtm!

PR #1868, opened 2021-06-12
[QUOTA] Chunk by chunk accounting of quota usage.

ankagrawal:
This PR defines a `QuotaChargeEventListener` interface which is passed to frontend router's request handlers to charge for each chunk being served for download or upload.


ambry-api/src/main/java/com/github/ambry/router/Router.java
```diff
@@ -34,10 +35,12 @@
    * @param blobId The ID of the blob for which blob data is requested.
    * @param options The options associated with the request. This cannot be null.
    * @param callback The callback which will be invoked on the completion of the request.
+   * @param quotaChargeEventListener Listener interface to charge quota cost for the operation.
    * @return A future that would eventually contain a {@link GetBlobResult} that can contain either
    *         the {@link BlobInfo}, the {@link ReadableStreamChannel} containing the blob data, or both.
    */
-  Future<GetBlobResult> getBlob(String blobId, GetBlobOptions options, Callback<GetBlobResult> callback);
+  Future<GetBlobResult> getBlob(String blobId, GetBlobOptions options, Callback<GetBlobResult> callback,
```
justinlin-linkedin, line 42: I think we should create a different method for this one, since there are some other callers who don't want to use a eventListener here.

ambry-api/src/main/java/com/github/ambry/router/Router.java
```diff
@@ -34,10 +35,12 @@
...
+      QuotaChargeEventListener quotaChargeEventListener);
```
justinlin-linkedin, line 43: And can we rename this to something else, like `ChunkEventListener`. I think it can be a generic callback not just for quota charging logic.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -774,11 +777,11 @@ public void onCompletion(Void result, Exception exception) {
           headCallback.markStartTime();
           router.getBlob(convertedId, new GetBlobOptionsBuilder().operationType(GetBlobOptions.OperationType.BlobInfo)
               .getOption(getOption)
-              .build(), headCallback);
+              .build(), headCallback, QuotaChargeEventListener.buildQuotaChargeEventListener(restRequest, quotaManager, false));
           break;
         case DELETE:
           deleteCallback.markStartTime();
-          router.deleteBlob(convertedId, getHeader(restRequest.getArgs(), Headers.SERVICE_ID, false), deleteCallback);
+          router.deleteBlob(convertedId, getHeader(restRequest.getArgs(), Headers.SERVICE_ID, false), deleteCallback, QuotaChargeEventListener.buildQuotaChargeEventListener(restRequest, quotaManager, false));
```
justinlin-linkedin, line 784: we probably don't need to do Chunk by Chunk on Delete.

ambry-api/src/main/java/com/github/ambry/router/Router.java
```diff
@@ -63,19 +67,22 @@
    *                       {@link ChunkInfo} object as a source of truth, so the caller should ensure that these
    *                       fields are set accurately.
    * @param callback The {@link Callback} which will be invoked on the completion of the request .
+   * @param quotaChargeEventListener Listener interface to charge quota cost for the operation.
    * @return A future that would contain the BlobId eventually.
    */
   Future<String> stitchBlob(BlobProperties blobProperties, byte[] userMetadata, List<ChunkInfo> chunksToStitch,
-      Callback<String> callback);
+      Callback<String> callback, QuotaChargeEventListener quotaChargeEventListener);
```
justinlin-linkedin, line 74: BTW, I don't think stitchBlob requires a callback here. All the chunk data has already been saved in ambry when we call stitchBlob.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
```diff
@@ -90,14 +91,8 @@ private double calculateCapacityUnitCost(RestRequest restRequest, RestResponseCh
     RequestPath requestPath = getRequestPath(restRequest);
     switch (restRequest.getRestMethod()) {
       case POST:
-        long contentSize = 0;
```
justinlin-linkedin, line 93: There is no need to remove this logic since the request cost is also used in C2S events.

ambry-frontend/src/test/java/com/github/ambry/frontend/FrontendRestRequestServiceTest.java
```diff
@@ -127,12 +128,14 @@
...
   private final static QuotaManager QUOTA_MANAGER;
   private final static String CLUSTER_NAME = "ambry-test";
+  private final static QuotaChargeEventListener QUOTA_CHARGE_EVENT_LISTENER;
 
   static {
     try {
       QUOTA_MANAGER =
           new AmbryQuotaManager(QuotaTestUtils.createQuotaConfig(Collections.emptyMap(), false, QuotaMode.TRACKING),
               new MaxThrottlePolicy(), Mockito.mock(AccountService.class), null, new MetricRegistry());
+      QUOTA_CHARGE_EVENT_LISTENER = () -> {};
```
justinlin-linkedin, line 138: we should definitely add new method in Router interface to avoid these empty callbacks.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnitThrottlePolicy.java
```diff
@@ -0,0 +1,4 @@
+package com.github.ambry.quota.capacityunit;
+
+public class CapacityUnitThrottlePolicy {
```
justinlin-linkedin, line 3: what this?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -153,7 +155,63 @@
   // The time spent by a chunk for data to be available in the channel.
   private long waitTimeForChannelDataAvailabilityMs;
 
-  private static final Logger logger = LoggerFactory.getLogger(PutOperation.class);
+  /**
+   * Construct a PutOperation with the given parameters. This private constructor is used for both blob uploads and
```
justinlin-linkedin, line 159: hmm, can we not move the code, it messes up the review.

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeEventListener.java
```diff
@@ -0,0 +1,43 @@
...
+      boolean shouldThrottle) {
+    return () -> {
+      ThrottlingRecommendation throttlingRecommendation = quotaManager.charge(restRequest);
+      if (throttlingRecommendation.shouldThrottle() && shouldThrottle) {
+        throw new RouterException("RequestQuotaExceeded", RouterErrorCode.TooManyRequests);
+      }
+    };
+  }
+
+  void onQuotaChargeEvent() throws RouterException;
```
justinlin-linkedin, line 42: I think we can change this to ChunkEventListener and passes some info this listener, like chunkId, chunkType, chunkSize, encryption , etc...

ambry-api/src/main/java/com/github/ambry/router/Router.java
ankagrawal: We can either create a different method, or pass null or empty lambda when callers don't want to use it. Keeping just one method helps keep Router less complicated in my opinion.
Still if you think otherwise, I can fix it.

ambry-api/src/main/java/com/github/ambry/router/Router.java
ankagrawal: The reason why I wanted to keep it as quota specific, instead of a generic interface, is that its not necessary that quota charge will happen on chunks. For example, in case of tll, delete and undelete, we are doing quota charge on the whole operation, and not on individual chunks. So I pass the `QuotaChargeEventListener` to each type `Operation` (e.g, `GetOperation`, `DeleteOperation` etc), and let the operation class itself decide where the quota charge needs to happen (on completion of main operation, or for each chunk). Let me know what you think about this.

The other alternative is to do quota charging at different places for different type of Ambry requests. I feel that this option scatters the quota code everywhere in the stack and something uniform is better.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
ankagrawal: Although we are passing `QuotaChargeEventListener` to all methods of `Router`, we are not charging quota on chunk operation for every operation. In delete, ttlupdate and undelete operation, we are charging quota when the main blob operation is complete.

ambry-api/src/main/java/com/github/ambry/router/Router.java
ankagrawal: stitch blob is charged for only one chunk. Since we create and save the metadata chunk in the stitch blob call, we charge one unit of quota.

ambry-frontend/src/main/java/com/github/ambry/frontend/UserQuotaRequestCostPolicy.java
ankagrawal: fixed.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnitThrottlePolicy.java
ankagrawal: removed.

justinlin-linkedin (changes requested):

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1292,6 +1299,15 @@ void checkAndMaybeComplete() {
...
       if (done) {
+        // the chunk is complete now. We can charge against quota for the chunk.
+        if (quotaChargeEventListener != null) {
+          try {
+            quotaChargeEventListener.onQuotaChargeEvent();
+          } catch (RouterException rEx) {
+            // For now we only log for quota charge exceptions for in progress requests.
+            logger.info("Exception {} while handling quota charge event", rEx.toString());
+          }
+        }
```
justinlin-linkedin, line 1310: We have to set the exception fo this PutOperation here in the catch statement, otherwise, it's not consistent with GetBlobOperation since we are setting exception in GetBlobOperation.

ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
```diff
@@ -76,6 +78,12 @@
   @Config(THROTTLING_MODE)
   public QuotaMode throttlingMode;
 
+  /**
+   * Should quota be throttled for in progress requests.
```
cgtz, line 82: "Should requests in progress be throttled if they exceed their quota"

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeEventListener.java
cgtz: I feel like the chunk size at the very least may be useful, even if we keep this quota specific. It is a bit of an unknown if the current strategy of treating each chunk as one CU will be granular enough, so maybe providing the size of the chunk may be useful for logging/fractional CU in the future.

ambry-api/src/main/java/com/github/ambry/quota/QuotaEnforcer.java
```diff
@@ -36,7 +36,16 @@
    * @param requestCostMap {@link Map} of {@link QuotaName} to the cost incurred to handle the request.
    * @return QuotaRecommendation object with the recommendation.
    */
-  QuotaRecommendation chargeAndRecommend(RestRequest restRequest, BlobInfo blobInfo, Map<QuotaName, Double> requestCostMap);
+  QuotaRecommendation chargeAndRecommend(RestRequest restRequest, BlobInfo blobInfo,
+      Map<QuotaName, Double> requestCostMap);
+
+  /**
+   * Makes an {@link QuotaRecommendation} using the information in {@link RestRequest}. This
+   * method also charges one unit of request cost against the quota.
```
cgtz, line 44: How does this method know which type of quota to charge against (usually given by QuotaName)?

ambry-api/src/main/java/com/github/ambry/router/Router.java
cgtz: Have we considered adding `quotaChargeEventListener` as a `GetBlobOption`/`PutBlobOption`? These have associated builders that can help with optional arguments like this.

We could create options classes for delete/undelete/ttlupdate to make it symmetric. Alternatively, do those methods even need to be passed a listener as opposed to the frontend handling this logic in the callback.

ambry-api/src/main/java/com/github/ambry/quota/QuotaEnforcer.java
ankagrawal: It looks at RestRequest's `getRestMethod` and converts it to write or read operation. For writes, we can charge for storage quota and WCU. For reads, we only charge for RCU.

ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
```diff
@@ -81,6 +81,6 @@ public int hashCode() {
    * Type of Ambry resource for which quota can be applied.
    */
   public enum QuotaResourceType {
-    ACCOUNT, CONTAINER
+    ACCOUNT, CONTAINER, SERVICE
```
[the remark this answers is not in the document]
ankagrawal: Its used in closed source code.

ambry-quota/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
```diff
@@ -50,18 +51,17 @@
   private static final int HTTP_STATUS_ALLOW = 200;
   private static final long NO_RETRY = -1L;
   private static final long BYTES_IN_GB = 1024 * 1024 * 1024;
-
+  private static final long CHUNK_COST_IN_BYTES = 4 * 1024;
```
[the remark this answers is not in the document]
ankagrawal: Yes this will need to be revised.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: Removed setting exception in GetBlobOperation as we are not throttling for in-progress requests for now.

ankagrawal (commented):
Initial response to review comments. Code changes for majority of the review comments coming up next.

ambry-quota/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-quota/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
```diff
@@ -119,6 +119,21 @@ public QuotaRecommendation chargeAndRecommend(RestRequest restRequest, BlobInfo
...
+  @Override
+  public QuotaRecommendation chargeAndRecommend(RestRequest restRequest) {
+    if (!RestUtils.isUploadRequest(restRequest)) {
+      return NO_QUOTA_VALUE_RECOMMENDATION;
+    }
+    if (!isReadRequest(restRequest)) {
+      // No cost for the desired QuotaName, then just call recommend
+      return recommend(restRequest);
+    }
+
```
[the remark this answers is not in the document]
ankagrawal: done.

ambry-quota/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
```diff
@@ -290,4 +305,13 @@ protected long getQuotaValueForContainer(Container container) {
...
   }
+
+  /**
+   * Check if the restRequest passed is for a READ request.
+   * @param restRequest {@link RestRequest} object.
+   * @return true is restRequest is a READ request. False otherwise.
+   */
+  private boolean isReadRequest(RestRequest restRequest) {
+    return (restRequest.getRestMethod() == RestMethod.GET || restRequest.getRestMethod() == RestMethod.HEAD);
+  }
```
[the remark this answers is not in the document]
ankagrawal: removed.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -75,6 +77,7 @@
   private final CryptoJobHandler cryptoJobHandler;
   private final AccountService accountService;
   private final Time time;
+  private final QuotaManager quotaManager;
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -829,6 +842,11 @@ protected void doOperationTowardsMaybeCompositeBlob(final String blobIdStr, Futu
             }
             currentOperationsCount.addAndGet(blobIdStrs.size());
             helper.doOperation.accept(blobIdStrs);
+            try {
+              quotaChargeEventListener.onQuotaChargeEvent();
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-api/src/main/java/com/github/ambry/router/Router.java
ankagrawal: Putting the listener or callback in Options class for each operation will make the APIs cleaner. We can take it as a follow up item after this PR is merged.

ambry-frontend/src/test/java/com/github/ambry/frontend/FrontendRestRequestServiceTest.java
ankagrawal: Using null now.

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeEventListener.java
ankagrawal: I will take this up as a follow up item after this PR is merged.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -85,6 +86,7 @@
   private final RetryExecutor retryExecutor = new RetryExecutor(Executors.newScheduledThreadPool(2));
   private final Set<RouterErrorCode> retriableRouterError =
       EnumSet.of(AmbryUnavailable, ChannelClosed, UnexpectedInternalError, OperationTimedOut);
+  private final QuotaChargeEventListener quotaChargeEventListener = () -> {};
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -242,6 +245,12 @@ private void onChunkOperationComplete(GetChunk chunk) {
               setOperationCompleted();
             }
             operationResult = new GetBlobResultInternal(new GetBlobResult(blobInfo, blobDataChannel), null);
+            try {
+              quotaChargeEventListener.onQuotaChargeEvent();
```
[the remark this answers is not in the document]
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -242,6 +245,12 @@ private void onChunkOperationComplete(GetChunk chunk) {
...
+            } catch (RouterException routerException) {
+              logger.info("Exception {} occurred during the quota charge event of blob {}", routerException, blobId.getID());
+              e = routerException;
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -753,7 +766,7 @@ protected void deleteBlob(final String blobIdStr, final String serviceId, Future
               if (callback != null) {
                 callback.onCompletion(result, exception);
               }
-            });
+            }, quotaChargeEventListener);
```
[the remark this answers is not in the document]
ankagrawal: We are delegating the quota charge callback responsibility to Operation classes. They can handle how and when they want to charge for quota.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -829,6 +842,11 @@ protected void doOperationTowardsMaybeCompositeBlob(final String blobIdStr, Futu
             }
             currentOperationsCount.addAndGet(blobIdStrs.size());
             helper.doOperation.accept(blobIdStrs);
+            try {
+              quotaChargeEventListener.onQuotaChargeEvent();
+            } catch (RouterException routerException) {
```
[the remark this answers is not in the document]
ankagrawal: The implementation of `onQuotaChargeEvent` makes sure that the only exception thrown out of that method will be RouterException.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -122,14 +122,19 @@ public ThrottlingRecommendation getThrottleRecommendation(RestRequest restReques
   @Override
   public ThrottlingRecommendation charge(RestRequest restRequest, BlobInfo blobInfo,
       Map<QuotaName, Double> requestCostMap) {
+    return charge(restRequest);
```
justinlin-linkedin, line 125: This method should probably stay for future usage.

ambry-router/src/main/java/com/github/ambry/router/BatchOperationCallbackTracker.java
```diff
@@ -42,13 +43,18 @@
    * @param callback the {@link Callback} to be triggered once acks are received for all blobs
    */
   BatchOperationCallbackTracker(List<BlobId> blobIds, FutureResult<Void> futureResult, Callback<Void> callback) {
+    this(blobIds, futureResult, callback, null);
+  }
+
+  BatchOperationCallbackTracker(List<BlobId> blobIds, FutureResult<Void> futureResult, Callback<Void> callback, Callback<Void> completionCallback) {
```
justinlin-linkedin, line 49: why do we need another callback here, we already have `callback`, there is no need to have `completionCallback`.

ambry-router/src/main/java/com/github/ambry/router/BatchOperationCallbackTracker.java
```diff
@@ -78,6 +84,9 @@
    * @param e the {@link Exception} that occurred (if any).
    */
   private void complete(Exception e) {
+    if(completionCallback != null) {
+      completionCallback.onCompletion(null, null);
+    }
```
justinlin-linkedin, line 89: The callback will be invoked in line 91. we don't need another one.

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
```diff
@@ -356,6 +360,7 @@ private void handleBody(InputStream payload, MessageMetadata messageMetadata, Me
     if (encryptionKey == null) {
       // if blob is not encrypted, move the state to Complete
       BlobInfo blobInfo = new BlobInfo(serverBlobProperties, userMetadata.array(), messageInfo.getLifeVersion());
+      handleQuotaEvent();
```
justinlin-linkedin, line 363: nit: can we do this in `checkAndMaybeComplete` method, just like what we did in `DeleteOperation`.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterFactory.java
```diff
@@ -112,6 +120,11 @@ public NonBlockingRouterFactory(VerifiableProperties verifiableProperties, Clust
     cryptoService = cryptoServiceFactory.getCryptoService();
     cryptoJobHandler = new CryptoJobHandler(routerConfig.routerCryptoJobsWorkerCount);
     defaultPartitionClass = clusterMapConfig.clusterMapDefaultPartitionClass;
+    FrontendConfig frontendConfig = new FrontendConfig(verifiableProperties);
+    AccountStatsStore accountStatsStore =
+        Utils.<AccountStatsStoreFactory>getObj(frontendConfig.accountStatsStoreFactory, verifiableProperties,
+            clusterMapConfig, new StatsManagerConfig(verifiableProperties),
+            clusterMap.getMetricRegistry()).getAccountStatsStore();
```
justinlin-linkedin, line 127: what is this?

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -320,21 +331,17 @@ private void checkAndMaybeComplete() {
         operationException.set(
             new RouterException("TtlUpdateOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
       }
+      if (quotaChargeCallback != null) {
+        try {
+          quotaChargeCallback.chargeQuota();
+        } catch (RouterException rEx) {
+          LOGGER.info("Exception {} while charging quota for ttl update operation.", rEx.toString());
+        }
+      }
```
justinlin-linkedin, line 340: we are charging too many for TTLUPDATE and UNDELETE. 
We should only charge once for TTL_UPDATE and UNDELETE, and we already checked it on TtlUpdateManager and UndeleteManager, we don't need to change it again in TtlUpdateOperation.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterQuotaCallbackTest.java
```diff
@@ -0,0 +1,200 @@
...
+import org.slf4j.LoggerFactory;
+
+import static com.github.ambry.utils.TestUtils.*;
+import static org.junit.Assert.*;
+
+
+/**
+ * Class to test the {@link NonBlockingRouter} with quota callbacks.
+ */
+public class NonBlockingRouterQuotaCallbackTest extends NonBlockingRouterTestBase {
```
justinlin-linkedin, line 41: I think taking QuotaCallback test out from NonBlockingRouterTest is unnecessary. We only have one test method and we are take huge chunk of code and create a base file. It's too many changes for too little gain.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
ankagrawal: The main implementation for this interface is in closed source code. The method exists there.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterFactory.java
ankagrawal: Not sure why its here. Maybe due to rebase.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
ankagrawal: ttlupdate and undelete have different flows for a simple blob and for a composite blob. This charge here will happen only for the case of a simple blob's ttl update. The reason being that for the BatchOperation, the quotaChargeCallback is passed as null. Hence for composite blobs, only one charge will happen when all the blobs are updated.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterQuotaCallbackTest.java
ankagrawal: The main motivation behind taking this out was that I want to test it for a much large number of parameters. And I don't want to test all the test cases of original NonBlockingRouterTest for all those parameters. This slows down our test time for the build.

ambry-router/src/main/java/com/github/ambry/router/BatchOperationCallbackTracker.java
ankagrawal: fixed.

ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/BatchOperationCallbackTracker.java
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/BatchOperationCallbackTracker.java
```diff
@@ -78,6 +85,13 @@
    * @param e the {@link Exception} that occurred (if any).
    */
   private void complete(Exception e) {
+    if(quotaChargeCallback != null) {
+      try {
+        quotaChargeCallback.chargeQuota();
+      } catch (RouterException rEx) {
+        LOGGER.info("Exception {} while charging quota for ttl operation", rEx.toString());
+      }
+    }
```
justinlin-linkedin, line 94: we should move this blob inside of the if blob below, like
```
    if (completed.compareAndSet(false, true)) {
        if(quotaChargeCallback != null) {
            try {
                quotaChargeCallback.chargeQuota();
            } catch (RouterException rEx) {
                LOGGER.info("Exception {} while charging quota for ttl operation", rEx.toString());
            }
      }
      NonBlockingRouter.completeOperation(futureResult, callback, null, e, false);
    }
```

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterFactory.java
```diff
@@ -15,19 +15,27 @@
 
 import com.codahale.metrics.MetricRegistry;
 import com.github.ambry.account.AccountService;
+import com.github.ambry.accountstats.AccountStatsStore;
```
justinlin-linkedin, line 18: These imports don't get used anywhere.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
justinlin-linkedin: that makes sense

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterFactory.java
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/BatchOperationCallbackTracker.java
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -321,21 +309,17 @@ private void checkAndMaybeComplete() {
         operationException.set(
             new RouterException("DeleteOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
       }
+      if(quotaChargeCallback != null) {
```
cgtz, line 312: I think there is an issue here. `checkAndMaybeComplete` is intended as an idempotent method that can be called multiple times. For example, let's look at the `handleResponse`  call. This is called for every response that comes back from a server. I think it may be possible for a response to come back such that the success threshold for the operation comes back, and this code block will be executed. Then, the 3rd successful response may come back and this would be called again.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -845,6 +848,14 @@ void checkAndMaybeComplete() {
         chunkCompleted = true;
       }
       if (chunkCompleted) {
+        if (quotaChargeCallback != null) {
```
cgtz, line 851: For the actions that are done from within checkAndMaybeComplete in all classes, I would suggest that we call the callback only if the completion status is not already ChunkState.Complete:

```
if (state != ChunkState.Complete && quotaChargeCallback != null) { chargeQuota }
```

This comment applies to all operation classes

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
cgtz: +1 let's centralize on checkAndMaybeComplete with extra state checking logic to ensure the callback is called only once.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
```diff
@@ -340,6 +362,13 @@ private void checkAndMaybeComplete() {
         operationException.set(
             new RouterException("UndeleteOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
       }
+      if(quotaChargeCallback != null) {
```
cgtz, line 365: all of these code blocks seem to be missing a space in between if and paren.

cgtz (approved):
LGTM after fixing issues to make sure callback only called once.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
ankagrawal: fixed.

cgtz (approved):

PR #1871, opened 2021-06-14
Refactor VCR cluster interfaces

zzmao:
This PR refactors VCR helix cluster related interfaces. 
A `VcrClusterAgentFactory` is created to provide methods: `getVcrClusterParticipant` and `getClusterSpectator`. 

`VcrClusterParticipant`, previously `VirtualReplicatorCluster`, is to handle participant logic. 
`VcrClusterSpectator`, previously `HelixClusterSpectator`, is to handle spectator logic.


ambry-api/src/main/java/com/github/ambry/clustermap/VcrClusterAgentFactory.java
```diff
@@ -18,12 +18,15 @@
 
 
 /**
- * A factory to create {@link ClusterSpectator} object.
+ * A factory interface to get {@link VcrClusterParticipant} and {@link VcrClusterSpectator}.
  */
-public class HelixClusterSpectatorFactory implements ClusterSpectatorFactory {
+public interface VcrClusterAgentFactory {
```
cgtz, line 23: nit: We may want to make agent plural to match `ClusterAgentsFactory`. Unless you think the way it is sounds better

build.gradle
```diff
@@ -181,7 +181,10 @@ project(':ambry-api') {
         compile "io.dropwizard.metrics:metrics-core:$metricsVersion"
         compile "com.fasterxml.jackson.core:jackson-core:$jacksonVersion"
         compile "com.fasterxml.jackson.core:jackson-annotations:$jacksonVersion"
-        compile "io.netty:netty-buffer:$nettyVersion"
+        compile "io.netty:netty-buffer:$nettyVersion",
+        compile ("org.apache.helix:helix-core:$helixVersion") {
```
cgtz, line 185: it would be nice to keep helix out of ambry-api's dependencies if possible since it is used by ambry-client and helix's transitive dependencies have caused conflict in the past. (long term we can solve this problem by isolating the classes/interfaces that are actually needed by the client into a separate module and keep this as an internal-api module).

How hard do you think it would be to change `VcrClusterSpectator` to not implement helix interfaces and instead deal with ambry defined listener interfaces? It may be challenging since we probably have code that deals with raw `InstanceConfig` and `LiveInstance` objects. Ideally, we'd have ambry classes to represent vcr nodes like we do in the server clustermap so replication would not be coupled with helix concepts.

zzmao: I also have the plan to do change like you said. A separate PR for this change would be clear for reviewers. 
Moving out helix dependencies from ambry-api in this PR.

ambry-api/src/main/java/com/github/ambry/clustermap/VcrClusterAgentFactory.java
```diff
@@ -18,12 +18,15 @@
...
+   * @return an instance of {@link VcrClusterParticipant} generated by the factory.
+   */
+  VcrClusterParticipant getVcrClusterParticipant() throws Exception;
+
 
-  @Override
-  public ClusterSpectator getClusterSpectator(CloudConfig cloudConfig, ClusterMapConfig clusterMapConfig) {
-    return new HelixClusterSpectator(cloudConfig, clusterMapConfig);
-  }
+  VcrClusterSpectator getVcrClusterSpectator(CloudConfig cloudConfig, ClusterMapConfig clusterMapConfig);
```
ankagrawal, line 31: javadocs

ambry-api/src/main/java/com/github/ambry/clustermap/VcrClusterSpectator.java
```diff
@@ -18,9 +18,9 @@
 
 
 /**
- * {@link ClusterSpectator} adds itself as a spectator of a cluster and registers some cluster change listeners.
+ * {@link VcrClusterSpectator} adds itself as a spectator of a cluster and registers some cluster change listeners.
  */
-public interface ClusterSpectator extends InstanceConfigChangeListener, LiveInstanceChangeListener {
+public interface VcrClusterSpectator extends InstanceConfigChangeListener, LiveInstanceChangeListener {
```
ankagrawal, line 23: nit: does it need to be called `VcrClusterSpectator` instead of `ClusterSpectator` ? It doesn't look to have any vcr specific methods.

ambry-api/src/main/java/com/github/ambry/clustermap/VcrClusterAgentFactory.java
ankagrawal, line 23: Should this. implement `ClusterAgentsFactory`? I was wondering if this class shouldn't return ClusterPartiticipant and ClusterSpectator at all. Instead of that it can have methods like -- getPeerCloudReplica(), getAssignedPartition(), getPeerStoreReplicas() etc. Things like `VcrClusterParticipant` and `VcrClusterSpectator` would then become internal implementation details, and vcr replication logic wouldn't need to know how to deal with participant and spectator.

build.gradle
cgtz: Sure, sounds like a plan.  maybe we could put all these interfaces in the ambry-commons module for now. This will be like it was before where the module was ambry-commons and the java package was com.github.ambry.clustermap.

ambry-api/src/main/java/com/github/ambry/clustermap/VcrClusterAgentFactory.java
zzmao: Basically, you are trying to model this as a `clustermap` and the code in this PR is trying to model this by roles. 

It's nice to have a unified interface to provide getAssignedPartition(), getPeerStoreReplicas() etc. 
However, `getAssignedPartition()` only belongs to `Participant`; getPeerReplica() can have two implementation, one for `Spectator` and the other for Participant`.  With this in mind, we need to provide role registration method in this `"clustermap"`.  This increases the implementation complexity. 

The code in this PR split functionalities by role. `Participant` provides `getAssignedPartition()`, `getPeerReplica()`;  `Spectator`  provides `getPeerReplica(replica)`. It's more natural. In addition, this can give node flexibility to be what role(s).

From abstraction wise, it's good to do sth like you suggested. But from implementation wise, it's clear to have two roles, each has their functionalities.

zzmao: but anyway, I will change VcrClusterSpectator to not implement helix interfaces and only deal with ambry defined listener interfaces.  (Maybe next PR)

zzmao: done

ambry-api/src/main/java/com/github/ambry/clustermap/VcrClusterAgentFactory.java
zzmao: added

ambry-api/src/main/java/com/github/ambry/clustermap/VcrClusterSpectator.java
zzmao: Based on Casey's suggestion, we will change VcrClusterSpectator to not implement helix interfaces and instead deal with ambry defined VCR listener interfaces. So this is will be a VCR specific spectator.

ankagrawal (approved):

cgtz (approved):

PR #1876, opened 2021-06-15
[named blobs] set correct ttl in metadata DB for permanent blobs

cgtz:
Previously, in the permanent blob upload flow for named blobs, the DB
would be populated with the TTL used for the initial router upload.
Later, the ambry blob TTL would be updated to make the blob permanent.
However, the metadata would still have the original expiry time.

This ensures that the metadata record is marked as permanent by not
sharing the same instance of BlobProperties with the router call. This
commit also supplies the correct initial TTL for stitchBlob calls.


SophieGuo410 (approved):
lgtm!

zzmao (approved):
LGTM

PR #1878, opened 2021-06-15
[StatsMigration] Using HikariCP connection pool in AccountStatsStore

justinlin-linkedin:
Staring using HikariCP connection pool in account stats mysql store in hope that some day we can get rid of MysqlDataAccessor.


ambry-quota/src/main/java/com/github/ambry/quota/storage/MySqlStorageUsageRefresher.java
```diff
@@ -150,8 +150,6 @@ private void initializeContainerStorageUsageMonthlyBase() {
           }
         } catch (Exception e) {
           throw new IllegalStateException("Unable to fetch monthly storage usage from mysql", e);
-        } finally {
-          accountStatsStore.closeConnection();
```
Arun-LinkedIn, line 154: Since we are using connection pool now, just thinking if we still need synchronization on access to `accountStatsStore` in few of the methods in this class? Seems like it is removed on other classes like `StatsManager` and `MySqlReportAggregatorTask`.

Arun-LinkedIn (commented):
Looks good to me. Just had one comment.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
```diff
@@ -117,57 +117,54 @@ public TaskResult run() {
...
-          if (clusterMapConfig.clustermapEnableDeleteInvalidDataInMysqlAggregationTask) {
-            removeInvalidAggregatedAccountAndContainerStats(results.getSecond());
-          }
-          accountStatsStore.storeAggregatedAccountStats(results.getSecond());
-        } else if (statsReportType == StatsReportType.PARTITION_CLASS_REPORT) {
-          if (clusterMapConfig.clustermapEnableDeleteInvalidDataInMysqlAggregationTask) {
-            removeInvalidAggregatedPartitionClassStats(results.getSecond());
-          }
-          accountStatsStore.storeAggregatedPartitionClassStats(results.getSecond());
+    long startTimeMs = System.currentTimeMillis();
```
cgtz, line 120: Is the only change here the removal of synchronization and the close connection at the end?

Was the synchronization before to protect the sql connection related data structures, but that is no longer required since the connection pool now handles it?

ambry-mysql/src/integration-test/java/com/github/ambry/accountstats/AccountStatsMySqlStoreIntegrationTest.java
```diff
@@ -505,17 +520,8 @@ private static StatsWrapper generateStatsWrapper(int numPartitions, int numAccou
...
-    for (String table : AccountStatsMySqlStore.TABLES) {
-      statement.executeUpdate("DELETE FROM " + table);
-    }
-  }
-
-  private void assertTableSize(MySqlDataAccessor dataAccessor, int expectedNumRows) throws SQLException {
-    Connection dbConnection = dataAccessor.getDatabaseConnection(true);
-    Statement statement = dbConnection.createStatement();
+  private void assertTableSize(Connection connection, int expectedNumRows) throws SQLException {
+    Statement statement = connection.createStatement();
```
cgtz, line 524: I know this is test code, but probably still a good practice to use try-with-resources or try-finally here.

ambry-quota/src/main/java/com/github/ambry/quota/storage/MySqlStorageUsageRefresher.java
justinlin-linkedin: updated, forgot to remove the synchroinzed block in this file.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/MySqlReportAggregatorTask.java
justinlin-linkedin: yes, that's the only change for these files. 

We used to have only one connection, now with the connection pool can we more than one connections, we don't need to synchronize those block anymore.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -615,6 +599,13 @@ public StatsSnapshot queryAggregatedPartitionClassStatsByClusterName(String clus
     return new StatsSnapshot(partitionClassNameValue, partitionClassNameSubMap);
   }
 
+  @Override
+  public void shutdown() {
+    if (dataSource instanceof HikariDataSource) {
```
cgtz, line 604: could this just be `instanceof Closeable` or are there some datasources we don't want to close?

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStoreFactory.java
```diff
@@ -67,44 +68,32 @@ public AccountStatsMySqlStoreFactory(VerifiableProperties verifiableProperties,
...
-        String sep = ep.getUrl().contains("?") ? "&" : "?";
-        String newUrl = new StringBuilder(ep.getUrl()).append(sep)
-            .append(REWRITE_BATCHED_STATEMENTS_LITERAL)
-            .append("=true")
-            .toString();
-        result.add(new DbEndpoint(newUrl, ep.getDatacenter(), ep.isWriteable(), ep.getUsername(), ep.getPassword()));
-      }
-    }
-    return result;
+  private HikariDataSource buildDataSource(DbEndpoint dbEndpoint) {
```
cgtz, line 82: not for this PR, but in the future, we can probably put this in a util class since it is pretty similar to the setup code used in named blobs.

ambry-mysql/src/main/java/com/github/ambry/mysql/BatchUpdater.java
```diff
@@ -64,29 +64,32 @@
...
       statement.clearBatch();
-      autoCommit = dataAccessor.getAutoCommmit();
-      dataAccessor.setAutoCommit(false);
+      autoCommit = connection.getAutoCommit();
+      connection.setAutoCommit(false);
     } catch (SQLException e) {
-      dataAccessor.onException(e, BatchUpdate);
+      metrics.batchUpdateFailureCount.inc();
       logger.error("Failed to prepare for batch insert on {}", tableName, e);
+      connection.close();
```
cgtz, line 92: we may want to close the prepared statement too if non-null since get/setAutoCommit somehow fails.

Or, does closing the connection automatically release any prepared statement resources?

cgtz (approved):
Thank you for doing this. This seems to simplify a lot of code!

Arun-LinkedIn (approved):

PR #1885, opened 2021-06-22
Add java setup to gh actions publish job

cgtz:
We need to have the basic java/gradle/repo setup steps in the publishing
job too.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -143,22 +143,27 @@ jobs:
       ARTIFACTORY_USER: ${{ secrets.ARTIFACTORY_USER }}
       ARTIFACTORY_API_KEY: ${{ secrets.ARTIFACTORY_API_KEY }}
       GIT_SECRET: ${{ secrets.GIT_SECRET }}
-    if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && secrets.PUBLISH == 'true'}}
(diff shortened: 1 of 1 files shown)


.github/workflows/github-actions.yml
```diff
@@ -145,6 +145,34 @@ jobs:
...
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Grant execute permission for gradlew
+        run: chmod +x gradlew
+
       - name: Publish To JFrog artifactory
         if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && env.PUBLISH == 'true'}}
```
cgtz, line 177: hmm, i think i actually have to move this if block to apply to all of these

cgtz: GH actions supports if conditions at the job level, so using that: https://github.blog/changelog/2019-10-01-github-actions-new-workflow-syntax-features/

cgtz: Turns out that neither `env` or `secrets` is accessible from a job level conditional, so for now we have to duplicate the if statement on each step: https://github.community/t/conditional-job-using-environment-variable/141995

Arun-LinkedIn (approved):
lgtm!

PR #1890, opened 2021-06-24
Update token generation when last log segment gets closed

SophieGuo410:
This pr support the following bug fix:
1. If the input is journal based token, but journal has been cleaned up by auto close last log segment feature, it should return the indexed based token with the last IndexSegment startOffset and first key.
2. If the input is index based token,  but journal has been cleaned up by auto close last log segment feature, it should moving forward until it reaches to the end of the last index segment.
3. Bypass throw exception since some of the previous check assume that if the token segmentStartOffset equals to indexSegments.lastKey(), it must be the journal based token and entry should be in journal. But since auto close last log segment will clean up journal, so we should bypass the exception reporting for this situation. 
4. If the segmentStartOffset of the index based token equals to last index segment, there's no need to check the journal since journal should be empty.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -1279,10 +1279,13 @@ class PersistentIndex {
           // Get entries starting from the first key in this offset.
           Map.Entry<Offset, IndexSegment> entry = indexSegments.floorEntry(offsetToStart);
           StoreFindToken newToken;
-          //if the token is point to the last index entry of last index segment and journal cleaned up after last log segment
-          //get auto closed, we should transform the journal based token into index based token.
-          if (entry != null && !entry.getKey().equals(indexSegments.lastKey()) || onlyContainsHeaderInLastLogSegment(
-              offsetToStart, indexSegments)) {
+          //There are several reasons that we could reach here:
+          //1.StoreFindToken is Uninitialized, and we have more index segments than journal can cover.
+          //2.StoreFindToken is JournalBased, but the journal somehow moved too much ahead, the offset in the FindToken is no longer within journal.
+          //  2.1 There might just be lots of new entries coming to journal so the journal moves ahead.
+          //  2.2 The journal might be cleared and new log segment is created for compaction (auto-close last segment)
+          if (entry != null && !entry.getKey().equals(indexSegments.lastKey())
+              || onlyContainsHeaderInfoInLastLogSegment()) {
             startTimeInMs = time.milliseconds();
             newToken = findEntriesFromSegmentStartOffset(entry.getKey(), null, messageEntries,
                 new FindEntriesCondition(maxTotalSizeOfEntries), indexSegments, storeToken.getInclusive());
@@ -1313,7 +1316,8 @@ class PersistentIndex {
         StoreFindToken newToken;
         //if index based token point to the last entry of last index segment, and journal size cleaned up after last log
         //segment auto close, we should return the original index based token instead of getting from journal.
-        if (onlyContainsHeaderInLastLogSegment(storeToken.getOffset(), indexSegments)) {
+        if (indexBasedTokenPointToLastEntryOfLastIndexSegment(storeToken, indexSegments)
+            && onlyContainsHeaderInfoInLastLogSegment()) {
           newToken = storeToken;
         } else {
           newToken = findEntriesFromSegmentStartOffset(storeToken.getOffset(), storeToken.getStoreKey(), messageEntries,
@@ -1567,11 +1571,9 @@ class PersistentIndex {
       List<MessageInfo> messageEntries, FindEntriesCondition findEntriesCondition,
       ConcurrentSkipListMap<Offset, IndexSegment> indexSegments, boolean inclusive) throws StoreException {
     Offset segmentStartOffset = initialSegmentStartOffset;
-    // if the journal based token point to the last entry of last index segment but journal cleaned up after to last log
-    //segment auto closed, we should return the corresponding index token instead of throw exception.
-    if (segmentStartOffset.equals(indexSegments.lastKey()) && !onlyContainsHeaderInLastLogSegment(segmentStartOffset,
-        indexSegments)) {
-      // We would never have given away a token with a segmentStartOffset of the latest segment.
+    if (segmentStartOffset.equals(indexSegments.lastKey()) && !onlyContainsHeaderInfoInLastLogSegment()) {
+      //We would never have given away a token with a segmentStartOffset of the latest segment unless the journal has
+      //been cleaned up by (auto close last log segment).
       throw new IllegalArgumentException(
           "Index : " + dataDir + " findEntriesFromOffset segment start offset " + segmentStartOffset
               + " is of the last segment");
@@ -1643,10 +1645,8 @@ class PersistentIndex {
         break; // we have entered and finished reading from the journal, so we are done.
       }
 
-      //if the journal based token point to the last entry of last index segment but journal cleaned up after to last log
-      //segment auto closed, we can bypass this check and return the index based token.
-      if (segmentStartOffset == validIndexSegments.lastKey() && !onlyContainsHeaderInLastLogSegment(segmentStartOffset,
-          validIndexSegments)) {
+      //if journal is empty due to auto close last log segment, there's no need to check this.
+      if (segmentStartOffset.equals(validIndexSegments.lastKey()) && !onlyContainsHeaderInfoInLastLogSegment()) {
         /* The start offset is of the latest segment, and was not found in the journal. This means an entry was added
          * to the index (creating a new segment) but not yet to the journal. However, if the journal does not contain
          * the latest segment's start offset, then it *must* have the previous segment's start offset (if it did not,
@@ -1700,16 +1700,30 @@ class PersistentIndex {
   }
 
   /**
-   * @param startOffset the index segment start {@link Offset}.
+   *
+   * @param token the {@link StoreFindToken} need to be checked.
+   * @param indexSegments the map of index segment start {@link Offset} to {@link IndexSegment} instances.
+   * @return {@code true} if the token point to the last entry of last log segment. Otherwise return {@code false}.
+   */
+  private boolean indexBasedTokenPointToLastEntryOfLastIndexSegment(StoreFindToken token,
+      ConcurrentSkipListMap<Offset, IndexSegment> indexSegments) {
+    StoreKey storeKey = token.getStoreKey();
+    IndexSegment lastIndexSegment = indexSegments.lastEntry().getValue();
+    StoreKey lastEntryInLastIndexSegment = lastIndexSegment.listIterator(lastIndexSegment.size()).previous().getKey();
+    logger.trace("DataDir {}: StoreKey for token {} is {}, StoreKey in last index segment is {}", dataDir, token,
+        storeKey, lastEntryInLastIndexSegment);
+    return lastEntryInLastIndexSegment.equals(storeKey) && token.getOffset().equals(lastIndexSegment.getStartOffset());
+  }
+
+  /**
    * @return {@link true} if last log segment only contains the header info after auto closing last log segment during compaction.
    */
-  private boolean onlyContainsHeaderInLastLogSegment(Offset startOffset,
-      ConcurrentSkipListMap<Offset, IndexSegment> indexSegments) {
-    logger.trace("Log segment name {} for start offset {}", startOffset.getName(), startOffset);
-    logger.trace("Current last log segment name {} when check start offset {}", log.getLastSegment().getName(),
-        startOffset);
-    return indexSegments != null && indexSegments.size() > 0 && startOffset.getOffset() == indexSegments.lastKey()
-        .getOffset() && startOffset.getName().compareTo(log.getLastSegment().getName()) < 0;
+  private boolean onlyContainsHeaderInfoInLastLogSegment() {
+    logger.trace("DataDir: {} journal size is : {}", dataDir, journal.getCurrentNumberOfEntries());
+    logger.trace("DataDir: {} last log segment start offset: {}, end offset: {}", dataDir,
+        log.getLastSegment().getStartOffset(), log.getLastSegment().getEndOffset());
+    return journal.getCurrentNumberOfEntries() == 0 && log.getLastSegment().isEmpty() && !log.getLastSegment()
+        .getName().equals(log.getFirstSegment().getName());
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -1285,7 +1285,7 @@ class PersistentIndex {
           //  2.1 There might just be lots of new entries coming to journal so the journal moves ahead.
           //  2.2 The journal might be cleared and new log segment is created for compaction (auto-close last segment)
           if (entry != null && !entry.getKey().equals(indexSegments.lastKey())
-              || onlyContainsHeaderInfoInLastLogSegment()) {
+              || isJournalEmptyAfterAutoClose()) {
             startTimeInMs = time.milliseconds();
             newToken = findEntriesFromSegmentStartOffset(entry.getKey(), null, messageEntries,
                 new FindEntriesCondition(maxTotalSizeOfEntries), indexSegments, storeToken.getInclusive());
@@ -1314,10 +1314,10 @@ class PersistentIndex {
         // Get entries starting from the token Key in this index.
         startTimeInMs = time.milliseconds();
         StoreFindToken newToken;
-        //if index based token point to the last entry of last index segment, and journal size cleaned up after last log
+        //if index based token points to the last entry of last index segment, and journal size cleaned up after last log
         //segment auto close, we should return the original index based token instead of getting from journal.
-        if (indexBasedTokenPointToLastEntryOfLastIndexSegment(storeToken, indexSegments)
-            && onlyContainsHeaderInfoInLastLogSegment()) {
+        if (indexBasedTokenPointsToLastEntryOfLastIndexSegment(storeToken, indexSegments)
+            && isJournalEmptyAfterAutoClose()) {
           newToken = storeToken;
         } else {
           newToken = findEntriesFromSegmentStartOffset(storeToken.getOffset(), storeToken.getStoreKey(), messageEntries,
@@ -1571,7 +1571,7 @@ class PersistentIndex {
       List<MessageInfo> messageEntries, FindEntriesCondition findEntriesCondition,
       ConcurrentSkipListMap<Offset, IndexSegment> indexSegments, boolean inclusive) throws StoreException {
     Offset segmentStartOffset = initialSegmentStartOffset;
-    if (segmentStartOffset.equals(indexSegments.lastKey()) && !onlyContainsHeaderInfoInLastLogSegment()) {
+    if (segmentStartOffset.equals(indexSegments.lastKey()) && !isJournalEmptyAfterAutoClose()) {
       //We would never have given away a token with a segmentStartOffset of the latest segment unless the journal has
       //been cleaned up by (auto close last log segment).
       throw new IllegalArgumentException(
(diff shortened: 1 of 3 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1290,7 +1279,10 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
           // Get entries starting from the first key in this offset.
           Map.Entry<Offset, IndexSegment> entry = indexSegments.floorEntry(offsetToStart);
           StoreFindToken newToken;
-          if (entry != null && !entry.getKey().equals(indexSegments.lastKey())) {
+          //if the token is point to the last index entry of last index segment and journal cleaned up after last log segment
+          //get auto closed, we should transform the journal based token into index based token.
+          if (entry != null && !entry.getKey().equals(indexSegments.lastKey()) || onlyContainsHeaderInLastLogSegment(
+              offsetToStart, indexSegments)) {
```
justinlin-linkedin, line 1285: I think this deserves more comments, like

The reasons why we are here could be 
1. StoreFindToken is Uninitialized, and we have more index segments than journal can cover
2. StoreFindToken is JournalBased, but the journal somehow moved to much ahead, the offset in the FindToken it no longer within journal.
    2.1 There might just be lots of new entries coming to journal so the journal moves ahead
    2.2 The journal might be cleared and new log segment is created for compaction (auto-close last segment)

These would be three reasons why we are here.
For reason 1, entry equals to first index segment and doesn't equals to last segment (we have more index segments  than journal can cover)
For reason 2.1. entry equals to some index segment in the middle, but not last segment
For reason 2.2 entry might equal to last segment is the new log segment doesn't have any new record.

That's why we are adding the new check  here.

And I have no idea how we can reach line 1297.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1318,9 +1310,15 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
         // Find the index segment corresponding to the token indexStartOffset.
         // Get entries starting from the token Key in this index.
         startTimeInMs = time.milliseconds();
-        StoreFindToken newToken =
-            findEntriesFromSegmentStartOffset(storeToken.getOffset(), storeToken.getStoreKey(), messageEntries,
-                new FindEntriesCondition(maxTotalSizeOfEntries), indexSegments, storeToken.getInclusive());
+        StoreFindToken newToken;
+        //if index based token point to the last entry of last index segment, and journal size cleaned up after last log
+        //segment auto close, we should return the original index based token instead of getting from journal.
+        if (onlyContainsHeaderInLastLogSegment(storeToken.getOffset(), indexSegments)) {
```
justinlin-linkedin, line 1316: I have doubts on this logic. 

Assume we have such set up

[Index1] [Index2].   [Index3] [Index4].    [Empty Journal]
 [Log Segment 1]      [Log Segment 2]    [new Log segment with only header]

Assuming that last time findEntriesSince returns a IndexBased storeFindToken and the offset is index4.startOffset, but the storeKey is some key in the middle of index4, which means we haven't finished all the IndexEntries in index4. In this case, we shouldn't just assign the storeToken to newToken and move on, we should keep reading the index segments.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1643,8 +1643,10 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
         break; // we have entered and finished reading from the journal, so we are done.
       }
 
-      if (segmentStartOffset == validIndexSegments.lastKey() && !onlyContainsHeaderInLastLogSegment(
-          segmentStartOffset)) {
+      //if the journal based token point to the last entry of last index segment but journal cleaned up after to last log
+      //segment auto closed, we can bypass this check and return the index based token.
+      if (segmentStartOffset == validIndexSegments.lastKey() && !onlyContainsHeaderInLastLogSegment(segmentStartOffset,
```
justinlin-linkedin, line 1648: I think this should be changed to `segmentStartOffet.equals(validIndexSemgents.lastKey())`. Prior to your change, if we reach the last index segment, the `segmentStartOffset` should be extracted by calling `indexSegments.higherKey(segmentStartOffset)` at line 1602, but now, it might just be the offset in storeFindToken we can't just compare the reference any more, we have to compare the actual value.

justinlin-linkedin (changes requested):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: If token reached to the last entry of the last index segment, and the last log segment has not been auto closed, which means token reached to the end of current log segment, at this time, we should return the original journal based token instead of transform it into index based.(line 1297)

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1701,10 +1703,13 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
    * @param startOffset the index segment start {@link Offset}.
    * @return {@link true} if last log segment only contains the header info after auto closing last log segment during compaction.
    */
-  private boolean onlyContainsHeaderInLastLogSegment(Offset startOffset) {
+  private boolean onlyContainsHeaderInLastLogSegment(Offset startOffset,
```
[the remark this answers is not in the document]
SophieGuo410: Updated to use "journal.getCurrentNumberOfEntries() == 0 && log.getLastSegment().isEmpty()".

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Updated the logic to check if the storeKey in indexBasedToken is point to the last entry of last index segment or not.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Updated the comments. I think when at beginning log segment is empty and index segment is empty, the start offset would be 18 and entry will be null since journal is null, the Uninitialized should be return directly which will reach line 1297.

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -1157,6 +1162,167 @@ public void tokenRevalidationTest() throws StoreException {
     }
   }
 
+  @Test
+  public void rebuildTokenWhenLastSegmentAutoClosedTestCase1() throws StoreException {
+    assumeTrue(isLogSegmented);
+    IndexSegment lastIndexSegment = stateForTokenTest.index.getIndexSegments().lastEntry().getValue();
+    //generate a journal based token mapping to first indexEntry in last index segment. it will return the next journal based token in journal.
+    Offset validOffset =
+        new Offset(lastIndexSegment.getLogSegmentName(), lastIndexSegment.getStartOffset().getOffset());
```
justinlin-linkedin, line 1171: this can just be
```
Offset validOffset = lastIndexSegment.getStartOffset();
```

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -1157,6 +1162,167 @@ public void tokenRevalidationTest() throws StoreException {
     }
   }
 
+  @Test
+  public void rebuildTokenWhenLastSegmentAutoClosedTestCase1() throws StoreException {
```
justinlin-linkedin, line 1166: this test case has nothing to do with auto-close last segment? Am I missing anyting?

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -1157,6 +1162,167 @@ public void tokenRevalidationTest() throws StoreException {
...
+    StoreFindToken token = (StoreFindToken) findInfo.getFindToken();
+    StoreFindToken expectedToken =
+        new StoreFindToken(nextOffset, stateForTokenTest.sessionId, stateForTokenTest.incarnationId, false,
+            lastIndexSegment.getResetKey(), lastIndexSegment.getResetKeyType(),
+            lastIndexSegment.getResetKeyLifeVersion());
+    compareTokens(expectedToken, token);
+  }
+
+  @Test
+  public void rebuildTokenWhenLastSegmentAutoClosedTestCase2() throws StoreException {
```
justinlin-linkedin, line 1187: so is this one.

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -1157,6 +1162,167 @@ public void tokenRevalidationTest() throws StoreException {
...
+    StoreFindToken expectedToken;
+    expectedToken =
+        new StoreFindToken(firstOffsetInJournal, stateForTokenTest.sessionId, stateForTokenTest.incarnationId, false,
+            lastIndexSegment.getResetKey(), lastIndexSegment.getResetKeyType(),
+            lastIndexSegment.getResetKeyLifeVersion());
+    compareTokens(expectedToken, token);
+  }
+
+  @Test
+  public void rebuildTokenWhenLastSegmentAutoClosedTestCase4() throws StoreException {
```
justinlin-linkedin, line 1227: This one is more like it. I suggest we put some comments here to describe what kind of tests are done in this testcase.

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -1157,6 +1162,167 @@ public void tokenRevalidationTest() throws StoreException {
...
+            lastIndexSegment.getResetKey(), lastIndexSegment.getResetKeyType(),
+            lastIndexSegment.getResetKeyLifeVersion());
+    autoCloseLastLogSegmentAndCleanUpJournal();
+    FindInfo findInfo = stateForTokenTest.index.findEntriesSince(startToken, 1);
+    StoreFindToken token = (StoreFindToken) findInfo.getFindToken();
+    StoreFindToken expectedToken =
+        new StoreFindToken(lastIndexSegment.iterator().next().getKey(), lastIndexSegment.getStartOffset(),
+            stateForTokenTest.sessionId, stateForTokenTest.incarnationId, lastIndexSegment.getResetKey(),
+            lastIndexSegment.getResetKeyType(), lastIndexSegment.getResetKeyLifeVersion());
+    compareTokens(expectedToken, token);
```
justinlin-linkedin, line 1244: can we use `expectedToken` as the StoreFindToken to call findEntriesSince method again. It's to test if we can move the token forward when the storeKey is not the first key in the last index segment.

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -1157,6 +1162,167 @@ public void tokenRevalidationTest() throws StoreException {
...
+    //generate a journal based token point to last entry of index segment, and close the last log segment and clean up journal
+    //add more entry in auto create log segment. it should return the index based token point to previous journal based token.
+    IndexSegment lastIndexSegment = stateForTokenTest.index.getIndexSegments().lastEntry().getValue();
+    Offset validOffset = stateForTokenTest.index.journal.getLastOffset();
+    StoreFindToken startToken =
+        new StoreFindToken(validOffset, stateForTokenTest.sessionId, stateForTokenTest.incarnationId, false,
+            lastIndexSegment.getResetKey(), lastIndexSegment.getResetKeyType(),
+            lastIndexSegment.getResetKeyLifeVersion());
+    autoCloseLastLogSegmentAndCleanUpJournal();
+    stateForTokenTest.addPutEntries(2, CuratedLogIndexState.PUT_RECORD_SIZE, Utils.Infinite_Time);
```
justinlin-linkedin, line 1300: I think we should add another test before adding PutEntries in the index segment. We can use journalbased token after autoCloseLastLogSegmentAndCleanUpJournal(), it would return the a IndexBased token withe last IndexSegment startOffset and first key .

justinlin-linkedin:
BTW, can you also type up some description for this PR, thanks

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1698,13 +1700,30 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
...
   /**
-   * @param startOffset the index segment start {@link Offset}.
+   *
+   * @param token the {@link StoreFindToken} need to be checked.
+   * @param indexSegments the map of index segment start {@link Offset} to {@link IndexSegment} instances.
+   * @return {@code true} if the token point to the last entry of last log segment. Otherwise return {@code false}.
+   */
+  private boolean indexBasedTokenPointToLastEntryOfLastIndexSegment(StoreFindToken token,
+      ConcurrentSkipListMap<Offset, IndexSegment> indexSegments) {
+    StoreKey storeKey = token.getStoreKey();
```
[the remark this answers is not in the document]
SophieGuo410: Sure, updated.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1698,13 +1700,30 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
   }
 
   /**
-   * @param startOffset the index segment start {@link Offset}.
+   *
+   * @param token the {@link StoreFindToken} need to be checked.
+   * @param indexSegments the map of index segment start {@link Offset} to {@link IndexSegment} instances.
+   * @return {@code true} if the token point to the last entry of last log segment. Otherwise return {@code false}.
```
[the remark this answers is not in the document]
SophieGuo410: Fixed.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1698,13 +1700,30 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
   }
 
   /**
-   * @param startOffset the index segment start {@link Offset}.
+   *
+   * @param token the {@link StoreFindToken} need to be checked.
```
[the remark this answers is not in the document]
SophieGuo410: Sure, updated.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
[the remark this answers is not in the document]
SophieGuo410: Fixed the typo.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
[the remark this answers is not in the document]
SophieGuo410: Fixed the typo.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1569,8 +1571,9 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
       List<MessageInfo> messageEntries, FindEntriesCondition findEntriesCondition,
       ConcurrentSkipListMap<Offset, IndexSegment> indexSegments, boolean inclusive) throws StoreException {
     Offset segmentStartOffset = initialSegmentStartOffset;
-    if (segmentStartOffset.equals(indexSegments.lastKey()) && !onlyContainsHeaderInLastLogSegment(segmentStartOffset)) {
-      // We would never have given away a token with a segmentStartOffset of the latest segment.
+    if (segmentStartOffset.equals(indexSegments.lastKey()) && !onlyContainsHeaderInfoInLastLogSegment()) {
+      //We would never have given away a token with a segmentStartOffset of the latest segment unless the journal has
+      //been cleaned up by (auto close last log segment).
```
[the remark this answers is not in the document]
SophieGuo410: Added more comments for it.

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
SophieGuo410: Included the details in comments.

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
SophieGuo410: Sure, thx for catching up this issue. Actually I added more test to let the token moving forward until it reaches to the end.

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
SophieGuo410: Actually I've already tested it. since add the 2 new entries to last log segment won't affecting the token generation cause if we put journal based token, it will return the indexed based token withe last IndexSegment startOffset and first key. But to be more clear, I move the addPutEntries below to the check, so it will be more clear.

justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1318,9 +1313,16 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
         // Find the index segment corresponding to the token indexStartOffset.
         // Get entries starting from the token Key in this index.
         startTimeInMs = time.milliseconds();
-        StoreFindToken newToken =
-            findEntriesFromSegmentStartOffset(storeToken.getOffset(), storeToken.getStoreKey(), messageEntries,
-                new FindEntriesCondition(maxTotalSizeOfEntries), indexSegments, storeToken.getInclusive());
+        StoreFindToken newToken;
+        //if index based token point to the last entry of last index segment, and journal size cleaned up after last log
+        //segment auto close, we should return the original index based token instead of getting from journal.
```
jsjtzyy, line 1318: I also wonder how index-based token would point to last index segment in normal case. Is it because previous journal-based token gets converted to index-based after last log segment is closed?

jsjtzyy (approved):
LGTM after minor comments are addressed.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1318,9 +1313,16 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
         // Find the index segment corresponding to the token indexStartOffset.
         // Get entries starting from the token Key in this index.
         startTimeInMs = time.milliseconds();
-        StoreFindToken newToken =
-            findEntriesFromSegmentStartOffset(storeToken.getOffset(), storeToken.getStoreKey(), messageEntries,
-                new FindEntriesCondition(maxTotalSizeOfEntries), indexSegments, storeToken.getInclusive());
+        StoreFindToken newToken;
+        //if index based token point to the last entry of last index segment, and journal size cleaned up after last log
```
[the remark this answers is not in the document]

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1290,7 +1279,13 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
...
           Map.Entry<Offset, IndexSegment> entry = indexSegments.floorEntry(offsetToStart);
           StoreFindToken newToken;
-          if (entry != null && !entry.getKey().equals(indexSegments.lastKey())) {
+          //There are several reasons that we could reach here:
+          //1.StoreFindToken is Uninitialized, and we have more index segments than journal can cover.
+          //2.StoreFindToken is JournalBased, but the journal somehow moved too much ahead, the offset in the FindToken is no longer within journal.
+          //  2.1 There might just be lots of new entries coming to journal so the journal moves ahead.
+          //  2.2 The journal might be cleared and new log segment is created for compaction (auto-close last segment)
+          if (entry != null && !entry.getKey().equals(indexSegments.lastKey())
+              || onlyContainsHeaderInfoInLastLogSegment()) {
```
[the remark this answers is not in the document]

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1698,13 +1705,33 @@ private StoreFindToken findEntriesFromSegmentStartOffset(Offset initialSegmentSt
...
+  }
+
+  /**
    * @return {@link true} if last log segment only contains the header info after auto closing last log segment during compaction.
    */
-  private boolean onlyContainsHeaderInLastLogSegment(Offset startOffset) {
-    logger.trace("Log segment name {} for start offset {}", startOffset.getName(), startOffset);
-    logger.trace("Current last log segment name {} when check start offset {}", log.getLastSegment().getName(), startOffset);
-    return config.storeAutoCloseLastLogSegmentEnabled && startOffset.getName().compareTo(log.getLastSegment().getName()) < 0;
+  private boolean onlyContainsHeaderInfoInLastLogSegment() {
```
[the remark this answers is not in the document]

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Discussed offline and point to the right place where the journal based token transformed to index based token.

PR #1900, opened 2021-07-03
Remove index Entry not belongs to last log segment from journal during bootstrap.

SophieGuo410:
Journal should only contain the index entry belongs to last log segment, otherwise the second last log segment will not be compacted.


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -498,7 +502,13 @@ void addToIndex(IndexEntry entry, FileSpan fileSpan) throws StoreException {
     } else {
       validIndexSegments.lastEntry().getValue().addEntry(entry, fileSpan.getEndOffset());
     }
-    journal.addEntry(entry.getValue().getOffset(), entry.getKey(), entry.getCrc());
+    if (config.storeAutoCloseLastLogSegmentEnabled && isForRecovery) {
```
zzmao, line 505: what if the config was True and then change to False? Can we just look at `isForRecovery`?

zzmao (commented):
One comment

zzmao (approved):
LGTM

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -478,12 +479,15 @@ void changeIndexSegments(List<File> segmentFilesToAdd, Set<Offset> segmentsToRem
...
+   * Adds a new entry to the index. For recovery, only the index segments belongs to last log segment should be put into
+   * journal, other wise it won't be compacted and we will facing the issue when last log segment is empty but journal size
+   * is non-zero.
    * @param entry The entry to be added to the index
    * @param fileSpan The file span that this entry represents in the log
+   * @param isForRecovery Determine if the addToIndex method is for recovery.
    * @throws StoreException
    */
-  void addToIndex(IndexEntry entry, FileSpan fileSpan) throws StoreException {
+  void addToIndex(IndexEntry entry, FileSpan fileSpan, boolean isForRecovery) throws StoreException {
```
justinlin-linkedin, line 490: all of the function invocations has the last parameter set to false except for one. I suggest in this case, we can add a new method with `isForRecovery` being false by default. In this way, we only have to change one function invocation.

justinlin-linkedin: BTW, can you change `isForRecovery` to `recovering`, just to use the same variable name as BlobStoreCompactor.commit.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -376,7 +376,8 @@ private void recover(MessageStoreRecovery recovery) throws StoreException, IOExc
           IndexValue newValue = new IndexValue(info.getSize(), runningOffset, IndexValue.FLAGS_DEFAULT_VALUE,
               info.getExpirationTimeInMs(), info.getOperationTimeMs(), info.getAccountId(), info.getContainerId(),
               info.getLifeVersion());
-          addToIndex(new IndexEntry(info.getStoreKey(), newValue, null), new FileSpan(runningOffset, infoEndOffset));
+          addToIndex(new IndexEntry(info.getStoreKey(), newValue, null), new FileSpan(runningOffset, infoEndOffset),
+              true);
```
justinlin-linkedin, line 380: I don't understand this? Why is this the only place that we set the last parameter to true? In this entire `recover` method, all the `IndexEntry` added to the persistent index should have the last parameter set to true. Like line 336, line 344, line 360. 
The problem is that those method `markAsDeleted`, `markAsUndelete` and `markAsPermanent` has another wrapper method around `addToIndex`, so it would be hard to change all these methods. So I suggest that we create a new object field `inRecovering` and set it true when starting recovering and set it false when finishing recovering.

justinlin-linkedin (changes requested):

justinlin-linkedin:
BTW, would be great if you can add some tests

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Sure. updated it and move the clean up journal method outside of recovery.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -212,6 +214,9 @@ public int compare(IndexEntry e1, IndexEntry e2) {
       if (cleanShutdown) {
         cleanShutdownFile.delete();
       }
+      //Journal should only include the index entry which belongs to the last log segment to make sure it only prevents the last log segment
+      //from being compacted.
+      removeIndexEntryNotBelongsToLastLogSegmentFromJournal();
```
justinlin-linkedin, line 219: we can do this after recovery, so we don't need to have this `inRecovery` variable.

justinlin-linkedin (approved):

PR #1901, opened 2021-07-12
Return a task result object from container deletion sync task.

ankagrawal:
We were returning null as `TaskResult` in `DeprecatedContainerCloudSyncTask`. While this doesn't affect the correctness for cloud container deletion, the `TaskRunner` of helix errors out on seeing null `TaskResult`. Hence the helix task is marked as failed, even though the task actually passed.
This PR fixes the issue by returning an appropriate object of `TaskResult` based on whether the task failed or passed.


SophieGuo410 (approved):
lgtm!

Arun-LinkedIn (approved):

PR #1904, opened 2021-07-12
Use jackson annotations to serialize and deserialize Account and Container

justinlin-linkedin:
Use jackson annotation to serialize and deserialize Account and Container, so we don't have to write toJson and fromJson method each time we have a new field added to account or container.
Still keep fromJson and toJson method to make sure this PR won't change anything.
Will remove it after it's tested with requests.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/AccountInfoMap.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AccountInfoMap.java
@@ -13,12 +13,14 @@
  */
 package com.github.ambry.account;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
-import org.json.JSONArray;
-import org.json.JSONException;
 import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -41,6 +43,7 @@ class AccountInfoMap {
   private final Map<Short, Account> idToAccountMap;
   // used to track last modified time of the accounts and containers in this cache
   private long lastModifiedTime = 0;
+  private final ObjectMapper objectMapper = new ObjectMapper();
 
   /**
    * Constructor for an empty {@code AccountInfoMap}.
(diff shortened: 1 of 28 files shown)


ambry-api/src/main/java/com/github/ambry/account/AccountCollectionSerde.java
```diff
@@ -58,6 +72,14 @@ public static JSONObject accountsToJson(Collection<Account> accounts) {
     }
   }
 
+  public static Collection<Account> accountsFromInputStreamInJson(ObjectMapper objectMapper, InputStream inputStream)
```
cgtz, line 75: Could we have a single static ObjectMapper for accounts that is used for all account/container ser/deser?

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
```diff
@@ -24,13 +27,15 @@
  * in two ways: 1) from an existing {@link Container} object; and 2) by supplying required fields of a {@link Container}.
  * This class is not thread safe.
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonPOJOBuilder(withPrefix="set")
 public class ContainerBuilder {
   // necessary
   private short id;
```
cgtz, line 34: We should take a careful look at how default values of primitives will behave when using the ContainerBuilder to deserialize json. For example, before an error would be thrown if id was not present in the json. Using jackson, it seems like a `Container` would be created with the default short value of 0 as the id.

cgtz (commented):
Thanks for starting this migration! Standardizing on jackson would be very helpful for the project

ambry-account/src/main/java/com/github/ambry/account/AccountInfoMap.java
```diff
@@ -41,6 +43,7 @@
   private final Map<Short, Account> idToAccountMap;
   // used to track last modified time of the accounts and containers in this cache
   private long lastModifiedTime = 0;
+  private final ObjectMapper objectMapper = new ObjectMapper();
```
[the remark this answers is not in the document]
justinlin-linkedin: updated

ambry-account/src/main/java/com/github/ambry/account/AccountInfoMap.java
```diff
@@ -61,27 +64,42 @@
...
+              "Invalid account record when reading accountMap because idKey and accountId do not match. idKey=" + idKey
+                  + " accountId=" + account.getId());
+        }
+        if (idToAccountMap.containsKey(account.getId()) || nameToAccountMap.containsKey(account.getName())) {
+          throw new IllegalStateException(
+              "Duplicate account id or name exists. id=" + account.getId() + " name=" + account.getName());
+        }
+        idToAccountMap.put(account.getId(), account);
+        nameToAccountMap.put(account.getName(), account);
+      } catch (JsonProcessingException e) {
```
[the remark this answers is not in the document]
justinlin-linkedin: updated

ambry-api/src/main/java/com/github/ambry/account/AccountCollectionSerde.java
justinlin-linkedin: updated

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
justinlin-linkedin: I initialize the container id and container name with a special static value  UNINITIALIZED_CONTAINER_ID and UNINITIALIZED_CONTAINER_NAME. So that if container id or container name is not present in json string, we can throw an exception. I did the same thing for account id and account name, but not other fields.

ambry-api/src/main/java/com/github/ambry/account/AccountCollectionSerde.java
```diff
@@ -99,4 +141,25 @@ public static JSONObject containersToJson(Collection<Container> containers) {
...
   }
+
+  /**
+   * Deserialize a collection of {@link Container}s in json from given InputStream.
+   * @param inputStream the {@link InputStream} that contains serialized json bytes.
+   * @param accountId the account id for these containers.
+   * @return a {@link Collection} of {@link Container}s.
+   */
+  public static Collection<Container> containersFromInputStreamInJson(InputStream inputStream, short accountId)
+      throws IOException {
```
zzmao, line 152: NIT: IOException won't happen here. We can remove it.

ambry-api/src/main/java/com/github/ambry/account/AccountCollectionSerde.java
```diff
@@ -58,6 +77,18 @@ public static JSONObject accountsToJson(Collection<Account> accounts) {
     }
   }
 
+  /**
+   * Deserialize a collection of {@link Account} in json from given InputStream.
+   * @param inputStream the {@link InputStream} that contains serialized json bytes.
+   * @return a {@link Collection} of {@link Account}s.
+   */
+  public static Collection<Account> accountsFromInputStreamInJson(InputStream inputStream) throws IOException {
```
zzmao, line 85: Nit: IOException is not used.

zzmao (approved):
LGTM, two supper minor comments.

ambry-api/src/main/java/com/github/ambry/account/AccountCollectionSerde.java
justinlin-linkedin: ObjectMapper.readValue should still throw out IOException. We have to keep it.

ambry-api/src/main/java/com/github/ambry/account/AccountCollectionSerde.java
justinlin-linkedin: same

cgtz (approved):
LGTM, just one optional comment, but it's up to you which way you prefer to go.

ambry-api/src/main/java/com/github/ambry/account/AccountBuilder.java
```diff
@@ -34,8 +34,8 @@
 @JsonPOJOBuilder(withPrefix = "")
 public class AccountBuilder {
   private short id = UNINITIALIZED_ACCOUNT_ID;
```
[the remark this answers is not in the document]
justinlin-linkedin: Sure, the Short object is better. Updated.

PR #1905, opened 2021-07-14
Fix compaction error on PUT with lifeVersion higher than 0

justinlin-linkedin:
When we have a PUT index value with lifeVersion higher than 0 to copy in compaction, we ignore the lifeVersion. This is a bug, we have to copy the lifeVersion as well.

### DIFF
--- a/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
+++ b/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
@@ -1492,6 +1492,10 @@ public class BlobStoreCompactorTest {
     state.addUndeleteEntry((MockId) p12.getKey(), (short) 2);
     state.addPutEntries(1, PUT_RECORD_SIZE, Utils.Infinite_Time);
 
+    // Index Segment 1.3 P2 -> P2
+    IndexEntry p13 = state.addPutEntries(1, PUT_RECORD_SIZE, Utils.Infinite_Time, (short)2).get(0);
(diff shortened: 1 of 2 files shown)


ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
```diff
@@ -1617,9 +1621,9 @@ public void undeleteSameIndexSegmentTest_NoTtlUpdate() throws Exception {
       compactor.close(0);
     }
     // Before compaction, the records in the log are
-    // P11 U11 P P P| P12 D12 U12 U12 P
+    // P11 U11 P P P| P12 D12 U12 U12 P | P13(2) P P P P
```
jsjtzyy, line 1624: Could you explain what the `P P P` are? (other PUT records?). Also why last 4 "P"s get removed after compaction? (Are they duplicate PUT records ?)

jsjtzyy (approved):
Thanks for fixing this bug.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
justinlin-linkedin: they are just random PUTs. The reason we have random puts here, is to make sure each index segment deal with one case of compaction. And the last 4 PUTs are not compacted. I just didn't bother to put it here since they are random PUTs.

zzmao (approved):
LGTM

PR #1906, opened 2021-07-15
Use fine grained caching and remove travis remnants

cgtz:
- Remove travis-CI related files and support for disabling publishing
  from github actions
- Use burrunan/gradle-cache-action to support fine grained caching of
  individual gradle task outputs instead of requiring a full directory
  upload of all cached artifacts. This should hopefully speed up
  builds for small changes.

More details about [burrunan/gradle-cache-action](https://github.com/burrunan/gradle-cache-action): This action starts up a local web server that supports the gradle remote cache protocol. It will contact the github cache under the hood but treat each individual gradle cache entry as a separate github cache entry instead of packaging all of the cache artifacts together like the default caching action does. This saves cache space, reduces unnecessary cache invalidation, and allows for parallel tasks to access cached data from each other.


.github/workflows/github-actions.yml
```diff
@@ -116,82 +92,64 @@ jobs:
...
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && env.PUBLISH == 'true'}}
 
-      - name: Cleanup Gradle Cache
-          # Remove some files from the Gradle cache which might have cached by earlier builds.
-        run: |
-          rm -f ~/.gradle/caches/modules-2/modules-2.lock
-          rm -f ~/.gradle/caches/modules-2/gc.properties
-        if: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && env.PUBLISH == 'true'}}
+      - name: Publish To JFrog artifactory
+        run: ./travis-publish.sh
```
Arun-LinkedIn, line 134: Typo for removing publishing command.

Arun-LinkedIn (approved):
Looks good to me! Just a typo that might be needs to be removed.

.github/workflows/github-actions.yml
cgtz: Thanks for catching this! I will remove it.

PR #1909, opened 2021-07-16
[Compaction] Fix NPE when TTL_UPDATE has not prior PUT

justinlin-linkedin:
This PR would fix a NPE caused by TTL_UPDATE index value having no PUT index value before. This case could happen when the PUT and TTL_UPDATE don't belong to the same log segment, but PUT's log segment is compacted and PUT is gone, while TTL_UPDATE is still there. 

The order of events that causes this case:
1. Create PUT in log segment 1
2. Create TTL_UPDATE in log segment 2
3. Create DELETE in log segment 3
4. DELETE is out of retention
5. Compact log segment 1 and PUT is removed
6. Compact log segment 3 and DELETE id removed because it's now a tombstone.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -1627,12 +1627,6 @@ class BlobStoreCompactor {
         } else {
           previousKey = currentKey;
           previousLatestState = currentLatestState = srcIndex.findKey(currentKey);
-          if (currentLatestState == null) {
-            // Only one case this would happen, an TTL_UPDATE indexValue that has no PUT before, this is because the put
-            // is already compacted but TTL_UPDATE is not compacted.
-            previousLatestState = currentLatestState =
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1655,6 +1655,12 @@ public boolean alreadyExists(PersistentIndex idx, FileSpan searchSpan, StoreKey
             }
           }
         } else if (currentValue.isTtlUpdate()) {
+          if (currentLatestState == null) {
+            // Only one case this would happen, an TTL_UPDATE indexValue that has no PUT before, this is because the put
```
jsjtzyy, line 1659: `an TTL_UPDATE indexValue that has no PUT before and no DELETE at the end, this is because ...`

jsjtzyy (approved):

SophieGuo410 (approved):
Thanks for fixing this issue!

PR #1911, opened 2021-07-19
Adding more log info and constraints for index segment persist

SophieGuo410:
This pr supports:
1.  Add more log info for easy trace.
2. Compaction might seal the last index segment, added more constraint to make sure only unsealed index segment will be persisted by this method -> writeIndexSegmentToFile

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -2279,6 +2279,7 @@ class PersistentIndex {
             toWrite.writeIndexSegmentToFile(toWrite.getEndOffset());
             toWrite.seal();
           }
+          //Compaction may seal the current last index segment.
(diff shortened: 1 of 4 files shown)


justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -2279,7 +2279,10 @@ public synchronized void write() throws StoreException {
             toWrite.writeIndexSegmentToFile(toWrite.getEndOffset());
             toWrite.seal();
           }
-          currentInfo.writeIndexSegmentToFile(indexEndOffsetBeforeFlush);
+          //Compaction may seal the current last index segment.
+          if (!currentInfo.isSealed()) {
+            currentInfo.writeIndexSegmentToFile(indexEndOffsetBeforeFlush);
```
zzmao, line 2284: Since compaction may seal it, this is not thread safe: compaction many seal it between 2282 and 2283.

zzmao (approved):
LGTM

PR #1912, opened 2021-07-19
[Quota] Add logic to turn request quota on and off individually

justinlin-linkedin:
Before this PR, we have "requestThrottlingEnabled" to control all the enforcers in the ambry quota manager. So it's very hard to turn off request quota enforcer at the same time keep storage quota enforcer. 
After this PR, we change the semantic of "requetThrottlingEnabled" to only control the request quota enforcer. In way, if we want to only enable throttling on storage quota enforcer, we should set this flag to false and request quota enforcer would continue working on tracking mode.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
@@ -37,12 +37,6 @@ public class QuotaConfig {
   public static final boolean DEFAULT_THROTTLE_IN_PROGRESS_REQUESTS = false;
   public StorageQuotaConfig storageQuotaConfig;
 
-  /**
-   * Config to enable throttling on customer's account or container.
-   */
-  @Config(REQUEST_THROTTLING_ENABLED)
-  @Default("false")
-  public boolean requestThrottlingEnabled;
 
   /**
    * Serialized json containing pairs of enforcer classes and corresponding source classes.
@@ -73,10 +67,18 @@ public class QuotaConfig {
   public String quotaManagerFactory;
 
   /**
-   * The mode in which quota throttling is being done (TRACKING/THROTTLING).
+   * The mode in which quota throttling is being done (TRACKING/THROTTLING). To throttle the requests, you have to change
+   * the mode to THROTTLING and turn on requestThrottlingEnabled or storageQuotaConfig.shouldThrottle (or both).
    */
   @Config(THROTTLING_MODE)
   public QuotaMode throttlingMode;
+
+  /**
+   * Config to enable request throttling on customer's account or container.
+   */
+  @Config(REQUEST_THROTTLING_ENABLED)
+  @Default("true")
+  public boolean requestThrottlingEnabled;
 
   /**
    * Should requests in progress be throttled if they exceed their quota.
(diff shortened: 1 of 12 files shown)


ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
```diff
@@ -73,11 +67,19 @@
...
+   * the mode to THROTTLING and turn on requestThrottlingEnabled or storageQuotaConfig.shouldThrottle (or both).
    */
   @Config(THROTTLING_MODE)
   public QuotaMode throttlingMode;
 
+  /**
+   * Config to enable request throttling on customer's account or container.
+   */
+  @Config(REQUEST_THROTTLING_ENABLED)
+  @Default("true")
```
ankagrawal, line 80: Why is request throttling true by default. I think it should be false by default, and we can make it true where we need it to be true.

ambry-quota/src/main/java/com/github/ambry/quota/MaxThrottlePolicy.java
```diff
@@ -37,14 +38,31 @@
   // Percentage usage level at or below this limit and above warning limit will generate critical warning. Anything above will be fatal.
   static final int CRITICAL_USAGE_LEVEL_LIMIT = 100;
 
+  private final QuotaConfig quotaConfig;
+
+  /**
```
ankagrawal, line 43: Caan you also update the javadocs of this class.

ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
justinlin-linkedin: The idea is that Mode controls whether we want to throttle or not, so by default, when we add an enforcer to the list, we are expecting it follows the global settings. So if we don't want it follow the setting, we will set it on purpose. Thus the default value should be "true".

ankagrawal (approved):

ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
Arun-LinkedIn: Just wanted to check that this means that we now have separate config to enable/disable for each quota type (`quotaConfig.requestThrottlingEnabled` for read/write quota and `quotaConfig.storageQuotaConfig.shouldThrottle` for storage quota)and there is no global config to enable/disable all quotas.  

Just a question, would it be beneficial to also have a global config to enable/disable all quotas (with default value as true)?

Arun-LinkedIn (approved):
Looks good to me. Just had one question.

ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
justinlin-linkedin: @Arun-LinkedIn We do have a global configuration, but not to turn off the quota enforcers, but to change the mode of quota enforcers. Currently there are two modes in quota enforcers, Tracking and Throttling. We have a global configuration to pick either mode for all quota enforcers. The default mode is Tracking, which means we are not going to throttle any request.

PR #1914, opened 2021-07-21
If crc is null, don't error out as COLLISION

SophieGuo410:
Previously if the message info crc is null or journal crc is null, inside the checkWriteSetStateInStore we will treat it as COLLIDING. Actually when the replication data is outside of journal, the crc could be null and in this case we will treat it as duplicate instead of COLLISION, because we will check if it's the duplicate data by below constraints:
index.findKey(info.getStoreKey(), fileSpan,
          EnumSet.of(PersistentIndex.IndexEntryType.PUT, PersistentIndex.IndexEntryType.DELETE)) != null 
And since we are not able to get the crc, we can't determine it as COLLISION.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -917,7 +917,7 @@ class PersistentIndex {
    */
   boolean wasRecentlySeen(MessageInfo info) {
     Long crcInJournal = journal.getCrcOfKey(info.getStoreKey());
-    return info.getCrc() != null && info.getCrc().equals(crcInJournal);
+    return info.getCrc().equals(crcInJournal);
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -911,13 +911,14 @@ class PersistentIndex {
   }
 
   /**
-   * Returns true if the given message was recently seen by this Index.
+   * Returns true if the given message was recently seen by this Index or crc is null from either {@link MessageInfo} or
+   * {@link Journal}.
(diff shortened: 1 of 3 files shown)


ambry-store/src/test/java/com/github/ambry/store/BlobStoreTest.java
```diff
@@ -1535,6 +1549,13 @@ public void putIdCollisionTest() throws Exception {
     // Ensure that all new entries were added.
     missingKeysAfter.clear();
     assertEquals(missingKeysAfter, store.findMissingKeys(allMockIdList));
+
+    // 5. first one absent, second one colliding and cleaned up journal.
```
justinlin-linkedin, line 1553: nit: we can better word the comments here, something like // 5. first be present, second one colliding but the journal is cleaned up. BTW, in case 4, we already put allMockIdList.get(3) to the store, so the first one is not absent.

justinlin-linkedin (approved):

justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -917,7 +917,7 @@ void validateSanityForUndeleteWithoutLifeVersion(StoreKey key, List<IndexValue>
    */
   boolean wasRecentlySeen(MessageInfo info) {
     Long crcInJournal = journal.getCrcOfKey(info.getStoreKey());
-    return info.getCrc() != null && info.getCrc().equals(crcInJournal);
+    return info.getCrc().equals(crcInJournal);
```
zzmao, line 920: It's weird to remove the null check. 

Is it possible to add `index.journal.getCrcOfKey(info.getStoreKey()) == null` in the function. You can also rename `wasRecentlySeen`.

zzmao (approved):
LGTM

PR #1915, opened 2021-07-21
Fix content length clamping issue in frontend

cgtz:
Netty has two  content length related helper methods:
  public static long getContentLength(HttpMessage message, long defaultValue)
  public static int getContentLength(HttpMessage message, int defaultValue)
We were calling getContentLength(message, -1) instead of getContentLength(message, -1L)
in a couple places, which lead to the length used to tell when the request is fully sent
out getting clamped to an int, which made requests pause after 2 GB of
data was sent.

After fixing this, the download  of greater than 2 GB objects with the content-length
response headers works. This bug did not affect
transfer-encoding:chunked serving since a different end condition for
telling when the response is fully sent is used there.


zzmao (approved):
LGTM!

justinlin-linkedin (approved):
awesome

robby:
🚢

PR #1917, opened 2021-07-22
If previous stats based didn't run successfully, still need to close last log segment

SophieGuo410:
If the stats based compaction has not been kicked off because the details are null, we still need to auto close the last log segment.


zzmao (approved):
LGTM

justinlin-linkedin (approved):
LGTM

PR #1918, opened 2021-07-23
Remove redundant logic for auto closing log segment and fix the logic when recovery failed.

SophieGuo410:
This pr supports:
1.  refactor the getLogSegmentsNotInJournal() method and remove redundant code.
2. if recover successfully, the active log segment should be the last log segment, otherwise, it should be log segment matching with last index segment, especially when recovery failed.


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -220,12 +220,8 @@ public int compare(IndexEntry e1, IndexEntry e2) {
       //from being compacted.
       removeIndexEntryNotBelongsToLastLogSegmentFromJournal();
       setEndOffsets();
-      if (config.storeAutoCloseLastLogSegmentEnabled) {
-        //if log segment is generated by auto close, at some point there's no index segment.
-        if (log.getEndOffset().getName().compareTo(getCurrentEndOffset().getName()) < 0) {
-          log.setActiveSegment(getCurrentEndOffset().getName());
-        }
-      } else {
```
justinlin-linkedin, line 228: this part really confuses me. So after you made this change, say if we have a auto close log segment without any new entry, assuming the last log segment that has entry is 10_0 and the new auto closed log segment is 11_0. so the log.getEndOffset().getName() would be 11_0 and getCurrentEndOffset().getName() is 10_0. The log end offset will already greater than the index end offset. Then how is this possbile?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1537,32 +1533,21 @@ private long getAbsolutePositionInLogForOffset(Offset offset,
...
         if (!logSegment.getName().equals(firstOffsetInJournal.getName())) {
           logSegmentNamesToReturn.add(logSegment.getName());
         } else {
           break;
         }
-        logSegment = log.getNextSegment(logSegment);
       }
+      logSegment = log.getNextSegment(logSegment);
     }
     return logSegmentNamesToReturn.size() > 0 ? logSegmentNamesToReturn : null;
```
justinlin-linkedin, line 1552: this is getting a bit complicated, can we just do something like
```
ArrayList<LogSegmentName> allLogSegments = new ArrayList<>();
LogSegment logSegment = log.getFirstSegment();
while(logSegment != null) {
    allLogSegment.add(logSegment.getName());
}
if (journal.isEmpty()) {
    return allLogSegments;
} else {
    Offset firstOffsetInJournal = journal.getFirstOffset();
    int index = allLogSegments.indexOf(firstOffsetInJournal.getName());
    if (index != -1) {
        return allLogSegments.subList(0, index);
    } else {
        return allLogSegments;
    }
}
```

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Yeah, I'm just trying to make this logic backward compatible.  The original design without my auto close log segment changes has this kind of logic: "log.setActiveSegment(getCurrentEndOffset().getName());" But for my case, I think it should never happened. And since during the log initialization, the active segment will be set as the last log segment, so maybe we can remove this part of logic? Let me know which way you prefer.

SophieGuo410: As I double checked, previously when index is missing, the corresponding logs should be gone. But since we have the recovery logic and logs should not be gone, instead, we should recover the index based on the log, is it correct? So I guess I can remove that part of logic and let the log initialization to set the active log segment.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Updated the logic with some minor difference.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: After a second thought, I realize that if the recovery failed, we need to set the active log segment to the last index segment related log segment. And if the recovery succeed and last log segment is empty, we should set the last log segment as active segment. Fixed the logic.

zzmao (approved):
LGTM

justinlin-linkedin (approved):

PR #1921, opened 2021-07-27
Add method to change quota mode in quota manager

justinlin-linkedin:
This PR adds two new methods to `QuotaManager` to set and return the quota mode so that in other places we can change quota mode based on other conditions.


ankagrawal (approved):
lgtm

Arun-LinkedIn (approved):
Looks good to me.

PR #1923, opened 2021-07-28
Adding test cases and relax constraint for setting active segment in PersistentIndex

SophieGuo410:
Previously when getCurrentEndOffset().equals(currentSecondLastLogSegmentEndOffSet) && log.getLastSegment().isEmpty(), we will notice that the last log segment don't have data and corresponding index, so at that time, we will set last log segment as active segment. There's one more case that if we have two log segments which has been auto created, the last log segment still need to be set as active segment, and the second last log segment will be compacted through compaction later.

### DIFF
--- a/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
+++ b/ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
@@ -2935,11 +2935,10 @@ public class BlobStoreCompactorTest {
         compactor.close(0);
       }
     }
-    if (compactor.closeLastLogSegmentIfQualified()) {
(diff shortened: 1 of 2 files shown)


ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
```diff
@@ -3878,7 +3877,7 @@ public void testCloseLastLogSegmentIfQualified() throws Exception {
     //create first active log segment
     refreshState(false, false, true);
     //leave two log segment space for auto close last log segment purpose
```
zzmao, line 3879: Fix comment

zzmao (approved):
LGTM. One minor comment

justinlin-linkedin (approved):
LGTM

PR #1924, opened 2021-07-28
Change log message for azure connectivity error.

ankagrawal:
Add better message for azure connectivity related startup error.


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationFactory.java
```diff
@@ -65,7 +65,7 @@ public CloudDestination getCloudDestination() throws IllegalStateException {
       dest.testAzureConnectivity();
       return dest;
     } catch (Exception e) {
-      logger.error("Error initializing Azure destination: {}", e.getMessage());
+      logger.error("Error initializing connection Azure destination: {}", e.getMessage());
```
cgtz, line 68: "connection to Azure destination"?

Does the exception message already have sufficient info about which connection is failing?

PR #1926, opened 2021-07-30
Fix a bug in throttler

justinlin-linkedin:
This PR would fix a bug in the throttler.

The bug works like this
1. If we have a throttler to throttle rate at 100 per second and we call maybeThrottle for 1000
2. It will sleep for 10 seconds so the rate would be 100 per second.
3. If we call maybeThrottle for 500 right away, we are expecting it to sleep for 5 seconds, but it will not sleep at all. 

This is the bug, it's caused by the error when setting the `periodStartNs`. If we have a sleep in the previous maybeThrottle call, then the `periodStartNs` will be set to a wrong value.
For example, if at t = 0, we call maybeThrottle for 1000, then we have to sleep for 10 second, but since the `periodStartNs` is set to now, which is 0, when we wake up from the sleep, our time still remains at 10 second ago.

This PR would fix that.

### DIFF
--- a/ambry-utils/src/main/java/com/github/ambry/utils/Throttler.java
+++ b/ambry-utils/src/main/java/com/github/ambry/utils/Throttler.java
@@ -79,7 +79,7 @@ public class Throttler {
             }
           }
         }
-        periodStartNs = now;
+        periodStartNs = time.nanoseconds();
         observedSoFar = 0;
       }
     }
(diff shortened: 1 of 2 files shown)


zzmao (approved):
This is a small change but might impact multiple places who use this class. However, this change slows down the rate, which should be fine.

cgtz (approved):

cgtz:
This PR would fix a bug in the throttler.

The bug works like this

    If we have a throttler to throttle rate at 100 per second and we call maybeThrottle for 1000
    It will sleep for 10 seconds so the rate would be 100 per second.
    If we call maybeThrottle for 500 right away, we are expecting it to sleep for 5 seconds, but it will not sleep at all.

This is the bug, it's caused by the error when setting the periodStartNs. If we have a sleep in the previous maybeThrottle call,
then the periodStartNs will be set to a wrong value.
For example, if at t = 0, we call maybeThrottle for 1000, then we have to sleep for 10 second, but since the periodStartNs is
set to now, which is 0, when we wake up from the sleep, our time still remains at 10 second ago.

PR #1927, opened 2021-08-03
Get previousStats from mysql db instead of localBackupFile

SophieGuo410:
Previously we will get the previousStats from localBackUpFile, which will cause some consistency issue between localBackupFile and mysql database. This pr is target to fix this issue.

### DIFF
--- a/ambry-mysql/src/integration-test/java/com/github/ambry/accountstats/AccountStatsMySqlStoreIntegrationTest.java
+++ b/ambry-mysql/src/integration-test/java/com/github/ambry/accountstats/AccountStatsMySqlStoreIntegrationTest.java
@@ -120,7 +120,28 @@ public class AccountStatsMySqlStoreIntegrationTest {
   }
 
   @Test
-  public void testEmptyStats() throws Exception {
+  public void testEmptyStatsWhenReadingPreviousStatsFromMysqlDb() throws Exception {
+    //write a new stats into database.
+    AccountStatsMySqlStore mySqlStore = createAccountStatsMySqlStore(clusterName1, hostname1, port1);
+    StatsWrapper stats = generateStatsWrapper(1, 1, 1, StatsReportType.ACCOUNT_REPORT);
+    mySqlStore.storeAccountStats(stats);
+
+    StatsWrapper obtainedStats = mySqlStore.queryAccountStatsByHost(hostname1, port1);
+    assertTrue(obtainedStats.getSnapshot().getSubMap().containsKey(Utils.statsPartitionKey((short) 0)));
+
+    //initialized the mySqlStore and write a new stats with the same partition.
+    mySqlStore = createAccountStatsMySqlStore(clusterName1, hostname1, port1);
+    StatsWrapper stats2 = generateStatsWrapper(0, 0, 0, StatsReportType.ACCOUNT_REPORT);
+    stats2.getSnapshot().getSubMap().put(Utils.statsPartitionKey((short) 0), new StatsSnapshot(0L, null));
+    mySqlStore.storeAccountStats(stats2);
+
+    // empty stats should remove all the data in the database
+    obtainedStats = mySqlStore.queryAccountStatsByHost(hostname1, port1);
+    assertFalse(obtainedStats.getSnapshot().getSubMap().containsKey(Utils.statsPartitionKey((short) 0)));
+  }
+
+  @Test
+  public void testEmptyStatsWhenReadingPreviousStatsFromLocalBackUpFile() throws Exception {
     AccountStatsMySqlStore mySqlStore = createAccountStatsMySqlStore(clusterName1, hostname1, port1);
     StatsWrapper stats = generateStatsWrapper(10, 10, 1, StatsReportType.ACCOUNT_REPORT);
     stats.getSnapshot().getSubMap().put(Utils.statsPartitionKey((short) 10), new StatsSnapshot(0L, null));
--- a/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
@@ -23,8 +23,6 @@ import com.github.ambry.server.StatsHeader;
 import com.github.ambry.server.StatsSnapshot;
 import com.github.ambry.server.StatsWrapper;
 import com.github.ambry.utils.Utils;
-import com.zaxxer.hikari.HikariDataSource;
-import java.io.Closeable;
 import java.io.File;
 import java.sql.Connection;
 import java.sql.SQLException;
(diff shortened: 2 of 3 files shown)


ambry-mysql/src/integration-test/java/com/github/ambry/accountstats/AccountStatsMySqlStoreIntegrationTest.java
```diff
@@ -499,6 +521,9 @@ private AccountStatsMySqlStore createAccountStatsMySqlStore(String clusterName,
     configProps.setProperty(AccountStatsMySqlConfig.UPDATE_BATCH_SIZE, String.valueOf(batchSize));
     configProps.setProperty(AccountStatsMySqlConfig.POOL_SIZE, String.valueOf(5));
     configProps.setProperty(StatsManagerConfig.STATS_OUTPUT_FILE_PATH, localBackupFilePath.toString());
+    if (readPreviousStatsFromMysqlDb) {
+      configProps.setProperty(AccountStatsMySqlConfig.READ_PREVIOUS_STATS_FROM_MYSQL_DB, "true");
+    }
```
justinlin-linkedin, line 526: you can just do
```
configProps.setProperty(AccountStatsMySqlConfig.READ_PREVIOUS_STATS_FROM_MYSQL_DB, String.valueOf(readPreviousStatsFromMysqlDb));
```

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -166,14 +164,23 @@ public AccountStatsMySqlStore(AccountStatsMySqlConfig config, DataSource dataSou
     partitionClassReportsDao = new PartitionClassReportsDao(dataSource, mySqlMetrics);
     this.hostnameHelper = hostnameHelper;
     storeMetrics = new AccountStatsMySqlStore.Metrics(registry);
-    if (!Strings.isNullOrEmpty(localBackupFilePath)) {
-      // load backup file and this backup is the previous stats
-      ObjectMapper objectMapper = new ObjectMapper();
+    if (config.readPreviousStatsFromMysqlDb) {
```
cgtz, line 167: Is there a situation where this config should be disabled? If this is not a dangerous feature to roll out and we want to use it in all circumstances, I would prefer not having another config for this.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -166,14 +164,23 @@ public AccountStatsMySqlStore(AccountStatsMySqlConfig config, DataSource dataSou
...
+      long startTimeMs = System.currentTimeMillis();
       try {
-        this.previousStats = objectMapper.readValue(new File(localBackupFilePath), StatsWrapper.class);
+        this.previousStats = queryAccountStatsBySimplifiedHostName(hostname, startTimeMs);
```
cgtz, line 170: why is the hostname already simplified in this situation? I may be missing something

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -313,6 +320,18 @@ private void deleteContainerAccountStats(Map<String, StatsSnapshot> prevPartitio
...
+
+  /**
+   * Query mysql database to get all the container storage usage for given {@code clusterName} and {@code queryHostname} and
+   * construct a {@link StatsSnapshot} from them.
+   * @param queryHostname The simplified hostname to query.
+   * @param startTimeMs The start time to query account stats by host.
+   * @return {@link StatsSnapshot} published by the given simplified hostname.
+   * @throws SQLException
+   */
+  private StatsWrapper queryAccountStatsBySimplifiedHostName(String queryHostname, long startTimeMs) throws SQLException {
```
justinlin-linkedin, line 334: we don't have to pass startTimeMs to this method, we can just call System.currentTimeMillis() to get this value within this method. it's fine to do this for `queryAccountStatsByHost` method since the line in 322 takes no time to finish.

justinlin-linkedin (approved):

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
```diff
@@ -166,14 +164,23 @@ public AccountStatsMySqlStore(AccountStatsMySqlConfig config, DataSource dataSou
...
-    if (!Strings.isNullOrEmpty(localBackupFilePath)) {
-      // load backup file and this backup is the previous stats
-      ObjectMapper objectMapper = new ObjectMapper();
+    if (config.readPreviousStatsFromMysqlDb) {
+      long startTimeMs = System.currentTimeMillis();
       try {
-        this.previousStats = objectMapper.readValue(new File(localBackupFilePath), StatsWrapper.class);
+        this.previousStats = queryAccountStatsBySimplifiedHostName(hostname, startTimeMs);
       } catch (Exception e) {
         this.previousStats = null;
```
cgtz, line 172: if the load from mysql fails, do we want to fall back to reading from the local backup file?

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
SophieGuo410: The hostname has already been simplified in AccountStatsMySqlStoreFactory line 59 right?

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
SophieGuo410: Sure, I think we can do that. Updated.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
SophieGuo410: Well, previously I'm planning to remove the config once it has been tested. But you are right, it's not a dangerous since if it's failed we can read from local backup file so I just removed the config.

ambry-mysql/src/main/java/com/github/ambry/accountstats/AccountStatsMySqlStore.java
SophieGuo410: Sure, updated.

ambry-mysql/src/integration-test/java/com/github/ambry/accountstats/AccountStatsMySqlStoreIntegrationTest.java
SophieGuo410: Removed the config based on Casey's suggestion.

PR #1930, opened 2021-08-03
Fix filespan for testing and stop auto close log segment when last log segment is empty.

SophieGuo410:
This pr target to support:
1. If the last log segment is empty, no need to auto close it.
2.Update the endOffsetOfPrevMsg to log.getEndOffset() for testing to match the real design.


zzmao (approved):
LGTM

justinlin-linkedin (approved):

PR #1932, opened 2021-08-04
Add a benchmark class for crc32 implementations

justinlin-linkedin:
Adding a benchmark tool to compare the ambry implementation and the java native implementation of crc32.


zzmao (approved):
LGTM

PR #1933, opened 2021-08-04
Adding java native crc32 implementation for PutRequest

justinlin-linkedin:
Adding java native crc32 implementation as a second choice for PutRequest to calculate CRC32
Adding configuration to RouterConfig to choose crc32 implementation.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -104,6 +104,9 @@ public class RouterConfig {
       "router.cross.colo.request.to.dc.with.most.replicas";
   public static final String ROUTER_BACKGROUND_DELETER_MAX_CONCURRENT_OPERATIONS =
       "router.background.deleter.max.concurrent.operations";
+  public static final String ROUTER_PUT_REQUEST_CRC32_ALGORITHM = "router.put.request.crc32.algorithm";
+  public static final String CRC32_ALGORITHM_AMBRY_UTIL = "AmbryUtil";
+  public static final String CRC32_ALGORITHM_JAVA_NATIVE = "JavaNative";
 
   /**
    * Number of independent scaling units for the router.
@@ -500,6 +503,10 @@ public class RouterConfig {
   @Default("0")
   public final int routerBackgroundDeleterMaxConcurrentOperations;
 
+  @Config(ROUTER_PUT_REQUEST_CRC32_ALGORITHM)
+  @Default(CRC32_ALGORITHM_AMBRY_UTIL)
+  public final String routerPutRequestCrc32Algorithm;
+
   /**
    * Create a RouterConfig instance.
    * @param verifiableProperties the properties map to refer to.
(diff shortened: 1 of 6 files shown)


ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequestCrc32Impl.java
```diff
@@ -0,0 +1,77 @@
...
+   */
+  long getValue();
+
+  /**
+   * Return java native implementation of crc32 algorithm.
+   * @return
+   */
+  static PutRequestCrc32Impl getJavaNativeInstance() {
+    return new PutRequestCrc32Impl() {
+      private CRC32 crc32 = new CRC32();
```
cgtz, line 44: nit: final for this and line 64

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -500,6 +503,10 @@
   @Default("0")
   public final int routerBackgroundDeleterMaxConcurrentOperations;
 
+  @Config(ROUTER_PUT_REQUEST_CRC32_IMPLEMENTATION)
+  @Default(CRC32_IMPLEMENTATION_AMBRY_UTIL)
+  public final String routerPutRequestCrc32Implementation;
```
cgtz, line 508: You could make this an enum config with `VerifiableProperties.getEnum`. It will avoid having to do a string comparison in `getCrc32Implementation`.

cgtz (approved):
looks good! It would also be interesting to profile at a future time how much CPU is used for CRC calculation on get requests (`Utils.readNettyByteBufFromCrcInputStream`), which uses a similar large buffer call to updateCrc for the Netty ByteBuf branch.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1397,7 +1399,24 @@ private void fetchRequests(RequestRegistrationCallback<PutOperation> requestRegi
...
-          buf.readableBytes(), BlobType.DataBlob, encryptedPerBlobKey != null ? encryptedPerBlobKey.duplicate() : null);
+          buf.readableBytes(), BlobType.DataBlob, encryptedPerBlobKey != null ? encryptedPerBlobKey.duplicate() : null,
+          getCrc32Implementation());
+    }
+
+    /**
+     * Select crc32 implementation based on {@link RouterConfig#routerPutRequestCrc32Implementation}.
+     * @return
+     */
+    protected PutRequestCrc32Impl getCrc32Implementation() {
```
zzmao, line 1410: Can we make it easy: just test Ture or False?
```
if(routerConfig.useJaveNativeCrc32) {
    return PutRequestCrc32Impl.getJavaNativeInstance();
} else {
    return PutRequestCrc32Impl.getAmbryUtilInstance();
}
```

I don't think we will introduce any other crc32 impl in the future, do we?

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequestCrc32Impl.java
```diff
@@ -0,0 +1,77 @@
...
+        return crc32.getValue();
+      }
+    };
+  }
+
+  /**
+   * Return ambry util implementation of crc32 algorithm.
+   * @return
+   */
+  static PutRequestCrc32Impl getAmbryUtilInstance() {
```
zzmao, line 62: rename it to `getAmbryCrc32Impl()` ?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
justinlin-linkedin: updated

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
justinlin-linkedin: i will use a boolean here.

zzmao (approved):
LGTM

PR #1934, opened 2021-08-05
Fix lag calculation granularity issue for index based token

SophieGuo410:
This pr supports to calculate the accurate value of lag when index based token points to last index segment, so when replication successfully finished, we are able to see the lag goes down to zero.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -1567,7 +1567,7 @@ class PersistentIndex {
       throw new IllegalArgumentException("Offset is invalid: " + offset + "; LogSegment: " + logSegment);
     }
     int numPrecedingLogSegments = getLogSegmentToIndexSegmentMapping(indexSegments).headMap(offset.getName()).size();
-    return numPrecedingLogSegments * log.getSegmentCapacity() + log.getSegmentCapacity() - getTotalOffsetAfterTokenPointedStoreKey(indexBasedToken);
+    return numPrecedingLogSegments * log.getSegmentCapacity() + log.getEndOffset().getOffset() - getTotalOffsetAfterTokenPointedStoreKey(indexBasedToken);
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -1439,7 +1439,19 @@ class PersistentIndex {
       ConcurrentSkipListMap<Offset, IndexSegment> indexSegments) {
     long bytesRead = 0;
     if (token.getType().equals(FindTokenType.IndexBased)) {
-      bytesRead = getAbsolutePositionInLogForOffset(token.getOffset(), indexSegments);
+      try {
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1532,6 +1544,50 @@ private long getAbsolutePositionInLogForOffset(Offset offset,
...
+   */
+  private long getAbsoluteReadBytesFromIndexBasedToken(StoreFindToken indexBasedToken, ConcurrentSkipListMap<Offset, IndexSegment> indexSegments)
+      throws StoreException {
+    Offset offset = indexBasedToken.getOffset();
+    LogSegment logSegment = log.getSegment(offset.getName());
+    if (logSegment == null || offset.getOffset() > logSegment.getEndOffset()) {
+      throw new IllegalArgumentException("Offset is invalid: " + offset + "; LogSegment: " + logSegment);
+    }
+    int numPrecedingLogSegments = getLogSegmentToIndexSegmentMapping(indexSegments).headMap(offset.getName()).size();
+    return numPrecedingLogSegments * log.getSegmentCapacity() + log.getSegmentCapacity() - getTotalOffsetAfterTokenPointedStoreKey(indexBasedToken);
```
justinlin-linkedin, line 1570: i don't think this is valid. we should probably do something like
```
numPrecedingLogSegments * log.getSegmentCapacity() + log.getEndOffset().getOffset() - getTotalOffsetAfterTokenPointedStoreKey(indexBasedToken);
```

SophieGuo410: Thx for catching it! Fixed it.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1532,6 +1544,50 @@ private long getAbsolutePositionInLogForOffset(Offset offset,
...
+    return numPrecedingLogSegments * log.getSegmentCapacity() + log.getEndOffset().getOffset() - getTotalOffsetAfterTokenPointedStoreKey(indexBasedToken);
+  }
+
+  /**
+   * Get the absolute read bytes after index based token {@link StoreKey}.
+   * @param indexBasedToken the point until which the log has been read.
+   * @return the absolute read bytes after index based token {@link StoreKey}.
+   * @throws StoreException
+   */
+  private long getTotalOffsetAfterTokenPointedStoreKey(StoreFindToken indexBasedToken) throws StoreException {
```
cgtz, line 1579: This does seem more of an estimate of position instead of an absolute number since the indexbased tokens are sorted by store key and not log offset.

My understanding is that even if this specific storekey had n bytes after it in the log, it does not mean that all keys prior to that log position were read already.

Is this okay for the use case? i.e. do we just care that the log is either completely read (lag == 0) or not completely read? If so, could we just test whether the store key in the token is equal to the last entry in the index segment or not?

cgtz: I understand this now. The indexing is always based on the index segment, the bytes remaining in the log after a key never comes into play.

This code looks at how many entries in the index segment have yet to be encountered and sums up their size.

PR #1937, opened 2021-08-06
Use AdaptiveOperationTracker in PutOperation

justinlin-linkedin:
We have some issues with SimpleOperationTracker with PutOperation. We want to change PutRequestParallelism to 2 and expect that only 2 copies of the same chunk will be sent out to ambry-server replicas. However, this is not what we see from the metrics. Even with parallelism changed to 2, we are still seeing 3 copies of the same chunk sent out to replicas. The reason of this is because of SimpleOperationTracker.

In SimpleOperationTracker, when we send out first two PutRequests, it will block because of the parallelism is 2. But as long as we receive the response from any of these 2 requests, the inflightCount would be decremented, thus we will immediately send out the third request. For that, we end up with 3 requests.

One way to fix this is to set the PutRequestParallelism to 1, in this case, we will send the first request out, wait and receive the first response. Then send the second request out and receive the second response. Assume that both responses are successes, we would stop and acknowledge this PutOperation as successful. But this way, we would drastically increase the latency of PutOperations, thus damage the overall upload performance.

The other way is to have different logic in the OperationTracker, to send out first 2 requests and wait until both responses come back. But this way, if we have a slow replica, then we are damaging the performance as well.

The best solution is to use AdaptiveOperationTracker, so that when we have slow replica, we would send out the third request right away. But before we can use AdaptiveOperationTracker, we have to change some logic in it.

AdaptiveOperationTracker has the same issue as the SimpleOperationTracker, where when the PutRequestParallelism is 2, it will send out the third request as soon as it receives the first response. This PR would fix that logic.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/AdaptiveOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/AdaptiveOperationTracker.java
@@ -173,6 +173,9 @@ class AdaptiveOperationTracker extends SimpleOperationTracker {
       case GetBlobOperation:
         trackerToReturn = isLocal ? routerMetrics.getBlobLocalDcLatencyMs : routerMetrics.getBlobCrossDcLatencyMs;
         break;
+      case PutOperation:
+        trackerToReturn = routerMetrics.putBlobLatencyMs;
+        break;
       default:
         throw new IllegalArgumentException("Unsupported router operation when getting whole DC latency tracker.");
     }
@@ -192,6 +195,9 @@ class AdaptiveOperationTracker extends SimpleOperationTracker {
         break;
       case GetBlobOperation:
         pastDueCounter = routerMetrics.getBlobPastDueCount;
+        break;
+      case PutOperation:
+        pastDueCounter = routerMetrics.putBlobPastDueCount;
         break;
       default:
         throw new IllegalArgumentException("Unsupported router operation when getting whole DC past due counter.");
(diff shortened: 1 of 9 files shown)


zzmao (approved):
LGTM

ambry-router/src/main/java/com/github/ambry/router/AdaptiveOperationTracker.java
```diff
@@ -244,12 +273,14 @@ public boolean hasNext() {
...
     public void remove() {
       replicaIterator.remove();
-      if (inflightCount >= getCurrentParallelism() && unexpiredRequestSendTimes.size() > 0) {
+      if (!shouldSendRequestWithoutConsideringMetrics() && unexpiredRequestSendTimes.size() > 0) {
         // we are here because oldest request is past due
         Map.Entry<ReplicaId, Pair<Boolean, Long>> oldestEntry = unexpiredRequestSendTimes.entrySet().iterator().next();
-        expiredRequestSendTimes.put(oldestEntry.getKey(), oldestEntry.getValue().getSecond());
-        unexpiredRequestSendTimes.remove(oldestEntry.getKey());
-        pastDueCounter.inc();
+        if (oldestEntry.getValue().getFirst()) {
```
cgtz, line 279: Could you explain why this condition was added?

cgtz (approved):

ambry-router/src/main/java/com/github/ambry/router/AdaptiveOperationTracker.java
justinlin-linkedin: double check. If we are in this if statement (link 276) then it means that the oldest entry should already past due and `oldestEntry.getValue().getFirst()` should return true. Putting this if statement here just for sanity check.

zzmao (approved):

PR #1939, opened 2021-08-09
Ignore client terminations when using Netty native IO

cgtz:
Netty prefixes exceptions thrown by their native IO (epoll) layer with a
method name. This means that our old code for detecting when a client
has terminated a connection no longer works when using the native IO
libraries. This change will do a suffix detection instead which will
catch both forms of the exception message.


zzmao (approved):
LGTM

PR #1940, opened 2021-08-10
Add some log messages to root cause partition map mismatch in recovery cluster

ankagrawal:
We have observed that during replication, the recovery test cluster map's static map is not able to find the partition being replicated from the replication data stream obtained from remote. Adding some logs here to root cause the issue.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/PartitionLayout.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/PartitionLayout.java
@@ -269,10 +269,15 @@ class PartitionLayout {
    */
   public Partition getPartition(InputStream stream) throws IOException {
     byte[] partitionBytes = Partition.readPartitionBytesFromStream(stream);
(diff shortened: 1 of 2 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/PartitionLayout.java
```diff
@@ -268,6 +269,9 @@ public void addNewReplicas(Partition partition, List<Disk> disks) {
    */
   public Partition getPartition(InputStream stream) throws IOException {
     byte[] partitionBytes = Partition.readPartitionBytesFromStream(stream);
+    logger.info("Partition bytes from stream: " + Arrays.toString(partitionBytes));
```
cgtz, line 272: This logging probably won't help exactly since the bytes read from stream will include the version number and the partition number concatenated

justinlin-linkedin (approved):

cgtz (approved):

PR #1944, opened 2021-08-13
[VCR_AUTOMATION] VCR Updater Implementation

zzmao:
Move core logic of HelixVcrPoplulateTool to HelixVcrUtil.java.
Implement VCR Helix Update logic on Ambry cluster change: The VCR who is ONLINE and backing up partition 1 will be the updater. It call vcr helix update in 2 cases: 1. ambry cluster change arrives 2. onOfflineBecomeOnline.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -382,7 +382,7 @@ public class CloudConfig {
    * The delay between an ambry cluster change notification arrive and VCR helix update action
    */
   @Config(VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS)
-  @Default("180")
+  @Default("60")
   public final int vcrHelixUpdateDelayTimeInSeconds;
 
   /**
@@ -453,7 +453,7 @@ public class CloudConfig {
     vcrHelixUpdaterParitionId = verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, "1");
     vcrHelixUpdateConfig =
         verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, DEFAULT_VCR_HELIX_UPDATE_CONFIG);
-    vcrHelixUpdateDelayTimeInSeconds = verifiableProperties.getInt(VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS, 180);
+    vcrHelixUpdateDelayTimeInSeconds = verifiableProperties.getInt(VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS, 60);
     vcrHelixUpdateDryRun = verifiableProperties.getBoolean(VCR_HELIX_UPDATE_DRY_RUN, false);
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -365,11 +365,11 @@ public class CloudConfig {
   public final String vcrHelixStateModelFactoryClass;
 
   /**
-   * The vcr node responsible for partition 1 is the vcr helix updater.
+   * The vcr node responsible for partition 1 is the vcr helix updater. Use empty string to disable vcr helix auto update.
    */
   @Config(VCR_HELIX_UPDATER_PARTITION_ID)
-  @Default("-1")
-  public final String vcrHelixUpdaterParitionId;
+  @Default("")
+  public final String vcrHelixUpdaterPartitionId;
 
   /**
    * The config string used when update VCR helix.
@@ -450,7 +450,7 @@ public class CloudConfig {
         Utils.splitString(verifiableProperties.getString(VCR_SOURCE_DATACENTERS, ""), ",", HashSet::new);
     vcrHelixStateModelFactoryClass = verifiableProperties.getString(VCR_HELIX_STATE_MODEL_FACTORY_CLASS,
         DEFAULT_VCR_HELIX_STATE_MODEL_FACTORY_CLASS);
-    vcrHelixUpdaterParitionId = verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, "-1");
+    vcrHelixUpdaterPartitionId = verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, "");
     vcrHelixUpdateConfig =
         verifiableProperties.getString(VCR_HELIX_UPDATE_CONFIG, DEFAULT_VCR_HELIX_UPDATE_CONFIG);
     vcrHelixUpdateDelayTimeInSeconds = verifiableProperties.getInt(VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS, 60);
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -63,6 +63,10 @@ public class CloudConfig {
   public static final String VCR_SOURCE_DATACENTERS = "vcr.source.datacenters";
   public static final String CLOUD_COMPACTION_NUM_THREADS = "cloud.compaction.num.threads";
   public static final String VCR_HELIX_STATE_MODEL_FACTORY_CLASS = "vcr.helix.state.model.factory.class";
+  public static final String VCR_HELIX_UPDATER_PARTITION_ID = "vcr.helix.updater.partition.id";
+  public static final String VCR_HELIX_UPDATE_CONFIG = "vcr.helix.update.config";
+  public static final String VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS = "vcr.helix.update.delay.time.in.seconds";
+  public static final String VCR_HELIX_UPDATE_DRY_RUN = "vcr.helix.update.dry.run";
 
   public static final String DEFAULT_CLOUD_DESTINATION_FACTORY_CLASS =
       "com.github.ambry.cloud.azure.AzureCloudDestinationFactory";
@@ -93,6 +97,15 @@ public class CloudConfig {
       "com.github.ambry.cloud.HelixVcrClusterAgentsFactory";
   public static final String DEFAULT_VCR_HELIX_STATE_MODEL_FACTORY_CLASS =
       "com.github.ambry.cloud.OnlineOfflineHelixVcrStateModelFactory";
+  public static final String DEFAULT_VCR_HELIX_UPDATE_CONFIG =
+      "{\n" + "  \"clusterConfigFields\": {\n" + "    \"maxOfflineInstancesAllowed\": 4,\n"
+          + "    \"numOfflineInstancesForAutoExit\": 2,\n" + "    \"allowAutoJoin\": true\n" + "  },\n"
+          + "  \"idealStateConfigFields\": {\n" + "    \"numReplicas\": 1,\n"
+          + "    \"stateModelDefRef\": \"OnlineOffline\",\n"
+          + "    \"rebalanceStrategy\": \"org.apache.helix.controller.rebalancer.strategy.CrushEdRebalanceStrategy\",\n"
+          + "    \"minActiveReplicas\": 0,\n"
+          + "    \"rebalancerClassName\": \"org.apache.helix.controller.rebalancer.DelayedAutoRebalancer\",\n"
+          + "    \"rebalanceDelayInMins\": 20\n" + "  }\n" + "}";
 
   /**
    * True for VCR node, false for live serving node.
@@ -351,6 +364,34 @@ public class CloudConfig {
   @Default(DEFAULT_VCR_HELIX_STATE_MODEL_FACTORY_CLASS)
   public final String vcrHelixStateModelFactoryClass;
 
+  /**
+   * The vcr node responsible for partition 1 is the vcr helix updater.
+   */
+  @Config(VCR_HELIX_UPDATER_PARTITION_ID)
+  @Default("1")
+  public final String vcrHelixUpdaterParitionId;
+
+  /**
+   * The config string used when update VCR helix.
+   */
+  @Config(VCR_HELIX_UPDATE_CONFIG)
+  @Default(DEFAULT_VCR_HELIX_UPDATE_CONFIG)
+  public final String vcrHelixUpdateConfig;
+
+  /**
+   * The delay between an ambry cluster change notification arrive and VCR helix update action
+   */
+  @Config(VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS)
+  @Default("180")
+  public final int vcrHelixUpdateDelayTimeInSeconds;
+
+  /**
+   * If true, VCR helix update will be in dry run mode. We can also use this to turn off automated VCR cluster update.
+   */
+  @Config(VCR_HELIX_UPDATE_DRY_RUN)
+  @Default("true")
+  public final boolean vcrHelixUpdateDryRun;
+
   public CloudConfig(VerifiableProperties verifiableProperties) {
 
     cloudIsVcr = verifiableProperties.getBoolean(CLOUD_IS_VCR, false);
@@ -409,5 +450,10 @@ public class CloudConfig {
         Utils.splitString(verifiableProperties.getString(VCR_SOURCE_DATACENTERS, ""), ",", HashSet::new);
     vcrHelixStateModelFactoryClass = verifiableProperties.getString(VCR_HELIX_STATE_MODEL_FACTORY_CLASS,
         DEFAULT_VCR_HELIX_STATE_MODEL_FACTORY_CLASS);
+    vcrHelixUpdaterParitionId = verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, "1");
+    vcrHelixUpdateConfig =
+        verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, DEFAULT_VCR_HELIX_UPDATE_CONFIG);
+    vcrHelixUpdateDelayTimeInSeconds = verifiableProperties.getInt(VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS, 180);
+    vcrHelixUpdateDryRun = verifiableProperties.getBoolean(VCR_HELIX_UPDATE_DRY_RUN, false);
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -368,7 +368,7 @@ public class CloudConfig {
    * The vcr node responsible for partition 1 is the vcr helix updater.
    */
   @Config(VCR_HELIX_UPDATER_PARTITION_ID)
-  @Default("1")
+  @Default("-1")
   public final String vcrHelixUpdaterParitionId;
 
   /**
@@ -450,9 +450,9 @@ public class CloudConfig {
         Utils.splitString(verifiableProperties.getString(VCR_SOURCE_DATACENTERS, ""), ",", HashSet::new);
     vcrHelixStateModelFactoryClass = verifiableProperties.getString(VCR_HELIX_STATE_MODEL_FACTORY_CLASS,
         DEFAULT_VCR_HELIX_STATE_MODEL_FACTORY_CLASS);
-    vcrHelixUpdaterParitionId = verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, "1");
+    vcrHelixUpdaterParitionId = verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, "-1");
     vcrHelixUpdateConfig =
-        verifiableProperties.getString(VCR_HELIX_UPDATER_PARTITION_ID, DEFAULT_VCR_HELIX_UPDATE_CONFIG);
+        verifiableProperties.getString(VCR_HELIX_UPDATE_CONFIG, DEFAULT_VCR_HELIX_UPDATE_CONFIG);
     vcrHelixUpdateDelayTimeInSeconds = verifiableProperties.getInt(VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS, 60);
     vcrHelixUpdateDryRun = verifiableProperties.getBoolean(VCR_HELIX_UPDATE_DRY_RUN, false);
   }
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -389,7 +389,7 @@ public class CloudConfig {
    * If true, VCR helix update will be in dry run mode. We can also use this to turn off automated VCR cluster update.
    */
   @Config(VCR_HELIX_UPDATE_DRY_RUN)
-  @Default("true")
+  @Default("false")
   public final boolean vcrHelixUpdateDryRun;
 
   public CloudConfig(VerifiableProperties verifiableProperties) {
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
@@ -14,6 +14,7 @@
 package com.github.ambry.cloud;
 
 import com.github.ambry.account.AccountService;
+import com.github.ambry.clustermap.AmbryPartition;
 import com.github.ambry.clustermap.CloudDataNode;
 import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.ClusterMapUtils;
@@ -21,10 +22,17 @@ import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.VcrClusterParticipant;
 import com.github.ambry.clustermap.VcrClusterParticipantListener;
+import com.github.ambry.commons.Callback;
+import com.github.ambry.commons.RetryExecutor;
+import com.github.ambry.commons.RetryPolicies;
+import com.github.ambry.commons.RetryPolicy;
 import com.github.ambry.config.CloudConfig;
 import com.github.ambry.config.ClusterMapConfig;
 import com.github.ambry.config.StoreConfig;
+import com.github.ambry.utils.ByteBufferInputStream;
 import com.github.ambry.utils.Utils;
+import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -33,6 +41,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.helix.HelixAdmin;
@@ -57,7 +67,6 @@ public class HelixVcrClusterParticipant implements VcrClusterParticipant {
   private final DataNodeId currentDataNode;
   private final String vcrClusterName;
   private final String vcrInstanceName;
-  private final Map<String, PartitionId> partitionIdMap;
   private final Set<PartitionId> assignedPartitionIds = ConcurrentHashMap.newKeySet();
   private final HelixVcrClusterMetrics metrics;
   private final List<VcrClusterParticipantListener> listeners = new ArrayList<>();
@@ -68,6 +77,9 @@ public class HelixVcrClusterParticipant implements VcrClusterParticipant {
   private HelixManager manager;
   private HelixAdmin helixAdmin;
   private VcrMetrics vcrMetrics;
+  private final RetryPolicy retryPolicy = RetryPolicies.exponentialPolicy(3, 2000);
+  private final RetryExecutor retryExecutor = new RetryExecutor(Executors.newScheduledThreadPool(2));
+  private final ClusterMap clusterMap;
 
   /**
    * Construct the helix VCR cluster.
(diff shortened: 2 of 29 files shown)


zzmao:
Let me know if you need a code walk-through.

cgtz (commented):
just a few starter comments to publish

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -105,21 +117,44 @@ public HelixVcrClusterParticipant(CloudConfig cloudConfig, ClusterMapConfig clus
...
+          } else {
+            logger.info("Partition {} is added to current VCR: {}. Number of assigned partitions: {}", partitionIdStr,
+                vcrInstanceName, assignedPartitionIds.size());
+          }
+        });
+  }
+
+  private void doAddPartition(String partitionIdStr, Callback<Object> callback) {
+    byte[] partitionBytes =
+        ClusterMapUtils.serializeShortAndLong(AmbryPartition.CURRENT_VERSION, Long.parseLong(partitionIdStr));
```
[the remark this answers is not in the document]
zzmao: sounds good. I also have this thought but forget to do this before publishing this PR.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -105,21 +117,44 @@ public HelixVcrClusterParticipant(CloudConfig cloudConfig, ClusterMapConfig clus
    * @param partitionIdStr The partitionIdStr notified by Helix.
    */
   public void addPartition(String partitionIdStr) {
-    PartitionId partitionId = partitionIdMap.get(partitionIdStr);
+    // Add partition with retry logic in case Ambry clustermap is not update to date.
+    retryExecutor.runWithRetries(retryPolicy, callback -> doAddPartition(partitionIdStr, callback), e -> true,
```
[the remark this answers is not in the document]
zzmao: I don't have to intend to do blocking or async execution. Both work to my understanding on Helix/

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -105,21 +117,44 @@ public HelixVcrClusterParticipant(CloudConfig cloudConfig, ClusterMapConfig clus
...
-        logger.info("Partition {} is added to current VCR: {}. Number of assigned partitions: {}", partitionIdStr,
-            vcrInstanceName, assignedPartitionIds.size());
         logger.debug("Assigned Partitions: {}", assignedPartitionIds);
       } else {
         logger.info("Partition {} exists on current VCR: {}", partitionIdStr, vcrInstanceName);
       }
+      callback.onCompletion(null, null);
     } else {
-      logger.error("Partition {} not in clusterMap on add.", partitionIdStr);
       metrics.partitionIdNotInClusterMapOnAdd.inc();
```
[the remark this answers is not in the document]
zzmao: good catch

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -292,4 +338,66 @@ protected String getReplicaThreadName(String datacenterToReplicateFrom, int thre
...
+   * 2. On Ambry cluster change
+   */
+  private void scheduleVcrHelix() {
+    if (vcrHelixUpdateFuture != null && vcrHelixUpdateFuture.cancel(false)) {
+      // If a vcrHelixUpdate task is scheduled, try to cancel it first.
+      logger.info("There was a scheduled vcrHelixUpdate task. Canceled.");
+    }
+    // either success cancel or not, we should schedule a new job to updateVcrHelix
+    vcrHelixUpdateFuture =
+        scheduler.schedule(() -> updateVcrHelix(), cloudConfig.vcrHelixUpdateDelayTimeInSeconds, TimeUnit.SECONDS);
```
ankagrawal, line 354: Why do we delay the `updateVcrHelix` execution? Is it to make sure that any previously scheduled update is cancelled before starting the task? If yes, should we rather try to wait for that update to be cancelled?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -91,6 +110,22 @@ public void start() throws ReplicationException {
     vcrClusterParticipant.addListener(new VcrClusterParticipantListener() {
       @Override
       public void onPartitionAdded(PartitionId partitionId) {
+        if (partitionId.isEqual(cloudConfig.vcrHelixUpdaterParitionId)) {
+          vcrHelixUpdateLock.lock();
+          try {
+            if (isAmbryListenerToUpdateVcrHelixRegistered == false) {
+              // Only register the listener once. Unfortunately, we can't unregister a listener, but it's OK since
```
ankagrawal, line 117: I guess it is possible for `isVcrHelixUpdater` to be true in two nodes. For example, lets say at time "t" node A is the owner of partition 1. Due to some partition reassignment (lets say new node addition), partition 1 get assigned to node B at time "t+1". In this case its possible for Node B to get notification of `addPartition` of partition 1 at "t+2" before Node A gets `removePartition` notification (at t+4). If a  main cluster update happens between "t+2" and "t+4", then two nodes might try to update vcr cluster at the same time.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -292,4 +338,66 @@ protected String getReplicaThreadName(String datacenterToReplicateFrom, int thre
...
+    logger.info("VcrHelixUpdate task scheduled. Will run in {} seconds.", cloudConfig.vcrHelixUpdateDelayTimeInSeconds);
+  }
+
+  /**
+   * The actual performer to update VCR Helix:
+   */
+  private void updateVcrHelix() {
+    String localDcZkStr = ((HelixClusterManager) clusterMap).getLocalDcZkConnectString();
+    logger.info("Going to update VCR Helix Cluster. Dryrun: {}", cloudConfig.vcrHelixUpdateDryRun);
+    logger.info("Current partitions in clustermap: ", clusterMap.getAllPartitionIds(null));
```
ankagrawal, line 364: Theoretically its possible for cluster map to get updated between line 363 and 364. It might be better to `getAll{artitoinIds` to a local list, and use the same list to log as well as for updating.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -292,4 +338,66 @@ protected String getReplicaThreadName(String datacenterToReplicateFrom, int thre
...
+    for (PartitionId partitionIdx : clusterMap.getAllPartitionIds(null)) {
+      logger.info(partitionIdx.toString());
+    }
+    try {
+      isVcrHelixUpdateInProgress = true;
+      HelixVcrUtil.updateResourceAndPartition(localDcZkStr, clusterMapConfig.clusterMapClusterName,
```
ankagrawal, line 370: Is it possible that this call fails (maybe temporary network issue)? In this case do we wait till the next time cluster map update happens on the main cluster?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
ankagrawal: I guess its not too much of a problem if in a rare case two nodes update the helix cluster concurrently. Helix should be able to handle this gracefully. But we should remove the comment from here.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -105,21 +117,44 @@ public HelixVcrClusterParticipant(CloudConfig cloudConfig, ClusterMapConfig clus
    * @param partitionIdStr The partitionIdStr notified by Helix.
    */
   public void addPartition(String partitionIdStr) {
-    PartitionId partitionId = partitionIdMap.get(partitionIdStr);
+    // Add partition with retry logic in case Ambry clustermap is not update to date.
+    retryExecutor.runWithRetries(retryPolicy, callback -> doAddPartition(partitionIdStr, callback), e -> true,
+        (result, exception) -> {
+          if (exception != null) {
+            logger.warn("Partition {} is not in clusterMap on addPartition and retry failed.", partitionIdStr,
```
ankagrawal, line 124: I guess we should also log a error message and a metric when we exhaust all the retries. Because we have no other way of knowing if a partition update totally failed.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -105,21 +117,44 @@ public HelixVcrClusterParticipant(CloudConfig cloudConfig, ClusterMapConfig clus
...
+    PartitionId partitionId = null;
+    try {
+      partitionId = clusterMap.getPartitionIdFromStream(new ByteBufferInputStream(ByteBuffer.wrap(partitionBytes)));
+    } catch (IOException e) {
+      callback.onCompletion(null, e);
+      return;
+    }
     if (partitionId != null) {
       if (assignedPartitionIds.add(partitionId)) {
         for (VcrClusterParticipantListener listener : listeners) {
```
ankagrawal, line 145: We are adding the new partition id to the `assignedPartitionIds `, and only when not already present, we call `onPartitionAdded `. Since we are retrying, this can cause issues. In the first try the `assignedPartitionIds.add` will succeed and `onPartitionAdded ` might fail. In the subsequent retries `assignedPartitionIds.add` will already have the partition id, and it will return false. This will basically make the retries no-op.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
ankagrawal: Also are we assuming that the only exception possible is partition not in clustemap?

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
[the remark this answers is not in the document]
cgtz: So it is okay for the doAddPartition work to go on in the background after addPartition returns? I guess we are also swallowing/logging any exception that happens after multiple retries in doAddPartition.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrMetrics.java
```diff
@@ -98,4 +104,13 @@ public VcrMetrics(MetricRegistry registry) {
   public MetricRegistry getMetricRegistry() {
     return registry;
   }
+
+  public void registerVcrHelixUpdateGuage(Gauge<Integer> updaterGaugeCount,
```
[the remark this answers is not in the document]
zzmao: fixed.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: If we add multiple partitions, VCR may receive multiple updates, but we only need to update it once. 
So we delay the update for a short time window, if there is another update comes in the short time window, we just cancel the old pending update.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: The rare case here is dangerous, let me double checking with Helix team. I will add the case you described to the comment.

zzmao: Worst case, we need a distributed lock.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: Added some log in `updateResourceAndPartition` to give us more precise information.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
zzmao: good catch. I am trying to get rid of `assignedPartitionIds `. Will update the logic here.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
zzmao: The actual vcr cluster update is a scheduled task(`updateVcrHelix`), so we can't know its result here.  The logic here is only to trigger the update.

Added an exception catch in `updateVcrHelix`.

All exception should be retried. Updated the log here.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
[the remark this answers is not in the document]
zzmao: So it is okay for the doAddPartition work to go on in the background after addPartition returns? : YES based on my current logic. Let me double check with Helix team.

 I guess we are also swallowing/logging any exception that happens after multiple retries in doAddPartition.
Didn't get this. If all retries failed, we could see the exception log in the user call back, no?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -292,4 +348,66 @@ protected String getReplicaThreadName(String datacenterToReplicateFrom, int thre
...
+    vcrHelixUpdateFuture =
+        scheduler.schedule(() -> updateVcrHelix(), cloudConfig.vcrHelixUpdateDelayTimeInSeconds, TimeUnit.SECONDS);
+    logger.info("VcrHelixUpdate task scheduled. Will run in {} seconds.", cloudConfig.vcrHelixUpdateDelayTimeInSeconds);
+  }
+
+  /**
+   * The actual performer to update VCR Helix:
+   */
+  private void updateVcrHelix() {
+    String localDcZkStr = ((HelixClusterManager) clusterMap).getLocalDcZkConnectString();
```
ankagrawal, line 372: I think its best to make the method `updateVcrHelix` synchronized. There are race conditions that might cause two threads to run this method simultaneously. For example, `vcrHelixUpdateFuture.cancel` method in `scheduleVcrHelix` will not cancel an already running thread. In that case two threads might try to `updateVcrHelix`.
Note that this method is not a performance critical method anyway, and we are already waiting for sometime before scheduling it to avoid a burst of calls. So its ok to add a lock here if it makes the code simpler and safe.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -105,21 +114,37 @@ public HelixVcrClusterParticipant(CloudConfig cloudConfig, ClusterMapConfig clus
    * @param partitionIdStr The partitionIdStr notified by Helix.
    */
   public void addPartition(String partitionIdStr) {
-    PartitionId partitionId = partitionIdMap.get(partitionIdStr);
+    // Add partition with retry logic in case Ambry clustermap is not update to date.
+    retryExecutor.runWithRetries(retryPolicy, callback -> doAddPartition(partitionIdStr, callback), e -> true,
+        (result, exception) -> {
+          if (exception != null) {
```
ankagrawal, line 120: If we try to handle only one type of exception in retries -- then it should say so.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
ankagrawal: The delay time should be greater than the time to upgrade an Ambry cluster. With that we will have a better chance that we have taken in all the cluster updates, before we go and update the vcr helix cluster.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -292,4 +348,67 @@ protected String getReplicaThreadName(String datacenterToReplicateFrom, int thre
...
+  synchronized
+  private void updateVcrHelix() {
+    logger.info("Going to update VCR Helix Cluster. Dryrun: {}", cloudConfig.vcrHelixUpdateDryRun);
+    logger.info("Current partitions in clustermap data structure: {}",
+        clusterMap.getAllPartitionIds(null).stream().map(Object::toString).collect(Collectors.joining(",")));
+    try {
+      String localDcZkStr = ((HelixClusterManager) clusterMap).getLocalDcZkConnectString();
+      isVcrHelixUpdateInProgress = true;
+      HelixVcrUtil.updateResourceAndPartition(localDcZkStr, clusterMapConfig.clusterMapClusterName,
+          cloudConfig.vcrClusterZkConnectString, cloudConfig.vcrClusterName, vcrHelixConfig,
```
ankagrawal, line 380: I would suggest that we should have a verification step, after calls to `HelixVcrUtil.updateResourceAndPartition`, in the finally block (so that it runs for the exception case as well). It should check 
1. Is the main cluster and helix cluster are up-to-date with each other.
2. If not, is a updateVcrHelix task in progress or already scheduled. If yes, then ignore. If not, then log an error message saying that "main and helix cluster have a mismatch" and generate a metric. Create an alert on that metric, because it will tell us that we need to upgrade manually.

This should be tied to my earlier comment to configure the task delay time to be greater than Ambry main cluster upgrade time, so that we have better surety that when this task runs, all the Ambry cluster updates are done.

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -351,6 +364,34 @@
   @Default(DEFAULT_VCR_HELIX_STATE_MODEL_FACTORY_CLASS)
   public final String vcrHelixStateModelFactoryClass;
 
+  /**
+   * The vcr node responsible for partition 1 is the vcr helix updater.
+   */
+  @Config(VCR_HELIX_UPDATER_PARTITION_ID)
+  @Default("-1")
```
cgtz, line 371: Since this is a string, could the default be empty string "" to disable the feature?

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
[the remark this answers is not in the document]
cgtz: Do we have a metric that we can monitor to tell if all partition additions succeeded?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -91,6 +112,30 @@ public void start() throws ReplicationException {
...
+              // Only register the listener once. Unfortunately, we can't unregister a listener, so we use
+              // isAmbryListenerToUpdateVcrHelixRegistered as the flag.
+              clusterMap.registerClusterMapListener(
+                  new com.github.ambry.cloud.VcrReplicationManager.AmbryListenerToUpdateVcrHelix());
+              isAmbryListenerToUpdateVcrHelixRegistered = true;
+              logger.info("VCR updater registered.");
+            }
+            // It's possible isVcrHelixUpdater to be true on two nodes.
+            // For example, lets say at time "t" node A is the owner of partition 1. Due to some partition reassignment
+            // (lets say new node addition), partition 1 get assigned to node B at time "t+1". In this case it's possible
```
cgtz, line 128: Let's say that we are trying to bootstrap a completely new vcr cluster after this change. Would we use the tool to create exactly one partition in that case and let this logic create all the other partitions?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -91,6 +112,30 @@ public void start() throws ReplicationException {
...
       @Override
       public void onPartitionAdded(PartitionId partitionId) {
+        if (partitionId.isEqual(cloudConfig.vcrHelixUpdaterParitionId)) {
+          vcrHelixUpdateLock.lock();
+          try {
+            if (!isAmbryListenerToUpdateVcrHelixRegistered) {
+              // Only register the listener once. Unfortunately, we can't unregister a listener, so we use
+              // isAmbryListenerToUpdateVcrHelixRegistered as the flag.
+              clusterMap.registerClusterMapListener(
+                  new com.github.ambry.cloud.VcrReplicationManager.AmbryListenerToUpdateVcrHelix());
```
cgtz, line 122: nit: is the fully specified classname needed here?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
```diff
@@ -42,7 +42,7 @@
   private long lastUpdatedSealedStateChangeCounter = 0;
 
   private static final short VERSION_FIELD_SIZE_IN_BYTES = 2;
-  private static final short CURRENT_VERSION = 1;
+  public static final short CURRENT_VERSION = 1;
```
cgtz, line 45: I think this could be package private now

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
```diff
@@ -447,7 +447,7 @@ public static void validatePorts(Port plainTextPort, Port sslPort, Port http2Por
    * @param longValue a Long value.
    * @return the serialized byte array.
    */
-  static byte[] serializeShortAndLong(Short shortValue, Long longValue) {
+  public static byte[] serializeShortAndLong(Short shortValue, Long longValue) {
```
cgtz, line 450: Could this still be package private?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -374,6 +375,17 @@ public PartitionId getPartitionIdFromStream(InputStream stream) throws IOExcepti
     return partition;
   }
 
+  @Override
+  public PartitionId getPartitionIdByName(String partitionIdStr) {
+    byte[] partitionBytes =
```
cgtz, line 380: could we just lookup `partitionNameToAmbryPartition.get(partitionIdStr)` here, or is there some inconsistency between the 2 data structures that I'm forgetting about?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: A longer delay window is preferred, but I don't understand why it should > time to upgrade vcr helix cluster.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: In this case, it highly possible that some fault error happened. My plan is to fire an alert and developers to look at it.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: `synchronized` added.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -105,21 +114,37 @@ public HelixVcrClusterParticipant(CloudConfig cloudConfig, ClusterMapConfig clus
...
+          } else {
+            logger.info("Partition {} is added to current VCR: {}. Number of assigned partitions: {}", partitionIdStr,
+                vcrInstanceName, assignedPartitionIds.size());
+          }
+        });
+  }
+
+  private void doAddPartition(String partitionIdStr, Callback<Object> callback) {
+    PartitionId partitionId = clusterMap.getPartitionIdByName(partitionIdStr);
     if (partitionId != null) {
```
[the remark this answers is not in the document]
zzmao: fixed.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -130,21 +155,17 @@ public void addPartition(String partitionIdStr) {
    * @param partitionIdStr The partitionIdStr notified by Helix.
    */
   public void removePartition(String partitionIdStr) {
-    PartitionId partitionId = partitionIdMap.get(partitionIdStr);
+    PartitionId partitionId = partitionIdMap.remove(partitionIdStr);
     if (partitionId != null) {
```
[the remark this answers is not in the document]
zzmao: fixed.

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
```diff
@@ -0,0 +1,373 @@
...
+    ConfigAccessor configAccessor = new ConfigAccessor(destZkClient);
+    ClusterConfig clusterConfig = configAccessor.getClusterConfig(destClusterName);
+    clusterConfig.setPersistBestPossibleAssignment(true);
+    ClusterConfigFields clusterConfigFields = config.getClusterConfigFields();
+    // if offline instances >= max offline instances allowed config, helix enters maintenance mode.
+    clusterConfig.setMaxOfflineInstancesAllowed(clusterConfigFields.getMaxOfflineInstancesAllowed());
+    // if offline instances <= number of offline instances config, helix exit maintenance mode.
+    clusterConfig.setNumOfflineInstancesForAutoExit(clusterConfigFields.getNumOfflineInstancesForAutoExit());
+    if (dryRun) {
+      logger.info("Will update cluster config to: {}", clusterConfig.toString());
```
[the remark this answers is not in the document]
zzmao: good catch. Fixed.

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -351,6 +364,34 @@
...
+  public final String vcrHelixUpdaterParitionId;
```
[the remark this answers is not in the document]
zzmao: fixed.

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
zzmao: yes. Empty string looks good.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: To bootstrap a VCR helix, my plan is to use previous CLI. Because setup up a helix involves too many configs. 

The scope of vcr automation is only to update VCR helix.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: fixed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
zzmao: fixed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
zzmao: fixed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
zzmao: changed to `partitionNameToAmbryPartition`.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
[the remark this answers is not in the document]
zzmao: It's hard to know `all partitions` because helix assign partition in a streaming style. 

We could add metric to indicate partition add failure count.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: Going to use helix distributed lock to solve this problem.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
zzmao: ?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: let's discuss this.

ankagrawal (approved):
Looks good to me overall. We should add a verification step at the end of `updateVcrHelix()` method in a follow up PR.

zzmao:
@ankagrawal @cgtz  Are we good to merge this PR?

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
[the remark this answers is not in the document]
cgtz: Oh, here I guess I meant that if there was an existing metric that tells the total number of partitions in the VCR cluster, we can know if the VCR cluster has converged with the server cluster. I guess we can also get this information from the helix admin UI.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
cgtz: We can discuss this offline. I was asking this with regards to automating the buildout of new datacenters. Helix is building a terraform provider for setting up new clusters programatically and I was trying to find out how far we are from being able to use that vs. making a new terraform step that would use the CLI tool we wrote.

PR #1953, opened 2021-08-24
[VCR_AUTOMATION] Add Helix lock and VCR automation integration test

zzmao:
Add VCR automation integration test
Add Helix Lock to make sure only one instance update vcr Helix.
Add a fixed rate task to check if ambry helix and vcr helix on sync.
Add resource drop handing logic in HelixVcrUtil
Fix a metric naming typo

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -66,6 +66,10 @@ public class CloudConfig {
   public static final String VCR_HELIX_UPDATER_PARTITION_ID = "vcr.helix.updater.partition.id";
   public static final String VCR_HELIX_UPDATE_CONFIG = "vcr.helix.update.config";
   public static final String VCR_HELIX_UPDATE_DELAY_TIME_IN_SECONDS = "vcr.helix.update.delay.time.in.seconds";
+  public static final String VCR_HELIX_SYNC_CHECK_INTERVAL_IN_SECONDS = "vcr.helix.sync.check.interval.in.seconds";
+  public static final String VCR_HELIX_LOCK_TIMEOUT_IN_MS = "vcr.helix.lock.timeout.in.ms";
+  public static final String VCR_WAIT_TIME_IF_HELIX_LOCK_NOT_OBTAINED_IN_MS =
+      "vcr.wait.time.if.helix.lock.not.obtained.in.ms";
   public static final String VCR_HELIX_UPDATE_DRY_RUN = "vcr.helix.update.dry.run";
 
   public static final String DEFAULT_CLOUD_DESTINATION_FACTORY_CLASS =
@@ -386,6 +390,27 @@ public class CloudConfig {
   public final int vcrHelixUpdateDelayTimeInSeconds;
 
   /**
+   * The period to check if Ambry Helix and VCR Helix on sync.
+   */
+  @Config(VCR_HELIX_SYNC_CHECK_INTERVAL_IN_SECONDS)
+  @Default("3600")
+  public final int vcrHelixSyncCheckIntervalInSeconds;
+
+  /**
+   * Helix lock timeout in ms. Use 10 minutes by default.
+   */
+  @Config(VCR_HELIX_LOCK_TIMEOUT_IN_MS)
+  @Default("600000")
+  public final long vcrHelixLockTimeoutInMs;
+
+  /**
+   * Sleep time if failed to obtain helix lock to update vcr helix.1
+   */
+  @Config(VCR_WAIT_TIME_IF_HELIX_LOCK_NOT_OBTAINED_IN_MS)
+  @Default("500")
+  public final long vcrWaitTimeIfHelixLockNotObtainedInMs;
+
+  /**
    * If true, VCR helix update will be in dry run mode. We can also use this to turn off automated VCR cluster update.
    */
   @Config(VCR_HELIX_UPDATE_DRY_RUN)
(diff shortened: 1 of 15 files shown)


zzmao:
This PR is based on https://github.com/linkedin/ambry/pull/1944 . Please only look at the last commit.

ambry-server/src/integration-test/java/com/github/ambry/server/VcrAutomationTest.java
```diff
@@ -0,0 +1,293 @@
...
+    helixControllerManager.syncStop();
+    vcrHelixControllerManager.syncStop();
+    zkInfoList.get(0).shutdown();
+  }
+
+  class NoOpCloudFactory implements CloudDestinationFactory {
+
+    @Override
+    public CloudDestination getCloudDestination() throws IllegalStateException {
+      return new CloudDestination() {
```
cgtz, line 213: You can likely use mockito here since it doesn't seem to be a complicated mock.

zzmao: fixed.

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -385,6 +389,27 @@
...
+
+  /**
+   * Helix lock timeout in ms. Use 10 minutes by default.
+   */
+  @Config(VCR_HELIX_LOCK_TIMEOUT_IN_MS)
+  @Default("600000")
+  public final long vcrHelixLockTimeoutInMs;
+
+  /**
+   * Sleep time if failed to obtain helix lock to update vcr helix.1
```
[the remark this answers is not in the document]
zzmao: fixed.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -115,21 +122,30 @@ public void onPartitionAdded(PartitionId partitionId) {
...
             if (!isAmbryListenerToUpdateVcrHelixRegistered) {
+              // Prepare the vcrUpdateDistributedLock. Only one instance can update vcr helix cluster at one time.
+              // It's possible isVcrHelixUpdater to be true on two nodes.
+              // For example, at time "t" node A is the owner of partition 1. Due to some partition reassignment
+              // (lets say new node addition), partition 1 get assigned to node B at time "t+1". In this case it's possible
+              // for Node B to get notification of addPartition of partition 1 at "t+2" before Node A gets removePartition
+              // notification (at t+4). If a main cluster update happens between "t+2" and "t+4", then two nodes might try
+              // to update vcr cluster at the same time. Therefore, we need this distributed lock.
+              LockScope distributedLockScope = new HelixLockScope(HelixLockScope.LockScopeProperty.CLUSTER,
+                  Arrays.asList(cloudConfig.vcrClusterName, cloudConfig.vcrClusterName));
```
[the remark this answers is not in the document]
zzmao: The HelixLockScope will try to create a lock in the path:

```
"/{clusterName}/LOCK/CLUSTER/{clusterName}");
```
2 vcrClusterName will be used to replace two clusterName above. 

I also feel a little bit weird.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -352,24 +371,35 @@ protected boolean shouldReplicateFromDc(String datacenterName) {
...
    * The actual performer to update VCR Helix:
    */
-  synchronized private void updateVcrHelix() {
-    logger.info("Going to update VCR Helix Cluster. Dryrun: {}", cloudConfig.vcrHelixUpdateDryRun);
-    logger.info("Current partitions in clustermap data structure: {}",
+  synchronized private void updateVcrHelix(String reason) {
+
+    logger.info("Going to update VCR Helix Cluster. Reason: {}, Dryrun: {}", reason, cloudConfig.vcrHelixUpdateDryRun);
+    while (!vcrUpdateDistributedLock.tryLock()) {
+      logger.warn("Could not obtain vcr update distributed lock. Sleep and retry.");
```
ankagrawal, line 394: The only way for a vcr node to get out of this while loop, is to have helix assign parttionid 1 to it, right?
What if that never happens., and helix never assigns partition 1 to this node. The vcr process will have to be restarted to get out of this loop. Should we limit the number of times sleep can happen in line 396? Something like, if we have slept and tried for lock 5 times, then its better to give up, because something else might be wrong. If we miss updating the cluster somehow, then anyways we will get notification due to alerts on the metric that checks if the cluster are in sync.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -383,10 +413,27 @@ synchronized private void updateVcrHelix() {
       logger.warn("VCR Helix cluster update failed: ", e);
     } finally {
       isVcrHelixUpdateInProgress = false;
+      vcrUpdateDistributedLock.unlock();
     }
     logger.info("VCR Helix cluster update done.");
   }
 
+  private void checkAmbryHelixAndVcrHelixOnSync() {
```
ankagrawal, line 421: javadocs?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -383,10 +413,27 @@ synchronized private void updateVcrHelix() {
...
+    try {
+      String localDcZkStr = ((HelixClusterManager) clusterMap).getLocalDcZkConnectString();
+      isSrcAndDstSync = HelixVcrUtil.isSrcDestSync(localDcZkStr, clusterMapConfig.clusterMapClusterName,
+          cloudConfig.vcrClusterZkConnectString, cloudConfig.vcrClusterName);
+    } catch (Exception e) {
+      logger.warn("Ambry Helix and Vcr Helix sync check runs into exception: ", e);
+    }
+    if (!isSrcAndDstSync) {
+      logger.warn("Ambry Helix cluster and VCR helix cluster are not on sync");
+      // Raise alert on this metric
```
ankagrawal, line 432: should we do this check here before logging the metric

if(vcrHelixUpdateFuture == null && !isSrcAndDstSync) {
    // this means no update is scheduled or is in progress, and we still find that src and dest are out of sync
    vcrMetrics.vcrHelixNotOnSync.inc();
}

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
```diff
@@ -266,6 +280,37 @@ private static boolean isDelayedRebalanceEnabled(String rebalancerClassName) {
...
+      if (HelixVcrUtil.ignoreResourceKeyWords.stream().anyMatch(resource::contains)) {
+        System.out.println("Resource " + resource + " from src cluster is ignored");
+        continue;
+      }
+      if (destResources.contains(resource)) {
+        // check if every partition exist.
+        Set<String> srcPartitions = srcAdmin.getResourceIdealState(srcClusterName, resource).getPartitionSet();
+        Set<String> destPartitions = destAdmin.getResourceIdealState(destClusterName, resource).getPartitionSet();
+        for (String partition : srcPartitions) {
+          if (!destPartitions.contains(partition)) {
```
ankagrawal, line 303: I know this might not happen, but should we also add condition (destPatitions.size() == srcPartitions.size()) and another check for (destResources.size() = srcResources.size()).  -- this will check if somehow vcr cluster has extra partitions or resources that main cluster somehow doesn't have (can't think of any way this scenario can happen, so feel free to ignore this comment if you want)

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: good point. added.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: added.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
zzmao: the lock is not associated to partition 1. It's a general distributed lock on zk. It will eventually get the lock no matter if the node has partition 1 or not. 

I agree with the retry count concern. I can add a config for max retry count.

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
zzmao: We can add the extra check when this happen. for now, I don't see this scenario could happen. 

If this happen, we need a vcr cluster rebuild, which is better to be done by manually cli.

cgtz (approved):

ankagrawal (approved):

PR #1956, opened 2021-08-31
Fix NullPointerException when journal is empty

SophieGuo410:
Fix race condition when we call journal.getFirstOffset method, it returns null because journal is just cleaned up due to autoCloseLastLogSegment but not covered by "entries != null".

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -312,8 +312,10 @@ class PersistentIndex {
    */
   private void removeIndexEntryNotBelongsToLastLogSegmentFromJournal() {
     LogSegmentName lastLogSegmentName = log.getLastSegment().getName();
-    while (journal.getFirstOffset() != null && !journal.getFirstOffset().getName().equals(lastLogSegmentName)) {
-      journal.removeSpecificValueInJournal(journal.getFirstOffset());
+    Offset journalFirstOffset = journal.getFirstOffset();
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1275,7 +1275,8 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
         // recheck to make sure that offsetToStart hasn't slipped out of the journal. It is possible (but highly
         // improbable) that the offset has slipped out of the journal and has been compacted b/w the time of getting
         // entries from the journal to when another view of the index segments was obtained.
-        if (entries != null && offsetToStart.compareTo(journal.getFirstOffset()) >= 0) {
+        if (entries != null && journal.getFirstOffset() != null
+            && offsetToStart.compareTo(journal.getFirstOffset()) >= 0) {
```
justinlin-linkedin, line 1279: this would not fix the race condition, we should probably do 
```
Offset journalFirstOffset = journal.getFirstOffset();
if (entries != null && journalFirstOffset != null && offsetToStart.compareTo(journalFirstOffset)) >= 0) {
}
```

justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Sure, Updated it and updated other parts which call journal.getFirstOffset() twice.

cgtz (approved):

PR #1960, opened 2021-09-08
Fix a race condition in IndexSegment

justinlin-linkedin:
There is a race condition between `IndexSegment.seal` and `IndexSegment.find` method.
In `seal` method, we first set the `sealed` to be true, then lock the write lock of `rwLock`, and load the data from file to `serEntries`.
In `find` method, we first lock the read lock of `rwLock` and then check if `sealed` is true or not. This would cause a race condition.

It would goes something like this
1. find calls rwLock.readLock().lock()
2. seal sets `sealed` to true
3. find checks if `sealed` is true or not. In this time, it's true, and find try to read entries from `serEntries`, which is empty. This is where NPE gets thrown.

A fix has to put in place to address this.
With the PR, now
1. find calls rwLock.readLock().lock()
2. seal calls rwLock.writeLock().lock() and it will be blocked by the read lock.
3. find checks if `sealed` is true or not. In this time, it's not true, and find try to read entries from map.


tbingol (commented):
Shouldn't you get read lock anytime you access "sealed", e,g, in size() ?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -828,10 +828,15 @@ void writeIndexSegmentToFile(Offset safeEndPoint) throws FileNotFoundException,
    * @throws StoreException if there are problems with the index
    */
   void seal() throws StoreException {
-    sealed.set(true);
-    map();
-    // we should be fine reading bloom filter here without synchronization as the index is read only
-    persistBloomFilter();
+    rwLock.writeLock().lock();
+    try {
+      sealed.set(true);
```
tbingol, line 833: if an exception is raised, should we undo sealed=true?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -828,10 +828,23 @@ void writeIndexSegmentToFile(Offset safeEndPoint) throws FileNotFoundException,
...
+      map();
+      // we should be fine reading bloom filter here without synchronization as the index is read only
```
cgtz, line 835: do we want to move this outside of the rwlock if its okay to read w/o synchronization? then we wouldn't need the index != null check in the catch block

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -828,10 +828,23 @@ void writeIndexSegmentToFile(Offset safeEndPoint) throws FileNotFoundException,
...
+      // If the in-memory map is already set to null, it means we are here because we failed writing the bloom filter
+      // to disk. The index segment can still function with bloom filter file being missing, so don't know anything
+      // in this case.
+      if (index != null) {
+        sealed.set(false);
+      }
+      throw e;
+    } finally {
+      rwLock.writeLock().unlock();
+    }
```
cgtz, line 847: There are a lot of places where sealed.get() is called and it seems like those could exhibit similar issues (albeit less frequently called methods than find). We may want to make a helper method to check for sealed state that always acquires the read lock

cgtz (changes requested):
LGTM after comments addressed

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -372,10 +381,15 @@ short getResetKeyLifeVersion() throws StoreException {
...
+    // With rwLock, "serEntries != null" is an atomic operation with "sealed.get() == true", which means,
+    // when sealed.get() == true, then serEntries will not be null.
+    // Also, getting reference of index before checking sealed.get() would make sure tha when sealed.get()
+    // returns false, the reference will always be not null.
+    NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> indexCopy = index;
     try {
-      if (!sealed.get()) {
-        ConcurrentSkipListSet<IndexValue> values = index.get(keyToFind);
+      if (!isSealedLocked()) {
+        ConcurrentSkipListSet<IndexValue> values = indexCopy.get(keyToFind);
```
tbingol, line 392: Once isSealedLocked() returns, no more lock is held. What if it returns false and just before you are executing the else branch it is set to true by another thread.

We need to get and keep the lock within the whole try block.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -275,13 +287,10 @@ File getFile() {
    * @return number of IndexEntry items in this segment.
    */
   int size() {
-    if (sealed.get()) {
-      if (serEntries == null) {
-        return 0;
-      }
+    if (isSealedLocked()) {
```
tbingol, line 290: You need to get and hold the lock for the whole duration of this function.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -838,8 +848,10 @@ void seal() throws StoreException {
    * @return index value of last PUT record in this index segment. Return {@code null} if no PUT is found
    */
   IndexValue getIndexValueOfLastPut() throws StoreException {
+
     IndexValue indexValueOfLastPut = null;
-    if (sealed.get()) {
+    NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> indexCopy = index;
+    if (isSealedLocked()) {
```
tbingol, line 854: Same as above. You execute the critical path without holding the lock. You need to ensure isSealed is consistent throughout the execution of this routine. So, you need to get the lock in the beginning of this routine and hold it until you return.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -1149,10 +1158,11 @@ boolean getEntriesSince(StoreKey key, FindEntriesCondition findEntriesCondition,
...
    */
   public Iterator<IndexEntry> iterator() {
-    if (!sealed.get()) {
-      return new UnsealedIndexSegmentEntryIterator();
+    NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> indexCopy = index;
+    if (!isSealedLocked()) {
+      return new UnsealedIndexSegmentEntryIterator(indexCopy);
     }
-    return new SealedIndexSegmentEntryIterator();
+    return new SealedIndexSegmentEntryIterator(serEntries);
```
tbingol, line 1165: Can you be sure UnsealedIndexSegmentEntryIterator() is called when isSealed=false? What if another thread sets it to true as soon as isSealedLocked() returns. You need to get and hold the lock throughout this routine.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -1175,10 +1185,11 @@ boolean getEntriesSince(StoreKey key, FindEntriesCondition findEntriesCondition,
...
    */
   ListIterator<IndexEntry> listIterator(int idx) {
-    if (!sealed.get()) {
-      return new UnsealedIndexSegmentEntryListIterator(idx);
+    NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> indexCopy = index;
+    if (!isSealedLocked()) {
+      return new UnsealedIndexSegmentEntryListIterator(indexCopy, idx);
     }
-    return new SealedIndexSegmentEntryListIterator(idx);
+    return new SealedIndexSegmentEntryListIterator(serEntries, idx);
```
tbingol, line 1192: Same as Iterator

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -1200,9 +1211,12 @@ boolean getIndexEntriesSince(StoreKey key, FindEntriesCondition findEntriesCondi
     if (!findEntriesCondition.proceed(currentTotalSizeOfEntriesInBytes.get(), getLastModifiedTimeSecs())) {
       return false;
     }
+
+    // These variables would change with sealed boolean
+    NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> indexCopy = index;
     NavigableSet<IndexValue> values = new TreeSet<>();
     List<IndexEntry> entriesLocal = new ArrayList<>();
-    if (sealed.get()) {
+    if (isSealedLocked()) {
```
tbingol, line 1219: Same as getIndexValueOfLastPut()

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -1259,7 +1273,7 @@ boolean getIndexEntriesSince(StoreKey key, FindEntriesCondition findEntriesCondi
    * @return the first key (in lexicographical order) of sealed index segment
    */
   StoreKey getFirstKeyInSealedSegment() throws StoreException {
-    if (!sealed.get()) {
+    if (!isSealedLocked()) {
```
tbingol, line 1276: same as getIndexValueOfLastPut()

tbingol (changes requested):

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
justinlin-linkedin: I have different idea about the lock here. It's actually fine to not hold the lock through out the entire method.
First, all the changes are done in read-only methods, so there will be no concurrent writes here. 
Secondly, sealed or unsealed, IndexSegment has the same information, just managed in different way. For example, this find method, if after `isSealedLock()` returns false, and other thread set it to true right way, we can still use `indexCopy` map to find the index values. Same logic applies to other methods.

tbingol: I am a little confused. If this is a single-threaded execution, why do we need to protect the "sealed" in the first place?

For the second part:
  - If the critical path does not involve the whole routine but just the subset of it, you can reduce the scope of the lock.
  - If segment's being locked or not does not make any difference:
    + Can we guarantee that it will always be true in the future? 
    + Can we consider having a single code path?

tbingol (changes requested):

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
justinlin-linkedin: this is not a single-thread execution of course. When an IndexSegment is created, it is unsealed. It has a map from blob id to index entries. This map is the variable `index`. After it becomes sealed, this map will be persisted to disk and loaded back to memory in a `ByteBuffer`. This `ByteBuffer` is the variable `serEntries`. 

The thread that changes index segment from unsealed to sealed is a background thread and we also have multiple read threads that returns index values from a given blob id. 

And checking if the index segment is sealed or not will help us determine which variable to use when returning index values. If index segment is sealed, then we should use `serEntries` buffer, if not, we should use `index` map. So as long as we know which variable to use, we can return the index values from this variable without locking, because if it's sealed, buffer `serEntries` is a read only buffer. if it's not sealed, `index` map is a concurrent hash map that can handles concurrent reads and writes.

tbingol: Thanks for the explanation.

Since you make a copy of `index` before calling isSealedLocked(), this would work. Having said that it is not a good practice in my opinion. Ideally, all the critical code path needs to be executed with the read-lock in order to be future-proof.

I don't expect you to make the changes in this PR but just noting the potential issue going forward.

justinlin-linkedin: No urgency for this PR, I will update all.

PR #1961, opened 2021-09-14
Update AccountStatsStore to include new ContainerStorageStats

justinlin-linkedin:
Update AccountStatsStore interfaces and its implementations

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/server/storagestats/ContainerStorageStats.java
+++ b/ambry-api/src/main/java/com/github/ambry/server/storagestats/ContainerStorageStats.java
@@ -157,7 +157,7 @@ public class ContainerStorageStats {
     }
 
     /**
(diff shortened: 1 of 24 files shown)


zzmao (approved):
LGTM

ambry-api/src/main/java/com/github/ambry/server/storagestats/ContainerStorageStats.java
```diff
@@ -148,6 +156,17 @@ public Builder(short containerId) {
       this.containerId = containerId;
     }
 
+    /**
+     * Constructor to instantiate a new {@link Builder} with the given {@link ContainerStorageStats}.
+     * @param origin The given {@link ContainerStorageStats}.
+     */
+    public Builder(ContainerStorageStats origin) {
```
ankagrawal, line 163: nit: Do we need both the Builder and the copy constructor? Looks like they are doing the same thing.

ankagrawal (approved):

PR #1962, opened 2021-09-14
HTTP2 support for VCR replication

zzmao:
HTTP2 support for VCR replication

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
@@ -170,8 +170,10 @@ public class VcrServer {
       if (clusterMapConfig.clusterMapEnableHttp2Replication) {
         connectionPool = new Http2BlockingChannelPool(sslFactory, new Http2ClientConfig(properties),
             new Http2ClientMetrics(registry));
+        logger.info("Using http2 for VCR replication.");
       } else {
         connectionPool = new BlockingChannelConnectionPool(connectionPoolConfig, sslConfig, clusterMapConfig, registry);
+        logger.info("Using blocking channel for VCR replication.");
       }
       connectionPool.start();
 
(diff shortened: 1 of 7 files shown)


zzmao:
Added @justinlin-linkedin  as reviewer.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
```diff
@@ -161,7 +166,13 @@ public void startup() throws InstantiationException {
       scheduler = Utils.newScheduler(serverConfig.serverSchedulerNumOfthreads, false);
       StoreKeyFactory storeKeyFactory = Utils.getObj(storeConfig.storeKeyFactory, clusterMap);
 
-      connectionPool = new BlockingChannelConnectionPool(connectionPoolConfig, sslConfig, clusterMapConfig, registry);
+      SSLFactory sslFactory = new NettySslHttp2Factory(sslConfig);
+      if (clusterMapConfig.clusterMapEnableHttp2Replication) {
+        connectionPool = new Http2BlockingChannelPool(sslFactory, new Http2ClientConfig(properties),
```
ankagrawal, line 171: nit: maybe log here if we are using http2 or blocking channel connection pool

ankagrawal (approved):

ambry-network/src/main/java/com/github/ambry/network/BlockingChannelInfo.java
```diff
@@ -169,6 +169,8 @@ private BlockingChannel getBlockingChannelBasedOnPortType(String host, int port)
       channel = new BlockingChannel(host, port, config);
     } else if (this.port.getPortType() == PortType.SSL) {
       channel = new SSLBlockingChannel(host, port, registry, config, sslSocketFactory, sslConfig);
+    } else {
+      logger.error("No port type: {}", this.port.getPortType());
```
cgtz, line 173: Returning null here in this case seems a bit weird to me instead of throwing an exception, or throwing an exception in the constructor if it is misconfigured. Is there a reason for that behavior? I was also wondering why this private method takes in the port number instead of using `port.getPort()` (I realize this may be very old code).

cgtz (approved):
LGTM. One Q about BlockingChannelInfo

ambry-network/src/main/java/com/github/ambry/network/BlockingChannelInfo.java
zzmao: Updated the code to throw an exception. (BTW, The piece of  code will be deprecated after we no longer need blocking channel.)

PR #1963, opened 2021-09-14
Fix race condition for token generation

SophieGuo410:
Check the journal before index segments to fix race condition because when we do the put, the indexSegments will be updated before journal.


zzmao:
You saw race condition caused issue on prod/ei?

zzmao:
Change LGTM.

SophieGuo410:
I saw some exception will be triggered, but it will be gone for the next round retry. It won't cause any functionality trouble, just to mitigate the exception.

justinlin-linkedin (approved):

PR #1964, opened 2021-09-19
Remove JSONObject from account and container

justinlin-linkedin:
Totally remove the JSONObject from account and container class.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
@@ -13,10 +13,12 @@
  */
 package com.github.ambry.account;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.ambry.commons.Notifier;
 import com.github.ambry.commons.TopicListener;
 import com.github.ambry.config.AccountServiceConfig;
 import com.github.ambry.server.StatsSnapshot;
+import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -43,6 +45,7 @@ import org.slf4j.LoggerFactory;
 public abstract class AbstractAccountService implements AccountService {
 
   private static final Logger logger = LoggerFactory.getLogger(AbstractAccountService.class);
+  private static final ObjectMapper objectMapper = new ObjectMapper();
   static final String ACCOUNT_METADATA_CHANGE_TOPIC = "account_metadata_change_topic";
   static final String FULL_ACCOUNT_METADATA_CHANGE_MESSAGE = "full_account_metadata_change";
   protected final AtomicReference<AccountInfoMap> accountInfoMapRef;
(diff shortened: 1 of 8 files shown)


zzmao (approved):
LGTM

ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
```diff
@@ -186,8 +189,13 @@ public Account getAccountById(short id) {
               if (existingContainer.isSameContainer(container)) {
                 // If an exactly same container already exists, treat as no-op (may be retry after partial failure).
                 // But include it in the list returned to caller to provide the containerId.
-                logger.info("Request to create container with existing name and properties: {}",
-                    existingContainer.toJson().toString());
+                String containerStr;
+                try {
+                  containerStr = objectMapper.writeValueAsString(existingContainer);
+                } catch (IOException e) {
+                  containerStr = existingContainer.toString();
```
ankagrawal, line 196: This basically means that processing the json has failed, meaning that there is something wrong with the container object. Should we bother trying to make progress after this? Maybe try throwing AccountServiceException here?

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -401,14 +401,20 @@ private void bindAccount(PreparedStatement statement, Account account, Statement
    */
   private void bindContainer(PreparedStatement statement, int accountId, Container container,
       StatementType statementType) throws SQLException {
+    String containerInJson;
+    try {
+      containerInJson = objectMapper.writeValueAsString(container);
+    } catch (IOException e) {
+      throw new SQLException("Fail to serialize container: " + container.toString(), e);
```
ankagrawal, line 408: It doesn't make sense to throw a java.sql.SqlException for deserialization failure. Maybe throw IOException?

ambry-api/src/test/java/com/github/ambry/account/AccountContainerTest.java
```diff
@@ -903,6 +903,137 @@ public void testAccountAndContainerSerDeWithoutIdAndName() throws IOException {
...
+            .lastModifiedTime(lastModifiedTime)
+            .aclInheritedByContainer(aclInheritedByContainer)
+            .containers(containers)
+            .build();
+      default:
+        throw new IllegalStateException("Unsupported account json version=" + metadataVersion);
+    }
+  }
+
+  private Container containerFromJson(JSONObject metadata, short parentAccountId) throws JSONException {
```
ankagrawal, line 944: javadocs?

ambry-api/src/test/java/com/github/ambry/account/AccountContainerTest.java
```diff
@@ -903,6 +903,137 @@ public void testAccountAndContainerSerDeWithoutIdAndName() throws IOException {
     }
   }
 
+  private Account accountFromJson(JSONObject metadata) throws JSONException {
```
ankagrawal, line 906: javadocs?

ankagrawal (commented):
Looks good overall. Some minor comments.

ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
justinlin-linkedin: serialize a container to a json in string usually doesn't fail. If it does, let's just print out the container name and id in the log message. This is what this code does. I don't think we should throw out an exception since we are serializing container just for log messages. I would rather just use container.toString() all the time if not for backward compatibility.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
justinlin-linkedin: This method only throws out an SQLException, that's why I am throwing out SQLException here even for serialization error.

ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
ankagrawal: Got it. Didn't realize that this was just for log message.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
ankagrawal: We can change the signature to include the correct exception.

ankagrawal (approved):

PR #1966, opened 2021-09-22
StatsManager starts collecting new storage stats

justinlin-linkedin:
StatsManager starts fetching the new storage stats from blobstorestats and save it to database.

### DIFF
--- a/ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
@@ -26,6 +26,9 @@ import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.clustermap.StateModelListenerType;
 import com.github.ambry.clustermap.StateTransitionException;
 import com.github.ambry.config.StatsManagerConfig;
+import com.github.ambry.server.storagestats.ContainerStorageStats;
+import com.github.ambry.server.storagestats.HostAccountStorageStats;
+import com.github.ambry.server.storagestats.HostPartitionClassStorageStats;
 import com.github.ambry.store.StorageManager;
 import com.github.ambry.store.Store;
 import com.github.ambry.store.StoreException;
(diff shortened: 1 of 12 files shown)


ambry-tools/src/main/java/com/github/ambry/cloud/azure/AzureCompactionTool.java
```diff
@@ -133,6 +133,11 @@ private PartitionPathId(String path) {
       return new byte[0];
     }
 
+    @Override
+    public long getId()  {
+      return 0;
```
zzmao, line 138: 0 is a valid id. maybe you can return -1 to catch issue earlier if this class is widely used.

zzmao (approved):
LGTM

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -420,17 +377,13 @@ public void run() {
...
+          HostAccountStorageStatsWrapper statsWrapper =
+              new HostAccountStorageStatsWrapper(statsHeader, new HostAccountStorageStats(hostStorageStatsMap));
           // 5. Persist this statsWrapper to mysql database if connection exists.
-          if (accountStatsStore != null) {
-            accountStatsStore.storeAccountStats(statsWrapper);
-          }
-          // 6. Persist this statsWrapper to local backup file.
-          publish(statsWrapper);
-          logger.info("Local account stats snapshot published to {}", statsOutputFile.getAbsolutePath());
+          accountStatsStore.storeHostAccountStorageStats(statsWrapper);
```
SophieGuo410, line 386: nit: I feel it's better to check accountStatsStore != null inside this method or constructor instead of outside of PartitionClassStatsPublisher.

SophieGuo410 (approved):
LGTM!

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
justinlin-linkedin: StatsManager would start two threads to run the publishers when the `accountStatStore` is not `null`. If we check the statement inside the publishers then we are creating the threads even if we don't do anything with the threads. That seems like a waste for threads.

SophieGuo410: I see. thx for the explanation.

PR #1970, opened 2021-09-28
Fix NPE error in changeIndexSegments

justinlin-linkedin:
getIndexSegmentDetails has this issue where it would return wrong file list for the given LogSegmentFile. It's because the PersistentIndex.getIndexSegmentFilesForLogSegment created a file filter based on matching filenames. It will include some unexpected files, for example

If we are currently undergoing a compaction and we have 0_1 and 0_10 log segments exist at the same time, using this method would return not only the index segment files for 0_1 but also the index segment files of 0_10. This would cause an issue in the changeIndexSegments method, since it removes all the index segment files of 0_1 at the end of compaction cycle. That's why we are trying to remove index segment files of 0_10 and getting a NPE.


SophieGuo410 (approved):
Thank you for fixing it!

PR #1971, opened 2021-09-28
Update BlobStoreStats after compaction

justinlin-linkedin:
ScanResults in BlobStoreStats contains some information that will be obsolete after each compaction, like log segment cache, like physical storage usage cache.

This PR would update the ScanResults after each compaction cycle to keep those in-memory cache updated.


ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -202,6 +208,37 @@
...
+    if (bucketCount > 0) {
+      // Compaction is finished, we need to reconstruct the log segments and physical storage usage in the scan result.
+      scanLock.lock();
+      try {
+        long delay = indexScannerScheduledFuture.getDelay(TimeUnit.SECONDS);
+        if (delay > 0) {
+          // We haven't started the next index scanner tasks, just cancel it and reschedule.
+          indexScannerScheduledFuture.cancel(false);
+          indexScannerScheduledFuture = longLiveTaskScheduler.scheduleAtFixedRate(indexScanner, 0,
+              TimeUnit.MILLISECONDS.toSeconds(bucketCount * bucketSpanTimeInMs), TimeUnit.SECONDS);
```
SophieGuo410, line 224: seems like the bucketSpanTimeInMs unit is minutes instead of MILLISECONDS

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -128,17 +132,19 @@
...
     this.enableBucketForLogSegmentReports = enableBucketForLogSegmentReports;
     this.enablePurgeDeleteTombstone = enablePurgeDeleteTombstone;
+    this.longLiveTaskScheduler = longLiveTaskScheduler;
+    this.shortLiveTaskScheduler = shortLiveTaskScheduler;
 
     if (bucketCount > 0) {
       indexScanner = new IndexScanner();
-      longLiveTaskScheduler.scheduleAtFixedRate(indexScanner, 0,
+      indexScannerScheduledFuture = longLiveTaskScheduler.scheduleAtFixedRate(indexScanner, 0,
           TimeUnit.MILLISECONDS.toSeconds(bucketCount * bucketSpanTimeInMs), TimeUnit.SECONDS);
```
cgtz, line 141: why convert from milliseconds back to seconds here as opposed to setting the unit arg to milliseconds?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -202,6 +208,37 @@
...
+   */
+  public void onCompactionFinished() {
+    if (bucketCount > 0) {
+      // Compaction is finished, we need to reconstruct the log segments and physical storage usage in the scan result.
+      scanLock.lock();
+      try {
+        long delay = indexScannerScheduledFuture.getDelay(TimeUnit.SECONDS);
+        if (delay > 0) {
+          // We haven't started the next index scanner tasks, just cancel it and reschedule.
+          indexScannerScheduledFuture.cancel(false);
```
cgtz, line 222: What's the reason for rescheduling the task to a later time in this case instead of just letting it run (since it has not started running yet)?

SophieGuo410 (approved):
LGTM.

PR #1972, opened 2021-09-28
Add gauge metric to record the number of stores skipping the compaction

justinlin-linkedin:


ambry-store/src/main/java/com/github/ambry/store/StorageManagerMetrics.java
```diff
@@ -97,6 +103,11 @@ public StorageManagerMetrics(MetricRegistry registry) {
     Gauge<Long> compactionsInProgressGauge = compactionsInProgress::longValue;
     registry.register(MetricRegistry.name(CompactionManager.class, "CompactionsInProgress"),
         compactionsInProgressGauge);
+
+    Gauge<Long> storeToSkipGauge =
+        () -> mountPathToStoreSetToSkipInCompactions.values().stream().mapToLong(Set::size).sum();
+    registry.gauge(MetricRegistry.name(CompactionManager.class, "CompactionNumberOfStoresToSkip"),
```
SophieGuo410, line 109: I'm not sure how long it will take for the lambda to be calculated, but better to be aware of it in case it delay the metric producing which trigger dotted metric issue?

SophieGuo410 (approved):

ambry-store/src/main/java/com/github/ambry/store/StorageManagerMetrics.java
justinlin-linkedin: this is fast, it's just a stream operations with no more than 130 elements.

zzmao (approved):
LGTM.  Double check CPU usage after deployement.

PR #1973, opened 2021-09-29
Don't keep all blobIds in blob store stats, only record number of put keys

justinlin-linkedin:
Don't keep a set of blob ids for each account container, since it has a huge memory footprint.


SophieGuo410 (approved):
Thanks for making those changes, lgtm!

cgtz (approved):

PR #1981, opened 2021-10-21
Avoid serialization of requests when using LocalNetworkClient

Arun-LinkedIn:
Avoid serialization/deserialization of request when using LocalNetworkClient for cloud facing router. These requests are sent from cloud router (in ambry frontend) to AmbryRequest API methods via local queues (running in the same process) and can be referenced directly via the java objects without serializing to byte streams.

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/LocalNetworkClient.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/LocalNetworkClient.java
@@ -16,8 +16,6 @@ package com.github.ambry.network;
 import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.network.LocalRequestResponseChannel.LocalChannelRequest;
 import com.github.ambry.utils.Time;
-import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufInputStream;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
@@ -76,16 +74,17 @@ public class LocalNetworkClient implements NetworkClient {
     // TODO: do anything with requestsToDrop?
     // The AmbryRequest sessions run in a thread pool, and each thread knows when the response is back so we can poll it
     long startTime = time.milliseconds();
-    try {
-      for (RequestInfo requestInfo : requestInfos) {
-        // TODO: inefficient to serialize request before sending, better to convert Send to an InputStream
-        // that handles the header skipping.
-        ByteBuf buffer = LocalRequestResponseChannel.byteBufFromPayload(requestInfo.getRequest());
-        channel.sendRequest(new LocalChannelRequest(requestInfo, processorId, new ByteBufInputStream(buffer)));
+    for (RequestInfo requestInfo : requestInfos) {
+      // Request is sent and received as it is without serializing to stream since they are handled in same process via
+      // local queues.
+      try {
+        channel.sendRequest(new LocalChannelRequest(requestInfo, processorId));
+      } catch (Exception e) {
+        logger.error("Received an unexpected error during sendAndPoll(): ", e);
+        // release related bytebuf
+        requestInfo.getRequest().release();
+        networkMetrics.networkClientException.inc();
       }
-    } catch (Exception e) {
-      logger.error("Received an unexpected error during sendAndPoll(): ", e);
-      networkMetrics.networkClientException.inc();
     }
 
     List<ResponseInfo> responses = channel.receiveResponses(processorId, pollTimeoutMs);
@@ -111,5 +110,19 @@ public class LocalNetworkClient implements NetworkClient {
   @Override
   public void wakeup() {
     channel.wakeup(processorId);
+  }
+
+  /**
+   * @return processorId. Only used in tests.
+   */
+  int getProcessorId() {
+    return processorId;
+  }
+
+  /**
+   * @return true if this instance is closed. Only used in tests.
+   */
+  boolean isClosed() {
+    return closed;
   }
 }
--- a/ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
@@ -14,15 +14,7 @@
 package com.github.ambry.network;
 
 import com.github.ambry.utils.BatchBlockingQueue;
-import com.github.ambry.utils.ByteBufferChannel;
-import io.netty.buffer.ByteBuf;
-import io.netty.buffer.ByteBufOutputStream;
-import io.netty.buffer.Unpooled;
-import java.io.IOException;
 import java.io.InputStream;
-import java.nio.ByteBuffer;
-import java.nio.channels.Channels;
-import java.nio.channels.WritableByteChannel;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.BlockingQueue;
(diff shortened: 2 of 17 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetManager.java
```diff
@@ -220,6 +221,12 @@ void handleResponse(ResponseInfo responseInfo) {
               }
               return serverError;
             });
+    if (responseInfo.getResponse() != null) {
```
zzmao, line 224: 1. Add comment to explain why this case happen?
2. Do others(PutManager, DeleteManager) need this? 
3. Can we use a setMethod to set the ByteBufInputStream? Is it possible to set this inside extractResponseAndNotifyResponseHandler?

zzmao: Looks like we still parse the content afterwards. Is my understanding correct?

zzmao: If the goal is to fully avoid deserialization, we probably need a little bit more work to avoid deserialization in GetBlobOperation: handleBody.

Let me know if you want to have a offline chat.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -110,46 +112,69 @@ public AmbryRequests(StoreManager storeManager, RequestResponseChannel requestRe
...
+      // accessed via {@code PutRequest#blobStream} at the receiver side where as it is carried via {@code PutRequest#blob}
+      // at the sender side. Also, using null crc value since the request is sent and received in the same process via
+      // local queues.
+      // Since we are creating ByteBufInputStream from the bytebuf object in sent request, setting releaseOnClose to
+      // true to release it once the stream is closed.
+      receivedRequest =
+          new PutRequest(sentRequest.getCorrelationId(), sentRequest.getClientId(), sentRequest.getBlobId(),
+              sentRequest.getBlobProperties(), sentRequest.getUsermetadata(), sentRequest.getBlobSize(),
+              sentRequest.getBlobType(), sentRequest.getBlobEncryptionKey(),
+              new ByteBufInputStream(sentRequest.getBlob(), true), null);
```
cgtz, line 172: What was the reason for not carrying over the CRC? The CRC is also used for some use cases like exact duplicate put entry detection (although some of these may not apply often for cloud use cases since there is no intracolo replication).

ambry-router/src/main/java/com/github/ambry/router/RouterUtils.java
```diff
@@ -245,4 +257,30 @@ static Port getPortToConnectTo(ReplicaId replicaId, boolean routerEnableHttp2Net
...
+   * the message content is held by {@code GetResponse#toSend} object at the sender side while it is held and accessed
+   * as {@code GetResponse#stream} at the receiver side. Hence, we need to construct new {@link Response} objects before
+   * using them. This method is applicable when sending and receiving responses as java objects (instead of serialized
+   * bytes) in the same process via {@link LocalNetworkClient}.
+   * At the moment, other responses like {@link PutResponse}, {@link DeleteResponse}, etc don't need any special
+   * handling and can be referenced as they are.
+   * @param sentResponse {@link Response} object constructed at sender side.
+   * @return {@link Response} object constructed at receiver side.
+   */
+  static Response mapToReceivedResponse(Response sentResponse) {
```
cgtz, line 272: this can be private

ambry-router/src/main/java/com/github/ambry/router/RouterUtils.java
```diff
@@ -245,4 +257,30 @@ static Port getPortToConnectTo(ReplicaId replicaId, boolean routerEnableHttp2Net
...
+    Response receivedResponse;
+    if (sentResponse instanceof GetResponse) {
+      GetResponse getResponse = (GetResponse) sentResponse;
+      // Not setting releaseOnClose to true when creating ByteBufInputStream since the underlying bytebuf is released
+      // when the Response object is closed as part of {@code ResponseInfo#release}.
+      receivedResponse = new GetResponse(getResponse.getCorrelationId(), getResponse.getClientId(),
```
cgtz, line 278: What would be the level of effort to build the functionality to expose `toSend.content()` and `blob.content()` as an input stream directly (by lazily constructing it) into the `PutRequest` and `GetResponse` objects? This code looks like it could be error prone if other constructors get added or other types of requests/responses need this logic.

Alternatively, what does it take to be able to stop using the `getInputStream` method in these cases?

ambry-network/src/test/java/com/github/ambry/network/LocalRequestResponseChannelTest.java
```diff
@@ -0,0 +1,102 @@
...
+import com.github.ambry.network.LocalRequestResponseChannel.LocalChannelRequest;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+
+public class LocalRequestResponseChannelTest {
```
cgtz, line 27: Are there specific tests with leak detection enabled where you have verified the ref count logic?

ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
```diff
@@ -114,24 +102,10 @@ public void wakeup(int processorId) {
...
-    return buffer;
-  }
-
   /**
    * A {@link NetworkRequest} implementation that works with {@link LocalRequestResponseChannel}.
    */
-  static class LocalChannelRequest implements NetworkRequest {
+  public static class LocalChannelRequest implements NetworkRequest {
     private RequestInfo requestInfo;
     private InputStream input;
```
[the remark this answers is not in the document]
Arun-LinkedIn: There are no cases left where `input` is supplied. Removed it as constructor argument.

ambry-network/src/main/java/com/github/ambry/network/LocalNetworkClient.java
```diff
@@ -78,10 +76,9 @@ public LocalNetworkClient(LocalRequestResponseChannel channel, int processorId,
...
-        // TODO: inefficient to serialize request before sending, better to convert Send to an InputStream
-        // that handles the header skipping.
-        ByteBuf buffer = LocalRequestResponseChannel.byteBufFromPayload(requestInfo.getRequest());
-        channel.sendRequest(new LocalChannelRequest(requestInfo, processorId, new ByteBufInputStream(buffer)));
+        // Request is sent and received as it is without serializing to stream since they are handled in same process via
+        // local queues.
+        channel.sendRequest(new LocalChannelRequest(requestInfo, processorId, null));
       }
     } catch (Exception e) {
       logger.error("Received an unexpected error during sendAndPoll(): ", e);
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Released the requests when we hit an exception.

ambry-network/src/test/java/com/github/ambry/network/LocalRequestResponseChannelTest.java
Arun-LinkedIn: For verifying the ref count logic, I ran test cases in `CloudRouterTest` with leak detection enabled. It seems to exercise putting and getting blobs from `LatchBasedInMemoryCloudDestination` via `LocalNetworkClient`.

zzmao (approved):

cgtz (approved):

PR #1984, opened 2021-10-28
Add http2 service on VCR

zzmao:
Added http2 service on VCR server.
Added tests.


ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
```diff
@@ -208,6 +226,32 @@ public void startup() throws InstantiationException {
 
       networkServer.start();
 
+      // Start netty http2 server
+      if (currentNode.hasHttp2Port()) {
```
ankagrawal, line 230: This part is shared between AmbryServer and VcrServer. Can we separate this out in an util method that both AmbryServer and VcrServer can share?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
```diff
@@ -233,6 +277,12 @@ public void shutdown() {
       if (networkServer != null) {
         networkServer.shutdown();
       }
+      if (requestHandlerPoolForHttp2 != null) {
```
ankagrawal, line 280: Similarly this part is shared too. Can we separate this out into a util method -- `shutDownAndHttp2`

build.gradle
```diff
@@ -426,7 +426,8 @@ project(':ambry-cloud') {
             project(':ambry-utils'),
             project(':ambry-commons'),
             project(':ambry-account'),
-            project(':ambry-replication')
+            project(':ambry-replication'),
+            project(':ambry-server')
```
ankagrawal, line 430: We should not introduce a dependency on `ambry-server` for `ambry-cloud`. This has a potential of bringing a lot of cyclical dependencies in future. Also this will bloat the jars will a lot of unnecessary libraries.

What we can do here is move `StorageServerNettyFactory` and `StorageServerNettyChannelInitializer` to `ambry-rest`. Maybe rename them as `StorageNettyFactory` and `StorageNettyChannelInitializer`. And then import `ambry-rest` instead of `ambry-server`, and also import `ambry-network` in `ambry-rest` project.

zzmao: Comment addressed.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
zzmao: I tried to created a common util for them to share. but the it requires many parameters(almost all variables). It not elegant unfortunately.

zzmao:
@ankagrawal @Arun-LinkedIn   this PR is ready. Hopefully we can merge it this week.

Arun-LinkedIn (approved):
lgtm!

ankagrawal (approved):

PR #1985, opened 2021-11-03
[StorageQuota] Adding EnforcementPolicy to enforce storage quota based on clustermap

justinlin-linkedin:
1. Add a new interface StorageQuotaEnforcementPolicy
2. Add two implementations of such interface
    2.1 A simple enforcement policy, once the current usage exceeds storage quota, throttle the request
    2.2 An adaptive enforcement policy, who also takes cluster capacity into consideration. If the cluster has more than 25% of partitions are writable, don't throttle.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StorageQuotaConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StorageQuotaConfig.java
@@ -124,7 +124,7 @@ public class StorageQuotaConfig {
   public final long adaptiveEnforcementClusterMapUpdateIntervalMs;
 
   /**
-   * Threshold in percentage when percentage of writable partitions drops below this threshold, we adaptive enforcement
+   * Threshold in percentage when percentage of writable partitions drops below this threshold, the adaptive enforcement
    * policy will start throttling upload requests.
    */
   @Config(ADAPTIVE_ENFORCEMENT_WRITEABLE_PARTITION_THRESHOLD)
--- a/ambry-api/src/main/java/com/github/ambry/config/StorageQuotaConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StorageQuotaConfig.java
@@ -27,6 +27,13 @@ public class StorageQuotaConfig {
   public static final String MYSQL_STORE_RETRY_MAX_COUNT = STORAGE_QUOTA_PREFIX + "mysql.store.retry.max.count";
   public static final String SHOULD_THROTTLE = STORAGE_QUOTA_PREFIX + "should.throttle";
   public static final String USE_PHYSICAL_STORAGE = STORAGE_QUOTA_PREFIX + "use.physical.storage";
+  public static final String ENFORCEMENT_POLICY_FACTORY = STORAGE_QUOTA_PREFIX + "enforcement.policy.factory";
+  public static final String ADAPTIVE_ENFORCEMENT_CLUSTERMAP_UPDATE_INTERVAL_MS =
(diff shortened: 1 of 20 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/storage/AdaptiveStorageQuotaEnforcementPolicyFactory.java
```diff
@@ -0,0 +1,74 @@
...
+  public AdaptiveStorageQuotaEnforcementPolicyFactory(ClusterMap clusterMap, StorageQuotaConfig config) {
+    policy = new AdaptiveStorageQuotaEnforcementPolicy(clusterMap, SystemTime.getInstance(), config);
+  }
+
+  @Override
+  public StorageQuotaEnforcementPolicy getPolicy() {
+    return policy;
+  }
+}
+
```
tbingol, line 42: Can you add some comments about this implementation?

ambry-quota/src/main/java/com/github/ambry/quota/storage/AdaptiveStorageQuotaEnforcementPolicyFactory.java
```diff
@@ -0,0 +1,74 @@
...
+        updateClusterMapWritablePartitionPercentage();
+      }
+    }
+    return currentUsage >= storageQuota
+        && currentClusterMapWritablePartitionPercentage <= config.adaptiveEnforcementWritablePartitionThreshold;
+  }
+
+  private void updateClusterMapWritablePartitionPercentage() {
+    this.currentClusterMapWritablePartitionPercentage =
+        100 * clusterMap.getWritablePartitionIds(null).size() / clusterMap.getAllPartitionIds(null).size();
```
tbingol, line 72: Though it should not happen, we better make sure to handle divide by zero case

ambry-api/src/main/java/com/github/ambry/config/StorageQuotaConfig.java
```diff
@@ -102,6 +109,28 @@
...
+
+  /**
+   * Interval to update clustermap data in adaptive enforcement policy.
+   */
+  @Config(ADAPTIVE_ENFORCEMENT_CLUSTERMAP_UPDATE_INTERVAL_MS)
+  @Default("10 * 60 * 1000")
+  public final long adaptiveEnforcementClusterMapUpdateIntervalMs;
+
+  /**
+   * Threshold in percentage when percentage of writable partitions drops below this threshold, we adaptive enforcement
```
tbingol, line 127: Nit: s/we adaptive/the adaptive

tbingol (approved):

ankagrawal (commented):
I don't think its a good idea to allow storage above quota. Storage once used is not a resource we can get back later. If, for example, due to a bug, a customer writes a lot of data to an account, and we allow the write to then continue on free space, then it will be a lot of operational effort for us to get that storage back. Imagine a case where writes due to bugs are also intermingled with legitimate writes. It will be operationally very time taking to figure out all the blobs that were erroneously written and then delete them , and then wait for compaction to catchup.

Even if we do decide to do this, we should do this only on a fraction "unpromised storage capacity" available. This will be the storage capacity of a cluster that is extra and not part of any customer's quota. Note that even with this, since our storage quota is per month, it will be hard to come up with an estimate of how much we can give out above quota.

So for storage quota, I think the cleanest approach will be to allocate storage quota generously, give out warnings and alerts proactively, but not allow customers to exceed quota.

PR #1987, opened 2021-11-03
Fixed the netty memory leak in LocalRequestResponseChannel

JingQianCloud:
Release payload after it was sent to the channel.
Verified the fix with CloudRouterTest unit tests.

Enabled Netty ByteBuf Leak Detector in CloudRouterTest. 
Netty was able to detect the payload memory leak.
With the fix, the CloudRouterTest passed.

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
@@ -125,6 +125,8 @@ public class LocalRequestResponseChannel implements RequestResponseChannel {
     payload.writeTo(new ByteBufferChannel(ByteBuffer.wrap(SIZE_BYTE_ARRAY)));
     WritableByteChannel byteChannel = Channels.newChannel(new ByteBufOutputStream(buffer));
     payload.writeTo(byteChannel);
+    // Just like MockSelector, LocalRequestResponseChannel needs to release the sends' resources after completed.
+    payload.release();
     return buffer;
   }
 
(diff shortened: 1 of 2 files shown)


ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
```diff
@@ -125,6 +125,8 @@ static ByteBuf byteBufFromPayload(Send payload) throws IOException {
     payload.writeTo(new ByteBufferChannel(ByteBuffer.wrap(SIZE_BYTE_ARRAY)));
     WritableByteChannel byteChannel = Channels.newChannel(new ByteBufOutputStream(buffer));
     payload.writeTo(byteChannel);
+    // Just like MockSelector, LocalRequestResponseChannel needs to release the sends' resources after completed.
+    payload.release();
```
justinlin-linkedin, line 129: I don't think this is the best place to release the payload, I would recommend releasing the payload in the caller methods instead of this method.

JingQianCloud: Thanks for the comment.

We have another PR which avoids serialization of requests when using LocalNetworkClient as below. 
https://github.com/linkedin/ambry/pull/1981 
https://github.com/linkedin/ambry/pull/1981/files#diff-f3bc94decabe1a16937bcab2ca59e84358d8e472facd2044745f84dffb46dbfe

Depending on the new code, the problem could still exist or don't. Need verify the memory leak with the new code.

So will onhold this PR.

JingQianCloud:
Close this PR since #1981 is refining related code.
#1981 will reverify this issue.

PR #1988, opened 2021-11-04
Fix a NPE in GetBlob in server caused by compaction

justinlin-linkedin:
Comments describe the reason for this PR. Any idea how to test this?

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -46,6 +46,8 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -124,6 +126,32 @@ class PersistentIndex {
   private static final Logger logger = LoggerFactory.getLogger(PersistentIndex.class);
   private final IndexPersistor persistor = new IndexPersistor();
   private final ScheduledFuture<?> persistorTask;
+
+  // ReadWriteLock to protect index values are always pointing to a valid log segments
+  // In compaction's final steps, persistent index and log will update their internal maps to reflect the result of
+  // compactions. This might have race condition with getBlobReadInfo method
+  // --------------------------------------------------------------------------------------
+  // compaction thread                          | getBlobReadInfo thread
+  // --------------------------------------------------------------------------------------
+  // 1. RenameTempLogSegmentFilenames           | 1. FindKeyToReturnIndexValue
+  // 2. AddNewLogSegmentToLogMap                |
+  // 3. UpdateIndexMap                          |
+  // 4. WaitForLogSegmentRefCountToZero         |
+  // 5. RemoveIndexSegmentFiles                 |
+  // 6. RemoveOldLogSegmentFromLogMap           |
+  // 7. RemoveLogSegmentFiles                   | 2. GetLogSegmentFromIndexValueFromLogMap
+  //                                            | 3. IncreaseLogSegmentRefCount
+  // -------------------------------------------------------------------------------------
+  // If two threads are executing in the above given order, then the GetLogSegmentFromIndexValueFromLogMap would
+  // throw an NullPointerException since RemoveOldLogSegmentFromLogMap would remove it from the map.
+  // This ReadWriteLock would prevent that from happening.
+  // 1. In getBlobReadInfo method, we use read lock to lock all three steps
+  // 2. In compaction thread, we use write lock to lock step 3 (in method changeIndexSegments)
+  // If getBlobReadInfo happens before compaction's step 3, then compaction thread would block at acquiring
+  // the write lock, so getBlobReadInfo's step 2 would not happen after compaction's step6.
+  // If getBlobReadInfo happens after compaction's step 3, then the IndexValue returned by findKey method would reflect
+  // the result of the compaction already, it will not point to any old log segment.
+  private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
 
   /**
    * Creates a new persistent index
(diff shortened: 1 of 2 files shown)


tbingol (changes requested):
I think the synchronization between BlobStore and compaction operations is broken. AFAICT, BlobsStoreCompactor::waitTillRefCountZero() attempts to wait until the reference count on a log segment drop to zero before it can proceed with the clean-up process. However, 
  - waitTillRefCountZero() waits for some certain time before it returns even the reference count on the log segment does not go back to zero. In other words, it is not guaranteed that all operations on the log segment would be drained by the time that function returns
  - I don't see anything that would prevent the ref count go back up to non-zero number again, i.e. there is no barrier for the new BlobStore operations to proceed.

We should properly synchronize the BlobStore operations with the compaction operation, probably using ReadWriteLock, rather than trying to handle exceptions as we are doing right now. BlobStore operations (put/get) would get a read lock on the log segment and the compaction process would get the write lock before proceeding with the clean up process.

justinlin-linkedin:
> I think the synchronization between BlobStore and compaction operations is broken. AFAICT, BlobsStoreCompactor::waitTillRefCountZero() attempts to wait until the reference count on a log segment drop to zero before it can proceed with the clean-up process. However,
> 
> * waitTillRefCountZero() waits for some certain time before it returns even the reference count on the log segment does not go back to zero. In other words, it is not guaranteed that all operations on the log segment would be drained by the time that function returns
> * I don't see anything that would prevent the ref count go back up to non-zero number again, i.e. there is no barrier for the new BlobStore operations to proceed.
> 
> We should properly synchronize the BlobStore operations with the compaction operation, probably using ReadWriteLock, rather than trying to handle exceptions as we are doing right now. BlobStore operations (put/get) would get a read lock on the log segment and the compaction process would get the write lock before proceeding with the clean up process.

You are right, I updated the PR to use a ReadWriteLock to protect the persistent index to make sure all the IndexValues fetched from persistent index would always point to a valid log segment.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -125,6 +127,32 @@ public int compare(IndexEntry e1, IndexEntry e2) {
...
+  // If two threads are executing in the above given order, then the GetLogSegmentFromIndexValueFromLogMap would
+  // throw an NullPointerException since RemoveOldLogSegmentFromLogMap would remove it from the map.
+  // This ReadWriteLock would prevent that from happening.
+  // 1. In getBlobReadInfo method, we use read lock to lock all three steps
+  // 2. In compaction thread, we use write lock to lock step 3 (in method changeIndexSegments)
+  // If getBlobReadInfo happens before compaction's step 3, then compaction thread would block at acquiring
+  // the write lock, so getBlobReadInfo's step 2 would not happen after compaction's step6.
+  // If getBlobReadInfo happens after compaction's step 3, then the IndexValue returned by findKey method would reflect
+  // the result of the compaction already, it will not point to any old log segment.
+  private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
```
tbingol, line 156: The scope of this lock is not correct. This lock needs to be owned by Blobstore and passed to this class. See the comment below.

Also, using a single lock seems to be too coarse. Do you want to block all reads, etc when changeIndexSegments run or prefer blocking access to the index/log segments only impacted by the compaction? I think creating a lock per index, log-segment tuple makes more sense.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1141,28 +1175,33 @@ IndexValue markAsUndeleted(StoreKey id, FileSpan fileSpan, MessageInfo info, lon
...
   BlobReadOptions getBlobReadInfo(StoreKey id, EnumSet<StoreGetOptions> getOptions) throws StoreException {
-    ConcurrentSkipListMap<Offset, IndexSegment> indexSegments = validIndexSegments;
-    IndexValue value = findKey(id);
-    BlobReadOptions readOptions;
-    if (value == null) {
-      throw new StoreException("Id " + id + " not present in index " + dataDir, StoreErrorCodes.ID_Not_Found);
-    } else if (value.isDelete()) {
-      if (!getOptions.contains(StoreGetOptions.Store_Include_Deleted)) {
-        throw new StoreException("Id " + id + " has been deleted in index " + dataDir, StoreErrorCodes.ID_Deleted);
+    rwLock.readLock().lock();
```
tbingol, line 1180: Getting the lock and releasing it upon the conclusion of this function is not sufficient. Blobstore::get function is accessing the log segment pointed by the BlobReadOptions returned by this function. In other words, when the Blobstore::get attempts to access the log segment, the log segment could be already deleted by the compaction. For that reason, this lock needs to be owned by the Blobstore and acquired within the scope of Blobstore::get operation, i.e. acquiring the lock should guarantee that index and log accesses will be coherent.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -467,35 +495,41 @@ private void setEndOffsets() throws StoreException {
...
   void changeIndexSegments(List<File> segmentFilesToAdd, Set<Offset> segmentsToRemove) throws StoreException {
-    Offset journalFirstOffset = journal.getFirstOffset();
-
-    TreeMap<Offset, IndexSegment> segmentsToAdd = new TreeMap<>();
-    for (File indexSegmentFile : segmentFilesToAdd) {
-      IndexSegment indexSegment = new IndexSegment(indexSegmentFile, true, factory, config, metrics, journal, time);
-      if (journalFirstOffset != null && indexSegment.getEndOffset().compareTo(journalFirstOffset) > 0) {
-        throw new IllegalArgumentException("One of the index segments has an end offset " + indexSegment.getEndOffset()
-            + " that is higher than the first offset in the journal " + journalFirstOffset);
+    rwLock.writeLock().lock();
```
tbingol, line 500: You need to acquire the lock only 4 through 7 in your chart above.

tbingol (changes requested):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: It's actually fine in this case since the compaction thread would wait for the log segment's refCount becoming 0. The Lock would make sure that these three steps 1. get index value from persistent index 2 getting log segment from index value 3 increase the refCount of the log segment   would become atomic.
That's why we don't have to lock the BlobStore.get, and only on the PersistentIndex.getBlobReadInfo. When getBlobReadInfo function returns a BlobReadOptions, it already increased the refCount of the log segment. We know the refCount is not 0, so the compaction thread will be waiting for the refCount to become 0 before it remove the log segment from the map and from the disk.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: Commented below, the scope should be sufficient to guarantee the correctness.

Using a single lock should be fine. The write lock would only be used in compaction, which happens either once in 24 hours or once in one hour. And the read lock's performance shouldn't be an issue.

tbingol (changes requested):
Thanks for the explanation. That should work.

How do we test this?

justinlin-linkedin:
> Thanks for the explanation. That should work.
> 
> How do we test this?

The unit test is going to be hard.

zzmao (approved):
LGTM. 

I am OK if unit test is too hard.

tbingol (approved):
Thanks for adding the test. Did we verify that without the fix, the test fails?

justinlin-linkedin:
> Thanks for adding the test. Did we verify that without the fix, the test fails?

I did, removing the rwLock from these two methods, we will get a NPE in getBlobReadInfo method.

tbingol:
> 

Great, thanks.

PR #1992, opened 2021-11-11
SafeRelease netty ByteBuf in GetBlobOperations and fix race condition in PutOperation

justinlin-linkedin:
We are experiencing a memory leak in ambry frontend. According to the ambry logs as well as the netty leak detector log messages, it seems like there are two different ways memory leak might happen in the frontend.

1.  I am seeing some exception stack trace in GetBlobOperation that this IllegalReferenceCounter caused BlobReadableStreamChannel to fail. This might cause memory leak.
```2021/11/07 12:17:35.291 ERROR [ResourceLeakDetector] [epollEventLoopGroup-4-27] [ambry-frontend] [] LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records:
#1:
        com.github.ambry.router.GetBlobOperation$BlobDataReadableStreamChannel$1.onCompletion(GetBlobOperation.java:422)
        com.github.ambry.router.GetBlobOperation$BlobDataReadableStreamChannel$1.onCompletion(GetBlobOperation.java:412)
        com.github.ambry.rest.NettyResponseChannel$Chunk.resolveChunk(NettyResponseChannel.java:909)
```

2. PutOperation has a race condition. PutManager calls PutOperation.fillChunks method in OperationController's thread. PutOperation's ChunkFillerChannel's callback is invoked at different thread. When the client closes the connection before Post is finished, frontend would detect that the channel is inactive and close the channel, which would result in closing the ChunkFillerChannel in PutOperation. Closing ChunkFillerChannel would then release all the data held by the "Complete" or "Ready" PutChunk. But at the same time, the fillChunks method, which runs in a different thread, might change PutChunk's state from "Building" to "Ready", which will not be released some time.

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
@@ -21,7 +21,6 @@ import com.github.ambry.network.NetworkClientErrorCode;
 import com.github.ambry.network.RequestInfo;
 import com.github.ambry.network.ResponseInfo;
 import com.github.ambry.protocol.RequestOrResponse;
-import com.github.ambry.utils.Utils;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
@@ -32,7 +31,6 @@ import io.netty.channel.pool.ChannelPoolMap;
 import io.netty.handler.codec.http.HttpObjectAggregator;
 import io.netty.handler.codec.http2.Http2StreamFrameToHttpObjectCodec;
 import io.netty.util.AttributeKey;
-import io.netty.util.ReferenceCountUtil;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.GenericFutureListener;
 import java.net.InetSocketAddress;
@@ -147,7 +145,6 @@ public class Http2NetworkClient implements NetworkClient {
                         }
                       }
                     }
-                    // When success, handler would release the request.
                   } else {
                     http2ClientMetrics.http2StreamWriteAndFlushErrorCount.inc();
                     logger.warn("Stream {} {} writeAndFlush fail. Cause: {}", streamChannel.hashCode(), streamChannel,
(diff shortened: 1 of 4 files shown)


ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -157,7 +154,6 @@ public void operationComplete(ChannelFuture future) throws Exception {
                       http2ClientResponseHandler.getResponseInfoQueue()
                           .put(new ResponseInfo(requestInfoFromChannelAttr, NetworkClientErrorCode.NetworkError, null));
                     }
-                    ReferenceCountUtil.safeRelease(requestInfo.getRequest());
```
cgtz, line 160: Discussed this offline. This was removed b/c it was found that writeAndFlush will release the buffer in both the success and failure case.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -290,6 +297,8 @@ private PutOperation(RouterConfig routerConfig, NonBlockingRouterMetrics routerM
     metadataPutChunk = new MetadataPutChunk();
     chunkFillerChannel = new ByteBufferAsyncWritableChannel(writableChannelEventListener);
     isEncryptionEnabled = passedInBlobProperties.isEncrypted();
+    restRequest = options.getRestRequest();
```
cgtz, line 300: Nit: since `makeTouchMessage` is only called in the constructor, you could pass restRequest in as a makeTouchMessage argument instead of making it a member variable.

cgtz (approved):

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
zzmao: just want to confirm this statement: "writeAndFlush will release the buffer in both the success and failure case" without any exception?

zzmao (approved):

PR #1993, opened 2021-11-15
Differentiate between not found errors and errors possibly due to offline replicas in OperationTracker

ankagrawal:
In cases where OperationTracker fails operation due to not found, differentiate it with unavailable error if some eligible replicas could have been offline during that time.


ankagrawal:
Reopening this PR to return a retryable error code when an operation fails unreliable due to a replica being down.

PR #1994, opened 2021-11-16
Corp Vcr Cluster Helix Config

ankagrawal:
This is the config to create corp vcr cluster.

### DIFF
--- a/config/vcrhelixconfig/CorpVcrClusterCreationHelixConfig.json
+++ b/config/vcrhelixconfig/CorpVcrClusterCreationHelixConfig.json
@@ -1,0 +1,15 @@
+{
+  "clusterConfigFields": {
+    "maxOfflineInstancesAllowed": 3,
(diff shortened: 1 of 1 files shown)


config/vcrhelixconfig/CorpVcrClusterCreationHelixConfig.json
```diff
@@ -0,0 +1,15 @@
+{
```
cgtz, line 1: I would suggest renaming this to something other than corp since this is our open source codebase. Could you also explain why the parameters differ for this use case?

ankagrawal: I guess I can remove this from here (and maybe check it in the closed source code?). The reason some parameters differ here is because the number of nodes are different in corp cluster.

zzmao (approved):

ankagrawal:
We do not need to add this config on open source code.

PR #1999, opened 2021-11-17
Cloud get blob operation test

JingQianCloud:
Created MockCompositeNetworkClient to test Get Blob Operation together with Cloud Backed data center.
Some limitations or comments:
1. Disk backed Server and Cloud backed Server have different mock implementations. May need unify mock layer interface for both of them. With unified mock layer, we can add more test cases.
2. So far we couldn't simulate different kinds of return error from cloud backed server, related to item 1.
3. Do we have standard alone test to verify that Cloud Backed data center will return status which is compatible with disk backed data center?
4. For all the GetBlobOperation tests inherited, we can run them in CloudGetBlobOperation. But for the ones, it won't add more value, I disable them.
5. I build the tree from the LocalRequestResponseChannel, so the memory leak fix is inherited. Will remove it in the next version.

There are some "Jing TODO" in the code, need some discussion with the team. For example,
1. Do we need test SPECIAL_PARTITION_CLASS for testBlobSizeReplacement?
2. lifeVersion got back from cloud backed server and disk backed server is different.
3. hasFailedOnNotFound, discussion about the fix.

### DIFF
--- a/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
@@ -234,9 +234,11 @@ public class GetBlobOperationTest {
     mockSelectorState.set(MockSelectorState.Good);
     VerifiableProperties vprops = new VerifiableProperties(getDefaultNonBlockingRouterProperties(true, LOCAL_DC));
     routerConfig = new RouterConfig(vprops);
-    mockClusterMap = new MockClusterMap(false, true, 9, 3, 3, false, includeCloudDc);
+    mockClusterMap = new MockClusterMap(false, true, 9, 3, 3, false, includeCloudDc, LOCAL_DC);
     this.includeCloudDc =  includeCloudDc;
     localDcName = mockClusterMap.getDatacenterName(mockClusterMap.getLocalDatacenterId());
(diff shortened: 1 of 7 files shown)


ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -186,55 +202,82 @@ public void after() {
...
 
   /**
    * Instantiate a router, perform a put, close the router. The blob that was put will be saved in the MockServer,
    * and can be queried by the getBlob operations in the test.
    * @param operationTrackerType the type of {@link OperationTracker} to use.
    * @param testEncryption {@code true} if blobs need to be tested w/ encryption. {@code false} otherwise
+   * @param includeCloudDc {@code true} if add cloud backed DC to the cluster map. {@code false} otherwise
    */
-  public GetBlobOperationTest(String operationTrackerType, boolean testEncryption) throws Exception {
+  public GetBlobOperationTest(String operationTrackerType, boolean testEncryption, boolean includeCloudDc) throws Exception {
```
justinlin-linkedin, line 227: I have a mixed feeling about adding cloud test in this file. I think this file is now too bloated to have more tests. Can we move these tests to a different file?

JingQianCloud: Yes, I have the similar feeling. Will make it a standalone Cloud test.

PR #2001, opened 2021-11-22
[BANDWIDTH_THROTTLING] Add methods to QuotaChargeCallback to check if request should be throttled and if quota exceeding is allowed.

ankagrawal:
For bandwidth throttling implementation, we will need to do separate checks in Operation Controller about:

- should the request be throttled for a resource (shouldThrottle)
- should a request be allowed even if quota is breached (quotaExceedAllowed).

Separate checks are needed for these two cases, because number of chunk requests to server, drained from the queue will depend on a quota resource's quota usage, as well as frontend's overall usage.

Currently the implementation of new methods is dummy. Future PRs will fill up the implementation.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
@@ -40,7 +40,7 @@ public interface QuotaChargeCallback {
     RequestCostPolicy requestCostPolicy = new UserQuotaRequestCostPolicy(quotaManager.getQuotaConfig());
     return new QuotaChargeCallback() {
       @Override
-      public void chargeQuota(long chunkSize) throws RouterException {
+      public void charge(long chunkSize) throws RouterException {
         try {
           Map<QuotaName, Double> requestCost = requestCostPolicy.calculateRequestQuotaCharge(restRequest, chunkSize)
               .entrySet()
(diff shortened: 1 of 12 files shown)


ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
```diff
@@ -84,4 +94,16 @@ public void chargeQuota() throws RouterException {
...
   void chargeQuota() throws RouterException;
+
+  /**
+   * Check if request should be throttled based on quota usage.
+   * @return {@code true} if request usage exceeds limit and request should be throttled. {@code false} otherwise.
+   */
+  boolean shouldThrottle();
+
+  /**
+   * Check if usage is allowed to exceed the quota limit.
```
tbingol, line 105: Nit: "limit" is redundant

tbingol (approved):

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
```diff
@@ -68,6 +68,16 @@ public void chargeQuota(long chunkSize) throws RouterException {
       public void chargeQuota() throws RouterException {
         chargeQuota(quotaManager.getQuotaConfig().quotaAccountingUnit);
       }
+
+      @Override
+      public boolean shouldThrottle() {
```
justinlin-linkedin, line 73: I would recommend to use the same method names as the Chargeable here, so we know the relation between the callback and the chargeable by just looking to the interface. We can probably rename the `chargeQuota` to `charge` and the `shouldThrottle` to `check`.

ankagrawal: done

justinlin-linkedin (approved):

PR #2002, opened 2021-11-22
[BANDWIDTH_THROTTLING] Create a Chargeable interface for charging quota on chunk requests.

ankagrawal:
Each chunk (GetChunk/PutChunk) will implement the Chargeable interface to define when a chunk request can be charged. This will also be used in `OperationController` to segregate each chunk request by the `QuotaResource` and to determine if quota resource goes above it quota.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
@@ -14,7 +14,7 @@
 package com.github.ambry.quota;
 
 /**
- * A {@link Chargeable} is a operation that can be charged against quota.
+ * A {@link Chargeable} is an operation that can be charged against quota.
  */
 public interface Chargeable {
 
@@ -26,7 +26,7 @@ public interface Chargeable {
   boolean checkAndCharge();
 
   /**
-   * Check if its allowed for usage to exceed quota.
+   * Check if usage allowed to exceed quota.
    *
    * @return {@code true} if usage is allowed to exceed quota. {@code false} otherwise.
    */
--- a/ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
@@ -1,0 +1,39 @@
+/**
+ * Copyright 2021 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.quota;
+
+/**
+ * A {@link Chargeable} is a operation that can be charged against quota.
+ */
+public interface Chargeable {
+
+  /**
+   * Charge the cost of the operation against relevant quota if usage is within quota.
+   *
+   * @return {@code true} if usage is within quota and quota was charged. {@code false} otherwise.
+   */
+  boolean checkAndCharge();
+
+  /**
+   * Check if its allowed for usage to exceed quota.
+   *
+   * @return {@code true} if usage is allowed to exceed quota. {@code false} otherwise.
+   */
+  boolean quotaExceedAllowed();
+
+  /**
+   * @return the {@link QuotaResource} whose operation is being charged.
+   */
+  QuotaResource getQuotaResource();
+}
--- a/ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
@@ -19,11 +19,18 @@ package com.github.ambry.quota;
 public interface Chargeable {
 
   /**
-   * Charge the cost of the operation against relevant quota if usage is within quota.
+   * Check if the usage is within quota for the quota resource of this operation.
    *
-   * @return {@code true} if usage is within quota and quota was charged. {@code false} otherwise.
+   * @return {@code true} if usage is within quota. {@code false} otherwise.
    */
-  boolean checkAndCharge();
+  boolean check();
+
+  /**
+   * Charge the request cost for this operation against quota of the quota resource of this operation.
+   *
+   * @return {@code true} if quota was charged. {@code false} otherwise.
+   */
+  boolean charge();
 
   /**
    * Check if usage allowed to exceed quota.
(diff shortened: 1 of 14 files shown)


ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
```diff
@@ -0,0 +1,39 @@
...
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.quota;
+
+/**
+ * A {@link Chargeable} is a operation that can be charged against quota.
```
tbingol, line 17: Nit: ... an operation ...

ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
```diff
@@ -0,0 +1,39 @@
...
+
+  /**
+   * Charge the cost of the operation against relevant quota if usage is within quota.
+   *
+   * @return {@code true} if usage is within quota and quota was charged. {@code false} otherwise.
+   */
+  boolean checkAndCharge();
+
+  /**
+   * Check if its allowed for usage to exceed quota.
```
tbingol, line 29: Nit: Repharase it, e.g. "Check if the usage is allowed to exceed the quota"

tbingol (approved):

ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
```diff
@@ -0,0 +1,39 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.quota;
+
+/**
+ * A {@link Chargeable} is a operation that can be charged against quota.
+ */
+public interface Chargeable {
```
justinlin-linkedin, line 19: I am not sure about the nature of this interface and how it will be used in operation controller?

The `getQuotaResource` would return account id or account/container id, which will be used to charge against a particular quota. But `checkAndCharge` doesn't take `QuotaManager` as its parameter, how would this method check and charge?

BTW, shouldn't "quotaExceedAllowed` be a configuration?

ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
ankagrawal: fixed

ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
ankagrawal: done.

ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
ankagrawal: This interface will be used by frontend to server operations -- like PutChunk, GetChunk etc. Basically the `OperationController` will maintain a collection of chunk requests that need to be sent to server for each account(or quotaresource). The `OperationController` will send these requests to server when the requests are within quota.
What the `OperationController` gets from each Operation (like PutOperation or GetOperation) are instances of `RequestInfo` classes. The `Chargeable` will be part of this `RequestInfo` class. This will help to make calls like `requestInfo.getChargeable().checkAndCharge()` to check if this chunk request is within quota and if yes then charge for it. If no, then this chunk will be delayed for processing. Note that charging in this case will not be as simple as just saying QuotaManager.charge(). The reason is that for the same chunk, there will be multiple requests sent to server nodes (due to local replica request, remote replica requests, router parallelism etc). But the charging should happen for each chunk only once. In order to manage all this complexity, each of the chunks (like PutChunk) will implement chargeable, and will keep track when and how to charge for the quota.

ankagrawal: Do note that this interface will be used by OperationController, but they will be implemented by individual chunk classes (like PutChunk).

ankagrawal: The `getQuotaResource` method will just help OperationController maintain a list of server requests for each `QuotaResource`. The `OperationController` doesn't really use it for charging quota. The problem `getQuotaResource` method in this interface solve is to associate the `RequestInfo` (that represents requests to server nodes) with the account or container to which the request belongs. This will help OperationController maintain a map of pending requests for each account or container.

ankagrawal: checkAndCharge doesn't need to take QuotaManager as its input. That part will be handled in the implementation of checkAndCharge inside each chunk object.

justinlin-linkedin: I mean at the end of the day, `checkAndCharge` would need `QuotaManager` to actually charge the quota from the account. So the `QuotaManager` would have to pass to GetOperation and PutOperation.  Right now the quota is charged through a QuotaChargeCallback. Are we going to use the same callback in the `checkAndCharge` implementations?

ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
```diff
@@ -0,0 +1,39 @@
...
+   *
+   * @return {@code true} if usage is allowed to exceed quota. {@code false} otherwise.
+   */
+  boolean quotaExceedAllowed();
+
+  /**
+   * @return the {@link QuotaResource} whose operation is being charged.
+   */
+  QuotaResource getQuotaResource();
+}
```
Arun-LinkedIn, line 39: nit: new line ending

Arun-LinkedIn (approved):

ambry-api/src/main/java/com/github/ambry/quota/Chargeable.java
```diff
@@ -0,0 +1,39 @@
...
+ * A {@link Chargeable} is an operation that can be charged against quota.
+ */
+public interface Chargeable {
+
+  /**
+   * Charge the cost of the operation against relevant quota if usage is within quota.
+   *
+   * @return {@code true} if usage is within quota and quota was charged. {@code false} otherwise.
+   */
+  boolean checkAndCharge();
```
Arun-LinkedIn, line 26: Thank you for describing in detail on how this interface would be used! Just a question. Since chunks can belong to different partition classes, is it possible that chunk belonging to a partition class with more local replicas get more bandwidth (compared to chunk belonging to partition class with lesser local replicas) since they would have more requests sent from frontend to server for same charge.

ankagrawal: Each chunk will be charged just once. So, while frontend can send multiple rest requests to servers, for each chunk charging will happen only once.
In the implementation, each chunk class (like PutChunk) will maintain a state called `AtomicBoolean isCharged`. First time the `charge` method is called, it will also set `isCharged` to true. When more calls are made for the `PutChunk`, it will check if `isCharged` is already true, and if yes, then it will not charge again.

justinlin-linkedin (approved):

PR #2011, opened 2021-12-04
[BANDWIDTH_THROTTLING] Chargeable implementation.

ankagrawal:


ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
```diff
@@ -0,0 +1,95 @@
...
+  @Override
+  public boolean check() {
+    if(quotaChargeCallback == null || isCharged) {
+      return true;
+    }
+    return quotaChargeCallback.check();
+  }
+
+  @Override
+  public boolean charge() {
```
zzmao, line 58: The current implementation is:

Create `QuotaChargeCallback` when submit a request to router. The `QuotaChargeCallback` is cost + action. When the code arrives `OperationQuotaCharger`, action is taken. 

I am thinking if we can simplify this to:
Only build `cost info` when submit a request. When the code arrives `OperationQuotaCharger`, it takes in `cost info` and charge(action).   All the actions(charge, check, etc) are inside OperationQuotaCharger.

zzmao (commented):
One comment. We can chat offline.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1109,6 +1115,54 @@ OperationTracker getOperationTrackerInUse() {
...
+    public boolean check() {
+      if (quotaChargeCallback == null || isCharged) {
+        return true;
+      }
+
+      return quotaChargeCallback.check();
+    }
+
+    @Override
+    public boolean charge() {
```
tbingol, line 1128: As mentioned in the other PR, check & charge should be atomic operations.

tbingol (changes requested):

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
```diff
@@ -0,0 +1,95 @@
...
+import com.github.ambry.quota.Chargeable;
+import com.github.ambry.quota.QuotaChargeCallback;
+import com.github.ambry.quota.QuotaResource;
+import com.github.ambry.rest.RestServiceException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * {@link Chargeable} implementation for cases where quota is charged just once for entire operation.
```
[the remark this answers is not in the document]
ankagrawal: UNDELETE, DELETE, UPDATE_TTL -- updated in comments.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: I want to take this up as a separate PR. The atomicity should actually come from the last layer of implementation which is QuotaSource. Will modify the call stack to reflect that.

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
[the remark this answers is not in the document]
ankagrawal: also GetBlobInfo

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -727,6 +732,53 @@ void poll(RequestRegistrationCallback<GetOperation> requestRegistrationCallback)
...
+      return quotaChargeCallback.check();
+    }
+
+    @Override
+    public boolean charge() {
+      if (quotaChargeCallback == null || isCharged) {
+        return true;
+      }
+      try {
+        quotaChargeCallback.charge(chunkSize);
```
tbingol, line 749: I think all these calls with eventually resolve to AmbryQuotaManager::charge(), however, that routine is not thread-safe.

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
ankagrawal: As discussed offline, since such a change spans multiple PR, we can take this up as a cleanup work after merging this.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
tbingol: Per our chat IRL, you can handle them in a follow up PR before feature makes the prod.

tbingol (approved):

zzmao (approved):
LGTM

PR #2013, opened 2021-12-05
[BANDWIDTH_THROTTLING] QuotaAwareOperationController

ankagrawal:


ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  /**
+   * Drain the request queue based on resource quota only and update the requests to be send to {@code requestsToSend}.
+   * @param requestsToSend a list of {@link RequestInfo} that will contain the requests to be sent out.
+   */
+  private void drainQuotaCompliantRequests(List<RequestInfo> requestsToSend) {
+    for (QuotaResource quotaResource : requestQueue.keySet()) {
+      if (quotaResource.equals(UNKNOWN_QUOTA_RESOURCE)) {
+        // If there are requests for which QuotaResource couldn't be found, this is most likely a bug.
```
zzmao, line 165: Do you want to add some metrcis for this case?

zzmao (commented):
Generally looks good to me. Will look at the other PR for the implementation to get more thoughts.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+        }
+      }
+      if (done) {
+        break;
+      }
+    }
+  }
+
+  /**
+   * Drain the request queue based on resource quota only and update the requests to be send to {@code requestsToSend}.
```
tbingol, line 159: ... to be sent ...

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+   * @param requestsToSend a list of {@link RequestInfo} that will contain the requests to be sent out.
+   */
+  private void drainQuotaCompliantRequests(List<RequestInfo> requestsToSend) {
```
tbingol, line 162: I don't think *drain* is a good name in this case. Maybe *handle* ?

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+        requestsToSend.addAll(requestQueue.get(UNKNOWN_QUOTA_RESOURCE));
+        requestQueue.remove(UNKNOWN_QUOTA_RESOURCE);
+        continue;
+      }
+      while (!requestQueue.get(quotaResource).isEmpty()) {
+        RequestInfo requestInfo = requestQueue.get(quotaResource).getFirst();
+        if (requestInfo.getChargeable().check()) {
+          requestsToSend.add(requestInfo);
+          requestQueue.get(quotaResource).removeFirst();
+          requestInfo.getChargeable().charge();
```
tbingol, line 176: Technically check() & charge() should be atomic. You may want to change your interface to CheckAndCharge() and it charges in a thread-safe manner if check() returns true.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
tbingol: This should be thread-safe ?

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+    getManager.poll(requestsToSend, requestsToDrop);
+    nonBlockingRouter.initiateBackgroundDeletes(getBackGroundDeleteRequests());
+    clearBackGroundDeleteRequests();
+    deleteManager.poll(requestsToSend, requestsToDrop);
+    ttlUpdateManager.poll(requestsToSend, requestsToDrop);
+    undeleteManager.poll(requestsToSend, requestsToDrop);
+  }
+
+  /**
+   * Add the specified {@link List} of {@link RequestInfo}s to the reqeust queue.
```
tbingol, line 113: s/reqeust/request

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+        if (requestInfo.getChargeable().quotaExceedAllowed()) {
+          requestInfo.getChargeable().charge();
+          requestsToSend.add(requestInfo);
+          requestQueue.get(quotaResource).removeFirst();
+          if (requestQueue.get(quotaResource).isEmpty()) {
+            requestQueue.remove(quotaResource);
+          }
+        } else {
+          done = true;
+          break;
```
tbingol, line 149: This is weird. Your exiting on the first request returning false for quotaExceedAllowed() implies that quotaExceedAllowed() is actually not a property of the request but rather a global setting. You should use a static class to access that setting and either completely skip the list or go through the all items.

tbingol (changes requested):

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
zzmao: `handle` sounds good to me.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+   */
+  private void drainRequestQueue(List<RequestInfo> requestsToSend) {
+    drainQuotaCompliantRequests(requestsToSend);
+
+    List<QuotaResource> quotaResources = new ArrayList<>(requestQueue.keySet());
+    Collections.shuffle(quotaResources);
+    boolean done = false;
+    while (!requestQueue.isEmpty()) {
+      for (QuotaResource quotaResource : quotaResources) {
+        RequestInfo requestInfo = requestQueue.get(quotaResource).getFirst();
```
Arun-LinkedIn, line 139: Like we discussed offline, just noting that charged and allowed replica requests of the same chunk could be split  across `send` cycles since we are choosing only one replica request in each iteration. Thanks for planning to optimize it in subsequent PRs.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+   * @param requestInfos {@link List} of {@link RequestInfo} objects.
+   */
+  private void addToRequestQueue(List<RequestInfo> requestInfos) {
+    for (RequestInfo requestInfo : requestInfos) {
+      QuotaResource quotaResource = requestInfo.getChargeable().getQuotaResource();
+      if (quotaResource == null) {
+        quotaResource = UNKNOWN_QUOTA_RESOURCE;
+      }
+      requestQueue.putIfAbsent(quotaResource, new LinkedList<>());
+      requestQueue.get(quotaResource).add(requestInfo);
```
Arun-LinkedIn, line 123: Like we discussed offline, just a note for later changes. Seems like we also have logic in router to drop requests if there is no response for them from server for some time (probably due to closed connections, etc?) . Looks like we may need to account for the time spent by them waiting in request queue.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * {@link OperationController} implementation that dispatches chunk requests to server based on quota.
+ */
+public class QuotaAwareOperationController extends OperationController {
+  private static final Logger logger = LoggerFactory.getLogger(QuotaAwareOperationController.class);
+  // QuotaResource instance to use cases where OperationController gets null QuotaResource.
```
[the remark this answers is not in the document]
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+      addToRequestQueue(newRequestsToSend);
+      drainRequestQueue(requestsToSend);
+    } catch (Exception e) {
+      logger.error("Operation Manager poll received an unexpected error: ", e);
+      routerMetrics.operationManagerPollErrorCount.inc();
+    }
+  }
+
+  /**
+   * Poll for new requests from each of the operation managers, and initiate backgroud deletes.
```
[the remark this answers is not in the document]
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: Makes sense. Have added a ticket for that as a follow up (shared with you separately).

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+      requestQueue.get(quotaResource).add(requestInfo);
+    }
+  }
+
+  /**
+   * Drain the request queue based on quota and system resources and update the requests to be send to {@code requestsToSend}.
+   * @param requestsToSend a list of {@link RequestInfo} that will contain the requests to be sent out.
+   */
+  private void drainRequestQueue(List<RequestInfo> requestsToSend) {
+    drainQuotaCompliantRequests(requestsToSend);
```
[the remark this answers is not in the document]
ankagrawal: Yes thats correct. We will first take up quota compliant requests. After all the quota compliant requests have been taken in, then we check if system has capacity to handle more requests (by calling isQuotaExceedAllowed() which checks if frontend has enough bandwidth left for more requests in the current time window). For the quota exceeded requests, we want to be fair to all accounts, so we take one request for each account at a time.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: shared the ticket you separately to track this work.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: And `handle` it will be!!!
This doesn't need to be thread safe, because OperationController's `poll` logic is always called in a single thread. All of `OperationController`'s logic isn't thread safe and doesn't need to be for this reason.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: I have a separate ticket for metrics and alerting. Will share the jira ticket with you guys separately.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: You are right about this. I will fix this in the next iteration. The atomicity should actually come from the last layer of implementation which is `QuotaSource`. Will modify the call stack to reflect that.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,193 @@
...
+        RequestInfo requestInfo = requestQueue.get(quotaResource).getFirst();
+        if (requestInfo.getChargeable().quotaExceedAllowed()) {
+          requestInfo.getChargeable().charge();
+          requestsToSend.add(requestInfo);
+          requestQueue.get(quotaResource).removeFirst();
+          if (requestQueue.get(quotaResource).isEmpty()) {
+            requestQueue.remove(quotaResource);
+          }
+        } else {
+          done = true;
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,188 @@
...
+/**
+ * {@link OperationController} implementation that dispatches chunk requests to server based on quota.
+ */
+public class QuotaAwareOperationController extends OperationController {
+  private static final Logger logger = LoggerFactory.getLogger(QuotaAwareOperationController.class);
+  // QuotaResource instance to use for cases where OperationController gets null QuotaResource.
+  private static final QuotaResource UNKNOWN_QUOTA_RESOURCE = new QuotaResource("UNKNOWN", QuotaResourceType.ACCOUNT);
+  private final Map<QuotaResource, LinkedList<RequestInfo>> requestQueue = new HashMap<>();
+  private final NonBlockingRouter nonBlockingRouter;
+  private final NonBlockingRouterMetrics routerMetrics;
```
Arun-LinkedIn, line 48: Seems like member variables `nonBlockingRouter` and `routerMetrics` are also defined in base class `OperationController`.  May be reuse them?

ankagrawal: done

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: You are right about this. I realized that quotaExceedAllowed() needs to be per request because for quota purposes there are two types of requests
- read requests that charge RCU
- write requests that charge WCU and Storage quota.
Frontend node accounts differently for WCU capacity and RCU capacity.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -0,0 +1,200 @@
...
+
+
+/**
+ * {@link OperationController} implementation that dispatches chunk requests to server based on quota.
+ */
+public class QuotaAwareOperationController extends OperationController {
+  private static final Logger logger = LoggerFactory.getLogger(QuotaAwareOperationController.class);
+  // QuotaResource instance to use for cases where OperationController gets null QuotaResource.
+  private static final QuotaResource UNKNOWN_QUOTA_RESOURCE = new QuotaResource("UNKNOWN", QuotaResourceType.ACCOUNT);
+  private final Map<QuotaResource, LinkedList<RequestInfo>> readRequestQueue = new HashMap<>();
```
ankagrawal, line 47: @zzmao @tbingol @Arun-LinkedIn Added separate queues for readRequests (Get/GetBlobInfo request that charge RCU) and write requests (Post, TtlUpdate, Delete that charge WCU and Storage quota).
Note that frontend has separate bandwidth capacity for write requests and read requests (full-duplex nic). The decision whether quota exceed is allowed will depend on the type of capacity a request wants to use. Separating out the types here help in making that decision in code efficiently without iterating through the entire queue everytime `pollForRequests` is called.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: Changed it to `poll` instead of `handle`. Let me know if it looks good.

tbingol (approved):

zzmao (approved):

Arun-LinkedIn (approved):

PR #2014, opened 2021-12-05
[BANDWIDTH_THROTTLING] Make OperationController class configurable.

ankagrawal:


ambry-router/src/main/java/com/github/ambry/router/OperationController.java
```diff
@@ -82,7 +82,7 @@
    * @param accountService the {@link AccountService} to use.
    * @throws IOException if the network components could not be created.
    */
-  OperationController(String suffix, String defaultPartitionClass, AccountService accountService,
+  public OperationController(String suffix, String defaultPartitionClass, AccountService accountService,
```
SophieGuo410, line 85: Does it need to be public?

SophieGuo410 (commented):
Overall lgtm.

tbingol (approved):

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
ankagrawal: This needs to be public because otherwise the reflection api used in Utils.obj() method that creates OperationController instance, will not be able to access this constructor.

SophieGuo410 (approved):

Arun-LinkedIn (approved):

Issue #913, opened 2018-04-27
Delete blobs that were successfully uploaded to some replicas but did not reach quorum


cgtz:
# Problem
Currently the router does not clean up blobs where the put to server nodes did not reach the quorum size but one or more blob replicas was created successfully. This blob could eventually be replicated and remain undeleted.

# Potential Solution
Keep track of blob IDs that were uploaded successfully to at least one replica and use the background deleter to clean it up.

PR #2015, opened 2021-12-06
Tbingol/delete slipped puts

tbingol:
Fix Issue #913: Delete blobs that were successfully uploaded to some replicas but failed to satisfy the quorum

PR maintains a set of chunk blob ids for the failed attempts during a put operation and issues delete requests on them upon the conclusion the operation.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/BackgroundDeleteRequest.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/BackgroundDeleteRequest.java
@@ -14,7 +14,6 @@
 
 package com.github.ambry.router;
 
-import com.github.ambry.commons.BlobId;
 import com.github.ambry.store.StoreKey;
 
 
--- a/ambry-router/src/main/java/com/github/ambry/router/BackgroundDeleteRequest.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/BackgroundDeleteRequest.java
@@ -14,6 +14,7 @@
 
 package com.github.ambry.router;
 
+import com.github.ambry.commons.BlobId;
 import com.github.ambry.store.StoreKey;
 
 
@@ -24,6 +25,7 @@ class BackgroundDeleteRequest {
   static final String SERVICE_ID_PREFIX = "ambry-background-delete-";
   private final StoreKey storeKey;
   private final String serviceId;
+  private final BlobId blobId;
 
   /**
    * @param storeKey The {@link StoreKey} to delete.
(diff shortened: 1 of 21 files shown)


ambry-router/src/main/java/com/github/ambry/router/BackgroundDeleteRequest.java
```diff
@@ -24,6 +25,7 @@
   static final String SERVICE_ID_PREFIX = "ambry-background-delete-";
   private final StoreKey storeKey;
   private final String serviceId;
+  private final BlobId blobId;
```
ankagrawal, line 28: BlobId is a StoreKey (it implements the storeKey interface). So why introduce both and set one or the other? Can't storeKey be used in every case?

ambry-router/src/main/java/com/github/ambry/router/RouterCallback.java
```diff
@@ -60,5 +62,17 @@ void scheduleDeletes(List<StoreKey> idsToDelete, String serviceIdSuffix) {
       backgroundDeleteRequests.add(new BackgroundDeleteRequest(storeKey, serviceIdSuffix));
     }
   }
+
+  /**
+   * Schedule the deletes of ids in the given list.
+   * @param idsToDelete the set of blob ids that need to be deleted.
+   * @param serviceIdSuffix the suffix to append to the service ID when deleting these blobs.
+   */
+  void scheduleDeletesForBlobIds(Set<BlobId> idsToDelete, String serviceIdSuffix) {
```
ankagrawal, line 71: maybe we can rename this to scheduleSlippedDeletes()? In fact we might not even need this method if we didn't change the `BackgroundDeleteRequest` class (see comments above).

ankagrawal (approved):
Look good to me overall. Can you please fix the tests. I guess just re-running them might help fix it -- failure don't seem to be related to this PR.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -715,6 +715,87 @@ public void onBlobDeleted(String blobId, String serviceId, Account account, Cont
     Assert.assertEquals("All operations should have completed", 0, router.getOperationsCount());
   }
 
+  /**
+   * Test that even when a composite blob put succeeds, the slipped put data chunks are deleted.
+   */
+  @Test
+  public void testSuccessfulPutDataChunkDelete() throws Exception {
```
cgtz, line 722: We should monitor how this test case performs on CI runs. If it fails intermittently, we may have to make some changes.

I understand why it is probabilistic though.

cgtz (approved):

PR #2016, opened 2021-12-06
Add more logging to PutOperation to debug memory leak

justinlin-linkedin:
Adding more logging information in the PutOperation to debug memory leak.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -161,7 +161,7 @@ class PutOperation {
 
   // Variables to keep track of netty chunks
   private final RestRequest restRequest;
-  private final String touchMessage;
+  private final String loggingContext;
 
   private static final Logger logger = LoggerFactory.getLogger(PutOperation.class);
 
@@ -298,7 +298,7 @@ class PutOperation {
     chunkFillerChannel = new ByteBufferAsyncWritableChannel(writableChannelEventListener);
     isEncryptionEnabled = passedInBlobProperties.isEncrypted();
     restRequest = options.getRestRequest();
-    touchMessage = makeTouchMessage();
+    loggingContext = makeLoggingContext();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -161,7 +161,7 @@
 
   // Variables to keep track of netty chunks
   private final RestRequest restRequest;
-  private final String touchMessage;
+  private final String loggingContext;
```
zzmao, line 164: Looks like you want to print channel information. How about name it to channelContext ?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1771,16 +1781,20 @@ protected PutRequest createPutRequest() {
...
-  private String makeTouchMessage() {
+  private String makeLoggingContext() {
     if (restRequest == null || restRequest.getRestRequestContext() == null) {
       return "";
     }
     String channelStr =
         ((NettyRequest.NettyRequestContext) restRequest.getRestRequestContext()).getChannel().toString();
-    return restRequest.getUri() + " " + channelStr;
+    if (isEncryptionEnabled) {
+      return "True " + channelStr;
```
zzmao, line 1794: What would the return string looks like? 

I feel we can make it more informative and formatted.  

maybe: 
channelStr + " Encrypted: True"

zzmao (approved):
Code LGTM

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
justinlin-linkedin: I also include some other information like encryption. And this is per-putoperation context for logging. I think loggingContext is fine.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
justinlin-linkedin: I added Encrypted to the context.

zzmao (approved):

cgtz (approved):

PR #2018, opened 2021-12-06
Add addl azure auth and key storage options

cgtz:
Provide a way to fetch a Cosmos DB from an azure keyvault protected by
RBAC instead of directly storing the key in configs.

Provide a StorageClient impl that uses the azure-identity library. This
library internally handles token refresh and may be preferable to the
lower level msal4j library after we have verified it.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudConfig.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudConfig.java
@@ -216,14 +216,16 @@ public class AzureCloudConfig {
   public final String azureIdentitySecret;
 
   /**
-   * Azure AAD identity client secret. For use with {@code ClientSecretCredential} auth.
+   * Azure AAD identity proxy host. This is a separate config from other services since there are cases where a proxy
+   * is required only for AAD (since AAD doesn't support private endpoints).
+   * For use with {@code ClientSecretCredential} auth.
    */
   @Config(AZURE_IDENTITY_PROXY_HOST)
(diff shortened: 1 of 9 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudConfig.java
```diff
@@ -169,6 +194,41 @@
...
+   * Azure AAD identity client secret. For use with {@code ClientSecretCredential} auth.
+   */
+  @Config(AZURE_IDENTITY_SECRET)
+  @Default("")
+  public final String azureIdentitySecret;
+
+  /**
+   * Azure AAD identity client secret. For use with {@code ClientSecretCredential} auth.
+   */
+  @Config(AZURE_IDENTITY_PROXY_HOST)
```
ankagrawal, line 221: javadocs.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudConfig.java
```diff
@@ -169,6 +194,41 @@
...
+  @Default("")
+  public final String azureIdentityProxyHost;
+
+  /**
+   * Azure AAD identity client secret. For use with {@code ClientSecretCredential} auth.
```
ankagrawal, line 226: javadocs

ankagrawal (approved):
Minor comments. Looks good to me otherwise.

zzmao (approved):
LGTM

PR #2024, opened 2022-01-06
Add accessControlAllowOrigin field at container level

SophieGuo410:


justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -111,6 +111,10 @@
      * Header to contain the Cookies
      */
     public final static String COOKIE = "Cookie";
+    /**
+     * Access_Control_Allow_Origin.
+     */
+    public static final String ACCESS_CONTROL_ALLOW_ORIGIN = "Access_Control_Allow_Origin";
```
cgtz, line 117: This header should use - instead _ : https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

SophieGuo410: Updated. Thanks for catching that.

cgtz (approved):

PR #2025, opened 2022-01-10
clean up chunks in put manager

justinlin-linkedin:
There is a bug in the PutManager that could cause memory leak.
When the client is uploading a blob and the connection is terminated, we would set the operation to be completed in the callback. However, we only release the ready and completed chunks in the callback method, not the building chunk. We are expecting the chunk filler thread would release the building chunk, however, the chunk filler thread might be sleeping and this operation would be removed by put manager. In this case, the building chunk is never released. 

This PR would fix this issue by releasing all the chunks in PutManager when the operation is removed.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -1082,6 +1082,14 @@ class PutOperation {
     }
 
     /**
+     * True is the data in this chunk is released.
+     * @return
+     */
+    synchronized boolean isDataReleased() {
+      return buf == null;
+    }
+
+    /**
      * @return {@code true} if chunk is a MetadataChunk. {@code false} otherwise. Since this is a regular chunk,
      * false is returned
      */
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -471,7 +471,8 @@ class PutOperation {
   }
 
   /**
-   * Release the blob content of chunks, including building chunk.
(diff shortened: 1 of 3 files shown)


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -470,9 +470,30 @@ private void releaseResource() {
     }
   }
 
+  /**
+   * Release the blob content of chunks, including building chunk.
+   */
+  private synchronized void releaseDataForAllChunks() {
+    for (PutChunk chunk : putChunks) {
+      if (chunk.isBuilding() || chunk.isReady() || chunk.isComplete()) {
```
cgtz, line 478: Do we need to handle the encrypting state as well? Is there harm to just releasing all chunks instead of having this if condition.

justinlin-linkedin: encrypted chunk has its owner thread to release the data. There might be some other race condition there. Let me take another look on the encrypted chunk and add a test case for this code.

andrewchoi5 (approved):

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
cgtz: It does look like the EncryptJob takes in `buf.retainedDuplicate()` though and handles releasing that. Do we also have to release the original `buf`?

justinlin-linkedin: we release the original `buf` in the encryptionCallback method. The first if statement checks if the chunk is data chunk, if it's, we release the `buf` no matter if there is an exception.

cgtz (approved):
lgtm

PR #2027, opened 2022-01-13
Create QuotaException class to deal with quota related exceptions.

ankagrawal:


ambry-router/src/test/java/com/github/ambry/router/OperationQuotaChargerTest.java
```diff
@@ -66,7 +67,10 @@ public void testCharge() throws Exception {
     operationQuotaCharger = new OperationQuotaCharger(quotaChargeCallback, BLOBID, "GetOperation");
 
     // charge should return false if quotaChargeCallback throws exception and isCharged is false.
-    Mockito.doThrow(new RouterException("", RouterErrorCode.TooManyRequests)).when(quotaChargeCallback).charge();
+    Mockito.doThrow(
+        new QuotaException("too many requests", new RouterException("", RouterErrorCode.TooManyRequests), true))
```
SophieGuo410, line 71: Are you planning to implement the logic to use isRetryable since the test set to true without real testing it.

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
```diff
@@ -16,6 +16,7 @@
 import com.github.ambry.commons.BlobId;
 import com.github.ambry.quota.Chargeable;
 import com.github.ambry.quota.QuotaChargeCallback;
+import com.github.ambry.quota.QuotaException;
 import com.github.ambry.quota.QuotaMethod;
 import com.github.ambry.quota.QuotaResource;
 import com.github.ambry.rest.RestServiceException;
```
SophieGuo410, line 22: nit: You might wanna remove the RestServiceException for most of the modified files since you replaced it with QuotaException.

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
```diff
@@ -51,22 +50,23 @@ public void charge(long chunkSize) throws RouterException {
...
+                  new RouterException("RequestQuotaExceeded", RouterErrorCode.TooManyRequests), false);
             } else {
               logger.debug("Quota exceeded for an in progress request.");
             }
           }
         } catch (Exception ex) {
-          if (ex instanceof RouterException && ((RouterException) ex).getErrorCode()
-              .equals(RouterErrorCode.TooManyRequests)) {
+          if (ex.getCause() instanceof RouterException && ((RouterException) ex.getCause()).getErrorCode()
+              .equals(RouterErrorCode.TooManyRequests) && quotaManager.getQuotaMode() == QuotaMode.THROTTLING) {
```
justinlin-linkedin, line 61: seems like if we have router error code being TooManyRequests, the quotaMode should already be THROTTLING, this is redundant.

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
```diff
@@ -51,22 +50,23 @@ public void charge(long chunkSize) throws RouterException {
           if (throttlingRecommendation != null && throttlingRecommendation.shouldThrottle() && shouldThrottle) {
```
justinlin-linkedin, line 50: I keep forgetting what is the purpose of passing another `shouldThrottle` parameter to the callback, why do we need it?

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
```diff
@@ -51,22 +50,23 @@ public void charge(long chunkSize) throws RouterException {
...
             if (quotaManager.getQuotaMode() == QuotaMode.THROTTLING
                 && quotaManager.getQuotaConfig().throttleInProgressRequests) {
```
justinlin-linkedin, line 52: so if the throttleInProgressRequests is false, we are not going to throttle any requests, including ttl update and delete?

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
justinlin-linkedin, line 50: also, I think the precondition to check if we should throttle is getting out of control, we have throttling recommendation, and should throttle parameter, and quota mode and throttleInProgressRequests, at some point, we should figure out a way to simplify all these.

ambry-api/src/main/java/com/github/ambry/quota/QuotaException.java
```diff
@@ -0,0 +1,49 @@
...
+    this.isRetryable = isRetryable;
+  }
+
+  /**
+   * Constructor for {@link QuotaManager}.
+   * @param message the exception message.
+   * @param cause {@link Throwable} object that is the actual cause of the exception.
+   * @param isRetryable {@code true} if the quota operation can be tried again. {@code false} otherwise.
+   */
+  public QuotaException(String message, Throwable cause, boolean isRetryable) {
```
justinlin-linkedin, line 38: why do we have isRetryable? how do we differentiate the case when it's not retryable?

ambry-api/src/main/java/com/github/ambry/quota/QuotaUtils.java
```diff
@@ -49,15 +52,20 @@ public static boolean isRequestResourceQuotaManaged(RestRequest restRequest) {
...
+    try {
+      Account account = RestUtils.getAccountFromArgs(restRequest.getArgs());
+      if (account.getQuotaResourceType() == QuotaResourceType.ACCOUNT) {
+        return QuotaResource.fromAccountId(account.getId());
+      } else {
+        Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
+        return QuotaResource.fromContainerId(account.getId(), container.getId());
+      }
+    } catch (RestServiceException rEx) {
+      LOGGER.error("Could not get quota resource for request.");
```
justinlin-linkedin, line 67: can we add any request specific info in the log message, and the exception as well? The exception will be thrown away after this method.

ambry-router/src/test/java/com/github/ambry/router/OperationQuotaChargerTest.java
ankagrawal: The implementation and test for isRetryable behavior will come in next PRs.

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
justinlin-linkedin: in that case, might as well just don't pass the quota callback.

ambry-api/src/main/java/com/github/ambry/quota/QuotaException.java
justinlin-linkedin: Hold on, what is the quota exception used for? So, we have r/w bandwidth quota, which will slow the operation down, but not rejecting the operations. In this case, we would probably never throw an exception. And we have storage quota, which will reject uploads when it's 100%. And this is not retryable.

Like when we deal with GetOperation, we don't really check storage quota, so it's only about the read bandwidth quota. Therefore, we shouldn't care about if it's retryable or not, since it will only be slowed down. But when we deal with PutOperation, we now have to deal with the results from storage quota and write bandwidth quota. If write bandwidth quota is 100%, we will slow it down. If the storage quota is 100%, we will reject the request, and it's not retryable.

Shouldn't this be the case?

ambry-api/src/main/java/com/github/ambry/quota/QuotaUtils.java
```diff
@@ -49,15 +52,21 @@ public static boolean isRequestResourceQuotaManaged(RestRequest restRequest) {
...
+    try {
+      Account account = RestUtils.getAccountFromArgs(restRequest.getArgs());
+      if (account.getQuotaResourceType() == QuotaResourceType.ACCOUNT) {
+        return QuotaResource.fromAccountId(account.getId());
+      } else {
+        Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
+        return QuotaResource.fromContainerId(account.getId(), container.getId());
+      }
+    } catch (RestServiceException rEx) {
+      LOGGER.error("Could not get quota resource for request: " + RestUtils.convertToStr(restRequest));
```
justinlin-linkedin, line 67: nit: could you also log out the exception?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -749,9 +750,9 @@ public boolean charge() {
       try {
         quotaChargeCallback.charge(chunkSize);
         isCharged = true;
-      } catch (RouterException rEx) {
+      } catch (QuotaException quotaException) {
         logger.warn(String.format("Quota charging failed in GetBlobOperation for blob %s due to %s ", blobId.toString(),
-            rEx.toString()));
+            quotaException.toString()));
```
justinlin-linkedin, line 755: nit: you don't have to use s String.format method here, logger has its own formatting.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -771,10 +772,10 @@ public QuotaResource getQuotaResource() {
       }
       try {
         return quotaChargeCallback.getQuotaResource();
-      } catch (RestServiceException rEx) {
+      } catch (QuotaException quotaException) {
         logger.error(
             String.format("Could create QuotaResource object during GetBlobOperation for the chunk %s due to %s. This should never happen.",
-                blobId.toString(), rEx.toString()));
+                blobId.toString(), quotaException.toString()));
```
justinlin-linkedin, line 778: same above.

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
```diff
@@ -63,9 +63,9 @@ public boolean charge() {
     try {
       quotaChargeCallback.charge();
       isCharged = true;
-    } catch (RouterException rEx) {
+    } catch (QuotaException quotaException) {
       LOGGER.warn(String.format("Quota charging failed in %s for blob %s due to %s ", operationName, blobId.toString(),
-          rEx.toString()));
+          quotaException.toString()));
```
justinlin-linkedin, line 68: same

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
```diff
@@ -85,10 +85,10 @@ public QuotaResource getQuotaResource() {
     }
     try {
       return quotaChargeCallback.getQuotaResource();
-    } catch (RestServiceException rEx) {
+    } catch (QuotaException quotaException) {
       LOGGER.error(String.format(
           "Could create QuotaResource object during %s operation for the chunk %s due to %s. This should never happen.",
-          operationName, blobId.toString(), rEx.toString()));
+          operationName, blobId.toString(), quotaException.toString()));
```
justinlin-linkedin, line 91: +1

ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
ankagrawal: I don't follow whats the argument here? This PR is basically for introducing a uniform QuotaException to use for quota use cases instead of using RestServiceException or RouterException. The above code is not being changed, and will be refactored out in future PRs.

ambry-api/src/main/java/com/github/ambry/quota/QuotaException.java
ankagrawal: Exception is never thrown for expected cases. QuotaException will be used for unexpected scenarios happening in the quota call stack (like some error through quota client, or maybe account or container name not found in the RestRequest).

The retryable is for if the exception (the unexpected scenario) can be resolved or not. For something coming from quota client maybe it can be resolved. But if we have a decision (either throttle or not), then no exception will ever be thrown. At no point we will charge for a request and also throw an exception about it. If a request is charged (or it is determined that it shouldn't be charged), then no exception will ever be thrown.

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
ankagrawal: done.

justinlin-linkedin (approved):

SophieGuo410 (approved):

PR #2028, opened 2022-01-16
Fix a bug in QuotaAwareOperationControllerTest where original put manager needs to be closed when assigning a mock object.

ankagrawal:
This is needed so that all the Router threads are closed (chunkfiller thread in this case) and cleanedup.

### DIFF
--- a/ambry-router/src/test/java/com/github/ambry/router/QuotaAwareOperationControllerTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/QuotaAwareOperationControllerTest.java
@@ -108,9 +108,7 @@ public class QuotaAwareOperationControllerTest {
     quotaAwareOperationController =
         new QuotaAwareOperationController(null, null, null, networkClientFactory, clusterMap, routerConfig, null, null,
             routerMetrics, null, null, null, null, nonBlockingRouter);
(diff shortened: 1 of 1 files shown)


ambry-router/src/test/java/com/github/ambry/router/QuotaAwareOperationControllerTest.java
```diff
@@ -107,6 +108,9 @@ public void setupMocks() throws Exception {
     quotaAwareOperationController =
         new QuotaAwareOperationController(null, null, null, networkClientFactory, clusterMap, routerConfig, null, null,
             routerMetrics, null, null, null, null, nonBlockingRouter);
+    Field privateField = OperationController.class.getDeclaredField("putManager");
+    privateField.setAccessible(true);
+    ((PutManager) privateField.get(quotaAwareOperationController)).close();
```
justinlin-linkedin, line 113: can we just expose PutManage as a package-level accessible object in QuotaAwareOperationController. This is convoluted.

ankagrawal: It was already package accessible. cleaned it up

justinlin-linkedin (approved):

PR #2030, opened 2022-01-23
[BANDWIDTH_THROTTLING] Refactor request cost logic.

ankagrawal:
Ambry calculates request cost for two separate concerns:
1. The cost for a user request that goes towards cost model of Ambry.
2. The quota charge for a user request that goes towards resource quota.

This change separates out the logic for the two concerns. Doesn't change any logic other than refactoring.


ambry-api/src/main/java/com/github/ambry/quota/QuotaChargeCallback.java
```diff
@@ -37,12 +37,12 @@
    */
   static QuotaChargeCallback buildQuotaChargeCallback(RestRequest restRequest, QuotaManager quotaManager,
       boolean shouldThrottle) {
-    RequestCostPolicy requestCostPolicy = new UserQuotaRequestCostPolicy(quotaManager.getQuotaConfig());
+    RequestQuotaCostPolicy requestQuotaCostPolicy = new SimpleRequestQuotaCostPolicy(quotaManager.getQuotaConfig());
```
ankagrawal, line 40: Ambry calculates request cost for two separate concerns:

The SCE cost for a user request that goes towards cost model of Ambry.
The quota charge for a user request that goes towards resource quota.
This change separates out the logic for the two concerns. Doesn't change any logic other than refactoring.

ambry-api/src/main/java/com/github/ambry/quota/SCERequestCostPolicy.java
```diff
@@ -0,0 +1,37 @@
...
+import com.github.ambry.rest.RestRequest;
+import com.github.ambry.rest.RestResponseChannel;
+import java.util.Map;
+
+
+/**
+ * Interface for the policy to calculate cost of a request for determining cost incurred via the SCE events.
+ * This forms the basis of Ambry's cost model for its customers.
+ */
+public interface SCERequestCostPolicy {
```
Arun-LinkedIn, line 26: Minor: Like checked offline, just had a question if SCE is a generic term for the OSS code.

ankagrawal: Good call out. Remove the term SCE, and refactored the interfaces and implementations to appropriate packages and directories.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -162,7 +160,7 @@ public void postProcessRequest(RestRequest restRequest, Callback<Void> callback)
 
   @Override
   public void processRequestCharges(RestRequest restRequest, RestResponseChannel responseChannel, BlobInfo blobInfo) {
-    Map<QuotaName, Double> requestCost = requestCostPolicy.calculateRequestCost(restRequest, responseChannel, blobInfo)
+    Map<QuotaName, Double> requestCost = ambryCostModelPolicy.calculateSCERequestCost(restRequest, responseChannel, blobInfo)
```
justinlin-linkedin, line 163: I assume that at the end of the day, we are going to remove the quota logic in this method and AmbryCostModelPolicy?

ankagrawal: So Ambry's cost calculation for quota purposes, and for cost-model have diverged. We are using different policies to calculate these costs. Thats the premise of this PR -- to separate out these two policies so that they can be extended independently in future, if any extension required.

There are two places in `AmbrySecurityService` where quota check is done currently. One in `postProcessRequest` and the other in `processResponse`(`processResquestCharges`). We can definitely get rid of quota check in `postProcessRequest`.
However the calculations in `processResquestCharges` are needed at the end of a user request, to send back the quota header responses as part of users' request response. Same applies to cost model policy calculations to send back cost headers as part of response. So the quota logic in `processRequestCharges` method should remain in my opinion.

There's an orthogonal question of whether `AmbryCostModelPolicy` needs the concept of `QuotaName` as part of the type of costs it handles. I have tried to remove any reference to `QuotaName` from the general interfaces. At the present moment though, the cost model policy implementation (`SimpleAmbryCostModelPolicy`) does use `QuotaName` and `QuotaUtils`. We can try to separate that out later, or maybe rename `QuotaName` enum such that it can be more generic.

justinlin-linkedin: I agree, the `AmbryCostModelPolicy` shouldn't have quota name since it has nothing to do with quota now.

justinlin-linkedin (approved):
LGTM

Arun-LinkedIn (approved):
Looks good to me.

PR #2031, opened 2022-01-23
[BANDWIDTH_THROTTLING] Remove httpStatus from recommendation coming from QuotaEnforcers.

ankagrawal:
This is a refactor only PR, with no logic changes.
1. Remove httpStatus from recommendation coming from QuotaEnforcers.
2. Change recommendedHttpStatus in ThrottlingRecommendation from int to HttpResponseStatus.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaRecommendation.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaRecommendation.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright 2021 LinkedIn Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -23,7 +23,6 @@ public class QuotaRecommendation {
   private final boolean shouldThrottle;
   private final float usagePercentage;
   private final QuotaName quotaName;
-  private final int recommendedHttpStatus;
   private final long retryAfterMs;
 
   /**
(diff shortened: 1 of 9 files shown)


ambry-api/src/main/java/com/github/ambry/quota/QuotaRecommendation.java
```diff
@@ -23,23 +23,19 @@
...
    * Constructor for {@link QuotaRecommendation}.
    * @param shouldThrottle boolean flag indicating throttling recommendation.
    * @param usagePercentage percentage of resource usage.
    * @param quotaName name of the enforcement that made the recommendation.
-   * @param recommendedHttpStatus recommended http status.
    * @param retryAfterMs time after which request can be retried.
    */
-  public QuotaRecommendation(boolean shouldThrottle, float usagePercentage, QuotaName quotaName,
-      int recommendedHttpStatus, long retryAfterMs) {
+  public QuotaRecommendation(boolean shouldThrottle, float usagePercentage, QuotaName quotaName, long retryAfterMs) {
```
ankagrawal, line 35: This is a refactor only PR with no logic changes.

justinlin-linkedin:
this all look good, but I am just wondering that if we should just return a boolean value from ThrottleRecommendation and let FrontendRestRequestService figure out which http status to send back. It seems like we only throw RestRequestException with a defined RestRequestErrorCode to FrontendRestRequestService. Maybe we should do the same thing for throttler.

This can come after this PR.

justinlin-linkedin (approved):

SophieGuo410 (approved):

PR #2034, opened 2022-01-24
[BANDWIDTH_THROTTLING] Introduce QuotaConfigs for specifying resource quotas and frontend CU capacity as json.

ankagrawal:
This PR does the ground work for implementation of `InMemoryCUQuotaSource` class (which will come as a subsequent PR once this is checked in). The `InMemoryCUQuotaSource` class will be the default Capacity Unit QuotaSource implementation for open source code.

1. This PR introduces `CapacityUnit` class that will be used by InMemoryCUQuotaSource class to encapsulate CU Quotas as well as CU Usage. Instances of CapacityUnit can be used to represent quota as well as usage.
2. This PR introduces QuotaConfigs for specifying resource quotas and frontend CU capacity as json. These configs will be used by `InMemoryCUQuotaSource` to setup quotas for accounts and containers during startup. 
3. This PR also introduces `JsonCUQuotaDataProviderUtil` class with utility methods to deserialize the new quota configs that provide the Capacity Unit quota data.

The `CapacityUnit` class and `JsonCUQuotaDataProviderUtil` will be used during instantiation of `InMemoryCUQuotaSource` class to initialize the QuotaSource implementation with quota values for Accounts and Containers.

### DIFF
--- a/ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
@@ -15,7 +15,7 @@ package com.github.ambry.quota.capacityunit;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.github.ambry.quota.QuotaName;
-import com.google.common.util.concurrent.AtomicDouble;
+import java.util.concurrent.atomic.AtomicLong;
 import org.codehaus.jackson.annotate.JsonIgnore;
 
 
@@ -26,21 +26,21 @@ public class CapacityUnit {
   static final String RCU_FIELD_NAME = "rcu";
   static final String WCU_FIELD_NAME = "wcu";
 
-  private final AtomicDouble rcu;
-  private final AtomicDouble wcu;
+  private final AtomicLong rcu;
+  private final AtomicLong wcu;
 
   /**
    * Constructor for {@link CapacityUnit}.
    */
   public CapacityUnit() {
-    rcu = new AtomicDouble(0);
-    wcu = new AtomicDouble(0);
+    rcu = new AtomicLong(0);
+    wcu = new AtomicLong(0);
   }
 
   @JsonIgnore
   public CapacityUnit(JsonNode jsonNode) {
-    this.rcu = new AtomicDouble(jsonNode.get(RCU_FIELD_NAME).asLong());
-    this.wcu = new AtomicDouble(jsonNode.get(WCU_FIELD_NAME).asLong());
+    this.rcu = new AtomicLong(jsonNode.get(RCU_FIELD_NAME).asLong());
+    this.wcu = new AtomicLong(jsonNode.get(WCU_FIELD_NAME).asLong());
   }
 
   @JsonIgnore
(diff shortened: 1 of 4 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Class encapsulating the Read_Capacity_Unit and Write_Capacity_Unit quotas or usage for resource.
+ */
+public class CapacityUnit {
+  static final String RCU_FIELD_NAME = "rcu";
+  static final String WCU_FIELD_NAME = "wcu";
+
+  private final AtomicDouble rcu;
+  private final AtomicDouble wcu;
```
justinlin-linkedin, line 30: why do we need an AtomicDouble for the RCU and WCU? Why are immutable once they are loaded from configuration file. We can just use a long type here.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
```diff
@@ -0,0 +1,90 @@
...
+   * @return Write Capacity Unit quota value.
+   */
+  public double getWcu() {
+    return wcu.get();
+  }
+
+  /**
+   * Set the Write Capacity Unit quota to the specified value.
+   */
+  public void setWcu(long wcu) {
```
justinlin-linkedin, line 87: I don't think we need setter for RCU and WCU.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
ankagrawal: The instances of CapacityUnit class can represent the quota value or the usage.
There will be a refresh flow in the (future) CUQuotaSource that will periodically refresh the usage of every QuotaResource to 0, to make sure that the usage isn't always monotonically increasing. Since this refresh can happen while requests attempt to read the usage values, we need atomic for thread safety.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
ankagrawal: The instances of CapacityUnit class can represent the quota value or the usage.
There will be a refresh flow in the (future) CUQuotaSource that will periodically refresh the usage of every QuotaResource to 0, to make sure that the usage isn't always monotonically increasing. The setters will be used in the refresh flow for usage.

tbingol (approved):

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
justinlin-linkedin: then let's make it AtomicLong instead of AtomicDouble.

ankagrawal: done

PR #2035, opened 2022-01-24
[BANDWIDTH_THROTTLING] Refactor QuotaChargeCallback implementation into its own file.

ankagrawal:
This PR also introduces a quota config to prepare for the QuotaChargeCallback implementation that will be used for Bandwidth Throttling. The main behavior difference between the current QuotaChargeCallback and the new upcoming implementation will be the fact that the new implementation will not reject requests that exceed their CU quota.

This is a refactor only PR with no logic change.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
@@ -35,14 +35,16 @@ public class QuotaConfig {
   public static final String QUOTA_ACCOUNTING_UNIT = QUOTA_CONFIG_PREFIX + "accounting.unit";
   public static final String RESOURCE_CU_QUOTA_IN_JSON = QUOTA_CONFIG_PREFIX + "resource.cu.quota.in.json";
   public static final String FRONTEND_CU_CAPACITY_IN_JSON = QUOTA_CONFIG_PREFIX + "frontend.cu.capacity.in.json";
+  public static final String BANDWIDTH_THROTTLING_FEATURE_ENABLED =
+      QUOTA_CONFIG_PREFIX + "bandwidth.throttling.feature.enabled";
   public static final String DEFAULT_QUOTA_MANAGER_FACTORY = "com.github.ambry.quota.AmbryQuotaManagerFactory";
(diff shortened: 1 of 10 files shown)


ambry-api/src/main/java/com/github/ambry/config/QuotaConfig.java
```diff
@@ -33,13 +33,14 @@
       QUOTA_CONFIG_PREFIX + "request.enforcer.source.pair.info.json";
   public static final String QUOTA_MANAGER_FACTORY = QUOTA_CONFIG_PREFIX + "manager.factory";
   public static final String QUOTA_ACCOUNTING_UNIT = QUOTA_CONFIG_PREFIX + "accounting.unit";
+  public static final String REJECT_ON_CU_QUOTA_BREACH = QUOTA_CONFIG_PREFIX + "reject.on.cu.quota.breach";
```
ankagrawal, line 36: This is a refactor only PR with no logic change, other than the fact that the instantiation of QuotaChargeCallback is now being controlled by this new config.

ambry-api/src/main/java/com/github/ambry/quota/RejectingQuotaChargeCallback.java
```diff
@@ -0,0 +1,95 @@
...
+import com.github.ambry.router.RouterErrorCode;
+import com.github.ambry.router.RouterException;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+
+/**
+ * A {@link QuotaChargeCallback} implementation that will reject requests that exceed their quota.
+ */
+public class RejectingQuotaChargeCallback implements QuotaChargeCallback {
```
Arun-LinkedIn, line 26: Just wondering if it is okay for this class to be in `ambry-api` module or if it should go in `ambry-quota` module since it is an implementation.

ankagrawal: Good call. Move around a few files to make that happen.

Arun-LinkedIn (approved):
Looks good to me!

justinlin-linkedin (approved):

PR #2036, opened 2022-01-25
Remove old stats format from frontend

justinlin-linkedin:
We currently support stats reports in two formats, this PR would remove the old format and only support new format.


Arun-LinkedIn (approved):
lgtm!

SophieGuo410 (approved):
lgtm.

PR #2037, opened 2022-01-26
[azure] Use correct configs for ClientSecretCredential

cgtz:
Replace usage of the legacy configs when constructing a
ClientSecretCredential.


ankagrawal (approved):
lgtm

PR #2038, opened 2022-01-31
[BANDWIDTH_THROTTLING] Introduce AmbryCUQuotaSource implementation.

ankagrawal:
This PR 
1. Changes QuotaSource interface to introduce new methods. The classes that implement this interface also change.
2. Introduces AmbryCUQuotaSource implementation. This will become the default CU QuotaSource implementation for Ambry. 
3. Introduces some new utility methods in QuotaUtils.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
@@ -13,6 +13,7 @@
  */
 package com.github.ambry.quota;
 
+import com.github.ambry.account.Account;
 import java.util.Collection;
 import java.util.List;
 
@@ -22,16 +23,71 @@ import java.util.List;
  * usage of a resource can be saved.
  */
 public interface QuotaSource {
+
+  /**
+   * Method to initialize the {@link QuotaSource}.
+   * @throws QuotaException in case of any exception.
+   */
+  void init() throws QuotaException;
+
+  /**
+   * @return {@code true} if the {@link QuotaSource} is initialized and ready to be used. {@code false} otherwise.
+   */
+  boolean isReady();
+
   /**
    * Get the {@link Quota} for specified resource and operation.
    * @param quotaResource {@link QuotaResource} object.
    * @param quotaName {@link QuotaName} object.
+   * @throws QuotaException in case of any exception.
    */
-  Quota getQuota(QuotaResource quotaResource, QuotaName quotaName);
+  Quota getQuota(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException;
+
+  /**
+   * Get the percent quota usage for the specified resource and operation.
+   * @param quotaResource {@link QuotaResource} object.
+   * @param quotaName {@link QuotaName} object.
+   * @return usage of quota in percentage.
+   * @throws QuotaException in case of any exception.
+   */
+  float getUsage(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException;
+
+  /**
+   * Charge the specified cost against quota for the specified {@link QuotaResource} and {@link QuotaName}.
+   * @param quotaResource {@link QuotaResource} object.
+   * @param quotaName {@link QuotaName} object.
+   * @param usageCost cost to charge against quota.
+   * @throws QuotaException in case of any exception.
+   */
+  void chargeUsage(QuotaResource quotaResource, QuotaName quotaName, double usageCost) throws QuotaException;
+
+  /**
+   * Get the percent usage of system resources relevant to the {@link QuotaName} specified. The exact definition of
+   * system resource is left for the source and enforcer to determine based on the type of quota. Return {@literal -1}
+   * if system resource usage doesn't make sense for a combination of quota, source or enforcer.
+   * @param quotaName {@link QuotaName} object.
+   * @return usage of relevant system resources in percentage.
+   * @throws QuotaException in case of any exception.
+   */
+  float getSystemResourceUsage(QuotaName quotaName) throws QuotaException;
+
+  /**
+   * Charge the specified cost against system resources relevant to the {@link QuotaName} specified. If system resource
+   * usage doesn't make sense for a combination of quota, source or enforcer, this method would do nothing.
+   * @param quotaName {@link QuotaName} object.
+   * @param usageCost of relevant system resources.
+   * @throws QuotaException in case of any exception.
+   */
+  void chargeSystemResourceUsage(QuotaName quotaName, double usageCost) throws QuotaException;
 
   /**
    * Update the quota for newly created {@link List} of {@link QuotaResource}s.
-   * @param quotaResources {@link List} of new created {@link QuotaResource}s.
+   * @param accounts {@link List} of new created {@link QuotaResource}s.
    */
-  void updateNewQuotaResources(Collection<QuotaResource> quotaResources);
+  void updateNewQuotaResources(Collection<Account> accounts);
+
+  /**
+   * Shutdown the {@link QuotaSource} and perform any cleanup.
+   */
+  void shutdown() throws Exception;
 }
--- a/ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
@@ -96,19 +96,15 @@ public class AmbryCUQuotaSource implements QuotaSource {
     checkSupported(quotaName, quotaResource);
 
     String resourceId = quotaResource.getResourceId();
-    if (cuQuota.containsKey(resourceId)) {
-      return new Quota<>(quotaName, cuQuota.get(resourceId).getQuotaValue(quotaName), quotaResource);
-    }
-    throw new QuotaException(String.format("Couldn't find quota for resource %s", quotaResource), true);
+    assertResourceId(resourceId);
+    return new Quota<>(quotaName, cuQuota.get(resourceId).getQuotaValue(quotaName), quotaResource);
   }
 
   @Override
   public float getUsage(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException {
     checkSupported(quotaName, quotaResource);
     String resourceId = quotaResource.getResourceId();
-    if (!cuQuota.containsKey(resourceId)) {
-      throw new QuotaException(String.format("Couldn't find quota for resource %s", quotaResource), true);
-    }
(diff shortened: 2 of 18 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
```diff
@@ -0,0 +1,205 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A {@link QuotaSource} implementation that keeps the quota and usage values in memory, and treats Ambry frontend's
+ * Read and Write Capacity Units as system resources for handling quota exceeded requests.
+ */
+public class InMemoryCUQuotaSource implements QuotaSource {
```
tbingol, line 44: I am not sure if InMemoryCUQuotaSource is a good name. Usage is calculated locally, in other words, it will always be stateless. I cannot think of any other implementation other than "InMemory". I suggest to remove the InMemory part

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
```diff
@@ -0,0 +1,205 @@
...
+  @Override
+  public boolean isReady() {
+    return isReady.get();
+  }
+
+  @Override
+  public Quota getQuota(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException {
+    checkSupported(quotaName, quotaResource);
+
+    String resourceId = quotaResource.getResourceId();
```
tbingol, line 98: nit: final String ...

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
```diff
@@ -0,0 +1,205 @@
...
+    if (cuQuota.containsKey(resourceId)) {
+      return new Quota<>(quotaName, cuQuota.get(resourceId).getQuotaValue(quotaName), quotaResource);
+    }
+    throw new QuotaException(String.format("Couldn't find quota for resource %s", quotaResource), true);
+  }
+
+  @Override
+  public float getUsage(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException {
+    checkSupported(quotaName, quotaResource);
+    String resourceId = quotaResource.getResourceId();
```
tbingol, line 108: nit: final String ...

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
```diff
@@ -0,0 +1,205 @@
...
+    if (cuUsage.containsKey(resourceId)) {
+      usage = cuUsage.get(resourceId).getQuotaValue(quotaName);
+    }
+    return QuotaUtils.getUsagePercentage(cuQuota.get(resourceId).getQuotaValue(quotaName), usage);
+  }
+
+  @Override
+  public void chargeUsage(QuotaResource quotaResource, QuotaName quotaName, double usageCost) throws QuotaException {
+    checkSupported(quotaName, quotaResource);
+    String resourceId = quotaResource.getResourceId();
```
tbingol, line 122: nit: final String ...

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
```diff
@@ -0,0 +1,205 @@
...
+    if (!cuQuota.containsKey(resourceId)) {
```
tbingol, line 123: nit: This pattern repeats in many functions. You may consider creating a private function for this something like

private void assertResourceId(String resourceId) throws QuotaException {
}

Should we also emit a metric & create an alert for resource ids running without a quota?

tbingol (approved):

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
ankagrawal: refactored the repeated pattern.
Will put out a separate PR for metrics. It possible that same metrics can be emitted at multiple levels (QuotaManager, QuotaSource, QuotaEnforcer or OperationController), and hence it will be easier to reason about metric placement once a good implementation is in shape.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
```diff
@@ -0,0 +1,210 @@
...
+    } else {
+      feUsage.incrementWcu((long) Math.ceil(usageCost));
+    }
+  }
+
+  @Override
+  public void updateNewQuotaResources(Collection<Account> accounts) {
+    Collection<QuotaResource> quotaResources = QuotaUtils.getQuotaResourcesFromAccounts(accounts);
+    quotaResources.forEach(quotaResource -> {
+      cuQuota.putIfAbsent(quotaResource.getResourceId(),
```
cgtz, line 145: Is `cuQuota` written/read from multiple threads? You may need a ConcurrentHashMap to prevent things like ConcurrentModificationException during iterating and, if there are multiple writers, corrupted internal state.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
ankagrawal: fixed it.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
ankagrawal: fixed.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/InMemoryCUQuotaSource.java
ankagrawal: fixed.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
ankagrawal: Fixed the concurrency issue. Also made feUsage atomic to avoid corruption.

ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
```diff
@@ -22,16 +23,71 @@
...
+  Quota getQuota(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException;
+
+  /**
+   * Get the percent quota usage for the specified resource and operation.
+   * @param quotaResource {@link QuotaResource} object.
+   * @param quotaName {@link QuotaName} object.
+   * @return usage of quota in percentage.
+   * @throws QuotaException in case of any exception.
+   */
+  float getUsage(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException;
```
cgtz, line 53: is there a reason that chargeUsage takes in doubles and getUsage returns float as opposed to using the same precision everywhere?

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
```diff
@@ -0,0 +1,215 @@
...
+    return QuotaUtils.getUsagePercentage(cuQuota.get(resourceId).getQuotaValue(quotaName), usage);
+  }
+
+  @Override
+  public void chargeUsage(QuotaResource quotaResource, QuotaName quotaName, double usageCost) throws QuotaException {
+    checkSupported(quotaName, quotaResource);
+    final String resourceId = quotaResource.getResourceId();
+    assertResourceId(resourceId);
+    if (quotaName == QuotaName.READ_CAPACITY_UNIT) {
+      cuUsage.get(resourceId).incrementRcu((long) Math.ceil(usageCost));
```
cgtz, line 125: There is a case where assertResourceId could return successfully but you get an NPE from cuUsage if updateNewQuotaResources is still running. Maybe you can protect against this with a similar existence check to getUsage.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  @Override
+  public float getUsage(QuotaResource quotaResource, QuotaName quotaName) throws QuotaException {
+    checkSupported(quotaName, quotaResource);
+    final String resourceId = quotaResource.getResourceId();
+    assertResourceId(resourceId);
+    double usage = 0;
+    if (cuUsage.containsKey(resourceId)) {
+      usage = cuUsage.get(resourceId).getQuotaValue(quotaName);
```
cgtz, line 114: This is a nit but, you can avoid looking up a value in a hashmap twice by doing:
```
CapacityUnit cu = cuUsage.get(resourceId);
double usage = cu != null? cu.getQuotaValue(quotaName) : 0;
```
(also could use optionals but we don't widely use them in our codebase)

cgtz (approved):
lgtm, just a couple more comments

ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
ankagrawal: So charge taken in double because the interface doesn't want to make assumptions about how large a cost could be. `getUsage` returns float, because it returns percentage usage.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
ankagrawal: fixed.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaSource.java
ankagrawal: fixed.

PR #2039, opened 2022-01-31
Move storageClient APIs in ambry-cloud module to async

Arun-LinkedIn:


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -90,35 +92,34 @@ public AzureBlobDataAccessor(CloudConfig cloudConfig, AzureCloudConfig azureClou
...
+          blobBatchAsyncClient, azureMetrics, blobLayoutStrategy, azureCloudConfig);
     } catch (ReflectiveOperationException roEx) {
       throw new IllegalArgumentException("Unable to instantiate storage client: " + roEx.getMessage(), roEx);
     }
     this.azureMetrics = azureMetrics;
     this.purgeBatchSize = AzureCloudConfig.DEFAULT_PURGE_BATCH_SIZE;
-    requestTimeout = null;
-    uploadTimeout = null;
-    batchTimeout = null;
+    requestTimeout = Duration.ofMillis(CloudConfig.DEFAULT_CLOUD_REQUEST_TIMEOUT);
```
ankagrawal, line 112: Same comment for these three lines -- we should use configs like `CloudConfig.cloudRequestTimeout`

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -90,35 +92,34 @@ public AzureBlobDataAccessor(CloudConfig cloudConfig, AzureCloudConfig azureClou
...
-   * @param blobBatchClient the {@link BlobBatchClient} to use.
+   * @param blobServiceAsyncClient the {@link BlobServiceAsyncClient} to use.
+   * @param blobBatchAsyncClient the {@link BlobBatchAsyncClient} to use.
    * @param clusterName the cluster name to use.
    * @param azureMetrics the {@link AzureMetrics} to use.
    * @param azureCloudConfig {@link AzureCloudConfig} object.
    */
-  AzureBlobDataAccessor(BlobServiceClient blobServiceClient, BlobBatchClient blobBatchClient, String clusterName,
-      AzureMetrics azureMetrics, AzureCloudConfig azureCloudConfig) {
+  AzureBlobDataAccessor(BlobServiceAsyncClient blobServiceAsyncClient, BlobBatchAsyncClient blobBatchAsyncClient,
```
ankagrawal, line 101: So we are replacing sync implementations with async implementations here. Would it be better to allow a config (maybe an existing one) to choose the implementation between sync or async?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -90,35 +92,34 @@ public AzureBlobDataAccessor(CloudConfig cloudConfig, AzureCloudConfig azureClou
...
+    batchTimeout = Duration.ofMillis(CloudConfig.DEFAULT_BATCH_TIMEOUT);
   }
 
   /**
    * Visible for testing.
    * @return the underlying {@link BlobServiceClient}.
    */
-  public BlobServiceClient getStorageClient() {
-    return storageClient.getStorageClient();
+  public BlobServiceAsyncClient getStorageAsyncClient() {
```
ankagrawal, line 121: Same here. Maybe we just leave `getStorageClient()` as is, and it will return either async client or sync client based on some config.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -145,27 +146,33 @@ ProxyOptions getProxyOptions() {
...
+          return false;
+        } else {
+          throw blobStorageException;
+        }
+      } else if (exception instanceof UncheckedIOException) {
+        // error processing input stream
+        throw ((UncheckedIOException) exception).getCause();
       } else {
-        throw e;
+        throw exception;
```
ankagrawal, line 175: I think `Exception` is too generic to throw from here.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
```diff
@@ -72,8 +68,6 @@
  */
 public abstract class StorageClient {
   Logger logger = LoggerFactory.getLogger(StorageClient.class);
-  private AtomicReference<BlobServiceClient> storageClientRef;
```
ankagrawal, line 75: Overall, a better abstraction would be that StorageClient implements methods like `getBlockBlobClient`, `getStorageClient` etc, and hides details about whether these are asyn implementation of sync implementation. It could then define a static method like `buildAsyncStorageClient` or `buildSyncStorageClient` based on a config. All the code everywhere else shouldn't worry about the StorageClient implementation details (whether this is sync implementation or async implemnentation). That way you can swap out any implementation with a new implementation with minimal code changes. It keeps the code extensible.

ambry-commons/src/main/java/com/github/ambry/commons/FutureUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+import static com.github.ambry.utils.Utils.*;
+
+
+/**
+ * A collection of utilities that expand the usage of {@link CompletableFuture}. Inspired from
+ * https://jar-download.com/artifacts/org.apache.flink/flink-runtime_2.11/1.7.0/source-code/org/apache/flink/runtime/concurrent/FutureUtils.java
```
ankagrawal, line 27: I wonder if such a reference is allowed as per our license.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
Arun-LinkedIn: Sure. Thank you for the suggestion. Yeah, like you said, it helps for us to have both sync and async methods. But, since the rest of code like users of StorageClient class is also moving to async, I was just wondering if we would still need sync methods. Also, it seems like the sync azure APIs also use async APIs underneath. Will sync up more with you offline.

ankagrawal: My main concern here is that a lot of stuff that should be ideally abstracted by StorageClient interface is being leaked out to callers. For example in `AzureBlobDataAccessor::uploadIfNotExists` , the role of `storageClient` should be to upload a blob and maybe return some response. So we should expect a call like `storageClient.uploadWithResponse`. Whether the upload is sync or async, and what goes into making that upload happen shouldn't leak out and be totally handled out by `storageClient`. The contract of `StorageClient` should be -- I can upload a blob to a storage backend (which happens to be azure in our case), I can donwload a blob from a storage backend (which happens to be azure in our case), I can delete a file, I can delete a batch,  I can get some properties for a blob etc. How its does it, is its own business, and callers shouldn't worry about that.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -180,14 +187,17 @@ public boolean uploadIfNotExists(BlobId blobId, long inputLength, CloudBlobMetad
...
-      throws BlobStorageException, IOException {
+  public void uploadFile(String containerName, String fileName, InputStream inputStream) throws Exception {
     try {
-      storageClient.uploadWithResponse(containerName, fileName, true, inputStream, inputStream.available(), null, null,
-          null, null, defaultRequestConditions, uploadTimeout);
-    } catch (UncheckedIOException e) {
-      // error processing input stream
-      throw e.getCause();
+      storageClient.uploadAsyncWithResponse(containerName, fileName, true, inputStream, inputStream.available(), null,
+          null, null, null, defaultRequestConditions).get(uploadTimeout.toMillis(), TimeUnit.MILLISECONDS);
```
cgtz, line 193: Does the async client have a way to pass a timeout in as an argument? Sometimes, waiting on a future with a timeout can leave resources still in use in the background as opposed to a "deep" timeout.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -268,23 +290,28 @@ public void downloadBlob(BlobId blobId, OutputStream outputStream) throws BlobSt
    * @return The {@link CloudBlobMetadata} if the blob was found, or null otherwise.
    * @throws BlobStorageException
    */
-  public CloudBlobMetadata getBlobMetadata(BlobId blobId) throws BlobStorageException {
-    BlobProperties blobProperties = null;
+  public CloudBlobMetadata getBlobMetadata(BlobId blobId) throws Exception {
```
cgtz, line 293: Are there many callers that expect these methods to throw BlobStorageException?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
Arun-LinkedIn: Yeah, since we would have moving the callers of StorageClient to be async as well, we are not having any cases where sync methods are being used. Hence removed them. In case there is a use case where sync functionality is needed, may be we can have callers of async client wait on the response?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
Arun-LinkedIn: Thank you for the explanation. Yeah, it is a good design for abstraction of classes. Like we discussed offline, in this case though we might not need sync implementations of storage client since rest of code is also going to refactored to be asynchronous. For example, we would be changing `AzureBlobDataAccessor::uploadIfNotExists` also to async (i.e. have return type as future which returns immediately and not wait for response) and hence only use storage client async methods.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
Arun-LinkedIn: It looks like we are currently checking for instance types of this exception when converting these exceptions to custom `CloudStorageException`.  I was wondering if we could probably handle more granular types there?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
Arun-LinkedIn: I tried checking but didn't find a way to pass timeout argument to async client. I was thinking that I might have to remove waiting on the response here in subsequent PRs since the `AzureBlobDataAccessor` APIs also would return `future`. But in that case too, I was thinking of completing the future exceptionally after a timeout which might leave resources in background. Just wondering if it is possible to clean resources when using `CompletableFutures` since they could be completed from different places.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
Arun-LinkedIn: Sorry, yeah may be it is too generic to catch and throw exceptions of type `Exception` class since it might also mask any checked exceptions (if they are added later by underneath layers). Modified it to throw subclass exceptions like `BlobStorageException`.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
Arun-LinkedIn: It looks like callers of these methods are catching all exceptions and wrapping it in `CloudStorageException` in `toCloudStorageException()` method. But, modified it back to throw sub-class exceptions like `BlobStorageException`, `IOException`.

ambry-commons/src/main/java/com/github/ambry/commons/FutureUtils.java
```diff
@@ -0,0 +1,95 @@
...
+        if (throwable != null) {
+          if (throwable instanceof CancellationException) {
+            resultFuture.completeExceptionally(throwable);
+          } else {
+            throwable = extractFutureExceptionCause(throwable);
+            if (!retryPredicate.test(throwable)) {
+              resultFuture.completeExceptionally(throwable);
+            } else if (retries > 0) {
+              retryErrorConsumer.accept(throwable);
+              retryOperation(resultFuture, operation, retries - 1, retryPredicate, retryErrorConsumer);
```
cgtz, line 82: It seems like this doesn't do any backoff. Is that okay for this use case?

cgtz (approved):
sorry for the long delay in final review

ambry-commons/src/main/java/com/github/ambry/commons/FutureUtils.java
Arun-LinkedIn: I was wondering if it should be okay since this is per our current behavior. This is for use case in `ADAuthBasedStorageClient` where we are refreshing our token and retrying blob operations immediately when we receive authorization failure due to token expiry. 

May be I can expand the utility method to take in backoff duration and retry accordingly when we use it for other use cases.

cgtz: This makes sense. ADAuthBasedStorageClient may be retired soon anyways.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
ankagrawal: nit: why not rename it to getStorageClient()

ambry-commons/src/main/java/com/github/ambry/commons/FutureUtils.java
```diff
@@ -0,0 +1,95 @@
...
+public class FutureUtils {
+
+  /**
+   * Retry the given operation the given number of times in case of a failure.
+   *
+   * @param <T> type of the result
+   * @param operation to executed
+   * @param retries if the operation failed
+   * @param retryPredicate Predicate to test whether an exception is retryable
+   * @param retryErrorConsumer
```
ankagrawal, line 39: nit: no description for some of the params?

ankagrawal (approved):

PR #2042, opened 2022-02-01
Add support for float configs in VerifiableProperties.

ankagrawal:
This PR add support for method `getGloatInRange` in VerifiableProperties. This will be used in future PRs (specially by QuotaConfig).

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/VerifiableProperties.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/VerifiableProperties.java
@@ -147,6 +147,24 @@ public class VerifiableProperties {
   }
 
   /**
+   * Read a float from the properties instance. Throw an exception if the value is not in the given range (inclusive).
+   * @param name The property name
+   * @param defaultVal The default value to use if the property is not found
+   * @param start The start of the range in which the value must fall (inclusive)
+   * @param end The end of the range in which the value must fall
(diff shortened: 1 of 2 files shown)


ambry-api/src/main/java/com/github/ambry/config/VerifiableProperties.java
```diff
@@ -146,6 +146,24 @@ public Double getDoubleInRange(String name, Double defaultVal, Double start, Dou
...
+  /**
+   * Read a float from the properties instance. Throw an exception if the value is not in the given range (inclusive).
+   * @param name The property name
+   * @param defaultVal The default value to use if the property is not found
+   * @param start The start of the range in which the value must fall (inclusive)
+   * @param end The end of the range in which the value must fall
+   * @return the {@link Float} value
+   */
+  public Float getFloatInRange(String name, Float defaultVal, Float start, Float end) {
+    Float v = containsKey(name) ? Float.valueOf(Float.parseFloat(getProperty(name))) : defaultVal;
```
cgtz, line 158: For the future: `Float.valueOf(String)` can be used here instead, but it looks like this more verbose way has been used in other places in this class. We can refactor this later. There is probably a lot of opportunity to reuse code better in VerifiableProperties, anyways

cgtz (approved):

PR #2045, opened 2022-02-02
[BANDWIDTH_THROTTLING] Introduce AmbryCUQuotaEnforcer which will act as the default QuotaEnforcer implementation for Capacity Unit quota.

ankagrawal:
This PR
1. Introduces AmbryCUQuotaEnforcer which will act as the default QuotaEnforcer implementation for Capacity Unit quota.
2. Introduces a quota config representing threshold of CU usage percentage of frontend to allow requests to exceed quota.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
@@ -15,13 +15,19 @@ package com.github.ambry.quota;
 
 import com.github.ambry.account.Account;
 import com.github.ambry.account.Container;
+import com.github.ambry.rest.RestRequest;
+import com.github.ambry.rest.RestServiceException;
+import com.github.ambry.rest.RestUtils;
 import java.util.Objects;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * Resource for which quota is specified for enforced.
  */
 public class QuotaResource {
+  private static final Logger LOGGER = LoggerFactory.getLogger(QuotaResource.class);
   public final static String DELIM = "_";
   private final String resourceId; // unique identifier for Ambry account, container or host.
   private final QuotaResourceType quotaResourceType;
@@ -73,6 +79,28 @@ public class QuotaResource {
    */
   public static QuotaResource fromAccountId(short accountId) {
     return new QuotaResource(Integer.toString(accountId), QuotaResourceType.ACCOUNT);
+  }
+
+  /**
+   * Create {@link QuotaResource} for the specified {@link RestRequest}.
+   * @param restRequest {@link RestRequest} object.
+   * @return QuotaResource extracted from headers of {@link RestRequest}.
+   * @throws QuotaException if appropriate headers aren't found in the {@link RestRequest}.
+   */
+  public static QuotaResource fromRestRequest(RestRequest restRequest) throws QuotaException {
+    try {
+      final Account account = RestUtils.getAccountFromArgs(restRequest.getArgs());
+      if (account.getQuotaResourceType() == QuotaResourceType.ACCOUNT) {
+        return QuotaResource.fromAccountId(account.getId());
+      }
+      Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
+      return QuotaResource.fromContainerId(account.getId(), container.getId());
+    } catch (RestServiceException rEx) {
+      LOGGER.error("Could not get quota resource for request: {} due to {}", RestUtils.convertToStr(restRequest),
+          rEx.getMessage());
+      throw new QuotaException("Could not get quota resource for request: " + RestUtils.convertToStr(restRequest),
+          false);
+    }
   }
 
   /**
--- a/ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
@@ -14,13 +14,16 @@
 package com.github.ambry.quota;
 
 import com.github.ambry.account.Account;
+import com.github.ambry.rest.RestRequest;
 import java.util.Collection;
 import java.util.List;
 
 
 /**
- * Interface representing the backend source from which quota for a resource can be fetched, and to which the current
- * usage of a resource can be saved.
+ * QuotaSource is the source of truth of quota for resources. It also tracks the usage of those resources. Resources could
+ * be classified into two types: Ambry's Accounts, Containers or Services as represented by {@link QuotaResource}, or system
+ * resources like a node's bandwidth.
+ * Some implementations might not need to track any system resource. Check {@link QuotaEnforcer} for more details.
  */
 public interface QuotaSource {
 
@@ -63,8 +66,8 @@ public interface QuotaSource {
 
   /**
    * Get the percent usage of system resources relevant to the {@link QuotaName} specified. The exact definition of
-   * system resource is left for the source and enforcer to determine based on the type of quota. Return {@literal -1}
-   * if system resource usage doesn't make sense for a combination of quota, source or enforcer.
+   * system resource is left for the implementation to determine.
+   * Return {@literal -1} if the implementation doesn't need to track any system resource.
    * @param quotaName {@link QuotaName} object.
    * @return usage of relevant system resources in percentage.
    * @throws QuotaException in case of any exception.
@@ -72,8 +75,7 @@ public interface QuotaSource {
   float getSystemResourceUsage(QuotaName quotaName) throws QuotaException;
 
   /**
-   * Charge the specified cost against system resources relevant to the {@link QuotaName} specified. If system resource
-   * usage doesn't make sense for a combination of quota, source or enforcer, this method would do nothing.
+   * Charge the specified cost against system resources relevant to the {@link QuotaName} specified.
    * @param quotaName {@link QuotaName} object.
    * @param usageCost of relevant system resources.
    * @throws QuotaException in case of any exception.
--- a/ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
@@ -22,6 +22,7 @@ import com.github.ambry.config.QuotaConfig;
 import com.github.ambry.messageformat.BlobInfo;
 import com.github.ambry.rest.RestRequest;
 import com.github.ambry.utils.Utils;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
@@ -30,7 +31,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
-import java.util.stream.Collectors;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.slf4j.Logger;
@@ -57,9 +57,9 @@ public class AmbryQuotaManager implements QuotaManager {
    * @param metricRegistry {@link MetricRegistry} object for creating quota metrics.
    * @throws ReflectiveOperationException in case of any exception.
    */
-  public AmbryQuotaManager(QuotaConfig quotaConfig,
-      QuotaRecommendationMergePolicy quotaRecommendationMergePolicy, AccountService accountService,
-      AccountStatsStore accountStatsStore, MetricRegistry metricRegistry) throws ReflectiveOperationException {
+  public AmbryQuotaManager(QuotaConfig quotaConfig, QuotaRecommendationMergePolicy quotaRecommendationMergePolicy,
+      AccountService accountService, AccountStatsStore accountStatsStore, MetricRegistry metricRegistry)
+      throws ReflectiveOperationException {
     Map<String, String> quotaEnforcerSourceMap =
         parseQuotaEnforcerAndSourceInfo(quotaConfig.requestQuotaEnforcerSourcePairInfoJson);
     Map<String, QuotaSource> quotaSourceObjectMap =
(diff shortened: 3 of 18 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -105,10 +104,15 @@ public ThrottlingRecommendation getThrottleRecommendation(RestRequest restReques
...
+      List<QuotaRecommendation> quotaRecommendations = new ArrayList<>();
+      for (QuotaEnforcer quotaEnforcer : requestQuotaEnforcers) {
+        try {
+          quotaRecommendations.add(quotaEnforcer.recommend(restRequest));
+        } catch (QuotaException quotaException) {
+          logger.warn("Could not get recommendation for quota {} due to exception: {}",
+              quotaEnforcer.supportedQuotaNames(), quotaException.getMessage());
+        }
+      }
       if (quotaRecommendations.size() == 0) {
```
tbingol, line 116: nit:s/quotaRecommendations.size() == 0/quotaRecommendations.isEmpty()

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
```diff
@@ -0,0 +1,137 @@
...
+          QuotaUtils.requestDetailsToString(restRequest, blobInfo), qEx.getMessage()), qEx.isRetryable());
+    } catch (Exception ex) {
+      throw new QuotaException(String.format("Could not charge for request %s due to unexpected exception %s",
+          QuotaUtils.requestDetailsToString(restRequest, blobInfo), ex.getMessage()), true);
+    }
+  }
+
+  @Override
+  public QuotaRecommendation recommend(RestRequest restRequest) throws QuotaException {
+    try {
```
tbingol, line 83: I think whatever checks you are making in charge(), you should also make them here. Otherwise, you may run into exceptions in charge() after passing the recommend(). That's bad if you consistently fail during the second enforcer as you will keep charging for the first one.

tbingol (changes requested):

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
ankagrawal: We should have a metric here to set up alerts on, so that we catch the situation of over charging system resource quickly.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
```diff
@@ -0,0 +1,141 @@
...
+   * @throws QuotaException in case action throws an exception.
+   */
+  private <T> T doAndHandleException(Callable<T> action, String errorMessagePrefix) throws QuotaException {
+    try {
+      return action.call();
+    } catch (Exception ex) {
+      if (ex instanceof QuotaException) {
+        throw new QuotaException(String.format("%s %s", errorMessagePrefix, ex.getMessage()),
+            ((QuotaException) ex).isRetryable());
+      }
```
justinlin-linkedin, line 137: this seems unnecessary.  we can just throw this QuotaException.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCapacityUnitQuotaEnforcer.java
```diff
@@ -15,19 +15,23 @@
...
+import java.util.List;
 import java.util.Map;
 
 
 /**
  * Implementation of {@link QuotaEnforcer} for Capacity Units of ambry resource.
  */
-// TODO: The current implementation allows all requests without any processing. It needs to be replaced by an implementation that enforces quota.
 public class AmbryCapacityUnitQuotaEnforcer implements QuotaEnforcer {
+  private final static List<QuotaName> SUPPORTED_QUOTA_NAMES =
```
justinlin-linkedin, line 33: we can just get rid of this implementation since all it's funcationality will be covered by the new implementation.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
```diff
@@ -0,0 +1,141 @@
...
+    this.throttleRetryAfterMs = quotaConfig.cuQuotaAggregationWindowInSecs;
+    this.maxFrontendCuUsageToAllowExceed = quotaConfig.maxFrontendCuUsageToAllowExceed;
+  }
+
+  @Override
+  public void init() {
+  }
+
+  @Override
+  public void charge(RestRequest restRequest, BlobInfo blobInfo, Map<QuotaName, Double> requestCostMap)
```
justinlin-linkedin, line 61: nit: i think we might be able to get rid of BlobInfo in the function signature, since it's not really used anywhere.

ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -128,15 +133,22 @@ public ThrottlingRecommendation charge(RestRequest restRequest, BlobInfo blobInf
...
     Timer.Context timer = quotaMetrics.quotaChargeTime.time();
     try {
-      throttlingRecommendation = quotaRecommendationMergePolicy.mergeEnforcementRecommendations(requestQuotaEnforcers.stream()
-          .map(quotaEnforcer -> quotaEnforcer.chargeAndRecommend(restRequest, blobInfo, requestCostMap))
-          .filter(Objects::nonNull)
-          .collect(Collectors.toList()));
+      for (QuotaEnforcer quotaEnforcer : requestQuotaEnforcers) {
+        try {
+          quotaEnforcer.charge(restRequest, blobInfo, requestCostMap);
+          quotaRecommendations.add(quotaEnforcer.recommend(restRequest));
```
justinlin-linkedin, line 142: hmm, we should probably reserve chargeAndRecommend method since it should be an atomic operation in this case.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
justinlin-linkedin: wait, I am talking about this if statement. when the exception is an instance of QuotaException, we don't have to convert it again to QuotaException, we can just throw it out. unless what you are trying to write in if condition is `!(ex instanceof QuotaException)`

justinlin-linkedin (approved):
Please fix that unnecessary type cast.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
ankagrawal: I want to add more context to the exception's message. At some point (most likely in `Chargeable` implementation), we will log the QuotaException message as error logs when needed. I want that log to contain enough information to that it will be easy for us to understand what happened through the chain of calls.

ambry-quota/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
```diff
@@ -47,6 +49,8 @@
  * An {@link QuotaEnforcer} implementation on each container's storage usage.
  */
 public class StorageQuotaEnforcer implements QuotaEnforcer {
+  private static final List<QuotaName> SUPPORTED_QUOTA_NAMES =
+      Collections.unmodifiableList(Arrays.asList(QuotaName.STORAGE_IN_GB));
   private static final Logger logger = LoggerFactory.getLogger(StorageQuotaEnforcer.class);
   private static final String STORAGE_QUOTA_SERVICE_PREFIX = "storage-quota-enforcer";
   private static final int HTTP_STATUS_THROTTLE = 429;
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-quota/src/main/java/com/github/ambry/quota/storage/StorageQuotaEnforcer.java
```diff
@@ -155,6 +158,16 @@ private QuotaRecommendation recommendBasedOnQuotaAndUsage(Pair<Long, Long> pair)
     return new QuotaRecommendation(shouldThrottle, usagePercentage, QuotaName.STORAGE_IN_GB, NO_RETRY);
   }
 
+  @Override
+  public boolean isQuotaExceedAllowed(RestRequest restRequest) throws QuotaException {
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * Build the {@link QuotaRecommendation} object from the specified usage and {@link QuotaName}.
+   * @param usage percentage usage.
+   * @param quotaName {@link QuotaName} object.
+   * @return QuotaRecommendation object.
+   */
+  private QuotaRecommendation buildQuotaRecommendation(float usage, QuotaName quotaName) {
+    boolean shouldThrottle = (usage >= 100);
```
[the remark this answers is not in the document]
ankagrawal: I think we should always recommend throttling only when customer has used up their quota. I don't see any use case where we would want to throttle someone before they used up quota.
We should definitely warn customers that their quota is going to be over soon. We have logic in `SimpleQuotaRecommendationMergePolicy` to generate that warning, and that will work irrespective of how we recommend `shouldThrottle` here.

I do agree with you that it will be better to add configs to define the thresholds at which we issue warning. Currently it is hardcoded in `SimpleQuotaRecommendationMergePolicy`. I will push out an RB to make those values configurable.

justinlin-linkedin (approved):

SophieGuo410 (approved):

ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
```diff
@@ -63,8 +65,8 @@
 
   /**
    * Get the percent usage of system resources relevant to the {@link QuotaName} specified. The exact definition of
-   * system resource is left for the source and enforcer to determine based on the type of quota. Return {@literal -1}
-   * if system resource usage doesn't make sense for a combination of quota, source or enforcer.
+   * system resource is left for the implementation to determine.
+   * Return {@literal -1} if system resource usage doesn't make sense.
```
tbingol, line 69: What do you mean by "if system resource usage doesn't make sense" ? What system resources do not make sense and why?

Same comment repeats in mant places

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcerFactory.java
```diff
@@ -0,0 +1,47 @@
...
+import com.github.ambry.config.QuotaConfig;
+import com.github.ambry.quota.QuotaEnforcer;
+import com.github.ambry.quota.QuotaEnforcerFactory;
+import com.github.ambry.quota.QuotaSource;
+import com.github.ambry.quota.QuotaSourceFactory;
+import java.io.IOException;
+
+
+/**
+ * An implementation of {@link QuotaEnforcerFactory} for that creates and returns {@link AmbryCUQuotaEnforcer} object.
```
tbingol, line 27: nit: "for" is redundant ?

ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
```diff
@@ -75,6 +81,29 @@ public static QuotaResource fromAccountId(short accountId) {
...
+   * @throws QuotaException if appropriate headers aren't found in the {@link RestRequest}.
+   */
+  public static QuotaResource fromRestRequest(RestRequest restRequest) throws QuotaException {
+    try {
+      Account account = RestUtils.getAccountFromArgs(restRequest.getArgs());
+      if (account.getQuotaResourceType() == QuotaResourceType.ACCOUNT) {
+        return QuotaResource.fromAccountId(account.getId());
+      } else {
+        Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
+        return QuotaResource.fromContainerId(account.getId(), container.getId());
```
tbingol, line 97: Nit: else is not needed

tbingol (approved):

ambry-api/src/main/java/com/github/ambry/quota/QuotaResource.java
```diff
@@ -75,6 +81,29 @@ public static QuotaResource fromAccountId(short accountId) {
...
 
+  /**
+   * Create {@link QuotaResource} for the specified {@link RestRequest}.
+   * @param restRequest {@link RestRequest} object.
+   * @return QuotaResource extracted from headers of {@link RestRequest}.
+   * @throws QuotaException if appropriate headers aren't found in the {@link RestRequest}.
+   */
+  public static QuotaResource fromRestRequest(RestRequest restRequest) throws QuotaException {
+    try {
+      Account account = RestUtils.getAccountFromArgs(restRequest.getArgs());
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-api/src/main/java/com/github/ambry/quota/QuotaSource.java
ankagrawal: I have updated this comment with "if the implementation doesn't need to track any system resource".

Note that tracking system resource along with quota is very implementation specific, so its hard to be more specific here. For example, in Ambry, the current storage quota implementation doesn't track the overall storage usage in a cluster (unlike bandwidth quota which does track overall bandwidth usage in a node). As a side-effect, a call to `isQuotaExceedAllowed` for storage quota's `QuotaEnforcer` implementation, will always return false.
Another implementation of storage quota could have, maybe, taken overall storage usage in a cluster into account, to decide the return value for `isQuotaExceedAllowed`.
So in cases where implementation doesn't need to track system resources (for whatever reason), the quota source will return the specified values as mentioned in the comment.

PR #2048, opened 2022-02-08
[BANDWIDTH_THROTTLING] Implement QuotaResourceMutexFactory to synchronize based on QuotaResource value.

ankagrawal:
QuotaManager implementation might need to synchronize quota usage accounting for the same QuotaResource. This synchronization needs to happen based on the QuotaResource object's value rather than on QuotaResource object instances (which is how java's monitor works by default). This PR creates a synchronization construct to synchronize based on the QuotaResource value.

### DIFF
--- a/ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.quota;
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.locks.ReentrantLock;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Class to provide synchronization constructs (lock and unlock) for all {@link QuotaResource} objects based on their value.
+ * This class can be used to enforce mutual exclusion between multiple {@link QuotaResource} objects that represent the
+ * same quota resource (as determined by {@link QuotaResource#equals}) method.
+ *
+ * To ensure the synchronization, this class internally creates a unique {@link ReentrantLock} for each quota resource,
+ * and maintains a {@link ConcurrentHashMap} to cache the mapping and reuse same {@link ReentrantLock} across multiple
+ * {@link QuotaResource} instances that represent the same quota resource.
+ *
+ * In order to prevent {@link ReentrantLock}s from being inadvertently reused outside this scope, this class doesn't
+ * leak the {@link ReentrantLock}s outside. It provides only lock and unlock constructs to the callers.
+ */
+public class QuotaResourceSynchronizer {
+  private static final Logger LOGGER = LoggerFactory.getLogger(QuotaResourceMutexCache.class);
+  protected final QuotaResourceMutexCache quotaResourceMutexCache = new QuotaResourceMutexCache();
+  private final QuotaMetrics quotaMetrics;
+
+  /**
+   * Constructor for {@link QuotaResourceSynchronizer}.
+   * @param quotaMetrics {@link QuotaMetrics} object.
+   */
+  public QuotaResourceSynchronizer(QuotaMetrics quotaMetrics) {
+    this.quotaMetrics = quotaMetrics;
+  }
+
+  /**
+   * Acquires the lock for the quota resource represented by the specified {@link QuotaResource}.
+   * @param quotaResource {@link QuotaResource} object representing the quota resource for which lock needs to be
+   *                      acquired.
+   */
+  public void lock(QuotaResource quotaResource) {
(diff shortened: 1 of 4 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
```diff
@@ -0,0 +1,133 @@
...
+        cleanupLock.readLock().unlock();
+      }
+    }
+
+    /**
+     * Method to cleanup the mutexes that are no longer in use.
+     */
+    protected void cleanupCache() {
+      try {
+        cleanupLock.writeLock().lock();
```
cgtz, line 122: Concurrent iteration is supported in ConcurrentHashMap. My understanding of the RW lock here is to account for the case where someone is getting a lock but has not locked it yet and this bg thread sees that no one is using the lock and removes it. Then, another thread tries to get a lock for that resource, but it actually gets a new lock. Is this correct?

Here i'm not even sure if that will fully protect it since the ReentrantLock::lock is called outside of the cleanupLock read scope.

ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
```diff
@@ -0,0 +1,102 @@
...
+     * 2. This cache never removes entries until the service (Ambry frontend) is restarted. In the worst case the size
+     *    of the cache might grow upto the total number of QuotaResources (accounts + containers + services) for which quota
+     *    is being enforced. Periodically cleaning up cache is one way to reduce the memory pressure, but will add more
+     *    complexity to maintain thread safety (and maybe more latency to access the cache too). Ambry's frontend keeps
+     *    the accounts and containers in memory at multiple places currently (e.g, AccountService implementation), with the
+     *    assumption that the total number of accounts and containers are small enough to keep in memory. As long as that
+     *    assumption holds, not cleaning up this cache is fine. If that assumption doesn't hold, a lot of code might need
+     *    restructuring, and this cache cleanup should be taken up as part of that effort.
+     */
+    protected final ConcurrentHashMap<QuotaResource, ReentrantLock> mutexCache = new ConcurrentHashMap<>();
```
tbingol, line 89: ReentrantLock lets the same thread lock the same object more than once. Is there any case where the same thread can serve to multiple requests for the same QuotaResource?

tbingol (changes requested):

ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
ankagrawal: The reason why I didn't consider anything other than `ReentrantLock` is because that's the default `Lock` implementation in java standard library (The other implementation is `ReadWriteLock` which doesn't apply to this case). Even the `sycnhronized` blocks in java are re-entrant. Although there shouldn't be any case where a thread can serve multiple requests for the same `QuotaResource` at the same time, `ReentrantLocks` generally also help with taking the same locks in other blocks (or methods) too. Otherwise care should be taken that a lock once held in a flow, isn't required to be held again, or else the thread will block itself.

ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
```diff
@@ -0,0 +1,102 @@
...
+    } else {
+      LOGGER.warn("Attempted unlock for {} when lock wasn't held.", quotaResource);
+      quotaMetrics.quotaResourceInvalidUnlockAttemptCount.inc();
+    }
+  }
+
+  /**
+   * A cache of unique mutexes for all {@link QuotaResource} objects that represent the same quota resource.
+   */
+  static class QuotaResourceMutexCache {
```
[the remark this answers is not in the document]
tbingol: I agree with the decision not to use the singletons, however, for my own curiosity, how is this helping a future cleanup method? Note that ConcurrentHashMap has the clear() method to clear all entries.

tbingol (approved):

ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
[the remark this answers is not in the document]
ankagrawal: By cleanup, I actually mean to cleanup entries for only those "QuotaResource"s that aren't being actively used. Something like a thread that wakes up periodically and iterates through the map to check if any "ReentrantLock" is not held and no one is waiting for it. Entries for such ReentrantLocks can be remove to clean up memory for unused locks.

ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
```diff
@@ -0,0 +1,102 @@
...
+    protected final ConcurrentHashMap<QuotaResource, ReentrantLock> mutexCache = new ConcurrentHashMap<>();
+
+    /**
+     * Returns the mutex in the form of {@link ReentrantLock} object for the specified {@link QuotaResource}.
+     * If the mutex is not already present in the cache, it creates a new mapping.
+     * @param quotaResource {@link QuotaResource} object.
+     * @return ReentrantLock object as mutex for the specified quotaResource.
+     */
+    ReentrantLock get(QuotaResource quotaResource) {
+      mutexCache.putIfAbsent(quotaResource, new ReentrantLock());
```
cgtz, line 98: nit: you can simplify this to one line (and avoid constructing the ReentrantLock if it already exists):
`return mutexCache.computeIfAbsent(quotaResource, k -> new ReentrantLock())`

cgtz (approved):
LGTM, one hashmap thing to fix

ambry-quota/src/main/java/com/github/ambry/quota/QuotaResourceSynchronizer.java
ankagrawal: fixed

PR #2050, opened 2022-02-09
Return recommendation from QuotaEnforcer charge method.

ankagrawal:


justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/quota/QuotaEnforcer.java
```diff
@@ -35,11 +35,12 @@
...
 
   /**
-   * Charges the request cost against the quota.
+   * Charges the request cost against the quota and returns the {@link QuotaRecommendation} after charging.
    * @param restRequest {@link RestRequest} object.
    * @param requestCostMap {@link Map} of {@link QuotaName} to the cost incurred to handle the request.
+   * @return QuotaRecommendation object with the recommendation after charging is done.
    */
-  void charge(RestRequest restRequest, Map<QuotaName, Double> requestCostMap)
+  QuotaRecommendation charge(RestRequest restRequest, Map<QuotaName, Double> requestCostMap)
```
SophieGuo410, line 43: nit: would it be better if the method renamed as something like chargeAndProvideRecommendations? Since it's not only charge the quota source.

ankagrawal: I want to keep the names simple here. The main job of `charge` is to charge for usages. Returning `QuotaRecommendation` is just a side effect which is ok for callers to ignore if they don't need it, and that won't change anything.
The semantics I am thinking of is something like list.remove or map.putIfAbsent in java. `list.remove` returns boolean, but thats just a side effect. also `Map::putIfAbsent` returns previous mapping if present, but doesn't call it out in the name.

If you still strongly believe that this should be renamed, I would be happy to do it.

SophieGuo410: I see, thx for the explanation, I'm ok with keep it simple.

SophieGuo410 (approved):

PR #2051, opened 2022-02-15
[BANDWIDTH_THROTTLING] Introduce QuotaAction as the recommended action to be taken in order to ensure requests are quota compliant.

ankagrawal:
For out of quota user/chunk requests, QuotaEnforcers can decide to either DELAY a request (in case of bandwidth/CU enforcers), or REJECT a request in case of storage request. This PR introduces QuotaAction to represent the recommended action by QuotaEnforcers and QuotaManager.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/quota/ThrottlingRecommendation.java
+++ b/ambry-api/src/main/java/com/github/ambry/quota/ThrottlingRecommendation.java
@@ -13,7 +13,6 @@
  */
 package com.github.ambry.quota;
 
-import io.netty.handler.codec.http.HttpResponseStatus;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -24,25 +23,22 @@ import java.util.Map;
  */
 public class ThrottlingRecommendation {
   public static final long NO_RETRY_AFTER_MS = -1; // No retry needed when request is not throttled.
-  private final boolean throttle;
+  private final QuotaAction quotaAction;
   private final Map<QuotaName, Float> quotaUsagePercentage;
-  private final HttpResponseStatus recommendedHttpStatus;
   private final long retryAfterMs;
   private final QuotaUsageLevel quotaUsageLevel;
 
   /**
    * Constructor for {@link ThrottlingRecommendation}.
-   * @param throttle flag indicating if request should be throttled.
+   * @param quotaAction {@link QuotaAction} object as the recommended action to take.
    * @param quotaUsagePercentage A {@link Map} of {@link QuotaName} to usage percentage.
-   * @param recommendedHttpStatus {@link HttpResponseStatus} representing overall recommended http status.
    * @param retryAfterMs time in ms after which request should be retried. -1 if request is not throttled.
    * @param quotaUsageLevel {@link QuotaUsageLevel} object.
    */
-  public ThrottlingRecommendation(boolean throttle, Map<QuotaName, Float> quotaUsagePercentage,
-      HttpResponseStatus recommendedHttpStatus, long retryAfterMs, QuotaUsageLevel quotaUsageLevel) {
-    this.throttle = throttle;
+  public ThrottlingRecommendation(QuotaAction quotaAction, Map<QuotaName, Float> quotaUsagePercentage,
+      long retryAfterMs, QuotaUsageLevel quotaUsageLevel) {
+    this.quotaAction = quotaAction;
     this.quotaUsagePercentage = new HashMap<>(quotaUsagePercentage);
-    this.recommendedHttpStatus = recommendedHttpStatus;
     this.retryAfterMs = retryAfterMs;
     this.quotaUsageLevel = quotaUsageLevel;
   }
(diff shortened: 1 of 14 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
```diff
@@ -113,9 +116,9 @@ public void shutdown() {
    * @return QuotaRecommendation object.
    */
   private QuotaRecommendation buildQuotaRecommendation(float usage, QuotaName quotaName) {
-    boolean shouldThrottle = (usage >= 100);
-    return new QuotaRecommendation(shouldThrottle, usage, quotaName,
-        shouldThrottle ? throttleRetryAfterMs : NO_THROTTLE_RETRY_AFTER_MS);
+    QuotaAction quotaAction = (usage >= 100) ? QuotaAction.DELAY : QuotaAction.ALLOW;
+    return new QuotaRecommendation(quotaAction, usage, quotaName,
+        (quotaAction == QuotaAction.DELAY) ? throttleRetryAfterMs : NO_THROTTLE_RETRY_AFTER_MS);
```
SophieGuo410, line 121: nit: I feel like the concept need to be consistent. From this logic, when quotaAction != DELAY, which could be ACTIVE or REJECT, then the getRetryAfterMs() will return -1. And the java doc in QuotaRecommendation.java mentioned "If request is not throttled then returns -1". But when I check this logic:
`  public boolean shouldThrottle() {
    return this.quotaAction == QuotaAction.REJECT || this.quotaAction == QuotaAction.DELAY;
  }`
It seems like REJECT and DELAY will both to be considered as should throttle.

ambry-api/src/main/java/com/github/ambry/quota/ThrottlingRecommendation.java
```diff
@@ -51,21 +47,28 @@ public ThrottlingRecommendation(boolean throttle, Map<QuotaName, Float> quotaUsa
    * @return true if recommendation is to throttle. false otherwise.
    */
   public boolean shouldThrottle() {
```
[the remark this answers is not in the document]
ankagrawal: Your understanding is correct. I have updated the naming to reflect that. Let me know if this makes sense now.

ambry-quota/src/main/java/com/github/ambry/quota/SimpleQuotaRecommendationMergePolicy.java
```diff
@@ -53,27 +54,24 @@ public SimpleQuotaRecommendationMergePolicy(QuotaConfig quotaConfig) {
...
     for (QuotaRecommendation recommendation : quotaRecommendations) {
-      boolean currentQuotaShouldThrottle = recommendation.shouldThrottle();
+      QuotaAction currentQuotaAction = recommendation.getQuotaAction();
       if (recommendation.getQuotaName() == QuotaName.READ_CAPACITY_UNIT
           || recommendation.getQuotaName() == QuotaName.WRITE_CAPACITY_UNIT) {
-        currentQuotaShouldThrottle &= quotaConfig.requestThrottlingEnabled;
+        currentQuotaAction = quotaConfig.requestThrottlingEnabled ? currentQuotaAction : QuotaAction.ALLOW;
       } else if (recommendation.getQuotaName() == QuotaName.STORAGE_IN_GB) {
-        currentQuotaShouldThrottle &= quotaConfig.storageQuotaConfig.shouldThrottle;
+        currentQuotaAction = quotaConfig.storageQuotaConfig.shouldThrottle ? currentQuotaAction : QuotaAction.ALLOW;
```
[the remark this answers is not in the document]
ankagrawal: This is sort of because the way we developed. You are right in pointing out that this needs refactoring. I will create a separate PR out for refactoring the quota configs. There are some configs which can be simplified into one, and there are a few that can just be removed.

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
ankagrawal: You are right that both REJECT and DELAY will be considered as "throttled", because both of them means that the request is out of quota and needs to be controlled. "REJECT" and "DELAY" are the recommended action for controlling.
Not every QuotaEnforcer needs to send both. In the current implementation, requests that are out of quota for bandwidth will never be rejected, but will be delayed. Hence CU implementations will set action as "DELAY" for out-of-quota requests. The requests that are out of quota for storage quota will be rejected. Hence storage quota sets "REJECT" as action and not "DELAY". In future there could be some fancy enforcer implementation that can set DELAY in certain conditions, while REJECT in certain other conditions (maybe something like DELAY for bandwidth out-of-quota, but if there are a lot of requests in queue, then REJECT).

So overall both REJECT and DELAY means that request is out-of-quota and needs to be controlled. 
We will only set NO_THROTTLE_RETRY_AFTER_MS if request is not throttled, i.e, if request is allowed.

SophieGuo410 (approved):
lgtm!

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
```diff
@@ -113,9 +116,9 @@ public void shutdown() {
    * @return QuotaRecommendation object.
    */
   private QuotaRecommendation buildQuotaRecommendation(float usage, QuotaName quotaName) {
-    boolean shouldThrottle = (usage >= 100);
-    return new QuotaRecommendation(shouldThrottle, usage, quotaName,
-        shouldThrottle ? throttleRetryAfterMs : NO_THROTTLE_RETRY_AFTER_MS);
+    QuotaAction quotaAction = (usage >= 100) ? QuotaAction.DELAY : QuotaAction.ALLOW;
```
tbingol, line 119: Consider replacing the 100 with a config param. In the future, we may let users to exceed their quota up to a certain point

tbingol (approved):

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/AmbryCUQuotaEnforcer.java
ankagrawal: Added it as a final variable for now, instead of hard coding.
The main reason why I am hesitant to create a config is that it adds one more config, that wouldn't be used now, which will make it confusing.

PR #2053, opened 2022-02-17
Update cosmos DB SDK to latest version 4.x

Arun-LinkedIn:
Changes to update cosmos client java SDK to 4.x from 2.x.

Reference guide: https://docs.microsoft.com/en-us/azure/cosmos-db/sql/migrate-java-v4-sdk?tabs=java-v4-async


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
```diff
@@ -254,16 +253,10 @@ private int compactPartitionBucketed(String partitionPath, String fieldName, lon
...
   List<CloudBlobMetadata> getDeadBlobs(String partitionPath, String fieldName, long startTime, long endTime,
-      int maxEntries) throws CloudStorageException {
-    try {
-      return cosmosDataAccessor.getDeadBlobs(partitionPath, fieldName, startTime, endTime, maxEntries);
-    } catch (DocumentClientException dex) {
-      throw AzureCloudDestination.toCloudStorageException(
-          "Failed to query deleted blobs for partition " + partitionPath, dex, azureMetrics);
-    }
+      int maxEntries) {
+    return cosmosDataAccessor.getDeadBlobs(partitionPath, fieldName, startTime, endTime, maxEntries);
```
ankagrawal, line 259: No exception check needed here?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
```diff
@@ -101,42 +109,55 @@
...
+    }
     if (azureCloudConfig.cosmosDirectHttps) {
       logger.info("Using CosmosDB DirectHttps connection mode");
-      connectionPolicy.setConnectionMode(ConnectionMode.Direct);
-    }
-    if (cloudConfig.vcrProxyHost != null) {
-      connectionPolicy.setProxy(cloudConfig.vcrProxyHost, cloudConfig.vcrProxyPort);
+      cosmosClientBuilder.directMode(new DirectConnectionConfig(), gatewayConnectionConfig);
+    } else {
+      cosmosClientBuilder.gatewayMode(gatewayConnectionConfig);
```
ankagrawal, line 136: Nit: maybe initialize the GatewayConnectionConfig here.

ankagrawal (commented):
Mostly looks good. Some minor comments.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageCompactor.java
Arun-LinkedIn: Thank you for catching. Added the exception check.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
Arun-LinkedIn: It looks like GatewayConnectionConfig is also being used in their direct connection mode (for some of the metadata operations). Due to that, I had to initialize it outside the `if` condition.

ankagrawal (approved):

cgtz (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudConfig.java
```diff
@@ -96,6 +101,24 @@
   @Config(COSMOS_DELETED_CONTAINER_COLLECTION_LINK)
   public final String cosmosDeletedContainerCollectionLink;
```
cgtz, line 102: Sorry for the late feedback, upon reading this, it seems like cosmosCollectionLink/cosmosDeletedContainerCollectionLink aren't used in the codebase anymore, so even thought they are present here, they won't be honored by this version of the code.

I think we can go ahead and remove them from linkedin/ambry. Configurations within LI will have to set both of these to be compatible with older versions of the code, but it would keep the newest code slightly cleaner and we could clean up the configs after we roll this out.

Another more complicated approach would be to keep the collection link configs as is, and just parse out the DB and container names from them. Benefit of this would be that a config migration isn't needed, but it leads to the java code being messier, imo.

build.gradle
```diff
@@ -434,6 +434,7 @@ project(':ambry-cloud') {
         compile "io.dropwizard.metrics:metrics-core:$metricsVersion"
         compile "io.dropwizard.metrics:metrics-jmx:$metricsVersion"
         compile "com.microsoft.azure:azure-cosmosdb:$azureCosmosDbVersion"
```
cgtz, line 436: Can you remove the dependency on the old cosmos lib

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudConfig.java
Arun-LinkedIn: Sure! Removed the old configs for cosmos collections from here. Will keep them in LI until the changes are rolled out.

PR #2054, opened 2022-02-23
[BANDWIDTH_THROTTLING] Parse enforcer and source pairs into a list of pairs instead of map.

ankagrawal:
This will allow multiple instances of same enforcer types in quota config.

### DIFF
--- a/ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
@@ -18,8 +18,8 @@ import com.github.ambry.account.Account;
 import com.github.ambry.account.AccountService;
 import com.github.ambry.accountstats.AccountStatsStore;
 import com.github.ambry.config.QuotaConfig;
-import com.github.ambry.messageformat.BlobInfo;
 import com.github.ambry.rest.RestRequest;
+import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Utils;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -30,6 +30,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.slf4j.Logger;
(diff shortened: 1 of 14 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/AmbryQuotaManager.java
```diff
@@ -191,22 +193,22 @@ protected void onAccountUpdateNotification(Collection<Account> updatedAccounts)
...
-    Map<String, String> quotaEnforcerSourceMap = new HashMap<>();
+    List<Pair<String, String>> quotaEnforcerSourcePairs = new ArrayList<>();
     JSONObject root = new JSONObject(quotaEnforcerSourceJson);
     JSONArray all = root.getJSONArray(QuotaConfig.QUOTA_ENFORCER_SOURCE_PAIR_INFO_STR);
     for (int i = 0; i < all.length(); i++) {
       JSONObject entry = all.getJSONObject(i);
       String enforcer = entry.getString(QuotaConfig.ENFORCER_STR);
       String source = entry.getString(QuotaConfig.SOURCE_STR);
-      quotaEnforcerSourceMap.put(enforcer, source);
+      quotaEnforcerSourcePairs.add(new Pair<>(enforcer, source));
```
SophieGuo410, line 209: May I know in which case the QuotaEnforcerFactory will mapping to multiple QuotaSourceFactory? Is this for migration? Also, can you update the test corresponding to the changes?

Arun-LinkedIn: Have same question as Sophie. When would we have multiple instances of same enforcer type?

ankagrawal: The methods changes are internal private methods of this class. All the tests that rely on AmbryQuotaManager, will automatically test the new changes. More explicit tests that exercise the logic for multiple QuotaEnforcer instances of the same type are present in the PR#2052.

As for the question about when multiple instances of the same enforcer type might be used, I realized this when writing some tests. It's theoretically possible that a same QuotaEnforcer implementation might differ in functionality based on some arguments passed to it while construction. In that case there will be two different instances but with the same type. In PR#2052, `AmbryQuotaManagerTest::buildAmbryQuotaManager` and `AmbryQuotaManagerTest::buildTestQuotaEnforcerSourceInfoPairJson` has an example for such a scenario.

PR #2061, opened 2022-02-28
[BANDWIDTH_THROTTLING] Add a boolean flag to ResponseInfo, so that OperationTrackers can react to quota related rejections, and ultimately non quota compliant requests can be rejected.

ankagrawal:
OperationControllers need to reject certain non quota compliant requests. This PR adds a boolean flag to ResponseInfo, so that OperationTrackers can react to quota related rejections, and ultimately non quota compliant requests can be rejected.

Note that each operation still needs to handle the exception in its `handleResponse` method and needs to throw appropriate exception back to user. That PR will come later.

OperationController doesn't throw exception to reject requests currently. That PR will come later.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/network/ResponseInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/network/ResponseInfo.java
@@ -21,14 +21,19 @@ import io.netty.util.ReferenceCountUtil;
 
 /**
  * The response from a {@link NetworkClient} comes in the form of an object of this class.
- * This class consists of the request associated with this response, along with either a non-null exception if there
- * was an error sending the request or a non-null ByteBuffer containing the successful response received for this
- * request. Also, this class contains {@link DataNodeId} to which the request is issued.
+ * This class consists of the following information about the response.
+ * Request associated with this response.
+ * Either a non-null exception if there was an error sending the request or a non-null ByteBuffer containing the
+ * successful response received for this request.
+ * A flag to indicate if this request was rejected due quota non-compliance.
+ * The {@link DataNodeId} to which the request is issued.
  */
 public class ResponseInfo extends AbstractByteBufHolder<ResponseInfo> {
   private final RequestInfo requestInfo;
   private final NetworkClientErrorCode error;
   private final DataNodeId dataNode;
+  private final boolean quotaRejected;
+
   /**
    * Response received from network in the form of serialized bytes.
    */
@@ -41,11 +46,20 @@ public class ResponseInfo extends AbstractByteBufHolder<ResponseInfo> {
   /**
    * Constructs a ResponseInfo with the given parameters.
    * @param requestInfo the {@link RequestInfo} associated with this response.
+   * @param quotaRejected {@code true} if this request was rejected due quota non-compliance. {@code false} otherwise.
+   */
+  public ResponseInfo(RequestInfo requestInfo, boolean quotaRejected) {
+    this(requestInfo, null, null, requestInfo == null ? null : requestInfo.getReplicaId().getDataNodeId(), quotaRejected);
+  }
+
+  /**
+   * Constructs a ResponseInfo with the given parameters.
+   * @param requestInfo the {@link RequestInfo} associated with this response.
    * @param error the error encountered in sending this request, if there is any.
    * @param content the response received for this request.
    */
   public ResponseInfo(RequestInfo requestInfo, NetworkClientErrorCode error, ByteBuf content) {
-    this(requestInfo, error, content, requestInfo == null ? null : requestInfo.getReplicaId().getDataNodeId());
+    this(requestInfo, error, content, requestInfo == null ? null : requestInfo.getReplicaId().getDataNodeId(), false);
   }
 
   /**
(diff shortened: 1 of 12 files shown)


justinlin-linkedin:
In `ResponseInfo` there is an `NetworkClientErrorCode`, we have a lot of logic relying on the `errorCode` being null. To make sure we handle all the corner case, I suggest we add a new NetworkClientErrorCode`, like `ExceededQuota`. And only when the error code is `ExceededQuota`, we try to get the quota exception from `ResponseInfo`.

ambry-router/src/main/java/com/github/ambry/router/AdaptiveOperationTracker.java
```diff
@@ -96,20 +96,21 @@
...
-    }
-    if (trackedRequestFinalState != TrackedRequestFinalState.TIMED_OUT
-        || !routerConfig.routerOperationTrackerExcludeTimeoutEnabled) {
-      getLatencyHistogram(replicaId).update(elapsedTime);
-      if (routerConfig.routerOperationTrackerMetricScope != OperationTrackerScope.Datacenter) {
-        // This is only used to report whole datacenter histogram for monitoring purpose
-        Histogram histogram =
-            replicaId.getDataNodeId().getDatacenterName().equals(datacenterName) ? localDcHistogram : crossDcHistogram;
-        histogram.update(elapsedTime);
+    if(trackedRequestFinalState != TrackedRequestFinalState.QUOTA_REJECTED) {
```
tbingol, line 99: nit: space after "if"

tbingol, line 99: no need to indent the whole code. Just do

if (trackedRequestFinalState == TrackedRequestFinalState.QUOTA_REJECTED) {
  return;
}

ambry-router/src/main/java/com/github/ambry/router/RouterUtils.java
```diff
@@ -205,6 +205,9 @@ static void replaceOperationException(AtomicReference<Exception> operationExcept
       NonBlockingRouterMetrics routerMetrics, ResponseInfo responseInfo, Deserializer<R> deserializer,
       Function<R, ServerErrorCode> errorExtractor) {
     R response = null;
+    if(responseInfo.getQuotaException() != null) {
```
tbingol, line 208: nit: add space after "if"

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
```diff
@@ -86,6 +86,9 @@ public boolean hasSucceeded() {
 
   @Override
   public boolean hasFailed() {
+    if(quotaRejected) {
```
tbingol, line 89: nit: add space after "if"

tbingol (approved):

ambry-router/src/main/java/com/github/ambry/router/AdaptiveOperationTracker.java
ankagrawal: done.

ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperationTracker.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/RouterUtils.java
ankagrawal: done.

ankagrawal:
> In `ResponseInfo` there is an `NetworkClientErrorCode`, we have a lot of logic relying on the `errorCode` being null. To make sure we handle all the corner case, I suggest we add a new NetworkClientErrorCode`, like `ExceededQuota`. And only when the error code is `ExceededQuota`, we try to get the quota exception from `ResponseInfo`.

I don't this it would be clean to add quota related error to `NetworkClientErrorCode`. `NetworkClientErrorCode` represents error received after the request has been sent over the network (or attempted to be sent over the network). Quota compliance errors happen much before that. The real invariant is that if we got quota related errors, then there shouldn't be a case where we will get NetworkClientError, because quota error would have happened much before that in `OperationController`. Also note that network error is used for setting other states related to liveness and latencies of storage nodes, while in case of quota errors no such state or metric should be updated.
In order to simplify some more, I have removed QuotaException object from ResponseInfo (because we will throw QuotaException out of OperationController only in case request is REJECTed).

ambry-api/src/main/java/com/github/ambry/network/ResponseInfo.java
```diff
@@ -14,21 +14,27 @@
 package com.github.ambry.network;
 
 import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.quota.QuotaException;
```
justinlin-linkedin, line 17: not used anywhere.

justinlin-linkedin (approved):
LGTM

PR #2064, opened 2022-03-02
Upgrade netty version

justinlin-linkedin:
upgrade netty version


cgtz (approved):

cgtz:
I see that 4.1.69 finally makes netty-all depend on other packages instead of bundling them! This is nice to prevent runtime conflicts for libraries that use the individual packages instead of netty-all.

https://netty.io/news/2021/10/11/4-1-69-Final.html

SophieGuo410 (approved):

PR #2065, opened 2022-03-07
[BANDWIDTH_THROTTLING] If bandwidth throttling is enabled, quota charge shouldn't happen after chunk requests are processed.

ankagrawal:
With bandwidth throttling enabled, QuotaAwareOperationController will do the quota charges before chunk requests are sent over network.

Also introduces get QuotaConfig method in QuotaChargeCallback to access QuotaConfig in Operation classes.

### DIFF
--- a/ambry-quota/src/main/java/com/github/ambry/quota/QuotaUtils.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/QuotaUtils.java
@@ -90,6 +90,16 @@ public class QuotaUtils {
     }
   }
 
+  /**
+   * @param quotaChargeCallback {@link QuotaChargeCallback} object.
+   * @return {@code true} if the charge needs to happen AFTER a request is processed. {@code false} otherwise.
+   */
+  public static boolean postProcessCharge(QuotaChargeCallback quotaChargeCallback) {
+    // Bandwidth throttling based implementation of quota enforcement will need ensure quota compliance BEFORE sending
+    // the chunk request to server. Hence return false if bandwidth throttling feature is enabled.
+    return quotaChargeCallback != null && !quotaChargeCallback.getQuotaConfig().bandwidthThrottlingFeatureEnabled;
+  }
+
   /*
    * @return QuotaName of the CU quota associated with {@link RestRequest}.
    */
--- a/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
@@ -31,6 +31,7 @@ import com.github.ambry.commons.LoggingNotificationSystem;
 import com.github.ambry.commons.ResponseHandler;
 import com.github.ambry.config.CryptoServiceConfig;
 import com.github.ambry.config.KMSConfig;
+import com.github.ambry.config.QuotaConfig;
 import com.github.ambry.config.RouterConfig;
 import com.github.ambry.config.VerifiableProperties;
 import com.github.ambry.messageformat.BlobAll;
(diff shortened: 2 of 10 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/QuotaUtils.java
```diff
@@ -90,6 +90,15 @@ public static QuotaChargeCallback buildQuotaChargeCallback(RestRequest restReque
     }
   }
 
+  /**
+   * @param quotaChargeCallback {@link QuotaChargeCallback} object.
+   * @return {@code true} if the charge needs to happen AFTER a request is processed. {@code false} otherwise.
+   */
+  public static boolean postProcessCharge(QuotaChargeCallback quotaChargeCallback) {
+    boolean retval = quotaChargeCallback != null && !quotaChargeCallback.getQuotaConfig().bandwidthThrottlingFeatureEnabled;
```
tbingol, line 98: Why do we return true when the bandwidthThrottlingFeature is not enabled?

Do we need retval here?

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -375,6 +386,9 @@ public void testSimpleBlobGetSuccess() throws Exception {
       options = new GetBlobOptionsInternal(new GetBlobOptionsBuilder().operationType(operationType).build(), false,
           routerMetrics.ageAtGet);
       getAndAssertSuccess(false, false, expectedLifeVersion);
+      if(!quotaChargeCallback.getQuotaConfig().bandwidthThrottlingFeatureEnabled) {
```
tbingol, line 389: nit: space after "if"

tbingol (changes requested):

ambry-quota/src/main/java/com/github/ambry/quota/QuotaUtils.java
ankagrawal: So there are going to be two implementations of QuotaChargeCallback going forward.
One will be what we have currently. It charges for quota, after the chunks are processed by the server. So it charged "post process". Its named `RejectingQuotaChargeCallback`, which is a bad name actually. 

The new "bandwidth throttling" based implementation will charge for quota "before" sending the chunk request to server. But note that the new implementation will also reject requests, because for storage quota we will reject the requests if they exceed quota.

So in summary the right set of changes are
1. Rename `RejectingQuotachargeCallback` to `PostProcessQuotaChargeCallback`.
2. New implementation should be called `PreProcessQuotaChargeCallback`.
3. If bandwidth throttling is enabled, then don't do stuff related to `PostProcessQuotaChargeCallback`. Or in other words, if bandwidth throttling is not enabled, then do the `PostProcessQuotaChargeCallback` stuff.

The logic above implements the point 3. PR [2069](https://github.com/linkedin/ambry/pull/2069/files) implements the point 1. PR [2070](https://github.com/linkedin/ambry/pull/2069/files) implements the point 2.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
ankagrawal: fixed.

tbingol (approved):

SophieGuo410 (approved):

ambry-quota/src/main/java/com/github/ambry/quota/QuotaUtils.java
```diff
@@ -90,6 +90,14 @@ public static QuotaChargeCallback buildQuotaChargeCallback(RestRequest restReque
     }
   }
 
+  /**
+   * @param quotaChargeCallback {@link QuotaChargeCallback} object.
+   * @return {@code true} if the charge needs to happen AFTER a request is processed. {@code false} otherwise.
```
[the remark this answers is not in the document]
ankagrawal: ### done.

PR #2071, opened 2022-03-09
Aggregate Get and Put qps to detect live traffic before container deletion

SophieGuo410:
We plan to automatically test live traffic before the container deletion.
In order to do that, Nuage team support a flow which can only take two metrics, one for put request and one for get request.
So this pr support 
1. Aggregate the per request QPS metric into the get and put QPS metric.
2. Code refactoring to move containerMetricsEnabledRequestTypes into open source, so we won't forget to define the request type in close source. Currently we forget to include UndeleteBlob and PutBlob into the list in close source for container metrics.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
@@ -41,6 +41,8 @@ public class FrontendConfig {
   public static final String NAMED_BLOB_DB_FACTORY = PREFIX + "named.blob.db.factory";
   public static final String CONTAINER_METRICS_EXCLUDED_ACCOUNTS = PREFIX + "container.metrics.excluded.accounts";
   public static final String ACCOUNT_STATS_STORE_FACTORY = PREFIX + "account.stats.store.factory";
+  public static final String CONTAINER_METRICS_ENABLED_REQUEST_TYPES = PREFIX + "container.metrics.enabled.request.types";
+  public static final String CONTAINER_METRICS_ENABLED_GET_REQUEST_TYPES = PREFIX + "container.metrics.enabled.get.request.types";
 
   // Default values
   private static final String DEFAULT_ENDPOINT = "http://localhost:1174";
@@ -49,6 +51,12 @@ public class FrontendConfig {
 
   private static final String DEFAULT_ACCOUNT_STATS_STORE_FACTORY =
       "com.github.ambry.accountstats.InmemoryAccountStatsStoreFactory";
+
+  private static final String DEFAULT_CONTAINER_METRICS_ENABLED_REQUEST_TYPES =
+      "DeleteBlob,GetBlob,GetBlobInfo,GetSignedUrl,PostBlob,UpdateBlobTtl,UndeleteBlob,PutBlob";
+
+  private static final String DEFAULT_CONTAINER_METRICS_ENABLED_GET_REQUEST_TYPES = "GetBlob,GetBlobInfo,GetSignedUrl";
+
 
   /**
    * Cache validity in seconds for non-private blobs for GET.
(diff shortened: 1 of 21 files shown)


ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
```diff
@@ -219,6 +228,20 @@
   @Default(DEFAULT_ACCOUNT_STATS_STORE_FACTORY)
   public final String accountStatsStoreFactory;
 
+  /**
```
cgtz, line 231: What's the reason for making these two configurable?

ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
```diff
@@ -50,6 +52,13 @@
   private static final String DEFAULT_ACCOUNT_STATS_STORE_FACTORY =
       "com.github.ambry.accountstats.InmemoryAccountStatsStoreFactory";
 
+  private static final String DEFAULT_CONTAINER_METRICS_ENABLED_REQUEST_TYPES =
+      "{\"DeleteBlob\", \"GetBlob\", \"GetBlobInfo\", \"GetSignedUrl\", \"PostBlob\", \"UpdateBlobTtl\", \"UndeleteBlob\", \"PutBlob\"}";
```
cgtz, line 56: I think this default value needs to just be a comma separated list for the string.split code in RestRequestMetricsGroup to work.

ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
SophieGuo410: In close source we have this cfg.containerMetricsEnabledRequestTypes which determine which metric need to emit container level metric. And in Container Metric we are hardcode the request type which enable the container metric. So if we update the open source, it's easy to forget modify the close source. So I think it's better if in close source we can get the containerMetricsEnabledRequestTypes from frontend config as well. In fact, we do forget to add UndeleteBlob and PutBlob in cfg.containerMetricsEnabledRequestTypes. And for the second config is to determine if the request is get or put type, in order to aggregate the per request metric to one put and one get metric since Nuage can only support reading two simple metric for live traffic check.

ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
SophieGuo410: Yeah, I've updated the code.

ambry-api/src/main/java/com/github/ambry/frontend/ContainerMetrics.java
```diff
@@ -50,6 +50,8 @@
   // 410
   private final Counter goneCount;
 
+  private final Counter totalQpsCount;
```
justinlin-linkedin, line 53: Naming is weird, we don't need qps in the name, Counter is an implicit qps. We can just call it totalCount.

ambry-api/src/main/java/com/github/ambry/frontend/ContainerMetrics.java
```diff
@@ -81,6 +85,8 @@
         metricRegistry.counter(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "ForbiddenCount"));
     notFoundCount = metricRegistry.counter(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "NotFoundCount"));
     goneCount = metricRegistry.counter(MetricRegistry.name(ContainerMetrics.class, metricPrefix + "GoneCount"));
+    String qpsMetricPrefix = accountName + SEPARATOR + containerName + SEPARATOR + (isGetRequest ? "GetRequestQps" : "PutRequestQps");
```
justinlin-linkedin, line 88: same.

justinlin-linkedin (commented):
LGTM

ambry-api/src/main/java/com/github/ambry/frontend/ContainerMetrics.java
SophieGuo410: Sure, updated.

ambry-api/src/main/java/com/github/ambry/frontend/ContainerMetrics.java
SophieGuo410: Updated.

justinlin-linkedin (approved):

cgtz (approved):

PR #2072, opened 2022-03-10
In cases where OperationTracker fails operation due to not found, differentiate it with unavailable error if some eligible replicas could have been offline during that time.

ankagrawal:
This PR introduces attempts to send a retryable Unavailable error for operations, in cases where Ambry frontend currently sends NotFound error, and there is a chance that offline replicas can make the operation successful.
The logic is guarded by a RouterConfig `routerUnavailableDueToOfflineReplicas` to try it out experimentally.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
@@ -309,6 +309,10 @@ class DeleteOperation {
     if (operationTracker.isDone() || operationCompleted) {
       if (operationTracker.hasSucceeded()) {
         operationException.set(null);
+      } else if(operationTracker.maybeFailedDueToOfflineReplicas()) {
+        operationException.set(
+            new RouterException("DeleteOperation failed possibly because some replicas are unavailable",
+                RouterErrorCode.AmbryUnavailable));
       } else if (operationTracker.hasFailedOnNotFound()) {
         operationException.set(
             new RouterException("DeleteOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
--- a/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
@@ -309,7 +309,7 @@ class DeleteOperation {
     if (operationTracker.isDone() || operationCompleted) {
       if (operationTracker.hasSucceeded()) {
         operationException.set(null);
-      } else if(operationTracker.maybeFailedDueToOfflineReplicas()) {
+      } else if (operationTracker.maybeFailedDueToOfflineReplicas()) {
         operationException.set(
             new RouterException("DeleteOperation failed possibly because some replicas are unavailable",
                 RouterErrorCode.AmbryUnavailable));
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -80,12 +80,12 @@ class SimpleOperationTracker implements OperationTracker {
   protected final int originatingDcNotFoundFailureThreshold;
   protected final int totalReplicaCount;
   protected final LinkedList<ReplicaId> replicaPool = new LinkedList<>();
+  protected final NonBlockingRouterMetrics routerMetrics;
   private final OpTrackerIterator otIterator;
   private final RouterOperation routerOperation;
   private final PartitionId partitionId;
   private final RouterConfig routerConfig;
   private final boolean crossColoEnabled;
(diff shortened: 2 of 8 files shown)


ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -357,24 +370,58 @@ public boolean hasSucceeded() {
...
+  @Override
+  public boolean maybeFailedDueToOfflineReplicas() {
+    if(!routerConfig.routerUnavailableDueToOfflineReplicas) {
+      return false;
+    }
+    // We mark the failure as due to offline replicas when we know that we couldn't find the blob in eligible replicas,
+    // and remaining replicas are enough to make the request successful, and there is atleast one offline replica in the
+    // remaining set of replicas. The offline replicas can come back up in future to make the request successful, and
+    // hence such a failure should be deemed as retryable.
+    if (hasFailedOnOriginatingDcNotFound()
```
justinlin-linkedin, line 382: I don't understand the logic here, Let's say we are dealing with originating dc, we have 3 replicas and 1 is offline, the other two returns NOT_FOUND and SUCCESS. It's basically what happened in the TTL_UPDATE getting 404. In this case, we want to return AmbryUnavailable so this method should return true. But, in this case, `hasFailedOnOriginatingDcNotFound` would return false. so we are not going to return true?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -149,6 +153,10 @@
     cloudReplicaParallelism = routerConfig.routerCloudRequestParallelism;
     List<ReplicaId> eligibleReplicas;
     List<ReplicaId> offlineReplicas = new ArrayList<>();
+    totalOfflineReplicaCount = getReplicasByState(partitionId, null, EnumSet.of(ReplicaState.OFFLINE)).values().size();
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
ankagrawal: In your case, `hasFailedOnOriginatingDcNotFound` will return false, but logic will try for cross-colo replicas as well, because for ttl-updates and deletes, cross-colo is enabled. And then it will eventually hit the second case for cross-colo not found.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -149,6 +153,10 @@
...
+    totalOfflineReplicaCount = getReplicasByState(partitionId, null, EnumSet.of(ReplicaState.OFFLINE)).values().size();
```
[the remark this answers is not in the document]
ankagrawal: fixed.

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
```diff
@@ -439,6 +439,9 @@ private void checkAndMaybeComplete() {
     if (progressTracker.isDone()) {
       if (progressTracker.hasSucceeded()) {
         operationException.set(null);
+      } else if (operationTracker.maybeFailedDueToOfflineReplicas()) {
```
cgtz, line 442: Why are we using this logic for GetBlobInfo/GetBlob operations too?

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -309,6 +309,10 @@ private void checkAndMaybeComplete() {
     if (operationTracker.isDone() || operationCompleted) {
       if (operationTracker.hasSucceeded()) {
         operationException.set(null);
+      } else if(operationTracker.maybeFailedDueToOfflineReplicas()) {
```
cgtz, line 312: formatting

cgtz (changes requested):

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
ankagrawal: removed.

cgtz (approved):

PR #2075, opened 2022-03-11
Dont create a crptoAgent when encrytion is disabled

justinlin-linkedin:
Don't create cryptoagent in vcr is the `vcrRequireEncryption` is false.


JingQianCloud (commented):
LGTM

cgtz:
One edge case is where some blobs in the cloud were previously encrypted with VCR encryption and require the crypto agent to decrypt them, even if vcrRequireEncryption is currently turned off. I don't think LI personally has this edge case, but it is something to keep in mind regarding whether `vcrRequireEncryption` config has caveats regarding if encryption was used in the past.

PR #2076, opened 2022-03-14
Introduce AmbryCloudRequests class for handling requests to cloud store

Arun-LinkedIn:
This change introduces new implementation of 'RequestsAPI' to handle requests from router to cloud store asynchronously. This implementation will be used in [CloudRouterFactory](https://github.com/linkedin/ambry/blob/master/ambry-router/src/main/java/com/github/ambry/router/CloudRouterFactory.java#L186).

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/store/MessageWriteSet.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/MessageWriteSet.java
@@ -34,8 +34,4 @@ public interface MessageWriteSet {
    * @return The list of message info about the message set
    */
   List<MessageInfo> getMessageSetInfo();
-
-  /**
-   * Set life Version for {@link StoreKey}.
-   */
 }
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/AmbryCloudRequests.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/AmbryCloudRequests.java
@@ -43,7 +43,6 @@ import com.github.ambry.server.ServerErrorCode;
 import com.github.ambry.server.StoreManager;
 import com.github.ambry.store.IdUndeletedStoreException;
 import com.github.ambry.store.MessageInfo;
-import com.github.ambry.store.Store;
 import com.github.ambry.store.StoreErrorCodes;
 import com.github.ambry.store.StoreException;
 import com.github.ambry.store.StoreGetOptions;
@@ -63,12 +62,11 @@ import org.slf4j.LoggerFactory;
  */
 public class AmbryCloudRequests extends AmbryRequests {
 
-  private static final Logger logger = LoggerFactory.getLogger(AmbryRequests.class);
+  private static final Logger logger = LoggerFactory.getLogger(AmbryCloudRequests.class);
 
   /**
    * @param storeManager {@link StoreManager} object to get stores for replicas
-   * @param requestResponseChannel the {@link RequestResponseChannel} to receive original ambry request and send
-   *                               original ambry response.
+   * @param requestResponseChannel the {@link RequestResponseChannel} to receive requests and send responses.
    * @param clusterMap the {@link ClusterMap} of the cluster
    * @param nodeId the data node Id.
    * @param registry the {@link MetricRegistry}
@@ -79,7 +77,12 @@ public class AmbryCloudRequests extends AmbryRequests {
   }
 
   @Override
-  public void handlePutRequest(NetworkRequest request) throws IOException, InterruptedException {
+  public void handlePutRequest(NetworkRequest request) throws InterruptedException {
+
+    if (!(request instanceof LocalRequestResponseChannel.LocalChannelRequest)) {
+      throw new IllegalArgumentException("The request must be of LocalChannelRequest type");
+    }
+
     PutRequest receivedRequest;
 
     // This is a case where handlePutRequest is called when frontends are writing to Azure. In this case, this method
@@ -111,7 +114,8 @@ public class AmbryCloudRequests extends AmbryRequests {
         MessageFormatWriteSet writeSet = getMessageFormatWriteSet(receivedRequest);
         CloudBlobStore cloudBlobStore =
             (CloudBlobStore) storeManager.getStore(receivedRequest.getBlobId().getPartition());
-        //TODO: Use executor instead of default thread pool
+        //TODO: Pass executor to have async completion stages run on configured thread pool instead of letting them
+        // run on default ForkJoinPool
         cloudBlobStore.putAsync(writeSet).whenCompleteAsync((unused, throwable) -> {
           if (throwable != null) {
             Exception ex = Utils.extractFutureExceptionCause(throwable);
@@ -133,8 +137,8 @@ public class AmbryCloudRequests extends AmbryRequests {
 
           try {
             requestResponseChannel.sendResponse(response.get(), request, null);
-          } catch (InterruptedException ignored) {
-
+          } catch (InterruptedException ie) {
+            logger.warn("Interrupted while enqueuing the response", ie);
           }
         });
       }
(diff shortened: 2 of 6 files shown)


ambry-api/src/main/java/com/github/ambry/store/Store.java
```diff
@@ -40,6 +42,18 @@
...
   StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> storeGetOptions) throws StoreException;
 
+  /**
+   * Returns the store info for the given ids asynchronously
+   * @param ids The list of ids whose messages need to be retrieved
+   * @param storeGetOptions A set of additional options that the store needs to use while getting the message
+   * @return a {@link CompletableFuture} that will eventually contain the {@link StoreInfo} for the given ids or the
+   *         {@link StoreException} if an error occurred
+   */
+  default CompletableFuture<StoreInfo> getAsync(List<? extends StoreKey> ids,
```
justinlin-linkedin, line 52: I realized that all the async operations are for AmbryCloudRequests, we probably don't have to change the interface, we can just cast the Store to CloudStore and only implement these async operations in CloudStore.

Later if we want to do async operations in ambry-server, we can come back and change the interface.

Arun-LinkedIn: Sure. Updated to define the asynchronous methods only in CloubBlobStore and down cast the class when using it in AmbryCloudRequests.

ambry-api/src/main/java/com/github/ambry/store/MessageWriteSet.java
```diff
@@ -36,6 +38,12 @@ public interface MessageWriteSet {
...
   /**
-   * Set life Version for {@link StoreKey}.
+   * Write the messages in this set to the given write channel asynchronously
+   * @param writeChannel The write interface to write the messages to
+   * @return a {@link CompletableFuture} that will eventually contain the size in bytes that was written to the
+   *         write interface or the {@link StoreException} if an error occurred
    */
-}
\ No newline at end of file
+  default CompletableFuture<Long> writeAsyncTo(Write writeChannel) {
```
ankagrawal, line 46: Will there be a case when both `writeTo` and `writeAsyncTo` should be present in the same class? If not I would suggest inherit MessageWriteSet into `SyncMessageWriteSet` and `AsyncMessageWriteSet`. The sync interface should have method `writeTo` and async should have method `writeAsyncTo`.

ambry-api/src/main/java/com/github/ambry/store/Write.java
```diff
@@ -39,4 +41,27 @@ public interface Write {
...
   void appendFrom(ReadableByteChannel channel, long size) throws StoreException;
+
+  /**
+   * Appends the buffer into the underlying write interface (eg: file) asynchronously. Returns the number of bytes
+   * successfully written
+   * @param buffer The buffer from which data needs to be written from
+   * @return a {@link CompletableFuture} that will eventually contain the number of bytes written to the write interface
+   *         or the {@link StoreException} if an error occurred
+   */
+  default CompletableFuture<Integer> appendAsyncFrom(ByteBuffer buffer) {
```
ankagrawal, line 52: Same comment as for `MessageWriteSet` above. If implementations don't need both sync and async, then maybe subclass this interface.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -186,6 +187,20 @@ public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> stor
...
   }
 
+  /**
+   * Returns the store info for the given ids asynchronously
+   * @param ids The list of ids whose messages need to be retrieved
+   * @param storeGetOptions A set of additional options that the store needs to use while getting the message
+   * @return a {@link CompletableFuture} that will eventually contain the {@link StoreInfo} for the given ids or the
+   *         {@link StoreException} if an error occurred
+   */
+  public CompletableFuture<StoreInfo> getAsync(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> storeGetOptions) {
```
ankagrawal, line 197: Can we have separate sync and asyn implementations for CloudBlobStore?

ambry-cloud/src/main/java/com/github/ambry/cloud/AmbryCloudRequests.java
```diff
@@ -0,0 +1,388 @@
...
+      if (error != ServerErrorCode.No_Error) {
+        logger.error("Validating put request failed with error {} for request {}", error, receivedRequest);
+        requestResponseChannel.sendResponse(
+            new PutResponse(receivedRequest.getCorrelationId(), receivedRequest.getClientId(), error), request, null);
+      } else {
+        MessageFormatWriteSet writeSet = getMessageFormatWriteSet(receivedRequest);
+        CloudBlobStore cloudBlobStore =
+            (CloudBlobStore) storeManager.getStore(receivedRequest.getBlobId().getPartition());
+        //TODO: Use executor instead of default thread pool
+        cloudBlobStore.putAsync(writeSet).whenCompleteAsync(((unused, throwable) -> {
```
justinlin-linkedin, line 115: there is an extra parenthesis in whenComplete method.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
Arun-LinkedIn: Since we are only adding async implementations for methods that handle requests (such as get, put, etc) and methods related to replication (findEntriesSince, etc) are still synchronous, I was just wondering if adding async methods in the existing CloudBlobStore implementation would be sufficient. What do you think?

ambry-api/src/main/java/com/github/ambry/store/MessageWriteSet.java
Arun-LinkedIn: Actually,  since asynchronous write method is only needed for cloud store, moved `writeAsyncTo` to `CloudMessageWriteSet` instead of having separate `AsyncMessageWriteSet`.

ambry-cloud/src/main/java/com/github/ambry/cloud/AmbryCloudRequests.java
```diff
@@ -0,0 +1,388 @@
...
+
+  @Override
+  public void handlePutRequest(NetworkRequest request) throws IOException, InterruptedException {
+    PutRequest receivedRequest;
+
+    // This is a case where handlePutRequest is called when frontends are writing to Azure. In this case, this method
+    // is called by request handler threads running within the frontend router itself. So, the request can be directly
+    // referenced as java objects without any need for deserialization.
+    PutRequest sentRequest =
+        (PutRequest) ((LocalRequestResponseChannel.LocalChannelRequest) request).getRequestInfo().getRequest();
```
justinlin-linkedin, line 89: this is all good, but let's have a check here to make sure the request is actually an instance of LocalChannelRequest, if not, we can return throw an IllegalArgumentException.

justinlin-linkedin (approved):
just some minor comments, the rest looks good to me

ambry-cloud/src/main/java/com/github/ambry/cloud/AmbryCloudRequests.java
Arun-LinkedIn: Done.

ambry-cloud/src/main/java/com/github/ambry/cloud/AmbryCloudRequests.java
```diff
@@ -0,0 +1,388 @@
...
+                  ServerErrorCode.Unknown_Error));
+            }
+          } else {
+            response.set(new PutResponse(receivedRequest.getCorrelationId(), receivedRequest.getClientId(),
+                ServerErrorCode.No_Error));
+          }
+
+          try {
+            requestResponseChannel.sendResponse(response.get(), request, null);
+          } catch (InterruptedException ignored) {
```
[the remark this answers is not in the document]
Arun-LinkedIn: Done

PR #2080, opened 2022-03-21
[BANDWIDTH_THROTTLING] Handle quota rejected response info in get operations.

ankagrawal:
This PR makes the changes to handle the quota rejected response from QuotaAwareOperationController in GetBlobOperation and GetBlobInfoOperation.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetOperation.java
@@ -195,14 +195,16 @@ abstract class GetOperation {
         return 3;
       case RangeNotSatisfiable:
         return 4;
-      case AmbryUnavailable:
+      case TooManyRequests:
         return 5;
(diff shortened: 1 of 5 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetOperation.java
```diff
@@ -203,6 +203,8 @@ protected int getPrecedenceLevel(RouterErrorCode routerErrorCode) {
         return 7;
       case BlobDoesNotExist:
         return 8;
+      case TooManyRequests:
+        return 9;
```
justinlin-linkedin, line 207: TooManyRequest should have precedence over other errors. If we have two errors, first one is BlobDoesNotExist, second one if TooManyRequest, then we should return TooManyRequest.

justinlin-linkedin (approved):

tbingol (approved):

PR #2082, opened 2022-03-23
Improve logic for termination of blob operations when they are not found in originating DC

Arun-LinkedIn:
Make sure there are at least 'PutSuccessTarget' replicas in leader or standby state in originating data center before terminating the operation with blob-not-found error.

Also, have a negative cache to store recently not-found blob IDs to avoid querying the servers again.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -32,6 +32,7 @@ public class RouterConfig {
   public static final String DEFAULT_CRYPTO_SERVICE_FACTORY = "com.github.ambry.router.GCMCryptoServiceFactory";
   public static final double DEFAULT_LATENCY_TOLERANCE_QUANTILE = 0.9;
   public static final long DEFAULT_OPERATION_TRACKER_HISTOGRAM_CACHE_TIMEOUT_MS = 1000L;
+  public static final long ROUTER_NOT_FOUND_CACHE_MAX_TTL_IN_MS = 24 * 60 * 1000L;
 
   // config keys
   public static final String ROUTER_SCALING_UNIT_COUNT = "router.scaling.unit.count";
@@ -112,6 +113,7 @@ public class RouterConfig {
   public static final String ROUTER_REQUEST_HANDLER_NUM_OF_THREADS = "router.request.handler.num.of.threads";
   public static final String ROUTER_STORE_KEY_CONVERTER_FACTORY = "router.store.key.converter.factory";
   public static final String ROUTER_UNAVAILABLE_DUE_TO_OFFLINE_REPLICAS = "router.unavailable.due.to.offline.replicas";
+  public static final String ROUTER_NOT_FOUND_CACHE_TTL_IN_MS = "router.not.found.cache.ttl.in.ms";
 
   /**
    * Number of independent scaling units for the router.
@@ -562,6 +564,14 @@ public class RouterConfig {
   public final boolean routerUnavailableDueToOfflineReplicas;
 
   /**
+   * Expiration time for Blob IDs stored in not-found cache. Default value is 15 seconds.
+   * Setting it to 0 would disable the cache and avoid storing any blob IDs.
+   */
+  @Config(ROUTER_NOT_FOUND_CACHE_TTL_IN_MS)
+  @Default("15*1000")
+  public final long routerNotFoundCacheTtlInMs;
+
+  /**
    * Create a RouterConfig instance.
    * @param verifiableProperties the properties map to refer to.
    */
@@ -682,5 +692,7 @@ public class RouterConfig {
         "com.github.ambry.store.StoreKeyConverterFactoryImpl");
     routerUnavailableDueToOfflineReplicas =
         verifiableProperties.getBoolean(ROUTER_UNAVAILABLE_DUE_TO_OFFLINE_REPLICAS, false);
+    routerNotFoundCacheTtlInMs = verifiableProperties.getLongInRange(ROUTER_NOT_FOUND_CACHE_TTL_IN_MS, 15 * 1000L, 0,
+        ROUTER_NOT_FOUND_CACHE_MAX_TTL_IN_MS);
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -566,6 +566,13 @@ public class RouterConfig {
   /**
    * Expiration time for Blob IDs stored in not-found cache. Default value is 15 seconds.
    * Setting it to 0 would disable the cache and avoid storing any blob IDs.
+   * TODO: With PR https://github.com/linkedin/ambry/pull/2072, when operation tracker fails due to blob-not-found and
+   *  some of eligible replicas are offline during the time of operation, we differentiate it with unavailable error and
+   *  and return 503 to client instead of 404. But we seem to do it only for 'ttl_update' & 'delete' but not for 'Get'.
+   *  When this cache is introduced, it is possible that blobs are cached for not-found on 'Get' and that could interfere
+   *  with above logic for 'TTL_Update' and 'Delete' as we would return 404 instead of 503. We might need to keep this
+   *  cache disabled (by setting it to 0 in configs) until we fix to return 503 for 'Get' calls as well when replicas are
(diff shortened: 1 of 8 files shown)


ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -60,6 +63,9 @@
   // Resources that need to be shut down when the router does.
   private final List<Closeable> resourcesToClose;
   private final AtomicInteger currentBackgroundOperationsCount = new AtomicInteger(0);
+  private static final int cacheMaxLimit = 1000;
+  // Cache to store blob IDs which were not found in servers recently.
+  private final Cache<String, Boolean> notFoundcache;
```
justinlin-linkedin, line 68: what is the motivation behind this cache? Are we seeing same blob ids are retried multiple times?

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -112,6 +113,8 @@
   public static final String ROUTER_REQUEST_HANDLER_NUM_OF_THREADS = "router.request.handler.num.of.threads";
   public static final String ROUTER_STORE_KEY_CONVERTER_FACTORY = "router.store.key.converter.factory";
   public static final String ROUTER_UNAVAILABLE_DUE_TO_OFFLINE_REPLICAS = "router.unavailable.due.to.offline.replicas";
+  public static final String ROUTER_CACHE_TTL_FOR_NOT_FOUND_RESPONSES_MS =
+      "router.cache.ttl.for.not.found.responses.ms";
```
justinlin-linkedin, line 117: I don't quite like the configuration name, it has nothing to do with response, can we rename it to router.not.found.cache.ttl.in.ms", something like that?

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
justinlin-linkedin, line 68: Also, let's rename the variable notFoundCache

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -202,13 +221,27 @@ private OperationController getOperationController() {
     routerMetrics.operationQueuingRate.mark();
     try {
       if (isOpen.get()) {
-        getOperationController().getBlob(blobIdStr, internalOptions, (internalResult, exception) -> {
-          GetBlobResult getBlobResult = internalResult == null ? null : internalResult.getBlobResult;
-          futureResult.done(getBlobResult, exception);
-          if (callback != null) {
-            callback.onCompletion(getBlobResult, exception);
+        if (notFoundcache.getIfPresent(blobIdStr) != null) {
```
justinlin-linkedin, line 224: let's put a log message here, so we know it's hitting the cache.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -202,13 +221,27 @@ private OperationController getOperationController() {
...
+          // If we know that blob doesn't exist, complete the operation.
```
justinlin-linkedin, line 225: also, i think we might want to have a metric here.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -314,12 +347,19 @@ private OperationController getOperationController() {
...
     FutureResult<Void> futureResult = new FutureResult<>();
     if (isOpen.get()) {
-      // Can skip attemptChunkDeletes if we can determine this is not a metadata blob
-      boolean attemptChunkDeletes = isMaybeMetadataBlob(blobId);
-      getOperationController().deleteBlob(blobId, serviceId, futureResult, callback, attemptChunkDeletes,
-          quotaChargeCallback);
-      if (!attemptChunkDeletes) {
-        routerMetrics.skippedGetBlobCount.inc();
+      if (notFoundcache.getIfPresent(blobId) != null) {
+        // If we know that blob doesn't exist, complete the operation
```
justinlin-linkedin, line 351: same above.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -516,8 +516,13 @@ public ReplicaId next() {
...
   private boolean hasFailedOnOriginatingDcNotFound() {
-    return originatingDcNotFoundFailureThreshold > 0
-        && originatingDcNotFoundCount >= originatingDcNotFoundFailureThreshold;
+    // Check that there are at least routerConfig.routerPutSuccessTarget number of replicas in leader or standby state
+    // in originating data center before failing the request. Also, in case originating data center no longer exists,
+    // avoid checking for this condition.
+    return originatingDcNotFoundFailureThreshold > 0 && (originatingDcNotFoundCount
+        >= originatingDcNotFoundFailureThreshold) && (this.originatingDcName != null
+        && getEligibleReplicas(this.originatingDcName, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER)).size()
+        >= routerConfig.routerPutSuccessTarget);
```
justinlin-linkedin, line 525: why it's put target? this is used mostly in get? we can make this change in the SimpleOperationTracker's constructor instead of in this method. So in the constructor we query for the number of eligible replicas in the originating dc, and if it's less than the target, we change the originatingDcNotFoundFailureThreshold to 0, so we disable this feature.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -561,6 +564,14 @@
   @Default("false")
   public final boolean routerUnavailableDueToOfflineReplicas;
 
+  /**
+   * Expiration time for Blob IDs stored in not-found cache. Default value is 60 seconds.
+   * Setting it to 0 would disable the cache and avoid storing any blob IDs.
+   */
+  @Config(ROUTER_CACHE_TTL_FOR_NOT_FOUND_RESPONSES_MS)
+  @Default("60*1000")
```
tbingol, line 572: can we use a smaller TTL, maybe 15 secs?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
tbingol, line 525: can you decompose this logic into more temp variables to improve the readability?

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -579,5 +634,40 @@ void incrementOperationsCount(int delta) {
...
+     */
+    public BlobOperationCallbackWrapper(String blobId, Callback<T> callback) {
+      this.blobId = blobId;
+      this.callback = callback;
+    }
+
+    @Override
+    public void onCompletion(T result, Exception exception) {
+      if (exception != null) {
+        if (exception instanceof RouterException && ((RouterException) exception).getErrorCode()
```
tbingol, line 661: Nested if is not needed. Append this condition to above if

tbingol (approved):

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
Arun-LinkedIn: Using the `getEligibleReplicas(originatingDcName, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER)).size()
 >= routerConfig.routerPutSuccessTarget)` to make sure that there is at least one or more replicas from the original PUT quorum in leader or standby state. I was wondering that when we receive not-found response from all replicas, this might help us to be sure that blob is missing. 

Sure. Moved the logic to constructor as suggested.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
Arun-LinkedIn: Negative cache was suggested by @tbingol.  Since we are making more strict checks for terminating the operations based on local not-found responses, we might have more cross colo look ups. Using not-found or negative cache might help us reduce the load on servers if clients were to do immediate look-ups again.

cgtz (approved):

ankagrawal (approved):

PR #2083, opened 2022-03-23
Only all numeric resource names to be synced to vcr cluster

justinlin-linkedin:
resource names are all numeric values in helix cluster, we should only allow those resources to be synced to vcr cluster.

### DIFF
--- a/ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
+++ b/ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
@@ -13,10 +13,8 @@
  */
 package com.github.ambry.clustermap;
 
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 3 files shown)


ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
```diff
@@ -43,10 +41,34 @@
...
+    }
+  }
+
+  /**
+   * Return {@code true} is the given resource name is a resource that contains partition information.
+   * @param resourceName The resource name.
+   * @return {@code true} when the given resource contains partition inforrmation
+   */
+  public static boolean isPartitionResourceName(String resourceName) {
+    // all the paritions are placed under resource named as integer values.
```
cgtz, line 61: nit: paritions -> partitions

The purpose of smallIntegers is to create a fast path to avoid needing to parse the string as an int in most cases?

cgtz (approved):
lgtm!

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
```diff
@@ -43,10 +41,34 @@
...
+  private static final Set<String> smallIntegers = new HashSet<>();
+
+  static {
+    for (int i = 0; i < 128; i++) {
+      smallIntegers.add(String.valueOf(i));
+    }
+  }
+
+  /**
+   * Return {@code true} is the given resource name is a resource that contains partition information.
```
ankagrawal, line 56: nit: "if" the given resource name

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
```diff
@@ -43,10 +41,34 @@
...
+   * @param resourceName The resource name.
+   * @return {@code true} when the given resource contains partition inforrmation
```
ankagrawal, line 58: nit: information

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
```diff
@@ -43,10 +41,34 @@
...
+
+  static {
+    for (int i = 0; i < 128; i++) {
+      smallIntegers.add(String.valueOf(i));
+    }
+  }
+
+  /**
+   * Return {@code true} is the given resource name is a resource that contains partition information.
+   * @param resourceName The resource name.
```
ankagrawal, line 57: Please add a comment here saying something like "A resource is considered to be a partition resource only if the resource id is an integer.

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
```diff
@@ -43,10 +41,34 @@
...
+   * @return {@code true} when the given resource contains partition inforrmation
+   */
+  public static boolean isPartitionResourceName(String resourceName) {
```
ankagrawal, line 60: I know this is out of scope for this commit, but should we do something to ensure that any new/old non-partition resources created by Ambry in helix are not integer?

ankagrawal (commented):
looks good to me overall. few minor comments.

ambry-commons/src/main/java/com/github/ambry/clustermap/HelixVcrUtil.java
justinlin-linkedin: yes, I only check one cluster and it has 117 resources, so I set the value to be 128, but I checked the other two clusters, they are both more than 128 and less than 512, I will set the maximum size to 512, this would cover all of them.

ankagrawal (approved):

PR #2085, opened 2022-03-24
[BANDWIDTH_THROTTLING] Ensure that PreProcessQuotaChargeCallback doesn't return DELAY or REJECT recommendation, when QuotaMode is TRACKING.

ankagrawal:
Returning DELAY or REJECT from PreProcessQuotaChargeCallback will lead to enforcement of quotas in QuotaAwareOperationController. This PR ensure that when QuotaMode is Tracking, the usage is tracked, but PreProcessQuotaChargeCallback returns ALLOW recommendation in all cases.

### DIFF
--- a/ambry-quota/src/main/java/com/github/ambry/quota/PreProcessQuotaChargeCallback.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/PreProcessQuotaChargeCallback.java
@@ -18,8 +18,6 @@ import com.github.ambry.rest.RestRequest;
 import com.github.ambry.router.RouterErrorCode;
 import java.util.Map;
 import java.util.stream.Collectors;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 
 /**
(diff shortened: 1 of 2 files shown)


ambry-quota/src/main/java/com/github/ambry/quota/PreProcessQuotaChargeCallback.java
```diff
@@ -54,7 +52,12 @@ public QuotaAction checkAndCharge(boolean shouldCheckQuotaExceedAllowed, boolean
         .entrySet()
         .stream()
         .collect(Collectors.toMap(entry -> QuotaName.valueOf(entry.getKey()), Map.Entry::getValue));
-    return quotaManager.chargeAndRecommend(restRequest, requestCost, shouldCheckQuotaExceedAllowed, forceCharge);
+    QuotaAction quotaAction =
+        quotaManager.chargeAndRecommend(restRequest, requestCost, shouldCheckQuotaExceedAllowed, forceCharge);
+    // Note that we track quota usages even if QuotaMode is TRACKING, so although we return ALLOW, checkAndCharge still
+    // needs to happen
+    quotaAction = quotaManager.getQuotaMode() == QuotaMode.TRACKING ? QuotaAction.ALLOW : quotaAction;
```
justinlin-linkedin, line 59: if we want to keep PostPorcessQuotaChargeCallback implementation, then we should probably move this logic to AmbryQuotaManager.

justinlin-linkedin (approved):

ambry-quota/src/main/java/com/github/ambry/quota/PreProcessQuotaChargeCallback.java
tbingol, line 59: Nit: return directly

tbingol: +1

tbingol (approved):

ambry-quota/src/main/java/com/github/ambry/quota/PreProcessQuotaChargeCallback.java
ankagrawal: I will refactor it as a cleanup effort. We will eventually get rid of PostProcessQuotaChargeCallback (and maybe the entire concept of callback class).

ankagrawal: fixed.

PR #2088, opened 2022-04-05
[CI] Remove usage of GIT_SECRET

cgtz:
Github actions sets up push permissions by default and does not require
supplying a personal access token anymore. This commit simplifies the
code to push a new release tag.

---

I followed the [example](https://github.com/actions/checkout#push-a-commit-using-the-built-in-token) on the checkout action readme

Tested this by running a similar [workflow](https://github.com/cgtz/actions-debug/runs/5842159040?check_suite_focus=true) in a debugging repo.

PR #2089, opened 2022-04-06
Split string on spaces when using 'commandLine' option in Gradle Executable tasks

Arun-LinkedIn:
Fixing typo in PR https://github.com/linkedin/ambry/pull/2088 to split string on spaces in the gradle Exec task.

For example, exec { commandLine "git" "push" "--tags" } seem to be executed as as 'gitpush --tags' instead of 'git push --tags' (https://github.com/linkedin/ambry/runs/5843222047?check_suite_focus=true#step:6:444).

### DIFF
--- a/gradle/ci-release.gradle
+++ b/gradle/ci-release.gradle
@@ -33,7 +33,7 @@ task gitPush {
 
     doLast {
       println "Pushing tag v$version"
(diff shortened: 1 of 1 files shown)


cgtz (approved):
Thanks for finding/fixing this. My bad for not testing more

PR #2090, opened 2022-04-07
Add Gradle plugin for retrying flaky tests

Arun-LinkedIn:
Reference: https://github.com/gradle/test-retry-gradle-plugin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,6 +33,7 @@ allprojects {
     apply plugin: 'idea'
     apply plugin: 'project-report'
     apply plugin: 'jacoco'
+    apply plugin: 'org.gradle.test-retry'
 
     repositories {
         mavenCentral()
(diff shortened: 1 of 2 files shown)


cgtz (approved):
I think this looks good and should improve our build stability. 
A helpful task for the future would be if you could come up with a process for auditing/discovering frequently failing tests, since this plugin can mask flaky tests if you don't look at the build scan or logs often.

build.gradle
```diff
@@ -115,6 +116,12 @@ subprojects {
         testLogging {
             exceptionFormat = 'full'
         }
+        retry {
+            // Allow for retrying flaky tests.
+            maxRetries = 3
+            maxFailures = 20
```
ankagrawal, line 122: Does this mean total number of tests failed during the build across all modules? Maybe it might be a good idea to comment that.

ankagrawal (approved):
looks good overall. Please consider adding the comment if it makes sense to you.

PR #2091, opened 2022-04-07
Support max ttl for chunk upload

SophieGuo410:
This PR support:
1. Treat chunkUploadInitialChunkTtlSecs as the default if the customer did not specify ttl
2. If the customer did specify x-ambry-ttl header, check that it is less than chunkUploadMaxChunkTtlSecs instead of chunkUploadInitialChunkTtlSecs

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.UUID;
 
 
+
 /**
  * Default implementation of {@link UrlSigningService} that currently only converts all headers that start with
  * "x-ambry-" into query parameters and does not actually sign the URL.
@@ -139,7 +140,9 @@ public class AmbryUrlSigningService implements UrlSigningService {
         maxUploadSize = chunkUploadMaxChunkSize;
         // They also have a non-optional blob TTL to ensure that chunks that were not stitched within a reasonable time
         // span are cleaned up.
-        argsForUrl.put(RestUtils.Headers.TTL, chunkUploadInitialChunkTtlSecs);
+        if (RestUtils.getLongHeader(args, RestUtils.Headers.TTL, false) == null) {
+          argsForUrl.put(RestUtils.Headers.TTL, chunkUploadInitialChunkTtlSecs);
+        }
         argsForUrl.put(RestUtils.Headers.CHUNK_UPLOAD, true);
         argsForUrl.put(RestUtils.Headers.SESSION, UUID.randomUUID().toString());
       }
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
@@ -265,12 +265,17 @@ class PostBlobHandler {
           frontendMetrics.postBlobMetricsGroup);
       BlobProperties blobProperties = RestUtils.buildBlobProperties(restRequest.getArgs());
       Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
+      Long chunkUploadTtl =
+          RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.CHUNK_UPLOAD_TTL, false);
       if (blobProperties.getTimeToLiveInSeconds() + TimeUnit.MILLISECONDS.toSeconds(
           blobProperties.getCreationTimeInMs()) > Integer.MAX_VALUE) {
         LOGGER.debug("TTL set to very large value in POST request with BlobProperties {}", blobProperties);
         frontendMetrics.ttlTooLargeError.inc();
       } else if (container.isTtlRequired() && (blobProperties.getTimeToLiveInSeconds() == Utils.Infinite_Time
-          || blobProperties.getTimeToLiveInSeconds() > frontendConfig.maxAcceptableTtlSecsIfTtlRequired)) {
+          || chunkUploadTtl == null
+          && blobProperties.getTimeToLiveInSeconds() > frontendConfig.maxAcceptableTtlSecsIfTtlRequired
+          || chunkUploadTtl != null
+          && blobProperties.getTimeToLiveInSeconds() > frontendConfig.chunkUploadMaxChunkTtlSecs)) {
         String descriptor = RestUtils.getAccountFromArgs(restRequest.getArgs()).getName() + ":" + container.getName();
         if (frontendConfig.failIfTtlRequiredButNotProvided) {
           throw new RestServiceException(
@@ -339,9 +344,12 @@ class PostBlobHandler {
         RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.SESSION, true);
         // validate that a max chunk size is set.
         RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.MAX_UPLOAD_SIZE, true);
+        Long chunkUploadTtl =
+            RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.CHUNK_UPLOAD_TTL, false);
         // validate that the TTL for the chunk is set correctly.
         long chunkTtl = blobProperties.getTimeToLiveInSeconds();
-        if (chunkTtl <= 0 || chunkTtl > frontendConfig.chunkUploadInitialChunkTtlSecs) {
+        if (chunkTtl <= 0 || chunkUploadTtl == null && chunkTtl > frontendConfig.chunkUploadInitialChunkTtlSecs
+            || chunkUploadTtl != null && chunkUploadTtl > frontendConfig.chunkUploadMaxChunkTtlSecs) {
           throw new RestServiceException("Invalid chunk upload TTL: " + chunkTtl, RestServiceErrorCode.InvalidArgs);
         }
       }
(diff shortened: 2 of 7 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -265,12 +265,17 @@ private BlobInfo getBlobInfoFromRequest() throws RestServiceException {
...
       Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
+      Long chunkUploadTtl =
+          RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.CHUNK_UPLOAD_TTL, false);
       if (blobProperties.getTimeToLiveInSeconds() + TimeUnit.MILLISECONDS.toSeconds(
           blobProperties.getCreationTimeInMs()) > Integer.MAX_VALUE) {
         LOGGER.debug("TTL set to very large value in POST request with BlobProperties {}", blobProperties);
         frontendMetrics.ttlTooLargeError.inc();
       } else if (container.isTtlRequired() && (blobProperties.getTimeToLiveInSeconds() == Utils.Infinite_Time
-          || blobProperties.getTimeToLiveInSeconds() > frontendConfig.maxAcceptableTtlSecsIfTtlRequired)) {
+          || chunkUploadTtl == null
```
tbingol, line 275: Nit: Would be good to reconcile the max_ttl first. Something like:

final long max_ttl = chunkUploadTtl != null ? frontendConfig.chunkUploadMaxChunkTtlSecs : frontendConfig.maxAcceptableTtlSecsIfTtlRequired;
.
.
if (... blobProperties.getTimeToLiveInSeconds() > max_ttl) {

}

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -339,9 +344,12 @@ private void checkUploadRequirements(BlobProperties blobProperties) throws RestS
         RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.SESSION, true);
         // validate that a max chunk size is set.
         RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.MAX_UPLOAD_SIZE, true);
+        Long chunkUploadTtl =
+            RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.CHUNK_UPLOAD_TTL, false);
         // validate that the TTL for the chunk is set correctly.
         long chunkTtl = blobProperties.getTimeToLiveInSeconds();
-        if (chunkTtl <= 0 || chunkTtl > frontendConfig.chunkUploadInitialChunkTtlSecs) {
+        if (chunkTtl <= 0 || chunkUploadTtl == null && chunkTtl > frontendConfig.chunkUploadInitialChunkTtlSecs
```
tbingol, line 351: same as above

tbingol (approved):

cgtz (commented):
Hey, so i'm thinking if we can limit the change to only touch the signed URL generation code and keep using only x-ambry-ttl header in PostBlobHandler instead of add x-ambry-chunk-upload-ttl.

Currently all data chunk uploads originate with a signed URL, so couldn't the UrlSigningService impl validate this in the getSignedUrl call and fail it if it isn't less than chunkUploadMaxChunkTtlSecs?

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -265,12 +265,17 @@ private BlobInfo getBlobInfoFromRequest() throws RestServiceException {
...
         LOGGER.debug("TTL set to very large value in POST request with BlobProperties {}", blobProperties);
         frontendMetrics.ttlTooLargeError.inc();
       } else if (container.isTtlRequired() && (blobProperties.getTimeToLiveInSeconds() == Utils.Infinite_Time
-          || blobProperties.getTimeToLiveInSeconds() > frontendConfig.maxAcceptableTtlSecsIfTtlRequired)) {
+          || chunkUploadTtl == null
+          && blobProperties.getTimeToLiveInSeconds() > frontendConfig.maxAcceptableTtlSecsIfTtlRequired
+          || chunkUploadTtl != null
+          && blobProperties.getTimeToLiveInSeconds() > frontendConfig.chunkUploadMaxChunkTtlSecs)) {
         String descriptor = RestUtils.getAccountFromArgs(restRequest.getArgs()).getName() + ":" + container.getName();
         if (frontendConfig.failIfTtlRequiredButNotProvided) {
```
cgtz, line 280: This config is not turned yet and was intended for another purpose. I think conflating this with the chunk TTL restrictions may be a bad idea.

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -339,9 +344,12 @@ private void checkUploadRequirements(BlobProperties blobProperties) throws RestS
         RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.SESSION, true);
         // validate that a max chunk size is set.
         RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.MAX_UPLOAD_SIZE, true);
+        Long chunkUploadTtl =
+            RestUtils.getLongHeader(restRequest.getArgs(), RestUtils.Headers.CHUNK_UPLOAD_TTL, false);
         // validate that the TTL for the chunk is set correctly.
         long chunkTtl = blobProperties.getTimeToLiveInSeconds();
-        if (chunkTtl <= 0 || chunkTtl > frontendConfig.chunkUploadInitialChunkTtlSecs) {
```
cgtz, line 344: If you make the change to the url signing service impls I described, I think you can just change this part of the code to see if chunkTtl <= frontendConfig.chunkUploadMaxChunkTtlSecs and get rid of the less than initialTtl clause.

cgtz (changes requested):

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
Arun-LinkedIn: Like Casey suggested, I was also wondering if making the change in url signing service impls is a good idea since chunk uploads seems to originate only with signed URL.

Also, just a question. Apart from extending the maximum TTL of chunk upload to 90 days, are we also trying to provide option for user to specify the TTL value (between 0 to 90 days). 

1. If yes, based on this [code](https://github.com/linkedin/ambry/blob/97623533690e6e0986265f395162dfe224d2862f/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java#L142) , it looks like we are currently always setting the TTL value in the signed URL to the config value `frontendConfig.chunkUploadInitialChunkTtlSecs` (with a default value of 28 days). Would it require modifying the code there too to instead take user provided value (provided via in x-ambry-ttl header) and set it in the signed URL?
2. If not, it looks like if we modify the config` frontendConfig.chunkUploadInitialChunkTtlSecs` to 90 days, it might satisfy user requirement. But, it might increase our storage space temporarily until the chunks are stitched and made permanent?

Also another question - do we need to validate the value of TTL in `checkUploadRequirements` for chunked requests since this TTL value is coming from signed URL and might already be checked during generation of the URL?

Arun-LinkedIn: Sorry, I was looking at OSS code above. In AmbryLI code, looks like we choose user TTL during generation of signed URLs for chunked uploads. Then just wondering that if we increase the config value `frontendConfig.chunkUploadInitialChunkTtlSecs` to 90 days, would it be sufficient to handle the requirement?

SophieGuo410: > If you make the change to the url signing service impls I described, I think you can just change this part of the code to see if chunkTtl <= frontendConfig.chunkUploadMaxChunkTtlSecs and get rid of the less than initialTtl clause.
I was thinking that constraint our client to use 28days ttl by default and if they have special cases, we can ask them to provide extra header if they are sure it won't affect GDPR compliance. If I understand your ideal correctly, seems that we won't gate them to have ttl larger than 28 days by default, then could we just increase the chunkUploadInitialChunkTtlSecs to 90 days?

Inside checkUploadRequirements, we only set the restriction when request isChunkUpload, are you suggesting instead of set the constraint at the upload blob stage, we should set it up at the getSignedUrl stage?

SophieGuo410: After discuss offline, we will use chunkUploadInitialChunkTtlSecs as the default ttl for chunkupload and allow user to specify longer ttl < 90 when they set x-ambry-ttl header.

cgtz (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
```diff
@@ -139,7 +140,9 @@ public String getSignedUrl(RestRequest restRequest) throws RestServiceException
         maxUploadSize = chunkUploadMaxChunkSize;
         // They also have a non-optional blob TTL to ensure that chunks that were not stitched within a reasonable time
         // span are cleaned up.
-        argsForUrl.put(RestUtils.Headers.TTL, chunkUploadInitialChunkTtlSecs);
+        if (RestUtils.getLongHeader(args, RestUtils.Headers.TTL, false) == null) {
+          argsForUrl.put(RestUtils.Headers.TTL, chunkUploadInitialChunkTtlSecs);
```
Arun-LinkedIn, line 144: Minor: Just wondering if we should have an `else` case to set the client provided TTL value in the URL arguments? Also, would it be helpful to check for Max TTL and throw error here itself (i.e. during generation of Signed URL) instead of doing it in `checkUploadRequirements` which seem to hit when client tries to upload the blob using the URL. It might probably let client know that his arguments were invalid quicker?

Just to note, it looks like in LI implementation of URLSigningService, above comments seem to be already handled. So, we might only need it for completeness in OSS code.

PR #2093, opened 2022-04-11
Take snapshot of all replicas of a partition only once in SimpleOperationTracker, so that there is no consistency issue due to any cluster map changes.

ankagrawal:
Take snapshot of all replicas of a partition only once in SimpleOperationTracker, so that there is no consistency issue due to any cluster map changes. Otherwise there is a chance of mismatch between multiple data structures of the OperationTracker if cluster map changes are concurrent with the progress of an operation in frontend.


tbingol (approved):
On a related note, operations in HelixClusterManager do not seem to be thread-safe. It is likely that you may retrieve inconsistent replicas.

SophieGuo410 (approved):

PR #2094, opened 2022-04-11
Add additional logging in not found error case of SimpleOperationController, to be able to better debug not found causes.

ankagrawal:


JingQianCloud (approved):
LBTM

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -419,8 +419,11 @@ public boolean hasFailedOnNotFound() {
     }
     if (hasFailedOnOriginatingDcNotFound()) {
       logger.info(
-          "Terminating {} on {} due to Not_Found failure. OriginatingDC Not_Found count: {}, failure threshold: {}",
-          routerOperation, partitionId, originatingDcNotFoundCount, originatingDcNotFoundFailureThreshold);
+          "Terminating {} on {} due to Not_Found failure. Originating Not_Found count: {}, failure threshold: {},"
+              + "originatingDcOfflineReplicaCount: {}, originatingDcNameTotalReplicaCount: {},"
```
Arun-LinkedIn, line 423: nit: Looks like we might not need string concatenation here.

Arun-LinkedIn (approved):

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
ankagrawal: The issue is that if I remove string concatenation, then the entire log statement has to be in the same line.

There is a tradeoff here, between how the code looks, and maybe a little bit of performance. Let me know if you want me to remove the concatenation, and if you think its ok to make the statement longer.

PR #2096, opened 2022-04-15
Add missing quota charge callback for background delete operations. Add logs if QuotaChargeCallback is null in OperationQuotaCharger.

ankagrawal:
This PR adds the quota charge callback that was missing from background deletes. It also adds logging to highlight cases where quota method was null.


SophieGuo410 (approved):
It looks like if quotaChargeCallback is null, we will put UNKNOWN_QUOTA_RESOURCE into writeRequestQueue, in the future, I think we need to handle this situation carefully. We don't want a read request with quotaChargeCallback==null adding into writeRequestQueue right?

ankagrawal:
> It looks like if quotaChargeCallback is null, we will put UNKNOWN_QUOTA_RESOURCE into writeRequestQueue, in the future, I think we need to handle this situation carefully. We don't want a read request with quotaChargeCallback==null adding into writeRequestQueue right?

You are right. I will make a subsequent change to ensure that quotaChargeCallback can never be passed as null in the constructor. That will eliminate this issue altogether.

tbingol (approved):

PR #2100, opened 2022-04-26
Check for null responses during cosmos updates to avoid NPE

Arun-LinkedIn:
cosmosDataAccessor.updateMetadata() can return null response if there is nothing to update, i.e. fields already have specified value. Check for null response during metadata updates to avoid NPE.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
@@ -335,7 +335,7 @@ public class AzureBlobDataAccessor {
    * @throws IllegalStateException on request timeout.
    */
   public AzureCloudDestination.UpdateResponse updateBlobMetadata(BlobId blobId, Map<String, Object> updateFields,
-      CloudUpdateValidator cloudUpdateValidator) throws BlobStorageException {
+      CloudUpdateValidator cloudUpdateValidator) throws Exception {
     Objects.requireNonNull(blobId, "BlobId cannot be null");
     updateFields.keySet()
         .forEach(field -> Objects.requireNonNull(updateFields.get(field), String.format("%s cannot be null", field)));
@@ -349,6 +349,8 @@ public class AzureBlobDataAccessor {
         String etag = blobProperties.getETag();
         Map<String, String> metadata = blobProperties.getMetadata();
 
+        // Validate the sanity of update operation by comparing the fields we want to update against existing metadata
+        // fields in Azure. This can throw StoreExceptions which are propagated to caller.
         if (!cloudUpdateValidator.validateUpdate(CloudBlobMetadata.fromMap(metadata), blobId, updateFields)) {
           return new AzureCloudDestination.UpdateResponse(false, metadata);
         }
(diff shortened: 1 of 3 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -335,7 +335,7 @@ public CloudBlobMetadata getBlobMetadata(BlobId blobId) throws BlobStorageExcept
    * @throws IllegalStateException on request timeout.
    */
   public AzureCloudDestination.UpdateResponse updateBlobMetadata(BlobId blobId, Map<String, Object> updateFields,
-      CloudUpdateValidator cloudUpdateValidator) throws BlobStorageException {
+      CloudUpdateValidator cloudUpdateValidator) throws Exception {
```
ankagrawal, line 338: Why do we need this change? We are either throwing BlobStorageException here, or RunTimeException.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -399,7 +399,9 @@ private UpdateResponse updateBlobMetadata(BlobId blobId, Map<String, Object> upd
       boolean updatedCosmos;
       try {
         CosmosItemResponse<CloudBlobMetadata> response = cosmosDataAccessor.updateMetadata(blobId, metadataMap);
-        updatedCosmos = response.getItem() != null;
+        // cosmosDataAccessor.updateMetadata() can return null response if there is nothing to update, i.e. fields
+        // already have specified value. Check for null response to see if metadata was updated.
+        updatedCosmos = response != null;
```
ankagrawal, line 404: No integration test needed for this?

SophieGuo410, line 404: if response != null, response.getItem() can still be null If responseBodyAsByteArray is empty?

Arun-LinkedIn: I was able to reproduce this with existing integration tests  (like [this](https://github.com/linkedin/ambry/blob/master/ambry-cloud/src/test/java/com/github/ambry/cloud/CloudBlobStoreIntegrationTest.java#L229) )  where there is no changes in metadata fields to update. Verified that we are not hitting NPE after the fix.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
Arun-LinkedIn: Looks like we could also throw `StoreException` during validation of update method when we call `cloudUpdateValidator.validateUpdate` within this function. Previously, `StoreException` was also incorrectly wrapped in `RunTimeException` which seemed incorrect. Due to that, I had to modify the signature of the method to throw `StoreException` explicitly.

SophieGuo410 (approved):

ankagrawal (approved):

PR #2101, opened 2022-04-26
Ensure the implementation of PartitionId returns the correct partitio…

tbingol:
Ensure the implementation of PartitionId returns the correct partition id with the getId() call. This is to ensure that partition-id can eventually be passed to StorageClient::deleteBatch during the compaction. Partition id will be used to discover the corresponding Azure storage account.

### DIFF
--- a/ambry-tools/src/main/java/com/github/ambry/cloud/azure/AzureCompactionTool.java
+++ b/ambry-tools/src/main/java/com/github/ambry/cloud/azure/AzureCompactionTool.java
@@ -45,7 +45,7 @@ import org.slf4j.LoggerFactory;
 /**
  * Tool to purge dead blobs for an Ambry partition from Azure storage,
  * or to report the compaction progress of all partitions.
- * Usage: java -cp /path/to/ambry.jar AzureCompactionTool -propsFile <property-file-path> [-purge partitionPath...]
+ * Usage: java -cp /path/to/ambry.jar AzureCompactionTool -propsFile <property-file-path> [-purge partitionId...]
  */
 public class AzureCompactionTool {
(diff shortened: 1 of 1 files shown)


ambry-tools/src/main/java/com/github/ambry/cloud/azure/AzureCompactionTool.java
```diff
@@ -122,10 +122,10 @@ public static void main(String[] args) throws Exception {
    * PartitionId implementation that returns only its path.
    */
   private static class PartitionPathId implements PartitionId {
```
ankagrawal, line 124: nit: update the javadocs (because it doesn't return only path now).

ankagrawal (approved):

ambry-tools/src/main/java/com/github/ambry/cloud/azure/AzureCompactionTool.java
```diff
@@ -122,10 +122,10 @@ public static void main(String[] args) throws Exception {
...
-    private final String path;
+    private final long id;
 
-    private PartitionPathId(String path) {
-      this.path = path;
+    private PartitionPathId(Long id) {
```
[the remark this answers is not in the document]
tbingol: Done

ambry-tools/src/main/java/com/github/ambry/cloud/azure/AzureCompactionTool.java
tbingol: Done

Arun-LinkedIn (approved):

PR #2102, opened 2022-04-29
Fix NPE due to referencing null partition key in CosmosQueryRequestOptions

Arun-LinkedIn:
This commit fixes NPE seen while accessing partition key in CosmosQueryRequestOptions. Also, it makes sure that field names are defined properly in classes (Ex: CosmosContainerDeletionEntry) used to represent records in Cosmos DB. Since latest Cosmos SDK APIs use Jackson's ObjectMapper internally to serialize and deserialize POJOs to/from JSON, we might need to make sure the field names and getter/setter methods for them are valid in the classes.

Ran integration tests manually using AzureIntegrationTest.java, AzureContainerCompactorIntegrationTest.java, CloudBlobStoreIntegrationTest.java.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosContainerDeletionEntry.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosContainerDeletionEntry.java
@@ -13,6 +13,7 @@
  */
 package com.github.ambry.cloud.azure;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.github.ambry.account.Container;
 import com.github.ambry.clustermap.ClusterMap;
 import java.util.Collection;
@@ -25,6 +26,7 @@ import org.json.JSONObject;
 /**
  * Class representing container deletion status in cloud.
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
 public class CosmosContainerDeletionEntry {
   static final String VERSION_KEY = "version";
   static final String CONTAINER_ID_KEY = "containerId";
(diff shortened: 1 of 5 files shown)


tbingol (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosDataAccessor.java
```diff
@@ -856,7 +857,9 @@ private CosmosPagedFlux<CloudBlobMetadata> executeCosmosQuery(SqlQuerySpec sqlQu
   <T> CosmosPagedFlux<T> executeCosmosQuery(String containerName, SqlQuerySpec sqlQuerySpec,
       CosmosQueryRequestOptions cosmosQueryRequestOptions, Class<T> classType, Timer timer) {
     azureMetrics.documentQueryCount.inc();
-    if (!Utils.isNullOrEmpty(cosmosQueryRequestOptions.getPartitionKey().toString())) {
+    // Check for null value of partition key.
```
ankagrawal, line 860: What would it mean for partitionKey to be null? Should we assert for it?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/CosmosContainerDeletionEntry.java
```diff
@@ -25,6 +26,7 @@
 /**
  * Class representing container deletion status in cloud.
  */
+@JsonIgnoreProperties(ignoreUnknown = true)
```
ankagrawal, line 29: I think we should be worried if we find unknown fields. Lets throw exception in those cases instead of ignoring them.

ankagrawal (commented):
Looks good to me overall. Added some minor comments.

PR #2104, opened 2022-05-02
Ensure all Router tests exercise the bandwidth throttling logic.

ankagrawal:
This PR ensures all Router tests exercise the bandwidth throttling logic enabled bandwidth throttling.
It also fixes an issue where null QuotaChargeCallback was being passed from batch operations (like delete of a composite blob).

Its a long PR, But most of the changes are due to changes in `Router.java` interface to pass `QuotaChargeCallback` for each operation. This ensures that all operations that go through `OperationController` can exercise the quota logic.

The main changes are:
1. Change in `Router.java` to ensure `QuotaChargeCallback` object is passed for each operation.
2. Changes to update all the code and tests to changes in `Router.java` interface.
3. Update `QuotaConfig` to enable bandwidth throttling and `RouterConfig` to enable `QuotaAwareOperationController` by default, so that all tests using `Router`, exercise the bandwidth throttling logic.
4. Update `OperationController.java` and `NonBlockingRouter.java` to ensure that `QuotaChargeCallback` is not passed as `null` for any operation.

### DIFF
--- a/ambry-router/src/test/java/com/github/ambry/router/ChargeTesterQuotaManagerFactory.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/ChargeTesterQuotaManagerFactory.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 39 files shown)


ambry-router/src/test/java/com/github/ambry/router/ChargeTesterQuotaManagerFactory.java
```diff
@@ -0,0 +1,105 @@
...
+import com.github.ambry.quota.QuotaMetrics;
+import com.github.ambry.quota.QuotaName;
+import com.github.ambry.quota.QuotaRecommendationMergePolicy;
+import com.github.ambry.quota.capacityunit.AmbryCUQuotaEnforcer;
+import com.github.ambry.rest.RestRequest;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
+
+
+public class ChargeTesterQuotaManagerFactory implements QuotaManagerFactory {
```
ankagrawal, line 36: Note that this was a pre-existing class that has been moved out of NonBlockingQuotaTest.java.

PR #2107, opened 2022-05-10
Add partition key during change feed query. Also update cosmos sdk version to 4.29.1.

Arun-LinkedIn:
This commit makes sure to provide partition key while querying cosmos change feed. It also updates cosmos java SDK version to 4.29.1. With the earlier version (4.25.0), we were hitting the OOM issue since there seems to be an issue with initialization of cache maintained internally by cosmos library for storing query execution plans. Due to that and some of our queries containing IN clause not being parameterized (due to limitation on cosmos side as described [here](https://github.com/Azure/azure-cosmosdb-java/issues/60#issuecomment-477376945)), the cache was growing unbounded. This seems to have been fixed in SDK versions > 4.26.0 with this [PR](https://github.com/Azure/azure-sdk-for-java/pull/26825).


tbingol (approved):
LGTM, however, it would be better to have two separate PRs, one for the partition path and the other for the version upgrade.

PR #2112, opened 2022-05-24
Move the handler logic out of FrontendRestRequestService

justinlin-linkedin:
This PR would move the delete, head, and get blob logic out of FrontendRestRequestService and create a handler for each of them respectively. This way, our http handling code would be more consistent.

There are some changes made to this existing logics.
1. Some metrics are no longer used
2. processResponse would delete with delete

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
@@ -195,7 +195,8 @@ class AmbrySecurityService implements SecurityService {
       }
       RequestPath requestPath = RestUtils.getRequestPath(restRequest);
       RestMethod restMethod = restRequest.getRestMethod();
-      if (blobInfo == null && !restMethod.equals(RestMethod.OPTIONS) && !restMethod.equals(RestMethod.PUT)) {
+      if (blobInfo == null && !restMethod.equals(RestMethod.OPTIONS) && !restMethod.equals(RestMethod.PUT)
+          && !restMethod.equals(RestMethod.DELETE)) {
         if (!requestPath.matchesOperation(Operations.GET_SIGNED_URL)) {
           throw new IllegalArgumentException("BlobInfo is null");
         }
@@ -268,6 +269,9 @@ class AmbrySecurityService implements SecurityService {
               responseChannel.setHeader(RestUtils.Headers.CREATION_TIME,
                   new Date(blobInfo.getBlobProperties().getCreationTimeInMs()));
             }
+            break;
+          case DELETE:
+            // There is nothing to do with delete
             break;
           default:
             exception = new RestServiceException("Cannot process response for request with method " + restMethod,
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
@@ -152,7 +152,12 @@ public class DeleteBlobHandler {
         restResponseChannel.setHeader(RestUtils.Headers.DATE, new GregorianCalendar().getTime());
         restResponseChannel.setHeader(RestUtils.Headers.CONTENT_LENGTH, 0);
         securityService.processResponse(restRequest, restResponseChannel, null, securityProcessResponseCallback());
-      }, restRequest.getUri(), LOGGER, finalCallback);
+      }, restRequest.getUri(), LOGGER, (r, e) -> {
+        // Even we failed in router operations, we already used some of the resources in router,
+        // so let's record the charges for this request.
+        securityService.processRequestCharges(restRequest, restResponseChannel, null);
+        finalCallback.onCompletion(null, e);
+      });
     }
 
     /**
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
@@ -32,11 +32,11 @@ import static com.github.ambry.rest.RestUtils.*;
 
 
 /**
- * Handler for all delete requests.
+ * Handler for all delete blob requests.
  */
-public class DeleteHandler {
+public class DeleteBlobHandler {
 
-  private static final Logger LOGGER = LoggerFactory.getLogger(DeleteHandler.class);
+  private static final Logger LOGGER = LoggerFactory.getLogger(DeleteBlobHandler.class);
 
   private final Router router;
   private final SecurityService securityService;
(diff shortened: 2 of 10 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -0,0 +1,253 @@
...
+      }, restRequest.getUri(), LOGGER, finalCallback);
+    }
+
+    /**
+     * After {@link IdConverter#convert} finishes, call {@link SecurityService#postProcessRequest} to perform
+     * request time security checks that rely on the request being fully parsed and any additional arguments set.
+     * @return a {@link Callback} to be used with {@link IdConverter#convert}.
+     */
+    private Callback<String> idConverterCallback() {
+      return buildCallback(metrics.getBlobIdConversionMetrics, convertedBlobId -> {
```
snalli, line 166: Can we avoid lambdas and put this code in a named function ?

justinlin-linkedin: that would be the next PR that would move handler logic away from callbacks and into the completablefuture world.

ambry-frontend/src/main/java/com/github/ambry/frontend/HeadHandler.java
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+     * After {@link IdConverter#convert} finishes, call {@link SecurityService#postProcessRequest} to perform
+     * request time security checks that rely on the request being fully parsed and any additional arguments set.
+     * @return a {@link Callback} to be used with {@link IdConverter#convert}.
+     */
+    private Callback<String> idConverterCallback() {
+      return buildCallback(metrics.headBlobIdConversionMetrics, convertedBlobId -> {
+        BlobId blobId = FrontendUtils.getBlobIdFromString(convertedBlobId, clusterMap);
+        accountAndContainerInjector.injectTargetAccountAndContainerFromBlobId(blobId, restRequest,
```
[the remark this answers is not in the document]
justinlin-linkedin: the idea here is to use less if statement to know what operation we are doing here. I guess I can check if we already have account and container in the request, if so we don't have to inject them again.

ambry-frontend/src/main/java/com/github/ambry/frontend/HeadBlobHandler.java
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * After {@link IdConverter#convert} finishes, call {@link SecurityService#postProcessRequest} to perform
+     * request time security checks that rely on the request being fully parsed and any additional arguments set.
+     * @return a {@link Callback} to be used with {@link IdConverter#convert}.
+     */
+    private Callback<String> idConverterCallback() {
+      return buildCallback(metrics.headBlobIdConversionMetrics, convertedBlobId -> {
+        BlobId blobId = FrontendUtils.getBlobIdFromString(convertedBlobId, clusterMap);
+        accountAndContainerInjector.injectTargetAccountAndContainerFromBlobId(blobId, restRequest,
```
Arun-LinkedIn, line 121: May be missing check to see if the Account and container is not already injected (for named blobs)?

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -269,6 +270,9 @@ public void processResponse(RestRequest restRequest, RestResponseChannel respons
                   new Date(blobInfo.getBlobProperties().getCreationTimeInMs()));
             }
             break;
+          case DELETE:
```
Arun-LinkedIn, line 273: Just noting that this might also require similar change in internal Ambry code (AmbryLiSecurityService). You might already be aware :)

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -0,0 +1,259 @@
...
+            case Segment:
+              router.getBlob(blobId.getID(), options, routerCallback(),
+                  QuotaUtils.buildQuotaChargeCallback(restRequest, quotaManager, true));
+              break;
+            case Replicas:
+              finalCallback.onCompletion(getReplicasHandler.getReplicas(blobId.getID(), restResponseChannel), null);
+              break;
+          }
+        }
+      }, restRequest.getUri(), LOGGER, finalCallback);
```
Arun-LinkedIn, line 208: Looks like if we are still [calculating](https://github.com/linkedin/ambry/blob/master/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java#L949) charges incurred for the GET request on router exceptions. Would we need same here too?

ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
```diff
@@ -0,0 +1,167 @@
...
+     * the blob in the storage layer.
+     * @param blobId the {@link BlobId} to undelete
+     * @return a {@link Callback} to be used with {@link SecurityService#postProcessRequest}.
+     */
+    private Callback<Void> securityPostProcessRequestCallback(BlobId blobId) {
+      return buildCallback(metrics.deleteBlobSecurityPostProcessRequestMetrics, result -> {
+        String serviceId = RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.SERVICE_ID, false);
+        router.deleteBlob(blobId.getID(), serviceId, routerCallback(),
+            QuotaUtils.buildQuotaChargeCallback(restRequest, quotaManager, false));
+      }, restRequest.getUri(), LOGGER, finalCallback);
```
Arun-LinkedIn, line 141: Looks like current code [calculates](https://github.com/linkedin/ambry/blob/master/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java#L1006) charges incurred for DELETE operations on router exceptions too. Would we need same here?

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
justinlin-linkedin: for sure

ambry-frontend/src/main/java/com/github/ambry/frontend/HeadBlobHandler.java
justinlin-linkedin: updated.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
justinlin-linkedin: it's weird, we already call the processRequestCharges in securityService.processResponse, it seems like we are doing duplicate work here.

ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
justinlin-linkedin: above.

SophieGuo410 (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
Arun-LinkedIn: I am just thinking that securityService.processResponse might not be called if there is a router exception since we directly invoke final callback. But not sure if we need to calculate request charges on failures. Our current code seems to be doing that.

Arun-LinkedIn (approved):

PR #2116, opened 2022-05-26
Fix a potential memory leak in http2networkclient

justinlin-linkedin:
Could this be the culprit? 
In drop request, we only retrieve the request info from the channel attribute map, which means we always keep a reference of request info in the attribute map. If this request info somehow has a memory leak, we will never detect it since there is always a valid reference. 

This should invalidate the reference of dropped requests in attribute map. Hopefully this would fix the memory spike issue or at leasts show more information in the log files.


SophieGuo410 (approved):

ambry-network/src/main/java/com/github/ambry/network/http2/Http2ClientResponseHandler.java
```diff
@@ -54,8 +54,9 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpResponse msg) {
           System.currentTimeMillis() - requestInfo.getStreamHeaderFrameReceiveTime());
       ResponseInfo responseInfo = new ResponseInfo(requestInfo, null, dup);
       responseInfoQueue.put(responseInfo);
-      releaseAndCloseStreamChannel(ctx.channel());
     }
+    // release the stream anyway
+    releaseAndCloseStreamChannel(ctx.channel());
```
Arun-LinkedIn, line 59: From this [code](https://github.com/justinlin-linkedin/ambry/blob/8e6309d582229ddb681b2b0d3a54f162be8e0793/ambry-network/src/main/java/com/github/ambry/network/http2/Http2Utils.java#L32), it looks like `releaseAndCloseStreamChannel` also has check for `requestInfo` before releasing the channel to pool. Since, we are resetting the requestInfo attribute to null when dropping the request, would it be possible that the channel associated with dropped request is not released back.

Arun-LinkedIn (approved):

PR #2121, opened 2022-05-30
[Http2] Add a configuration to treat timeout as network error in http2 network client

justinlin-linkedin:
when ambry server has disk failures, it would not respond to any of the frontend requests. Frontend nodes would sees lots of timeout error from this pariticular ambry server node. But since timeout is not treated as the network error, frontend nodes would still consider this ambry server live and keep sending requests to this ambry server node. 

This PR adds a configuration that would make http2 network client return network error instead of timeout when seeing a timeout. This would force frontend nodes stop sending requests to the non-responding ambry server nodes.

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
@@ -92,7 +92,11 @@ public class Http2NetworkClient implements NetworkClient {
           // We need stream reset to cancel the stream in transmitting.
           RequestInfo requestInfo = streamChannel.attr(Http2NetworkClient.REQUEST_INFO).getAndSet(null);
           if (requestInfo != null) {
-            readyResponseInfos.add(new ResponseInfo(requestInfo, NetworkClientErrorCode.TimeoutError, null));
+            NetworkClientErrorCode errorCode = NetworkClientErrorCode.TimeoutError;
+            if (http2ClientConfig.http2TimeoutAsNetworkError) {
+              errorCode = NetworkClientErrorCode.NetworkError;
+            }
+            readyResponseInfos.add(new ResponseInfo(requestInfo, errorCode, null));
             ReferenceCountUtil.safeRelease(requestInfo.getRequest());
           }
         }
@@ -199,7 +203,8 @@ public class Http2NetworkClient implements NetworkClient {
                 successCount.incrementAndGet();
               } else {
                 failCount.incrementAndGet();
-                responseInfoList.add(new ResponseInfo(null, NetworkClientErrorCode.NetworkError, null, dataNodeId, false));
+                responseInfoList.add(
+                    new ResponseInfo(null, NetworkClientErrorCode.NetworkError, null, dataNodeId, false));
                 logger.error("Couldn't acquire stream channel to {}:{} . Cause: {}.", dataNodeId.getHostname(),
                     dataNodeId.getHttp2Port(), future.cause());
               }
--- a/ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
@@ -37,6 +37,7 @@ public class Http2ClientConfig {
   public static final String HTTP2_BLOCKING_CHANNEL_POOL_SHUTDOWN_TIMEOUT_MS =
       "http2.blocking.channel.pool.shutdown.timeout.ms";
   public static final String HTTP2_PEER_CERTIFICATE_SAN_REGEX = "http2.peer.certificate.san.regex";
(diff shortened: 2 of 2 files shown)


ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -91,7 +91,11 @@ public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Inte
           // We need stream reset to cancel the stream in transmitting.
           RequestInfo requestInfo = streamChannel.attr(Http2NetworkClient.REQUEST_INFO).get();
           if (requestInfo != null) {
-            readyResponseInfos.add(new ResponseInfo(requestInfo, NetworkClientErrorCode.TimeoutError, null));
+            NetworkClientErrorCode errorCode = NetworkClientErrorCode.TimeoutError;
```
ankagrawal, line 94: Not getting a response from server node might also mean network connection issues on the frontend itself. For example, if the connection to the server nodes timeout due to network issues on frontend, it should not mark those nodes with network errors. The server nodes might be totally fine in this case.
Should we actually look into the server code instead, and make sure that in case of disk failures, the server node returns appropriate error message, instead of just timing out?

ankagrawal: Comments on GetBlobOperation::GetChunk::cleanupExpiredInflightRequests line 801 & PutOperation::PutChunk::cleanupExpiredInflightRequests, also seem indicate something similar.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -91,7 +91,11 @@ public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Inte
...
+            if (http2ClientConfig.http2TimeoutAsNetworkError) {
+              errorCode = NetworkClientErrorCode.NetworkError;
+            }
+            readyResponseInfos.add(new ResponseInfo(requestInfo, errorCode, null));
```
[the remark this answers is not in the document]
justinlin-linkedin: we can do that, but the line would be too long.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
justinlin-linkedin: the server code does return correct error code, it's disk_unavailable. And frontend does mark disks as unavailable. 

The problem i am getting in this case, is that in server, the metrics shows there is no incoming requests after disk failures and in frontend, the log shows all the requests going to this particular server time out. But since we are not treating time out as network error, frontend keeps sending requests to this server and keep getting time outs and keep dropping requests. 

Treating timeout as network isn't an ideal solution here, but the problem is that we are only getting timeout from this failing host and that should say something about the health of this server host.

SophieGuo410: Once the code has been verified, are we plan to replace the NetworkClientErrorCode.TimeoutError with NetworkClientErrorCode.NetworkError? 
Since based on below logic, it will treat NetworkError as ReplicaEventType.Node_Timeout anyway.
private void onNetworkEvent(ReplicaId replicaId, NetworkClientErrorCode errorCode) {
    switch (errorCode) {
      case NetworkError:
        clusterMap.onReplicaEvent(replicaId, ReplicaEventType.Node_Timeout);
        break;
      default:
        break;
    }

SophieGuo410 (approved):

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
justinlin-linkedin: that's right, we will treat timeout as network error and mark this data node as down and stop sending requests to this data node.

One thing I think worth mentioning is that, NetworkErrorCode.Timeout is added with http2 network client. In SocketNetworkClient, there are only ConnectionUnavailable and NetworkError. ConnectionUnavailable means there are too many outgoing requests happening at the same time and we can't check out any socket connection anymore. And NetworkError means as long as there is an error in the socket connection. Timeout is one of those errors. In SocketNetworkClient, we treat timeout (dropped requests) as network error already. 

I suppose the reason why we are not doing that in Http2NetworkClient is that we don't have a way to close the stream like the way we close the socket connection in SocketNetworkClient. But I will try this and see how things go.

justinlin-linkedin:
> Approving this so we can test the changes. However, we do need to think about cases where the network issue is with the frontend node. Just removing server replicas in that case will lead the frontend to think that server nodes are unavailable. We should atleast plan to log appropriate error logs for those cases.

the server node will temporarily be marked as soft down. There is a fixed backoff logic to save this server node back from unavailability. Basically, after 6 consecutive time out, this server would be marked as down. And after 5 minutes, we will retry this server again.

PR #2122, opened 2022-05-30
Fix onSuccess call in get in blobstore

justinlin-linkedin:
In BlobStore.get method, it call onSuccess after get method returns a StoreMessageReadSet back to the caller. But get method never performs any io operation to read bytes from disks, so we can't really call onSuccess or onError in get method. We have to call onSuccess and onError after reading the bytes in StoreMessageReadSet.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -309,7 +309,7 @@ public class BlobStore implements Store {
         }
       }
 
-      MessageReadSet readSet = new StoreMessageReadSet(readOptions, new StoreMessageReadSet.IOOperationHandler() {
+      MessageReadSet readSet = new StoreMessageReadSet(readOptions, new StoreMessageReadSet.IOPHandler() {
         @Override
         public void onSuccess() {
           BlobStore.this.onSuccess("GET");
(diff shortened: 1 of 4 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -888,7 +900,6 @@ public FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries, St
         remoteTokenTracker.updateTokenFromPeerReplica(token, hostname, remoteReplicaPath);
       }
       FindInfo findInfo = index.findEntriesSince(token, maxTotalSizeOfEntries);
-      onSuccess();
```
justinlin-linkedin, line 891: findEntriesSince doesn't perform any disk io since all the index files are already loaded in the memory.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -309,14 +309,26 @@ public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> stor
...
+          }
+        }
+      });
       // We ensure that the metadata list is ordered with the order of the message read set view that the
       // log provides. This ensures ordering of all messages across the log and metadata from the index.
       List<MessageInfo> messageInfoList = new ArrayList<MessageInfo>(readSet.count());
       for (int i = 0; i < readSet.count(); i++) {
         messageInfoList.add(indexMessages.get(readSet.getKeyAt(i)));
       }
-      onSuccess();
```
JingQianCloud, line 319: Overall looks good to me.
Do you mind add your comment why we remove the onSuccess to the handler? It'll help to understand the code since the onSuccess for Get is different from all the other functions now.

JingQianCloud (approved):
Overall looks good to em.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: sure

PR #2123, opened 2022-06-02
Integrate with Helix vm freeze event handler

JingQianCloud:
1. Bump up Helix core from 1.0.1 to 1.0.3.
2. For frontend, integrate with Helix VM event handler.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -35,9 +35,15 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import org.apache.helix.AccessOption;
+import org.apache.helix.cloud.event.helix.CloudEventCallbackProperty;
+import org.apache.helix.cloud.event.helix.CloudEventCallbackProperty.HelixOperation;
 import org.apache.helix.HelixAdmin;
+import org.apache.helix.HelixCloudProperty;
 import org.apache.helix.HelixManager;
+import org.apache.helix.HelixManagerProperty;
 import org.apache.helix.InstanceType;
+import org.apache.helix.manager.zk.ZKHelixManager;
+import org.apache.helix.model.CloudConfig;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.store.HelixPropertyStore;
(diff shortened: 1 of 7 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -95,7 +101,31 @@ public HelixParticipant(ClusterMapConfig clusterMapConfig, HelixFactory helixFac
...
+    // Cloud Event Callback Property: A property to define the maintenance event handling behavior for a Helix manager.
+    // User can specify which callback implementation class they want to use for Helix operations, if not specified, it will use org.apache.helix.cloud.event.helix.DefaultCloudEventCallbackImpl.class by default
+    CloudEventCallbackProperty callbackProperty = new CloudEventCallbackProperty(Collections.emptyMap());
+
+    // Enable/disable Helix optional operations for a Helix manager via CloudEventCallbackProperty.
+    callbackProperty.setHelixOperationEnabled(HelixOperation.ENABLE_DISABLE_INSTANCE, true);
+    callbackProperty.setHelixOperationEnabled(HelixOperation.MAINTENANCE_MODE, true);
+
+    // Helix Cloud Property: General cloud-related property for a Helix manager, includes CloudEventCallbackProperty.
+    HelixCloudProperty cloudProperty = new HelixCloudProperty(new CloudConfig(new ZNRecord(clusterName)));
```
mgao0, line 118: As we discussed offline, there are two ways to supply a cloud config, you could build one in code, or read from zk in ZKHelixManager constructor. `new CloudConfig(new ZNRecord(clusterName)` this means a default cloud config which is usually used when cloud is not enabled. When on cloud, if the cloud config is to be read from zk, then it's fine; but if a cloud config is meant to be built in code, then you will need to do more here.

mgao0 (commented):
Generally looks good to me. Left one comment to consider. Discussed offline about having listeners registered for non-cloud environment won't do harm because no signal is going to trigger the listeners.

gradle/dependency-versions.gradle
```diff
@@ -19,7 +19,7 @@ ext {
     javaxVersion = "3.0.1"
     nettyVersion = "4.1.74.Final"
     nettyTcnativeVersion = "2.0.49.Final"
-    helixVersion = "1.0.1"
```
xyuanlu, line 22: Please don't use 1.0.3. This version will be deprecated.

JingQianCloud: Ack. Will wait for the next Helix release.

JingQianCloud:
Discard the PR.

PR #2125, opened 2022-06-02
Fix dropped request in http2 network client

justinlin-linkedin:
We are getting NPE in TtlUpdateManager, where a correlation id would point to a null TtlUpdateOperation. The only possible explanation is that we received two responses with the same request. This could only happen when we dropped a request and return the response  and later the real response comes back. 

This PR would close the channel right away after we drop requests.

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
@@ -86,7 +86,7 @@ public class Http2NetworkClient implements NetworkClient {
       for (int correlationId : requestsToDrop) {
         Channel streamChannel = correlationIdInFlightToChannelMap.remove(correlationId);
         if (streamChannel != null) {
-          RequestInfo requestInfo = streamChannel.attr(Http2NetworkClient.REQUEST_INFO).getAndSet(null);
+          RequestInfo requestInfo = releaseAndCloseStreamChannel(streamChannel);
           // Drop request just generates a ResponseInfo with TimeoutError to router.
           // The stream is still transmitting, but router will ignore ResponseInfo with the same correlationId.
           // We need stream reset to cancel the stream in transmitting.
@@ -99,7 +99,6 @@ public class Http2NetworkClient implements NetworkClient {
             }
             readyResponseInfos.add(new ResponseInfo(requestInfo, errorCode, null));
             ReferenceCountUtil.safeRelease(requestInfo.getRequest());
-            releaseAndCloseStreamChannel(streamChannel);
           } else {
             logger.warn("Drop request {} on streamChannel: {}", correlationId, streamChannel);
           }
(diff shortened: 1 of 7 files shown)


ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
```diff
@@ -86,18 +86,22 @@ public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Inte
...
           if (requestInfo != null) {
+            logger.warn("Drop request {}:{} on streamChannel: {}", requestInfo.getRequest().getRequestOrResponseType(),
+                correlationId, streamChannel);
             NetworkClientErrorCode errorCode = NetworkClientErrorCode.TimeoutError;
             if (http2ClientConfig.http2TimeoutAsNetworkError) {
               errorCode = NetworkClientErrorCode.NetworkError;
             }
             readyResponseInfos.add(new ResponseInfo(requestInfo, errorCode, null));
             ReferenceCountUtil.safeRelease(requestInfo.getRequest());
+            releaseAndCloseStreamChannel(streamChannel);
```
ankagrawal, line 102: The `releaseAndCloseStreamChannel` won't do anything here. Because we are already calling `streamChannel.attr(Http2NetworkClient.REQUEST_INFO).getAndSet(null);` in line 89 above. So within `releaseAndCloseStreamChannel`, the requestInfo will be null.

We should maybe call `releaseAndCloseStreamChannel` in line 89 itself?

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateManager.java
```diff
@@ -160,8 +160,13 @@ void handleResponse(ResponseInfo responseInfo) {
         RouterUtils.extractResponseAndNotifyResponseHandler(responseHandler, routerMetrics, responseInfo,
             TtlUpdateResponse::readFrom, TtlUpdateResponse::getError);
     RequestInfo routerRequestInfo = responseInfo.getRequestInfo();
-    int correlationId = ((TtlUpdateRequest) routerRequestInfo.getRequest()).getCorrelationId();
+    int correlationId = routerRequestInfo.getRequest().getCorrelationId();
     TtlUpdateOperation ttlUpdateOperation = correlationIdToTtlUpdateOperation.remove(correlationId);
+    if (ttlUpdateOperation == null) {
+      LOGGER.warn("No TtlUpdateOperation found for correlation id: {}", correlationId);
```
ankagrawal, line 166: Can we do this same logging in the else of line 187 below?

justinlin-linkedin: updated.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2NetworkClient.java
justinlin-linkedin: updated

ankagrawal (approved):

SophieGuo410 (approved):
LGTM!

PR #2127, opened 2022-06-03
[Frontend] Increase MAX_JSON_REQUEST_SIZE_BYTES_KEY

snalli:
This patch increases the MAX_JSON_REQUEST_SIZE_BYTES_KEY from 20mb to
128Mb. This is the maximum size of json sent as part of a stitch
request. The stitch request for 1tb blob is 64mb. Setting it to 128mb
allows 2tb stitch requests.

Tested on localhost mac by uploading a 64mb stitch request.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
@@ -218,7 +218,7 @@ public class FrontendConfig {
    * The maximum size in bytes for the JSON body of a "POST /stitch" request.
    */
   @Config(MAX_JSON_REQUEST_SIZE_BYTES_KEY)
-  @Default("20 * 1024 * 1024")
+  @Default("128 * 1024 * 1024")
   public final int maxJsonRequestSizeBytes;
 
   /**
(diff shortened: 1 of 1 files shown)


ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
```diff
@@ -314,7 +314,7 @@ public FrontendConfig(VerifiableProperties verifiableProperties) {
     maxAcceptableTtlSecsIfTtlRequired = verifiableProperties.getIntInRange(MAX_ACCEPTABLE_TTL_SECS_IF_TTL_REQUIRED_KEY,
         (int) TimeUnit.DAYS.toSeconds(30), 0, Integer.MAX_VALUE);
     maxJsonRequestSizeBytes =
-        verifiableProperties.getIntInRange(MAX_JSON_REQUEST_SIZE_BYTES_KEY, 20 * 1024 * 1024, 0, Integer.MAX_VALUE);
+        verifiableProperties.getIntInRange(MAX_JSON_REQUEST_SIZE_BYTES_KEY, 128 * 1024 * 1024, 0, Integer.MAX_VALUE);
```
justinlin-linkedin, line 317: do we need 128MB to be the default value? or can we just change the value for this configuration in some cluster?

snalli: good idea. yes we can change the config for a specific cluster.

PR #2132, opened 2022-06-14
[Frontend] Add metadata cache in router

snalli:
This patch adds a cache to save metadata chunks of composite blobs.
The NonBlockingRouter instantiates a volatile MetadataCache. The cache
maps a String blobID to a MetadataChunk object. We use Guava Cache and
configure it to hold a maximum of routerMetadataCacheMaxItemCount items.
Localhost experiments show we save 110-125ms on RANGE requests on
a 1TB blob with range sizes varying from 1MB to 1024MB.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -114,6 +114,8 @@ public class RouterConfig {
   public static final String ROUTER_STORE_KEY_CONVERTER_FACTORY = "router.store.key.converter.factory";
   public static final String ROUTER_UNAVAILABLE_DUE_TO_OFFLINE_REPLICAS = "router.unavailable.due.to.offline.replicas";
   public static final String ROUTER_NOT_FOUND_CACHE_TTL_IN_MS = "router.not.found.cache.ttl.in.ms";
+  private static final String ROUTER_METADATA_CACHE = "router.metadata.cache" ;
+  private static final String ROUTER_METADATA_CACHE_MAX_ITEM_COUNT = "router.metadata.cache.max.item.count";
 
   /**
    * Number of independent scaling units for the router.
@@ -578,6 +580,11 @@ public class RouterConfig {
   @Default("15*1000")
   public final long routerNotFoundCacheTtlInMs;
 
+  @Config(ROUTER_METADATA_CACHE)
+  public final boolean routerMetadataCache;
+
+  @Config(ROUTER_METADATA_CACHE_MAX_ITEM_COUNT)
+  public final int routerMetadataCacheMaxItemCount;
   /**
    * Create a RouterConfig instance.
    * @param verifiableProperties the properties map to refer to.
@@ -643,6 +650,8 @@ public class RouterConfig {
         verifiableProperties.getIntInRange(ROUTER_UNDELETE_REQUEST_PARALLELISM, 3, 1, Integer.MAX_VALUE);
     routerUseGetBlobOperationForBlobInfo =
         verifiableProperties.getBoolean(ROUTER_USE_GET_BLOB_OPERATION_FOR_BLOB_INFO, false);
+    routerMetadataCache = verifiableProperties.getBoolean(ROUTER_METADATA_CACHE, false);
+    routerMetadataCacheMaxItemCount = verifiableProperties.getInt(ROUTER_METADATA_CACHE_MAX_ITEM_COUNT, 100);
     List<String> customPercentiles =
         Utils.splitString(verifiableProperties.getString(ROUTER_OPERATION_TRACKER_CUSTOM_PERCENTILES, ""), ",");
     routerOperationTrackerCustomPercentiles =
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -114,7 +114,7 @@ public class RouterConfig {
   public static final String ROUTER_STORE_KEY_CONVERTER_FACTORY = "router.store.key.converter.factory";
   public static final String ROUTER_UNAVAILABLE_DUE_TO_OFFLINE_REPLICAS = "router.unavailable.due.to.offline.replicas";
   public static final String ROUTER_NOT_FOUND_CACHE_TTL_IN_MS = "router.not.found.cache.ttl.in.ms";
-  private static final String ROUTER_METADATA_CACHE = "router.metadata.cache" ;
+  private static final String ROUTER_METADATA_CACHE_ENABLED = "router.metadata.cache.enabled" ;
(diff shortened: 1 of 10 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -99,6 +100,7 @@ class GetBlobOperation extends GetOperation {
   private final Map<Integer, GetChunk> correlationIdToGetChunk = new HashMap<>();
   // Callback to charge against quota for each chunk that's fetched.
   private final QuotaChargeCallback quotaChargeCallback;
+  private final GetManager getManager;
```
justinlin-linkedin, line 103: I suggest we create a class, which is just a thin wrap of the Guava Cache, and we create an object of this class and pass this object to GetBlobOperation, so that we don't have to implements the same method in all these classes.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -114,6 +114,8 @@ public class RouterConfig {
   public static final String ROUTER_STORE_KEY_CONVERTER_FACTORY = "router.store.key.converter.factory";
   public static final String ROUTER_UNAVAILABLE_DUE_TO_OFFLINE_REPLICAS = "router.unavailable.due.to.offline.replicas";
   public static final String ROUTER_NOT_FOUND_CACHE_TTL_IN_MS = "router.not.found.cache.ttl.in.ms";
+  private static final String ROUTER_METADATA_CACHE = "router.metadata.cache" ;
```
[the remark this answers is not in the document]
snalli: done

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -578,6 +580,11 @@ public class RouterConfig {
   @Default("15*1000")
   public final long routerNotFoundCacheTtlInMs;
 
+  @Config(ROUTER_METADATA_CACHE)
+  public final boolean routerMetadataCache;
+
+  @Config(ROUTER_METADATA_CACHE_MAX_ITEM_COUNT)
+  public final int routerMetadataCacheMaxItemCount;
```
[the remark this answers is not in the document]
snalli: done

ambry-commons/src/main/java/com/github/ambry/commons/MetadataChunk.java
```diff
@@ -0,0 +1,29 @@
+package com.github.ambry.commons;
+
+import com.github.ambry.messageformat.BlobInfo;
+import com.github.ambry.messageformat.CompositeBlobInfo;
+
+public class MetadataChunk {
```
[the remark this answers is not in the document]
snalli: done

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: done

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1289,6 +1338,9 @@ GetOption getGetOption() {
      */
     @Override
     MessageFormatFlags getOperationFlag() {
+      if (mdCacheLookupResult == mdCacheLookUpStatus.mdCacheHit) {
+        return MessageFormatFlags.BlobInfo;
```
Arun-LinkedIn, line 1342: It looks like if we the blob ID in metadata cache, we are modifying GetBlobOperation to get blobInfo instead of blob for the first chunk to validate the blob. 

Just thinking that one other way could be to fetch the getBlobInfo from [NonBlockingRouter](https://github.com/linkedin/ambry/blob/master/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java#L240) (a bit higher level) itself if the blob ID is found in cache. If the getBlobInfo() is successful, we could issue getBlob() call. Else, we could report the error to user. With this approach, advantage seems to be that we don't need to integrate BlobInfo operation inside GetBlobOperation and handle the exceptions at upper level. But, we might have to do the operations we do during handling of metadata response such as calculating the needed [chunk IDs](https://github.com/linkedin/ambry/blob/master/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java#L1523), [initializing the data chunk objects](https://github.com/linkedin/ambry/blob/master/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java#L1571), etc in the [GetBlobOperation constructor](https://github.com/linkedin/ambry/blob/master/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java#L154) since there is no request-response flow for the metadata chunk. What do you think?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1372,15 +1424,24 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
           rawPayloadBuffer = ByteBuffer.wrap(payloadBytes);
         }
 
-        if (getOperationFlag() == MessageFormatFlags.Blob) {
+        if (getOperationFlag() == MessageFormatFlags.BlobInfo) {
+          blobType = BlobType.MetadataBlob;
+          blobData = null;
+          serverBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
+          ByteBuffer userMd = MessageFormatRecord.deserializeUserMetadata(payload);
+          blobInfo = new BlobInfo(serverBlobProperties, userMd.array(), messageInfo.getLifeVersion());
```
Arun-LinkedIn, line 1432: I think the user metadata in the metadata blob might be encrypted. We might have to use the user metadata/blobInfo from the metadata cache.

snalli:
Closing this. Will refactor and split into multiple PRs for ease of review.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -578,6 +580,20 @@ public class RouterConfig {
   @Default("15*1000")
   public final long routerNotFoundCacheTtlInMs;
 
+  /**
+   * Toggles metadata cache improvement in NonBlockingRouter.
+   * If true, metadata chunk of a composite blob is cached.
+   * If false, there is no caching of metadata chunks.
+   */
+  @Config(ROUTER_METADATA_CACHE_ENABLED)
+  public final boolean routerMetadataCacheEnabled;
```
Arun-LinkedIn, line 589: Looks like setting the cache entry size to 0 also disables the cache by evicting every entry. But, may be it is better to config to be more explicit and avoid touching the cache.

snalli:
Created a new one with refactored code. Closing this as I messed up my local branches.

PR #2135, opened 2022-06-16
Changing SecurityService, IdConverter, Router interface to add CompletableFuture

justinlin-linkedin:
Adding CompletableFuture to SecurityService, Router and IdConverter interface.

With these interface supporting CompletableFuture, we can move FrontRestRequestService's HttpHandlers away from Callback hell and into CompletableFuture.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/frontend/IdConverter.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/IdConverter.java
@@ -13,10 +13,12 @@
  */
 package com.github.ambry.frontend;
 
+import com.github.ambry.commons.CallbackUtils;
 import com.github.ambry.messageformat.BlobInfo;
 import com.github.ambry.rest.RestRequest;
 import com.github.ambry.commons.Callback;
 import java.io.Closeable;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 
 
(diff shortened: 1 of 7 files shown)


ambry-api/src/main/java/com/github/ambry/frontend/IdConverter.java
```diff
@@ -37,6 +39,18 @@ public interface IdConverter extends Closeable {
    */
   Future<String> convert(RestRequest restRequest, String input, Callback<String> callback);
 
+  /**
+   * Converts an ID.
+   * @param restRequest {@link RestRequest} representing the request.
+   * @param input the ID that needs to be converted.
+   * @return a {@link CompletableFuture} that will eventually contain the converted ID.
+   */
+  default CompletableFuture<String> convert(RestRequest restRequest, String input) {
```
Arun-LinkedIn, line 48: Looks like it is [suggested](https://stackoverflow.com/questions/47571117/what-is-the-difference-between-completionstage-and-completablefuture/49158702#49158702) to have APIs return `CompletionStage` which is an interface of `CompletableFuture` since the `future` is usually expected to be completed in the implementation and we don't expose `complete` methods to caller. Just wondering if it would be okay to return `CompletionStage` if we don't see need for caller of these APIs to complete the `future`.  (I think I may also have to do the same where I added `CompletableFutures` to other classes before).

ambry-api/src/main/java/com/github/ambry/frontend/SecurityService.java
```diff
@@ -81,36 +82,36 @@ void processResponse(RestRequest restRequest, RestResponseChannel responseChanne
    * Similar to {@link #preProcessRequest(RestRequest, Callback)} but returns a {@link Future} instead of requiring
```
Arun-LinkedIn, line 82: nit: 'but returns a {@link CompletableFuture}'

ambry-api/src/main/java/com/github/ambry/frontend/SecurityService.java
```diff
@@ -81,36 +82,36 @@ void processResponse(RestRequest restRequest, RestResponseChannel responseChanne
...
-    preProcessRequest(restRequest, futureResult::done);
-    return futureResult;
+  default CompletableFuture<Void> preProcessRequest(RestRequest restRequest) {
+    CompletableFuture<Void> future = new CompletableFuture<>();
+    preProcessRequest(restRequest, CallbackUtils.fromCompletableFuture(future));
+    return future;
   }
 
   /**
    * Similar to {@link #processRequest(RestRequest, Callback)} but returns a {@link Future} instead of requiring
```
Arun-LinkedIn, line 94: nit: but returns a {@link CompletableFuture}

ambry-api/src/main/java/com/github/ambry/frontend/IdConverter.java
justinlin-linkedin: In principal that's true. But 1. we are already returning CompletableFuture in NamedBlobDb 2. CompletionStage has a [method](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletionStage.html#toCompletableFuture--) `toCompletableFuture` to return the underlying implementation. It's weird.

snalli (approved):

PR #2136, opened 2022-06-16
Fixed Cloud VCR sync Helix cluster failure.

JingQianCloud:
The VCR replication failure is due to Helix update didn't work.
We picked the wrong source Helix Zk connection string. For Azure VCR, we couldn't use local DC Zk String to find the Helix resource. It didn't exist.

1. Patched on Azure VCR node and tested the fix. It replication works now.
2. ./gradlew build locally failed with 24 unit test case. But the existing code has the same issue. Shouldn't be related to the code change. PR test run passed.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
@@ -52,10 +52,12 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
+import org.apache.helix.HelixAdmin;
 import org.apache.helix.lock.DistributedLock;
 import org.apache.helix.lock.LockScope;
 import org.apache.helix.lock.helix.HelixLockScope;
 import org.apache.helix.lock.helix.ZKDistributedNonblockingLock;
+import org.apache.helix.manager.zk.ZKHelixAdmin;
 
 
 /**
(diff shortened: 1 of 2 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -110,6 +118,43 @@ public VcrReplicationManager(CloudConfig cloudConfig, ReplicationConfig replicat
...
+            HelixAdmin srcAdmin = new ZKHelixAdmin(zkStr);
+            List<String> srcResources = srcAdmin.getResourcesInCluster(syncSrcClusterName);
+            syncSrcZkStr = zkStr;
+            break;
+          } catch (Exception e) {
+            logger.warn("Ambry Helix hit one invalid sync source {} {} ", zkStr, syncSrcClusterName, e);
+          }
+        }
+      }
+    }
```
justinlin-linkedin, line 155: I think we might have a better way doing this.
You see, each vcr configuration has vcrSourceDatacenters. This is a list of datacenters (fabrics) that VCR has to replicate blobs from. We can just use the first element in this list to know which datacenter's zookeeper has the partition information. And we already have a map from datacenter to zookeeper url, it's in clusterMapDcsZkConnectString.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -110,6 +118,43 @@ public VcrReplicationManager(CloudConfig cloudConfig, ReplicationConfig replicat
...
+
+      // try the remaining ZK strings if local ZK string doesn't work
+      if (syncSrcZkStr == null) {
+        ArrayList<String> zkStrings = ((HelixClusterManager) clusterMap).getAllDcZkConnectionString();
+        logger.info("Ambry Helix zkstring count {} strings {} ", zkStrings.size(), zkStrings);
+
+        for (String zkStr : zkStrings) {
+          try {
+            HelixAdmin srcAdmin = new ZKHelixAdmin(zkStr);
+            List<String> srcResources = srcAdmin.getResourcesInCluster(syncSrcClusterName);
```
ankagrawal, line 147: I am not sure how resource heavy (in terms of data transferred, latency and work done by helix) is the API `getResourcesInCluster`. Can we maybe use something lightweight like `ZkUtil.isClusterSetup()` here, since we aren't really going to use the resources.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -110,6 +118,43 @@ public VcrReplicationManager(CloudConfig cloudConfig, ReplicationConfig replicat
...
+            syncSrcZkStr = zkStr;
+            break;
+          } catch (Exception e) {
+            logger.warn("Ambry Helix hit one invalid sync source {} {} ", zkStr, syncSrcClusterName, e);
```
ankagrawal, line 151: If we aren't just failing here, then we can as well do this within the `schduleVcrHelix` method.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
ankagrawal: I think we can do what Justin suggested here, and don't need to do any checks in the constructors. Because we aren't failing the VcrReplicationManger, so we can log the error message when trying to sync. Maybe we can also add a metric can we can be alerted on.

JingQianCloud:
Discard the PR.

PR #2137, opened 2022-06-19
Moving NamedBlob's Get and Delete out to separate handlers

justinlin-linkedin:
We have too many logic stuffed into ambry's core read/write path. This PR would clean that up a bit. This PR moves NamedBlob's Get and Delete logic out from core Get and Delete handling logic but reuse most of the code. In this way, we separate core read/write path from NamedBlob's read/write path.


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -16,6 +16,7 @@
 import com.codahale.metrics.MetricRegistry;
 import com.github.ambry.commons.Callback;
 import com.github.ambry.commons.CallbackUtils;
+import com.github.ambry.commons.FutureUtils;
```
mikeyang923, line 19: Is this 'FutureUtils' auto added by IDE? looks it's not in use in this file.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -74,7 +74,9 @@ class FrontendRestRequestService implements RestRequestService {
   private SecurityService securityService = null;
   private GetPeersHandler getPeersHandler;
   private GetSignedUrlHandler getSignedUrlHandler;
-  private NamedBlobListHandler listNamedBlobsHandler;
+  private NamedBlobListHandler namedBlobListHandler;
+  private NamedBlobGetHandler namedBlobGetHandler;
+  private NamedBlobDeleteHandler namedBlobDeleteHandler;
   private GetBlobHandler getBlobHandler;
   private PostBlobHandler postBlobHandler;
   private NamedBlobPutHandler namedBlobPutHandler;
```
mikeyang923, line 82: nit, for better readability, the 'NamedBlobPutHandler' (line 82) can be moved to be right after 'NamedBlobDeleteHandler' (line 79) so that the four NamedBlob handlers are together.

mikeyang923 (approved):
lgtmbc.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
```diff
@@ -0,0 +1,109 @@
...
+    NamedBlobPath namedBlobPath = NamedBlobPath.parse(namedBlobId, Collections.emptyMap());
+    // This is a three steps operation
+    // 1. Get regular blob id from named blob id.
+    // 2. Delete regular blob id from Router
+    // 3. Delete database row.
+    // If we failed at step 1, or step 1 returns NotFound, then  we just return the error back to client, indicating failure
+    // If we failed at step 2, client can resend the request to ambry so we can resume the operation. If regular blob is already
+    //    deleted, Router would do a nop and return success
+    // If we failed at step 3, client can still resend the delete request to delete this blob again.
+    namedBlobDb.get(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(), namedBlobPath.getBlobName())
```
SophieGuo410, line 79: Would it be better if we call convertId in AmbryIdConverterFactory for all namedBlobPath -> blobId mapping?  And inside the NamedBlobDB::delete, we already handle if namedBlob not found issue, do we still need to get the blob first?

Arun-LinkedIn: +1. Yeah, wondering if it would be better to handle namedBlobPath -> blobId operations using `AmbryIdConverter` to hide the namedBlobDb operations from rest of the classes. Since the request path is replaced once the conversion is done, regular blob delete/get handling convert() methods would act on regular blob ID.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            callback.onCompletion(null, Utils.extractFutureExceptionCause(exception));
+            return;
+          }
+          RequestPath newRequestPath =
+              new RequestPath(requestPath.getPrefix(), requestPath.getClusterName(), requestPath.getPathAfterPrefixes(),
+                  "/ " + blobId, requestPath.getSubResource(), requestPath.getBlobSegmentIdx());
+          // Replace RequestPath in the RestRequest and call GetBlobHandler.handle.
+          restRequest.setArg(InternalKeys.REQUEST_PATH, newRequestPath);
+          try {
+            accountAndContainerInjector.injectAccountAndContainerForNamedBlob(restRequest,
```
Arun-LinkedIn, line 92: Like Sophie suggested [here](https://github.com/linkedin/ambry/pull/2137/files#r904272769), would it be better to check for account and container at the beginning so that we avoid network call to NamedBlob DB in case account or container is invalid.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
```diff
@@ -0,0 +1,109 @@
...
+      throws RestServiceException {
+    RequestPath requestPath = getRequestPath(restRequest);
+    // Get blob id and reconstruct RequestPath
+    // this is the three-part named blob id, including account name, container name and the custom blob name.
+    String namedBlobId = requestPath.getOperationOrBlobId(false);
+    NamedBlobPath namedBlobPath = NamedBlobPath.parse(namedBlobId, Collections.emptyMap());
+    // This is a three steps operation
+    // 1. Get regular blob id from named blob id.
+    // 2. Delete regular blob id from Router
+    // 3. Delete database row.
```
Arun-LinkedIn, line 74: @SophieGuo410 can correct me. Seems like currently it is 2 step process, we delete (soft delete) the Named blob ID first and then delete the regular Blob ID. In case step 2 fails, it seems like the blob could be orphaned (Not sure though if this is actually the case. May be we need to implement to delete orphaned blobs asynchronously)?

If we make it as 3 step process to GET blob ID first from NamedBlobDB, delete the regular blob ID from Ambry and delete DB row, seems like we won't need need asynchronous orphaned blob clean up. But, would it increase the DELETE latency due to 2 network calls to DB?

SophieGuo410: Once Mike's update named blob feature is done, I think we are able to handle it in a better way. We will memorize the old blob in namedBlobDB until it gets purged by background deleter. 
One thing I want to point out is, if step3 failed, we will return fail the customer, which means if delete failed, user should still be able to get the blob. But in this sequence, the real blob will be gone, which I think will bring trouble.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
```diff
@@ -0,0 +1,109 @@
...
+        .thenApply(NamedBlobRecord::getBlobId)
+        .whenComplete((blobId, exception) -> {
+          if (exception != null) {
+            callback.onCompletion(null, Utils.extractFutureExceptionCause(exception));
+            return;
+          }
+          RequestPath newRequestPath =
+              new RequestPath(requestPath.getPrefix(), requestPath.getClusterName(), requestPath.getPathAfterPrefixes(),
+                  "/ " + blobId, requestPath.getSubResource(), requestPath.getBlobSegmentIdx());
+          // Replace RequestPath in the RestRequest and call GetBlobHandler.handle.
```
mikeyang923, line 89: Typo: 'call GetBlobHandler.handle' ----> 'call DeleteBlobHandler.handle'

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
justinlin-linkedin: the contract of delete is not like this. In regular blobs, once the client issues a delete, it's considered as deleted, whether ambry returns 202 or 503. We should use the same logic here.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
justinlin-linkedin: I disagree. This is one of the major reason why I am doing this, which is to separate named blob operations into it's own components, so that if one day we don't want named blob operations, we can just remove this file and not worry about id converter since id converter has nothing to do with named blob. 

That's how I see named blob, it's an add-on, extended component, not part of core ambry read/write path. Core ambry should know as little as possible about named blob. 

I will admit this is opinionated. We can talk more about this offline.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
```diff
@@ -0,0 +1,128 @@
...
+          }
+          RequestPath newRequestPath =
+              new RequestPath(requestPath.getPrefix(), requestPath.getClusterName(), requestPath.getPathAfterPrefixes(),
+                  "/ " + blobId, requestPath.getSubResource(), requestPath.getBlobSegmentIdx());
+          // Replace RequestPath in the RestRequest and call DeleteBlobHandler.handle.
+          restRequest.setArg(InternalKeys.REQUEST_PATH, newRequestPath);
+          try {
+            deleteBlobHandler.handle(restRequest, restResponseChannel, (v, e) -> {
+              if (e != null) {
+                callback.onCompletion(null, e);
```
Arun-LinkedIn, line 115: If step 3 fails and user reattempts Delete operation, we may get 404 not found from router. To handle that, would we need to check for 404 here and delete named blob DB row?

Arun-LinkedIn (commented):
Looks good to me overall. Just had one additional comment.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
justinlin-linkedin: I suppose we can do that.

Arun-LinkedIn (approved):
lgtm!

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobDeleteHandler.java
```diff
@@ -0,0 +1,105 @@
...
+            callback.onCompletion(null, dbException);
+            return;
+          }
+          // What if this blobName is already deleted?
+          // We still want to issue DELETE to router, since router might still have a valid blob because we delete blobName
+          // from database before deleting blob from router. We might fail after we delete from database and the client retries
+          // delete. Even if client doesn't retry delete, we still have a background job to delete the orphan blob from router.
+          RequestPath newRequestPath =
+              new RequestPath(requestPath.getPrefix(), requestPath.getClusterName(), requestPath.getPathAfterPrefixes(),
+                  "/ " + blobId, requestPath.getSubResource(), requestPath.getBlobSegmentIdx());
```
SophieGuo410, line 95: QQ: there's a space after slash, would that be a problem?

justinlin-linkedin: thank you for pointing this out, now the question is, why the tests worked??

justinlin-linkedin: I got it now, base64 decoder would ignore the leading space ....

SophieGuo410 (approved):

PR #2138, opened 2022-06-23
[Frontend] Ambry blob metadata cache - 1

snalli:
This patch adds a cache to the non-blocking router. The cache saves
metadata of composite blobs.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -578,11 +578,30 @@ public class RouterConfig {
   @Default("15*1000")
   public final long routerNotFoundCacheTtlInMs;
 
+  public static final String ROUTER_BLOB_METADATA_CACHE_ID = "router.blob.metadata.cache.id";
+  @Config(ROUTER_BLOB_METADATA_CACHE_ID)
+  public final String routerBlobMetadataCacheId;
+
+  public static final String ROUTER_BLOB_METADATA_CACHE_ENABLED = "router.blob.metadata.cache.enabled";
+  @Config(ROUTER_BLOB_METADATA_CACHE_ENABLED)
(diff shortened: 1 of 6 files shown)


ambry-commons/src/main/java/com/github/ambry/commons/AmbryCacheObject.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.github.ambry.commons;
+
+/**
+ * Implements a generic cache object for the AmbryCache
+ */
+public class AmbryCacheObject {
```
justinlin-linkedin, line 20: please set this to an interface.

mikeyang923:
So this is to add the cache, but the cache is not in use yet? Will you raise another PR to plug the cache into the ambry read pipeline?

snalli:
> So this is to add the cache, but the cache is not in use yet? Will you raise another PR to plug the cache into the ambry read pipeline?

Yes Mike, I plan for it to be 3 part PR where I'll integrate the cache into Ambry frontend. Based on some previous attempts, I realized the PR is going to be bigger than I originally thought and it'd be easier to review if I split it.

ambry-commons/src/main/java/com/github/ambry/commons/AmbryCacheObject.java
snalli: Done

mikeyang923 (approved):
lgtm

snalli:
> lgtm

Thanks. I accidentally re-clicked request for review after your approval. Not sure if it cancels the approval and re-sends a review request to the same reviewer.

justinlin-linkedin (approved):
LGTM

PR #2139, opened 2022-06-23
Add GetMode in NamedBlobDb to get deleted and expired named blob record

justinlin-linkedin:
In get requests, we can add get-options in the header to fetch deleted and expired regular blob. But we can't do this in the named blob due to NamedBlobDb's lack of support of retrieving deleted and expired named blob record. 

This PR adds that support to NamedBlobDb.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
+++ b/ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
@@ -1,4 +1,4 @@
-/*
+/**
  * Copyright 2020 LinkedIn Corp. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -12,7 +12,6 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  *
  */
-
 package com.github.ambry.named;
 
 import com.github.ambry.frontend.Page;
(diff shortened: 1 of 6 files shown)


ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
```diff
@@ -24,6 +23,39 @@
...
+  /**
+   * Look up a {@link NamedBlobRecord} by name.
+   * @param accountName the name of the account.
+   * @param containerName the name of the container.
+   * @param blobName the name of the blob.
+   * @param mode The {@link GetMode} for this get method.
+   * @return a {@link CompletableFuture} that will eventually contain either the {@link NamedBlobRecord} for the named
+   *         blob or an exception if an error occurred.
+   */
+  CompletableFuture<NamedBlobRecord> get(String accountName, String containerName, String blobName, GetMode mode);
```
SophieGuo410, line 57: Can we directly use getOption instead of the new generated GetMode enum. It sort of duplicate the code, and in the future if the getOption changed, we need to remember to change this enum accordingly avoid confusion, which is sort of an anti pattern to me.

justinlin-linkedin: The reason why I want to have a different Get option in named blob is because I don't want the named blob db relying on http request related object, like different component has different set of errors.

But I suppose if it's one to one mapping and GetOption is generic enough, we can re-use it here.

Will update.

SophieGuo410 (approved):

Arun-LinkedIn (approved):

PR #2140, opened 2022-06-24
Only need N-1 NotFound to return NotFound

justinlin-linkedin:
We only need N - 1 NotFound from originating DC to return NotFound.

N is the number of replicas in originating DC. It includes all replica states including offline for Get, Delete and Update.
This means, we can tolerate up to 1 unavailable replica, whether it's due to disk failure, host failure or just the replica is in bad state.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -161,8 +161,9 @@ class SimpleOperationTracker implements OperationTracker {
         (Map<ReplicaState, List<ReplicaId>>) partitionId.getReplicaIdsByStates(EnumSet.allOf(ReplicaState.class), null);
     List<ReplicaId> eligibleReplicas;
     List<ReplicaId> offlineReplicas = new ArrayList<>();
-    totalOfflineReplicaCount = getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE))
-            .getOrDefault(ReplicaState.OFFLINE, Collections.emptyList()).size();
+    totalOfflineReplicaCount =
+        getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE)).getOrDefault(ReplicaState.OFFLINE,
+            Collections.emptyList()).size();
     allReplicas = allDcReplicasByState.values().stream().flatMap(Collection::stream).collect(Collectors.toList());
     allReplicaCount = allReplicas.size();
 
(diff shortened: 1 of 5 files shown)


ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -227,7 +228,7 @@ class SimpleOperationTracker implements OperationTracker {
     // Order the replicas so that local healthy replicas are ordered and returned first,
     // then the remote healthy ones, and finally the possibly down ones.
     List<? extends ReplicaId> replicas =
-        routerConfig.routerGetEligibleReplicasByStateEnabled ? eligibleReplicas : allReplicas;
+        routerConfig.routerGetEligibleReplicasByStateEnabled ? eligibleReplicas : partitionId.getReplicaIds();
```
JingQianCloud, line 231: What's the reason to change allReplicas to getReplicaIds.
Before looks like we want to have local healthy replicas at the beginning. After the change, the order will be different. Is that a concern?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -337,25 +331,24 @@ class SimpleOperationTracker implements OperationTracker {
 
     int numActiveReplicasInOriginatingDc =
         getEligibleReplicas(originatingDcName, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER)).size();
-    if (routerConfig.routerOperationTrackerTerminateOnNotFoundEnabled && numReplicasInOriginatingDc > 0
+    if (routerConfig.routerOperationTrackerTerminateOnNotFoundEnabled
```
JingQianCloud, line 334: If routerOperationTrackerTerminateOnNotFoundEnabled, with the logic, it'll fix the 503 error.
But what about it's not enabled? Will we still see 503 response?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -337,25 +331,24 @@ class SimpleOperationTracker implements OperationTracker {
...
-      // This condition accounts for following cases:
-      // 1. Intermediate state of moving replicas (there could be 6 replicas in originating dc temporarily).
-      // 2. Looks at all replicas (instead of routerPutSuccessTarget) in originating DC since one of the replicas
-      // in which the blob was PUT originally could be in error state or is being rebuilt from scratch. By looking at
-      // all replicas, we make sure that we don't miss the other replica in which blob was PUT.
-      // 3. Uses this feature only if there are at least 'routerPutSuccessTarget' number of replicas in leader or
-      // standby state. This ensures that all the replicas in PUT quorum are not down (or being rebuilt) which means
-      // that there is at least one active replica containing the blob.
-      originatingDcNotFoundFailureThreshold = numReplicasInOriginatingDc;
+      // There are two conditions to meet in order to use this feature
```
JingQianCloud, line 336: Just to confirm,
If routerOperationTrackerTerminateOnNotFoundEnabled, with the logic, it'll fix the 503 error.
If it's not enabled, in the case(8 replica returned not found and 1 returned not available), we'll still see 503, right?

JingQianCloud (commented):
Have some questions embedded.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
justinlin-linkedin: the order will be enforced the the code after this line.

The reason for this change is for test. In some tests, we set up error responses in certain order that is different from the allReplicas, but same with partitionId.getReplicaIds().

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
justinlin-linkedin: that's right.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -337,25 +331,24 @@ class SimpleOperationTracker implements OperationTracker {
...
+      // 1. TerminateOnNotFound has to be enabled.
+      // 2. We need enough active replicas in the originating DC
+      // How many active replicas is considered as enough? We have to consider some extreme cases.
+      // When the put success target is 2, we at least need 2 replicas in active state so we know the replicas are up to
+      // date. If we have only one replica in active state, we might run into a scenario that we are getting NotFound
+      // from the other two inactive replicas and we should not ignore the active one.
+      originatingDcNotFoundFailureThreshold = originatingDcTotalReplicaCount - routerConfig.routerPutSuccessTarget + 1;
```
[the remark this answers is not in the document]
justinlin-linkedin: if the real originating DC is decommissioned, then this blob should already be replicated to all other colos for a long time, which means, we should have this blob in all three replicas in any colo.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: What about the following code? Since the replicas list's order is changed. the order we put to the Pool is different. Is this a concern?
        if (isLocalDcReplica) {
          addToBeginningOfPool(replicaId);
        } 


    for (ReplicaId replicaId : replicas) {
      examinedReplicas.add(replicaId);
      String replicaDcName = replicaId.getDataNodeId().getDatacenterName();
      boolean isLocalDcReplica = replicaDcName.equals(datacenterName);
      boolean isOriginatingDcReplica = replicaDcName.equals(originatingDcName);
      if (isOriginatingDcReplica) {
        numReplicasInOriginatingDc++;
      }
      if (!replicaId.isDown()) {
        if (isLocalDcReplica) {
          addToBeginningOfPool(replicaId);
        } else if (crossColoEnabled && isOriginatingDcReplica) {
          addToEndOfPool(replicaId);
        } else if (crossColoEnabled) {
          backupReplicas.addFirst(replicaId);
        }
      } else {
        if (isLocalDcReplica) {
          downReplicas.addFirst(replicaId);
        } else if (crossColoEnabled) {
          downReplicas.addLast(replicaId);
        }
      }
    }

justinlin-linkedin: and the allReplicas doesn't have a deterministic order anyway. It has replicas from all state, leader, standby, bootstrap, inactive, offline. When allReplicas is created, it's created by iterating through a unordered map, which doesn't guarantee the order of each state.

JingQianCloud (approved):

Arun-LinkedIn (approved):

PR #2141, opened 2022-06-24
[WIP] First step to simplify the SimpleOperationTracker

justinlin-linkedin:
In this PR, I made some changes to the SimpleOperationTracker to make it simpler.

1. Remove the cloud code
2. Remove the quota rejection code. Quota rejection should be done in the Operation itself.
3. Remove the code to return 503 when there is a offline replicas in originating DC. We can remove this code by adding all originating DC's replicas to the replicaPool. This way, we would not ignore the offline replicas in TTL_UPDATE and DELETE. The reason why we added this logic in the first place is because some replica is restarting, and it's marked as OFFLINE in helix. If we add all the replicas in originating DC regardless of its state, then we will hit this restarting replica and get a network error, which will return 503.
4. Remove the cross-colo not found logic. Let's rely on originating DC not found logic.
5. Change the originating DC not found logic. This time, we will go through all the replicas in the originating DC and then see how many NOT FOUND error we've seen. This way, we will never miss any replicas in originating DC.


snalli:
Will need some time to understand this. May sync with you offline

PR #2142, opened 2022-06-27
[Frontend] Ambry blob metadata cache - 2

snalli:
This patch integrates adds metadata cache to frontend non-blocking
router. We conditionally lookup cache on a RANGE request and use
cached metadata. We delete cached metadata during DELETE request.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -589,7 +589,12 @@ public class RouterConfig {
   public static final String ROUTER_BLOB_METADATA_CACHE_MAX_SIZE_BYTES = "router.blob.metadata.cache.max.size.bytes";
   @Config(ROUTER_BLOB_METADATA_CACHE_MAX_SIZE_BYTES)
   public final long routerBlobMetadataCacheMaxSizeBytes;
-  long numBytesInOneMb = (long) Math.pow(1024, 2);
+  public static final long NUM_BYTES_IN_ONE_MB = (long) Math.pow(1024, 2);
+
+  public static final String ROUTER_SMALLEST_BLOB_FOR_METADATA_CACHE = "router.smallest.blob.for.metadata.cache";
+  @Config(ROUTER_SMALLEST_BLOB_FOR_METADATA_CACHE)
+  public final long routerSmallestBlobForMetadataCache;
+  public static final long NUM_BYTES_IN_ONE_TB = (long) Math.pow(1024, 4);
 
   /**
    * Create a RouterConfig instance.
@@ -601,7 +606,9 @@ public class RouterConfig {
     routerBlobMetadataCacheEnabled =
         verifiableProperties.getBoolean(ROUTER_BLOB_METADATA_CACHE_ENABLED, false);
     routerBlobMetadataCacheMaxSizeBytes =
-        verifiableProperties.getLong(ROUTER_BLOB_METADATA_CACHE_MAX_SIZE_BYTES, 64 * numBytesInOneMb);
+        verifiableProperties.getLong(ROUTER_BLOB_METADATA_CACHE_MAX_SIZE_BYTES, 64 * NUM_BYTES_IN_ONE_MB);
+    routerSmallestBlobForMetadataCache =
+        verifiableProperties.getLong(ROUTER_SMALLEST_BLOB_FOR_METADATA_CACHE, NUM_BYTES_IN_ONE_TB);
     routerScalingUnitCount = verifiableProperties.getIntInRange(ROUTER_SCALING_UNIT_COUNT, 1, 1, Integer.MAX_VALUE);
     routerHostname = verifiableProperties.getString(ROUTER_HOSTNAME);
     routerDatacenterName = verifiableProperties.getString(ROUTER_DATACENTER_NAME);
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -15,6 +15,7 @@ package com.github.ambry.router;
 
 import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.commons.AmbryCache;
 import com.github.ambry.commons.BlobId;
 import com.github.ambry.commons.BlobIdFactory;
 import com.github.ambry.commons.Callback;
@@ -123,6 +124,8 @@ class GetBlobOperation extends GetOperation {
   private BlobDataReadableStreamChannel blobDataChannel;
   // the CompositeBlobInfo that will be set if (and when) this blob turns out to be a composite blob.
   private CompositeBlobInfo compositeBlobInfo;
+  // A cache for blob metadata of composite blobs
+  private AmbryCache blobMetadataCache;
 
   /**
    * Construct a GetBlobOperation
@@ -140,18 +143,75 @@ class GetBlobOperation extends GetOperation {
    * @param cryptoJobHandler {@link CryptoJobHandler} to assist in the execution of crypto jobs
    * @param time the Time instance to use.
    * @param isEncrypted if the encrypted bit is set based on the original blobId string of a {@link BlobId}.
+   * @param blobMetadataCache A cache to save blob metadata for composite blobs
    */
   GetBlobOperation(RouterConfig routerConfig, NonBlockingRouterMetrics routerMetrics, ClusterMap clusterMap,
       ResponseHandler responseHandler, BlobId blobId, GetBlobOptionsInternal options,
       Callback<GetBlobResultInternal> callback, RouterCallback routerCallback, BlobIdFactory blobIdFactory,
       KeyManagementService kms, CryptoService cryptoService, CryptoJobHandler cryptoJobHandler, Time time,
-      boolean isEncrypted, QuotaChargeCallback quotaChargeCallback) {
+      boolean isEncrypted, QuotaChargeCallback quotaChargeCallback, AmbryCache blobMetadataCache) {
     super(routerConfig, routerMetrics, clusterMap, responseHandler, blobId, options, callback, kms, cryptoService,
         cryptoJobHandler, time, isEncrypted);
     this.routerCallback = routerCallback;
     this.blobIdFactory = blobIdFactory;
     this.quotaChargeCallback = quotaChargeCallback;
-    firstChunk = new FirstGetChunk();
+    this.blobMetadataCache = blobMetadataCache;
+    BlobMetadata blobMetadata = shouldLookupMetadataCache() ? (BlobMetadata) blobMetadataCache.getObject(blobId.toString()) : null;
+    firstChunk = (blobMetadata == null) ? new FirstGetChunk() : new CachedFirstChunk(blobMetadata);
+  }
+
+  /**
+   * Returns whether the current request is a range request or not.
+   * @return True if range request, else False.
+   */
+  boolean isRangeRequest() {
+    return options.getBlobOptions.getRange() != null;
+  }
+
+  /**
+   * Decides if we should look up the metadata cache during a GET request
+   * @return True if and only if
+   *  - cache exists
+   *  - it is a RANGE request
+   *  - it is not a raw mode request
+   *  - it is not a GET request for a segment
+   *  - the blobID is for a metadata blob (this info is present only in v5 and v6 version of blobID)
+   *  False, otherwise.
+   */
+  boolean shouldLookupMetadataCache() {
+    return (blobMetadataCache != null &&
+        isRangeRequest() &&
+        options.getBlobOptions.hasBlobSegmentIdx() == false &&
+        options.getBlobOptions.isRawMode() == false &&
+        blobId.getBlobDataType() == BlobId.BlobDataType.METADATA);
+  }
+
+  /**
+   * Conditionally saves blob metadata for composite blobs
+   * @return True if metadata was saved successfully, else False.
+   */
+  boolean saveMetadata() {
+    boolean putResult = false;
+    if (firstChunk.shouldSaveMetadata()) {
+      BlobMetadata blobMetadata = new BlobMetadata(blobId.toString(), blobInfo, compositeBlobInfo);
+      putResult = blobMetadataCache.putObject(blobMetadata.getBlobId(), blobMetadata);
+      logger.trace("Saving metadata for blobId = {}, result = {}", blobId, putResult);
+    }
+    return putResult;
+  }
+
+  /**
+   * Unconditionally deletes blob metadata for composite blobs.
+   * @param reason String describing reason for deletion, useful for triaging.
+   * @return True if metadata was deleted successfully, else False.
+   */
+  boolean deleteMetadata(String reason) {
+    if (blobMetadataCache == null) {
+      return false;
+    }
+    boolean deleteResult = blobMetadataCache.deleteObject(blobId.toString());
+    logger.trace("Deleting metadata for blobId = {}, reason = {}, result = {}", blobId, reason, deleteResult);
+    return deleteResult;
   }
 
   /**
@@ -201,6 +261,15 @@ class GetBlobOperation extends GetOperation {
         blobDataChannel.completeRead();
       }
     }
+    if (abortCause != null && abortCause instanceof RouterException) {
+      switch(((RouterException) abortCause).getErrorCode()) {
+        /* If the blob is not found, then delete its metadata from frontend cache. */
+        case BlobDoesNotExist:
+        case BlobDeleted:
+        case BlobExpired:
+          deleteMetadata(((RouterException) abortCause).getErrorCode().toString());
+      }
+    }
     setOperationCompleted();
   }
 
@@ -219,6 +288,8 @@ class GetBlobOperation extends GetOperation {
     if (chunk == firstChunk) {
       if (operationCallbackInvoked.compareAndSet(false, true)) {
         if (chunk.chunkBlobId.getBlobDataType() == BlobId.BlobDataType.METADATA) {
+          /* Save metadata on successful completion. Everything is decrypted at this point. */
+          saveMetadata();
           routerMetrics.getMetadataChunkLatencyMs.update(time.milliseconds() - chunk.initializedTimeMs);
         }
         if (options.getChunkIdsOnly) {
@@ -616,7 +687,7 @@ class GetBlobOperation extends GetOperation {
     // size of the chunk
     private long chunkSize;
     // whether the operation on the current chunk has completed.
-    private boolean chunkCompleted;
+    protected boolean chunkCompleted;
     // Tracks quota charging for this chunk.
     private OperationQuotaCharger operationQuotaCharger;
     protected long initializedTimeMs;
@@ -1273,16 +1344,41 @@ class GetBlobOperation extends GetOperation {
   private class FirstGetChunk extends GetChunk {
 
     // refers to the blob type.
-    private BlobType blobType;
-    private List<CompositeBlobInfo.ChunkMetadata> chunkMetadataList;
-    private BlobProperties serverBlobProperties;
-    private short lifeVersion;
+    protected BlobType blobType;
+    protected List<CompositeBlobInfo.ChunkMetadata> chunkMetadataList;
+    protected BlobProperties serverBlobProperties;
+    protected short lifeVersion;
 
     /**
      * Construct a FirstGetChunk and initialize it with the {@link BlobId} of the overall operation.
      */
     FirstGetChunk() {
       super(-1, new CompositeBlobInfo.ChunkMetadata(blobId, 0L, -1L));
+    }
+
+    /**
+     * Decides if we should cache metadata during a GET request
+     * This method should be called after decryption is complete.
+     * @return True if and only if
+     *  - cache exists
+     *  - it's a RANGE request
+     *  - it's not a GET request for a segment
+     *  - it is not a raw mode request
+     *  - the blobID is for a metadata blob (this info is present only in v5 and v6 version of blobID)
+     *  - the blob type is metadata blob
+     *  - firstChunk is complete
+     *  - there are no exceptions
+     *  False, otherwise.
+     */
+    protected boolean shouldSaveMetadata() {
+      return (blobMetadataCache != null &&
+          isRangeRequest() &&
+          options.getBlobOptions.hasBlobSegmentIdx() == false &&
+          options.getBlobOptions.isRawMode() == false &&
+          blobId.getBlobDataType() == BlobId.BlobDataType.METADATA &&
+          blobType == BlobType.MetadataBlob &&
+          totalSize >= routerConfig.routerSmallestBlobForMetadataCache &&
+          isComplete() && getChunkException() == null && getOperationException() == null);
     }
 
     /**
@@ -1568,7 +1664,7 @@ class GetBlobOperation extends GetOperation {
     /**
      * Initialize data chunks and few other cast for metadata chunk
      */
-    private void initializeDataChunks() {
+    protected void initializeDataChunks() {
       if (options.getChunkIdsOnly
           || options.getBlobOptions.getOperationType() == GetBlobOptions.OperationType.BlobInfo) {
         chunkIdIterator = null;
@@ -1644,6 +1740,19 @@ class GetBlobOperation extends GetOperation {
     }
 
     /**
+     * Wrapper over resolveRange
+     * @param rangeTotalSize byte range to be used for range resolution
+     * @return {@code true} if range resolution succeeded or no range resolution was done. {@code false} otherwise
+     */
+    protected boolean resolveRangeProper(long rangeTotalSize) {
+      /*
+       * resolveRange is a weird function that returns false on success and true on failure !
+       * Hence, this wrapper to make code readable !
+       */
+      return !resolveRange(rangeTotalSize);
+    }
+
+    /**
      * On an invalid range, set a {@link RouterErrorCode#RangeNotSatisfiable} exception for this chunk, mark the chunk
      * as unconditionally completed, and set the chunk counters such that the operation will be completed.
      * @param exception the reason that the range was invalid.
@@ -1656,6 +1765,113 @@ class GetBlobOperation extends GetOperation {
       dataChunks = null;
       numChunksTotal = 0;
       numChunksRetrieved.set(0);
+    }
+  }
+
+  private class CachedFirstChunk extends FirstGetChunk {
+
+    /**
+     * Construct a CachedFirstChunk and initialize it with the {@link BlobMetadata} of the overall operation.
+     * @param blobMetadata Cached blob metadata for composite blobs
+     */
+    CachedFirstChunk(BlobMetadata blobMetadata) {
+      super();
+      /*
+       * If cache hit, extract all relevant metadata and save in appropriate variables.
+       * If the entry gets evicted later, we are still safe as we have saved all the metadata.
+       */
+      blobType = BlobType.MetadataBlob;
+      compositeBlobInfo = blobMetadata.getCompositeBlobInfo();
+      totalSize = compositeBlobInfo.getTotalSize();
+      chunkMetadataList = compositeBlobInfo.getChunkMetadataList();
+      blobInfo = blobMetadata.getBlobInfo();
+      lifeVersion = blobInfo.getLifeVersion();
+      serverBlobProperties = blobInfo.getBlobProperties();
+      chunkIndexToBuf = new ConcurrentHashMap<>();
+      chunkIndexToBufWaitingForRelease = new ConcurrentHashMap<>();
+      successfullyDeserialized = true;
+    }
+
+    /**
+     * Return {@link MessageFormatFlags} to associate with cached first chunk.
+     * For a cached metadata chunk, we just need to get blobInfo to validate it.
+     * We don't have to fetch the entire metadata chunk again.
+     * @return {@link MessageFormatFlags#BlobInfo}
+     */
+    @Override
+    MessageFormatFlags getOperationFlag() {
+      return MessageFormatFlags.BlobInfo;
+    }
+
+    /**
+     * Decides if we should save metadata or not.
+     * @return False for cached first chunk. There is nothing to cache.
+     */
+    @Override
+    protected boolean shouldSaveMetadata() { return false; }
+
+    /**
+     * On an invalid cache entry, set a {@link RouterErrorCode} exception for first chunk, mark the chunk
+     * as completed, and set the chunk counters such that the operation will be completed.
+     * @param routerException Exception due to which cache entry was invalid.
+     */
+    protected void onInvalidCacheEntry(RouterException routerException) {
+      setChunkException(routerException);
+      chunkCompleted = true;
+      retainChunkExceptionOnSuccess = true;
+      chunkIdIterator = null;
+      dataChunks = null;
+      numChunksTotal = 0;
+      numChunksRetrieved.set(0);
+    }
+
+    /**
+     * Validate cached metadata chunk and proceed to request data chunks
+     * @param payload Bytes received from backend
+     * @param messageMetadata  Metadata associated with messages sent out
+     * @param messageInfo Message info class that contains basic info about a message
+     * @throws IOException
+     * @throws MessageFormatException
+     */
+    @Override
+    void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInfo messageInfo)
+        throws IOException, MessageFormatException {
+      /* If we find blob metadata in frontend cache, validate it and use it. */
+      BlobProperties receivedBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
+      /*
+       * Some cached variables are mutable.
+       * Update them and then compare if other immutable variables are same or not.
+       */
+      updateTtlIfRequired(serverBlobProperties, messageInfo);
+      lifeVersion = messageInfo.getLifeVersion();
+      blobInfo.setLifeVersion(lifeVersion);
+      if (!serverBlobProperties.equals(receivedBlobProperties)) {
+        /*
+         * This is a rare error case.
+         * If we find blob properties is in our frontend cache, we request just blobInfo from backend to validate
+         * the cache entry. If cached blob properties does not match received blob properties, due to a random bit flip for eg.,
+         * we must delete the cache entry and throw a retriable exception to the user.
+         * The next request will re-fill the cache with a valid entry if the blob stills exists.
+         * We do not need to decrypt anything at this point.
+         */
+        String reason = "Cached blob property does not match received blob property";
+        logger.error("{} for blobId = {}, cached blob property = {}, received blob property = {}", reason, blobId,
+            serverBlobProperties, receivedBlobProperties);
+        deleteMetadata(reason);
+        onInvalidCacheEntry(new RouterException(reason, RouterErrorCode.UnexpectedInternalError));
+        return;
+      }
+      getOptions().ageAtAccessTracker.trackAgeAtAccess(serverBlobProperties.getCreationTimeInMs());
+
+      /*
+       * If we reach here, then the cached metadata is valid and already decrypted from a previous GET/RANGE request.
+       * Just use it directly and initiate fetching data chunks.
+       */
+      if (isRangeRequest() && resolveRangeProper(totalSize)) {
+        chunkMetadataList = compositeBlobInfo.getStoreKeysInByteRange(resolvedByteRange.getStartOffset(),
+            resolvedByteRange.getEndOffset());
+        initializeDataChunks();
+      }
     }
   }
 }
--- a/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
@@ -97,6 +97,14 @@ class NonBlockingRouter implements Router {
     ResponseHandler responseHandler = new ResponseHandler(clusterMap);
     this.kms = kms;
     this.cryptoJobHandler = cryptoJobHandler;
+    /*
+     * Initialize blobMetadata cache before the operation controllers
+     * as it will be passed further down and we don't want to pass down a null value.
+     */
+    blobMetadataCache = new AmbryCache(routerConfig.routerBlobMetadataCacheId,
+        routerConfig.routerBlobMetadataCacheEnabled,
+        routerConfig.routerBlobMetadataCacheMaxSizeBytes,
+        routerMetrics.getMetricRegistry());
     ocCount = routerConfig.routerScalingUnitCount;
     ocList = new ArrayList<>();
     for (int i = 0; i < ocCount; i++) {
@@ -122,10 +130,6 @@ class NonBlockingRouter implements Router {
         .build();
     routerMetrics.initializeNotFoundCacheMetrics(notFoundCache);
     routerMetrics.initializeQuotaOCMetrics(ocList);
-    blobMetadataCache = new AmbryCache(routerConfig.routerBlobMetadataCacheId,
-        routerConfig.routerBlobMetadataCacheEnabled,
-        routerConfig.routerBlobMetadataCacheMaxSizeBytes,
-        routerMetrics.getMetricRegistry());
   }
 
   /**
@@ -133,6 +137,14 @@ class NonBlockingRouter implements Router {
    */
   Cache<String, Boolean> getNotFoundCache() {
     return notFoundCache;
+  }
+
+  /**
+   * Returns an instance of blob metadata cache.
+   * @return Returns an instance of blobMetadata cache
+   */
+  public AmbryCache getBlobMetadataCache() {
+    return blobMetadataCache;
   }
 
   /**
@@ -502,6 +514,9 @@ class NonBlockingRouter implements Router {
           deleteRequests.add(new BackgroundDeleteRequest(storeKey, serviceId, quotaChargeCallback));
         }
         initiateBackgroundDeletes(deleteRequests);
+        if (blobMetadataCache != null) {
+          blobMetadataCache.deleteObject(blobIdStr);
+        }
       }
       currentBackgroundOperationsCount.decrementAndGet();
     };
--- a/ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
+++ b/ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
@@ -15,6 +15,7 @@ package com.github.ambry.messageformat;
 
 import com.github.ambry.utils.SystemTime;
 import com.github.ambry.utils.Utils;
+import java.util.Objects;
 
 
 /**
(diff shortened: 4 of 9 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1375,6 +1471,16 @@ protected void maybeProcessCallbacks() {
     @Override
     void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInfo messageInfo)
         throws IOException, MessageFormatException {
+
+      if (isMetadataCacheHit()) {
+        getOptions().ageAtAccessTracker.trackAgeAtAccess(serverBlobProperties.getCreationTimeInMs());
+        handleMetadataBlob(null, blobInfo.getUserMetadata(), null);
+        // Do not call updateTtlIfRequired(). If the blob expires, the frontend will find out from backend during the next GET request
+        // TODO: Deserialize payload and do basic validations to check if cached blob properties are correct.
+        // TODO: Should we decrypt ? Probably no.
```
snalli, line 1480: One thing I am not very sure of is what to do if the blobInfo we fetch to validate the cache entry is corrupted or not found. Do we invalidate the cache entry and retry to fetch the entire metadata from backend ? @justinlin-linkedin

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -591,6 +591,11 @@ public class RouterConfig {
   public final long routerBlobMetadataCacheMaxSizeBytes;
   long numBytesInOneMb = (long) Math.pow(1024, 2);
 
+  public static final String ROUTER_SMALLEST_BLOB_FOR_METADATA_CACHE = "router.smallest.blob.for.metadata.cache";
+  @Config(ROUTER_SMALLEST_BLOB_FOR_METADATA_CACHE)
+  public final long routerSmallestBlobForMetadataCache;
+  long numBytesInOneTb = (long) Math.pow(1024, 4);
```
justinlin-linkedin, line 597: we can use a static value for this and change the variable name to all upper case. 

And 1 TB, really?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1278,13 +1282,100 @@ private class FirstGetChunk extends GetChunk {
...
+     *  - cache exists
+     *  - it is a RANGE request
+     *  - it is not a raw mode request
+     *  - the blobID is for a metadata blob
+     *  - the cache not yet been consulted
+     *  False, otherwise.
+     */
+    boolean shouldLookupMetadataCache() {
+      boolean result = (blobMetadataCache != null &&
+          options.getBlobOptions.getRange() != null &&
```
justinlin-linkedin, line 1325: I think we can remove the `options.getBlobOptions.getRange() != null`. Even if we are trying to get the whole blob, we can still leverage cache.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1278,13 +1282,100 @@ private class FirstGetChunk extends GetChunk {
...
+     * @return True if and only if
+     *  - cache exists
+     *  - it is a RANGE request
+     *  - it is not a raw mode request
+     *  - the blobID is for a metadata blob
+     *  - the cache not yet been consulted
+     *  False, otherwise.
+     */
+    boolean shouldLookupMetadataCache() {
+      boolean result = (blobMetadataCache != null &&
```
justinlin-linkedin, line 1324: we don't need a boolean variable, we can jus return this statement.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1278,13 +1282,100 @@ private class FirstGetChunk extends GetChunk {
...
+        if (blobMetadata != null) {
+          blobMetadataCacheEntryStatus = AmbryCacheEntryStatus.cacheHit;
+          blobType = BlobType.MetadataBlob;
+          compositeBlobInfo = blobMetadata.getCompositeBlobInfo();
+          blobInfo = blobMetadata.getBlobInfo();
+          lifeVersion = blobInfo.getLifeVersion();
+          serverBlobProperties = blobInfo.getBlobProperties();
+          chunkIndexToBuf = new ConcurrentHashMap<>();
+          chunkIndexToBufWaitingForRelease = new ConcurrentHashMap<>();
+          successfullyDeserialized = true;
```
justinlin-linkedin, line 1306: hmm, this is tricky, we need to set FirstGetChunk to ChunkState.Complete so we don't have to send requests to server nodes. But that being said, we still have to get chunkMetadataList and set chunkIdIterator.  Check out code from line 1616 to line 1636 and method initializeDataChunks.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1278,13 +1282,100 @@ private class FirstGetChunk extends GetChunk {
...
+     *  - cache exists
+     *  - it is RANGE request
+     *  - it is not a raw mode request
+     *  - the blobID is for a metadata blob
+     *  - the cache was looked up and we got a cache miss
+     *  False, otherwise.
+     */
+    boolean shouldSaveMetadata() {
+      boolean result = (blobMetadataCache != null &&
+          options.getBlobOptions.getRange() != null &&
```
justinlin-linkedin, line 1345: same above.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1278,13 +1282,100 @@ private class FirstGetChunk extends GetChunk {
...
+    /**
+     * Conditionally saves blob metadata for composite blobs
+     * @return True if metadata was saved successfully, else False.
+     */
+    boolean saveMetadata() {
+      boolean result = false;
+      if (shouldSaveMetadata()) {
+        BlobMetadata blobMetadata = new BlobMetadata(blobId.toString(), blobInfo, compositeBlobInfo);
+        result = blobMetadataCache.putObject(blobMetadata.getBlobId(), blobMetadata);
+        if (result == true) {
```
justinlin-linkedin, line 1362: if (result) {
}

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1308,8 +1399,13 @@ GetOption getGetOption() {
...
     MessageFormatFlags getOperationFlag() {
-      return options.getBlobOptions.getOperationType() == GetBlobOptions.OperationType.Data ? MessageFormatFlags.Blob
-          : MessageFormatFlags.All;
+      if (isMetadataCacheHit()) {
+        return MessageFormatFlags.BlobInfo;
+      } else if (options.getBlobOptions.getOperationType() == GetBlobOptions.OperationType.Data) {
+        return MessageFormatFlags.Blob;
+      } else {
+        return MessageFormatFlags.All;
+      }
```
justinlin-linkedin, line 1408: this operationFlag is for requests to server node, it shouldn't have anything to do with cache.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1375,6 +1471,16 @@ protected void maybeProcessCallbacks() {
...
+
+      if (isMetadataCacheHit()) {
+        getOptions().ageAtAccessTracker.trackAgeAtAccess(serverBlobProperties.getCreationTimeInMs());
+        handleMetadataBlob(null, blobInfo.getUserMetadata(), null);
+        // Do not call updateTtlIfRequired(). If the blob expires, the frontend will find out from backend during the next GET request
+        // TODO: Deserialize payload and do basic validations to check if cached blob properties are correct.
+        // TODO: Should we decrypt ? Probably no.
+        return;
+      }
+
```
justinlin-linkedin, line 1483: handleBody is the method to call when handling response from server node, it shouldn't have anything to do with cache.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1498,10 +1604,13 @@ public List<CompositeBlobInfo.ChunkMetadata> getChunkMetadataList() {
...
-      ByteBuf serializedMetadataContent = blobData.content();
+      ByteBuf serializedMetadataContent = null;
       try {
-        compositeBlobInfo =
-            MetadataContentSerDe.deserializeMetadataContentRecord(serializedMetadataContent.nioBuffer(), blobIdFactory);
+        if (!isMetadataCacheHit()) {
+          serializedMetadataContent = blobData.content();
+          compositeBlobInfo =
+              MetadataContentSerDe.deserializeMetadataContentRecord(serializedMetadataContent.nioBuffer(), blobIdFactory);
+        }
```
justinlin-linkedin, line 1613: same

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1529,7 +1638,7 @@ private void handleMetadataBlob(BlobData blobData, byte[] userMetadata, ByteBuff
           rangeResolutionFailure = true;
         }
         if (!rangeResolutionFailure) {
-          if (options.getChunkIdsOnly || getOperationFlag() == MessageFormatFlags.Blob || encryptionKey == null) {
+          if (options.getChunkIdsOnly || getOperationFlag() == MessageFormatFlags.Blob || isMetadataCacheHit() || encryptionKey == null) {
```
justinlin-linkedin, line 1641: same

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -502,6 +510,7 @@ void initiateChunkDeletesIfAny(final String blobIdStr, final String serviceId,
           deleteRequests.add(new BackgroundDeleteRequest(storeKey, serviceId, quotaChargeCallback));
         }
         initiateBackgroundDeletes(deleteRequests);
+        blobMetadataCache.deleteObject(blobIdStr);
```
justinlin-linkedin, line 513: it's good to remove it from cache in delete, but I think we should remove it from cache in ttl update as well, because ttl update would change the ttl for metadata. So the blob is outdate after ttl update.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -325,6 +328,7 @@ void poll(RequestRegistrationCallback<GetOperation> requestRegistrationCallback)
       }
       if (firstChunk.isComplete()) {
         onChunkOperationComplete(firstChunk);
+        firstChunk.saveMetadata();
```
justinlin-linkedin, line 331: this is not the best place to call saveMetadata. poll method will be invoked by network client every 50 ms, so we would call saveMetadata over and over again. I recommend calling saveMetadata in onChunkOperationComplete method after we check operationCallbackInvoked.

justinlin-linkedin (changes requested):

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: this is the handleBody for the first chunk - metadata chunk. So if we have a cacheHit, we don't need to do anything below this clause. Datachunks have a different handleBody

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: explained above

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: explained above

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
snalli: Yeah we discussed this and Tugrul suggested letting the cache entry remain and let the frontend talk to backend to validate the cache entry during the next RANGE request.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
Arun-LinkedIn: Hmm, are you going with approach of checking cache in GetBlobOperation (and modifying first request to fetch blobInfo instead of metadata chunk) instead of checking cache in NonBlockingRouter to issue separate GetBlobInfo requests? Still thinking that if we go with latter, we wouldn't need with modifications in other methods such as `getOperationFlag`, `handleBody`, etc :)

snalli: To summarise for others:
1/ Check cache in GetBlobOperation and modify code related to FirstGetChunk
2/ Check cache at the Router level. If cache hit, issue GetBlobInfo followed by GetBlob. If cache miss, just issue GetBlob.
Yes I am going with the first approach. All core changes are within FirstGetChunk class. Modifications to methods getOperationFlag, handleBody aren't too difficult. Second approach may avoid this but may require changes elsewhere. We would just be shifting the changes to another place and also scatter it across classes probably. Also, second approach requires callbacks within callbacks which is confusing.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
justinlin-linkedin: The question is more about why do we differentiate metadata chunk at different size? I don't see why we would not cache some metadata chunk but cache the other.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
justinlin-linkedin: That seems to be the case. But we might have multiple clients issuing get requests.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
justinlin-linkedin: we have a atomicboolean to safe guard the code to be only executed once.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: Sorry, my reply above was for shouldSaveMetadata. But for shouldLookupMetadataCache, in reply to your comment, we should not look up cache on each GET request. We discussed this in the design doc and Ashish suggested to lookup the cache conditionally for eg. only on range requests. This makes sense since vast majority of blobs are small, most requests are GET for small blobs and will not see much benefits from caching. Looking up the cache on each GET just adds more latency. I have some perf numbers to back it up.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: actually, no. The result var is the result of inserting an object in the cache and not the result of shouldSaveMetadata. So this code is correct.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: > That seems to be the case. But we might have multiple clients issuing get requests.

Just saw this. I think we both commented at the same time. Yes, you are right, we can get multiple GET requests for the same blob from multiple clients. I will change the code to leverage that but we need to set some threshold for blob size below which we won't cache.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
snalli: Actually, this reminds me that if the blob doesn't exist in the backend, we should invalidate its cache entry in the frontend. I need to add that code.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: I thought more about this. If we save metadata even for every GET request on a blob (subject to some min size), then we may want to look it up for subsequent GETs on the same blob. The problem is that we don't know what the size of a blob is just by looking at the blobID. So then we need to lookup the cache for each GET which just adds extra latency for those blobs whose metadata wasn't cached. I discussed this issue in the design doc and adding latency to GETs that will mostly encounter a cache miss wasn't acceptable.

justinlin-linkedin:
The PR looks good to me, but I want to propose a new way here, can we create a new class called CachedFirstGetChunk. This CachedFirsrGetChunk would extend from FirstGetChunk. So in the GetBlobOperation's constructor, we would check if this MetadataChunk is in the cached and create a CachedFirstGetChunk. And of course we would fall back to FirstGetChunk when the MetadataChunk is not cached. What do you think? Can we try it?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1375,6 +1500,36 @@ protected void maybeProcessCallbacks() {
...
+        BlobProperties receivedBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
+        updateTtlIfRequired(serverBlobProperties, messageInfo);
+        if (serverBlobProperties.toString() != receivedBlobProperties.toString()) {
+          /**
+           * If we find blob properties is in our frontend cache, we request just blobInfo from backend to validate
+           * the cache entry. If cached blob properties does not match received blob properties,
+           * we must delete the cache entry and throw a retriable exception to the user.
+           * The next request will re-fill the cache with a valid entry if the blob stills exists.
+           * We do not need to decrypt anything at this point.
+           */
```
justinlin-linkedin, line 1517: I don't think so.

The blob property, except for ttl, is immutable. So if the blob property from server is different from cache, them it's because the ttl is different, which means client updated ttl before get. This just means that we have to update the ttl in the cache, we don't have to through an error back to client.

snalli: I agree. I update the TTL and then compare the cached and received blob properties. This if-clause is for the rare case when something goes wrong.

snalli:
> The PR looks good to me, but I want to propose a new way here, can we create a new class called CachedFirstGetChunk. This CachedFirsrGetChunk would extend from FirstGetChunk. So in the GetBlobOperation's constructor, we would check if this MetadataChunk is in the cached and create a CachedFirstGetChunk. And of course we would fall back to FirstGetChunk when the MetadataChunk is not cached. What do you think? Can we try it?

This is a great idea ! I have updated my PR. The code however isn't entirely encapsulated in CachedFirstChunk.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1752,107 @@ private void onInvalidRange(Exception exception) {
...
+      compositeBlobInfo = blobMetadata.getCompositeBlobInfo();
+      totalSize = compositeBlobInfo.getTotalSize();
+      chunkMetadataList = compositeBlobInfo.getChunkMetadataList();
+      blobInfo = blobMetadata.getBlobInfo();
+      lifeVersion = blobInfo.getLifeVersion();
+      serverBlobProperties = blobInfo.getBlobProperties();
+      chunkIndexToBuf = new ConcurrentHashMap<>();
+      chunkIndexToBufWaitingForRelease = new ConcurrentHashMap<>();
+      successfullyDeserialized = true;
+      logger.info(String.format("Metadata cache-hit for blobId = %s", blobId));
```
justinlin-linkedin, line 1778: let's do log trace for this. And logger methods are by default formatting your string. we can do it in this way:
```
logger.trace("Metadata cache-hit for blobId = {}", blobId);
```

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1752,107 @@ private void onInvalidRange(Exception exception) {
...
+    }
+
+    /**
+     * Return the {@link MessageFormatFlags} to associate with the first getBlob chunk operation.
+     * @return {@link MessageFormatFlags#Blob} for {@link GetBlobOptions.OperationType#Data}, or
+     *         {@link MessageFormatFlags#All} by default.
+     */
```
justinlin-linkedin, line 1785: we should update the comments here. Let's point out that for cached metadata chunk, we should request blob info from server to verify the existence of this chunk and update TTL.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1752,107 @@ private void onInvalidRange(Exception exception) {
...
+    @Override
+    void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInfo messageInfo)
+        throws IOException, MessageFormatException {
+
+      /**
+       * If we find blob metadata in frontend cache, validate it and use it.
+       */
+      BlobProperties receivedBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
+      updateTtlIfRequired(serverBlobProperties, messageInfo);
+      if (!serverBlobProperties.toString().equals(receivedBlobProperties.toString())) {
```
justinlin-linkedin, line 1810: verify the equality of blob properties should be done in a different way. we should implement an `equal` method for BlobProperties.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1752,107 @@ private void onInvalidRange(Exception exception) {
...
+         * the cache entry. If cached blob properties does not match received blob properties,
+         * we must delete the cache entry and throw a retriable exception to the user.
+         * The next request will re-fill the cache with a valid entry if the blob stills exists.
+         * We do not need to decrypt anything at this point.
+         */
+        String reason = "Cached blob property does not match received blob property";
+        deleteMetadata(reason);
+        logger.error(
+            String.format("%s for blobId = %s, cached blob property = %s, received blob property = %s", reason, blobId,
+                serverBlobProperties.toString(), receivedBlobProperties));
```
justinlin-linkedin, line 1823: same, logger.error has builtin formatting.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1752,107 @@ private void onInvalidRange(Exception exception) {
...
+        throw new MessageFormatException(reason, MessageFormatErrorCodes.IO_Error);
```
justinlin-linkedin, line 1824: we shouldn't throw MessageFormatException here.

Let's figure out why there is a difference between cached blob properties and requested blob properties. The only case i can think of is ttl.

Rather than that, I can't think so any other case. 

If we updated ttl  and they are still different, then something is probably very wrong.

We should terminate this operation by setting chunkException. And MessageFormatException is not a good one, we should use RouterException.

```
      BlobProperties receivedBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
      updateTtlIfRequired(serverBlobProperties, messageInfo);
      lifeVersion = messageInfo.getLifeVersion;
      if (!serverBlobProperties.equals(receivedBlobProperties)) {
            deleteMetadataInCache(reason);
            setChunkException(new RouterException(reason, RouterErrorCode.UnexpectedInternalError));
            return;
      }
```

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1752,107 @@ private void onInvalidRange(Exception exception) {
...
+            chunkMetadataList = compositeBlobInfo.getStoreKeysInByteRange(resolvedByteRange.getStartOffset(),
+                resolvedByteRange.getEndOffset());
+          }
+        }
+        initializeDataChunks();
+      } catch (IllegalArgumentException e) {
+        onInvalidRange(e);
+      }
+    }
+  }
```
justinlin-linkedin, line 1855: can we make a method for this code so we can reuse them?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -140,18 +144,93 @@ class GetBlobOperation extends GetOperation {
...
+  /**
+   * Conditionally saves blob metadata for composite blobs
+   * @return True if metadata was saved successfully, else False.
+   */
+  boolean saveMetadata() {
+    boolean put_result = false;
+    if (shouldSaveMetadata()) {
+      BlobMetadata blobMetadata = new BlobMetadata(blobId.toString(), blobInfo, compositeBlobInfo);
+      put_result = blobMetadataCache.putObject(blobMetadata.getBlobId(), blobMetadata);
+      logger.info(String.format("Saving metadata for blobId = %s, result = %s", blobId, put_result));
```
justinlin-linkedin, line 216: do logger.trace please. And you don't have to call String.format, logger's methods by default do string formatting for you.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -140,18 +144,93 @@ class GetBlobOperation extends GetOperation {
...
+        blobId.getBlobDataType() == BlobId.BlobDataType.METADATA &&
+        totalSize >= routerConfig.routerSmallestBlobForMetadataCache);
+  }
+
+  /**
+   * Conditionally saves blob metadata for composite blobs
+   * @return True if metadata was saved successfully, else False.
+   */
+  boolean saveMetadata() {
+    boolean put_result = false;
```
justinlin-linkedin, line 212: let's rename this variable to `putResult`.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -140,18 +144,93 @@ class GetBlobOperation extends GetOperation {
...
+   * If it was never saved, then we don't have to delete it.
+   * @return True if metadata was deleted successfully, else False.
+   * @param reason
+   */
+  boolean deleteMetadata(String reason) {
+    if (blobMetadataCache == null) {
+      return false;
+    }
+    boolean delete_result = blobMetadataCache.deleteObject(blobId.toString());
+    logger.info(String.format("Deleting metadata for blobId = %s, reason = %s, result = %s", blobId, reason, delete_result));
```
justinlin-linkedin, line 232: same

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
justinlin-linkedin: And set `chunkCompleted` to true as well.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: good catch!

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: got it. I thought of the same first !

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: should we update lifeVersion here using `blobInfo.setLifeVersion(lifeVersion);`

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -201,6 +280,15 @@ void abort(Exception abortCause) {
         blobDataChannel.completeRead();
       }
     }
+    switch(((RouterException) abortCause).getErrorCode()) {
+      /**
+       * If the blob is not found, then delete its metadata from frontend cache.
```
Arun-LinkedIn, line 285: nit: probably use end of line comments ( // If the blob is not found ... ) or block comments (/* * */) for multiple lines since we seem to do that in rest of code?

snalli: agreed, will fix

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -219,6 +316,12 @@ private void onChunkOperationComplete(GetChunk chunk) {
     if (chunk == firstChunk) {
       if (operationCallbackInvoked.compareAndSet(false, true)) {
         if (chunk.chunkBlobId.getBlobDataType() == BlobId.BlobDataType.METADATA) {
+          /**
+           * Save metadata on successful completion. Everything is decrypted at this point.
+           */
```
justinlin-linkedin, line 321: I realize that you have a lot of comments that are using /** */. we should just use // for multiple lines.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1643,12 +1746,25 @@ private boolean resolveRange(long totalSize) {
...
+     * @param dataRangeSize byte range to be used for range resolution
+     * @return {@code true} if range resolution succeeded. {@code false} otherwise
+     */
+    protected boolean resolveRangeProper(long dataRangeSize) {
+      /**
+       * resolveRange is a weird function that returns false on success and true on failure !
+       * Hence, this wrapper to make code readable !
+       */
+      return !resolveRange(dataRangeSize);
+    }
```
justinlin-linkedin, line 1760: or we should just change the return value in resolveRange method.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1774,127 @@ private void onInvalidRange(Exception exception) {
...
+    void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInfo messageInfo)
+        throws IOException, MessageFormatException {
+      // If we find blob metadata in frontend cache, validate it and use it.
+      BlobProperties receivedBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
+      /**
+       * Some cached variables are mutable.
+       * Update them and then compare if other immutable variables are same or not.
+       */
+      updateTtlIfRequired(serverBlobProperties, messageInfo);
+      blobInfo.setLifeVersion(lifeVersion);
```
justinlin-linkedin, line 1847: blobInfo.setLifeVersion(messageInfo.getLifeVersion()).

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1774,127 @@ private void onInvalidRange(Exception exception) {
...
+        onInvalidCacheEntry(new RouterException(reason, RouterErrorCode.UnexpectedInternalError));
+        return;
+      }
+      getOptions().ageAtAccessTracker.trackAgeAtAccess(serverBlobProperties.getCreationTimeInMs());
+
+      /**
+       * If we reach here, then the cached metadata is valid and already decrypted from a previous GET/RANGE request.
+       * Just use it directly and initiate fetching data chunks.
+       */
+      requestDataChunks();
```
justinlin-linkedin, line 1871: can we rename this to something else, just prepareDataChunks...

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1774,127 @@ private void onInvalidRange(Exception exception) {
...
+      // If we find blob metadata in frontend cache, validate it and use it.
+      BlobProperties receivedBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
+      /**
+       * Some cached variables are mutable.
+       * Update them and then compare if other immutable variables are same or not.
+       */
+      updateTtlIfRequired(serverBlobProperties, messageInfo);
+      blobInfo.setLifeVersion(lifeVersion);
+      lifeVersion = messageInfo.getLifeVersion();
+      if (!serverBlobProperties.equals(receivedBlobProperties)) {
```
justinlin-linkedin, line 1849: we have to override BlobProperties's equals method.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1774,127 @@ private void onInvalidRange(Exception exception) {
...
+       * Just use it directly and initiate fetching data chunks.
+       */
+      requestDataChunks();
+    }
+
+    /**
+     * Resolve range, if needed, and request data chunks
+     * @throws IllegalArgumentException
+     */
+    void requestDataChunks() throws IllegalArgumentException {
```
justinlin-linkedin, line 1878: can we reuse this method in FirstGetChunk.handleMetadataChunk?

justinlin-linkedin (commented):
We should add some tests for the cached first chunk.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -140,20 +145,103 @@ class GetBlobOperation extends GetOperation {
...
+   * @return True if and only if
+   *  - cache exists
+   *  - it is a RANGE request
+   *  - it is not a raw mode request
+   *  - the blobID is for a metadata blob
+   *  False, otherwise.
+   */
+  boolean shouldLookupMetadataCache() {
+    return (blobMetadataCache != null &&
+        options.getBlobOptions.getRange() != null &&
```
Arun-LinkedIn, line 186: Should we enable this for segmented GETs too (`options.getBlobOptions.hasBlobSegmentIdx()`) where user can request for segments of a stitched blob via URI like <blobId>/Segment/<SegmentNumber>? It can be useful for that feature too. Looks like you are already handling initializing data chunks for this case in `CachedFirstChunk::requestDataChunks()`

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
Arun-LinkedIn, line 1847: Should this come after line 1848 where updated life version is fetched?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1774,127 @@ private void onInvalidRange(Exception exception) {
...
+      if (options.getBlobOptions.hasBlobSegmentIdx()) {
+        int requestedSegment = options.getBlobOptions.getBlobSegmentIdx();
+        if (requestedSegment < 0 || requestedSegment >= chunkMetadataList.size()) {
+          throw new IllegalArgumentException(
+              "Bad segment number: " + requestedSegment + ", num of keys: " + chunkMetadataList.size());
+        }
+        chunkMetadataList = chunkMetadataList.subList(requestedSegment, requestedSegment + 1);
+        dataRangeSize = chunkMetadataList.get(0).getSize();
+      }
+      if (resolveRangeProper(dataRangeSize)) {
```
Arun-LinkedIn, line 1889: If we decide to look up cache for segmented GETs too in `shouldLookupMetadataCache`, we might need to check if this a range request by `if (options.getBlobOptions.getRange() != null) ` here.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1774,127 @@ private void onInvalidRange(Exception exception) {
...
+       */
+      requestDataChunks();
+    }
+
+    /**
+     * Resolve range, if needed, and request data chunks
+     * @throws IllegalArgumentException
+     */
+    void requestDataChunks() throws IllegalArgumentException {
+      long dataRangeSize = totalSize;
```
Arun-LinkedIn, line 1879: nit: may be rename variable `dataRangeSize` to `rangeTotalSize`

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -219,6 +316,12 @@ private void onChunkOperationComplete(GetChunk chunk) {
     if (chunk == firstChunk) {
       if (operationCallbackInvoked.compareAndSet(false, true)) {
         if (chunk.chunkBlobId.getBlobDataType() == BlobId.BlobDataType.METADATA) {
+          /**
+           * Save metadata on successful completion. Everything is decrypted at this point.
```
Arun-LinkedIn, line 320: nit: same as above. Think it can be implementation comments (either // or /* * */) and not java doc (/** * */) comments.

Arun-LinkedIn (commented):
Mostly looks good. Just had few comments.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -140,20 +145,103 @@ class GetBlobOperation extends GetOperation {
...
+      logger.trace("Saving metadata for blobId = {}, result = {}", blobId, putResult);
+    }
+    return putResult;
+  }
+
+  /**
+   * Unconditionally deletes blob metadata for composite blobs.
+   * If it was never saved, then we don't have to delete it.
+   * @return True if metadata was deleted successfully, else False.
+   * @param reason
```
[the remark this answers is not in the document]
snalli: Good point. The return val from this method is not currently used but, in general, i think its a good practice to return something assuming it may be used in the future.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: I need to understand segmented GETs

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: Yeah i thought of it and its going to break a lot of code

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: yes done

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: correct

snalli: we need to set the lifeVersion member of this class as well.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: Yes we could but I don't want to touch that right now since its very critical to all GETs

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: agreed, i need to read up a bit on segmented GETs

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: done

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: done

snalli:
> We should add some tests for the cached first chunk.

yeah working on it

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: I can send out another PR to reuse this method in handleMetadataBlob

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: I can fix existing code in another PR

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: For this, we only reach this code for a range request. So no need to check getRange() != null. This a policy controlled by shouldLookUpMetadataCache()

snalli:
Addressed all comments. WIP on unit tests.
LMK if I missed any comments.
The policy right now is:
1. Populate cache on each GET/RANGE/SEGMENTed GET if blobSize >= 1 TB (configurable)
2. Lookup cache only for RANGE/SEGMENT requests

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: ok updated code to use cache for segmented GETs also

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: ok updated code to use cache for segmented GETs also

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1658,6 +1751,133 @@ private void onInvalidRange(Exception exception) {
...
+        }
+        chunkMetadataList = chunkMetadataList.subList(requestedSegment, requestedSegment + 1);
+        rangeTotalSize = chunkMetadataList.get(0).getSize();
+      }
+      if (options.getBlobOptions.getRange() != null && resolveRangeProper(rangeTotalSize)) {
+        if (!options.getBlobOptions.hasBlobSegmentIdx()) {
+          chunkMetadataList = compositeBlobInfo.getStoreKeysInByteRange(resolvedByteRange.getStartOffset(),
+              resolvedByteRange.getEndOffset());
+        }
+        initializeDataChunks();
```
Arun-LinkedIn, line 1877: Would you need to call `initializeDataChunks` outside the `if (options.getBlobOptions.getRange() != null)` since we also need to initialize it for segmented GETs now. But you might still need to only initialize if range resolution passed before initializing.

snalli: problem with segmented get
fixBlobSizeIfRequired(blobInfo.getBlobProperties().getBlobSize(),
                  options.getBlobOptions.hasBlobSegmentIdx() ? firstChunk.getChunkMetadataList().get(0).getSize()
                      : totalSize);

This changes the size of the blob in the metadata cache and then the cached entry doesn't match blobinfo coming from backend. Ignoring SEGMENTED GET as discussed offline

snalli:
Fixed unittest failures
The caching policy right now is quite conservative as below. If this passes EI, then we can think of expanding the policy.

1. Populate cache on each RANGE request if blobSize >= 1 TB (configurable)
2. Lookup cache only for RANGE requests

snalli:
just running some tests with the cache enabled. WIP.

justinlin-linkedin (approved):
LGTM

Arun-LinkedIn (approved):
Looks good to me too

PR #2143, opened 2022-06-28
[Named Blob Migration] Change MySQL related code to be ready for named blob data transition test and verification

mikeyang923:
This PR change the named blob mysql CRUD logic with newly added fields (version, blob_state). This together with the DB migration will make mysql layer ready for Named Blob Update API. And next step is to utilize the version value in the frontend/router layer and update ambry-client.


ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -86,44 +88,37 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
 
   /**
    * Select records up to a specific limit where the blob name starts with a string prefix. The fourth parameter can
    * be used for pagination.
    */
   private static final String LIST_QUERY = String.format("SELECT %1$s, %2$s, %3$s, %4$s FROM %5$s "
-          + "WHERE (%6$s, %7$s) = (?, ?) AND %1$s LIKE ? AND %1$s >= ? ORDER BY %1$s ASC LIMIT ?", BLOB_NAME, BLOB_ID,
-      EXPIRES_TS, DELETED_TS, NAMED_BLOBS, ACCOUNT_ID, CONTAINER_ID);
+          + "WHERE (%6$s, %7$s) = (?, ?) AND %8$s AND %1$s LIKE ? AND %1$s >= ? ORDER BY %1$s ASC LIMIT ?", BLOB_NAME,
+      BLOB_ID, EXPIRES_TS, DELETED_TS, NAMED_BLOBS, ACCOUNT_ID, CONTAINER_ID, STATE_MATCH);
```
snalli, line 100: This is getting harder to read and will get harder as we add more args. Is there a java equivalent of python format using named args ?  MessageFormat perhaps ?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -86,44 +88,37 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
-          CONTAINER_ID, BLOB_NAME, BLOB_ID, EXPIRES_TS);
-
-  /**
-   * If a record already exists for a named blob, attempt an update if the record in the DB represents an expired or
-   * deleted blob.
-   */
-  private static final String UPDATE_IF_DELETED_OR_EXPIRED_QUERY =
-      String.format("UPDATE %s SET %s = ?, %s = ?, %s = null WHERE %s AND %s", NAMED_BLOBS, BLOB_ID, EXPIRES_TS,
-          DELETED_TS, PK_MATCH, IS_DELETED_OR_EXPIRED);
+      String.format("INSERT INTO %s (%s, %s, %4$s, %5$s, %6$s, %7$s, %8$s) VALUES (?, ?, ?, ?, ?, ?, ?)", NAMED_BLOBS,
```
snalli, line 106: please use arg index like %1$s, %2$s

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -86,44 +88,37 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
    */
   private static final String INSERT_QUERY =
-      String.format("INSERT INTO %s (%s, %s, %4$s, %5$s, %6$s) VALUES (?, ?, ?, ?, ?)", NAMED_BLOBS, ACCOUNT_ID,
-          CONTAINER_ID, BLOB_NAME, BLOB_ID, EXPIRES_TS);
-
-  /**
-   * If a record already exists for a named blob, attempt an update if the record in the DB represents an expired or
-   * deleted blob.
-   */
-  private static final String UPDATE_IF_DELETED_OR_EXPIRED_QUERY =
```
snalli, line 110: why did u remove UPDATE_IF_DELETED_OR_EXPIRED_QUERY

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
```diff
@@ -5,7 +5,9 @@ CREATE TABLE IF NOT EXISTS named_blobs (
     blob_id varbinary(50) NOT NULL,
     deleted_ts datetime(6) DEFAULT NULL,
     expires_ts datetime(6) DEFAULT NULL,
-    PRIMARY KEY (account_id, container_id, blob_name)
+    version long(20) NOT NULL,
+    blob_state varchar(20) NOT NULL,
+    PRIMARY KEY (account_id, container_id, blob_name, version)
```
justinlin-linkedin, line 10: what are the steps to adding a new column and change the primary key of a table?
It seems like if we don't prepare for the schema change, we would add version as a new column with NULL value. 

Should we walk out a plan to do schema migration?
Maybe we should do something like this
1. Add version but allow NULL
2. Starting writing version to table
3. Backfill version 
4. Disallow NULL on version
5. Change primary key  to include version.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -417,6 +395,17 @@ Map<String, DataSource> getDataSources() {
         .collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getDataSource()));
   }
 
+  /**
+   * Build the version for Named Blob row based on timestamp and uuid postfix.
+   * @return a long number whose rightmost 5 digits are uuid postfix, and the remaining digits are current timestamp
+   */
+  private static long buildVersion() {
+    long currentTime = System.currentTimeMillis();
+    UUID uuid = UUID.randomUUID();
```
snalli, line 404: how about using datacenter id - lva, ltx, lor instead of randomUUID so that we are more deterministic in execution ?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
cgtz: MessageFormat works to let you use curly brace replacements, but perhaps at this point it is more readable to just accept some duplication but write out the queries in full without any replacements.

ambry-named-mysql/src/main/java/com/github/ambry/named/NamedBlobState.java
```diff
@@ -0,0 +1,6 @@
+package com.github.ambry.named;
+
+public enum NamedBlobState {
+  WIP,
```
snalli, line 4: nit: IN_PROGRESS, COMPLETE

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: Yeah, and you could also say "select * from <tablename> where <colname1> ...". Basically select all columns and then filter out what you don't need the get() method below.

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
snalli: Would this backfill need updating all rows where version is NULL ? If so, we could allow NULL and treat it as T0.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
cgtz: If `select *` is used, please be careful about getting by column index in the `ResultSet`.  golden gate schemas have a few extra columns to keep in mind.

ambry-api/src/main/java/com/github/ambry/named/NamedBlobRecord.java
```diff
@@ -72,6 +73,20 @@ public String getBlobId() {
     return blobId;
   }
 
+  /**
+   * @set the version for the named blob map.
+   */
+  public void setVersion(long versionNew) {
```
cgtz, line 79: Could this just be added to the constructor?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -241,34 +239,11 @@ public CompletableFuture<PutResult> put(NamedBlobRecord record) {
             } else {
               statement.setTimestamp(5, null);
             }
+            statement.setLong(6, buildVersion());
+            statement.setString(7, NamedBlobState.READY.name());
```
cgtz, line 243: Are there still plans to support a Put without overwrite option? If so, what would the query for that look like given timestamp based version ids?

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
mikeyang923: Yeah, good point, I raised a ticket to MySQL DBA (https://jira01.corp.linkedin.com:8443/browse/MI-10173), will work with the DBA for the detailed steps of schema migration. 
Since we are adding version to primary key, the version field can't be null

mikeyang923: Yeah, good point, I raised a ticket to MySQL DBA (https://jira01.corp.linkedin.com:8443/browse/MI-10173), will work with the DBA for the detailed steps of schema migration. 
Since we are adding version to primary key, the version field can't be null

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Updated to only use the replacement for the 'from/where' clause (totally 5 replacements).

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Because it's not needed anymore, previously we assume there is only one row per blob_id, so this 'UPDATE_IF_DELETED_OR_EXPIRED_QUERY' was applied when the row for a blob_id 'AlreadyExist', but now we are using blob_id+version, that logic is not needed anymore.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: I didn't use datacenter id to avoid concurrency within a datacenter (eg multiple concurrent requests for the same named blob in datacenter lva).

ambry-named-mysql/src/main/java/com/github/ambry/named/NamedBlobState.java
mikeyang923: I think WIP is a well known abbr. for 'Work In Progress', so readability-wise it's ok. And it can save some storage/bits so :)

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Good point, think it's doable, I will think about and add that.

snalli:
Does this any reviews or does it need any updates from author ?

mikeyang923:
> 

It's ready for review now. thanks!

mikeyang923:
> Does this any reviews or does it need any updates from author ?

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
```diff
@@ -20,3 +20,29 @@ CREATE INDEX named_blobs_et ON named_blobs(expires_ts);
...
 CREATE INDEX named_blobs_id ON named_blobs(blob_id);
+
+/* New Table */
+CREATE TABLE IF NOT EXISTS named_blobs_v2 (
+    account_id int NOT NULL,
+    container_id int NOT NULL,
+    blob_name varchar(191) NOT NULL,
+    blob_id varbinary(50) NOT NULL,
+    deleted_ts datetime(6) DEFAULT NULL,
+    expires_ts datetime(6) DEFAULT NULL,
```
SophieGuo410, line 31: As we discussed, are we suppose to remove the expires_ts and use deleted_ts only?

mikeyang923: Yes, just did that in latest commit [6f3c303](https://github.com/linkedin/ambry/pull/2143/commits/6f3c3031ab43eda4bd25a40f9968884870eced8f)

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -118,22 +139,30 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
 
   private final AccountService accountService;
   private final String localDatacenter;
   private final List<String> remoteDatacenters;
   private final RetryExecutor retryExecutor;
   private final Map<String, TransactionExecutor> transactionExecutors;
   private final MySqlNamedBlobDbConfig config;
+  private final Counter namedDataInconsistentGetCount;
+  private final Counter namedDataInconsistentListCount;
+  private final Counter namedDataInconsistentDeleteCount;
```
justinlin-linkedin, line 162: I suggest making an inner class specifically for metrics. Something like
```
private static class Metrics {
}
```
But make sure you have MySqlNamedBlobDb to initialize the name of the metrics.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -148,6 +177,12 @@ class MySqlNamedBlobDb implements NamedBlobDb {
                 dataSourceFactory.getDataSource(dbEndpoint),
                 localDatacenter.equals(dbEndpoint.getDatacenter()) ? config.localPoolSize : config.remotePoolSize)));
     this.remoteDatacenters = MySqlUtils.getRemoteDcFromDbInfo(config.dbInfo, localDatacenter);
+    this.namedDataInconsistentGetCount = metricRegistry.counter(
+        MetricRegistry.name(MySqlNamedBlobDb.class, "namedDataInconsistentGetCount"));
+    this.namedDataInconsistentListCount = metricRegistry.counter(
+        MetricRegistry.name(MySqlNamedBlobDb.class, "namedDataInconsistentListCount"));
+    this.namedDataInconsistentDeleteCount = metricRegistry.counter(
+        MetricRegistry.name(MySqlNamedBlobDb.class, "namedDataInconsistentDeleteCount"));
```
justinlin-linkedin, line 185: we should capitalize the names

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -156,157 +191,70 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
...
-          } else {
-            return new NamedBlobRecord(accountName, containerName, blobName, blobId, timestampToMs(expirationTime));
-          }
-        }
+      NamedBlobRecord record, recordV2 = null;
+      record = run_get(accountName, containerName, blobName, option, accountId, containerId, connection);
+      try {
+        recordV2 = run_get_v2(accountName, containerName, blobName, option, accountId, containerId, connection);
+      } catch (Exception e) {
+        logger.error("NAMED GET: Data Error: old table succeed while new table failed: ", e);
```
justinlin-linkedin, line 199: this doesn't seems right. There will be a period of time when v1 table has more data than v2 table, some blobname would return a valid record in v1 but not_found in v2. and this shouldn't be an issue. 

we should probably do something like this to cover this case
```
record = run_get();
try {
    recordV2 = run_get_v2();
    if (!record.equalsWithoutVersion(recordV2)) {
        // log error
        // metrics
    }
} catch (Exception e) {
    if (isNotFoundException(e)) {
        // record metric here
    }
    // log an error 
    // record metric here
}
```

ambry-api/src/main/java/com/github/ambry/named/NamedBlobRecord.java
```diff
@@ -88,8 +110,17 @@ public boolean equals(Object o) {
       return false;
     }
     NamedBlobRecord record = (NamedBlobRecord) o;
-    return expirationTimeMs == record.expirationTimeMs && Objects.equals(accountName, record.accountName)
-        && Objects.equals(containerName, record.containerName) && Objects.equals(blobName, record.blobName)
-        && Objects.equals(blobId, record.blobId);
+    return Objects.equals(accountName, record.accountName) && Objects.equals(containerName, record.containerName)
+        && Objects.equals(blobName, record.blobName) && Objects.equals(blobId, record.blobId);
+  }
```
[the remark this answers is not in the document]
mikeyang923: 1. The version can be added back after the data transition period (otherwise the old table and new table's NamedBlobRecords won't match).
2. @SophieGuo410 and I discussed recently about merging 'expires_ts' into 'deleted_ts' and remove the 'expires_ts' column, due to the MySQL auto-purge only work with 'deleted_ts'. So now the NamedBlobRecord's `expirationTimeMs` is used to hold the 'deleted_ts' value which will be min of (expires_ts, deleted_ts). With this background info, we need to remove this '`expirationTimeMs`' in equality check, otherwise below line will fail:
https://github.com/linkedin/ambry/blob/b481e8c89e5355a73a59c804c2101e10d65ce279/ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java#L138
Because 'record''s expirationTimeMs is default value -1, while 'recordFromStore''s expirationTimeMs is holding the deleted time which is not -1. Moreover, I feel the 'expirationTimeMs' is not the keys to check whether two records are the same, as long as the other properties are equal, the records are equal.

ambry-api/src/main/java/com/github/ambry/named/DeleteResult.java
```diff
@@ -44,4 +47,20 @@ public String getBlobId() {
...
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    DeleteResult record = (DeleteResult) o;
+    return Objects.equals(getBlobId(), record.getBlobId()) && Objects.equals(isAlreadyDeleted(), record.isAlreadyDeleted());
```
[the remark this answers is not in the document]
mikeyang923: Good point, done in c7b422b

ambry-api/src/main/java/com/github/ambry/named/DeleteResult.java
```diff
@@ -44,4 +47,20 @@ public String getBlobId() {
...
+  }
+
+  public String toString() {
```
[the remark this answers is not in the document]
mikeyang923: Done in  c7b422b

ambry-api/src/main/java/com/github/ambry/named/NamedBlobRecord.java
```diff
@@ -88,8 +110,17 @@ public boolean equals(Object o) {
...
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(blobId);
+  }
+
+  public String toString() {
```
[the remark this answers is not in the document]
mikeyang923: Done in 
c7b422b

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
```diff
@@ -20,3 +20,25 @@ CREATE INDEX named_blobs_et ON named_blobs(expires_ts);
...
 CREATE INDEX named_blobs_id ON named_blobs(blob_id);
+
+/* New Table */
+CREATE TABLE IF NOT EXISTS named_blobs_v2 (
+    account_id int NOT NULL,
+    container_id int NOT NULL,
+    blob_name varchar(191) NOT NULL,
+    blob_id varbinary(50) NOT NULL,
+    deleted_ts datetime(6) DEFAULT NULL,
+    version bigint NOT NULL,
```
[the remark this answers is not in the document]
mikeyang923: Done in 
c7b422b

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
```diff
@@ -20,3 +20,25 @@ CREATE INDEX named_blobs_et ON named_blobs(expires_ts);
...
+    blob_state varchar(20) NOT NULL,
```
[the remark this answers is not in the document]
mikeyang923: Ok, done in c7b422b

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done in 
c7b422b

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: I did it this way to be consistent with the other metrics definitions in ambry, where the names are not capitalized.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Right, but there won't be too many such cases, and this is just for logging/metrics purpose, I don't think we are aiming for perfect metrics (eg. 0), we are good if the metrics is small enough?

justinlin-linkedin: We only have to run sync old to new table once after deployment. My idea is
1. Deploy the code that writes to both table 
2. Copy data from old table to new 

In this way, we will guarantee all the old data would be synced to new. 

However, we don't really know when we are going to do the step 2 so after step 1 there will be lots of missing data in new table. 

That's my plan here.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
justinlin-linkedin: but most of the metrics are capitalized. If you see non-capitalized metrics, they are not norm.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: I see, just updated in commit [359fb6f](https://github.com/linkedin/ambry/pull/2143/commits/359fb6f29d7756fe4bb290255c553791128f0142). Can you take another look?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -95,21 +107,30 @@ class MySqlNamedBlobDb implements NamedBlobDb {
   private static final String LIST_QUERY = String.format("SELECT %1$s, %2$s, %3$s, %4$s FROM %5$s "
           + "WHERE (%6$s, %7$s) = (?, ?) AND %1$s LIKE ? AND %1$s >= ? ORDER BY %1$s ASC LIMIT ?", BLOB_NAME, BLOB_ID,
       EXPIRES_TS, DELETED_TS, NAMED_BLOBS, ACCOUNT_ID, CONTAINER_ID);
+  private static final String LIST_QUERY_V2 = String.format("SELECT t1.blob_name, t1.blob_id, t1.version, t1.deleted_ts "
+          + "FROM named_blobs_v2 t1 INNER JOIN (SELECT account_id, container_id, blob_name, max(version) as version FROM named_blobs_v2 "
+          + "WHERE (account_id, container_id) = (?, ?) AND %1$s GROUP BY account_id, container_id, blob_name) t2 ON "
+          + "(t1.account_id,t1.container_id,t1.blob_name,t1.version) = (t2.account_id,t2.container_id,t2.blob_name,t2.version) "
+          + "WHERE t1.blob_name LIKE ? AND t1.blob_name >= ? ORDER BY t1.blob_name ASC LIMIT ?",
+      STATE_MATCH);
```
justinlin-linkedin, line 115: is this really necessary? I am not an expert on sql statement, but it seems like we can do something like
```
SELECT blob_name, blob_id, max(version), deleted_ts FROM named_blobs_v2 WHERE (account_id, container_id) = (?, ?) AND blob_state = Ready AND  blob_name LIKE ? AND blob_name >= ?  GROUP BY account_id, container_id, blob_name ORDER BY blob_name ASC LIMIT ?
```
Does this look right to you ?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -417,6 +377,293 @@ Map<String, DataSource> getDataSources() {
...
+  private PutResult run_put(NamedBlobRecord record, short accountId, short containerId, Connection connection)
+      throws Exception {
+    // 1. Pulling existing row content so that we can get the min(expires_ts, deleted_ts) of existing and new records
+    NamedBlobRecord existingRow;
+    try {
+      existingRow = run_get(record.getAccountName(), record.getContainerName(), record.getBlobName(),
+          GetOption.Include_All, accountId, containerId, connection);
+    } catch (Exception e) {
+      existingRow = null;
+    }
```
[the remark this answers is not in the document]
mikeyang923: As per the comment above that code:
`// 1. Pulling existing row content so that we can get the min(expires_ts, deleted_ts) of existing and new records`
Suppose there is existing row with:
x, y, z, ... deleted_ts_1, ...

And new data to be inserted has min (expires_ts/deleted_ts) as 'deleted_ts_2'.
Then we need to insert the data as min (deleted_ts_1, deleted_ts_2), so we need that get query.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -156,157 +203,70 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
...
-            }
-            statement.executeUpdate();
-          } catch (SQLIntegrityConstraintViolationException e) {
-            rowAlreadyExists = true;
+          PutResult putResult;
+          putResult = run_put(record, accountId, containerId, connection);
+          try {
+            run_put_v2(record, accountId, containerId, connection);
+          } catch (Exception e) {
+            logger.error("NAMED PUT: Data Error: old table succeed while new table failed: ", e);
```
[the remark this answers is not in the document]
mikeyang923: Ok, done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -156,157 +203,70 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
...
-          // use the current time
-          statement.setTimestamp(1, currentTime);
-          statement.setInt(2, accountId);
-          statement.setInt(3, containerId);
-          statement.setString(4, blobName);
-          statement.executeUpdate();
-        }
+      if (!deleteResult.equals(deleteResultV2)) {
+        logger.warn("NAMED DELETE: Data Inconsistence: old record='{}', new record='{}'", deleteResult, deleteResultV2);
+        metricsRecoder.namedDataInconsistentDeleteCount.inc();
```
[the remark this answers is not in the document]
mikeyang923: Ok, done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: No, your query has problem because:
"All column names in SELECT list must appear in GROUP BY clause unless name is used only in an aggregate function" https://stackoverflow.com/questions/5986127/do-all-columns-in-a-select-list-have-to-appear-in-a-group-by-clause

I actually spent quite some time to come up with above query, it looks long, but the idea is simple: 

```
1. Get the max version per each (account_id, container_id, blob_name) as t2.
2. Join the original table with t2 on the primary keys to filter out historical data with older versions.
3. And then we can apply the blob_name like ?, and >, order by and limit per page size.
```

ambry-named-mysql/src/main/java/com/github/ambry/named/NamedBlobState.java
```diff
@@ -0,0 +1,27 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.github.ambry.named;
+
+public enum NamedBlobState {
+  WIP(0),
+  READY(1);
```
justinlin-linkedin, line 19: I think we can just rely on ordinal value, we don't have to assign another value for it.

ambry-named-mysql/src/main/java/com/github/ambry/named/NamedBlobState.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.github.ambry.named;
+
+public enum NamedBlobState {
+  WIP(0),
```
justinlin-linkedin, line 18: We should use InProgress now we are going to use integral value.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -156,157 +224,85 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
...
+      try {
+        recordV2 = run_get_v2(accountName, containerName, blobName, option, accountId, containerId, connection);
+        if (!record.equals(recordV2)) {
+          logger.warn("NAMED GET: Data Inconsistence: old record='{}', new record='{}'", record, recordV2);
+          metricsRecoder.namedDataInconsistentGetCount.inc();
+        }
+      } catch (Exception e) {
+        if (e instanceof RestServiceException &&
+            ((RestServiceException) e).getErrorCode() == RestServiceErrorCode.NotFound) {
+          logger.warn("NAMED GET: Data Not Found: old table succeed while new table got NotFound error: ", e);
```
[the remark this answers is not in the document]
mikeyang923: Done in [7403e50](https://github.com/linkedin/ambry/pull/2143/commits/7403e5066fef239a52857ab4f2cf4d8fb1b137d3)

ambry-named-mysql/src/main/java/com/github/ambry/named/NamedBlobState.java
mikeyang923: Done in [7403e50](https://github.com/linkedin/ambry/pull/2143/commits/7403e5066fef239a52857ab4f2cf4d8fb1b137d3)

ambry-named-mysql/src/main/java/com/github/ambry/named/NamedBlobState.java
mikeyang923: Done in [7403e50](https://github.com/linkedin/ambry/pull/2143/commits/7403e5066fef239a52857ab4f2cf4d8fb1b137d3)
Can you take another look?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -417,6 +413,293 @@ Map<String, DataSource> getDataSources() {
...
+      if (record.getExpirationTimeMs() != Utils.Infinite_Time) {
+        statement.setTimestamp(5, new Timestamp(record.getExpirationTimeMs()));
+      } else {
+        statement.setTimestamp(5, null);
+      }
+      statement.setLong(6, buildVersion());
+      statement.setInt(7, NamedBlobState.READY.ordinal());
+      statement.executeUpdate();
+    } catch (SQLIntegrityConstraintViolationException e) {
+      throw e;
```
justinlin-linkedin, line 616: let's call build exception here.

ambry-api/src/main/java/com/github/ambry/named/NamedBlobRecord.java
```diff
@@ -92,4 +114,15 @@ public boolean equals(Object o) {
...
   }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(blobId);
+  }
+
+  @Override
+  public String toString() {
+    return "NamedBlobRecord[accountName=" + accountName + ",containerName=" + containerName + ",blobName=" + blobName +
```
justinlin-linkedin, line 125: let's print out the version as well.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -95,6 +108,12 @@ class MySqlNamedBlobDb implements NamedBlobDb {
   private static final String LIST_QUERY = String.format("SELECT %1$s, %2$s, %3$s, %4$s FROM %5$s "
           + "WHERE (%6$s, %7$s) = (?, ?) AND %1$s LIKE ? AND %1$s >= ? ORDER BY %1$s ASC LIMIT ?", BLOB_NAME, BLOB_ID,
       EXPIRES_TS, DELETED_TS, NAMED_BLOBS, ACCOUNT_ID, CONTAINER_ID);
+  private static final String LIST_QUERY_V2 = String.format("SELECT t1.blob_name, t1.blob_id, t1.version, t1.expires_ts, t1.deleted_ts "
+          + "FROM named_blobs_v2 t1 INNER JOIN (SELECT account_id, container_id, blob_name, max(version) as version FROM named_blobs_v2 "
+          + "WHERE (account_id, container_id) = (?, ?) AND %1$s GROUP BY account_id, container_id, blob_name) t2 ON "
+          + "(t1.account_id,t1.container_id,t1.blob_name,t1.version) = (t2.account_id,t2.container_id,t2.blob_name,t2.version) "
+          + "WHERE t1.blob_name LIKE ? AND t1.blob_name >= ? ORDER BY t1.blob_name ASC LIMIT ?",
+      STATE_MATCH);
```
justinlin-linkedin, line 116: nit: can we format this sql statement
```
// @formatter:off
String.format(""
          + "SELECT t1.blob_name, t1.blob_id, t1.version, t1.expires_ts, t1.deleted_ts "
          + "FROM named_blobs_v2 t1 " 
          + "INNER JOIN "
          + "SELECT account_id, container_id, blob_name, max(version) as version "
          + "FROM named_blobs_v2 "
          + "WHERE (account_id, container_id) = (?, ?) AND %1$s "
          + "        GROUP BY account_id, container_id, blob_name) t2 "
          + "ON (t1.account_id = t2.account_id AND t1.container_id = t2.contaienr_id AND t1.blob_name = t1.blob_name AND t1.version = t2.version) "
          + "WHERE t1.blob_name LIKE ? AND t1.blob_name >= ? ORDER BY t1.blob_name ASC LIMIT ?");
// @formatter:on
```

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -148,6 +176,47 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
+    public final Counter namedDataErrorDeleteCount;
+    public final Counter namedDataInconsistentDeleteCount;
+
+    public final Counter namedDataErrorPutCount;
+
+    /**
+     * Constructor to create the Metrics.
+     * @param metricRegistry The {@link MetricRegistry}.
+     */
+    public Metrics(MetricRegistry metricRegistry) {
```
justinlin-linkedin, line 199: now we are on metrics here, can we add latency metrics for each sql methods?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: The issue is that the 'buildException' method need 'accountName', 'containerName', 'blobName':
```
  private static RestServiceException buildException(String message, RestServiceErrorCode errorCode, String accountName,
      String containerName, String blobName) {
```
While we only have 'accountId', 'containerId' at above line. To call build exception, we need refactoring of the methods signatures.

Plus that any exception raised here will be caught in the 'put' method for logging/metrics track purpose only, it won't be exposed to client, so I don't see benefits of call build exception here.
```
          try {
            run_put_v2(record, accountId, containerId, connection);
          } catch (Exception e) {
            logger.error("NAMED PUT: Data Error: old table succeed while new table failed: ", e);
            metricsRecoder.namedDataErrorPutCount.inc();
          }
```

ambry-api/src/main/java/com/github/ambry/named/NamedBlobRecord.java
mikeyang923: Good point, done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Good, done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Sure, added. can you take another look?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -156,157 +246,92 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
...
-          // use the current time
-          statement.setTimestamp(1, currentTime);
-          statement.setInt(2, accountId);
-          statement.setInt(3, containerId);
-          statement.setString(4, blobName);
-          statement.executeUpdate();
+      } catch (Exception e) {
+        if (e instanceof RestServiceException &&
+            ((RestServiceException) e).getErrorCode() == RestServiceErrorCode.NotFound) {
+          logger.warn("NAMED DELETE: Data Not Found: old table succeed while new table got NotFound error: ", e);
```
justinlin-linkedin, line 326: let's not add a warn log message here, since just like get, this might happen a lot before backfill.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -156,157 +246,92 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
...
-          } else {
-            return new NamedBlobRecord(accountName, containerName, blobName, blobId, timestampToMs(expirationTime));
-          }
+      long startTime = System.currentTimeMillis();
+      NamedBlobRecord record, recordV2;
+      record = run_get(accountName, containerName, blobName, option, accountId, containerId, connection);
+      try {
+        recordV2 = run_get_v2(accountName, containerName, blobName, option, accountId, containerId, connection);
+        if (!record.equals(recordV2)) {
+          logger.warn("NAMED GET: Data Inconsistence: old record='{}', new record='{}'", record, recordV2);
```
justinlin-linkedin, line 255: for list, I suggest we don't add a warn log. List is not like a point lookup query. For get, warn log message makes sense. If we see two records are different in old and new table, this is a warning level error. But in list, before we backfill, we are going to see lots of differences.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
justinlin-linkedin: if that's the case let's not catch this exception, there is no point to just catch an exception and then throw it right away without doing anything.

justinlin-linkedin (approved):
left a couple nitpicking comments, most of them looks good.
btw, could you please invite an sql expert to review your sql statement?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done in latest commit.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: I just removed the warn log in LIST part.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done in latest commit.

mikeyang923:
> 

Sounds good, @bei-xu

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -417,6 +440,291 @@ Map<String, DataSource> getDataSources() {
...
+      statement.setInt(1, accountId);
+      statement.setInt(2, containerId);
+      statement.setString(3, record.getBlobName());
+      statement.setBytes(4, Base64.decodeBase64(record.getBlobId()));
+      if (record.getExpirationTimeMs() != Utils.Infinite_Time) {
+        statement.setTimestamp(5, new Timestamp(record.getExpirationTimeMs()));
+      } else {
+        statement.setTimestamp(5, null);
+      }
+      statement.setLong(6, buildVersion());
```
SophieGuo410, line 639: looks like run_put_v2 will take all the put request, but for run_put, it will return conflict when blob still alive.
So when you do get, would it be different? So you will never know if db and db_v2 is consistent or not?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -95,6 +109,18 @@ class MySqlNamedBlobDb implements NamedBlobDb {
   private static final String LIST_QUERY = String.format("SELECT %1$s, %2$s, %3$s, %4$s FROM %5$s "
           + "WHERE (%6$s, %7$s) = (?, ?) AND %1$s LIKE ? AND %1$s >= ? ORDER BY %1$s ASC LIMIT ?", BLOB_NAME, BLOB_ID,
       EXPIRES_TS, DELETED_TS, NAMED_BLOBS, ACCOUNT_ID, CONTAINER_ID);
+  // @formatter:off
+  private static final String LIST_QUERY_V2 = String.format(""
```
SophieGuo410, line 113: Can you add some comments here to explain what's your query is doing?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Right, that's for supporting versioned named-blob, so run_put_v2, will create new version instead of raise conflict. 
But look at the put method, I don't catch any exception from 'run_put', so if it failed in 'run_put' due to conflicts, then it won't call run_put_v2. Therefore, it should be ok regarding the db vs db_v2 comparison.
```
          putResult = run_put(record, accountId, containerId, connection);
          try {
            run_put_v2(record, accountId, containerId, connection);
          } catch (Exception e) {
            logger.error("NAMED PUT: Data Error: old table succeed while new table failed: ", e);
            metricsRecoder.namedDataErrorPutCount.inc();
          }
```

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Sure, Bei Xu is reviewing this query, if she is good with it, I can add a comment then.

justinlin-linkedin (commented):
Maybe we should introduce a configuration for the new table. So even if we merge the code it will not take affect.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
SophieGuo410: Yeah, I understand it will be good for PUT. But for GET, let's say when people trying to get a named blob, but the record is different since the old DB won't have the latest value. Would that be a problem? I thought finally you will compare all the record and make sure data in DB is consistent with data in DB_V2?

mikeyang923: As we chatted, the run_put_v2 won't run if the run_put failed. So we are good with this.

mikeyang923:
> Maybe we should introduce a configuration for the new table. So even if we merge the code it will not take affect.

Good idea, done in latest commit.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Bei reviewed this query and give it a go. Just added blob_name per Bei's suggestion, and added the java doc for this query.

PR #2145, opened 2022-06-29
Add platform specific build for netty

justinlin-linkedin:
A quick workaround to re-enable openssl.

Real released war should be built in a linux box so it will eventually include linux-native library.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -178,6 +178,27 @@ subprojects {
     }
 }
 
+ext {
+    osArch = System.getProperty("os.arch")
+    osName = System.getProperty("os.name").toLowerCase()
+    osNameShort = ""
+    if (osName.contains("mac"))
+        osNameShort = "osx"
+    else if (osName.contains("Linux"))
+        osNameShort = "linux"
+    else
+        osNameShort = "windows"
+    osClass = "${osNameShort}_$osArch"
+    osArchNetty = ""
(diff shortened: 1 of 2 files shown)


snalli (approved):

build.gradle
```diff
@@ -396,7 +404,11 @@ project(':ambry-rest') {
         compile "io.dropwizard.metrics:metrics-jmx:$metricsVersion"
         compile "net.sf.jopt-simple:jopt-simple:$joptSimpleVersion"
         compile "io.netty:netty-all:$nettyVersion"
-        compile "io.netty:netty-tcnative-boringssl-static:$nettyTcnativeVersion"
+        runtime "io.netty:netty-tcnative-boringssl-static:$nettyTcnativeVersion:linux-x86_64"
```
cgtz, line 407: what's the difference between runtime and runtimeOnly?

justinlin-linkedin: Not very familiar with Gradle, but it seems like runtime is a deprecated configuration, just like compile. The new gradle would use implementation to replace compile and runtimeOnly to replace runtime.

justinlin-linkedin: https://docs.gradle.org/6.8.1/userguide/java_plugin.html#tab:configurations

Arun-LinkedIn (approved):

cgtz (approved):
lgtm

PR #2150, opened 2022-07-12
Handle regular http requests in http2 server in ambry-server

justinlin-linkedin:
We are adding regular http handlers in ambry-server to handle regular http request.
In order to do that, we have to change a couple interfaces.
1. NetworkRequest
2. RequestAPI

In NetworkRequest, we return an FullHttpRequest from netty when this is a regular http request.
In RequestAPI, we have a method to handle this http request.

NetworkRequest goes through a pipeline from AmbryNetworkRequestHandler to NettyServerRequestResponseChannel, and to RequestHandler and RequestHandlerPool and finally ends up in AmbryServerRequests. 

Once the regular http request reaches the AmbryServerRequests, we need to multiplexing it with different http handlers. Http Handler is created to handle different http requests. To handle new http requests, we have to add new http handlers that handle this request.

Some tests will be added later.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/network/NetworkRequest.java
+++ b/ambry-api/src/main/java/com/github/ambry/network/NetworkRequest.java
@@ -14,6 +14,7 @@
 package com.github.ambry.network;
 
 import io.netty.buffer.ByteBuf;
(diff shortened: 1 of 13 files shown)


justinlin-linkedin:
Not doing this anymore, will close this one.

PR #2151, opened 2022-07-12
Issue cosmos queries serially to avoid 429s

Arun-LinkedIn:
After refactoring AzureCloudDestination to issue requests to cloud asynchronously, multiple cosmos queries were being issued in parallel. This could be the reason for increase in 429s from Cosmos on the canary node. To avoid that, this change issues cosmos chains the query operations and issues them serially.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
@@ -20,7 +20,6 @@ import com.azure.storage.blob.BlobServiceAsyncClient;
 import com.azure.storage.blob.BlobServiceClient;
 import com.azure.storage.blob.batch.BlobBatchAsyncClient;
 import com.azure.storage.blob.models.BlobErrorCode;
-import com.azure.storage.blob.models.BlobProperties;
 import com.azure.storage.blob.models.BlobRequestConditions;
 import com.azure.storage.blob.models.BlobStorageException;
 import com.codahale.metrics.Timer;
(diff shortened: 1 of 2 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureBlobDataAccessor.java
```diff
@@ -418,43 +417,41 @@ public CompletableFuture<UpdateResponse> updateBlobMetadataAsync(BlobId blobId,
...
-              deleteBatchOperationFuture.complete(null);
+              deletedBlobs.add(blobMetadata);
             }
           });
-      deleteBatchOperationFutureList.add(deleteBatchOperationFuture);
+      if (resultFuture == null) {
+        resultFuture = currentBatchDeleteFuture;
+      } else {
+        // Issue batch deletes serially one after another since issuing all in parallel can be expensive resulting in 429s.
+        resultFuture = resultFuture.thenCompose(unused -> currentBatchDeleteFuture);
```
justinlin-linkedin, line 450: This seems to defeat the purpose of passing a function that would return a CompletionStage.

The idea is to not send the delete request unless the first delete batch request has returned, right? But creating a CompletableFuture on each delete batch in this for loop would trigger delete request be sent out right away. We should do something like this

```
CompletableFuture<Void> resultFuture = CompletableFuture.completedFuture(null);
for (List<CloudBlobMetadata> batchOfBlobs : partitionedLists) {
    resultFuture = resultFuture.thenCompose(unused -> FutureUtils.orTimeout(....));
}
```

In this way, each delete batch would be encapsulated in the Function that is passed to thenCompose, so only when the previous delete batch response comes back , we would start composing the next completable future.

justinlin-linkedin (changes requested):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestination.java
```diff
@@ -390,57 +390,44 @@ public Map<String, CloudBlobMetadata> getBlobMetadata(List<BlobId> blobIds) thro
...
-          resultFuture.completeExceptionally(
+          throw new CompletionException(
               toCloudStorageException("Failed to query metadata for blob" + blobIds.get(0), ex));
         } else {
           if (cloudBlobMetadata == null) {
-            resultFuture.complete(Collections.emptyMap());
+            return Collections.emptyMap();
           } else {
-            resultFuture.complete(Collections.singletonMap(cloudBlobMetadata.getId(), cloudBlobMetadata));
+            return Collections.singletonMap(cloudBlobMetadata.getId(), cloudBlobMetadata);
```
justinlin-linkedin, line 411: though I think this can be as simple as 
```
return azureBlobDataAccessor.getBlobMetadataAsync(blobIds.get(0)).thenApply(md -> md == null? Collections.emptyMap()? Collections.singletonMap(md.getId(), md));
```

justinlin-linkedin (approved):
LGTM

ankagrawal (approved):

PR #2155, opened 2022-07-19
Add a configuration to storage quota to reject request without storage quota

justinlin-linkedin:
Adding a configuration to storage enforcer so that we can starting rejecting upload requests that has no quota in targeted account or container.


SophieGuo410 (approved):

ankagrawal (approved):
Approving this PR. Having this functionality doesn't hurt.

However, we should be careful in turning this on. Ideally there should be no way for a user to create an account without quotas. As long as this is true, this feature shouldn't be turned on, because once a user has quotas, then it is Ambry's (or its downstream's) responsibility to fetch those quota values. Instead of penalizing user requests for quota failures (and getting into more support and availability issues), we should create alerts for quota missing scenarios, and act on those alerts promptly.

PR #2156, opened 2022-07-19
[Frontend] Metadata cache unit tests

snalli:
Add unittests to evaluate metadata cache

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/CloudRouterFactory.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CloudRouterFactory.java
@@ -23,6 +23,7 @@ import com.github.ambry.clustermap.CloudDataNode;
 import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.clustermap.ReplicaType;
+import com.github.ambry.commons.AmbryCache;
 import com.github.ambry.commons.BlobIdFactory;
 import com.github.ambry.commons.SSLFactory;
 import com.github.ambry.commons.ServerMetrics;
@@ -147,7 +148,7 @@ public class CloudRouterFactory implements RouterFactory {
       CompositeNetworkClientFactory networkClientFactory = getCompositeNetworkClientFactory(requestHandlerPool);
       NonBlockingRouter router =
           new NonBlockingRouter(routerConfig, routerMetrics, networkClientFactory, notificationSystem, clusterMap, kms,
-              cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass);
+              cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass, null);
       // Make sure requestHandlerPool is shut down properly
       router.addResourceToClose(requestHandlerPool);
       router.addResourceToClose(cloudDestination);
--- a/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterFactory.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterFactory.java
@@ -16,6 +16,7 @@ package com.github.ambry.router;
 import com.codahale.metrics.MetricRegistry;
 import com.github.ambry.account.AccountService;
 import com.github.ambry.clustermap.ClusterMap;
+import com.github.ambry.commons.AmbryCache;
 import com.github.ambry.commons.SSLFactory;
 import com.github.ambry.config.ClusterMapConfig;
 import com.github.ambry.config.Http2ClientConfig;
@@ -122,8 +123,12 @@ public class NonBlockingRouterFactory implements RouterFactory {
   @Override
   public Router getRouter() {
     try {
+      AmbryCache blobMetadataCache = new AmbryCache(routerConfig.routerBlobMetadataCacheId,
+          routerConfig.routerBlobMetadataCacheEnabled,
+          routerConfig.routerBlobMetadataCacheMaxSizeBytes,
+          routerMetrics.getMetricRegistry());
       return new NonBlockingRouter(routerConfig, routerMetrics, networkClientFactory, notificationSystem, clusterMap,
-          kms, cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass);
+          kms, cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass, blobMetadataCache);
     } catch (IOException | ReflectiveOperationException e) {
       throw new IllegalStateException("Error instantiating NonBlocking Router ", e);
     }
--- a/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
@@ -22,6 +22,7 @@ import com.github.ambry.clustermap.MockPartitionId;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.PartitionState;
 import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.commons.AmbryCache;
 import com.github.ambry.commons.BlobId;
 import com.github.ambry.commons.BlobIdFactory;
 import com.github.ambry.commons.ByteBufferAsyncWritableChannel;
@@ -97,6 +98,7 @@ import org.junit.runners.Parameterized;
 
 import static com.github.ambry.router.PutManagerTest.*;
 import static com.github.ambry.router.RouterTestHelpers.*;
+import static junit.framework.Assert.*;
 import static org.junit.Assume.*;
 
 
@@ -133,6 +135,8 @@ public class GetBlobOperationTest {
   private final RouterCallback routerCallback;
   private final String operationTrackerType;
   private final boolean testEncryption;
+  private final boolean enableMetadataCache;
+  private final AmbryCache blobMetadataCache;
   private MockKeyManagementService kms = null;
   private MockCryptoService cryptoService = null;
   private CryptoJobHandler cryptoJobHandler = null;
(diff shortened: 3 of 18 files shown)


ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -1367,6 +1367,326 @@ public void onEvent(ReplicaId replicaId, Object e) {
...
+      // delete simple blob
+      router.deleteBlob(blobId, null).get(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      assertTrue("Must delete metadata in cache", deleteCacheEntryLatch.await(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
+      assertNull("Blob metadata must be absent in metadata cache", router.getBlobMetadataCache().getObject(blobId));
+    } finally {
+      router.close();
+    }
+  }
+
+  public void routerMetadataCacheErrorOnGetCompositeBlob(ServerErrorCode serverErrorCode, RouterErrorCode routerErrorCode, boolean deleteCacheEntryOnFail) throws Exception {
```
justinlin-linkedin, line 1542: make this method private please.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -1367,6 +1367,326 @@ public void onEvent(ReplicaId replicaId, Object e) {
...
+  @Test
+  public void testRouterMetadataCacheBlobExpiredOnGetCompositeBlob() throws Exception {
+    routerMetadataCacheErrorOnGetCompositeBlob(ServerErrorCode.Blob_Expired, RouterErrorCode.BlobExpired, true);
+  }
+  @Test
+  public void testRouterMetadataCacheUnexpectedInternalErrorOnGetCompositeBlob() throws Exception {
+    routerMetadataCacheErrorOnGetCompositeBlob(ServerErrorCode.Unknown_Error, RouterErrorCode.UnexpectedInternalError, false);
+  }
+
+  public void routerMetadataCacheErrorOnDeleteCompositeBlob(ServerErrorCode serverErrorCode, RouterErrorCode routerErrorCode) throws Exception {
```
justinlin-linkedin, line 1625: no need to create this function, we can just copy all the code to the test function below.

justinlin-linkedin (changes requested):
The tests look good.
Can we also add a test on GetBlobOperationTest file. This file shares a lot of similarity with NonBlockingRouterTest file.
In GetBlobOperationTest file, we can add a test case to verify the GetRequest would be requesting BlobInfo instead of BlobAll when there is a cache hit in the MockNetworkClient.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -1367,6 +1367,326 @@ public void onEvent(ReplicaId replicaId, Object e) {
...
+      };
+
+      setRouter(properties, layout, loggingNotificationSystem);
+      assertNotNull("Metadata cache must be created", router.getBlobMetadataCache());
+
+      // put a simple blob and test it is absent in the metadata cache
+      setOperationParams((int) (8 * Math.pow(1024, 2)), TTL_SECS);
+      String blobId =
+          router.putBlob(putBlobProperties, putUserMetadata, putChannel, new PutBlobOptionsBuilder().build())
+              .get(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
```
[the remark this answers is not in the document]
snalli: thanks

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
snalli: agreed

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
snalli: yes but this would be needed if we want to test with different error combos. Right now there is just one, but I was hoping we'd add more in the future

snalli:
> The tests look good. Can we also add a test on GetBlobOperationTest file. This file shares a lot of similarity with NonBlockingRouterTest file. In GetBlobOperationTest file, we can add a test case to verify the GetRequest would be requesting BlobInfo instead of BlobAll when there is a cache hit in the MockNetworkClient.

yes will do

ambry-commons/src/main/java/com/github/ambry/commons/AmbryCache.java
```diff
@@ -54,12 +56,15 @@ public class AmbryCache {
    * @param cacheEnabled Toggles cache. If true, cache is enabled. Else, cache is disabled.
    * @param cacheMaxSizeBytes Maximum memory footprint of the cache
    * @param metricRegistry Instance of metrics registry to record stats
+   * @param notificationSystem
    */
-  public AmbryCache(String cacheId, boolean cacheEnabled, long cacheMaxSizeBytes, MetricRegistry metricRegistry) {
+  public AmbryCache(String cacheId, boolean cacheEnabled, long cacheMaxSizeBytes, MetricRegistry metricRegistry,
```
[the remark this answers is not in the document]
snalli: On second thought, I think I found a way to do it without notifications. I'll implement and publish.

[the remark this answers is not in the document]
justinlin-linkedin: The point is to not use NotificationSystem, since NotificationSystem was created for a specific reason and adding cache hit and miss event to NotificationSystem is not good idea. Let's make sure each class follow the true purpose of their creation. 
That being said, I don't oppose passing a "CacheEventCallback" or something like that, as long as it's not NotificationSystem.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -1367,6 +1365,258 @@ public void onEvent(ReplicaId replicaId, Object e) {
...
+      assertTrue("Must encounter a cache miss", ambryCacheWithMetrics.getCacheMissCountDown().await(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
+      assertTrue("Must put metadata in cache", ambryCacheWithMetrics.getPutObjectCountDown().await(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS));
+      assertNotNull("Blob metadata must be present in metadata cache", router.getBlobMetadataCache().getObject(blobId));
+
+      // failed delete
+      mockServerLayout.getMockServers().forEach(mockServer -> mockServer.setServerErrorForAllRequests(serverErrorCode));
+      TestCallback<Void> testCallback = new TestCallback<>();
+      Future<Void> future = router.deleteBlob(blobId, null, testCallback, null);
+      assertFailureAndCheckErrorCode(future, testCallback, routerErrorCode);
+      assertNotNull("Blob metadata must be present in metadata cache", router.getBlobMetadataCache().getObject(blobId));
```
Arun-LinkedIn, line 1597: This is not related to tests. But just wanted to note that similar to how we are invalidating cache when blob is not found during GetBlob operations, we should update logic sometime later to invalidate cache when blob is not found during GETBlobInfo, TTLUpdate or DELETE operations too. 

This will avoid BlobInfo call during later GET operations. Just wondering that even though one BlobInfo is not costly, since we already saw the blob is gone, we  should probably make use of it.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -1367,6 +1365,258 @@ public void onEvent(ReplicaId replicaId, Object e) {
...
+      setRouterWithMetadataCache(properties, new AmbryCacheMetrics(1,1,1,0));
+      AmbryCacheWithMetrics ambryCacheWithMetrics = (AmbryCacheWithMetrics) router.getBlobMetadataCache();
+      assertNotNull("Metadata cache must be created", ambryCacheWithMetrics);
+
+      // put a simple blob and test it is absent in the metadata cache
+      setOperationParams(PUT_CONTENT_SIZE, TTL_SECS);
+      blobId =
+          router.putBlob(putBlobProperties, putUserMetadata, putChannel, new PutBlobOptionsBuilder().build())
+              .get(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+      assertTrue("BlobID must be version 6 blobID", BlobId.getVersion(blobId) == BlobId.BLOB_ID_V6);
```
Arun-LinkedIn, line 1384: nit: This check can be removed too since this is not related to this test.

Arun-LinkedIn (commented):
Tests in NonBlockingRouterTest look good. But, like you and Justin are discussing, we should probably add unit tests in GetBlobOperationTest too to probably check cases like firstChunk is instance of CachedFirstChunk, metadata is saved correctly in Cache, operation fails when blob information is different from cache, etc.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
snalli: got it

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterFactory.java
```diff
@@ -123,7 +123,7 @@ public NonBlockingRouterFactory(VerifiableProperties verifiableProperties, Clust
   public Router getRouter() {
     try {
       return new NonBlockingRouter(routerConfig, routerMetrics, networkClientFactory, notificationSystem, clusterMap,
-          kms, cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass);
+          kms, cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass, null);
```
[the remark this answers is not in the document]
snalli: good catch !

ambry-router/src/main/java/com/github/ambry/router/CloudRouterFactory.java
```diff
@@ -145,9 +146,13 @@ public Router getRouter() throws InstantiationException {
...
           getRequestHandlerPool(verifiableProperties, clusterMap, cloudDestination, cloudConfig);
       CompositeNetworkClientFactory networkClientFactory = getCompositeNetworkClientFactory(requestHandlerPool);
+      AmbryCache blobMetadataCache = new AmbryCache(routerConfig.routerBlobMetadataCacheId,
+          routerConfig.routerBlobMetadataCacheEnabled,
+          routerConfig.routerBlobMetadataCacheMaxSizeBytes,
+          routerMetrics.getMetricRegistry());
       NonBlockingRouter router =
           new NonBlockingRouter(routerConfig, routerMetrics, networkClientFactory, notificationSystem, clusterMap, kms,
-              cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass);
+              cryptoService, cryptoJobHandler, accountService, time, defaultPartitionClass, blobMetadataCache);
```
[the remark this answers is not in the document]
snalli: good catch! this should've been in the previous file

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -206,7 +218,7 @@ public static List<Object[]> data() {
    * @param isBandwidthThrottlingEnabled value for
    * {@link com.github.ambry.config.QuotaConfig#bandwidthThrottlingFeatureEnabled}.
    */
-  public GetBlobOperationTest(String operationTrackerType, boolean testEncryption, boolean isBandwidthThrottlingEnabled)
+  public GetBlobOperationTest(String operationTrackerType, boolean testEncryption, boolean isBandwidthThrottlingEnabled, boolean enableMetadataCache)
```
Arun-LinkedIn, line 221: nit: @param may be missing for enableMetadataCache

Arun-LinkedIn (approved):

mikeyang923 (approved):
lgtm

PR #2158, opened 2022-07-22
Check for time outs correctly before dropping a request at Router

Arun-LinkedIn:
Currently when we drop a request at router due to no response from network, we check for the time elapsed since its creation at Router. But it could be possible that some of the requests spent majority of time at router layer (due to bandwidth throttling, etc) before reaching Network layer. Such requests would expire even before they are sent to network or wouldn't be given sufficient time to receive response from network. To handle such cases, this change tries to check for time elapsed since a request was received at network layer. In addition, it also checks over all time elapsed at router layer to avoid negative cases where requests get stuck at router layer itself. 

Also removed wrapper classes like `ChunkPutRequestInfo`, `ChunkGetRequestInfo`, etc and used `RequestInfo` itself for book keeping at router.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -33,6 +33,10 @@ public class RouterConfig {
   public static final double DEFAULT_LATENCY_TOLERANCE_QUANTILE = 0.9;
   public static final long DEFAULT_OPERATION_TRACKER_HISTOGRAM_CACHE_TIMEOUT_MS = 1000L;
   public static final long ROUTER_NOT_FOUND_CACHE_MAX_TTL_IN_MS = 24 * 60 * 1000L;
+  public static final int MAX_NETWORK_TIMEOUT_VALUE_FOR_A_REQUEST_IN_MS = 60 * 1000;
+  // This is a theoretical maximum value. Configured value may be much smaller since we might need to respond back to
+  // client with either success or failure much sooner.
+  public static final int MAX_OVERALL_TIMEOUT_VALUE_FOR_A_REQUEST_IN_MS = 60 * 60 * 1000;
 
   // config keys
   public static final String ROUTER_SCALING_UNIT_COUNT = "router.scaling.unit.count";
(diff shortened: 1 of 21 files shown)


ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
```diff
@@ -85,27 +105,67 @@ public ReplicaId getReplicaId() {
...
+   */
+  public long getResponseHeaderReceiveTime() {
+    return responseHeaderReceiveTime;
+  }
+
+  /**
+   * @return {@code True} if request has been received by network layer.
+   */
+  public boolean isRequestReceivedByNetworkLayer() {
+    return requestEnqueueTime != 1;
```
ankagrawal, line 168: != -1

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
```diff
@@ -142,20 +143,25 @@ void poll(RequestRegistrationCallback<GetOperation> requestRegistrationCallback)
...
-      Map.Entry<Integer, GetRequestInfo> entry = inFlightRequestsIterator.next();
+      Map.Entry<Integer, RequestInfo> entry = inFlightRequestsIterator.next();
       int correlationId = entry.getKey();
-      GetRequestInfo info = entry.getValue();
-      if (time.milliseconds() - info.startTimeMs > routerConfig.routerRequestTimeoutMs) {
+      RequestInfo requestInfo = entry.getValue();
+      // If there is no response from network layer or if request itself didn't get chance to reach chance to reach
+      // network layer (due to some throttling in router, etc), drop the request after some time.
+      long currentTimeInMs = time.milliseconds();
+      if ((requestInfo.isRequestReceivedByNetworkLayer()
```
ankagrawal, line 155: We should abstract this if condition into a separate method (maybe RouterUtils.isRouterRequestExpired()). Its the same condition for all Operations.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -182,12 +183,19 @@ public class RouterConfig {
   public final int routerConnectionCheckoutTimeoutMs;
 
   /**
-   * Timeout for requests issued by the router to the network layer.
+   * Timeout for requests waiting at the router layer.
    */
   @Config(ROUTER_REQUEST_TIMEOUT_MS)
-  @Default("2000")
+  @Default("4000")
   public final int routerRequestTimeoutMs;
```
ankagrawal, line 190: We should set a very high default time for this (maybe 1 hour?). Its possible for an account to issue a ttlupdate for a composite blob, and the last of those chunk requests to be stuck in quota queue for long because the account doesn't have sufficient quota.

ankagrawal (commented):
Mostly looks good. Couple of comments.

justinlin-linkedin (approved):

ankagrawal (approved):

ambry-network/src/main/java/com/github/ambry/network/http2/Http2ClientStreamStatsHandler.java
```diff
@@ -44,10 +44,10 @@ protected void channelRead0(ChannelHandlerContext ctx, Http2Frame frame) throws
     RequestInfo requestInfo = ctx.channel().attr(Http2NetworkClient.REQUEST_INFO).get();
     if (requestInfo != null) {
       requestInfo.responseFramesCount++;
-      long time = System.currentTimeMillis() - requestInfo.getStreamSendTime();
+      long time = System.currentTimeMillis() - requestInfo.getRequestSendTime();
```
SophieGuo410, line 47: QQ: do we need to specially handle the case when requestInfo.getRequestSendTime() == -1?

SophieGuo410 (approved):
Overall LGTM, just have some questions.

ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
```diff
@@ -50,6 +51,25 @@ public RequestInfo(String host, Port port, SendWithCorrelationId request, Replic
...
+  /**
+   * Construct a RequestInfo with the given parameters
+   * @param host the host to which the data is meant for
+   * @param port the port on the host to which the data is meant for
+   * @param request the data to be sent.
+   * @param replicaId the {@link ReplicaId} associated with this request.
+   * @param chargeable the {@link Chargeable} associated with this request.
+   * @param creationTime the creation time of this request in msec.
+   */
+  public RequestInfo(String host, Port port, SendWithCorrelationId request, ReplicaId replicaId, Chargeable chargeable,
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sorry for the confusion. Yeah, they are actually same in production code. I had to do so that it is compatible with test code. Since we are using `MockTime` in tests to mimic time movement, etc, had to use `time.milliseconds` instead of directly calling system time function.

ambry-network/src/main/java/com/github/ambry/network/http2/Http2ClientStreamStatsHandler.java
Arun-LinkedIn: Hmm, I think this is called when we get response from server. Due to that, just thinking `requestSendTime` should have a valid time by now.

PR #2160, opened 2022-07-25
Fixed the setEndOffsets null pointer issue 

JingQianCloud:
Fixed the setEndOffsets null pointer issue when there was some compaction log temp file.

To associate index segment files with log segment file, index segment file should start with name.startsWith(logSegmentName.toString()+BlobStore.SEPARATOR),
not only name.startsWith(logSegmentName.toString()) because 0_49_18_index is not a index segment of  0_4_log

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -2303,7 +2303,7 @@ class PersistentIndex {
    * @param logSegmentName the name of the log segment whose index segment files are required.
    * @return the list of {@link IndexSegment} files that refer to the log segment with name {@code logSegmentName}.
    */
-  static File[] getIndexSegmentFilesForLogSegment(String dataDir, final LogSegmentName logSegmentName) { // ERROR
+  static File[] getIndexSegmentFilesForLogSegment(String dataDir, final LogSegmentName logSegmentName) {
     return new File(dataDir).listFiles(new FilenameFilter() {
       @Override
       public boolean accept(File dir, String name) {
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -2311,7 +2311,7 @@ class PersistentIndex {
         // index name has the format of <offset>_index, for example 0_index, 265_index
         // In this case, input parameter logSegmentName == ""
         if (logSegmentName.toString().isEmpty())
-          return name.startsWith(logSegmentName.toString()) && name.endsWith(IndexSegment.INDEX_SEGMENT_FILE_NAME_SUFFIX);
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -2303,11 +2303,17 @@ void setGetBlobReadInfoTestCallback(Runnable getBlobReadInfoTestCallback) {
...
+  static File[] getIndexSegmentFilesForLogSegment(String dataDir, final LogSegmentName logSegmentName) { // ERROR
     return new File(dataDir).listFiles(new FilenameFilter() {
       @Override
       public boolean accept(File dir, String name) {
-        return name.startsWith(logSegmentName.toString()) && name.endsWith(IndexSegment.INDEX_SEGMENT_FILE_NAME_SUFFIX);
+        // If not segmented, log name is log_current
+        // index name has the format of <offset>_index, for example 0_index, 265_index
+        // In this case, input parameter logSegmentName == ""
+        if (logSegmentName.toString().isEmpty())
+          return name.startsWith(logSegmentName.toString()) && name.endsWith(IndexSegment.INDEX_SEGMENT_FILE_NAME_SUFFIX);
```
justinlin-linkedin, line 2314: we don't have to check startsWith if we know the logSegmentName.toString() is empty, we can just check the endsWith part.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -2303,11 +2303,17 @@ void setGetBlobReadInfoTestCallback(Runnable getBlobReadInfoTestCallback) {
    * @param logSegmentName the name of the log segment whose index segment files are required.
    * @return the list of {@link IndexSegment} files that refer to the log segment with name {@code logSegmentName}.
    */
-  static File[] getIndexSegmentFilesForLogSegment(String dataDir, final LogSegmentName logSegmentName) {
+  static File[] getIndexSegmentFilesForLogSegment(String dataDir, final LogSegmentName logSegmentName) { // ERROR
```
justinlin-linkedin, line 2306: also method `cleanupIndexSegmentFilesForLogSegment` has to be updated as well.

JingQianCloud: Good catch.

SophieGuo410, line 2306: nit: can you remove this comment "// ERROR"?

SophieGuo410 (approved):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
JingQianCloud: Oh. I missed it. Thanks for catching it.

justinlin-linkedin (approved):

PR #2161, opened 2022-07-26
[Frontend] Add logs in AmbryCache

snalli:
This patch adds more debug logging to AmbryCache.
It sets defaults for localhost testing as well.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
@@ -512,7 +512,10 @@ class NonBlockingRouter implements Router {
         }
         initiateBackgroundDeletes(deleteRequests);
         if (blobMetadataCache != null) {
-          blobMetadataCache.deleteObject(blobIdStr);
+          boolean deleteResult = blobMetadataCache.deleteObject(blobIdStr);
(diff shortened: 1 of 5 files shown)


ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -512,7 +512,10 @@ void initiateChunkDeletesIfAny(final String blobIdStr, final String serviceId,
         }
         initiateBackgroundDeletes(deleteRequests);
         if (blobMetadataCache != null) {
-          blobMetadataCache.deleteObject(blobIdStr);
+          boolean deleteResult = blobMetadataCache.deleteObject(blobIdStr);
+          logger.debug("[{}] Issued delete-metadata for blobId = {}, reason = {}, result = {}", blobMetadataCache.getCacheId(),
```
SophieGuo410, line 516: nit: if it's just constant string you can directly add in the first string input.
reason = Background delete operation
or probably operation = Background delete operation makes more sense.

snalli: done

SophieGuo410 (approved):

Arun-LinkedIn (approved):

PR #2162, opened 2022-07-27
Log the timeout reason for router request timeouts.

ankagrawal:


ambry-router/src/main/java/com/github/ambry/router/RouterUtils.java
```diff
@@ -296,11 +296,34 @@ public static Response mapToReceivedResponse(Response sentResponse) {
...
+  }
+
+  /**
+   * {@link Enum} All the reasons for router request expiry.
+   */
+  public enum RouterRequestExpiryReason {
+    /**
+     * No timeout.
+     */
+    NO_TIMEOUT,
```
Arun-LinkedIn, line 319: Minor: Would it easier to follow if we check if request expired or not and then check reason only if it expired?   

Could think of below structure. 

```
 public class RouterRequestExpiryState{
     boolean isExpired;
     RequestExpiryReason reason;
}
```

Arun-LinkedIn (approved):

PR #2163, opened 2022-07-27
[Frontend] Update TTL in metadata cache path

snalli:
This patch updates the ttl of the received metadata so that it can
be compared with the cached metadata. Apparently, the received metadata
isn't up-to-date and must be fixed upon receipt.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -1847,6 +1847,7 @@ class GetBlobOperation extends GetOperation {
         throws IOException, MessageFormatException {
       /* If we find blob metadata in frontend cache, validate it and use it. */
       BlobProperties receivedBlobProperties = MessageFormatRecord.deserializeBlobProperties(payload);
+      updateTtlIfRequired(receivedBlobProperties, messageInfo);
       /*
        * Some cached variables are mutable.
        * Update them and then compare if other immutable variables are same or not.
@@ -1864,7 +1865,9 @@ class GetBlobOperation extends GetOperation {
          * We do not need to decrypt anything at this point.
          */
         String reason = "Cached blob property does not match received blob property";
-        logger.error("{} for blobId = {}, cached blob property = {}, received blob property = {}", reason, blobId,
+        logger.error("[{}] {} for blobId = {}, cached blob property = {}, received blob property = {}",
+            blobMetadataCache.getCacheId(),
+            reason, blobId,
             serverBlobProperties, receivedBlobProperties);
         deleteMetadata(reason);
         onInvalidCacheEntry(new RouterException(reason, RouterErrorCode.UnexpectedInternalError));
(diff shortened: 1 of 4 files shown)


SophieGuo410 (approved):

Arun-LinkedIn (approved):

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1864,7 +1865,9 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
          * We do not need to decrypt anything at this point.
          */
         String reason = "Cached blob property does not match received blob property";
-        logger.error("{} for blobId = {}, cached blob property = {}, received blob property = {}", reason, blobId,
+        logger.error("[{}] {} for blobId = {}, cached blob property = {}, received blob property = {}",
+            blobMetadataCache.getCacheId(),
+            reason, blobId,
```
justinlin-linkedin, line 1870: nit: please format this file.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -1492,6 +1492,42 @@ public void testRouterMetadataCacheDeleteCompositeBlob() throws Exception {
     }
   }
 
+  @Test
+  public void testRouterMetadataCacheUpdateTTLCompositeBlob() throws Exception {
+    try {
+      Properties properties = getNonBlockingRouterProperties("DC1");
+      properties.setProperty("router.blob.metadata.cache.enabled", Boolean.toString(true));
+      properties.setProperty("router.smallest.blob.for.metadata.cache", Long.toString(0));
+      setRouterWithMetadataCache(properties, new AmbryCacheMetrics(1,1,1,1));
```
justinlin-linkedin, line 1501: Sorry I was sick a couple days ago and didn't have enough time to review the unit test code, but I think the name of "AmbryCacheMetrics" is confusing, especially on "Metrics" part. 
We have a convention to use "Metric" name when it's actually the metrics, like histogram, timer, counter etc.. This object is not metrics, it's more of stats. I suggest we change the name to something like AmbryCacheStats? What do you think?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: this file or this line ?

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
snalli: I am fine with name but I'd prefer to do it in a separate follow up patch rather than this one. There would be many renames across many files.

snalli: I mean I am fine with renaming, but in a follow up patch.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: done

snalli:
Got two approvals. If the maintainers can merge, I could deploy.

justinlin-linkedin (approved):

PR #2164, opened 2022-07-27
[Frontend] Reduce impact of delete on metadata cache

snalli:
The delete-metadata codepath in GetBlobOperation is unguarded.
As a result, we issue delete-metadata even for simple blobs that do not
exist. The cache sees a massive load of such requests. This patch aims
to reduce the impact of such requests.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -185,6 +185,26 @@ class GetBlobOperation extends GetOperation {
   }
 
   /**
+   * This helper decides if metadata must be deleted from cache or not.
+   * This mainly reduces the impact on metadata cache due to a large number of GET requests for
+   * simple blobs that do not exist.
+   * @param abortCause Reason to abort GetBlobOperation
+   * @return True if metadata must be deleted, false otherwise.
+   */
+  boolean shouldDeleteMetadata(Exception abortCause) {
+    if (abortCause != null && abortCause instanceof RouterException) {
+      switch (((RouterException) abortCause).getErrorCode()) {
+        /* If the blob is not found, then delete its metadata from frontend cache. */
+        case BlobDoesNotExist:
+        case BlobDeleted:
+        case BlobExpired:
+          return (blobMetadataCache != null && blobId.getBlobDataType() == BlobId.BlobDataType.METADATA);
+      }
+    }
+    return false;
+  }
+
+  /**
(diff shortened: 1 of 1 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -184,6 +184,26 @@ boolean shouldLookupMetadataCache() {
...
+  /**
+   * This helper decides if metadata must be deleted from cache or not.
+   * This mainly reduces the impact on metadata cache due to a large number of GET requests for
+   * simple blobs that do not exist.
+   * @param abortCause Reason to abort GetBlobOperation
+   * @return True if metadata must be deleted, false otherwise.
+   */
+  boolean shouldDeleteMetadata(Exception abortCause) {
+    if (abortCause != null && abortCause instanceof RouterException) {
+      switch (((RouterException) abortCause).getErrorCode()) {
```
SophieGuo410, line 196: In this case, you could just use the previous way for switch and add break under the case right? 
Can you also add the default for the switch statement. 
And QQ: since BlobExpired is the only case we want to take action, why don't you just use a if statement.

snalli: Sorry not clear what you mean by the first suggestion of adding a break.
And we do want to handle all three err codes and not just BlobExpired. Here is why:
1. node 1, putBlob - success
2. node 1, getBlob(range request) - success, caches metadata
3. node 2, deleteBlob 
4. node 1, getBlob(range request) - gets a blobDoesNotExist and we must clear its cache

snalli: Also the cache is only for composite blobs. So I jsut want to filter out the simple blobs. Hence this PR.

snalli: If you are wondering if the old code and new code are same, they're not. There is one difference. In the old, we'd attempt deleting metadata even for non-existent simple blobs. But in the new one, we'd only attempt deleting metadata for composite blobs.

Arun-LinkedIn (approved):

justinlin-linkedin (approved):
LGTM

SophieGuo410 (approved):

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
SophieGuo410: thx for the explanation, I missed that part.

PR #2166, opened 2022-07-28
Increase network timeout for router requests during heavy load 

Arun-LinkedIn:
Since it is possible that the response times for requests sent from ambry-frontend to ambry-server may increase under heavy load (For example, we could have lot of simultaneous TTLUpdate operations on chunks of a large composite blob), we might need to set higher time outs for requests going out so that they don't expire. 

This change uses the number of in-flight requests between frontend router and server to calculate approximate time needed for the response and accordingly increases the time out of the request.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
@@ -38,6 +38,8 @@ public class Http2ClientConfig {
       "http2.blocking.channel.pool.shutdown.timeout.ms";
   public static final String HTTP2_PEER_CERTIFICATE_SAN_REGEX = "http2.peer.certificate.san.regex";
   public static final String HTTP2_TIMEOUT_AS_NETWORK_ERROR = "http.timeout.as.network.error";
+  public static final String HTTP2_REQUEST_ADDITIONAL_TIMEOUT_MS = "http2.request.additional.timeout.ms";
+  public static final String HTTP2_REQUEST_COUNT_FOR_SCALING_TIMEOUT = "http2.request.count.for.scaling.timeout";
 
   /**
    * HTTP/2 connection idle time before we close it. -1 means no idle close.
@@ -167,6 +169,22 @@ public class Http2ClientConfig {
   @Default("false")
   public final boolean http2TimeoutAsNetworkError;
 
+  /**
+   * Denotes how much extra time should be given for requests to receive response under heavy load.
+   */
+  @Config(HTTP2_REQUEST_ADDITIONAL_TIMEOUT_MS)
+  @Default("2500")
+  public final int http2RequestAdditionalTimeoutMs;
+
+  /**
+   * Denotes the number of requests to consider when increasing the network time out. For example, if
+   * {@link this#http2RequestAdditionalTimeoutMs} is 2500 and this value is 100000, it means that we will
+   * increment network timeout by 2.5 seconds for every 100K requests.
+   */
+  @Config(HTTP2_REQUEST_COUNT_FOR_SCALING_TIMEOUT)
+  @Default("100000")
+  public final int http2RequestCountForScalingTimeout;
+
   public Http2ClientConfig(VerifiableProperties verifiableProperties) {
     idleConnectionTimeoutMs = verifiableProperties.getLong(HTTP2_IDLE_CONNECTION_TIMEOUT_MS, -1);
     http2MinConnectionPerPort =
@@ -191,5 +209,7 @@ public class Http2ClientConfig {
         verifiableProperties.getInt(HTTP2_BLOCKING_CHANNEL_POOL_SHUTDOWN_TIMEOUT_MS, 3000);
     http2PeerCertificateSanRegex = verifiableProperties.getString(HTTP2_PEER_CERTIFICATE_SAN_REGEX, "");
     http2TimeoutAsNetworkError = verifiableProperties.getBoolean(HTTP2_TIMEOUT_AS_NETWORK_ERROR, false);
+    http2RequestAdditionalTimeoutMs = verifiableProperties.getInt(HTTP2_REQUEST_ADDITIONAL_TIMEOUT_MS, 2500);
+    http2RequestCountForScalingTimeout = verifiableProperties.getInt(HTTP2_REQUEST_COUNT_FOR_SCALING_TIMEOUT, 100000);
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
@@ -15,7 +15,6 @@ package com.github.ambry.network;
 
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.quota.Chargeable;
-import java.util.concurrent.atomic.AtomicLong;
 
 
 /**
@@ -32,8 +31,12 @@ public class RequestInfo {
   private long requestEnqueueTime = -1;
   private long requestSendTime = -1;
   private long responseHeaderReceiveTime = -1;
-  private final AtomicLong networkTimeOutMs = new AtomicLong(0);
-  private final AtomicLong finalTimeOutMs = new AtomicLong(0);
+  // Determines the time to wait for response once it is sent to server
+  private long networkTimeOutMs;
+  // Determines the time to wait for response once it is created. It is possible that request remains in router queues
+  // due to insufficient quota. This time out helps the request to avoid being stuck in queues for a long time and
+  // also enables to send response to client in a bounded time.
+  private final long finalTimeOutMs;
   public int responseFramesCount = 0;
 
   /**
@@ -55,8 +58,8 @@ public class RequestInfo {
     this.replicaId = replicaId;
     this.chargeable = chargeable;
     this.requestCreateTime = creationTime;
-    this.networkTimeOutMs.set(networkTimeOutMs);
-    this.finalTimeOutMs.set(finalTimeOutMs);
+    this.networkTimeOutMs = networkTimeOutMs;
+    this.finalTimeOutMs = finalTimeOutMs;
   }
 
   /**
@@ -177,24 +180,24 @@ public class RequestInfo {
   }
 
   /**
-   * @return the time to wait for response from server.
+   * @return the time to wait for response once this request is sent to server.
    */
   public long getNetworkTimeOutMs() {
-    return networkTimeOutMs.get();
+    return networkTimeOutMs;
   }
 
   /**
-   * Increase the time to wait for response from server.
+   * Increases the time to wait for response from server.
    * @param delta time in milliseconds.
    */
   public void incrementNetworkTimeOutMs(long delta) {
-    networkTimeOutMs.getAndAdd(delta);
+    networkTimeOutMs += delta;
   }
 
   /**
-   * @return the overall wait time for a request in router.
+   * @return the overall wait time for this request in router.
    */
   public long getFinalTimeOutMs() {
-    return finalTimeOutMs.get();
+    return finalTimeOutMs;
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
@@ -15,6 +15,7 @@ package com.github.ambry.network;
 
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.quota.Chargeable;
+import java.util.concurrent.atomic.AtomicLong;
 
 
 /**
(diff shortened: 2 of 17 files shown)


justinlin-linkedin (approved):
LGTM

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -193,8 +193,8 @@ boolean saveMetadata() {
     if (firstChunk.shouldSaveMetadata()) {
       BlobMetadata blobMetadata = new BlobMetadata(blobId.toString(), blobInfo, compositeBlobInfo);
       putResult = blobMetadataCache.putObject(blobMetadata.getBlobId(), blobMetadata);
-      logger.debug("[{}] Issued save-metadata for blobId = {}, result = {}", blobMetadata.getBlobId(),
-          blobId, putResult);
+      logger.debug("[{}] Issued save-metadata for blobId = {}, result = {}", blobMetadata.getBlobId(), blobId,
```
snalli, line 196: these are formatting changes. can these be avoided ?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -209,8 +209,8 @@ boolean deleteMetadata(String reason) {
       return false;
     }
     boolean deleteResult = blobMetadataCache.deleteObject(blobId.toString());
-    logger.debug("[{}] Issued delete-metadata for blobId = {}, reason = {}, result = {}", blobMetadataCache.getCacheId(),
-        blobId, reason, deleteResult);
+    logger.debug("[{}] Issued delete-metadata for blobId = {}, reason = {}, result = {}",
+        blobMetadataCache.getCacheId(), blobId, reason, deleteResult);
```
snalli, line 213: these are formatting changes. can these be avoided ?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -896,10 +896,11 @@ private void cleanupExpiredInFlightRequests(RequestRegistrationCallback<GetOpera
...
           requestRegistrationCallback.registerRequestToDrop(correlationId);
           inFlightRequestsIterator.remove();
-        } else {
-          // the entries are ordered by correlation id and time. Break on the first request that has not timed out.
-          break;
         }
+        // Note: Even though the entries are ordered by correlation id and their creation time, we cannot break out of
+        // the while loop when we find an unexpired entry. This is because time outs for all requests may not be equal
+        // now. For example, we assign higher time outs when we find that there are lot of outstanding requests by
+        // letting Network client dynamically update the timeout via RequestInfo#incrementNetworkTimeOutMs.
```
snalli, line 903: This comment is not clear. Is the time out per operation or per request within the operation ? If we find the first request that has not yet timed, can't we assert that subsequent ones would not have timed out since their expiry would be further in the future ?

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
```diff
@@ -182,7 +183,8 @@ private void fetchRequests(RequestRegistrationCallback<GetOperation> requestRegi
       Port port = RouterUtils.getPortToConnectTo(replicaId, routerConfig.routerEnableHttp2NetworkClient);
       GetRequest getRequest = createGetRequest(blobId, getOperationFlag(), options.getBlobOptions.getGetOption());
       RequestInfo requestInfo =
-          new RequestInfo(hostname, port, getRequest, replicaId, operationQuotaCharger, time.milliseconds());
+          new RequestInfo(hostname, port, getRequest, replicaId, operationQuotaCharger, time.milliseconds(),
+              routerConfig.routerRequestNetworkTimeoutMs, routerConfig.routerRequestTimeoutMs);
```
snalli, line 187: Also why make this change for all operations ? GetBlobInfo would never timeout no matter how large the blob.

ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
```diff
@@ -167,6 +169,22 @@ public class Http2ClientConfig {
   @Default("false")
   public final boolean http2TimeoutAsNetworkError;
 
+  /**
+   * Denotes how much extra time should be given for requests to receive response under heavy load.
+   */
+  @Config(HTTP2_REQUEST_ADDITIONAL_TIMEOUT_MS)
```
ankagrawal, line 175: Nit: technically this is not heavy load. This will be needed even if there is only one request, and that request is a very large update request. Also this will apply only for ttlupdate right? Because for GET and PUT, number of server requests will be bound by parallelism. And deletes are background.

So maybe we should say something like `http2.request.large.blob.ttlupdate.timeout.ms`

ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
```diff
@@ -167,6 +169,22 @@ public class Http2ClientConfig {
...
+  public final int http2RequestAdditionalTimeoutMs;
+
+  /**
+   * Denotes the number of requests to consider when increasing the network time out. For example, if
+   * {@link this#http2RequestAdditionalTimeoutMs} is 2500 and this value is 100000, it means that we will
+   * increment network timeout by 2.5 seconds for every 100K requests.
+   */
+  @Config(HTTP2_REQUEST_COUNT_FOR_SCALING_TIMEOUT)
+  @Default("100000")
+  public final int http2RequestCountForScalingTimeout;
```
ankagrawal, line 186: Can we use chunkCount instead of request count? Also make it explicit that this is for update ttl only.

ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
```diff
@@ -31,6 +32,8 @@ public class RequestInfo {
   private long requestEnqueueTime = -1;
   private long requestSendTime = -1;
   private long responseHeaderReceiveTime = -1;
+  private final AtomicLong networkTimeOutMs = new AtomicLong(0);
```
ankagrawal, line 35: Why atomic? For one `RequestInfo` object, would multiple threads try to update the timeout value?

ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
```diff
@@ -31,6 +32,8 @@ public class RequestInfo {
...
+  private final AtomicLong finalTimeOutMs = new AtomicLong(0);
```
ankagrawal, line 36: Nit: these aren't timeout values. These are timeTaken/timeSpent.

ambry-network/src/test/java/com/github/ambry/network/http2/RequestsStatsHandlerTest.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  static class TestHandler extends ChannelInboundHandlerAdapter {
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+      throw new RuntimeException("Network error");
+    }
+  }
+}
```
ankagrawal, line 80: nit: newline?

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -284,10 +285,11 @@ private void cleanupExpiredInflightRequests(
         onErrorResponse(requestInfo.getReplicaId(),
             RouterUtils.buildTimeoutException(correlationId, requestInfo.getReplicaId().getDataNodeId(), blobId));
         requestRegistrationCallback.registerRequestToDrop(correlationId);
-      } else {
-        // the entries are ordered by correlation id and time. Break on the first request that has not timed out.
-        break;
       }
+      // Note: Even though the entries are ordered by correlation id and their creation time, we cannot break out of
```
ankagrawal, line 289: Can this have any performance effect?

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
ankagrawal: I guess the `RequestInfo` for `GetBlobInfo` request would be queued to the account's quota queue, and might not be taken up until ttlupdates for all chunks issued before the `GetBlobInfo` are done.
This is actually worrying. This means that if a user issues a ttlupdate for a 1TiB blob, all other requests to the same frontend would take a very very long time to make any progress. We need to think of a solution for this case.

ankagrawal (changes requested):

ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
snalli: There is some confusion between the terminologies here. By request do u mean one user request or do u mean the requests between frontend and backend ?

ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
snalli: +1

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
snalli: Same question. This change makes us loop all requests

ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
Arun-LinkedIn: @snalli Yeah, these are requests between frontend and backend. Since these are used in the context of network layer between Frontend/Server, may be 'requests' terminology should be okay.

@ankagrawal Yeah, currently will only have such large number of requests for TTLUpdate operations. But, am trying to keep this PR generic and increase time outs dynamically based on request count. For example, may be having very high QPS of operations at Frontend, etc?. 

Regarding the naming, just wondering keeping 'request' is still okay since this a Network layer configuration which only knows of request-response and might now know of operations, chunks, etc.

ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
Arun-LinkedIn: Hmm, similar as above comment. Since we are using this configuration at frontend/server network layer which only knows of request-responses, wondering if `request count` is still okay.

ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
Arun-LinkedIn: Hmm, these are actually timeout values for this request. Just that they are incremented at run time on top of configured value.

ambry-api/src/main/java/com/github/ambry/network/RequestInfo.java
Arun-LinkedIn: Hmm, yeah was just thinking if it is possible that read and increment could be called in parallel. But, since each `requestInfo` will be accessed by only one request handler thread during `sendAndPoll` operation, may be we don't need Atomic operation. Will remove it.

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
Arun-LinkedIn: I was just wondering if it might be okay for now size of the Map is currently fixed/small since we can have probably maximum of 3 requestInfos (determined by the router request parallelism factor) for each operation at a time  or may be more if we use adaptive operation tracker.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
Arun-LinkedIn: These are picked up automatically when trying to format the file. Since there are not many formatting changes, left them here. But in case there are too many, will remove them to help with review.

ambry-api/src/main/java/com/github/ambry/config/Http2ClientConfig.java
justinlin-linkedin: I would argue that there is a real chance that other requests are going to see the same the pattern. 
When there is a TTL_UPDATE request on a huge blob, there will be lots of ttl_update frontend requests going to server in Http2NetworkClient. If other requests arrive at the same time, they will be affected as well since they are sharing on NetworkClient.

It's true that BackgroundDeleter doesn't share the same NetworkClient with the regular OperationController, but if we are trying to delete a huge blob, (just like ttl_update a huge blob), background deleter would probably gets stuck with a heavy load as well.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
Arun-LinkedIn: Yeah, in almost all of the cases, when we find a request that has not yet timed out, subsequent requests would also be valid. But, since the time outs of all requests is not fixed now, it is possible that later request in the queue has shorter time out. This helps to clean up such requests. I think performance wise it might be okay since the map would only hold few elements at a time (i.e. number of outstanding requests which should be like 3 or 4). 

Updated comments in code to make them more clear.

ankagrawal (approved):

justinlin-linkedin (approved):

PR #2168, opened 2022-07-31
Log a warning when rejecting request due to reject recommendation.

ankagrawal:
Rename some quota metrics to a more accurate representation of their usage.


SophieGuo410 (approved):

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -196,6 +196,8 @@ private void pollQuotaCompliantRequests(List<RequestInfo> requestsToSend,
             break;
           case REJECT:
             quotaAvailable = false;
+            logger.warn(
+                "Rejecting request for quota resource %s because of reject recommendation.", quotaResource.toString());
```
Arun-LinkedIn, line 200: nit: Probably need to be {} instead of %s for the argument.

Arun-LinkedIn (approved):

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
ankagrawal: fixed.

PR #2170, opened 2022-08-02
Include Disk HealthCheck in Ambry State HealthCheck w/Tests

maskndafi-linkedin:
This PR changes the Ambry HealthCheck Admin Endpoint(AHAE) to include disk healthchecking. The disk healthcheck implementation will CRUD a file asynchronously every minute to determine the health of each disk in the background. The AHAE will fetch the current state of the disk and if any disk isn't healthy or any helix state isn't standby/leader then the server is deemed BAD, otherwise GOOD.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
@@ -32,11 +32,35 @@ public class DiskManagerConfig {
   @Default("false")
   public final boolean diskManagerEnableSegmentPooling;
 
+  @Config("disk.manager.disk.healthcheck.enabled")
+  @Default("false")
+  public final boolean diskManagerDiskHealthCheckEnabled;
+
+  @Config("disk.manager.disk.healthcheck.interval.seconds")
+  @Default("60")
+  public final int diskManagerDiskHealthCheckIntervalSeconds;
+
+  @Config("disk.manager.disk.healthcheck.file.path")
+  @Default("/ambry-data/")
+  public final String diskManagerDiskHealthCheckFilePath;
+
+  @Config("disk.manager.disk.healthcheck.operation.timeout.seconds")
+  @Default("5")
+  public final int diskManagerDiskHealthCheckOperationTimeoutSeconds;
+
   public DiskManagerConfig(VerifiableProperties verifiableProperties) {
     diskManagerReserveFileDirName =
         verifiableProperties.getString("disk.manager.reserve.file.dir.name", "reserve-pool");
     diskManagerRequiredSwapSegmentsPerSize =
         verifiableProperties.getIntInRange("disk.manager.required.swap.segments.per.size", 1, 0, 1000);
     diskManagerEnableSegmentPooling = verifiableProperties.getBoolean("disk.manager.enable.segment.pooling", false);
+    diskManagerDiskHealthCheckEnabled =
+        verifiableProperties.getBoolean("disk.manager.disk.healthcheck.enabled", false);
+    diskManagerDiskHealthCheckIntervalSeconds =
+        verifiableProperties.getInt("disk.manager.disk.healthcheck.interval.seconds", 60);
+    diskManagerDiskHealthCheckFilePath =
+        verifiableProperties.getString("disk.manager.disk.healthcheck.file.path", "/ambry-data/");
+    diskManagerDiskHealthCheckOperationTimeoutSeconds =
+        verifiableProperties.getInt("disk.manager.disk.healthcheck.operation.timeout.seconds", 5);
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
@@ -37,7 +37,7 @@ public class DiskManagerConfig {
   public final boolean diskManagerDiskHealthCheckEnabled;
 
   @Config("disk.manager.disk.healthcheck.interval.seconds")
-  @Default("60")
+  @Default("180")
   public final int diskManagerDiskHealthCheckIntervalSeconds;
 
   @Config("disk.manager.disk.healthcheck.file.path")
--- a/ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
@@ -54,8 +54,7 @@ public class DiskManagerConfig {
     diskManagerRequiredSwapSegmentsPerSize =
         verifiableProperties.getIntInRange("disk.manager.required.swap.segments.per.size", 1, 0, 1000);
     diskManagerEnableSegmentPooling = verifiableProperties.getBoolean("disk.manager.enable.segment.pooling", false);
-    diskManagerDiskHealthCheckEnabled =
-        verifiableProperties.getBoolean("disk.manager.disk.healthcheck.enabled", false);
+    diskManagerDiskHealthCheckEnabled = verifiableProperties.getBoolean("disk.manager.disk.healthcheck.enabled", false);
     diskManagerDiskHealthCheckIntervalSeconds =
         verifiableProperties.getInt("disk.manager.disk.healthcheck.interval.seconds", 60);
     diskManagerDiskHealthCheckFilePath =
--- a/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
@@ -95,7 +95,7 @@ public class AmbryServerRequests extends AmbryRequests {
   AmbryServerRequests(StoreManager storeManager, RequestResponseChannel requestResponseChannel, ClusterMap clusterMap,
       DataNodeId nodeId, MetricRegistry registry, ServerMetrics serverMetrics, FindTokenHelper findTokenHelper,
       NotificationSystem operationNotification, ReplicationAPI replicationEngine, StoreKeyFactory storeKeyFactory,
-      ServerConfig serverConfig, DiskManagerConfig diskManagerConfig,StoreKeyConverterFactory storeKeyConverterFactory,
+      ServerConfig serverConfig, DiskManagerConfig diskManagerConfig, StoreKeyConverterFactory storeKeyConverterFactory,
       StatsManager statsManager, ClusterParticipant clusterParticipant) {
     super(storeManager, requestResponseChannel, clusterMap, nodeId, registry, serverMetrics, findTokenHelper,
         operationNotification, replicationEngine, storeKeyFactory, storeKeyConverterFactory);
@@ -430,14 +430,15 @@ public class AmbryServerRequests extends AmbryRequests {
   private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, AdminRequest adminRequest)
       throws StoreException, IOException {
     ServerHealthStatus serverHealthStatus = ServerHealthStatus.BAD; //used to determine if the server is ever unhealthy
-    JSONArray brokenDisks = new JSONArray();
+    JSONArray brokenDisks = new JSONArray();        // indicate which disks didn't pass the disk healthcheck
+    JSONArray unstablePartitions = new JSONArray(); // indicate which partitions aren't in leader/standby state
 
     if (this.storeManager instanceof StorageManager) {
       StorageManager storageManager = (StorageManager) this.storeManager;
       serverHealthStatus = ServerHealthStatus.GOOD;
 
       //Finds all partitions on this host
-      List<PartitionId> partitionsInThisHost =Collections.list(storageManager.getPartitionToDiskManager().keys());
+      List<PartitionId> partitionsInThisHost = Collections.list(storageManager.getPartitionToDiskManager().keys());
 
       /*
        Checks if the Host's BlobStore exists,started and is Leader/Standby ReplicaState for all Partitions
@@ -447,42 +448,49 @@ public class AmbryServerRequests extends AmbryRequests {
         BlobStore hostBlobStore = (BlobStore) storageManager.getStore(partitionId);
 
         //if any fail, this host isn't healthy
-        if (hostBlobStore == null || !hostBlobStore.isStarted() ||
-            ((hostBlobStore.getCurrentState() != ReplicaState.STANDBY) &&
-                (hostBlobStore.getCurrentState() != ReplicaState.LEADER))) {
+        if (hostBlobStore == null || !hostBlobStore.isStarted() || (
+            (hostBlobStore.getCurrentState() != ReplicaState.STANDBY) && (hostBlobStore.getCurrentState()
+                != ReplicaState.LEADER))) {
+
           serverHealthStatus = ServerHealthStatus.BAD;
-          break;
+          JSONObject unstablePartition = new JSONObject();
+          unstablePartition.put("partitionId", partitionId);
+          unstablePartition.put("reason", hostBlobStore.getCurrentState());
+          unstablePartitions.put(unstablePartition);
         }
       }
 
       /*
        Checks if every disk that's assigned to a partition is also mounted and disk is healthy
        */
-      HashSet<DiskManager> partitionsDiskManagers = new HashSet<>(storageManager.getPartitionToDiskManager().values());
-      for(Map.Entry<DiskId,DiskManager> entry : storageManager.getDiskToDiskManager().entrySet()){
-        DiskManager mountedDiskManager = entry.getValue();
+      if (diskManagerConfig.diskManagerDiskHealthCheckEnabled) {
+        HashSet<DiskManager> partitionsDiskManagers =
+            new HashSet<>(storageManager.getPartitionToDiskManager().values());
+        for (Map.Entry<DiskId, DiskManager> entry : storageManager.getDiskToDiskManager().entrySet()) {
+          DiskManager mountedDiskManager = entry.getValue();
 
-        //Checks Disk healthchecking is enabled and
-        // if a disk isn't healthy then store that disk to return in the response
-        if(diskManagerConfig.diskManagerDiskHealthCheckEnabled &&
-            (!partitionsDiskManagers.contains(mountedDiskManager) ||
-            mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY)){
+          //Checks Disk healthchecking is enabled and
+          // if a disk isn't mounted as expected or isn't healthy then store it as part of the response
+          if (!partitionsDiskManagers.contains(mountedDiskManager)
+              || mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY) {
 
-          serverHealthStatus = ServerHealthStatus.BAD;
-          JSONObject brokenDisk = new JSONObject();
-          brokenDisk.put("disk",entry.getKey().getMountPath());
-          brokenDisk.put("reason",mountedDiskManager.getDiskHealthStatus());
-          brokenDisks.put(brokenDisk);
+            serverHealthStatus = ServerHealthStatus.BAD;
+            JSONObject brokenDisk = new JSONObject();
+            brokenDisk.put("disk", entry.getKey().getMountPath());
+            brokenDisk.put("reason", mountedDiskManager.getDiskHealthStatus());
+            brokenDisks.put(brokenDisk);
+          }
         }
       }
     }
 
     //Initializing the content of the response as a json
     JSONObject contentJSON = new JSONObject();
-    contentJSON.put("health",serverHealthStatus);
-    contentJSON.put("brokenDisks",brokenDisks);
+    contentJSON.put("health", serverHealthStatus);
+    contentJSON.put("brokenDisks", brokenDisks);
+    contentJSON.put("unstablePartitions", unstablePartitions);
 
-    byte[] content =  contentJSON.toString().getBytes(StandardCharsets.UTF_8);
+    byte[] content = contentJSON.toString().getBytes(StandardCharsets.UTF_8);
     ServerErrorCode error = ServerErrorCode.No_Error;
 
     return new AdminResponseWithContent(adminRequest.getCorrelationId(), adminRequest.getClientId(), error, content);
--- a/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
@@ -429,13 +429,13 @@ public class AmbryServerRequests extends AmbryRequests {
    */
   private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, AdminRequest adminRequest)
       throws StoreException, IOException {
-    ServerHealthStatus serverHealthStatus = ServerHealthStatus.BAD; //used to determine if the server is ever unhealthy
+    ServerHealthStatus serverHealthStatus = this.storeManager instanceof StorageManager ? ServerHealthStatus.GOOD
+        : ServerHealthStatus.BAD; //used to determine if the server is ever unhealthy
     JSONArray brokenDisks = new JSONArray();        // indicate which disks didn't pass the disk healthcheck
     JSONArray unstablePartitions = new JSONArray(); // indicate which partitions aren't in leader/standby state
 
     if (this.storeManager instanceof StorageManager) {
       StorageManager storageManager = (StorageManager) this.storeManager;
-      serverHealthStatus = ServerHealthStatus.GOOD;
 
       //Finds all partitions on this host
       List<PartitionId> partitionsInThisHost = Collections.list(storageManager.getPartitionToDiskManager().keys());
@@ -452,7 +452,6 @@ public class AmbryServerRequests extends AmbryRequests {
             (hostBlobStore.getCurrentState() != ReplicaState.STANDBY) && (hostBlobStore.getCurrentState()
                 != ReplicaState.LEADER))) {
 
-          serverHealthStatus = ServerHealthStatus.BAD;
           JSONObject unstablePartition = new JSONObject();
           unstablePartition.put("partitionId", partitionId);
           unstablePartition.put("reason", hostBlobStore.getCurrentState());
@@ -474,7 +473,6 @@ public class AmbryServerRequests extends AmbryRequests {
           if (!partitionsDiskManagers.contains(mountedDiskManager)
               || mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY) {
 
-            serverHealthStatus = ServerHealthStatus.BAD;
             JSONObject brokenDisk = new JSONObject();
             brokenDisk.put("disk", entry.getKey().getMountPath());
             brokenDisk.put("reason", mountedDiskManager.getDiskHealthStatus());
@@ -483,6 +481,9 @@ public class AmbryServerRequests extends AmbryRequests {
         }
       }
     }
+    //if there are any issues on the partition or disk state then considered BAD
+    serverHealthStatus =
+        unstablePartitions.length() > 0 || brokenDisks.length() > 0 ? ServerHealthStatus.BAD : serverHealthStatus;
 
     //Initializing the content of the response as a json
     JSONObject contentJSON = new JSONObject();
--- a/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
@@ -18,6 +18,7 @@ import com.codahale.metrics.MetricRegistry;
 import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.ClusterParticipant;
 import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.clustermap.DiskId;
 import com.github.ambry.clustermap.HardwareState;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.PartitionState;
@@ -25,6 +26,7 @@ import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.clustermap.ReplicaState;
 import com.github.ambry.clustermap.ReplicaStatusDelegate;
 import com.github.ambry.commons.ServerMetrics;
+import com.github.ambry.config.DiskManagerConfig;
 import com.github.ambry.config.ServerConfig;
 import com.github.ambry.network.NetworkRequest;
 import com.github.ambry.network.RequestResponseChannel;
@@ -44,6 +46,8 @@ import com.github.ambry.replication.FindTokenHelper;
 import com.github.ambry.replication.ReplicationAPI;
 import com.github.ambry.replication.ReplicationManager;
 import com.github.ambry.store.BlobStore;
+import com.github.ambry.store.DiskHealthStatus;
+import com.github.ambry.store.DiskManager;
 import com.github.ambry.store.StorageManager;
 import com.github.ambry.store.Store;
 import com.github.ambry.store.StoreException;
@@ -56,9 +60,13 @@ import java.nio.charset.StandardCharsets;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import org.json.JSONArray;
+import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -78,6 +86,7 @@ public class AmbryServerRequests extends AmbryRequests {
           RequestOrResponseType.UndeleteRequest);
   private static final Logger logger = LoggerFactory.getLogger(AmbryServerRequests.class);
   private final ServerConfig serverConfig;
+  private final DiskManagerConfig diskManagerConfig;
   private final StatsManager statsManager;
   private final ClusterParticipant clusterParticipant;
   private final ConcurrentHashMap<RequestOrResponseType, Set<PartitionId>> requestsDisableInfo =
@@ -86,10 +95,11 @@ public class AmbryServerRequests extends AmbryRequests {
   AmbryServerRequests(StoreManager storeManager, RequestResponseChannel requestResponseChannel, ClusterMap clusterMap,
       DataNodeId nodeId, MetricRegistry registry, ServerMetrics serverMetrics, FindTokenHelper findTokenHelper,
       NotificationSystem operationNotification, ReplicationAPI replicationEngine, StoreKeyFactory storeKeyFactory,
-      ServerConfig serverConfig, StoreKeyConverterFactory storeKeyConverterFactory, StatsManager statsManager,
-      ClusterParticipant clusterParticipant) {
+      ServerConfig serverConfig, DiskManagerConfig diskManagerConfig,StoreKeyConverterFactory storeKeyConverterFactory,
+      StatsManager statsManager, ClusterParticipant clusterParticipant) {
     super(storeManager, requestResponseChannel, clusterMap, nodeId, registry, serverMetrics, findTokenHelper,
         operationNotification, replicationEngine, storeKeyFactory, storeKeyConverterFactory);
+    this.diskManagerConfig = diskManagerConfig;
     this.serverConfig = serverConfig;
     this.statsManager = statsManager;
     this.clusterParticipant = clusterParticipant;
@@ -419,18 +429,19 @@ public class AmbryServerRequests extends AmbryRequests {
    */
   private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, AdminRequest adminRequest)
       throws StoreException, IOException {
-    boolean hostHealthy = false; //used to determine if the host is ever unhealthy
+    ServerHealthStatus serverHealthStatus = ServerHealthStatus.BAD; //used to determine if the server is ever unhealthy
+    JSONArray brokenDisks = new JSONArray();
 
     if (this.storeManager instanceof StorageManager) {
       StorageManager storageManager = (StorageManager) this.storeManager;
-      hostHealthy = true;
+      serverHealthStatus = ServerHealthStatus.GOOD;
 
       //Finds all partitions on this host
       List<PartitionId> partitionsInThisHost =Collections.list(storageManager.getPartitionToDiskManager().keys());
 
       /*
-       * Checks if the Host's BlobStore exists,started and is Leader/Standby ReplicaState for all Partitions
-       * this Host is apart of
+       Checks if the Host's BlobStore exists,started and is Leader/Standby ReplicaState for all Partitions
+       this Host is apart of
        */
       for (PartitionId partitionId : partitionsInThisHost) {
         BlobStore hostBlobStore = (BlobStore) storageManager.getStore(partitionId);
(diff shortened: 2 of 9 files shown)


ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -186,6 +200,10 @@ void start() throws InterruptedException {
       compactionManager.enable();
       reportUnrecognizedDirs();
       running = true;
+      if(diskHealthCheck.enabled) {
```
justinlin-linkedin, line 203: please expose an isEnabled method from diskHealthCheck object.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -439,14 +447,39 @@ private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, Ad
...
+      /*
+       Checks if every disk that's assigned to a partition is also mounted and disk is healthy
+       */
+      HashSet<DiskManager> partitionsDiskManagers = new HashSet<>(storageManager.getPartitionToDiskManager().values());
+      for(Map.Entry<DiskId,DiskManager> entry : storageManager.getDiskToDiskManager().entrySet()){
+        DiskManager mountedDiskManager = entry.getValue();
+
+        //Checks Disk healthchecking is enabled and
+        // if a disk isn't healthy then store that disk to return in the response
+        if(mountedDiskManager.diskHealthCheck != null && mountedDiskManager.diskHealthCheck.enabled &&
```
justinlin-linkedin, line 464: let's not do this, let's just do diskManagerConfig.enabled ...

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+  }
+
+  /**
+    Manages the disk healthchecking tasks such as creating,writing, reading, and deleting a file
+   */
+  public class DiskHealthCheck{
+    private int intervalSeconds;             //time between checking the state of the disk
+    public int operationTimeoutSeconds;      //how long a read/write operation is allotted
+    public boolean enabled;                  //enabled healthchecking
+    private String mountPath;                //filepath of where the file will be created
```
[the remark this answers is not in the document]
maskndafi-linkedin: I had made those two variables public to be manipulated in testing. Should I just make them private with getter/setter but I wouldn't be able to make them final?

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: There is no diskManagerConfig in AmbryServerRequests. Should I just rely on mountedDiskManager.diskHealthCheck.isEnabled?

ambry-server/src/main/java/com/github/ambry/server/ServerHealthStatus.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2021 LinkedIn Corp. All rights reserved.
```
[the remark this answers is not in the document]
maskndafi-linkedin: Thanks, couldn't tell when I copied them. Fixed

ambry-store/src/main/java/com/github/ambry/store/DiskHealthStatus.java
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright 2021 LinkedIn Corp. All rights reserved.
```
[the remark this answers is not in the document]
maskndafi-linkedin: Fixed

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
   }
+
+  public DiskHealthStatus getDiskHealthStatus(){
+    return diskHealthStatus;
+  }
+
+  /**
+    Manages the disk healthchecking tasks such as creating,writing, reading, and deleting a file
+   */
+  public class DiskHealthCheck{
```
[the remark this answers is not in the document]
maskndafi-linkedin: Made it private

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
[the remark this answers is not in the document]
maskndafi-linkedin: Made them private and final

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+                StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE);
+        fileChannel.force(true); //writes to disk immediately
+      }catch(IOException e){
+        diskHealthStatus = DiskHealthStatus.CREATE_IO_EXCEPTION;
+        return;
+      }
+
+      //Finds today's date and stores it
+      String todayDate = new SimpleDateFormat("yyyy-mm-dd hh:mm:ss").format(Calendar.getInstance().getTime());
+      ByteBuffer buffer = ByteBuffer.allocate(19);
```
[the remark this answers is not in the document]
maskndafi-linkedin: Yeah, you're right. I changed it. Thanks

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+      //Write to file in operationTimeoutSeconds seconds
+      Future<Integer> writeOperation = fileChannel.write(buffer, 0);
+      try{
+        writeOperation.get(operationTimeoutSeconds,TimeUnit.SECONDS);
+      }catch(Exception e){
+        writeOperation.cancel(true);
+        diskHealthStatus = DiskHealthStatus.WRITE_TIMEOUT;
+        deleteFile(fileChannel);
+        return;
+      }
```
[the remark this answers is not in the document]
maskndafi-linkedin: That's a good point, I made those changes to both write/read.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+        readOperation.cancel(true);
+        diskHealthStatus = DiskHealthStatus.READ_TIMEOUT;
+        deleteFile(fileChannel);
+        return;
+      }
+
+      //Checks if the content matches what was written
+      String fileContent = new String(buffer.array()).trim();
+      if(!fileContent.equals(todayDate)){
+        diskHealthStatus = DiskHealthStatus.READ_DIFFERENT;
```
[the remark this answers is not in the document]
maskndafi-linkedin: I changed it to include close/delete portion

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Created and utilized the isEnabled function

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: Distributed diskmanagerconfig so we no longer rely on mountedDiskManager.diskHealthCheck.isEnabled

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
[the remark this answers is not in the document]
maskndafi-linkedin: Yeah, that could be confusing. I changed it. Thanks

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -419,18 +429,19 @@ private AdminResponse handleBlobStoreControlRequest(DataInputStream requestStrea
    */
   private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, AdminRequest adminRequest)
       throws StoreException, IOException {
-    boolean hostHealthy = false; //used to determine if the host is ever unhealthy
+    ServerHealthStatus serverHealthStatus = ServerHealthStatus.BAD; //used to determine if the server is ever unhealthy
```
snalli, line 432: this can be initiailiazed to GOOD and L437 can be removed.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -439,14 +450,39 @@ private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, Ad
         if (hostBlobStore == null || !hostBlobStore.isStarted() ||
             ((hostBlobStore.getCurrentState() != ReplicaState.STANDBY) &&
                 (hostBlobStore.getCurrentState() != ReplicaState.LEADER))) {
-          hostHealthy = false;
+          serverHealthStatus = ServerHealthStatus.BAD;
           break;
```
snalli, line 454: instead of breaking on first bad partition, how about iterate over all partitions and get an entire list of all bad partitions ?

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -942,50 +950,144 @@ public void crossColoMetricsUpdateTest() throws Exception {
...
+
+    AdminRequest adminRequest = new AdminRequest(AdminRequestOrResponseType.HealthCheck, partitionId, correlationId,
+        clientId);
+
+    AdminResponseWithContent response =
+        (AdminResponseWithContent) sendRequestGetResponse(adminRequest, ServerErrorCode.No_Error);
+    String content = new String(response.getContent(),StandardCharsets.UTF_8);
+
+    //Ensures the response is AdminResponse, json and as expected
+    assertTrue("Response not of type AdminResponse", response instanceof AdminResponse);
```
snalli, line 988: Response must be of type AdminResponse

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -942,50 +950,144 @@ public void crossColoMetricsUpdateTest() throws Exception {
...
+    assertTrue("Response is not a json object",isJSONValid(content));
```
snalli, line 989: Response must be a json object

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -942,50 +950,144 @@ public void crossColoMetricsUpdateTest() throws Exception {
...
+    JSONObject contentJSON  = new JSONObject(content);
+    assertEquals(description,expectedResponse.toString(), contentJSON.toString());
```
snalli, line 991: space after comma

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+
+      //Creates directory for the file if not already there
+      File dir = new File(mountPath);
+      if (!dir.exists()) dir.mkdirs();
+
+      //Checks if there is enough space for the file's content
+      if( new File(mountPath).getFreeSpace() < 19){
+        diskHealthStatus = DiskHealthStatus.INSUFFICIENT_SPACE;
+        return;
+      }
```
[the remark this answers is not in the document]
maskndafi-linkedin: Great, good to know. Made the change

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+    }
+
+    /**
+     * Performing Disk HealthCheck Test in 5 steps
+     * 1. Checks if there is sufficient space on disk
+     * 2. Creates the file on the filepath
+     * 3. Writes today's date to disk
+     * 4. Read and compares what was written to disk
+     * 5. Deletes the file
+     * @return True/False depending on the close/delete was successful
```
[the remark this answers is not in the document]
maskndafi-linkedin: Fixed

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+
+  public DiskHealthStatus getDiskHealthStatus(){
+    return diskHealthStatus;
+  }
+
+  /**
+    Manages the disk healthchecking tasks such as creating,writing, reading, and deleting a file
+   */
+  public class DiskHealthCheck{
+    private int intervalSeconds;             //time between checking the state of the disk
```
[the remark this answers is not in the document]
maskndafi-linkedin: Excluded it from the class. Thanks

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+      //Checks if there is enough space for the file's content
+      if( new File(mountPath).getFreeSpace() < 19){
+        diskHealthStatus = DiskHealthStatus.INSUFFICIENT_SPACE;
+        return;
+      }
+
+      //Creates and opens the file
+      AsynchronousFileChannel fileChannel;
+      try {
+        Path path = Paths.get(mountPath + "temp");
```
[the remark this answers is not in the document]
maskndafi-linkedin: Good to know. Changed

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+        path.toFile().createNewFile();
```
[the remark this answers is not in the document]
maskndafi-linkedin: Fixed

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +588,121 @@ private void reportUnrecognizedDirs() {
...
+        fileChannel =
+            AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.READ,
+                StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE);
+        fileChannel.force(true); //writes to disk immediately
+      }catch(IOException e){
+        diskHealthStatus = DiskHealthStatus.CREATE_IO_EXCEPTION;
```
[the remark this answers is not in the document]
maskndafi-linkedin: I now logged all exception in the disk healtcheck portion

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -942,50 +950,144 @@ public void crossColoMetricsUpdateTest() throws Exception {
...
    */
   @Test
-  public void healthCheckTest() throws InterruptedException, IOException {
+  public void healthCheckTest()
+      throws InterruptedException, StoreException, IOException {
 
     //retrieves writeable partitions to test with
     List<? extends PartitionId> partitionIds = clusterMap.getWritablePartitionIds(DEFAULT_PARTITION_CLASS);
+    JSONObject expectedJSONReponse = new JSONObject();
+    expectedJSONReponse.put("brokenDisks",new JSONArray());
```
snalli, line 1032: space after comma

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+     * 2. Creates the file on the filepath
+     * 3. Writes today's date to disk
+     * 4. Read and compares what was written to disk
+     * 5. Deletes the file
+     */
+    public void diskHealthTest(){
+      if(!enabled){ return;} //ensures disk checks are enabled
+
+      //Creates directory for the file if not already there
+      File dir = new File(heathcheckDir);
```
snalli, line 652: please move all disk operations into the try block. This mkdir could also fail if the disk or filesystem is bad

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+      //Checks if the content matches what was written
+      String fileContent = new String(buffer.array()).trim();
+      if(!fileContent.equals(todayDate)){
+        diskHealthStatus = DiskHealthStatus.READ_DIFFERENT;
+        deleteFile(fileChannel);
+        return;
+      }
+
+      //Deletes this test file
+      if(deleteFile(fileChannel)){
```
snalli, line 722: This delete also should be in a try-catch since any disk related operation could fail at any time

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+                StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE);
+        fileChannel.force(true); //writes to disk immediately
+      }catch(IOException e){
+        diskHealthStatus = DiskHealthStatus.CREATE_IO_EXCEPTION;
+        logger.error("Exception occurred when creating a file for disk healthcheck", e);
+        return;
+      }
+
+      //Finds today's date and stores it
+      String todayDate = new SimpleDateFormat("yyyy-mm-dd hh:mm:ss").format(Calendar.getInstance().getTime());
```
snalli, line 670: do we need write currdate or can we just write a constant string ?

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+      ByteBuffer buffer = ByteBuffer.allocate(todayDate.getBytes().length);
+      buffer.put(todayDate.getBytes());
+      buffer.flip();
+
+      //Write to file in operationTimeoutSeconds seconds
+      Future<Integer> writeOperation = fileChannel.write(buffer, 0);
+      try{
+        writeOperation.get(operationTimeoutSeconds,TimeUnit.SECONDS);
+      }catch(TimeoutException e){
```
snalli, line 679: the formatting is off. please add some spaces after } and )

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+    private final int operationTimeoutSeconds;      //how long a read/write operation is allotted
+    private final boolean enabled;                  //enabled healthchecking
+    private final String heathcheckDir;             //filepath of where the file will be created
+
+
+    DiskHealthCheck(int operationTimeoutSeconds, boolean enabled, String heathcheckDir){
+      this.enabled = enabled;
+      this.heathcheckDir = heathcheckDir;
+      this.operationTimeoutSeconds = operationTimeoutSeconds;
+
```
snalli, line 611: rm newline

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -73,7 +82,10 @@ public class DiskManager {
   private final MessageStoreHardDelete hardDelete;
   private final List<String> unexpectedDirs = new ArrayList<>();
   private final AccountService accountService;
+  private final DiskManagerConfig diskManagerConfig;
   private boolean running = false;
+  private DiskHealthStatus diskHealthStatus = DiskHealthStatus.UNKNOWN_HEALTH;
```
snalli, line 87: why init to unknown ? init to GOOD. The healthcheck will set it to BAD

ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
```diff
@@ -32,11 +32,35 @@ public class DiskManagerConfig {
   @Default("false")
   public final boolean diskManagerEnableSegmentPooling;
 
+  @Config("disk.manager.disk.healthcheck.enabled")
+  @Default("false")
+  public final boolean diskManagerDiskHealthCheckEnabled;
+
+  @Config("disk.manager.disk.healthcheck.interval.seconds")
+  @Default("60")
```
snalli, line 40: is 60 good enuf ?

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
maskndafi-linkedin: Thanks, I'll remove it

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
maskndafi-linkedin: Thanks, I'll remove it

ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
maskndafi-linkedin: Yeah it was discussed with the team that it would be a good frequency

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Because if the ambry healthcheck endpoint was accessed and the disk healthcheck job hasn't finished running the first time then it would appear as though all the disks are healthy when they aren't. We need confirmation that the disk healthcheck passed before we can say the disk is healthy and return a GOOD response to ambry healthcheck

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: You have a great point. The current date acts as a timestamp for when the issued occurred if there ever was one.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Yeah so the deleteFile function is in a try catch

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: If the storageManager isn't a StorageManager then the healthstatus will return GOOD. So by checkng what storageManager is first, we know if we can check the replica/disk states otherwise we have no information about the server

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Will do. Thanks, I didn't notice

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: You have a point. It's hard to distinguish if the issue is from the replica bootstrapping or from the disk's hardware if we don't have both pieces of information. Let me ask the media-sre team just to confirm

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -439,14 +450,39 @@ private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, Ad
...
+            (!partitionsDiskManagers.contains(mountedDiskManager) ||
+            mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY)){
+
+          serverHealthStatus = ServerHealthStatus.BAD;
+          JSONObject brokenDisk = new JSONObject();
+          brokenDisk.put("disk",entry.getKey().getMountPath());
+          brokenDisk.put("reason",mountedDiskManager.getDiskHealthStatus());
+          brokenDisks.put(brokenDisk);
+        }
+      }
```
justinlin-linkedin, line 477: we can put `diskManagerConfig.diskManagerDiskHealthCheckEnabled` outside of this for loop.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+      if (!dir.exists()) dir.mkdirs();
+
+      //Creates and opens the file
+      AsynchronousFileChannel fileChannel;
+      try {
+        Path path = Paths.get(heathcheckDir,"temp");
+        fileChannel =
+            AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.READ,
+                StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE);
+        fileChannel.force(true); //writes to disk immediately
```
justinlin-linkedin, line 662: I don't think there is anything to force at this moment, we can remove this line.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+        logger.error("Timeout Exception occurred when writing to a file for disk healthcheck", e);
+        deleteFile(fileChannel);
+        return;
+      }catch(Exception e){
+        writeOperation.cancel(true);
+        diskHealthStatus = DiskHealthStatus.WRITE_EXCEPTION;
+        logger.error("Exception occurred when writing to a file for disk healthcheck", e);
+        deleteFile(fileChannel);
+        return;
+      }
```
justinlin-linkedin, line 691: it seems like we can reuse some of the exception handling code,
```
      Future<Integer> writeOperation = fileChannel.write(buffer, 0);
      Exception ex = null;
      try{
        writeOperation.get(operationTimeoutSeconds,TimeUnit.SECONDS);
      }catch(TimeoutException e){
        ex = e;
        diskHealthStatus = DiskHealthStatus.WRITE_TIMEOUT;
        return;
      }catch(Exception e){
        ex = e;
        diskHealthStatus = DiskHealthStatus.WRITE_EXCEPTION;
        return;
      }
      if (ex != null) {
        writeOperation.cancel(true);
        logger.error("Exception occurred when writing to a file for disk healthcheck", e);
        deleteFile(fileChannel);
        return;
      }
```

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,138 @@ private void reportUnrecognizedDirs() {
...
+     * Attempts to close the filechannel and delete the file
+     * @param fileChannel the file used for disk healthcheck
+     * @return True/False depending on the close/delete was successful
+     */
+    private boolean deleteFile(AsynchronousFileChannel fileChannel){
+      if(fileChannel.isOpen()){
+        try{
+          fileChannel.close();
+        }catch(IOException e){
+          diskHealthStatus = diskHealthStatus.DELETE_IO_EXCEPTION;
```
justinlin-linkedin, line 632: should be 
```
if (diskHealthStatus == HEALTHY) {
    diskHealthStatus = diskHealthStatus.DELETE_IO_EXCEPTION;
}
```

justinlin-linkedin (commented):
Please also format your code.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: This is meant for the future write/read calls to be written to the storage device immediately as to avoid staying in the filechannel's buffer. At least that's what I understood from 
'Forces any updates to this channel's file to be written to the storage device that contains it.'

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: I changed the if-statement to account for any issue that's happened thus far during the healthcheck so as to not overlap with that diskHealthStatus
`if (successfulThusFar) {
    diskHealthStatus = diskHealthStatus.DELETE_IO_EXCEPTION;
}`

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Great idea. I combined read/write logic since they only differed by a line and refactored that section to make it easier to read

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: Thanks. Changed it

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Removed

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
maskndafi-linkedin: Formatted that file. Thanks

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
maskndafi-linkedin: Changed. Thanks

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: I changed the logic to store the non-standby/leader partitions as well. Thanks for that

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -419,34 +429,68 @@ private AdminResponse handleBlobStoreControlRequest(DataInputStream requestStrea
...
       throws StoreException, IOException {
-    boolean hostHealthy = false; //used to determine if the host is ever unhealthy
+    ServerHealthStatus serverHealthStatus = ServerHealthStatus.BAD; //used to determine if the server is ever unhealthy
+    JSONArray brokenDisks = new JSONArray();        // indicate which disks didn't pass the disk healthcheck
+    JSONArray unstablePartitions = new JSONArray(); // indicate which partitions aren't in leader/standby state
 
     if (this.storeManager instanceof StorageManager) {
       StorageManager storageManager = (StorageManager) this.storeManager;
-      hostHealthy = true;
+      serverHealthStatus = ServerHealthStatus.GOOD;
```
snalli, line 438: Can this.storeManager be an instance of something else ?
Also, I find the assignment to this variable a bit redundant BAD -> GOOD -> BAD.
One way around this is to eliminate L438
and add a snippet at the end of this loop

if (unstablePartitions.length()) then status = BAD.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -439,14 +450,39 @@ private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, Ad
...
+      for(Map.Entry<DiskId,DiskManager> entry : storageManager.getDiskToDiskManager().entrySet()){
+        DiskManager mountedDiskManager = entry.getValue();
+
+        //Checks Disk healthchecking is enabled and
+        // if a disk isn't healthy then store that disk to return in the response
+        if(diskManagerConfig.diskManagerDiskHealthCheckEnabled &&
+            (!partitionsDiskManagers.contains(mountedDiskManager) ||
+            mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY)){
+
+          serverHealthStatus = ServerHealthStatus.BAD;
```
snalli: instead of assigning this variable over and over again for each bad diskMgr, we could add a snippet at the end of the loop to set it to BAD if the len of brokenDisk >= 1

ambry-server/src/main/java/com/github/ambry/server/ServerHealthStatus.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package com.github.ambry.server;
+
+/**
+ * Server Health Status Levels
+ * Each level determines the health of the server. This can be extended to include bootstrapping
+ */
+public enum ServerHealthStatus {
+
+  /** The server is healthy based on the replica state's and disk state*/
```
snalli, line 22: space before */

ambry-server/src/main/java/com/github/ambry/server/ServerHealthStatus.java
```diff
@@ -0,0 +1,28 @@
...
+  GOOD,
+
+  /** There was at least one issue with either the replica state or disk state on this server*/
```
snalli, line 25: space before */

snalli: And i believe for single line comments it is // and not /**

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
snalli: What you are mentioning is the DIRECT_IO mode. The open call should have a flag to open the file in that mode. Is there a flag ?

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -419,34 +429,68 @@ private AdminResponse handleBlobStoreControlRequest(DataInputStream requestStrea
    */
   private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, AdminRequest adminRequest)
```
snalli, line 430: Who sends a healthCheckRequest to ambry ?

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: Great suggestion. I added it to the end so it now looks like 
`serverHealthStatus =
        unstablePartitions.length() > 0 || brokenDisks.length() > 0 ? ServerHealthStatus.BAD : serverHealthStatus;`

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: storeManager can be CloudStorageManager which doesn't have the resources we need
But as for the the redundant logic. I see your point and just added a ternary operation now. If we can't perform a disk healthcheck then we wouldn't know the state of the disk so it's considered bad

ambry-server/src/main/java/com/github/ambry/server/ServerHealthStatus.java
maskndafi-linkedin: Enums have multi-line comment structure as what I've seen around the repo. Thanks I changed it to make each comment multi-line now

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: ambry-healthchecker will be the one sending requests to know about the server's health

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: I have now. Appreciate it

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Yeah this can be done automatically with the flag like @snalli suggested, I included the `DSYNC` flag which will synchronize with disk faster than `SYNC` as `DSYNC` doesn't need the metadata to write the file's updates to disk since we don't check for the metadata in the healthcheck.

snalli: This filesystem in this case will sync metadata since you are creating file and allocating block. Just syncing data would lead to crash inconsistency. The FS would sync data only when you are overwriting file pages without extending or shrinking the file. Just use SYNC in your code. It would only be a few millisecs slower than DSYNC, which is tolerable given the frequency of health check is in the order of secs.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,165 @@ private void reportUnrecognizedDirs() {
...
+    private AsynchronousFileChannel createFileChannel() {
+      try {
+        File dir = new File(healthcheckDir);
+        if (!dir.exists() && !dir.mkdir()) {
+          throw new Error(String.format("%s directory wasn't created", healthcheckDir));
+        }
+        Path path = Paths.get(healthcheckDir, "temp");
+        AsynchronousFileChannel fileChannel =
+            AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.READ,
+                StandardOpenOption.DSYNC, StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE);
```
justinlin-linkedin: let's not use DSYNC for now, it might impact the disk performance.
At least, let's use a configuration here so we can easily turn this on and off.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,165 @@ private void reportUnrecognizedDirs() {
...
+        if (!successful) {
+          diskHealthStatus = DiskHealthStatus.READ_DIFFERENT;
+        }
+      }
+
+      //Deletes this test file and only specifies healthy if everything successful
+      if (deleteFile(fileChannel, successful) && successful) {
+        diskHealthStatus = DiskHealthStatus.HEALTHY;
+      }
+      ;
```
justinlin-linkedin, line 751: empty statement here.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -570,4 +590,165 @@ private void reportUnrecognizedDirs() {
...
+      try {
+        if (isWriting) {
+          result = fileChannel.write(buffer, 0);
+        } else {
+          result = fileChannel.read(buffer, 0);
+        }
+        result.get(operationTimeoutSeconds, TimeUnit.SECONDS);
+        return true;
+      } catch (TimeoutException e) {
+        diskHealthStatus = timeoutException;
```
justinlin-linkedin, line 663: nit: rename this to timeoutStatus and otherStatus.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
snalli: If disk perf is a concern, then just buffer the writes in memory and flush just before closing the file. 
From what I understand, force(true) is the same as DSYNC ?

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Changed. Thanks

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
snalli: The amount of data written is really small. It won't affect disk perf. Either we buffer the data/metadata and flush it on close or do a direct_io. Without either of these, the test is incorrect.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -419,34 +429,69 @@ private AdminResponse handleBlobStoreControlRequest(DataInputStream requestStrea
...
-                (hostBlobStore.getCurrentState() != ReplicaState.LEADER))) {
-          hostHealthy = false;
-          break;
+        if (hostBlobStore == null || !hostBlobStore.isStarted() || (
+            (hostBlobStore.getCurrentState() != ReplicaState.STANDBY) && (hostBlobStore.getCurrentState()
+                != ReplicaState.LEADER))) {
+
+          JSONObject unstablePartition = new JSONObject();
+          unstablePartition.put("partitionId", partitionId);
+          unstablePartition.put("reason", hostBlobStore.getCurrentState());
```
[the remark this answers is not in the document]
maskndafi-linkedin: Yeah, you're right

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: Included the SYNC flag now

justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java
snalli: I think 60 is too aggressive. Can you make it 180 ? We do not want the health-check interfering with regular traffic. If there are no issues in the first deployment for two weeks, we can reduce the interval.

maskndafi-linkedin: The media-sre team thought a minute would be good to avoid stale information especially to know if the disks are mounted or not. They were thinking at most 120 in the last meeting with the stakeholders, do you think that wouldn’t be too aggressive or just ramp up from 180 in the coming weeks like you said?

maskndafi-linkedin: I spoke with my team and scheduling the disk health to occur every 3 minutes for now is fine. As long as we scale down to 1 minute if the deployment w/ 3 minutes doesn't causes issues.

snalli (approved):

PR #2172, opened 2022-08-05
[WIP]Support get chunkId for composite blob

SophieGuo410:


ambry-tools/src/main/java/com/github/ambry/tools/admin/GetChunkIdTool.java
```diff
@@ -0,0 +1,243 @@
+/**
+ * Copyright 2020 LinkedIn Corp. All rights reserved.
```
ankagrawal, line 2: 2022

cgtz:
I would suggest making this into an "admin" subresource similar to the "Replicas" frontend api instead of a CLI tool. I remember that the CLI tool was a bit hard to get running in an urgent situation, since it requires supplying a lot of configuration.

SophieGuo410:
> 

yeah, I'll work on it. This is a temporarily work around for this use case.

snalli:
does this need any review or can this be closed ?

SophieGuo410:
> does this need any review or can this be closed ?

it's a draft pr and it's still work in progress.  no need to review for now.

PR #2173, opened 2022-08-08
Fixed the DR process "partition is null" exception. 

JingQianCloud:
Blob ID in the composition Blob may not be in the static clustermap.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
@@ -36,6 +36,7 @@ public class RecoveryTestClusterManager implements ClusterMap {
   final Map<AmbryDisk, Disk> ambryDiskToDiskMap;
   final Map<AmbryDataNode, DataNode> ambryDataNodeToDataNodeMap;
 
+  private static final Logger logger = LoggerFactory.getLogger(Replica.class);
   /**
    * Construct a RecoveryTestClusterManager instance.
    * @param staticClusterManager the {@link StaticClusterManager} object.
@@ -52,7 +53,19 @@ public class RecoveryTestClusterManager implements ClusterMap {
   public PartitionId getPartitionIdFromStream(InputStream stream) throws IOException {
     DuplicatingInputStream duplicatingInputStream = new DuplicatingInputStream(stream);
     duplicatingInputStream.mark(0);
-    PartitionId partitionIdStatic = staticClusterManager.getPartitionIdFromStream(duplicatingInputStream);
+    PartitionId partitionIdStatic = null;
+    try {
+      partitionIdStatic = staticClusterManager.getPartitionIdFromStream(duplicatingInputStream);
+    } catch (IOException e) {
+      // for composite blob, it has the list of Blob IDs. It may not exist in the static cluster map.
+      // swallow the exception and try with helixClusterManager
+    }
+
+    if (partitionIdStatic == null) {
+      duplicatingInputStream.reset();
+      partitionIdStatic = helixClusterManager.getPartitionIdFromStream(duplicatingInputStream);
+      logger.trace("DR: partition doesn't exist in staticClusterManager, try helixClusterManager {}", partitionIdStatic);
+    }
     return partitionIdStatic;
   }
 


ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
```diff
@@ -52,7 +53,19 @@ public class RecoveryTestClusterManager implements ClusterMap {
...
+    } catch (IOException e) {
+      // for composite blob, it has the list of Blob IDs. It may not exist in the static cluster map.
+      // swallow the exception and try with helixClusterManager
+    }
+
+    if (partitionIdStatic == null) {
+      duplicatingInputStream.reset();
+      partitionIdStatic = helixClusterManager.getPartitionIdFromStream(duplicatingInputStream);
+      logger.trace("DR: partition doesn't exist in staticClusterManager, try helixClusterManager {}", partitionIdStatic);
+    }
```
justinlin-linkedin, line 68: this looks good, can we add a test for this?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
```diff
@@ -52,7 +53,19 @@ public class RecoveryTestClusterManager implements ClusterMap {
   public PartitionId getPartitionIdFromStream(InputStream stream) throws IOException {
     DuplicatingInputStream duplicatingInputStream = new DuplicatingInputStream(stream);
     duplicatingInputStream.mark(0);
-    PartitionId partitionIdStatic = staticClusterManager.getPartitionIdFromStream(duplicatingInputStream);
+    PartitionId partitionIdStatic = null;
+    try {
+      partitionIdStatic = staticClusterManager.getPartitionIdFromStream(duplicatingInputStream);
+    } catch (IOException e) {
+      // for composite blob, it has the list of Blob IDs. It may not exist in the static cluster map.
```
snalli, line 60: Please log an exception here or move the trace msg that says trying here before actually trying the helixMap

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
```diff
@@ -52,7 +53,19 @@ public class RecoveryTestClusterManager implements ClusterMap {
...
+      // swallow the exception and try with helixClusterManager
+    }
+
+    if (partitionIdStatic == null) {
+      duplicatingInputStream.reset();
+      partitionIdStatic = helixClusterManager.getPartitionIdFromStream(duplicatingInputStream);
```
snalli, line 66: This code could be moved in the catch block of the try-catch above. 
And are we sure that the helixClusterManager will always have the partitiionID. Even if it does, can it throw some excpetion that must be caught ?

JingQianCloud: If we remove the code to the catch, it has the dependency on the implementation of the staticClusterManager.getPartitionIdFromStream.
If the staticClusterManager.getPartitionIdFromStream decides to "don't throw exception" but just return "null", we'll miss the case.

Check "if ==null" is more generic. 

If helixClusterManager returns null or exception, caller will fail somewhere.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
JingQianCloud: I think log when partitionIdStatis is null is more generic. It can cover two cases,
1. staticClusterManager.getPartitionIdFromStream throws exception.
2. staticClusterManager.getPartitionIdFromStream returns null for some reason.

JingQianCloud: But you are right, without the log here, we may not know exception ever happened. I'll add the extra log here.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
snalli: cool thanks

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
snalli: cool thanks

snalli (approved):
just add the log msg in the catch path and i am good with this PR. LGTM.

justinlin-linkedin (approved):

PR #2174, opened 2022-08-10
Removing Filepath from Disk Manager Config

maskndafi-linkedin:
This PR aims to remove file path from the disk healthchecker as getMountPath() already contains the path needed for disk healthchecker


justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -203,6 +202,7 @@ void start() throws InterruptedException {
       reportUnrecognizedDirs();
       running = true;
       if (diskHealthCheck.isEnabled()) {
+        logger.trace("Starting Disk Healthchecker");
```
justinlin-linkedin, line 205: if we are only doing this for once, we can use info log
I think we can do a trace log on each test?

maskndafi-linkedin: Changed to include info log and trace log in each test

justinlin-linkedin (approved):

PR #2176, opened 2022-08-11
Correct mount checks and diskManager health checks

maskndafi-linkedin:
This PR corrects how disk healthchecks checks a mount by initializing a disk to mount_not_accessible and utilizing checkMountAccessible() at each disk healthceck test. Also, this PR fixes how it only chooses disks that belong to a partitionId as part of the healthcheck response.

### DIFF
--- a/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
@@ -469,9 +469,9 @@ public class AmbryServerRequests extends AmbryRequests {
           DiskManager mountedDiskManager = entry.getValue();
 
           //Checks Disk healthchecking is enabled and
-          // if a disk isn't mounted as expected or isn't healthy then store it as part of the response
-          if (!partitionsDiskManagers.contains(mountedDiskManager)
-              || mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY) {
+          // if a disk manager is assigned to a partition and isn't healthy then store it as part of the response
+          if (partitionsDiskManagers.contains(mountedDiskManager)
+              && mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY) {
 
             JSONObject brokenDisk = new JSONObject();
             brokenDisk.put("disk", entry.getKey().getMountPath());
--- a/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
@@ -683,12 +683,16 @@ public class DiskManager {
             AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.READ,
                 StandardOpenOption.SYNC, StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE);
         return fileChannel;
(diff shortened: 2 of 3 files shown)


justinlin-linkedin (approved):
LGTM

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -469,9 +469,9 @@ private AdminResponse handleHealthCheckRequest(DataInputStream requestStream, Ad
           DiskManager mountedDiskManager = entry.getValue();
 
           //Checks Disk healthchecking is enabled and
-          // if a disk isn't mounted as expected or isn't healthy then store it as part of the response
-          if (!partitionsDiskManagers.contains(mountedDiskManager)
-              || mountedDiskManager.getDiskHealthStatus() != DiskHealthStatus.HEALTHY) {
+          // if a disk manager is assigned to a partition and isn't healthy then store it as part of the response
```
snalli, line 472: what if the partitionsDiskManager does not contain mountedDiskManager ?
The two booleans are not equal. Is that intended ?

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -679,17 +677,15 @@ private boolean performOperation(AsynchronousFileChannel fileChannel, ByteBuffer
...
+        checkMountPathAccessible();
+        Path path = Paths.get(disk.getMountPath(), "temp");
         AsynchronousFileChannel fileChannel =
             AsynchronousFileChannel.open(path, StandardOpenOption.WRITE, StandardOpenOption.READ,
                 StandardOpenOption.SYNC, StandardOpenOption.CREATE, StandardOpenOption.DELETE_ON_CLOSE);
         return fileChannel;
       } catch (Exception e) {
-        diskHealthStatus = DiskHealthStatus.CREATE_EXCEPTION;
+        diskHealthStatus =
+            e instanceof StoreException ? DiskHealthStatus.MOUNT_NOT_ACCESSIBLE : DiskHealthStatus.CREATE_EXCEPTION;
```
snalli, line 688: Instead of just checking if e is an instance of StoreException, check specifically that the message or errcode from L539 matches your expectation.

snalli: And add a new catch blocks to catch StoreException

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
maskndafi-linkedin: Great question. So we only care about disk managers that are assigned to partitions. So, as long as the disk is assigned to a partition then we care about the health. This avoids caring about backup/unused disks. The two booleans shouldn't be the same then

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
maskndafi-linkedin: You have a good point, good practice to differentiate the exceptions and be exact about it. Fixed

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
snalli: ok got it

snalli (approved):

PR #2178, opened 2022-08-13
Fix issues related to Simple Operation Tracker response logic

mikeyang923:
Fix issues related to Simple Operation Tracker response logic:
1. Return 503 Error instead of NOT_FOUND when there is a success in replica
2. Set allReplicas to include all replicas regardless of its state

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -16,11 +16,9 @@ package com.github.ambry.router;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.clustermap.ReplicaState;
-import com.github.ambry.clustermap.ReplicaType;
 import com.github.ambry.commons.BlobId;
 import com.github.ambry.config.RouterConfig;
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
@@ -199,7 +197,7 @@ class SimpleOperationTracker implements OperationTracker {
     totalOfflineReplicaCount =
         getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE)).getOrDefault(ReplicaState.OFFLINE,
             Collections.emptyList()).size();
-    allReplicas = allDcReplicasByState.values().stream().flatMap(Collection::stream).collect(Collectors.toList());
+    allReplicas = partitionId.getReplicaIds().stream().collect(Collectors.toList());
     allReplicaCount = allReplicas.size();
 
     switch (routerOperation) {
@@ -472,6 +470,11 @@ class SimpleOperationTracker implements OperationTracker {
   }
 
   @Override
+  public int getSuccessCount() {
+    return replicaSuccessCount;
+  }
+
+  @Override
   public boolean isDone() {
     return hasSucceeded() || hasFailed();
   }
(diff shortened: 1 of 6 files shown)


justinlin-linkedin (approved):
This looks good to me as a hot fix, can you add some test cases?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -199,7 +197,7 @@ class SimpleOperationTracker implements OperationTracker {
     totalOfflineReplicaCount =
         getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE)).getOrDefault(ReplicaState.OFFLINE,
             Collections.emptyList()).size();
-    allReplicas = allDcReplicasByState.values().stream().flatMap(Collection::stream).collect(Collectors.toList());
+    allReplicas = partitionId.getReplicaIds().stream().collect(Collectors.toList());
```
SophieGuo410, line 200: The test case didn't cover this change. Can you add some test regarding this change(for example, one replica not found, one replica no state, it will do cross colo ttl update? So later when people accidentally change it back or change to something else, it will trigger the tests.

mikeyang923: Done in latest commit.

ambry-router/src/test/java/com/github/ambry/router/OperationTrackerTest.java
```diff
@@ -461,6 +461,49 @@ public void getOperationWithReplicaStateTest() {
...
+    List<String> mountPaths = Collections.singletonList("mockMountPath");
+    // set up one node per data center for testing
+    datanodes = new ArrayList<>(Arrays.asList(new MockDataNodeId(portList, mountPaths, "dc-0"),
+        new MockDataNodeId(portList, mountPaths, "dc-1")));
+    mockPartition = new MockPartitionId();
+    localDcName = datanodes.get(0).getDatacenterName();
+    mockClusterMap = new MockClusterMap(false, datanodes, 1, Collections.singletonList(mockPartition), localDcName);
+    // put two STANDBY replicas in each data center (note that "populateReplicaList" method alternatively distributes
+    // the replica, so here we set 4 for two dc in total)
+    populateReplicaList(4, ReplicaState.STANDBY);
```
SophieGuo410, line 477: Actually the test that I want you to check is for original dc, there's one replica down, one replica in non-state, and one can find the value. If you don't do the change, it will return unavailable. But with your change, it will return succeed since cross colo can do ttl update.

mikeyang923: Ok, updated.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -361,8 +361,14 @@ private void checkAndMaybeComplete() {
         operationException.set(new RouterException("TtlUpdateOperation failed possibly because of offline replicas",
             RouterErrorCode.AmbryUnavailable));
       } else if (operationTracker.hasFailedOnNotFound()) {
-        operationException.set(
-            new RouterException("TtlUpdateOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
+        if (operationTracker.getSuccessCount() > 0) {
+          operationException.set(
```
justinlin-linkedin, line 365: let's add a metric here so we would know this happened.

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
mikeyang923: Done in the latest commit.

SophieGuo410 (approved):

PR #2180, opened 2022-08-15
Fixed VCR Helix participant didn't report http2 port issue.

JingQianCloud:
Fixed the issue "VCR Helix participant didn't report http2 port because VCR_HELIX_CONFIG_READY is true."
Added VcrInstanceConfig, if the current instance config is different from the helix InstanceConfig, update the helix record.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
@@ -14,6 +14,7 @@
 package com.github.ambry.cloud;
 
 import com.github.ambry.account.AccountService;
+import com.github.ambry.cloud.azure.VcrInstanceConfig;
 import com.github.ambry.clustermap.CloudDataNode;
 import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.ClusterMapUtils;
@@ -195,18 +196,15 @@ public class HelixVcrClusterParticipant implements VcrClusterParticipant {
     helixAdmin = manager.getClusterManagmentTool();
 
     InstanceConfig instanceConfig = helixAdmin.getInstanceConfig(vcrClusterName, vcrInstanceName);
-    if (!instanceConfig.getRecord().getBooleanField(VCR_HELIX_CONFIG_READY, false)) {
-      if (cloudConfig.vcrSslPort != null) {
-        instanceConfig.getRecord().setSimpleField(SSL_PORT_STR, Integer.toString(cloudConfig.vcrSslPort));
-      }
-      if (cloudConfig.vcrHttp2Port != null) {
-        instanceConfig.getRecord().setSimpleField(HTTP2_PORT_STR, Integer.toString(cloudConfig.vcrHttp2Port));
-      }
-      // Set HELIX_ENABLED to be true. Listeners take action only when this value is True.
-      instanceConfig.getRecord().setBooleanField(VCR_HELIX_CONFIG_READY, true);
-      logger.info("Set VCR_HELIX_CONFIG_READY to true.");
+    VcrInstanceConfig helixInstanceConfig = VcrInstanceConfig.toVcrInstanceConfig(instanceConfig);
+    VcrInstanceConfig newInstanceConfig = new VcrInstanceConfig(cloudConfig.vcrSslPort, cloudConfig.vcrHttp2Port, true);
+    if (!helixInstanceConfig.equals(newInstanceConfig))  {
+      // Set VCR_HELIX_CONFIG_READY to be true. Listeners take action only when this value is True.
+      newInstanceConfig.updateInstanceConfig(instanceConfig);
+
+      logger.info("Update VCR InstanceConfig and set VCR_HELIX_CONFIG_READY to true.");
     } else {
-      logger.info("VCR_HELIX_CONFIG_READY is true.");
+      logger.info("VCR InstanceConfig is the same and VCR_HELIX_CONFIG_READY is true.");
     }
     helixAdmin.setInstanceConfig(vcrClusterName, vcrInstanceName, instanceConfig);
     logger.info("Participated in HelixVcrCluster successfully.");
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/VcrInstanceConfig.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/VcrInstanceConfig.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ */
+
+package com.github.ambry.cloud.azure;
+
+import org.apache.helix.model.InstanceConfig;
+import java.util.Objects;
+import static com.github.ambry.clustermap.ClusterMapUtils.*;
+import static com.github.ambry.config.CloudConfig.*;
+
+/**
+ * A data object for configs scoped to a single Vcr node.
+ */
+public class VcrInstanceConfig {
+  private final Integer sslPort;
+  private final Integer http2Port;
+  private final boolean configReady;
+
+  /**
+   * @param sslPort the ssl port, or {@code null} if the server does not have one.
+   * @param http2Port the HTTP2 port, or {@code null} if the server does not have one.
+   * @param configReady it's true if participant is ready
+   */
+  public VcrInstanceConfig(Integer sslPort, Integer http2Port, boolean configReady) {
+    this.sslPort = sslPort;
+    this.http2Port = http2Port;
+    this.configReady = configReady;
+  }
+
+  /**
+   * Compare two {@link VcrInstanceConfig}
+   * @param o the other {@link VcrInstanceConfig} to compare with this {@link VcrInstanceConfig}.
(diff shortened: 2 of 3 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
```diff
@@ -59,6 +59,7 @@ public PartitionId getPartitionIdFromStream(InputStream stream) throws IOExcepti
     } catch (IOException e) {
       // for composite blob, it has the list of Blob IDs. It may not exist in the static cluster map.
       // swallow the exception and try with helixClusterManager
+      logger.trace("RecoveryTestClusterManager getPartitionIdFromStream exception", e);
```
snalli, line 62: logger.error ?

snalli: logger.error instead of trace ?

JingQianCloud: Yes, I'm thinking about the level. 
Actually it hits very frequently, basically for all the composite blob, we hit for every item in the meta blob. The log will be overwhelmed by it.
That's why in original PR, if partitionIDStatus == null, I log the message but only use trace.

    if (partitionIdStatic == null) {
      duplicatingInputStream.reset();
      partitionIdStatic = helixClusterManager.getPartitionIdFromStream(duplicatingInputStream);
      logger.trace("DR: partition doesn't exist in staticClusterManager, try helixClusterManager {}", partitionIdStatic);
    }

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -195,7 +195,8 @@ public void participate() throws Exception {
     helixAdmin = manager.getClusterManagmentTool();
 
     InstanceConfig instanceConfig = helixAdmin.getInstanceConfig(vcrClusterName, vcrInstanceName);
-    if (!instanceConfig.getRecord().getBooleanField(VCR_HELIX_CONFIG_READY, false)) {
+    if (!instanceConfig.getRecord().getBooleanField(VCR_HELIX_CONFIG_READY, false) ||
```
justinlin-linkedin, line 198: is there any cases where vcrHelixConfigReady is false? I mean, I don't see HelixVcrClusterParticipant ever update this to false.

JingQianCloud: I didn't see any case either.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
JingQianCloud: Hi Sanketh,
Thanks for the suggestion.
1.  For the IOException block, I'll change accordingly.  
2. The reason that I prefer adding log in partitionIdStatic == null block is no matter staticClusterManager throws exception or returns null, it's the same level of error. Throwing exception in this case is not more severe than return null. Basically we couldn't find the Blob in the static map. 
3. And staticClusterManager.getPartitionIdFromStream can change its implementation to return null only instead of throwing exception. In this case, it's still the same level of error. Checking null or not is more generic and our code logic won't be dependent on the implementation of the static cluster map.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
justinlin-linkedin: i think in this case, we should not use any boolean value to indicate if we should should update the instance config. I think we should define an VcrInstanceConfig class and compare the object we obtain from helix and the object we expect. If they are not the same, we update the helix.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
snalli: Yeah I agree with points 2 and 3. I don't have any comments there. My comment was only for point 1.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
```diff
@@ -59,6 +59,7 @@ public PartitionId getPartitionIdFromStream(InputStream stream) throws IOExcepti
     } catch (IOException e) {
       // for composite blob, it has the list of Blob IDs. It may not exist in the static cluster map.
       // swallow the exception and try with helixClusterManager
+      logger.error("staticClusterManager getPartitionIdFromStream throws {}", e.getMessage());
```
snalli, line 62: You mentioned this would overwhelm the log ? The log.error is good but I suggested putting this inside an if-case to prevent overwhelming the log like this 
```
if (e.getMessage.equals("...") == true or false){ 
    log.error("...")
}
```

snalli:
Lets merge this. Please make the logging change by conditionally logging and not overwhelming the log

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
JingQianCloud: I think this is the only exception message we have so far. So I guess we'll log the same amount of error messages.

JingQianCloud:
@snalli I have another comment to address for this PR. Sorry for the slow response.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/RecoveryTestClusterManager.java
snalli: I see, so this won't overwhelm the log ?

snalli: ok understood

snalli:
> @snalli I have another comment to address for this PR. Sorry for the slow response.

ok lmk when done and i will merge

snalli:
Are there any more comments to be addressed on this ?

JingQianCloud:
Yes, I need address Justin's comment.
Add VcrInstanceConfig data structure. Using its comparison to replace vcrHelixConfigReady.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
JingQianCloud: I've added the VcrInstanceConfig
1. We'd better use VcrInstanceConfig in HelixVcrClusterParticipant and HelixVcrClusterSpectator and other related vcr code. It needs a code refactoring. It's out of the scope of the PR and won't address in this PR.
2. I'd like to still keep the VCR_HELIX_CONFIG_READY. Although in our code, we don't modify it or set it to false anywhere. But it's a flag we can use to change the vcr using some other Operational tooling. So I prefer keeping it.

snalli (approved):

snalli:
approved

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/VcrInstanceConfig.java
```diff
@@ -0,0 +1,66 @@
...
+   * @param sslPort the ssl port, or {@code null} if the server does not have one.
+   * @param http2Port the HTTP2 port, or {@code null} if the server does not have one.
+   */
+  public VcrInstanceConfig(Integer sslPort, Integer http2Port) {
+    this.sslPort = sslPort;
+    this.http2Port = http2Port;
+  }
+
+  @Override
+  public boolean equals(Object o) {
```
snalli, line 39: docstring

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/VcrInstanceConfig.java
```diff
@@ -0,0 +1,66 @@
...
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+    VcrInstanceConfig that = (VcrInstanceConfig) o;
+    return Objects.equals(sslPort, that.sslPort) && Objects.equals(http2Port, that.http2Port);
+  }
+
+  public static VcrInstanceConfig toVcrInstanceConfig(InstanceConfig instanceConfig) {
+      String sslPortStr = instanceConfig.getRecord().getSimpleField(SSL_PORT_STR);
+      Integer sslPort = (sslPortStr == null) ? null : Integer.valueOf(sslPortStr);
```
snalli, line 52: docstring

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/VcrInstanceConfig.java
```diff
@@ -0,0 +1,66 @@
...
+      String httpPortStr = instanceConfig.getRecord().getSimpleField(HTTP2_PORT_STR);
+      Integer http2Port = (httpPortStr == null) ? null : Integer.valueOf(httpPortStr);
+      return new VcrInstanceConfig(sslPort, http2Port);
+  }
+
+  public void updateInstanceConfig(InstanceConfig instanceConfig) {
```
snalli, line 58: docstring

snalli (commented):
docstring

snalli (changes requested):
docstring

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/VcrInstanceConfig.java
JingQianCloud: Added the docstring. Thanks,

snalli (approved):
lgtm

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
JingQianCloud: Filed new revision. Added VCR_HELIX_CONFIG_READY to the VcrInstanceConfig.

justinlin-linkedin (approved):

PR #2184, opened 2022-08-18
[Build] Upgrade ZK client for ACL

snalli:
Upgrade Zookeeper client and add dependencies in preparation for Helix
ACL

### DIFF
--- a/ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
+++ b/ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
@@ -16,7 +16,7 @@ package com.github.ambry.quota.capacityunit;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.github.ambry.quota.QuotaName;
 import java.util.concurrent.atomic.AtomicLong;
-import org.codehaus.jackson.annotate.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 
 
 /**
--- a/ambry-test-utils/src/main/java/com/github/ambry/utils/TestUtils.java
+++ b/ambry-test-utils/src/main/java/com/github/ambry/utils/TestUtils.java
@@ -283,6 +283,8 @@ public class TestUtils {
     public ZkServerWrapper(String dataDir, String logDir, int port) {
       this.dataDir = new File(dataDir);
       this.dataLogDir = new File(logDir);
+      this.dataDir.delete();
+      this.dataLogDir.delete();
       this.dataDir.mkdirs();
       this.dataLogDir.mkdirs();
       this.port = port;
--- a/build.gradle
+++ b/build.gradle
@@ -95,6 +95,8 @@ subprojects {
 
     dependencies {
         compile "org.slf4j:slf4j-api:$slf4jVersion"
+        compile "org.apache.zookeeper:zookeeper:$zookeeperVersion"
+        compile "org.xerial.snappy:snappy-java:$snappyVersion"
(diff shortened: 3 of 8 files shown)


build.gradle
```diff
@@ -95,6 +95,8 @@ subprojects {
 
     dependencies {
         compile "org.slf4j:slf4j-api:$slf4jVersion"
+        compile "org.apache.zookeeper:zookeeper:$zookeeperVersion"
```
Arun-LinkedIn, line 98: QQ: I think we interact with zookeeper only via helix library client. Would we need this direct dependency?

snalli: Either way works but I prefer this as it is upfront and explicit. And if we import zookeeper directly, we are guaranteed we use the latest version.

SophieGuo410: Actually I prefer always leveraging the transitive dependency if it works since we don't need to take care of upgrading this version all the time. Let's say when we upgrade the helix lib next time, we don't need to remember to upgrade this zk version.

snalli: removed

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
```diff
@@ -16,7 +16,7 @@
 import com.fasterxml.jackson.databind.JsonNode;
 import com.github.ambry.quota.QuotaName;
 import java.util.concurrent.atomic.AtomicLong;
-import org.codehaus.jackson.annotate.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnore;
```
SophieGuo410, line 19: QQ: any reason to change this?

ambry-test-utils/src/main/java/com/github/ambry/utils/TestUtils.java
```diff
@@ -283,6 +283,8 @@ static class ZkServerWrapper {
     public ZkServerWrapper(String dataDir, String logDir, int port) {
       this.dataDir = new File(dataDir);
       this.dataLogDir = new File(logDir);
+      this.dataDir.delete();
```
SophieGuo410, line 286: Same here, why you need to add this delete(), just trying to understand your intension, thx!

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
snalli: code doesn't compile without this

ambry-test-utils/src/main/java/com/github/ambry/utils/TestUtils.java
snalli: This is just to ensure that the ZK log folder from previous run is cleaned up and we start from an empty dir

SophieGuo410 (approved):

ambry-quota/src/main/java/com/github/ambry/quota/capacityunit/CapacityUnit.java
snalli: Jing's PR https://github.com/linkedin/ambry/pull/2123/files#diff-63fa9099c720e34c848596bd8c1eb05ecf92da8810196b579ef47611fa1bc15bR206

justinlin-linkedin: the change is good, we shouldn't be using codehaus library anymore. fasterxml library is the newer release for the same library.

snalli:
Can i get a shipit on this ?

Arun-LinkedIn (approved):

PR #2185, opened 2022-08-18
Fix a NullPointerException in QuotaAwareOperationController

justinlin-linkedin:
Fix a NullPointerException.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
@@ -196,8 +196,8 @@ public class QuotaAwareOperationController extends OperationController {
             break;
           case REJECT:
             quotaAvailable = false;
-            logger.warn("Rejecting request for quota resource {} because of reject recommendation.",
-                quotaResource.toString());
+            logger.warn(
+                "Rejecting request for quota resource {} because of reject recommendation.", quotaResource.toString());
             routerMetrics.rejectedRequestRate.mark();
             nonCompliantRequests.add(requestInfo);
             requestQueue.get(quotaResource).removeFirst();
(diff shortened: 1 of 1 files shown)


ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -240,16 +241,21 @@ private void pollQuotaExceedAllowedRequestsIfAny(List<RequestInfo> requestsToSen
         requestQueue.get(quotaResource).removeFirst();
         if (requestQueue.get(quotaResource).isEmpty()) {
           requestQueue.remove(quotaResource);
+          // This quotaResource has exhausted, we should remove it from next for loop
+          quotaResourcesRemoved.add(quotaResource);
         }
       }
+      quotaResources.removeAll(quotaResourcesRemoved);
```
justinlin-linkedin, line 248: if the quotaResource is exhausted, we need to remove it from quotaResources list so next time we do the for loop in line 226, we don't have to deal with it anymore.

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -221,6 +221,7 @@ private void pollQuotaExceedAllowedRequestsIfAny(List<RequestInfo> requestsToSen
     Timer.Context timer = routerMetrics.pollExceedAllowedRequestTime.time();
     List<QuotaResource> quotaResources = new ArrayList<>(requestQueue.keySet());
     Collections.shuffle(quotaResources);
+    List<QuotaResource> quotaResourcesRemoved = new ArrayList<>();
     while (!requestQueue.isEmpty()) {
       for (QuotaResource quotaResource : quotaResources) {
         RequestInfo requestInfo = requestQueue.get(quotaResource).getFirst();
```
ankagrawal, line 227: Thanks for catching this bug. Would doing something like below here help with the performance a little?
if(!requestQueue.containsKey(quotaResource) { continue; }

justinlin-linkedin: make some change, I think this probably would be the most efficient way.

snalli (approved):

ankagrawal (approved):

PR #2187, opened 2022-08-22
Support getBlobChunkIds API for composite blob

SophieGuo410:
Will set it as admin request in close source.


snalli (commented):
Can you add some context in the commit msg as to why this is needed ? Paste the GCN number if u have one

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -69,6 +70,7 @@ class NonBlockingRouter implements Router {
   private final Cache<String, Boolean> notFoundCache;
   private final AmbryCache blobMetadataCache;
 
+
```
snalli, line 73: rm newline

ambry-router/src/test/java/com/github/ambry/router/GetManagerTest.java
```diff
@@ -534,6 +545,18 @@ private void compareBlobInfo(BlobInfo blobInfo, long segmentSize) {
...
 
+  /**
+   * Compare the size of chunk BlobId list with the expected chunk size + metadata chunk.
+   * @param chunkBlobIds List of chunk blob Ids for the composite blob.
+   */
+  private void compareChunkBlobIds(long blobSize, List<StoreKey> chunkBlobIds) {
+    if (blobSize <= CHUNK_SIZE) {
+      Assert.assertNull("chunkBlobIds should be null for simple blob", chunkBlobIds);
+    } else {
+      Assert.assertEquals("the size of chunk BlobId list should match", blobSize / CHUNK_SIZE + 1, chunkBlobIds.size());
```
snalli, line 556: Is this always true ? There could be a 4MB composite blob comprising four 1-MB chunks.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -1743,6 +1743,7 @@ private GetBlobOperation getAndAssertSuccess(final boolean getChunksBeforeRead,
                   blobInfo.getBlobProperties().getBlobSize());
               Assert.assertNull("Unexpected blob data in operation result", result.getBlobResult.getBlobDataChannel());
               Assert.assertEquals("LifeVersion mismatch", expectedLifeVersion, blobInfo.getLifeVersion());
+              break;
```
snalli, line 1746: not needed.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -162,7 +163,7 @@ private Callback<Void> securityPostProcessRequestCallback(BlobId blobId) {
           RestRequestMetrics restRequestMetrics = metricsGroup.getRestRequestMetrics(restRequest.isSslUsed(), true);
           restRequest.getMetricsTracker().injectMetrics(restRequestMetrics);
         }
-        if (subResource == null) {
+        if (subResource == null || subResource == SubResource.BlobChunkIds) {
```
justinlin-linkedin, line 166: This is a bit odd, we should probably do something like
```
if (subResource == SubResource.Replicas) {
    finalCallback.onCompletion(getReplicasHandler.getReplicas(blobId.getID(), restResponseChannel), null);
} else {
    router.getBlob(blobId.getID(), options, routerCallback(), QuotaUtils.buildQuotaChargeCallback(restRequest, quotaManager, true));
}
```

ambry-api/src/main/java/com/github/ambry/router/GetBlobResult.java
```diff
@@ -23,15 +25,18 @@
 public class GetBlobResult {
   private final BlobInfo blobInfo;
   private final ReadableStreamChannel blobDataChannel;
+  private final List<StoreKey> blobChunkIds;
```
justinlin-linkedin, line 28: we already have a class called GetBlobResultInternal, which includes List of store keys. Though that serves as an internal structure, we should probably remove that class all together now that we have this one.

ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
```diff
@@ -413,7 +413,7 @@ private void handleBody(InputStream payload, MessageMetadata messageMetadata, Me
     if (encryptionKey == null) {
       // if blob is not encrypted, move the state to Complete
       BlobInfo blobInfo = new BlobInfo(serverBlobProperties, userMetadata.array(), messageInfo.getLifeVersion());
-      operationResult = new GetBlobResultInternal(new GetBlobResult(blobInfo, null), null);
+      operationResult = new GetBlobResultInternal(new GetBlobResult(blobInfo, null, null), null);
```
justinlin-linkedin, line 416: This is what I am talking about, should just get rid of GetBlobResultInternal.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -312,7 +312,9 @@ private void onChunkOperationComplete(GetChunk chunk) {
           // GetManager remove this operation.
           setOperationCompleted();
           List<StoreKey> chunkIds = e == null && compositeBlobInfo != null ? compositeBlobInfo.getKeys() : null;
-          operationResult = new GetBlobResultInternal(null, chunkIds);
+          operationResult = new GetBlobResultInternal(
+              options.getBlobOptions.getOperationType() == GetBlobOptions.OperationType.BlobChunkIds
```
justinlin-linkedin, line 316: In `GetBlobOptionsInternal` there is a member called `getChunkIdsOnly`, we should probably get rid of that as well now that we have BlobChunkId as SubResource.

snalli (changes requested):
requesting changes

ambry-router/src/test/java/com/github/ambry/router/GetManagerTest.java
SophieGuo410: For all these tests used this getBlobAndCompareContent method, both stitched blob and composite blob has larger size than the one chunk. So I use this way to make the test simple. Let me know if you have any other preference.

ambry-api/src/main/java/com/github/ambry/router/GetBlobResult.java
SophieGuo410: You are right, I removed the GetBlobResultInternal.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
SophieGuo410: I checked the undelete and delete also need this getChunkIdsOnly in GetBlobOptionsInternal, so I just keep using it. Let me know if I miss anything. thx!

PR #2196, opened 2022-08-24
Adding metric to record the direct memory usage for index segment files

justinlin-linkedin:
Adding metric so record the direct memory usage for index segment files, so combined with the netty bytebuf pool, we can get a much better understanding of the direct memory usage of ambry-server.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
@@ -20,6 +20,7 @@ import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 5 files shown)


ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -259,6 +263,13 @@ void initializeIndexGauges(String storeId, final PersistentIndex index, final lo
           () -> index.getLogUsedCapacity() - blobStoreStats.getCachedValidSize().getSecond();
       registry.register(MetricRegistry.name(Log.class, prefix + "CurrentInvalidDataSize"), currentInvalidDataSize);
     }
+    if (enableStoreIndexDirectMemoryUsageMetric) {
+      indexes.putIfAbsent(storeId, index);
+      Gauge<Long> indexDirectMemoryUsage =
```
SophieGuo410, line 268: Would this PersistentIndex::getDirectMemoryUsage take long time which will delay the emitting of metric?

justinlin-linkedin: My guesstimation is that this might be slow, but it should be able to finish in the scope of 100s of milliseconds at worst. It depends on how many index segments we have on all the disks. 
I remember metric would be taken every minute, so it shouldn't be an issue. But just in case, I put a configuration here .

SophieGuo410: nit: probably you can add some debug log to get an idea of how long it will be take in difference cases, up to you.

SophieGuo410 (approved):

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
justinlin-linkedin: done

mikeyang923 (approved):

PR #2197, opened 2022-08-25
Dont update state delete state when undelete is disabled

justinlin-linkedin:
When undelete is disabled, DELETE will be the final state of the blob's life cycle. 
In this case, when we iterate through IndexSegment in findEntriesSince, we have to call update state for the DELETE record, since we already know it's the final state.

This PR should greatly improve the performance of replicating through a log segment that is full of DELETEs.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -126,6 +126,8 @@ class PersistentIndex {
   private static final Logger logger = LoggerFactory.getLogger(PersistentIndex.class);
   private final IndexPersistor persistor = new IndexPersistor();
   private final ScheduledFuture<?> persistorTask;
+  // When undelete is enabled, DELETE is not the final state of the blob, since a DELETEd blob can be UNDELTEd.
+  private final boolean isDeleteFinalStateOfBlob;
 
   // ReadWriteLock to make sure index values are always pointing to valid log segments.
   // In compaction's final steps, persistent index and log will update their internal maps to reflect the result of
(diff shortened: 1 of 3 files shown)


SophieGuo410 (approved):

snalli:
some unittest for this would be good.

justinlin-linkedin:
> some unittest for this would be good.

Added more test cases in IndexTest

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -225,6 +227,13 @@ public int compare(IndexEntry e1, IndexEntry e2) {
     this.incarnationId = incarnationId;
     this.maxInMemoryIndexSizeInBytes = config.storeIndexMaxMemorySizeBytes;
     this.maxInMemoryNumElements = config.storeIndexMaxNumberOfInmemElements;
+    if (config.storeCompactionFilter
```
ankagrawal, line 230: .equals() for string comparison

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1947,8 +1956,11 @@ private void updateStateForMessages(List<MessageInfo> messageEntries) throws Sto
     ListIterator<MessageInfo> messageEntriesIterator = messageEntries.listIterator();
     while (messageEntriesIterator.hasNext()) {
       MessageInfo messageInfo = messageEntriesIterator.next();
-      // for all the message info, we use most recent ref even for delete. since a deleted entry can be undeleted.
-      // ok to use most recent ref
+      // If the message is already deleted and DELETE is the final state of the blob, then we don't have
+      // to update the state for this message.
+      if (isDeleteFinalStateOfBlob && messageInfo.isDeleted()) {
+        continue;
```
ankagrawal, line 1962: Should we add unit tests for this?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: updated.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: We already have findEntriesSinceTest in the IndexTest file. I added a new parameter to that unit test class for this.

ankagrawal (approved):

PR #2199, opened 2022-08-27
Common compression interface and factory, plus LZ4 and ZSTD support.

panchow:
This is the first part of the Ambry compression support.
It contains the generic re-usable compression library and interface.  It is not called by Ambry yet.


ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Compress the source data specified in {@code sourceData}.  This method will allocate buffer to store the
+   * compressed data, the name, and original data size.  The returned buffer may be bigger than sourceData
+   * in case sourceData is incompressible.  The return value contains both the buffer and actual usage in the buffer.
+   *
+   * The compressed buffer format:
+   * - Algorithm name size (1 byte) - size of AlgorithmName in bytes.
+   * - Algorithm name (N bytes) - name of compression algorithm, up to the max size specified in CompressionFactory.
+   * - Original Data size (4 bytes) - up to 2 GB.
+   * - Compressed data - the output of the compression algorithm.
```
justinlin-linkedin, line 63: I don't quite get the necessity of storing these information in compressed data.
We know in some cases we are not going to compress some data, which means that in order to know if a data is compressed or not, we have to store this piece of information somewhere. It's most likely that we would store this information in blob properties. Why not just sore all these information, algorithm name and original data size in blob properties as well?

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -0,0 +1,82 @@
...
+   *
+   * Note that the compressed data does not contain original data size, but decompress() requires the original data
+   * size.  The original data size is stored as 4-bytes integer in the compressed buffer.
+   *
+   * @param sourceData The source uncompressed data to compress.  It cannot be null or empty.
+   * @return Pair that contains the compressed buffer and the buffer usage size in bytes.
+   */
+  Pair<Integer, byte[]> compress(byte[] sourceData);
```
justinlin-linkedin, line 71: This looks good, but I also want to propose another method, which takes source byte array and target byte array, so we can create a target byte array in our own way. We are heavily rely on netty to manage byte buffer for us, without this, we would have to copy the compress byte buffer to netty after compression.
That being said, we should also expose a method to tell us the size of compressed byte array.

justinlin-linkedin (commented):
Most part looks good, just some questions for the interface.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
panchow: That's a good suggestion.  As we discussed, we have 2 options: store the compression algorithm, size, and compressed binary into the blob, or store the compressed binary into the blob and add compression algorithm and original size in metadata.  Since we need to store a metadata to indicate whether blob is compressed, it makes sense to add the compression algorithm and original size to the metadata too.  However, the compressed size is unique per chunk and we currently do not support per-chunk metadata, so this suggestion would not work unless we add per-chunk metadata.

I will add a version # (1 byte size) to the beginning in case we need to alter the structure in the future.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
panchow: Another good suggestion.  I'll make that change.

ambry-commons/src/main/java/com/github/ambry/compression/ZstdCompression.java
```diff
@@ -0,0 +1,138 @@
...
+   * @param compressedBufferSize Size of the buffer where compression can write to.  This size should be same as
+   *                             estimateMaxCompressedDataSize().
+   * @return The size of the compressed data in bytes.
+   */
+  @Override
+  protected int compressData(byte[] sourceData, int sourceDataOffset, int sourceDataSize,
+      byte[] compressedBuffer, int compressedBufferOffset, int compressedBufferSize) {
+
+    long compressedSize = Zstd.compressByteArray(compressedBuffer, compressedBufferOffset,
+        compressedBuffer.length - compressedBufferOffset, sourceData, 0, sourceData.length,
```
justinlin-linkedin, line 97: shouldn't this be `soourceDataOffset` instead of 0.

ambry-commons/src/main/java/com/github/ambry/compression/ZstdCompression.java
```diff
@@ -0,0 +1,138 @@
...
+   * @param compressedBufferOffset The offset in compressedBuffer where the decompression should start reading.
+   * @param compressedBufferSize Size of the compressed buffer returned from compressData().
+   * @param sourceDataBuffer The buffer to store decompression output (the original source data).
+   * @param sourceDataOffset Offset where to write the decompressed data.
+   * @param sourceDataSize Size of the buffer to hold the decompressed data.  It should be size of original data.
+   */
+  @Override
+  protected void decompressData(byte[] compressedBuffer, int compressedBufferOffset, int compressedBufferSize,
+      byte[] sourceDataBuffer, int sourceDataOffset, int sourceDataSize) {
+    long decompressedSize = Zstd.decompressByteArray(sourceDataBuffer, 0, sourceDataBuffer.length,
```
justinlin-linkedin, line 126: same above

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
```diff
@@ -0,0 +1,124 @@
...
+    return registeredCompressionMap.get(algorithmName);
+  }
+
+  /**
+   * Set the default compressor to use when compressing data.
+   * @param newCompressor The new compressor to use.
+   */
+  public void setDefaultCompressor(Compression newCompressor) {
+    defaultCompressor = newCompressor;
+  }
```
justinlin-linkedin, line 96: let's do a sanity check that this compressor is already registered.

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Given the compressed buffer generated by compressor's compress() method,
+   * find the decompressor based on the algorithm name stored in compressedBuffer.
+   *
+   * @param compressedBuffer The compressed buffer generated by the compress() method.
+   * @return The decompressor that can decompress this compressed buffer.  Null if no decompressor found.
+   */
+  public Compression getDecompressor(byte[] compressedBuffer) {
```
justinlin-linkedin, line 114: I don't think we differentiate Compressor from Decompressor. So it's probably a bit confusing as an interface. Maybe change the name to `getCompressionFromCompressedData`, or something short.

ambry-commons/src/test/java/com/github/ambry/compression/BaseCompressionTest.java
```diff
@@ -0,0 +1,219 @@
...
+import org.mockito.Mockito;
+
+public class BaseCompressionTest {
+
+  @Test
+  public void testGetAlgorithmNameBinary() throws Throwable {
+    // Test: A single-letter algorithm name.
+    BaseCompression compressor = Mockito.mock(BaseCompression.class);
+    Mockito.when(compressor.getAlgorithmName()).thenReturn("A");
+    byte[] binary = (byte[]) MethodUtils.invokeMethod(compressor, true, "getAlgorithmNameBinary");
```
justinlin-linkedin, line 30: what is the purpose of using this indirect method call? we can just do compressor.getAlgorithemNameBinary.

ambry-commons/src/test/java/com/github/ambry/compression/ZstdCompressionTest.java
```diff
@@ -0,0 +1,59 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.compression;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class ZstdCompressionTest {
```
justinlin-linkedin, line 19: we can merge ZstdCompressionTest and ZstdCompressionFailedTest, just like the LZ4 tests.

ambry-commons/src/test/java/com/github/ambry/compression/ZstdCompressionTest.java
```diff
@@ -0,0 +1,59 @@
...
+  @Test
+  public void testCompressAndDecompress_MinimumLevel() {
+    compression.setCompressionLevel(compression.getMinimumCompressionLevel());
+    LZ4CompressionTest.compressDataAndDecompressDataTest(compression, "Test minimum compression using minimum level.");
+  }
+
+  @Test
+  public void testCompressAndDecompress_MaximumLevel() {
+    compression.setCompressionLevel(compression.getMaximumCompressionLevel());
+    LZ4CompressionTest.compressDataAndDecompressDataTest(compression, "Test maximum compression using maximum level.");
```
justinlin-linkedin, line 56: maybe create a util class in the same test package for this shared method?

ambry-test-utils/src/main/java/com/github/ambry/utils/TestUtils.java
```diff
@@ -503,4 +506,59 @@ public static boolean isSrcDestSync(String srcZkString, String srcClusterName, S
...
+   * Call a method by name using reflection, such as private method, and return the exception thrown.
+   * @param instance The private method instance.
+   * @param methodName Name of the method to call, such as the private method name.
+   * @param parameters The parameters.
+   * @return The exception thrown by the method.  Null if call did not throw exception.
+   * @throws ExecutionException if failed to invoke the method.  The method is not expected to throw this exception.
+   */
+  public static Throwable invokeAndGetException(Object instance, String methodName, Object... parameters)
+      throws ExecutionException {
+    try {
```
justinlin-linkedin, line 520: please format this file

build.gradle
```diff
@@ -100,7 +100,13 @@ subprojects {
         testCompile "org.apache.logging.log4j:log4j-slf4j-impl:$log4jVersion"
         testCompile "junit:junit:$junitVersion"
         testCompile "org.mockito:mockito-core:2.+"
+        // Add PowerMock to support more unit test scenarios.
+        // See https://github.com/powermock/powermock/wiki/Mockito for detail about PowerMock.
+        testCompile "org.powermock:powermock-api-mockito2:2.+"
```
justinlin-linkedin, line 105: please put the version to the dependency_version file.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -0,0 +1,220 @@
...
+   * This method will allocate buffer to store the compressed data.
+   * The returned buffer may be bigger than sourceData in case sourceData is incompressible.
+   * The return value contains both the buffer and actual usage in the buffer.
+   *
+   * @param sourceData The source uncompressed data to compress.  It cannot be null or empty.
+   * @return Pair that contains the compressed buffer and the buffer usage size in bytes.
+   */
+  default Pair<Integer, byte[]> compress(byte[] sourceData) {
+    if (sourceData == null) {
+      throw new IllegalArgumentException("sourceData cannot be null.");
```
ankagrawal, line 151: the javadocs say that `sourceData` cannot be empty. Should we include an empty check too?

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Decompress portion of the shared buffer in {@code compressedBuffer}.  Both the {@code compressedBuffer} and
+   * {@code sourceData} may be shared buffers and that's why both provide their offset and size as parameters.
+   * The compressed buffer is the output from compress() that contains the version, algorithm name, original data size,
+   * and compressed binary.
+   *
+   * @param compressedBuffer The compressed buffer generated in compress() method.
+   * @param compressedBufferOffset Offset in the compressedBuffer to start reading.
+   * @param compressedBufferSize Number of bytes to read in compressedBuffer.
+   * @param sourceData The buffer to hold the decompressed/original data.
```
ankagrawal, line 182: nit: Should we call it `decompressedData`?

ambry-api/src/main/java/com/github/ambry/compression/CompressionException.java
```diff
@@ -0,0 +1,36 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.compression;
+
+/**
+ * Generic compression failure exception.
+ */
+public class CompressionException extends RuntimeException {
```
ankagrawal, line 19: Any reason why we should create it as a RuntimeException?

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -0,0 +1,220 @@
...
+    return getDecompressBufferSize(compressedBuffer, 0, compressedBuffer.length);
+  }
+
+  /**
+   * Compress portion of the shared buffer specified in {@code sourceData}.  Both {@code sourceData}
+   * and {@code compressedBuffer} can be shared/composite buffers.  That's why both buffers need to provide
+   * their size and offset.  The caller is responsible for managing memory allocation.
+   *
+   * The compressed buffer is a structure in this format:
+   * - Version (1 byte) - the structure/format version.  It starts with value '1'.
```
ankagrawal, line 120: Would we also need to encode the compression level along with the algorithm name?

ambry-api/src/test/java/com/github/ambry/compression/CompressionExceptionTest.java
```diff
@@ -0,0 +1,31 @@
...
+  @Test
+  public void testConstructor() {
+    CompressionException ex = new CompressionException("error");
+    Assert.assertNotNull(ex.getMessage());
+
+    ex = new CompressionException("failed", new Exception());
+    Assert.assertNotNull(ex.getMessage());
+    Assert.assertNotNull(ex.getCause());
+  }
+}
```
ankagrawal, line 31: nit: missing newline at the end of file.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -0,0 +1,290 @@
...
+   * @param sourceData The source uncompressed data.  It has already been null and empty verified.
+   * @param sourceDataOffset Offset in sourceData to start reading.
+   * @param sourceDataSize Number of bytes in sourceData to compress.
+   * @param compressedBuffer  The buffer to hold the compressed data.
+   * @param compressedBufferOffset The offset in compressedBuffer where the compression output should write.
+   * @param compressedBufferSize Size of the buffer where compression can write to.  This size should be same as
+   *                             estimateMaxCompressedDataSize().
+   * @return The size of the compressed data in bytes.
+   */
+  protected abstract int compressData(byte[] sourceData, int sourceDataOffset, int sourceDataSize,
```
ankagrawal, line 58: How is it different from Compression::compress? Curious why do we need this?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -0,0 +1,290 @@
...
+   * The decompression implementation is algorithm specific.
+   *
+   * @param compressedBuffer The compressed buffer.
+   * @param compressedBufferOffset The offset in compressedBuffer where the decompression should start reading.
+   * @param compressedBufferSize Size of the compressed buffer returned from compressData().
+   * @param sourceDataBuffer The buffer to store decompression output (the original source data).
+   * @param sourceDataOffset Offset where to write the decompressed data.
+   * @param sourceDataSize Size of the buffer to hold the decompressed data.  It should be size of original data.
+   */
+  protected abstract void decompressData(byte[] compressedBuffer, int compressedBufferOffset, int compressedBufferSize,
```
ankagrawal, line 73: same as above.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * The base compression algorithm implementation.
+ * It contains common algorithm implementation methods.
+ */
+public abstract class BaseCompression implements Compression {
+
+  // The maximum length of the algorithm name in bytes.
+  // This prevents algorithms from using long names that occupy more disk space.
+  public static final byte MAX_ALGORITHM_NAME_LENGTH = 10;
```
ankagrawal, line 26: Do you think making it configurable is a good idea?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -0,0 +1,290 @@
...
+   * - algorithm name binary.
+   * - original data size (4 bytes)
+   * - worst-case compression output.
+   * This is an estimate calculation it is usually slightly larger than the sourceDataSize.
+   *
+   * @param sourceDataSize The source data size.
+   * @return The estimated buffer size required to hold the compressed data.
+   */
+  public int getCompressBufferSize(int sourceDataSize) {
+    return estimateMaxCompressedDataSize(sourceDataSize) + getAlgorithmNameBinary().length + 5;
```
ankagrawal, line 124: The size could also depend on the version. A newer version might introduce a new header field.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -0,0 +1,290 @@
...
+   * compression.  That is why it accepts the offset and size parameters.
+   * @param compressedBuffer The compressed buffer.
+   * @param compressedBufferOffset The offset in the compressed buffer where compression data is stored.
+   * @param compressedBufferSize The size in compressedBuffer to read.
+   * @return The size of original data.
+   */
+  public int getDecompressBufferSize(byte[] compressedBuffer, int compressedBufferOffset, int compressedBufferSize) {
+    // Verify each parameter.
+    verifyCompressedBuffer(compressedBuffer, compressedBufferOffset, compressedBufferSize);
+    if (compressedBufferSize < 6) {
```
ankagrawal, line 140: This check could depend on the version.

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
```diff
@@ -0,0 +1,130 @@
...
+  public Compression getCompressorByName(String algorithmName) {
+    return registeredCompressionMap.get(algorithmName);
+  }
+
+  /**
+   * Set the default compressor to use when compressing data.  The new compressor can be null to clear default.
+   * If a new compressor is specified, it must be registered.
+   * @param newCompressor The new compressor to use.
+   */
+  public void setDefaultCompressor(Compression newCompressor) {
```
ankagrawal, line 97: nit: maybe we can also register the compressor here in the same call.

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Get the shared instance of compression factory.
+   * The shared instance contains a pre-registered set of supported compression algorithms.
+   * @return The shared compression factory instance.
+   */
+  public static CompressionFactory getDefault() {
+    if (defaultFactory != null) return defaultFactory;
+
+    synchronized(CompressionFactory.class) {
+      if (defaultFactory == null) {
```
ankagrawal, line 45: Before we start a process (either frontend or server), we know which compression algorithm(s) to use. Therefore, I would recommend to take the compression class names only from config. Ideally there shouldn't be any case where we don't know what `Compression` to use before starting the process.

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
```diff
@@ -0,0 +1,130 @@
...
+  /**
+   * Given the compressed buffer generated by compressor's compress() method,
+   * find the decompressor based on the algorithm name stored in compressedBuffer.
+   *
+   * @param compressedBuffer The compressed buffer generated by the compress() method.
+   * @return The decompressor that can decompress this compressed buffer.  Null if no decompressor found.
+   */
+  public Compression getCompressionFromCompressedData(byte[] compressedBuffer) {
+    // Compressed buffer contains at least version (1 byte), algorithm name size (1 byte) and original size (4 bytes).
+    if (compressedBuffer == null || compressedBuffer.length < 6) {
```
ankagrawal, line 122: the length check here could be header version dependent.

ambry-commons/src/main/java/com/github/ambry/compression/LZ4Compression.java
```diff
@@ -0,0 +1,160 @@
...
+  // For performance purpose, cache the factory.
+  private static final LZ4Factory LZ4_FACTORY = LZ4Factory.fastestInstance();
+
+  /**
+   * Get the unique name of this compression algorithm.
+   * WARNING - Do not change the algorithm name.  See Compression interface for detail.
+   * @return Name of this algorithm.
+   */
+  @Override
+  public String getAlgorithmName() { return "LZ4"; }
```
ankagrawal, line 50: nit: make the algorithm name a public static final string?

ambry-commons/src/main/java/com/github/ambry/compression/LZ4Compression.java
```diff
@@ -0,0 +1,160 @@
...
+   */
+  @Override
+  public int getMinimumCompressionLevel() { return 0; }
+
+  /**
+   * Get the maximum compression level.
+   * @return The maximum compression level.
+   */
+  @Override
+  public int getMaximumCompressionLevel() { return 17; }
```
ankagrawal, line 64: nit: make the max and min compression level a public static final int?

ambry-commons/src/main/java/com/github/ambry/compression/ZstdCompression.java
```diff
@@ -0,0 +1,137 @@
...
+public class ZstdCompression extends BaseCompressionWithLevel {
+
+  /**
+   * Get the unique name of this compression algorithm.
+   * WARNING - Do not change the algorithm name.  See Compression interface for detail.
+   * @return Name of this algorithm.
+   */
+  @Override
+  public String getAlgorithmName() {
+    return "ZSTD";
```
ankagrawal, line 36: public static final string/int for constants?

ankagrawal: Also maybe use enums for supported algorithm names.

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
panchow: This factory is used for decompression.  It registers all the support decompression.  The compression algorithm is set in setDefaultCompressor().  I do not think the we should select the default compressor in config either.  I discussed my point of view about config above.  Let's chat if we need to make it a config.

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
panchow: It's debatable.  The API is just setting the default compressor.  If the impl also registers the compressor, it's considered side-effect.  Caller can call always call register() if it wants.

ambry-commons/src/main/java/com/github/ambry/compression/CompressionFactory.java
panchow: This is related to the 6 check above.  This size 6 check is optional.  The size 6 is minimal that should be applicable to all versions.  When we add a new version, the change can revisit this minimal requirement.

ambry-commons/src/main/java/com/github/ambry/compression/LZ4Compression.java
panchow: It can't be static since it's part of the interface.  If I make it final, it becomes non-inheritable.  Since there is no setAlgortihmName, it's okay to leave it as is.

ambry-commons/src/main/java/com/github/ambry/compression/LZ4Compression.java
panchow: It can't be static or final since they are part of the CompressionLevel interface.

ambry-commons/src/main/java/com/github/ambry/compression/ZstdCompression.java
panchow: No enum for this case. Making it enum means adding a new algorithm will need to modify the enum.  Say, another team wants to use our library and they want to provide their own compression algorithm, they can't modify our enum class to add their class.  Enum is useful to make code readable but limits expansion

justinlin-linkedin (approved):
I don't see major blocker for approval, but please format your files. Seems some indentation is off in some of your files.

ankagrawal (approved):
Looks good to me. Please make sure that the tests are passing.

PR #2201, opened 2022-08-30
Dump replicas' state when hit 503

JingQianCloud:
Dump eligible replica list when we hit 503.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -429,9 +429,10 @@ class SimpleOperationTracker implements OperationTracker {
       logger.info(
           "Terminating {} on {} due to disk down count and total Not_Found count from eligible replicas and some "
               + "other replicas being unavailable. CrossColoEnabled: {}, DiskDownCount: {}, TotalNotFoundCount: {}, "
-              + "TotalReplicaCount: {}, replicaSuccessTarget: {}, OfflineReplicaCount: {}, allReplicaCount: {} {}",
+              + "TotalReplicaCount: {}, replicaSuccessTarget: {}, OfflineReplicaCount: {}, allReplicaCount: {} {} "
+              + "replicasByState = {}",
           routerOperation, partitionId, crossColoEnabled, diskDownCount, totalNotFoundCount, totalReplicaCount,
-          replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, getBlobIdLog());
+          replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, getBlobIdLog(), allDcReplicasByState);
       routerMetrics.failedMaybeDueToTotalOfflineReplicasCount.inc();
       return true;
     }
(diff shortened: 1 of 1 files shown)


ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -429,9 +429,10 @@ public boolean maybeFailedDueToOfflineReplicas() {
       logger.info(
           "Terminating {} on {} due to disk down count and total Not_Found count from eligible replicas and some "
               + "other replicas being unavailable. CrossColoEnabled: {}, DiskDownCount: {}, TotalNotFoundCount: {}, "
-              + "TotalReplicaCount: {}, replicaSuccessTarget: {}, OfflineReplicaCount: {}, allReplicaCount: {} {}",
+              + "TotalReplicaCount: {}, replicaSuccessTarget: {}, OfflineReplicaCount: {}, allReplicaCount: {} {} "
+              + "replicaPool = {}",
           routerOperation, partitionId, crossColoEnabled, diskDownCount, totalNotFoundCount, totalReplicaCount,
-          replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, getBlobIdLog());
+          replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, getBlobIdLog(), replicaPool);
```
justinlin-linkedin, line 435: I am afraid that dumping replicaPool is not enough. Replica Pool only tells us the "good" replicas, we still don't know the "bad" replicas' states.

We should probably construct a map from replica Id to its replica state for all replicas and then dump this map here.

JingQianCloud: You are right. Submit the second revision.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -426,12 +426,16 @@ public boolean maybeFailedDueToOfflineReplicas() {
...
+              ReplicaState.OFFLINE, ReplicaState.ERROR, ReplicaState.DROPPED));
       logger.info(
           "Terminating {} on {} due to disk down count and total Not_Found count from eligible replicas and some "
               + "other replicas being unavailable. CrossColoEnabled: {}, DiskDownCount: {}, TotalNotFoundCount: {}, "
-              + "TotalReplicaCount: {}, replicaSuccessTarget: {}, OfflineReplicaCount: {}, allReplicaCount: {} {}",
+              + "TotalReplicaCount: {}, replicaSuccessTarget: {}, OfflineReplicaCount: {}, allReplicaCount: {} {} "
+              + "replicasByState = {}",
           routerOperation, partitionId, crossColoEnabled, diskDownCount, totalNotFoundCount, totalReplicaCount,
-          replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, getBlobIdLog());
+          replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, getBlobIdLog(), replicasByState);
```
justinlin-linkedin, line 438: there is already a class member called allDcReplicasByState, we should use that member.

snalli:
Can u fix the commit msg ? it might be too verbose and has some Li internal info

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -426,12 +426,16 @@ public boolean maybeFailedDueToOfflineReplicas() {
     }
     if (hasFailedOnCrossColoNotFound() && allReplicaCount - totalNotFoundCount >= replicaSuccessTarget
         && totalOfflineReplicaCount > 0) {
+      Map<ReplicaState, List<ReplicaId>> replicasByState = getReplicasByState(null,
+          EnumSet.of(ReplicaState.BOOTSTRAP, ReplicaState.STANDBY, ReplicaState.LEADER, ReplicaState.INACTIVE,
```
snalli, line 430: can u put this enum in a common file so that u and others can reuse ?

JingQianCloud: 1. About the message, good point. I've removed the printout.
2. Will use one existing variable. So the line will be removed.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: I missed that. Thanks.

snalli (approved):

PR #2202, opened 2022-08-31
Add new request type BlobChunkIds

SophieGuo410:
Add new request type BlobChunkIds to retrieve data-chunk IDs of composite blobs.
This will help us do get data chunk Ids for composite blob to see each chunk detail for testing purpose.
Ignore this test testRouterMetadataCacheFirstChunkOperationFlag() which needs further investigation and not related to this pr.


snalli:
Fix the title and commit description to be more readable.

Title:
Add new request type BlobChunkIds

Commit:
Add new request type BlobChunkIds to retrieve data-chunk IDs of composite blobs.
This will help us do XYZ.
<No need to mention anything else about 2187, testing, closed src etc.>

ambry-router/src/main/java/com/github/ambry/router/GetManager.java
```diff
@@ -292,18 +292,3 @@ class GetBlobOptionsInternal {
...
-  /**
-   * Construct a GetBlobResultInternal instance.
-   * @param getBlobResult The {@link GetBlobResult} associated with this instance, if there is one..
-   * @param storeKeys The store keys associated with this instance, if there are any.
-   */
-  public GetBlobResultInternal(GetBlobResult getBlobResult, List<StoreKey> storeKeys) {
-    this.getBlobResult = getBlobResult;
-    this.storeKeys = storeKeys;
-  }
-}
```
justinlin-linkedin, line 309: we can also remove GetBlobOptionsInternal's getChunkIdsOnly and solely rely on the GetBlobOptions' OperationType.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -210,27 +205,76 @@ private Callback<GetBlobResult> routerCallback() {
...
+
+    /**
+     * set chunk blob Ids for composite blob in response channel.
+     * @param result The result of the request i.e a {@link GetBlobResult} object with the properties of the blob.
+     * @return a {@link ReadableStreamChannel} that represents the response to the {@code restRequest}.
+     * @throws RestServiceException if there's any error during serialization.
+     */
+    private ReadableStreamChannel setChunkBlobIdsInResponseChannel(GetBlobResult result)
+        throws RestServiceException {
+      List<StoreKey> blobChunkIds = result.getBlobChunkIds();
```
justinlin-linkedin, line 272: blobChunkIds could be null when the blob is not a composite blob. we have to deal with it. What do we want to return in this case? An empty map? or a map but BLOB_CHUNK_IDS_KEY's value is an empty list?

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -162,21 +167,11 @@ private Callback<Void> securityPostProcessRequestCallback(BlobId blobId) {
           RestRequestMetrics restRequestMetrics = metricsGroup.getRestRequestMetrics(restRequest.isSslUsed(), true);
           restRequest.getMetricsTracker().injectMetrics(restRequestMetrics);
         }
-        if (subResource == null) {
+        if (subResource == SubResource.Replicas) {
+          finalCallback.onCompletion(getReplicasHandler.getReplicas(blobId.getID(), restResponseChannel), null);
+        } else {
```
snalli, line 172: Please don't refactor. Make a separate PR if you want to refactor.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -210,27 +205,76 @@ private Callback<GetBlobResult> routerCallback() {
...
-              BlobInfo blobInfo = result.getBlobInfo();
-              if (restRequest.getArgs().containsKey(SEND_USER_METADATA_AS_RESPONSE_BODY) && (boolean) restRequest.getArgs()
-                  .get(SEND_USER_METADATA_AS_RESPONSE_BODY)) {
-                restResponseChannel.setHeader(Headers.CONTENT_TYPE, "application/octet-stream");
-                restResponseChannel.setHeader(Headers.CONTENT_LENGTH, blobInfo.getUserMetadata().length);
-                response = new ByteBufferReadableStreamChannel(ByteBuffer.wrap(blobInfo.getUserMetadata()));
-              } else {
-                restResponseChannel.setHeader(Headers.CONTENT_LENGTH, 0);
-                response = new ByteBufferReadableStreamChannel(EMPTY_BUFFER);
+            if (subResource == null) {
```
snalli, line 208: This refactor is ok since it vastly improves reaadability

snalli: But I still recommend a separate PR since I just want to focus on your changes alone

snalli: A bit tedious to separate out what is really your change and what is not.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetReplicasHandler.java
```diff
@@ -17,14 +17,12 @@
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.commons.BlobId;
-import com.github.ambry.commons.ByteBufferReadableStreamChannel;
```
snalli, line 20: remove this change please. This doesn't add to this PR.

ambry-frontend/src/test/java/com/github/ambry/frontend/FrontendRestRequestServiceTest.java
```diff
@@ -923,6 +924,28 @@ public void getReplicasTest() throws Exception {
     }
   }
 
+  @Test
```
snalli, line 927: short docstring

ambry-router/src/main/java/com/github/ambry/router/GetOperation.java
```diff
@@ -45,7 +44,7 @@ abstract class GetOperation {
   protected final NonBlockingRouterMetrics routerMetrics;
   protected final ClusterMap clusterMap;
   protected final ResponseHandler responseHandler;
-  protected final Callback<GetBlobResultInternal> getOperationCallback;
+  protected final Callback<GetBlobResult> getOperationCallback;
```
snalli, line 47: Are any of these changes really needed to add a new request type to get BlobChunkIds ? The PR is expanding and its getting harder to rread

snalli:
So this PR has grown too big.. Could you please filter out all refactoring into a separate PR ?

snalli:
Some unittest failures. I restarted the job but it failed again.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
SophieGuo410: I will return a map with empty list, just to keep it consistent.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
SophieGuo410: Actually after a second thought, I decided not to refactor this code. Instead I switch back to the original coding style so in the future when people add more subResource, they will quickly get this code need to be updated.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
SophieGuo410: For this part, if I don't optimize the code, with my new changes it will be much harder to read. So I would prefer to use this way instead.
The only new feature I added is case BlobChunkIds.

ambry-router/src/main/java/com/github/ambry/router/GetOperation.java
SophieGuo410: It's not a new type, It's the existing type which we should use. Previously we use GetBlobResultInternal is because we don't have blobChunkIds in GetBlobResult. With my new changes, we should use GetBlobResult.

ambry-router/src/main/java/com/github/ambry/router/GetManager.java
SophieGuo410: Let me check and I'll use a separate pr to update this part.

SophieGuo410:
> Some unittest failures. I restarted the job but it failed again.

Test passed after I close the router for testGetBlobChunkIds() in NonBlockingRouterTest.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendUtils.java
```diff
@@ -13,6 +13,7 @@
  */
 package com.github.ambry.frontend;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
```
justinlin-linkedin, line 16: not used anywhere.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendUtils.java
```diff
@@ -33,6 +34,7 @@
 import java.io.Writer;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.List;
```
justinlin-linkedin, line 37: same above.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -210,27 +219,80 @@ private Callback<GetBlobResult> routerCallback() {
...
+        restResponseChannel.setHeader(Headers.CONTENT_LENGTH, blobInfo.getUserMetadata().length);
+        return new ByteBufferReadableStreamChannel(ByteBuffer.wrap(blobInfo.getUserMetadata()));
+      } else {
+        restResponseChannel.setHeader(Headers.CONTENT_LENGTH, 0);
+        return new ByteBufferReadableStreamChannel(EMPTY_BUFFER);
+      }
+    }
+
+    /**
+     * set chunk blob Ids for composite blob in response channel.
```
justinlin-linkedin, line 279: nit: let's capitalize the comments. `Set chunk blob ...`

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -38,13 +38,19 @@
...
+import com.github.ambry.store.StoreKey;
 import com.github.ambry.utils.MockTime;
 import com.github.ambry.utils.SystemTime;
 import com.github.ambry.utils.TestUtils;
 import com.github.ambry.utils.Utils;
 import java.io.ByteArrayOutputStream;
+import java.io.OutputStreamWriter;
+import java.io.Writer;
 import java.nio.ByteBuffer;
+import java.nio.charset.StandardCharsets;
```
justinlin-linkedin, line 53: don't thing this is used anywhere.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -1817,6 +1823,28 @@ public void testBlobNotFoundCache() throws Exception {
     }
   }
 
+  /**
+   * Test get blob chunk Ids for composite blob.
+   * @throws Exception
+   */
+  @Test
+  public void testGetBlobChunkIds() throws Exception {
```
justinlin-linkedin, line 1831: can you also throw a test case to for simple blob, which should return an empty list.

justinlin-linkedin (commented):
looks mostly good to me, just a couple nitpicking

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -235,7 +237,9 @@ public Future<GetBlobResult> getBlob(String blobIdStr, GetBlobOptions options, f
     }
     currentOperationsCount.incrementAndGet();
     final FutureResult<GetBlobResult> futureResult = new FutureResult<>();
-    GetBlobOptionsInternal internalOptions = new GetBlobOptionsInternal(options, false, routerMetrics.ageAtGet);
+    GetBlobOptionsInternal internalOptions =
+        new GetBlobOptionsInternal(options, options.getOperationType() == GetBlobOptions.OperationType.BlobChunkIds,
```
snalli, line 241: Make a new constructor like this.

`  
GetBlobOptionsInternal(GetBlobOptions getBlobOptions,  NonBlockingRouterMetrics.AgeAtAccessMetrics ageAtAccessTracker) {
    this.getBlobOptions = getBlobOptions;
    this.getChunkIdsOnly = getBlobOptions.getOperationType() == GetBlobOptions.OperationType.BlobChunkIds;
    this.ageAtAccessTracker = ageAtAccessTracker;
  }
`

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
```diff
@@ -301,19 +302,20 @@ protected void doOperationTowardsMaybeCompositeBlob(final String blobIdStr, Futu
...
                   helper.getOpName()), RouterErrorCode.UnexpectedInternalError);
           NonBlockingRouter.completeOperation(futureResult, callback, null, exception, false);
         } else {
           List<String> blobIdStrs = new ArrayList<>();
           blobIdStrs.add(blobIdStr);
-          if (result.storeKeys != null) {
-            result.storeKeys.forEach(key -> blobIdStrs.add(key.getID()));
+          List<StoreKey> blobChunkIds = result == null ? new ArrayList<>() : result.getBlobChunkIds();
+          if (blobChunkIds != null) {
+            blobChunkIds.forEach(key -> blobIdStrs.add(key.getID()));
```
snalli, line 318: 1> Why is result null ? A better way is for it to be never null and just be an empty list if we have nothing to serve

snalli: Where is blobChunkIds used outside of this callback ? If not, why init it ?

snalli: This is better.
```
if (result != null && result.getBlobChunkIds() != null) {
result.getBlobChunkIds().forEach(key -> blobIdStrs.add(key.getID()));
}
```

snalli: If result.getBlobChunkIds() is never null, then you can skip that boolean

snalli: Ok so one way to avoid this member BlobChunkIds from being null is to control its initialization in the constructor itself. This gets rid of all the null checks and eliminates any possiblity of null ptr exception. 

```
  public GetBlobResult(BlobInfo blobInfo, ReadableStreamChannel blobDataChannel, List<StoreKey> blobChunkIds) {
    this.blobInfo = blobInfo;
    this.blobDataChannel = blobDataChannel;
    this.blobChunkIds = (blobChunkIds == null) ? new ArrayList<>() : blobChunkIds;
  }
```

snalli: Unless null is interpreted in a special way to mean something else elsewhere, I prefer to not have any nulls as much as possible.

snalli: Fewer null checks in if-clauses makes code easier to read.

snalli: I'd prefer result also to be not null, but that is beyond this PR

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
snalli: +1

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
SophieGuo410: 1. If we use GetBlobOptionsInternal, I think set the getChunkIdsOnly outside of constructor would be more readable. Otherwise you will see something like new GetBlobOptionsInternal(options, false, routerMetrics.ageAtGet) but inside the constructor it set to true at some case. It's kind of not clear to me. 
2. As we discussed, I will have a follow pr to replace the GetBlobOptionsInternal with GetBlobOptions.

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
SophieGuo410: blobChunkIds could be null when we GET without calling BlobChunkId subResource.
It's based on legacy code like how we handle GetBlobInternalResult. 
But I think you're right, I'm gonna have a follow up pr to set blobChunkIds as empty list if it don't have value at producer end. 
I would not prefer to handle it either in consumer or producer side instead of in the constructor.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
SophieGuo410: Since right now we handle null to empty list mapping in consumer side(getBlobHandler response) to keep consistent with legacy code in this pr, the new test I've added will return null instead of empty list. But I will have a follow up pr to set it as empty list from producer side, so the test will be updated to return an empty list in the future pr.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
snalli: > Otherwise you will see something like new GetBlobOptionsInternal(options, false, routerMetrics.ageAtGet) but inside the constructor it set to true at some case

Why will this happen ? They are two distinct ctors

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
snalli: > I'm gonna have a follow up pr to set blobChunkIds as empty list if it don't have value at producer end.

```
    this.blobChunkIds = (blobChunkIds == null) ? new ArrayList<>() : blobChunkIds;
```

Its just a one line change in this pr. Follow up is not required.

snalli (approved):

justinlin-linkedin (approved):

PR #2204, opened 2022-09-06
[frontend] Fix randomness in test

snalli:
This patch fixes a test that expects a composite blob but could
end up getting a simple blob due to wrong random expression.
The test evaluates metadata cache that only works for composite blobs.

### DIFF
--- a/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
@@ -1304,7 +1304,9 @@ public class GetBlobOperationTest {
    */
   @Test
   public void testRouterMetadataCacheFirstChunkOperationFlag() throws Exception {
-    blobSize = random.nextInt(maxChunkSize) + maxChunkSize * random.nextInt(10);
+    // Note that random() can return 0. So we add 1.
+    // This _will_ generate a composite blob.
+    blobSize = maxChunkSize + random.nextInt(maxChunkSize) + 1;
     int randomOne = random.nextInt(blobSize);
     int randomTwo = random.nextInt(blobSize);
     doPut();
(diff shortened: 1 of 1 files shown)


ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -1315,18 +1317,20 @@ public void testRouterMetadataCacheFirstChunkOperationFlag() throws Exception {
         .build(), false, routerMetrics.ageAtGet);
     // First range request gets the entire blob including blobInfo
     GetBlobOperation op1 = getAndAssertSuccess(false, false, (short) 0);
-    assertTrue(String.format("Expected operation flag %s but received %s", MessageFormatFlags.All,
-        op1.getFirstChunkOperationFlag()), op1.getFirstChunkOperationFlag() == MessageFormatFlags.All);
+    assertSame(String.format("Expected operation flag %s but received %s", MessageFormatFlags.All,
+        op1.getFirstChunkOperationFlag()), op1.getFirstChunkOperationFlag(), MessageFormatFlags.All);
```
SophieGuo410, line 1321: You are comparing op1.getFirstChunkOperationFlag() with op1.getFirstChunkOperationFlag() instead of op2.getFirstChunkOperationFlag().
You should use assertEquals. since assertEquals() Asserts that two objects are equal. assertSame() Asserts that two objects refer to the same object.

snalli: done

snalli:
ok fixed the test

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
snalli: done fixed

SophieGuo410 (approved):

PR #2205, opened 2022-09-06
Changes to use helix aggregation view for getting cluster information 

Arun-LinkedIn:
This PR adds changes to use helix local aggregation view cluster for getting global cluster information instead of querying from helix source clusters in individual data centers. This should speed up the start time of helix spectators like frontends and servers.

In addition, this change also refactors following code:
1.  Removes `SimpleClusterChangeHandler` since we support dynamically adding/removing replicas by default now.
2. Removes `CloudServiceClusterChangeHandler` since we don't need cloud code now.
3. Renames `DynamicClusterChangeHandler` to `HelixClusterChangeHandler` and makes it as an inner class of HelixClusterManager so that the shared data structures storing global cluster information can be easily updated and synchronized.


Arun-LinkedIn:
Will publish another PR.

Arun-LinkedIn:
Closing this one. Will break this down into smaller pull requests.

PR #2210, opened 2022-09-10
Fix ConcurrentModificationException during frontend start up

Arun-LinkedIn:
Sample error:

```
Caused by: java.util.ConcurrentModificationException
        at java.util.ArrayList$Itr.checkForComodification(ArrayList.java:1043) ~[?:?]
        at java.util.ArrayList$Itr.next(ArrayList.java:997) ~[?:?]
        at com.github.ambry.router.OperationController.<init>(OperationController.java:127) ~[ambry-router-0.3.834.jar:?]
```

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/OperationController.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/OperationController.java
@@ -40,6 +40,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
@@ -107,13 +108,15 @@ public class OperationController implements Runnable {
     this.responseHandler = responseHandler;
     this.nonBlockingRouter = nonBlockingRouter;
     // Warm up connections to dataNodes in local and remote DCs.
-    List<ResponseInfo> responseInfos = Collections.synchronizedList(new ArrayList<>());
     String localDatacenter = clusterMap.getDatacenterName(clusterMap.getLocalDatacenterId());
     Map<Boolean, List<DataNodeId>> localAndRemoteNodes = clusterMap.getDataNodeIds()
         .stream()
         // ignore any in-process "data nodes" without TCP ports
         .filter(dataNodeId -> dataNodeId.getPort() != DataNodeId.UNKNOWN_PORT)
         .collect(Collectors.partitioningBy(dataNodeId -> localDatacenter.equals(dataNodeId.getDatacenterName())));
+    // 'responseInfos' list needs to be thread safe since it it possible that the list can be modified as we are reading
+    // it for scenario where we returned from networkClient.warmUpConnections() due to time out.
+    List<ResponseInfo> responseInfos = new CopyOnWriteArrayList<>();
     logger.info("Warming up local datacenter connections to {} nodes. Connections warmup percentage: {}%.",
         localAndRemoteNodes.get(true).size(), routerConfig.routerConnectionsLocalDcWarmUpPercentage);
     networkClient.warmUpConnections(localAndRemoteNodes.get(true),
--- a/ambry-router/src/main/java/com/github/ambry/router/OperationController.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/OperationController.java
@@ -108,15 +108,13 @@ public class OperationController implements Runnable {
     this.responseHandler = responseHandler;
     this.nonBlockingRouter = nonBlockingRouter;
     // Warm up connections to dataNodes in local and remote DCs.
+    List<ResponseInfo> responseInfos = Collections.synchronizedList(new ArrayList<>());
     String localDatacenter = clusterMap.getDatacenterName(clusterMap.getLocalDatacenterId());
     Map<Boolean, List<DataNodeId>> localAndRemoteNodes = clusterMap.getDataNodeIds()
         .stream()
         // ignore any in-process "data nodes" without TCP ports
         .filter(dataNodeId -> dataNodeId.getPort() != DataNodeId.UNKNOWN_PORT)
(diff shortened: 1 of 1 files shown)


ambry-router/src/main/java/com/github/ambry/router/OperationController.java
```diff
@@ -106,13 +107,15 @@ public OperationController(String suffix, String defaultPartitionClass, AccountS
...
-    List<ResponseInfo> responseInfos = Collections.synchronizedList(new ArrayList<>());
     String localDatacenter = clusterMap.getDatacenterName(clusterMap.getLocalDatacenterId());
     Map<Boolean, List<DataNodeId>> localAndRemoteNodes = clusterMap.getDataNodeIds()
         .stream()
         // ignore any in-process "data nodes" without TCP ports
         .filter(dataNodeId -> dataNodeId.getPort() != DataNodeId.UNKNOWN_PORT)
         .collect(Collectors.partitioningBy(dataNodeId -> localDatacenter.equals(dataNodeId.getDatacenterName())));
+    // 'responseInfos' list needs to be thread safe since it it possible that the list can be modified as we are reading
+    // it for scenario where we returned from networkClient.warmUpConnections() due to time out.
+    List<ResponseInfo> responseInfos = new CopyOnWriteArrayList<>();
```
justinlin-linkedin, line 118: maybe just use forEach to replace the for loop in line 127. for loop would return a iterator, which is not protected by mutex and requires locking by users. But forEach is protected by the mutex in the SynchronizedList. This way, we won't see the ConcurrentModificationException.

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
```diff
@@ -106,13 +107,15 @@ public OperationController(String suffix, String defaultPartitionClass, AccountS
     this.responseHandler = responseHandler;
     this.nonBlockingRouter = nonBlockingRouter;
     // Warm up connections to dataNodes in local and remote DCs.
-    List<ResponseInfo> responseInfos = Collections.synchronizedList(new ArrayList<>());
```
snalli, line 109: What changed that is causing this exception now ?

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
snalli, line 118: COW array is not a good idea. It just creates a new array on each write and consumes more memory. 
Does synchronizedList not guard the array using a lock ?

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
```diff
@@ -39,6 +39,7 @@
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.CopyOnWriteArrayList;
```
snalli, line 42: Where is concurrent write happening on L127 ? It appears there are only reads on the array in question

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
snalli: +1

snalli, line 118: What about other methods such as using a synchronized block, or waiting till all writer-threads have completed before iterating the list ?

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
snalli: Nevermind, found the writes

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
snalli: nvm, i found that syncList is not thread-safe when iterating and a concurrent write (!). COW is ok iff the list is small and there are a few writes.

snalli: nvm, they may not work since the writes are scattered across code.

snalli: nvm, i think cow is better in this case iff array is small and writes are few. forEach is limited in the things it can do.

https://stackoverflow.com/questions/16635398/java-8-iterable-foreach-vs-foreach-loop

Arun-LinkedIn: Thank you for the suggestions. For now, used forEach with SynchronizedList() since there is not too much logic in the block and no exceptions to handle etc. 

Since this list is used to keep track of connection failures during warming up connections from frontends to servers, we can probably have more writes if there are lot of server errors and CopyOnWriteArrayList may be sub-optimal. For example, in below, there seem to be around 733 failures/writes during connection warm up.

// local datacenter connection warm up
2022/09/09 19:26:40.290 INFO [Http2NetworkClient] [main] [ambry-frontend] [] HTTP2 connection warm up done. Tried: 2628, Succeeded: 2046, Failed: 549, Time elapsed: 5251 ms
// remote datacenter connection warm up
2022/09/09 19:26:45.475 INFO [Http2NetworkClient] [main] [ambry-frontend] [] HTTP2 connection warm up done. Tried: 930, Succeeded: 743, Failed: 184, Time elapsed: 5185 ms

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
Arun-LinkedIn: This seems to have been present always. We only see if we return from `networkClient.warmUpConnections` on timeout (5 seconds) and the `responseInfo` List is updated by Netty threads as we try to iterate over it. Due to that, the probability might have been less.

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
justinlin-linkedin, line 42: nit: no used anywhere.

justinlin-linkedin (approved):

snalli (approved):

ambry-router/src/main/java/com/github/ambry/router/OperationController.java
snalli: had an offline discussion with arun to discuss pros and cons of foreach vs cow

snalli:
rebase on master to disable the large blob test

PR #2213, opened 2022-09-13
Add a new FindMissingKey implementation to make it faster

justinlin-linkedin:
In ReplicaThread, when we get a list of MessageInfos from the source host, we call findMissingKeys to see if those keys are missing in the local store. However, this is really slow since for each key, we have to do a search on all the index files. But those keys should be pretty close to each other, so they might be stored in the same index segment file or file that are close to the one that has the first key.  In same extreme cases, those keys returned in the same ReplicaMetadataRequest all belongs to one index segment. This is true when the replication token is reset by compaction in source host and now we are replicating from the beginning of the log again.

We should leverage the replication key locality to make replication faster.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -470,6 +470,26 @@ public class StoreConfig {
   @Default("200")
   public final int storeDiskIoReservoirTimeWindowMs;
 
+  /**
+   * True to enable batch mode in findMissingKeys
+   */
+  @Config(storeEnableFindMissingKeysInBatchModeName)
+  @Default("false")
+  public final boolean storeEnableFindMissingKeysInBatchMode;
+  public static final String storeEnableFindMissingKeysInBatchModeName = "store.enable.find.missing.keys.in.batch.mode";
+
+  @Config(storeCacheSizeForFindMissingKeysInBatchModeName)
+  @Default("3")
+  public final int storeCacheSizeForFindMissingKeysInBatchMode;
+  public static final String storeCacheSizeForFindMissingKeysInBatchModeName =
+      "store.cache.size.for.find.missing.keys.in.batch.mode";
+
+  @Config(storeNumOfCacheMissForFindMissingKeysInBatchModeName)
+  @Default("5")
+  public final int storeNumOfCacheMissForFindMissingKeysInBatchMode;
+  public static final String storeNumOfCacheMissForFindMissingKeysInBatchModeName =
+      "store.num.of.cache.miss.for.find.missing.keys.in.batch.mode";
+
   public StoreConfig(VerifiableProperties verifiableProperties) {
 
     storeKeyFactory = verifiableProperties.getString("store.key.factory", "com.github.ambry.commons.BlobIdFactory");
@@ -583,5 +603,11 @@ public class StoreConfig {
             Integer.MAX_VALUE);
     storeDiskIoReservoirTimeWindowMs =
         verifiableProperties.getIntInRange("store.disk.io.reservoir.time.window.ms", 200, 0, Integer.MAX_VALUE);
+    storeEnableFindMissingKeysInBatchMode =
+        verifiableProperties.getBoolean(storeEnableFindMissingKeysInBatchModeName, false);
+    storeCacheSizeForFindMissingKeysInBatchMode =
+        verifiableProperties.getIntInRange(storeCacheSizeForFindMissingKeysInBatchModeName, 3, 1, 100);
+    storeNumOfCacheMissForFindMissingKeysInBatchMode =
+        verifiableProperties.getIntInRange(storeNumOfCacheMissForFindMissingKeysInBatchModeName, 5, 3, 100);
   }
 }
--- a/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
@@ -461,6 +461,29 @@ class IndexSegment implements Iterable<IndexEntry> {
   }
 
   /**
+   * Return all the keys in the IndexSegment in a set.
+   * @return
+   */
+  Set<StoreKey> getAllStoreKeys() throws StoreException {
+    Set<StoreKey> keys = new HashSet<>();
+    rwLock.readLock().lock();
+    boolean isSealed = sealed.get();
+    NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> indexCopy = index;
+    ByteBuffer buffer = serEntries;
+    rwLock.readLock().unlock();
+    if (isSealed) {
+      buffer = buffer.duplicate();
+      int numOfIndexEntries = numberOfEntries(buffer);
+      for (int i = 0; i < numOfIndexEntries; i++) {
+        keys.add(getKeyAt(buffer, i));
+      }
+    } else {
+      keys.addAll(indexCopy.keySet());
+    }
+    return keys;
+  }
+
+  /**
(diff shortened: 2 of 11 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1298,9 +1306,189 @@ Set<StoreKey> findMissingKeys(List<StoreKey> keys) throws StoreException {
...
+      NavigableSet<IndexValue> values = nextSegment.find(key);
+      if (values == null || values.isEmpty()) {
+        shouldUseCachedKeys = findKeyAndAddIndexSegmentToCache(key, missingKeys);
+      } else {
+        if (nextSegment.isSealed()) {
+          Set<StoreKey> allStoreKeys = nextSegment.getAllStoreKeys();
+          // Adding this index segment to the front of the list
+          logger.info("Index {}: In findMissingKeyInbBatch, adding next segment : {}", dataDir,
+              nextSegment.getStartOffset());
+          cachedKeys.addFirst(new Pair<>(nextSegment.getStartOffset(), allStoreKeys));
```
SophieGuo410, line 1467: QQ: one thing confused me a little bit is, if we can find the key in next index segment, we add it to the cache, but we are not sure if the next segment has the final state of the key. But findKeyAndAddIndexSegmentToCache will always find the key's final state and add corresponding index segment. So can you explain a little bit about when do we need to find the final state?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1298,9 +1306,189 @@ Set<StoreKey> findMissingKeys(List<StoreKey> keys) throws StoreException {
...
+        // Adding this index segment to the front of the list
+        logger.trace("Index {}: In findMissingKeysInbBatch, adding new segment : {}", dataDir,
+            segment.getStartOffset());
+        cachedKeys.addFirst(new Pair<>(entry.getKey(), allStoreKeys));
+        if (cachedKeys.size() > config.storeCacheSizeForFindMissingKeysInBatchMode) {
+          cachedKeys.removeLast();
+        }
+      } else {
+        logger.trace("Index {}: In findMissingKeysInbBatch, hitting unsealed index segment, stop using cache : {}",
+            dataDir, segment.getStartOffset());
```
SophieGuo410, line 1421: if segment == null, it will through NullPointException.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -460,6 +460,29 @@ NavigableSet<IndexValue> find(StoreKey keyToFind) throws StoreException {
...
 
+  /**
+   * Return all the keys in the IndexSegment in a set.
+   * @return
+   */
+  Set<StoreKey> getAllStoreKeys() throws StoreException {
+    Set<StoreKey> keys = new HashSet<>();
+    rwLock.readLock().lock();
+    boolean isSealed = sealed.get();
+    NavigableMap<StoreKey, ConcurrentSkipListSet<IndexValue>> indexCopy = index;
```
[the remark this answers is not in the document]
justinlin-linkedin: it's fine, in the end, we don't really copy the keys in unsealed index segment.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: The reason why we are using findKey is not to find the final state, it's just a tool to find any segment that contains this key. The ultimate goal is to check if this key exists or not, we don't really care about the final state. But I don't have better tool to use than just calling findKey to the keys missing from the cache.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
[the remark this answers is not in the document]
justinlin-linkedin: you are right, updated.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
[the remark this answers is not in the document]
SophieGuo410: So which means we could use index directly right? Just don't get why do we need to copy the reference?  The indexCopy would have same value as index since it's just copy the reference.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1298,9 +1306,205 @@ Set<StoreKey> findMissingKeys(List<StoreKey> keys) throws StoreException {
...
+
+    // This method should only be invoked in one replication thread, but to make sure there is no race condition, we use
+    // a lock to protect it.
+
+    missingKeysLock.lock();
+    try {
+      for (int i = 0; i < keys.size(); i++) {
+        StoreKey key = keys.get(i);
+        if (!shouldUseCachedKeys) {
+          // if the cache is disabled, then just use a regular findKey method to check the existence of the key
```
Arun-LinkedIn, line 1353: nit: To avoid this check, may be we could also break from for loop and iterate in another loop when we find cache is disabled.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -159,6 +163,10 @@ public int compare(IndexEntry e1, IndexEntry e2) {
 
   private Runnable getBlobReadInfoTestCallback = null;
 
+  // The lock to protect cachedKeys in the findMissingKeysInBatch
+  private final Lock missingKeysLock = new ReentrantLock();
+  private LinkedList<Pair<Offset, Set<StoreKey>>> cachedKeys = new LinkedList<>();
```
Arun-LinkedIn, line 168: Like we were discussing offline, may be we might need separate cached key set for each remote replica since each of them might be at different replication sync offset.

Arun-LinkedIn (commented):
Thank you for explaining the use case offline. Mostly looks good.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -470,6 +470,26 @@ public class StoreConfig {
   @Default("200")
   public final int storeDiskIoReservoirTimeWindowMs;
 
+  /**
+   * True to enable batch mode in findMissingKeys
+   */
+  @Config(storeEnableFindMissingKeysInBatchModeName)
+  @Default("false")
+  public final boolean storeEnableFindMissingKeysInBatchMode;
+  public static final String storeEnableFindMissingKeysInBatchModeName = "store.enable.find.missing.keys.in.batch.mode";
```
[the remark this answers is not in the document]
justinlin-linkedin: this is the static configuration name, the configuration variable name doesn't have "Name" in the end.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: Updated.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: Updated

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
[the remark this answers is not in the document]
justinlin-linkedin: the index might be set to null due to the index segment becoming sealed. So in the lock, we have to obtain a copy of reference to the index.

Arun-LinkedIn (approved):

SophieGuo410 (approved):

PR #2216, opened 2022-09-17
Optimize Updates: Ensure blobInfo is set for encrypted blobs' getChunkIdsOnly option response.

ankagrawal:
Due to this for encrypted blobs, update operations were not getting optimized.

Also, Add metric for cases where update operation for already updated blob is not optimized.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -62,6 +62,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -1532,9 +1533,15 @@ class GetBlobOperation extends GetOperation {
           getOptions().ageAtAccessTracker.trackAgeAtAccess(serverBlobInfo.getBlobProperties().getCreationTimeInMs());
           blobData = blobAll.getBlobData();
           encryptionKey = blobAll.getBlobEncryptionKey();
-          if (encryptionKey == null) {
-            // set blobInfo only if decryption is not required. If not, maybeProcessCallbackAndComplete() will set appropriate
-            // value
+          if (Objects.isNull(encryptionKey) || options.getChunkIdsOnly) {
+            // set blobInfo iff:
+            //   1. blob is not encrypted OR,
+            //   2. getChunkIdsOnly is set.
+            // If the blob is encrypted, then the user metadata is encrypted too. So in that case,
+            // maybeProcessCallbackAndComplete() will set the appropriate value for blobInfo after decryption.
+            // In the case of getChunkIdsOnly call, we don't need the user metadata but still need blob
+            // properties to evaluate if the metadata chunk is already updated for update requests. So we allow blobInfo
+            // to be set here.
             blobInfo = serverBlobInfo;
             blobInfo.setLifeVersion(messageInfo.getLifeVersion());
           } else {
--- a/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterMetrics.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterMetrics.java
@@ -186,6 +186,7 @@ public class NonBlockingRouterMetrics {
   public final Counter compositeBlobSizeMismatchCount;
   public final Counter unknownPartitionClassCount;
   public final Counter updateOptimizedCount;
+  public final Counter updateUnOptimizedCount;
   // Number of unnecessary blob gets avoided via use of BlobDataType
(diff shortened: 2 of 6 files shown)


snalli:
Try to keep the commit title small something like
[frontend] Set blobInfo for getChunkIdsOnly on encrypted blob

Helps with one-liner git show.
You may elaborate in the description.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1532,9 +1532,12 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
...
           encryptionKey = blobAll.getBlobEncryptionKey();
-          if (encryptionKey == null) {
-            // set blobInfo only if decryption is not required. If not, maybeProcessCallbackAndComplete() will set appropriate
-            // value
+          if (encryptionKey == null || options.getChunkIdsOnly) {
+            // set blobInfo only if decryption is not required, or if its a chunk id only request. This is because
+            // blobInfo has user metadata which is encrypted for encrypted blobs. When decryption is required,
+            // maybeProcessCallbackAndComplete() will set appropriate value after decryption.
+            // Note that in case of getChunkIdsOnly call, we don't need the user metadata but still might need blob
+            // properties.
```
SophieGuo410, line 1540: nit: It's better to include the reason of why we need the blob properties for future reference.

SophieGuo410 (approved):

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1532,9 +1532,12 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
           getOptions().ageAtAccessTracker.trackAgeAtAccess(serverBlobInfo.getBlobProperties().getCreationTimeInMs());
           blobData = blobAll.getBlobData();
           encryptionKey = blobAll.getBlobEncryptionKey();
-          if (encryptionKey == null) {
-            // set blobInfo only if decryption is not required. If not, maybeProcessCallbackAndComplete() will set appropriate
-            // value
+          if (encryptionKey == null || options.getChunkIdsOnly) {
```
snalli, line 1535: this comment is unclear.
this case handles blobs that are not encrypted.
why describe what happens when the blob is encrypted ?
Also the code change doesn't match the commit title

snalli, line 1535: can u use a helper like isEncrypted() { return enc_key != null } ?

ankagrawal: An example of a case it doesn't handle is "blob is encrypted, but its a getBlob call without getChunkIdsOnly set". In this case, for example, both `encryptionKey == null` will evaluate to false, and `option.getChunkIdsOnly` will evaluate to false.

ankagrawal: Changed it to `Objects.isNull`. I believe that `enc_key == null` check to identify an unencrypted blob, works in this case when we are looking at the deserialized blob we just read from the server. It might not apply as a sufficient condition for a generic `isEncrypted()` in all cases. So I am trying to avoid a method like `isEncrypted` for this check.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterMetrics.java
```diff
@@ -481,7 +482,9 @@ public NonBlockingRouterMetrics(ClusterMap clusterMap, RouterConfig routerConfig
     unknownPartitionClassCount =
         metricRegistry.counter(MetricRegistry.name(PutOperation.class, "UnknownPartitionClassCount"));
     updateOptimizedCount =
-        metricRegistry.counter(MetricRegistry.name(OperationController.class, "updateOptimizedCount"));
+        metricRegistry.counter(MetricRegistry.name(OperationController.class, "UpdateOptimizedCount"));
+    updateNotOptimizedCount =
+        metricRegistry.counter(MetricRegistry.name(OperationController.class, "UpdateNotOptimizedCount"));
```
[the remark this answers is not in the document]
ankagrawal: Looks like both convey the same meaning with the same clarity. But let me know if you are really worried, and I can make the name change.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
snalli: ok so a better comment might be

`
set blobInfo iff 
1. blob is not encrypted or,
2. getChunkIdsOnly is set since we need blobProperties member of blobInfo.
 
Else, maybeProcessCallbackAndComplete() will set it after decryption.
`

snalli (approved):

snalli (approved):

PR #2217, opened 2022-09-20
Temporarily return 404 when two replicas not found but one succeed

SophieGuo410:
There’s a case that due to some reason two replicas don’t have the data(maybe due to OSUA wiping out some data), but the third one still has the DELETE records. 
In current scenario, we return 503 which is not appropriate.
This is the short term fix to revert the changes to the previous version which return 404.
For the long term fix, we should check all replica status for delete operation and return succeed(200) when one replica succeed.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -122,8 +122,8 @@ public class RouterConfig {
   public static final String ROUTER_NOT_FOUND_CACHE_TTL_IN_MS = "router.not.found.cache.ttl.in.ms";
   public static final String ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS =
       "router.update.op.metadata.reliance.timestamp.in.ms";
-  public static final String ROUTER_UNAVAILABLE_DUE_TO_SUCCESS_COUNT_IS_NON_ZERO =
-      "router.unavailable.due.to.success.count.is.non.zero";
+  public static final String ROUTER_UNAVAILABLE_DUE_TO_SUCCESS_COUNT_IS_NON_ZERO_FOR_DELETE =
+      "router.unavailable.due.to.success.count.is.non.zero.for.delete";
 
   /**
    * Number of independent scaling units for the router.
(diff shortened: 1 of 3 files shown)


ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -361,21 +361,10 @@ private void checkAndMaybeComplete() {
...
-              new RouterException("DeleteOperation failed possibly because of unavailable replicas",
-                  RouterErrorCode.AmbryUnavailable));
-        } else {
-          operationException.set(
-              new RouterException("DeleteOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
-        }
+        //TODO: Temporarily return 404 when delete operation returns two NOT_FOUND.
+        //TODO: Will check all local replicas and return succeed if one replica return succeed in this case.
+        operationException.set(
+            new RouterException("DeleteOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
```
justinlin-linkedin, line 367: let's create a configuration for this and only turn it on for delivery cluster to return 503.

ankagrawal: What's the problem with returning 503 in this case? If for delete, the requirement of quorum is that there should be 2 successes, then why not let the user retry for eventual success?
Getting a 503 error from Ambry means that due to some replica unavailability, Ambry doesn't know what to do with the request, So please retry later.

justinlin-linkedin: Delete is different from updateTTL. For updateTtl, when you get NOT_FOUND from a replica, you can wait until this replica replicating the PUT back. But for delete, replication won't solve the problem. For example. replica 1 has the PUT and DELETE, but replica 2 and replica 3 somehow had data wipeout in the past and they replicated from replica 1. However, when they replicate, the would see this blob id as deleted so they would total bypass this blob id. This would result in having no records for this blob id in replica 2 and 3 when replica 1 has the PUT and DELETE. This issue can't be solved waiting for more replication.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -578,6 +580,13 @@ public class RouterConfig {
   @Default("false")
   public final boolean routerUnavailableDueToOfflineReplicas;
 
+  /**
+   * If true the simple operation tracker will check if there's one replica return success, router will return unavailable error.
+   */
+  @Config(ROUTER_UNAVAILABLE_DUE_TO_SUCCESS_COUNT_IS_NON_ZERO)
+  @Default("true")
+  public final boolean routerUnavailableDueToSuccessCountIsNonZero;
```
justinlin-linkedin, line 588: can we add "for.delete" in the name?

justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
SophieGuo410: Updated.

ankagrawal (approved):

Arun-LinkedIn (approved):

PR #2222, opened 2022-09-22
[frontend] Remove unnecessary if-clause in PUT

snalli:
Remove unnecessary if-clause in PutOperation

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -1184,21 +1184,11 @@ class PutOperation {
     private void prepareForSending() {
       try {
         // Determine data type to set
-        BlobDataType blobDataType = null;
-        if (isMetadataChunk()) {
-          blobDataType = BlobDataType.METADATA;
-        } else if (chunkIndex == 0) {
-          // TODO: need more data to distinguish between Simple and Data Chunk
-          blobDataType = BlobDataType.DATACHUNK;
-        } else {
-          blobDataType = BlobDataType.DATACHUNK;
-        }
+        BlobDataType blobDataType = isMetadataChunk() ? BlobDataType.METADATA : BlobDataType.DATACHUNK;
 
-        // if this is part of a retry, make sure no previously attempted partitions are retried.
-        if (partitionId != null) {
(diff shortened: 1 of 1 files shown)


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1184,21 +1184,11 @@ private void prepareForBuilding(int chunkIndex) {
...
-        } else if (chunkIndex == 0) {
-          // TODO: need more data to distinguish between Simple and Data Chunk
-          blobDataType = BlobDataType.DATACHUNK;
-        } else {
-          blobDataType = BlobDataType.DATACHUNK;
-        }
+        BlobDataType blobDataType = isMetadataChunk() ? BlobDataType.METADATA : BlobDataType.DATACHUNK;
 
-        // if this is part of a retry, make sure no previously attempted partitions are retried.
-        if (partitionId != null) {
```
ankagrawal, line 1198: Why remove this check and add to `attemptedPartitionIds`?

snalli: code with less if-clauses is just flows and easier to read.
The old code and this one are equivalent.
`attemptedPartitionIds` is not used anywhere else.
In this new code, if we fail and retry, we already have recorded the prev partition and will exclude it. Same as old code. Although if we succeed, the size of attemptedPartitionIds will be 1 in this case and 0 in old code.

snalli: Not neccesary to merge but something i noticed while reading the code today.

ankagrawal: I must be missing something -- Why do you say that `attemptedPartitionIds` is not used anywhere? Its used in the immediately next line after the line you deleted.

> we already have recorded the prev partition and will exclude it

How and where have we recorded the previous partition and where's the code to exclude it?

ankagrawal: I see that you added the `attemptedPartitionIds` in the next line.

ankagrawal (approved):

ankagrawal (approved):

Arun-LinkedIn (approved):

PR #2227, opened 2022-09-26
Compression Config

panchow:
- Configuration settings used by compression.
- Minor updates to Utils and CompressionMap
- Add more unit tests


ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
+/*
+ * Copyright 2017 LinkedIn Corp. All rights reserved.
```
snalli, line 2: 2022

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+
+import com.github.ambry.utils.Utils;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+
+/**
+ * Configuration parameters required by a CompressionService.
+ * Configurations are read-only.  The only way to alter the values is by modifying the config files.
+ * The configuration applies to compression only.  Decompression is automatic and has no configuration.
```
snalli, line 24: if there is no compression, will there be decompression ?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+ */
+public class CompressionConfig {
+
+  // Boolean whether compression is enabled in PUT operation.
+  private static final String CONFIG_COMPRESSION_ENABLED = "router.compression.enabled";
+  private static final boolean DEFAULT_COMPRESSION_ENABLED = true;
+
+  // Whether to skip compression if content-encoding present.
+  private static final String CONFIG_SKIP_CONTENT_ENCODING = "router.compression.skip.content.encoding";
```
snalli, line 33: is this to skip encoding or skip compression when encoding present ?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  private static final boolean DEFAULT_SKIP_CONTENT_ENCODING = true;
+
+  // A comma-separated list of compressible fully content-types, like "text/javascript, text/log; Charset=UTF8"
+  private static final String CONFIG_CONTENT_TYPES_COMPRESSIBLE = "router.compression.content.types.compressible";
+  private static final String DEFAULT_CONTENT_TYPES_COMPRESSIBLE = null;
+
+  // A comma-separated list of incompressible fully content-types, like "text/javascript, text/log; Charset=UTF8"
+  private static final String CONFIG_CONTENT_TYPES_INCOMPRESSIBLE = "router.compression.content.types.incompressible";
```
snalli, line 41: what happens if text/javascript is present in both configs - compressible and incompressible ?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  private static final String CONFIG_CONTENT_PREFIXES_COMPRESSIBLE = "router.compression.content.prefixes.compressible";
+  private static final String DEFAULT_CONTENT_PREFIXES_COMPRESSIBLE = "text";
+
+  // A comma-separated list of incompressible content-prefixes, such as "image" for image/*.
+  private static final String CONFIG_CONTENT_PREFIXES_INCOMPRESSIBLE = "router.compression.content.prefixes.incompressible";
+  private static final String DEFAULT_CONTENT_PREFIXES_INCOMPRESSIBLE = "image";
+
+  // Boolean whether to compress other content-types not specified in content-types and content-prefixes.
+  // It applies to PUT operation only.
+  private static final String CONFIG_COMPRESS_OTHER_CONTENT_TYPE = "router.compression.other.content.types";
```
snalli, line 54: what is an example of other content type ? If there some content type that can be compressed, why can't we just include it in the config in L37?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  private static final boolean DEFAULT_COMPRESS_OTHER_CONTENT_TYPE = false;
+
+  // The minimal source/chunk size in bytes in order to qualify for compression.
+  private static final String CONFIG_MINIMAL_DATA_SIZE_IN_BYTES = "router.compression.minimal.content.size";
+  private static final int DEFAULT_MINIMAL_DATA_SIZE_IN_BYTES = 1024;
+
+  // The minimal compression ratio (uncompressedSize/compressSize) in order to keep compressed content.
+  // The higher the compression ratio the better with 1 means the same size.
+  private static final String CONFIG_MINIMAL_COMPRESS_RATIO = "router.compression.minimal.ratio";
```
snalli, line 63: how does this ratio help ? what if its not met ?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  /**
+   * Combine the compressible and incompressible list of content-types into a map of content-type to boolean.
+   *
+   * @param compressibleContentTypes List of compressible content-types.
+   * @param incompressibleContentTypes List of incompressible content-types.
+   * @return A map from content-type to boolean (compressible or incompressible).
+   */
+  private Map<String, Boolean> parseContentTypes(String compressibleContentTypes, String incompressibleContentTypes) {
+    Map<String, Boolean> result = new HashMap<>();
+    addContentTypesToMap(result, compressibleContentTypes, true);
```
snalli, line 175: some check to ensure they are mutually exclusive ?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  private static final String CONFIG_MINIMAL_DATA_SIZE_IN_BYTES = "router.compression.minimal.content.size";
+  private static final int DEFAULT_MINIMAL_DATA_SIZE_IN_BYTES = 1024;
+
+  // The minimal compression ratio (uncompressedSize/compressSize) in order to keep compressed content.
+  // The higher the compression ratio the better with 1 means the same size.
+  private static final String CONFIG_MINIMAL_COMPRESS_RATIO = "router.compression.minimal.ratio";
+  private static final float DEFAULT_MINIMAL_COMPRESS_RATIO = 1.1f;
+
+  // The name of the algorithm to use.  This name must be pre-registered in CompressionMap class.
+  private static final String CONFIG_ALGORITHM_NAME = "router.compression.algorithm.name";
```
ankagrawal, line 67: So does this mean that everytime we try to compress data, we check if the compression ratio >= "router.compression.minimal.ratio"? If not, we save the data uncompressed?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  private static final String DEFAULT_ALGORITHM_NAME = null;
+
+  /**
+   * Whether compression is enabled.
+   */
+  public boolean isCompressionEnabled;
```
ankagrawal, line 73: For each of the configs, we should use the decorators @Config and @Default (default is optional). For example, in CloudConfig.java we have

```
  @Config(CLOUD_DESTINATION_FACTORY_CLASS)
  @Default(DEFAULT_CLOUD_DESTINATION_FACTORY_CLASS)
  ```

File: ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java [hunk 10]
```diff
@@ -0,0 +1,289 @@
...
+   * Whether compression is skipped when the content-encoding is present in the request.
+   */
+  public boolean isSkipWithContentEncoding;
+
+  /**
+   * The map contains list of content-types and whether they are compressible.
+   * The map key (content type) must be lower-case only.
+   * The map value should either be true for compressible content-type or false for incompressible content-type.
+   */
+  public Map<String, Boolean> compressContentTypes;
```
ankagrawal, line 85: The @Config and @Default decorators comment apply to all the configs in this file.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
snalli: I mean shouldn't the same config control both compression and decompression ?

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
snalli: +1

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Compression will store the algorithm name inside the binary, so that decompression can lookup the algorithm by this name.  That's why the compression config is different from decompression.  There aren't that many decompression config.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: That is correct.  If encoding is present, skip compression.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: If same setting is in both incompressible and compressible, incompressible takes priority because the map from content-type to boolean adds compressible content-type first, followed by incompressible content-types.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Other content-type is for content-type not specified in compressible nor incompressible list.  Content-type is technically free text.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: In this case, there is no need for exclusivity.  Instead it's import to specify the order the rules are applied.  As described in comment, the order is:
- full content-type check.
- content-type prefix check.
- other content-type setting.

For example, what if I want to block everything image/* except image/bmp?  Then add "image" to incompressible prefix, and add "image/bmp" to compressible content-type.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  // A comma-separated list of compressible fully content-types, like "text/javascript, text/log; Charset=UTF8"
+  private static final String CONFIG_CONTENT_TYPES_COMPRESSIBLE = "router.compression.content.types.compressible";
+  private static final String DEFAULT_CONTENT_TYPES_COMPRESSIBLE = null;
+
+  // A comma-separated list of incompressible fully content-types, like "text/javascript, text/log; Charset=UTF8"
+  private static final String CONFIG_CONTENT_TYPES_INCOMPRESSIBLE = "router.compression.content.types.incompressible";
+  private static final String DEFAULT_CONTENT_TYPES_INCOMPRESSIBLE = null;
+
+  // A comma-separated list of compressible content-prefixes, such as "text" for text/*
+  private static final String CONFIG_CONTENT_PREFIXES_COMPRESSIBLE = "router.compression.content.prefixes.compressible";
```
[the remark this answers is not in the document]
panchow: It is necessary to support prefix.  For example, most images are incompressible.  Instead of explicitly listing all image content-types, just use "image/*" to exclude them all.

One option is combine prefix list and full content-type list, but can config distinguish which is prefix and which is not?  This will require prefix to use "/*".  The other option is to separate them.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,289 @@
...
+  private static final String DEFAULT_CONTENT_PREFIXES_COMPRESSIBLE = "text";
+
+  // A comma-separated list of incompressible content-prefixes, such as "image" for image/*.
+  private static final String CONFIG_CONTENT_PREFIXES_INCOMPRESSIBLE = "router.compression.content.prefixes.incompressible";
```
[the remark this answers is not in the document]
panchow: See if the answer makes sense.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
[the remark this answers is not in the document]
panchow: This is for other content-types.  We can't possibly list all content-types in both the full content-type or prefix.  If the content-type doesn't fall to any of the list, should we compress the chunk?  The answer falls to this "other content-type" setting.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Need to talk to you about how these two annotations are used.  I looked around the code and couldn't find how it is used.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Yes, if the compression has too little value, don't waste the effort of decompression.  So if the compression result is insignificant (incompressible content), then don't use compression.  Just store the blob as raw binary.

For a simple blob, the compression ratio determines whether to keep the compressed data or discard compression and store raw.  For composite blobs, the metadata chunk isn't compressed.  The first data chunk is compressed and check its compression ratio.  If it's good, the compressed data is kept and all other chunks will be compressed.  If the first chunk is incompressible, compressed data will be discarded and all other chunks will not be compressed.   The reason for this behavior is due to there is only 1 blob properties per blob, not per chunk.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,319 @@
...
+
+/**
+ * Configuration parameters required by a CompressionService.
+ * Configurations are read-only.  The only way to alter the values is by modifying the config files.
+ * The configuration applies to compression only.  Decompression is automatic and has no configuration.
+ */
+public class CompressionConfig {
+
+  // Boolean whether compression is enabled in PUT operation.
+  private static final String CONFIG_COMPRESSION_ENABLED = "router.compression.enabled";
```
justinlin-linkedin, line 29: configuration name can be public, it would be easier for unit tests if they are public. In unit tests, we often create a new Config based on certain options. If these are public, we can directly use these names in the unit tests.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,319 @@
...
+  private static final boolean DEFAULT_COMPRESSION_ENABLED = true;
```
justinlin-linkedin, line 30: the default value should be false. We shouldn't turn this new feature on before the GetBlobOperation and PutBlobOperation both have the support for compression.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,319 @@
...
+
+  // Whether to skip compression if content-encoding present.
+  private static final String CONFIG_SKIP_CONTENT_ENCODING = "router.compression.skip.content.encoding";
```
justinlin-linkedin, line 33: I can't think of any cases we would need to compress when the content-encoding is present. I don't think this is necessary, but having it won't hurt.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,319 @@
...
+  private static final boolean DEFAULT_SKIP_CONTENT_ENCODING = true;
+
+  // A comma-separated list of compressible fully content-types, like "text/javascript, text/log; Charset=UTF8"
```
justinlin-linkedin, line 36: Bad example, we don't really care about Charset, and Ambry is probably not going to work with charset in the content-type in some cases.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,319 @@
...
+  private static final String CONFIG_CONTENT_TYPES_INCOMPRESSIBLE = "router.compression.content.types.incompressible";
+  private static final String DEFAULT_CONTENT_TYPES_INCOMPRESSIBLE = null;
+
+  // A comma-separated list of compressible content-prefixes, such as "text" for text/*
+  private static final String CONFIG_CONTENT_PREFIXES_COMPRESSIBLE = "router.compression.content.prefixes.compressible";
+  private static final String DEFAULT_CONTENT_PREFIXES_COMPRESSIBLE = "text";
+
+  // A comma-separated list of incompressible content-prefixes, such as "image" for image/*.
+  private static final String CONFIG_CONTENT_PREFIXES_INCOMPRESSIBLE = "router.compression.content.prefixes.incompressible";
+  private static final String DEFAULT_CONTENT_PREFIXES_INCOMPRESSIBLE = "image";
```
justinlin-linkedin, line 50: I am afraid these configs are making it complicated. Now we have many configs to control whether we are going to compress certain content types. I suggest we just have one config, "router.compression.content.types.compressible" and only compress the blob if the content type is in this list.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,319 @@
...
+  // A comma-separated list of compressible fully content-types, like "text/javascript, text/log; Charset=UTF8"
+  private static final String CONFIG_CONTENT_TYPES_COMPRESSIBLE = "router.compression.content.types.compressible";
+  private static final String DEFAULT_CONTENT_TYPES_COMPRESSIBLE = null;
+
+  // A comma-separated list of incompressible fully content-types, like "text/javascript, text/log; Charset=UTF8"
+  private static final String CONFIG_CONTENT_TYPES_INCOMPRESSIBLE = "router.compression.content.types.incompressible";
+  private static final String DEFAULT_CONTENT_TYPES_INCOMPRESSIBLE = null;
+
+  // A comma-separated list of compressible content-prefixes, such as "text" for text/*
+  private static final String CONFIG_CONTENT_PREFIXES_COMPRESSIBLE = "router.compression.content.prefixes.compressible";
```
justinlin-linkedin, line 45: prefix would be merged withe the router.compression.content.types.compressible since prefix is just the major type in the mime type.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,319 @@
...
+  private static final int DEFAULT_MINIMAL_DATA_SIZE_IN_BYTES = 1024;
+
+  // The minimal compression ratio (uncompressedSize/compressSize) in order to keep compressed content.
+  // The higher the compression ratio the better with 1 means the same size.
+  private static final String CONFIG_MINIMAL_COMPRESS_RATIO = "router.compression.minimal.ratio";
+  private static final float DEFAULT_MINIMAL_COMPRESS_RATIO = 1.1f;
+
+  // The name of the algorithm to use.  This name must be pre-registered in CompressionMap class.
+  private static final String CONFIG_ALGORITHM_NAME = "router.compression.algorithm.name";
+  private static final String DEFAULT_ALGORITHM_NAME = null;
```
justinlin-linkedin, line 68: let's put some default algorithm here.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
justinlin-linkedin: BTW, we can remove the CONFIG prefix on each variable name.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Thank you.. fixed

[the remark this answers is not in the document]
panchow: Fixed

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: This setting actually skips compression if content-encoding is present.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Thank you.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: We don't care about Charset that's why if it's present, we discard it.  We can't stop client from sending it so we'll discard it if present.  However, if the CFG specifies one, we honor it, but it's not required.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: We can't just have content-types alone.  It would be nice to specify prefix, unless we combine prefix and full content-types together in one list.  I agree that we have too many configs throughout Ambry.  The existing configs give the impression that Ambry has detailed granular configurations.  I'm up for simplicity.  Let me remove some of the configs.  I'll keep the config properties in the same class so it's easy to find but remove their configurability.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Let me combine the prefix list and the full content-type list into 1.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Sure. Since the name constant is in common and can't access from here.  I'd have to put a string value here.

ankagrawal (approved):

justinlin-linkedin (approved):
LGTM

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -0,0 +1,204 @@
...
+   *   - text/html; charset=UTF-8
+   *   - multipart/form-data; boundary=something
+   * <p>
+   * Content type format is "prefix/specific; option=value".
+   * Content check is applied in this order.  If not in the list, it's not compressible.
+   * 1. Check whether the full contentType parameter is specified in content-type set.  Example "text/html"
+   * 2. If contentType parameter contains option (separated by ";"), remove the option and check again.
+   *    For example, parameter content-type "text/html; charset=UTF-8", check config by content-type "text/html".
+   * 3. Get the content-type prefix (the string before the "/" separator) and lookup prefix again.
+   *    For example, parameter content-type "text/html; charset=UTF-8", check config by content-prefix "text".
```
Arun-LinkedIn, line 166: Minor: Sorry for late comment. Would it have less comparisons if we first check if prefix matches (option 3) and check for exact match in case if fails? 

For example, just rearranging your order 
1. Remove the options if they are present and keep only prefix/specific.
2. Check if prefix is present in compressible list. If yes, return true.
3. Else, check if there is an exact match. If yes, return true.
4. Else, return false.

PR #2233, opened 2022-09-27
Temporary Disable BatchOperationCallbackTrackerTest before we figure out router.getOperationsCount() is not zero problem. 

JingQianCloud:
Temporary Disable BatchOperationCallbackTrackerTest.
Currently after the test, router.getOperationsCount() is not zero.


snalli (approved):

snalli:
if you can un-draft this, it can be merged.

PR #2236, opened 2022-09-28
Remove CloudServiceClusterChangeHandler class

Arun-LinkedIn:
`CloudServiceClusterChangeHandler` was used to support cloud replicas in ClusterMap. Since we no longer need cloud code, removing this class and it's associated code in `HelixClusterManager`.


ambry-clustermap/src/main/java/com/github/ambry/clustermap/CloudServiceClusterChangeHandler.java
```diff
@@ -1,157 +0,0 @@
...
-  private final ClusterMapConfig clusterMapConfig;
-  private final ClusterChangeHandlerCallback changeHandlerCallback;
-  private final CloudServiceDataNode cloudServiceDataNode;
-
-  public CloudServiceClusterChangeHandler(String dcName, ClusterMapConfig clusterMapConfig,
-      ClusterChangeHandlerCallback changeHandlerCallback) throws Exception {
-    this.dcName = dcName;
-    this.clusterMapConfig = clusterMapConfig;
-    this.changeHandlerCallback = changeHandlerCallback;
-    cloudServiceDataNode = new CloudServiceDataNode(dcName, clusterMapConfig);
```
panchow, line 53: Since we no longer support cloud-based data nodes, does this mean we can remove CloudServiceDataNode too?  Please make sure this change will not affect VCR.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -133,20 +133,6 @@ public HelixClusterManager(ClusterMapConfig clusterMapConfig, String instanceNam
           initializationFailureMap.putIfAbsent(initializer.getDcName(), e);
         }
       }
-
-      for (DcInfo dcInfo : dcToDcInfo.values()) {
-        if (dcInfo.clusterChangeHandler instanceof ClusterMapChangeListener) {
```
panchow, line 138: Why remove this code?  This code seems like generic clusterChangeListener that's not cloud DC specific.

mikeyang923: @panchow  Based on comments below looks this section is "tailored to the CloudServiceClusterChangeHandler, which only needs to be provided one replica per partition"

ambry-clustermap/src/main/java/com/github/ambry/clustermap/CloudServiceClusterChangeHandler.java
Arun-LinkedIn: Yeah, we can remove CloudServiceDataNode too. I can try to put up a separate PR for remove rest of code related to cloud-based data nodes. 

Yeah, this shouldn't affect VCR. I think CloudServiceClusterChangeHandler was added to support cloud-based replicas in cluster map.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: Yeah, above code is sort of tailored for `CloudServiceClusterChangeHandler`.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/DatacenterInitializer.java
```diff
@@ -109,15 +109,10 @@ public void start() {
...
-            break;
-          case CLOUD_BACKED:
-            dcInfo = initializeCloudDatacenter();
-            break;
-          default:
-            throw new UnsupportedOperationException("Unknown replica type: " + dcZkInfo.getReplicaType());
+        if (dcZkInfo.getReplicaType() == ReplicaType.DISK_BACKED) {
+          dcInfo = initializeHelixDatacenter();
+        } else {
+          throw new UnsupportedOperationException("Unknown replica type: " + dcZkInfo.getReplicaType());
```
justinlin-linkedin, line 115: So the next PR would be remove this ReplicaType, right?

ankagrawal: ReplicaType.CLOUD_BACKED is needed for recovery flow.

ankagrawal (approved):

PR #2238, opened 2022-09-28
Implement the ReplicateBlobRequest handling in AmbryServer.

JingQianCloud:
Implement the ReplicateBlobRequest handling in AmbryServer.

### DIFF
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
@@ -170,6 +170,7 @@ public class MessageSievingInputStream extends InputStream {
       }
       msg = output.getMsg();
     }
+
     if (output.getException() != null) {
       if (output.getException() instanceof MessageFormatException) {
         logger.error("Error validating/transforming the message {} and hence skipping the message", msgInfo,
@@ -178,15 +179,15 @@ public class MessageSievingInputStream extends InputStream {
       } else {
         throw new IOException("Encountered exception during transformation", output.getException());
       }
-    } else if (output.getMsg() == null) {
+    }
+    if (output.getMsg() == null) {
       logger.info("Transformation is on, and the message with id {} does not have a replacement and was discarded.",
           msgInfo.getStoreKey());
       return null;
-    } else {
-      logger.trace("Original message length {}, transformed bytes read {}", msgInfo.getSize(),
-          output.getMsg().getMessageInfo().getSize());
-      return output.getMsg();
     }
+    logger.trace("Original message length {}, transformed bytes read {}", msgInfo.getSize(),
+        output.getMsg().getMessageInfo().getSize());
+    return output.getMsg();
   }
 
   /**
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
@@ -156,38 +156,36 @@ public class MessageSievingInputStream extends InputStream {
   public static Message transferInputStream(List<Transformer> transformers, InputStream inStream, MessageInfo msgInfo)
       throws IOException {
     Message msg = new Message(msgInfo, inStream);
-    logger.trace("transferInputStream for message {}", msgInfo);
-
     if (transformers == null || transformers.isEmpty()) {
       // Write the message without any transformations.
       return msg;
-    } else {
-      TransformationOutput output = null;
-      for (Transformer transformer : transformers) {
-        output = transformer.transform(msg);
-        if (output.getException() != null || output.getMsg() == null) {
-          break;
-        } else {
-          msg = output.getMsg();
-        }
+    }
+
+    logger.trace("transferInputStream for message {}", msgInfo);
+    TransformationOutput output = null;
+    for (Transformer transformer : transformers) {
+      output = transformer.transform(msg);
+      if (output.getException() != null || output.getMsg() == null) {
+        break;
       }
-      if (output.getException() != null) {
-        if (output.getException() instanceof MessageFormatException) {
-          logger.error(
-              "Error validating/transforming the message {} and hence skipping the message", msgInfo, output.getException());
-          return null;
-        } else {
-          throw new IOException("Encountered exception during transformation", output.getException());
-        }
-      } else if (output.getMsg() == null) {
-        logger.info("Transformation is on, and the message with id {} does not have a replacement and was discarded.",
-            msgInfo.getStoreKey());
+      msg = output.getMsg();
+    }
+    if (output.getException() != null) {
+      if (output.getException() instanceof MessageFormatException) {
+        logger.error("Error validating/transforming the message {} and hence skipping the message", msgInfo,
+            output.getException());
         return null;
       } else {
-        logger.trace("Original message length {}, transformed bytes read {}", msgInfo.getSize(),
-            output.getMsg().getMessageInfo().getSize());
-        return output.getMsg();
+        throw new IOException("Encountered exception during transformation", output.getException());
       }
+    } else if (output.getMsg() == null) {
+      logger.info("Transformation is on, and the message with id {} does not have a replacement and was discarded.",
+          msgInfo.getStoreKey());
+      return null;
+    } else {
+      logger.trace("Original message length {}, transformed bytes read {}", msgInfo.getSize(),
+          output.getMsg().getMessageInfo().getSize());
+      return output.getMsg();
     }
   }
 
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
@@ -58,6 +58,7 @@ public class MessageSievingInputStream extends InputStream {
    * @param inStream The stream from which bytes need to be read. If the underlying stream is SocketInputStream, it needs
    *               to be blocking
    * @param messageInfoList List of MessageInfo which contains details about the messages in the stream
+   * @param transformers the list of the {@link Transformer} to use (in order).
    * @param metricRegistry Metric register to register metrics
    * @throws java.io.IOException
    */
@@ -143,6 +144,51 @@ public class MessageSievingInputStream extends InputStream {
     sievedStream = new SequenceInputStream(inputStreamEnumeration);
     this.sievedStreamSize = (int) sievedMessageInfoList.stream().mapToLong(MessageInfo::getSize).sum();
     logger.trace("Completed validation of message stream ");
+  }
+
+  /**
+   * Transfer one message with a list of {@link Transformer}
+   * @param inStream The stream from which bytes need to be read.
+   * @param msgInfo MessageInfo which contains details about the message in the stream
+   * @param transformers the list of the {@link Transformer} to use (in order).
+   * @throws java.io.IOException
+   */
+  public static Message transferInputStream(List<Transformer> transformers, InputStream inStream, MessageInfo msgInfo)
+      throws IOException {
+    Message msg = new Message(msgInfo, inStream);
+    logger.trace("transferInputStream for message {}", msgInfo);
+
+    if (transformers == null || transformers.isEmpty()) {
+      // Write the message without any transformations.
+      return msg;
+    } else {
+      TransformationOutput output = null;
+      for (Transformer transformer : transformers) {
+        output = transformer.transform(msg);
+        if (output.getException() != null || output.getMsg() == null) {
+          break;
+        } else {
+          msg = output.getMsg();
+        }
+      }
+      if (output.getException() != null) {
+        if (output.getException() instanceof MessageFormatException) {
+          logger.error(
+              "Error validating/transforming the message {} and hence skipping the message", msgInfo, output.getException());
+          return null;
+        } else {
+          throw new IOException("Encountered exception during transformation", output.getException());
+        }
+      } else if (output.getMsg() == null) {
+        logger.info("Transformation is on, and the message with id {} does not have a replacement and was discarded.",
+            msgInfo.getStoreKey());
+        return null;
+      } else {
+        logger.trace("Original message length {}, transformed bytes read {}", msgInfo.getSize(),
+            output.getMsg().getMessageInfo().getSize());
+        return output.getMsg();
+      }
+    }
   }
 
   /**
--- a/ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
+++ b/ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
@@ -19,7 +19,6 @@ import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.PartitionState;
-import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.clustermap.ReplicaType;
 import com.github.ambry.commons.BlobId;
 import com.github.ambry.commons.ErrorMapping;
@@ -40,7 +39,6 @@ import com.github.ambry.network.ConnectedChannel;
 import com.github.ambry.network.ConnectionPool;
 import com.github.ambry.network.LocalRequestResponseChannel.LocalChannelRequest;
 import com.github.ambry.network.NetworkRequest;
-import com.github.ambry.network.Port;
 import com.github.ambry.network.RequestResponseChannel;
 import com.github.ambry.network.Send;
 import com.github.ambry.network.ServerNetworkResponseMetrics;
@@ -66,6 +64,7 @@ import com.github.ambry.store.StoreKeyConverter;
 import com.github.ambry.store.StoreKeyConverterFactory;
 import com.github.ambry.store.StoreKeyFactory;
 import com.github.ambry.store.Transformer;
+import com.github.ambry.utils.NettyByteBufDataInputStream;
 import com.github.ambry.utils.SystemTime;
 import com.github.ambry.utils.Utils;
 import io.netty.buffer.ByteBufInputStream;
@@ -77,7 +76,6 @@ import java.util.Collections;
 import java.util.EnumSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -107,7 +105,7 @@ public class AmbryRequests implements RequestAPI {
   protected static final Logger publicAccessLogger = LoggerFactory.getLogger("PublicAccessLogger");
   private static final Logger logger = LoggerFactory.getLogger(AmbryRequests.class);
 
-  private static String ON_DEMAND_REPLICATION_CLIENTID_PREFIX = "replication-ondemand-fetch";
+  private static String ON_DEMAND_REPLICATION_CLIENTID_PREFIX = "replication-ondemand-fetch-";
 
   public AmbryRequests(StoreManager storeManager, RequestResponseChannel requestResponseChannel, ClusterMap clusterMap,
       DataNodeId nodeId, MetricRegistry registry, ServerMetrics serverMetrics, FindTokenHelper findTokenHelper,
@@ -741,15 +739,16 @@ public class AmbryRequests implements RequestAPI {
     // Get the two parameters from the replicateBlobRequest: the blobId and the remoteHostName
     BlobId blobId = replicateBlobRequest.getBlobId();
     String remoteHostName = replicateBlobRequest.getSourceHostName();
+    int remoteHostPort = replicateBlobRequest.getSourceHostPort();
+    DataNodeId remoteDataNode = clusterMap.getDataNodeId(remoteHostName, remoteHostPort);
+    GetResponse getResponse = null;
     try {
-      Port remoteHostPort;
-      Optional<DataNodeId> remoteDataNode = blobId.getPartition().getReplicaIds().stream().map(ReplicaId::getDataNodeId)
-          .filter(node -> node.getHostname().equals(remoteHostName)).findFirst();
-      if (remoteDataNode.isPresent()) {
-        remoteHostPort = remoteDataNode.get().getPortToConnectTo();
-      } else {
-        logger.error("ReplicateBlobRequest {} couldn't find the remote host {} in the clustermap.", blobId, remoteHostName);
-        throw new IOException("ReplicateBlobRequest " + blobId + " couldn't find the remote host " + remoteHostName);
+      if (remoteDataNode == null) {
+        logger.error("ReplicateBlobRequest {} couldn't find the remote host {} {} in the clustermap.", blobId,
+            remoteHostName, remoteHostPort);
+        throw new IOException(
+            "ReplicateBlobRequest " + blobId + " couldn't find the remote host " + remoteHostName + " port "
+                + remoteHostPort);
       }
       // ON_DEMAND_REPLICATION_TODO: Add configuration as replicationConfig.replicationConnectionPoolCheckoutTimeoutMs
       final int connectionPoolCheckoutTimeoutMs = 1000;
@@ -761,19 +760,22 @@ public class AmbryRequests implements RequestAPI {
       partitionRequestInfoList.add(partitionInfo);
       // Get the Blob including expired and deleted.
       final String clientId = ON_DEMAND_REPLICATION_CLIENTID_PREFIX + currentNode.getHostname();
-      GetRequest getRequest = new GetRequest(replicateBlobRequest.getCorrelationId(),
-          clientId, MessageFormatFlags.All, partitionRequestInfoList, GetOption.Include_All);
+      GetRequest getRequest = new GetRequest(replicateBlobRequest.getCorrelationId(), clientId, MessageFormatFlags.All,
+          partitionRequestInfoList, GetOption.Include_All);
 
       // step 1: get the blob from the remote replica
-      ConnectedChannel connectedChannel = connectionPool.checkOutConnection(remoteHostName, remoteHostPort, connectionPoolCheckoutTimeoutMs);;
+      ConnectedChannel connectedChannel =
+          connectionPool.checkOutConnection(remoteHostName, remoteDataNode.getPortToConnectTo(),
+              connectionPoolCheckoutTimeoutMs);
       ChannelOutput channelOutput = connectedChannel.sendAndReceive(getRequest);
-      GetResponse getResponse = GetResponse.readFrom(channelOutput.getInputStream(), clusterMap);
+      getResponse = GetResponse.readFrom(channelOutput.getInputStream(), clusterMap);
       if ((getResponse.getError() != ServerErrorCode.No_Error) || (getResponse.getPartitionResponseInfoList() == null)
           || (getResponse.getPartitionResponseInfoList().size() == 0)) {
-        logger.error("ReplicateBlobRequest failed to get blob {} from the remote node {} {} {}",
-            blobId, remoteHostName, remoteHostPort, getResponse.getError());
-        response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
-            ServerErrorCode.Unknown_Error);
+        logger.error("ReplicateBlobRequest failed to get blob {} from the remote node {} {} {}", blobId, remoteHostName,
+            remoteHostPort, getResponse.getError());
+        response =
+            new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
+                ServerErrorCode.Unknown_Error);
       } else {
         // step 2: get the blob message from the response and write to the store
         // only have one partition. And checked at least it has one entry above.
@@ -792,15 +794,15 @@ public class AmbryRequests implements RequestAPI {
               MessageInfo orgMsgInfo = messageInfoList.get(0);
               Message output = MessageSievingInputStream.transferInputStream(Collections.singletonList(transformer),
                   getResponse.getInputStream(), orgMsgInfo);
-              if (output == null || !output.getMessageInfo().getStoreKey().equals(blobId)) {
+              if (output == null) {
                 logger.error("ReplicateBlobRequest transferInputStream {} returned null, {} {} {}", orgMsgInfo,
                     remoteHostName, remoteHostPort, blobId);
-                response =
-                    new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
-                        ServerErrorCode.Unknown_Error);
+                response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(),
+                    replicateBlobRequest.getClientId(), ServerErrorCode.Unknown_Error);
               } else {
-                MessageFormatWriteSet writeset = new MessageFormatWriteSet(output.getStream(), Collections.singletonList(output.getMessageInfo()),
-                    false);
+                MessageFormatWriteSet writeset =
+                    new MessageFormatWriteSet(output.getStream(), Collections.singletonList(output.getMessageInfo()),
+                        false);
                 Store store = storeManager.getStore(blobId.getPartition());
                 store.put(writeset);
 
@@ -813,8 +815,8 @@ public class AmbryRequests implements RequestAPI {
                 }
                 logger.info("ReplicateBlobRequest replicated Blob {} from remote host {} {}", blobId, remoteHostName,
                     remoteHostPort);
-                response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
-                    ServerErrorCode.No_Error);
+                response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(),
+                    replicateBlobRequest.getClientId(), ServerErrorCode.No_Error);
               } // if (output == null) {
             } // if (messageInfoList == null || messageInfoList.size() != 1)
           } catch (StoreException e) { // catch the store write exception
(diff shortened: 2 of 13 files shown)


ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
```diff
@@ -145,6 +146,51 @@ public InputStream nextElement() {
...
+        logger.info("Transformation is on, and the message with id {} does not have a replacement and was discarded.",
+            msgInfo.getStoreKey());
+        return null;
+      } else {
+        logger.trace("Original message length {}, transformed bytes read {}", msgInfo.getSize(),
+            output.getMsg().getMessageInfo().getSize());
+        return output.getMsg();
+      }
+    }
+  }
```
justinlin-linkedin, line 192: this seems to be something we can merge with the other method.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +714,147 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+        long processingTime = SystemTime.getInstance().milliseconds() - startTime;
+        totalTimeSpent += processingTime;
+        publicAccessLogger.info("{} {} processingTime {}", replicateBlobRequest, response, processingTime);
+        metrics.replicateBlobProcessingTimeInMs.update(processingTime);
+        requestResponseChannel.sendResponse(response, request,
+            new ServerNetworkResponseMetrics(metrics.replicateBlobResponseQueueTimeInMs, metrics.replicateBlobSendTimeInMs,
+                metrics.replicateBlobTotalTimeInMs, null, null, totalTimeSpent));
+        return;
+      }
+    }
```
justinlin-linkedin, line 783: can we merge step 1 and step 2 so we don't have a finally breaking this method.
And in the end, we have to release the input stream in GetResponse since it's Netty ByteBuf.

ambry-replication/src/test/java/com/github/ambry/replication/MockConnectionPool.java
```diff
@@ -80,9 +88,16 @@ public void shutdown() {
 
   @Override
   public ConnectedChannel checkOutConnection(String host, Port port, long timeout) {
+    // For mock layer, we ignore the port provided.
+    // For mock layer, each replica is using incremental port number.
+    // When we send ReplicateBlobRequest, don't know the remote port number.
```
justinlin-linkedin, line 93: we do, we can get the remote port number from clustermap.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,154 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
 
   @Override
   public void handleReplicateBlobRequest(NetworkRequest request) throws IOException, InterruptedException {
-    throw new UnsupportedOperationException("Request type not supported");
+    if (connectionPool == null || transformer == null) {
```
Arun-LinkedIn, line 719: Minor: May be we don't need to check if `transformer == null` since `MessageSievingInputStream.transferInputStream(transformer)` returns original message if transformer is null?

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,154 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+    BlobId blobId = replicateBlobRequest.getBlobId();
+    String remoteHostName = replicateBlobRequest.getSourceHostName();
+    Port remoteHostPort;
+    Optional<DataNodeId> remoteDataNode = blobId.getPartition().getReplicaIds().stream().map(ReplicaId::getDataNodeId)
+        .filter(node -> node.getHostname().equals(remoteHostName)).findFirst();
+    if (remoteDataNode.isPresent()) {
+      remoteHostPort = remoteDataNode.get().getPortToConnectTo();
+    } else {
+      logger.error("ReplicateBlobRequest {} couldn't find the remote host {} in the clustermap.", blobId, remoteHostName);
+      throw new IOException("ReplicateBlobRequest " + blobId + " couldn't find the remote host " + remoteHostName);
```
Arun-LinkedIn, line 750: Would sending `ReplicateBlobResponse` with internal error be better. Looks like we throw `Exception` from this method, it just closes the connection with client (frontend) without sending any response.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,154 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+    final int connectionPoolCheckoutTimeoutMs = 1000;
+
+    List<PartitionRequestInfo> partitionRequestInfoList = new ArrayList<PartitionRequestInfo>();
+    List<BlobId> blobIds = new ArrayList<BlobId>();
+    blobIds.add(blobId);
+    PartitionRequestInfo partitionInfo = new PartitionRequestInfo(blobId.getPartition(), blobIds);
+    partitionRequestInfoList.add(partitionInfo);
+    // Get the Blob including expired and deleted.
+    GetRequest getRequest = new GetRequest(replicateBlobRequest.getCorrelationId(),
+        replicateBlobRequest.getClientId(), MessageFormatFlags.All, partitionRequestInfoList,
```
Arun-LinkedIn, line 762: One quick Q. Since this is a GetRequest from server to server, should we send a different client ID like `replication-ondemand-fetch-[server hostname]` so that it will help to identify these requests in logs?

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,154 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+    // only have one partition. And checked at least it has one entry above.
+    PartitionResponseInfo partitionResponseInfo = getResponse.getPartitionResponseInfoList().get(0);
+    if (partitionResponseInfo.getErrorCode() == ServerErrorCode.No_Error) {
+      try {
+        List<MessageInfo> messageInfoList = partitionResponseInfo.getMessageInfoList();
+        if (messageInfoList == null || messageInfoList.size() != 1) {
+          logger.error("ReplicateBlobRequest PartitionResponseInfo response from GetRequest {} {} {} {} {} returned null.",
+              partitionResponseInfo, messageInfoList, remoteHostName, remoteHostPort, blobId);
+          response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
+              ServerErrorCode.Blob_Not_Found);
```
Arun-LinkedIn, line 804: Just a Q. Should we return internal error in this case since target server would actually return a response with valid error code when blob is not found.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,154 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+              ServerErrorCode.Unknown_Error);
+        }
+      }
+    } else if (partitionResponseInfo.getErrorCode() == ServerErrorCode.Blob_Deleted) {
+      // Since GetOption is Include_All, even it's deleted on the remote replica, we'll still get the PutBlob.
+      // One exception is that because of compaction or other reasons, the PutRecord is gone.
+      logger.error("ReplicateBlobRequest Blob {} of {} is delete from the source replica", blobId, replicateBlobRequest);
+      response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
+          ServerErrorCode.Blob_Not_Found);
+    } else if (partitionResponseInfo.getErrorCode() == ServerErrorCode.Blob_Authorization_Failure) {
```
Arun-LinkedIn, line 849: QQ: Would we need to handle this error case. Since it is server-server communication, would we get authorization failure?

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
Arun-LinkedIn: +1

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
justinlin-linkedin, line 750: let's not throw an exception here, but return an response with a solid server error.

I think have all this logic in a big try-catch statement is good,  just like other methods, so we can deal with exceptions and returning proper response in one single place.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,154 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+      logger.error("ReplicateBlobRequest authorization failure for replicateBlobRequest {}", replicateBlobRequest);
+      response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
+          ServerErrorCode.Blob_Authorization_Failure);
+    } else {
+      logger.error("ReplicateBlobRequest unknown failure {} for replicateBlobRequest {}",
+          partitionResponseInfo.getErrorCode(), replicateBlobRequest);
+      response = new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
+          ServerErrorCode.Unknown_Error);
```
justinlin-linkedin, line 857: we can just relay the error to the frontend.

ambry-test-utils/src/main/java/com/github/ambry/clustermap/MockDataNodeId.java
```diff
@@ -58,7 +58,19 @@ public MockDataNodeId(String hostname, List<Port> ports, List<String> mountPaths
...
+
+  /**
+   * Create a {@link MockDataNodeId} object from given {@code ports}, {@code mountPaths}, {@code dataCenter}.
+   * @param ports Ports associated with server on node.
+   * @param mountPaths Mount paths for replicas on node.
+   * @param dataCenter Name of datacenter.
+   * @param uniqueName true if want to have unique host name. false, host name is the same.
+   */
+  public MockDataNodeId(List<Port> ports, List<String> mountPaths, String dataCenter, boolean uniqueName) {
+    this(uniqueName? "localhost" + count : "localhost", ports, mountPaths, dataCenter);
```
justinlin-linkedin, line 72: why not just always set uniqueName to true?

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Modified. 
It cleaned up the final break but at the same time it adds up the if/else layers. :-)

JingQianCloud: Thanks for the reminder to release the GetResponse.
I'm just curious why nettyByteBufLeakHelper couldn't catch it.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Actually I have a question here.
If the transfer is null for some reason, shall we keep the original message and write to the store or fail the request?
For example, we may fail to get the transformer below. 
I'm not exactly sure what the transfer does. If we are supposed to do the transfer but fails to do that, does that mean the content is not we want and we shouldn't write to the store?


      try {
        StoreKeyConverter keyConverter = storeKeyConverterFactory.getStoreKeyConverter();
        transformer = Utils.getObj(serverConfig.serverMessageTransformer, storeKeyFactory, keyConverter);
      } catch (Exception e) {
        logger.error("Failed to create transformer", e);
      }

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Looks like to me, in the handleReplicaMetadataRequest and handleRequests, we may throw UnsupportedOperationException. I guess the caller will deal with that and wrap a response for it?

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Good idea. It's be helpful for debugging. Added.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: I'm not sure about that. Probably you are right, we won't hit this case.
But since replication code handles it, I keep it here for safe.
I hope it's ok.

ambry-test-utils/src/main/java/com/github/ambry/clustermap/MockDataNodeId.java
JingQianCloud: Yes, first I tried to always generate unique host name.
But int-test kept falling one after another like FrontendQuotaIntegrationTest. I guess quite some unit test assume the host name is always "localhost".
It may take some effort to understand these integration test cases and clean them up.
So I revert back the change and keep the original way.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: But as you suggested, I've returned Response for this case.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: I assume the GetRequest may have a couple of partition/blob IDs to get.
Do you mean if there is a single BlobID is missing, the remote server will return a error code?

I see. Returning UNKNOWN_ERROR.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Done.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
```diff
@@ -145,6 +146,51 @@ public InputStream nextElement() {
...
+   */
+  public static Message transferInputStream(List<Transformer> transformers, InputStream inStream, MessageInfo msgInfo)
+      throws IOException {
+    Message msg = new Message(msgInfo, inStream);
+    logger.trace("transferInputStream for message {}", msgInfo);
+
+    if (transformers == null || transformers.isEmpty()) {
+      // Write the message without any transformations.
+      return msg;
+    } else {
```
tbingol, line 164: Else is not needed after returning from the if branch

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
```diff
@@ -145,6 +146,51 @@ public InputStream nextElement() {
...
+      TransformationOutput output = null;
+      for (Transformer transformer : transformers) {
+        output = transformer.transform(msg);
+        if (output.getException() != null || output.getMsg() == null) {
+          break;
+        } else {
```
tbingol, line 170: Nit: Else is not needed after breaking in the if branch

ambry-test-utils/src/main/java/com/github/ambry/clustermap/MockDataNodeId.java
justinlin-linkedin: This is not the right way to do it. There are two different ways moving forward.
1. Adding port to the ReplicateBlobRequest (sorry I was against it at the first place, but not it seems like it's necessary). The port would be part of the key to identity the data node, not the port to connect to. For example, if a data node has plaintext port 12001, and ssl port 13001 and http2 port 14001. Then the port number in ReplicateBlobRequest should be 12001, since it will be used to identify the data node. But we still would connect to http2 port.
2. Fix the FrontendQuotaIntegrationTest.

If it's easier to do 1, please go for it. Sorry I misled you at the first place.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
Arun-LinkedIn: Hmm, just thinking that since functionality of `transformer` is abstracted out, we could carry on in this method even if it is null. But, may be either way is okay.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +718,156 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+      // ON_DEMAND_REPLICATION_TODO: Add configuration as replicationConfig.replicationConnectionPoolCheckoutTimeoutMs
+      final int connectionPoolCheckoutTimeoutMs = 1000;
+
+      List<PartitionRequestInfo> partitionRequestInfoList = new ArrayList<PartitionRequestInfo>();
+      List<BlobId> blobIds = new ArrayList<BlobId>();
+      blobIds.add(blobId);
+      PartitionRequestInfo partitionInfo = new PartitionRequestInfo(blobId.getPartition(), blobIds);
+      partitionRequestInfoList.add(partitionInfo);
+      // Get the Blob including expired and deleted.
+      final String clientId = ON_DEMAND_REPLICATION_CLIENTID_PREFIX + currentNode.getHostname();
```
Arun-LinkedIn, line 763: nit: May be a separator (-) between `replication-ondemand-fetch` prefix and hostname

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,162 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+      GetRequest getRequest = new GetRequest(replicateBlobRequest.getCorrelationId(), clientId, MessageFormatFlags.All,
+          partitionRequestInfoList, GetOption.Include_All);
+
+      // step 1: get the blob from the remote replica
+      ConnectedChannel connectedChannel =
+          connectionPool.checkOutConnection(remoteHostName, remoteDataNode.getPortToConnectTo(),
+              connectionPoolCheckoutTimeoutMs);
+      ChannelOutput channelOutput = connectedChannel.sendAndReceive(getRequest);
+      getResponse = GetResponse.readFrom(channelOutput.getInputStream(), clusterMap);
+      if ((getResponse.getError() != ServerErrorCode.No_Error) || (getResponse.getPartitionResponseInfoList() == null)
```
justinlin-linkedin, line 772: nit: we don't need parentheses around each boolean statement

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -135,7 +140,7 @@
  * Tests for {@link AmbryServerRequests}.
  */
 @RunWith(Parameterized.class)
-public class AmbryServerRequestsTest {
+public class AmbryServerRequestsTest extends ReplicationTestHelper {
```
justinlin-linkedin, line 143: why do we need this?

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -914,6 +922,247 @@ public void undeleteTest() throws Exception {
...
+    int correlationId = TestUtils.RANDOM.nextInt();
+    String clientId = TestUtils.getRandomString(10);
+    BlobId blobId = (BlobId) storeKey;
+
+    ReplicateBlobRequest request =
+        new ReplicateBlobRequest(correlationId, clientId, blobId, remoteHost.dataNodeId.getHostname(),
+            remoteHost.dataNodeId.getPort());
+
+    storageManager.resetStore();
+    RequestOrResponseType requestType = request.getRequestType();
```
justinlin-linkedin, line 960: requestType not used anywhere.

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -914,6 +922,247 @@ public void undeleteTest() throws Exception {
...
+    Response response = sendRequestGetResponse(request, ServerErrorCode.No_Error);
+    assertEquals("Operation received at the store not as expected", RequestOrResponseType.PutRequest,
+        MockStorageManager.operationReceived);
+    ReplicateBlobResponse replicateBlobResponse = (ReplicateBlobResponse) response;
+    assertEquals("expect ReplicateBlobRequest is successful. ", replicateBlobResponse.getError(),
+        ServerErrorCode.No_Error);
```
justinlin-linkedin, line 966: can you also check if we can get the blob out from the local store now that replicate blob request succeeded.

ambry-server/src/test/java/com/github/ambry/server/AmbryServerRequestsTest.java
```diff
@@ -914,6 +922,247 @@ public void undeleteTest() throws Exception {
...
+    // add delete to the remote host
+    addDeleteMessagesToReplicasOfPartition(partitionId, storeKey, Arrays.asList(remoteHost));
+
+    MockConnectionPool connectionPool = new MockConnectionPool(hosts, clusterMap, 5);
+    ambryRequests = new AmbryServerRequests(storageManager, requestResponseChannel, clusterMap, dataNodeId,
+        clusterMap.getMetricRegistry(), serverMetrics, findTokenHelper, null, replicationManager, storeKeyFactory,
+        serverConfig, diskManagerConfig, storeKeyConverterFactory, statsManager, helixParticipant, connectionPool);
+
+    Assume.assumeTrue(
+        MessageFormatRecord.getCurrentMessageHeaderVersion() >= MessageFormatRecord.Message_Header_Version_V3);
```
justinlin-linkedin, line 996: same above

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageSievingInputStream.java
```diff
@@ -145,6 +146,49 @@ public InputStream nextElement() {
...
+      // Write the message without any transformations.
+      return msg;
+    }
+
+    logger.trace("transferInputStream for message {}", msgInfo);
+    TransformationOutput output = null;
+    for (Transformer transformer : transformers) {
+      output = transformer.transform(msg);
+      if (output.getException() != null || output.getMsg() == null) {
+        break;
```
snalli, line 169: This is a bit verbose and difficult to read with too many if-clauses.

This can be written as 

```
for (...) {

  if (output.getException() != null) {
     // do something and return
  } else if (getMsg == null) {
     // do something and return
  }
  // do someting

}
```

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
snalli: i'd recommend substituting this boolean with a small helper like isSomethingX(). Makes it so much more readable.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,162 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+    ReplicateBlobRequest replicateBlobRequest;
+    if (request instanceof LocalChannelRequest) {
+      // This is a case where handleReplicateBlobRequest is called when frontends are talking to Azure. In this case, this method
+      // is called by request handler threads running within the frontend router itself. So, the request can be directly
+      // referenced as java objects without any need for deserialization.
+      replicateBlobRequest = (ReplicateBlobRequest) ((LocalChannelRequest) request).getRequestInfo().getRequest();
+    } else {
+      replicateBlobRequest = ReplicateBlobRequest.readFrom(new DataInputStream(request.getInputStream()), clusterMap);
+    }
+    long requestQueueTime = SystemTime.getInstance().milliseconds() - request.getStartTimeInMs();
```
snalli, line 732: you can eliminate requestQueueTime and just use totalTimeSpent

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,162 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+            if (messageInfoList == null || messageInfoList.size() != 1) {
+              logger.error(
+                  "ReplicateBlobRequest PartitionResponseInfo response from GetRequest {} {} {} {} {} returned null.",
+                  partitionResponseInfo, messageInfoList, remoteHostName, remoteHostPort, blobId);
+              response =
+                  new ReplicateBlobResponse(replicateBlobRequest.getCorrelationId(), replicateBlobRequest.getClientId(),
+                      ServerErrorCode.Blob_Not_Found);
+            } else {
+              MessageInfo orgMsgInfo = messageInfoList.get(0);
+              Message output = MessageSievingInputStream.transferInputStream(Collections.singletonList(transformer),
```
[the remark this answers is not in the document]
snalli: So something like

if (error) {
  log.error
  return completeRequest(...);
}

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +717,177 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+          connectionPool.checkOutConnection(remoteHostName, remoteDataNode.getPortToConnectTo(),
+              connectionPoolCheckoutTimeoutMs);
+      ChannelOutput channelOutput = connectedChannel.sendAndReceive(getRequest);
+      GetResponse getResponse = GetResponse.readFrom(channelOutput.getInputStream(), clusterMap);
+      if (getResponse.getError() != ServerErrorCode.No_Error) {
+        logger.error("ReplicateBlobRequest failed to get blob {} from the remote node {} {} {}", blobId, remoteHostName,
+            remoteHostPort, getResponse.getError());
+        return new Pair(getResponse.getError(), getResponse);
+      }
+      if ((getResponse.getPartitionResponseInfoList() == null) || (getResponse.getPartitionResponseInfoList().size()
```
justinlin-linkedin, line 853: nit: we don't have to add parenthese around these two statements.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +717,177 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+              replicateBlobRequest);
+          return new Pair(ServerErrorCode.Blob_Deleted, getResponse);
+        case Blob_Authorization_Failure:
+          logger.error("ReplicateBlobRequest authorization failure for replicateBlobRequest {}", replicateBlobRequest);
+          return new Pair(ServerErrorCode.Blob_Authorization_Failure, getResponse);
+        default:
+          logger.error("ReplicateBlobRequest unknown failure {} for replicateBlobRequest {}",
+              partitionResponseInfo.getErrorCode(), replicateBlobRequest);
+          return new Pair(partitionResponseInfo.getErrorCode(), getResponse);
+      }
```
justinlin-linkedin, line 886: This can be merged into just one statement. something like

``` 
if (partitionResponseInfo.getErrorCode() == No_Error) {
          xxxxxx
} else {
          logger.error("ReplicateBlobRequest failure {} for replicateBlobRequest {}",
              partitionResponseInfo.getErrorCode(), replicateBlobRequest);
          return new Pair(partitionResponseInfo.getErrorCode(), getResponse);
}
```

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
snalli: If you're ok with both, I recommend eliminating it. Consistency is not needed.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +716,162 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
 
   @Override
   public void handleReplicateBlobRequest(NetworkRequest request) throws IOException, InterruptedException {
```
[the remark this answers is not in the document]
snalli: I had the same comment y'day but I can't find it rn. 
But my comment was that this method is overly verbose and indented with too many if-clauses and a lot of duplicate code.

I recommend refactoring common code into one method for example

```
if (error) {
  // do something
  // return completeRequest(args); --> common code, return fast
}

// continue with happy case
```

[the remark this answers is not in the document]
JingQianCloud: Are you reading my last revision, I've refactored it.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +717,177 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+
+  /**
+   * to get the Blob from the remote replica
+   * @param replicateBlobRequest the {@link ReplicateBlobRequest}
+   * @param blobId blobId of the ReplicateBlob request
+   * @param remoteHostName the name of the remote host to get the blob from
+   * @param remoteHostPort the port of the remote host to get the blob from
+   * @return a pair of the {@link ServerErrorCode} and the {@link GetResponse}.
+   */
+  private Pair<ServerErrorCode, GetResponse> getBlobFromRemoteReplicate(ReplicateBlobRequest replicateBlobRequest,
```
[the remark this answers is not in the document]
JingQianCloud: Cleaned up.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: It's my coding style. It's for easy reading and "no need to worry about precedence". 
I hope it's ok that I keep it.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: I simplified it. 
But I'd like to keep the some comments. Hope it's ok.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Removed it.

justinlin-linkedin (approved):
looking good

SophieGuo410 (approved):

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +717,166 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+        new ServerNetworkResponseMetrics(metrics.replicateBlobResponseQueueTimeInMs, metrics.replicateBlobSendTimeInMs,
+            metrics.replicateBlobTotalTimeInMs, null, null, totalTimeSpent));
+  }
+
+  /**
+   * to get the Blob from the remote replica
+   * @param replicateBlobRequest the {@link ReplicateBlobRequest}
+   * @return a pair of the {@link ServerErrorCode} and the {@link GetResponse}.
+   */
+  private Pair<ServerErrorCode, GetResponse> getBlobFromRemoteReplicate(ReplicateBlobRequest replicateBlobRequest) {
```
snalli, line 814: getBlobFromRemoteReplica

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +717,166 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+    totalTimeSpent += processingTime;
+    publicAccessLogger.info("{} {} processingTime {}", replicateBlobRequest, response, processingTime);
+    metrics.replicateBlobProcessingTimeInMs.update(processingTime);
+    requestResponseChannel.sendResponse(response, request,
+        new ServerNetworkResponseMetrics(metrics.replicateBlobResponseQueueTimeInMs, metrics.replicateBlobSendTimeInMs,
+            metrics.replicateBlobTotalTimeInMs, null, null, totalTimeSpent));
+  }
+
+  /**
+   * to get the Blob from the remote replica
```
snalli, line 810: Get blob from remote replica

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
snalli: Please condense verbose booleans into helpers like isSomethingXX

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -683,7 +717,166 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+        logger.error(
+            "ReplicateBlobRequest PartitionResponseInfo response from GetRequest {} {} {} {} {} returned null.",
+            partitionResponseInfo, messageInfoList, remoteHostName, remoteHostPort, blobId);
+        return new Pair(ServerErrorCode.Blob_Not_Found, getResponse);
+      }
+
+      return new Pair(ServerErrorCode.No_Error, getResponse);
+    } catch (Exception e) { // catch the getBlob exception
+      logger.error("ReplicateBlobRequest getBlob {} from the remote node {} hit exception ", blobId, remoteHostName, e);
+      return new Pair(ServerErrorCode.Unknown_Error, null);
```
snalli, line 878: This is good, i like it

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Fixed the typo.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: Do you mean wrapping the "getResponse.getPartitionResponseInfoList() == null) || (getResponse.getPartitionResponseInfoList().size()==0"?
I feel it's a little bit overkilled.

snalli: Yes, it would be readable to have such a helper even if its a one-liner. I can never understand long booleans and they keep growing.

Arun-LinkedIn (approved):

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: To my opinion, two condition is not long. And it is not used much in other places, so I skip this change.

PR #2239, opened 2022-09-28
Add compression metrics.

panchow:
- Define compression metrics for overall compression and metrics per algorithm.
- Compression metrics is available under router metrics.


ambry-api/src/main/java/com/github/ambry/router/ReadableStreamChannel.java
```diff
@@ -31,7 +31,7 @@ public interface ReadableStreamChannel extends Channel {
    * Return the size of stream that is available on this channel. If -1, then size is unknown.
    * @return the size of the stream available on this channel. -1 if size is unknown.
    */
-  public long getSize();
+  long getSize();
```
snalli, line 34: why remove public ?

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
```diff
@@ -0,0 +1,141 @@
...
+  public static class AlgorithmMetrics {
+
+    // Algorithm's compress() metrics.
+    public final Meter compressRate;      // Rate compress() is called.
+    public final Counter compressError;   // Rate compress() throws exception.
+    public final Histogram compressRatioPercent;    // compression ratio % = 100 * UncompressedSize/CompressedSize
+    public final Histogram fullSizeCompressTimeInMicroseconds;
+    public final Histogram fullSizeCompressSpeedMBPerSec;
+    public final Histogram partialFullCompressTimeInMicroseconds;
+    public final Histogram partialFullCompressSpeedMBPerSec;
```
snalli, line 21: what is partialFull ?

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
```diff
@@ -0,0 +1,141 @@
...
+      partialFullDecompressSpeedMBPerSec = registry.histogram(MetricRegistry.name(ownerClass, algorithmName,
+          "PartialFullDecompressSpeedMBPerSec"));
+    }
+  }
+
+  // Group all compression related metrics under one owner "Compression".
+  // These metrics are per chunk, not per operation.
+  private final static Class<?> ownerClass = Compression.class;
+  private final MetricRegistry metricsRegistry;
+  private final Map<String, AlgorithmMetrics> algorithmToMetricsMap = new ConcurrentHashMap<>();
```
snalli, line 68: are there going to be multiple algorithms operating at the same time ?

ambry-router/src/test/java/com/github/ambry/router/CompressionMetricsTest.java
```diff
@@ -0,0 +1,19 @@
+package com.github.ambry.router;
```
snalli, line 1: license

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
```diff
@@ -0,0 +1,141 @@
...
+    // Decompression metrics.
+    decompressSuccessRate = registry.meter(MetricRegistry.name(ownerClass,"DecompressSuccessRate"));
+    decompressErrorRate = registry.meter(MetricRegistry.name(ownerClass,"DecompressErrorRate"));
+    decompressExpandSizeBytes = registry.counter(MetricRegistry.name(ownerClass,"DecompressExpandSizeBytes"));
+    decompressError_BufferConversion = registry.counter(MetricRegistry.name(ownerClass,
+        "DecompressErrorBufferConversion"));
+    decompressError_BufferTooSmall = registry.counter(MetricRegistry.name(ownerClass,
+        "DecompressErrorBufferTooSmall"));
+    decompressError_UnknownAlgorithmName = registry.counter(MetricRegistry.name(ownerClass,
+        "DecompressErrorUnknownAlgorithmName"));
```
snalli, line 126: will this ever happen ? if this happens, then it means we can't read the blob ?

ambry-api/src/main/java/com/github/ambry/router/ReadableStreamChannel.java
panchow: It's an interface.  Everything is public by default.  The "public" has no impact.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
panchow: 4MB is full size.  < 4MB is partial full.  The metrics for < 4 MB is highly skewed due to overheads.  For compressing a 1KB text is probably less efficient than compressing 100 KB, so the metrics for compressing 1 KB will generate many misleading metrics (like low compression ratio, low throughput (due to overhead)).  So looking at the 4MB compression metrics provides a more consistent and comparable picture of the algorithm performance.  At first, I thought about emitting metrics for 4 MB chunks only, but that may miss issues when the chunk size is < 4 MB.  That's why I separated the metrics for full chunks from the partial full chunks.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
panchow: There will be only 1 selected compression algorithm, but decompression may use different ones.  Yes, we do support multiple concurrent decompression algorithms.  That's why we store the algorithm name inside the compressed chunk.

ambry-router/src/test/java/com/github/ambry/router/CompressionMetricsTest.java
panchow: Yes :)  I missed this file too.  I'll add it.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
panchow: This shouldn't happen.  If they do, we need to investigate immediately because it could mean data lost.  That's why we need to monitor these metrics and setup alarms.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
snalli: I see. Please put this in comments, otherwise no one would guess this.
Also consider an alternate name such as fullChunkCompressTime and subChunkCompressTime.
The word chunk rings a bell for 4MB.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
snalli: hmm so if a blob is compressed using algo X, can it be decompressed using algo Y ? 
or do u mean blob1 is compressed/decompressed using algo X, blob2 using algo Y and blob3 using algo Z ?

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -777,5 +780,15 @@ public RouterConfig(VerifiableProperties verifiableProperties) {
...
+
+    compressionConfig = new CompressionConfig(verifiableProperties);
+  }
+
+  /**
+   * Get the compression-related configs.
+   * @return The compression-related configs.
+   */
+  public CompressionConfig getCompressionConfig() {
+    return compressionConfig;
```
snalli, line 792: we should consider https://projectlombok.org/features/GetterSetter

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -638,6 +638,9 @@ public class RouterConfig {
   @Config(ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS)
   public final long routerUpdateOpMetadataRelianceTimestampInMs;
 
+  // Group compression-related configs in the CompressConfig class.
+  private final CompressionConfig compressionConfig;
```
justinlin-linkedin, line 642: you can set this to public, it's a readonly object. just like any other members of this RouterConfig class.

ambry-api/src/main/java/com/github/ambry/router/ReadableStreamChannel.java
```diff
@@ -45,5 +45,5 @@ public interface ReadableStreamChannel extends Channel {
    *                 into the {@code asyncWritableChannel} or if there is an exception in doing so. This can be null.
    * @return the {@link Future} that will eventually contain the result of the operation.
    */
-  public Future<Long> readInto(AsyncWritableChannel asyncWritableChannel, Callback<Long> callback);
+  Future<Long> readInto(AsyncWritableChannel asyncWritableChannel, Callback<Long> callback);
```
justinlin-linkedin, line 48: thanks for removing this.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+
+public class CompressionMetrics {
+  public static class AlgorithmMetrics {
+
+    // Algorithm's compress() metrics.
+    public final Meter compressRate;      // Rate compress() is called.
+    public final Counter compressError;   // Rate compress() throws exception.
+    public final Histogram compressRatioPercent;    // compression ratio % = 100 * UncompressedSize/CompressedSize
+    public final Histogram fullSizeCompressTimeInMicroseconds;
```
justinlin-linkedin, line 31: let's simplify the name to TimeInMs.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
justinlin-linkedin: we can categorize blobs based on size and then create speed metrics for those categories. something like smallBlobCompressSpeed, mediumBlobCompressSpeed and largeBlobCompressionSpeed. Please search ServerMetrics class, there is a similar settings in that class.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
```diff
@@ -0,0 +1,154 @@
...
+  // Compression metrics, regardless of compression algorithm.
+  public final Meter compressSkipRate;      // The rate compression is skipped due to config settings.
+  public final Meter compressErrorRate;     // The overall compress failure rate, eg: compress() throws exception.
+  public final Meter compressAcceptRate;    // The rate compression is applied and accepted after filtering.
+  public final Counter compressReduceSizeBytes;  // For accepted, bytes reduce = OriginalSize - CompressedSize.
+
+  // Compression error metrics.  These metrics must be monitored and raise alarm if as soon as they occurred.
+  public final Counter compressError_BufferConversion;
+  public final Counter compressError_ConfigInvalidCompressorName;
+  public final Counter compressError_CompressFailed;
```
justinlin-linkedin, line 92: Not a big fan of such naming pattern, care to change to camel case?

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterMetrics.java
```diff
@@ -267,6 +267,7 @@ public class NonBlockingRouterMetrics {
   private final Map<DataNodeId, NodeLevelMetrics> dataNodeToMetrics;
   private final RouterConfig routerConfig;
   private ScheduledExecutorService scheduler = null;
+  private CompressionMetrics compressionMetrics;
```
justinlin-linkedin, line 270: we can just instantiate this object here, doing it in the method below might suffer some concurrency issue.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
panchow: If a blob is compressed using algo X, it must be decompressed by algo X.  However, if a new or better algo is available, we would want to switch to the new algo.  However, previously compressed blobs still must be decompressed using original compressor.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
panchow: I don't think it's necessary to break it to multiple categories.  I believe the full size and partial-full size is good enough.  The goal of these metrics aren't for analysis.  We just want to make sure the algorithm isn't acting crazy.  So most of the time we'll just keep an eye on the time of the full size chunks.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
panchow: Yes, we should use lombok.  It saves us from writing getters and setters.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
panchow: Yes, I could make it public just to make it like others.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
panchow: Milliseconds is a big unit.  Typical responsive REST calls take 1-10 ms.  Compression time could take less than 1 ms.  In generally, it's okay to use Milliseconds to measure length operations like external calls, but for internal calls, milliseconds doesn't work unless the value type is a double instead of long.  For this case, it's actually better to use microseconds.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
panchow: Yes, I can change it to camel case.

Regarding categorizing the compression metrics.  Are we going use the metrics for analysis?  I presumption is we'll use it just for performance and stability purpose.  I don't think we should break it down to multiple categories unless there is a need for it.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterMetrics.java
panchow: Yes, we can do that.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
snalli: awesome ! these getters and setters are mostly noise.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
justinlin-linkedin: please make it public.

justinlin-linkedin (approved):
One last comment.

tbingol (approved):

snalli (approved):

snalli:
Approved but please merge after addressing all comments

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
ankagrawal: Do note that lombok can have its own issues. If an IDE doesn't support lombok, then the code will show up as error in the IDE. We are open source code, so not sure if tying ourselves to certain IDEs is a good idea. As lombok releases new versions, if they are backward incompatible, the IDEs will need sometime to update their support.
Also putting breakpoints and creating javadocs will be hard. Sometimes code reading could also become harder, because you won't be able to find the method you are looking for

Not saying that we should not consider lombok. But lets do make sure we understand the pros and cons, and weigh it considering that we are open source. Maybe we can strictly define the cases where we use lombok (only pure POJOs?), and ensure that our style checker can validate that.

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
ankagrawal: If we are ensuring that we will not compress smaller blobs, do we need this metric still?

ambry-router/src/main/java/com/github/ambry/router/CompressionMetrics.java
```diff
@@ -0,0 +1,170 @@
...
+/**
+ * Metrics for both compression service and compression algorithm.
+ * The compression service metrics counts why compression is skip due to configuration and its error rate.
+ * The algorithm-specific metrics measure the time and speed of the algorithm.
+ * The service metric names use the format [Owner].[MetricName].
+ * These algorithm-specific metric names use the format [Owner].[AlgorithmName].[MetricName].
+ * Example algorithm names include "LZ4" and "ZSTD".
+ */
+public class CompressionMetrics {
+  public static class AlgorithmMetrics {
```
ankagrawal, line 33: nit: javadocs for `AlgorithmMetrics`.

ankagrawal (approved):

PR #2241, opened 2022-09-30
Start using IndexSegment offsets from compaction log to generate new token in PersistentIndex

justinlin-linkedin:
This PR would enable persistent index to start using index segment offsets from compaction log to generate reset token if the index segment can't be found due compaction.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -163,6 +163,7 @@ class PersistentIndex {
   private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
   private Runnable getBlobReadInfoTestCallback = null;
 
+  private volatile boolean shouldRebuildTokenBasedOnCompactionHistory = false;
   private volatile boolean sanityCheckFailed = false;
   // A map from index segment start offset before compaction to index segment start offset after compaction.
   // This map will only be modified in the compaction thread and accessed in the replication threads.
@@ -650,6 +651,14 @@ class PersistentIndex {
    */
   Map<Long, Set<Offset>> getCompactionTimestampToIndexSegmentOffsets() {
     return Collections.unmodifiableMap(compactionTimestampToIndexSegmentOffsets);
+  }
+
+  /**
+   * Enable RebuildTokenBasedOnCompactionHistory
+   */
+  void enableRebuildTokenBasedOnCompactionHistory() {
+    logger.info("Index {}: Enable rebuild token based on compaction history", dataDir);
+    shouldRebuildTokenBasedOnCompactionHistory = true;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -2650,11 +2663,15 @@ private StoreFindToken revalidateStoreFindToken(StoreFindToken token,
...
-            revalidatedToken = generateTokenBasedOnResetKey(token, indexSegments);
+          if (shouldRebuildTokenBasedOnCompactionHistory && !sanityCheckFailed) {
+            revalidatedToken = generateTokenBasedOnIndexSegmentOffset(token, indexSegments);
           } else {
-            revalidatedToken = new StoreFindToken();
+            if (!ignoreResetKey && token.getResetKey() != null) {
+              // reset key is present in the token
+              revalidatedToken = generateTokenBasedOnResetKey(token, indexSegments);
+            } else {
+              revalidatedToken = new StoreFindToken();
```
SophieGuo410, line 2673: Do we want a metric to calculate how much time it has been reduced for the token to be reset to the beginning of legSegments?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -2707,6 +2724,76 @@ private StoreFindToken generateTokenBasedOnResetKey(StoreFindToken token,
...
+      segment = indexSegments.get(afterOffset);
+      if (segment != null) {
+        logger.trace("Index {}: Offset {} maps to {} after compaction, reset token to this index segment", dataDir,
+            offset, afterOffset);
+        metrics.offsetFoundInBeforeAfterMapCount.inc();
+        return new StoreFindToken(FindTokenType.IndexBased, segment.getStartOffset(),
+            segment.getFirstKeyInSealedSegment(), sessionId, incarnationId, true, StoreFindToken.CURRENT_VERSION,
+            segment.getResetKey(), segment.getResetKeyType(), segment.getResetKeyLifeVersion());
+      } else {
+        logger.trace("Index {}: Offset {} maps to {} after compaction, but it doesn't exist in index", dataDir, offset,
```
SophieGuo410, line 2790: Will this cause the while loop loop forever?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: I don't understand, what do we want to record here? The time we to generate a new token?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: The sanity check should prevent that. Hmm, let me add something to the sanity check to avoid forever loop.

SophieGuo410 (approved):

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
SophieGuo410: Nvm, we can monitoring the replication lag.

Arun-LinkedIn (approved):

PR #2243, opened 2022-09-30
Added replicateBlob interface to the router.

JingQianCloud:
Added replicateBlob interface to the router. It's not supported yet.

Wanted to open the interface up. And add http interface to support curl as well.
Had discussion with Justin. Justin suggested that we don't open the ReplicateBlob up to the client. We may use it only internally.

Change the PR to draft first.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/router/Router.java
+++ b/ambry-api/src/main/java/com/github/ambry/router/Router.java
@@ -13,6 +13,7 @@
  */
 package com.github.ambry.router;
 
+import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.commons.Callback;
 import com.github.ambry.commons.CallbackUtils;
 import com.github.ambry.messageformat.BlobInfo;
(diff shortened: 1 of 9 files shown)


ambry-api/src/main/java/com/github/ambry/router/Router.java
```diff
@@ -173,6 +186,20 @@ default CompletableFuture<String> putBlob(BlobProperties blobProperties, byte[]
...
+   * about whether the request succeeded or not.
+   * @param blobId The ID of the blob to be replicated.
+   * @param serviceId The service ID of the service replicating the blob. This can be null if unknown.
+   * @param sourceDataNode The source {@link DataNodeId} to get the blob from.
+   * @return A future that would contain information about whether the replicateBlob succeeded or not, eventually.
+   */
+  default CompletableFuture<Void> replicateBlob(String blobId, String serviceId, DataNodeId sourceDataNode) {
+    CompletableFuture<Void> future = new CompletableFuture<>();
+    replicateBlob(blobId, serviceId, sourceDataNode, CallbackUtils.fromCompletableFuture(future), null);
+    return future;
```
snalli, line 200: bit confused here. All 3 routers below - non-blocking, in-mem, perf - don't support replicateBlob. So which router actually supports it ?

JingQianCloud: This is the PR to add the interface. The next PR will add the implementation in Non Blocking Router.

But Justin suggested don't add interface and don't open up it to the client. And we only use the ReplicateBlobRequest internally.

JingQianCloud:
Closing this PR now.
We may don't expose the ReplicateBlob to the router interface and don't allow client to call the ReplicateBlob.

PR #2244, opened 2022-09-30
Remove some unused BlobStoreCompactor configuration

justinlin-linkedin:
In this PR, we remove two unused configurations for BlobStoreCompactor.
1. store.always.enable.target.index.duplicate.checking: This configuration was used to always check if there is a duplicate in target index in compaction. We don't have to do this since only the crashed compaction would need to check duplicates in target index.
2. store.compaction.enable.basic.info.on.missing.duplicate: This configuration is used to print out some basic index and entry information when we are expecting duplicates and not getting any. Change it to trace log.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -1430,23 +1430,6 @@ class BlobStoreCompactor {
       copyCandidates.removeIf(copyCandidate ->
           isDuplicate(copyCandidate, duplicateSearchSpan, indexSegment.getStartOffset(), checkAlreadyCopied) || (
               config.storeContainerDeletionEnabled && isFromDeprecatedContainer(copyCandidate)));
-      if (duplicateSearchSpan != null && validEntriesSize == copyCandidates.size()) {
-        // When duplicate search span is not null, which mean it's highly likely that we would see duplicates from
-        // source. If we are here, then there is no duplicate at all, this might be an error.
-        logger.trace(
-            "Processing IndexSegment {} with duplicate search span {} in {}, we probably should have duplicates.",
-            indexSegment.toString(), duplicateSearchSpan, storeId);
-        logger.trace("IndexSegments in source persistent index with {}", storeId);
-        srcIndex.getIndexSegments().values().forEach(seg -> logger.info("{}", seg.getFile()));
-        Map<PersistentIndex.IndexEntryType, Long> collect = copyCandidates.stream()
-            .collect(Collectors.groupingBy(entry -> entry.getValue().getIndexValueType(), Collectors.counting()));
-        logger.trace("{} with {}", collect.entrySet()
-            .stream()
-            .map(ent -> ent.getValue().toString() + " " + String.valueOf(ent.getKey()))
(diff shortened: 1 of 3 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1435,22 +1430,22 @@ public List<IndexEntry> getValidEntry(IndexSegment indexSegment, FileSpan duplic
       copyCandidates.removeIf(copyCandidate ->
           isDuplicate(copyCandidate, duplicateSearchSpan, indexSegment.getStartOffset(), checkAlreadyCopied) || (
               config.storeContainerDeletionEnabled && isFromDeprecatedContainer(copyCandidate)));
-      if (duplicateSearchSpan != null && validEntriesSize == copyCandidates.size()
-          && config.storeCompactionEnableBasicInfoOnMissingDuplicate) {
```
SophieGuo410, line 1439: QQ: It seems like previously we always use the default value which is false. And since it has no error messages pops out, I feel no one will pay attention to this error unless someone enable trace log on purpose.

ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
```diff
@@ -2914,19 +2907,12 @@ private void testDuplicatePutsHelper() throws Exception {
...
-        compactor.compact(details, bundleReadBuffer);
-        fail("Should fail with duplicate put");
-      } catch (StoreException e) {
-        assertEquals(StoreErrorCodes.Unknown_Error, e.getErrorCode());
-        assertTrue(e.getMessage().contains("duplicate PUT"));
-      }
+      fail("Should fail with duplicate put");
+    } catch (StoreException e) {
+      assertEquals(StoreErrorCodes.Unknown_Error, e.getErrorCode());
+      assertTrue(e.getMessage().contains("duplicate PUT"));
```
mikeyang923, line 2915: Should we upgrade Junit to 5 and use `assertThrows`?

mikeyang923 (approved):
lgtm

ambry-store/src/test/java/com/github/ambry/store/BlobStoreCompactorTest.java
justinlin-linkedin: upgrading to Junit 5 takes a long time. Junit 5 changes the assert interface to put message after the values. In Junit 4, we can do
```
Assert.assertEquals("Something is wrong", expectedValue, obtainedValue);
```
But in Junit 5, we have to do 
```
Assert.assertEquals(expectedValue, obtainedValue, "Something is wrong");
```
It will take a long to fix all of that.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: The configuration was created for a specific reason back then. We saw some Put IndexEntries having duplicates. Later we figured out it's because we have a list of log segments under compaction but they are not consecutive. After we fixed that this never happened again.

ankagrawal (approved):

PR #2245, opened 2022-10-03
Compression service for Ambry.

panchow:
The compression service applies configs to compress and decompress ByteBuf. It uses metrics to collect statistics and errors.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
@@ -162,7 +162,8 @@ public class CompressionService {
       compressionMetrics.compressErrorConfigInvalidCompressorName.inc();
       logger.error("The default compressor algorithm, " + algorithmName
           + ", specified in config does not exist.  This config is ignored.");
-      defaultCompressor = allCompressions.get(ZstdCompression.ALGORITHM_NAME);
+      algorithmName = ZstdCompression.ALGORITHM_NAME;
+      defaultCompressor = allCompressions.get(algorithmName);
     }
 
     // Apply compression.
--- a/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
@@ -43,6 +43,7 @@ public class CompressionService {
 
   private final CompressionConfig compressionConfig;
   private final CompressionMetrics compressionMetrics;
+  private Compression defaultCompressor;
 
   // When a new compression algorithm is introduced, add the new algorithm here.
   // WARNING: Do not remove compression algorithms from this list.  See {@code Compression} interface for detail.
@@ -57,6 +58,18 @@ public class CompressionService {
   public CompressionService(CompressionConfig config, CompressionMetrics metrics) {
     compressionConfig = config;
     compressionMetrics = metrics;
+
+    // Check whether the default compressor name is valid in the config.  If not, fall back to Zstd.
+    Compression compressor = allCompressions.get(compressionConfig.algorithmName);
+    if (compressor == null) {
+      compressor = allCompressions.get(ZstdCompression.ALGORITHM_NAME);
+
+      // Emit metrics for default compressor name in config does not exist.
+      compressionMetrics.compressErrorConfigInvalidCompressorName.inc();
+      logger.error("The default compressor algorithm, " + compressionConfig.algorithmName
+          + ", specified in config does not exist.  This default has changed to " + ZstdCompression.ALGORITHM_NAME);
+    }
+    defaultCompressor = compressor;
   }
 
   /**
@@ -65,6 +78,23 @@ public class CompressionService {
    */
   public CompressionMap getAllCompressions() {
     return allCompressions;
+  }
+
+  /**
+   * Get the default compressor.
+   * @return The default compressor.
+   */
+  public Compression getDefaultCompressor() {
+    return defaultCompressor;
+  }
+
+  /**
+   * Set the default compressor to use.
+   * @param newDefaultCompressor The new default compressor.
+   */
+  public void setDefaultCompressor(Compression newDefaultCompressor) {
+    Objects.requireNonNull(newDefaultCompressor, "newDefaultCompressor");
+    defaultCompressor = newDefaultCompressor;
   }
 
   /**
@@ -150,20 +180,8 @@ public class CompressionService {
       return null;
     }
 
-    // Get default compressor.  If config missing, fallback to Zstandard.
-    String algorithmName = compressionConfig.algorithmName;
-    Compression defaultCompressor = allCompressions.get(algorithmName);
-    if (defaultCompressor == null) {
-      // Emit metrics for default compressor name does not exist.
-      compressionMetrics.compressErrorConfigInvalidCompressorName.inc();
-      logger.error("The default compressor algorithm, " + algorithmName
-          + ", specified in config does not exist.  This config is ignored.");
-      algorithmName = ZstdCompression.ALGORITHM_NAME;
-      defaultCompressor = allCompressions.get(algorithmName);
-    }
-
     // Apply compression.
-    CompressionMetrics.AlgorithmMetrics algorithmMetrics = compressionMetrics.getAlgorithmMetrics(algorithmName);
+    CompressionMetrics.AlgorithmMetrics algorithmMetrics = compressionMetrics.getAlgorithmMetrics(defaultCompressor.getAlgorithmName());
     Pair<Integer, byte[]> compressResult;
     try {
       algorithmMetrics.compressRate.mark();
@@ -265,7 +283,8 @@ public class CompressionService {
       decompressedBuffer = decompressor.decompress(buffer, bufferOffset, dataSize);
       long durationMicroseconds = (System.nanoTime() - startTime)/1000;
 
-      long speedInMBPerSec = (long) (BytePerMicrosecondToMBPerSec * decompressedBuffer.length / (double) durationMicroseconds);
+      long speedInMBPerSec = durationMicroseconds == 0 ? 0 :
+          (long) (BytePerMicrosecondToMBPerSec * decompressedBuffer.length / (double) durationMicroseconds);
       if (decompressedBuffer.length == fullChunkSize) {
         algorithmMetrics.fullSizeDecompressTimeInMicroseconds.update(durationMicroseconds);
         algorithmMetrics.fullSizeDecompressSpeedMBPerSec.update(speedInMBPerSec);
--- a/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
@@ -1,0 +1,349 @@
+/*
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.router;
+
+import com.github.ambry.compression.Compression;
+import com.github.ambry.compression.CompressionException;
+import com.github.ambry.compression.CompressionMap;
+import com.github.ambry.compression.LZ4Compression;
+import com.github.ambry.compression.ZstdCompression;
+import com.github.ambry.config.CompressionConfig;
+import com.github.ambry.messageformat.BlobProperties;
+import com.github.ambry.utils.Pair;
+import com.github.ambry.utils.Utils;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import java.nio.ByteBuffer;
+import java.util.Objects;
+import org.apache.commons.lang3.tuple.Triple;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Compression service is thread-safe singleton that apply compression and decompression.
+ * Its implementation is backed by the CompressionMap in the common library.
+ */
+public class CompressionService {
+
+  private static final Logger logger = LoggerFactory.getLogger(CompressionService.class);
+
+  // MB/sec = (byte/microsecond * MB/1024*1024bytes * 1000000Microseconds/sec) = 1 MB/1.048576 sec
+  private static final double BytePerMicrosecondToMBToSec = 1/1.048576;
+
+  private final CompressionConfig compressionConfig;
+  private final CompressionMetrics compressionMetrics;
+
+  // When a new compression algorithm is introduced, add the new algorithm here.
+  // WARNING: Do not remove compression algorithms from this list.  See {@code Compression} interface for detail.
+  private final CompressionMap allCompressions = CompressionMap.of(new ZstdCompression(), new LZ4Compression());
+
+  /**
+   * Create an instance of the compression service using the compression config and metrics.
+   * The default compressor is loaded from config, if available.  If not available, ZStd is the default.
+   * @param config Compression config.
+   * @param metrics Compression metrics.
+   */
+  public CompressionService(CompressionConfig config, CompressionMetrics metrics) {
+    compressionConfig = config;
+    compressionMetrics = metrics;
+  }
+
+  /**
+   * Get the compression map that contains all compression algorithms.
+   * @return The compression map.
+   */
+  public CompressionMap getAllCompressions() {
+    return allCompressions;
+  }
+
+  /**
+   * Check whether the blob is compressible based on the blob properties.
+   *
+   * @param blobProperties The blob property that is common to all chunks. It helps to detect whether blob is
+   *                       compressible or not.
+   * @return True if blob is compressible based on the blob properties; FALSE if blob is not compressible.
+   */
+  public boolean isBlobCompressible(BlobProperties blobProperties) {
+    Objects.requireNonNull(blobProperties, "blobProperties cannot be null.");
+
+    // Return false if compression is disabled.  No need to emit metrics.
+    if (!compressionConfig.isCompressionEnabled) {
+      logger.info("No compression applied because compression is disabled.");
+      return false;
+    }
+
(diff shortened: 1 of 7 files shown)


ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
```diff
@@ -0,0 +1,343 @@
...
+  @Test
+  public void testCompressWithLowRatio() throws CompressionException {
+    CompressionConfig config = new CompressionConfig();
+    byte[] sourceBuffer = "Test Message for testing purpose.  The Message is part of the testing message.".getBytes();
+    ByteBuf sourceByteBuf = Unpooled.wrappedBuffer(sourceBuffer);
+    CompressionMetrics metrics = new CompressionMetrics(new MetricRegistry());
+    config.minimalSourceDataSizeInBytes = 1;
+
+    // Test: Compression skipped because compression ratio is too low.
+    metrics = new CompressionMetrics(new MetricRegistry());
```
ankagrawal, line 126: double initialization?

ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
```diff
@@ -0,0 +1,343 @@
...
+        "Test Message for testing purpose.  The Message is part of the testing message." +
+        "Test Message for testing purpose.  The Message is part of the testing message.").getBytes();
+    ByteBuf sourceByteBuf = Unpooled.wrappedBuffer(sourceBuffer);
+    CompressionMetrics metrics = new CompressionMetrics(new MetricRegistry());
+
+    // Test: Happy case - compression succeed partial full.
+    config.minimalCompressRatio = 1.0;
+    config.minimalSourceDataSizeInBytes = 1;
+    config.algorithmName = "BAD"; // algorithm not found, will fall back to Zstd.
+    metrics = new CompressionMetrics(new MetricRegistry());
```
ankagrawal, line 150: double initialization

ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
panchow: Thanks

ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
panchow: Thanks

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+              + "chunkBuffer.hasArray() = %s, chunkBuffer.readableBytes = %d, chunkBuffer.arrayOffset = %d",
+          chunkBuffer.hasArray(), chunkBuffer.readableBytes(), chunkBuffer.arrayOffset()), ex);
+
+      // Emit metrics for count how many compressions skipped due to buffer conversion error.
+      compressionMetrics.compressErrorRate.mark();
+      compressionMetrics.compressErrorBufferConversion.inc();
+      return null;
+    }
+
+    // Get default compressor.  If config missing, fallback to Zstandard.
```
ankagrawal, line 157: nit: maybe add the logic to get the default compressor in a separate method.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+    // Return false if compression is disabled.  No need to emit metrics.
+    if (!compressionConfig.isCompressionEnabled) {
+      logger.info("No compression applied because compression is disabled.");
+      return false;
+    }
+
+    // Check content-encoding. BlobProperty.ContentEncoding contains a string, do not apply compression.
+    // Note: Calling getBlobProperties() throws if blob properties is null, so it never returns null.
+    String contentEncoding = blobProperties.getContentEncoding();
+    if (compressionConfig.isSkipWithContentEncoding && !Utils.isNullOrEmpty(contentEncoding)) {
```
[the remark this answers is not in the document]
panchow: Compression will be disabled by default until feature launch and during testing.  It will be disabled in case of emergency too.  Normally, it will be enabled.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+    // Compression is accepted, emit metrics.
+    compressionMetrics.compressAcceptRate.mark();
+    compressionMetrics.compressReduceSizeBytes.inc(sourceDataSize - compressedDataSize);
+    return Unpooled.wrappedBuffer(ByteBuffer.wrap(compressResult.getSecond(), 0, compressedDataSize));
+  }
+
+  /**
+   * Decompress the specified compressed buffer.
+   *
+   * @param compressedBuffer The compressed buffer.
```
ankagrawal, line 223: nit: missing param?

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+      // Emit metrics to count how many compressions skipped due to encoded data.
+      compressionMetrics.compressSkipRate.mark();
+      compressionMetrics.compressSkipContentEncoding.inc();
+      return false;
+    }
+
+    // Check the content-type.
+    String contentType = blobProperties.getContentType();
+    if (!compressionConfig.isCompressibleContentType(contentType)) {
+      logger.info("No compression applied because content-type " + contentType + " is incompressible.");
```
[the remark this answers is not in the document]
panchow: It's for performance reason.  Behind the scene, string builder (like string.format) is expensive.  String concatenation of a few short messages is faster than using string builder.  For 10+ concatenations, it's better to use format for readability reason.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: Getting the default compressor isn't called by another method (not shared).  Also, it's not big, just 3 lines.  Breaking it up to another method shortens this method creates more methods which doesn't reduce code side.  I don't think it makes a difference by breaking it out, unless it has readability issue.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: Thank you.

ankagrawal (approved):

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+   *                       compressible or not.
+   * @return True if blob is compressible based on the blob properties; FALSE if blob is not compressible.
+   */
+  public boolean isBlobCompressible(BlobProperties blobProperties) {
+    Objects.requireNonNull(blobProperties, "blobProperties cannot be null.");
+
+    // Return false if compression is disabled.  No need to emit metrics.
+    if (!compressionConfig.isCompressionEnabled) {
+      return false;
+    }
```
snalli, line 83: I like this style of coding !

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+
+    // Check content-encoding. BlobProperty.ContentEncoding contains a string, do not apply compression.
+    // Note: Calling getBlobProperties() throws if blob properties is null, so it never returns null.
+    String contentEncoding = blobProperties.getContentEncoding();
+    if (compressionConfig.isSkipWithContentEncoding && !Utils.isNullOrEmpty(contentEncoding)) {
+      logger.info("No compression applied because compression content is encoded with " + contentEncoding);
```
snalli, line 89: This will overwhelm the log

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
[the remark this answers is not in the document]
snalli: Can you change it to trace ?

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
snalli, line 89: compression content -> blob content

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+    // For all other case, compression should be enabled.
+    return true;
+  }
+
+  /**
+   * Apply compression to a chunk buffer if compression is enabled and content-type is compressible.
+   * It applies compression config to determine whether to accept or reject the compressed result.
+   * For example, if the compressed data size is smaller than original data by certain % (exceeds threshold),
+   * the compressed data is accepted and return compressed data; otherwise, compression is discarded.
+   * This method emits metrics along the process.  It returns null if failed instead of throwing exceptions.
```
snalli, line 115: I like that it returns null instead of throwing exc

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+   * This method is typically called for the first chunk to determine whether to compress the entire blob or not.
```
snalli, line 116: how do u extrapolate this from just the first data chunk ?

ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
```diff
@@ -0,0 +1,341 @@
...
+    Assert.assertEquals(1, metrics.decompressSuccessRate.getCount());
+    Assert.assertTrue(metrics.decompressExpandSizeBytes.getCount() > 0);
+    Assert.assertArrayEquals(decompressedBuffer.array(), sourceBuffer);
+
+    // Zstd is the default compression algorithm.
+    CompressionMetrics.AlgorithmMetrics algorithmMetrics = metrics.getAlgorithmMetrics(ZstdCompression.ALGORITHM_NAME);
+    Assert.assertEquals(1, algorithmMetrics.decompressRate.getCount());
+    Assert.assertTrue(algorithmMetrics.fullSizeDecompressTimeInMicroseconds.getCount() > 0);
+  }
+}
```
snalli, line 341: Test with unknown algorithm should default to zstd

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+    }
+
+    // Get default compressor.  If config missing, fallback to Zstandard.
+    String algorithmName = compressionConfig.algorithmName;
+    Compression defaultCompressor = allCompressions.get(algorithmName);
+    if (defaultCompressor == null) {
+      // Emit metrics for default compressor name does not exist.
+      compressionMetrics.compressErrorConfigInvalidCompressorName.inc();
+      logger.error("The default compressor algorithm, " + algorithmName
+          + ", specified in config does not exist.  This config is ignored.");
```
justinlin-linkedin, line 163: if the default compression doesn't exist in the CompressionMap, then it will never exist unless we change the config or code and release a new version. If this is the case, then we should do this in constructor so we don't see this error message over and over again.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+      defaultCompressor = allCompressions.get(algorithmName);
+    }
+
+    // Apply compression.
+    CompressionMetrics.AlgorithmMetrics algorithmMetrics = compressionMetrics.getAlgorithmMetrics(algorithmName);
+    Pair<Integer, byte[]> compressResult;
+    try {
+      algorithmMetrics.compressRate.mark();
+      long startTime = System.nanoTime();
+      compressResult = defaultCompressor.compress(sourceBuffer, sourceBufferOffset, sourceDataSize);
```
justinlin-linkedin, line 174: at same point we have to create ByteBuf outside of compress method. Do you have a plan for that?

ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
```diff
@@ -0,0 +1,341 @@
...
+
+  @Test
+  public void testIsBlobCompressible() {
+    BlobProperties blobProperties = new BlobProperties(1234L, "testServiceID", (short) 2222, (short) 3333, true);
+    CompressionMetrics metrics = new CompressionMetrics(new MetricRegistry());
+    CompressionConfig config = new CompressionConfig();
+    CompressionService service = new CompressionService(config, metrics);
+
+    // Test: compression disabled.  Compression does nothing.
+    config.isCompressionEnabled = false;
```
justinlin-linkedin, line 42: hmm, configuration's members are not read-only. i would rather make it read-only and create a new configuration from a Properties.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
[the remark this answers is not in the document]
panchow: Yes, I can change them to trace.  I would like to know how we use the log levels.  Usually, I use info() for important steps, like a decisions.  I use debug and trace for parameter logging, a step has started and completed.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: Let's talk about how we use the logs.  It's subjective how much to log.  When things go well, it seems to log too much.  When something goes wrong, we wish we have more logs.  Would it help if I change this to trace?

panchow: That comment is irrelevant, I should remove it.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: I agree.  I would throw an exception if something went wrong and the code can't handle it.  I would return null if it's expected and can be handled by the method.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: Sorry, it was out-dated comment.  Let me remove it since it causes confusion.  Compression is applied per chunk basis, not per blob.

ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
panchow: Line 148 tests invalid algorithm name and it fell back to Zstd.  There is already a test case for it.  This case is when the compression algorithm is not specified in config, it would use the default Zstd.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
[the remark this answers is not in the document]
panchow: ok

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+   * @param isFullChunk Whether this is a full size chunk (4MB) or smaller.
+   * @return Returns a new compressed buffer.  It returns null is no compression is applied.
+   */
+  public ByteBuf compressChunk(ByteBuf chunkBuffer, boolean isFullChunk) {
+    Objects.requireNonNull(chunkBuffer, "blobProperties cannot be null.");
+
+    // Check the blob size.  If it's too small, do not compress.
+    int sourceDataSize = chunkBuffer.readableBytes();
+    if (sourceDataSize < compressionConfig.minimalSourceDataSizeInBytes) {
+      logger.info("No compression applied because source data size " + sourceDataSize + " is smaller than minimal size "
```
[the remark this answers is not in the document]
panchow: yup, I've turned them to trace.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: This log should appear only once because the code after this log sets the default to Zstd.  In my opinion, it is not a good practice to put logic to constructor.   Constructor is intended for initialization and parameter checking, not to perform any real action/work.  Actions should be in methods.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -0,0 +1,350 @@
...
+   */
+  public ByteBuf decompress(ByteBuf compressedBuffer, int fullChunkSize) throws CompressionException {
+    Objects.requireNonNull(compressedBuffer, "compressedBuffer");
+
+    // buffer, offset, and size.
+    Triple<byte[], Integer, Integer> bufferInfo;
+    try {
+      bufferInfo = convertByteBufToByteArray(compressedBuffer);
+    }
+    catch (Exception ex) {
```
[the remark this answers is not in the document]
panchow: Let me correct this.

ambry-router/src/test/java/com/github/ambry/router/CompressionServiceTest.java
panchow: Whether to make it read-only or not really depends on what "config" is.  If we use "config" as for a one-time remote-control purpose, we can make it read-only and feel free to make as many copies as wanted.  Any changes will require restarting the service that is very expensive.  If the purpose of config is to control run-time behavior of a service, then config should be read-write and singleton.  Admin should be able to change run-time behavior without having to restart the service.  I think configuration should be run-time behavior controller that can alter at run-time without restarting services and the service should not cache config values.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: I need more info about this comment.  Currently compression service does not create ByteBuf, the caller (PutOperation and GetOperation) will capture the method output, byte[], and convert it to ByteBuf or ByteBuffer, whatever works for them.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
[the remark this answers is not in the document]
snalli: Yeah that is correct. I follow this https://stackoverflow.com/questions/2031163/when-to-use-the-different-log-levels.
I have not found any rule in the team. You can turn on tracing in prod to debug compression.

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
snalli: Yup trace is better. If something goes wrong, you can enable tracing in prod.

snalli (approved):

PR #2246, opened 2022-10-05
Implemented ReplicateBlobOperation and ReplicateBlobManager in the frontend.

JingQianCloud:
Implemented ReplicateBlobOperation and ReplicateBlobManager in the frontend.

This is based on the server side's ReplicateBlobRequest implementation.
For the ReplicateBlob, caller will choose one particular source DataNode to replicate the Blob from.
But for the target data node, we rely on the SimpleOperationTracker to pick one. In this way, error handling is much easier. Caller needn't worry about target DataNode's condition. And we may make sure enough quorum is reached during the replication.

### DIFF
--- a/ambry-commons/src/main/java/com/github/ambry/commons/LoggingNotificationSystem.java
+++ b/ambry-commons/src/main/java/com/github/ambry/commons/LoggingNotificationSystem.java
@@ -58,28 +58,26 @@ public class LoggingNotificationSystem implements NotificationSystem {
 
   @Override
   public void onBlobDeleted(String blobId, String serviceId, Account account, Container container) {
-    logger.debug("onBlobDeleted {}", blobId,
-        ", " + serviceId + ", accountName " + (account == null ? null : account.getName()) + ", accountId" + (
-            account == null ? null : account.getId()) + ", containerName " + (container == null ? null
-            : container.getName()) + ", containerId " + (container == null ? null : container.getId()));
+    logger.debug("onBlobDeleted {}, serviceId {}, accountName {}, accountId {}, containerName {}, containerId {}",
+        blobId, serviceId, account == null ? null : account.getName(), account == null ? null : account.getId(),
+        container == null ? null : container.getName(), container == null ? null : container.getId());
   }
 
   @Override
   public void onBlobUndeleted(String blobId, String serviceId, Account account, Container container) {
-    logger.debug("onBlobUndeleted {}", blobId,
-        ", " + serviceId + ", accountName " + (account == null ? null : account.getName()) + ", accountId" + (
-            account == null ? null : account.getId()) + ", containerName " + (container == null ? null
-            : container.getName()) + ", containerId " + (container == null ? null : container.getId()));
+    logger.debug("onBlobUndeleted {}, serviceId {}, accountName {}, accountId {}, containerName {}, containerId {}",
+        blobId, serviceId, account == null ? null : account.getName(), account == null ? null : account.getId(),
+        container == null ? null : container.getName(), container == null ? null : container.getId());
   }
 
   @Override
   public void onBlobReplicated(String blobId, String serviceId, Account account, Container container,
       DataNodeId sourceHost) {
-    logger.debug("onBlobReplicated {}", blobId,
-        ", " + serviceId + ", accountName " + (account == null ? null : account.getName()) + ", accountId" + (
-            account == null ? null : account.getId()) + ", containerName " + (container == null ? null
-            : container.getName()) + ", containerId " + (container == null ? null : container.getId())
-            + ", sourceHostName " + sourceHost.getHostname() + ", sourceHostPort " + sourceHost.getPort());
+    logger.debug(
+        "onBlobReplicated {}, serviceId {}, accountName {}, accountId {}, containerName {}, containerId {}, host {} {}",
+        blobId, serviceId, account == null ? null : account.getName(), account == null ? null : account.getId(),
+        container == null ? null : container.getName(), container == null ? null : container.getId(),
+        sourceHost.getHostname(), sourceHost.getPort());
   }
 
   @Override
--- a/ambry-commons/src/main/java/com/github/ambry/commons/LoggingNotificationSystem.java
+++ b/ambry-commons/src/main/java/com/github/ambry/commons/LoggingNotificationSystem.java
@@ -15,6 +15,7 @@ package com.github.ambry.commons;
 
 import com.github.ambry.account.Account;
 import com.github.ambry.account.Container;
+import com.github.ambry.clustermap.DataNodeId;
 import com.github.ambry.messageformat.BlobProperties;
 import com.github.ambry.notification.BlobReplicaSourceType;
 import com.github.ambry.notification.NotificationBlobType;
@@ -69,7 +70,16 @@ public class LoggingNotificationSystem implements NotificationSystem {
         ", " + serviceId + ", accountName " + (account == null ? null : account.getName()) + ", accountId" + (
             account == null ? null : account.getId()) + ", containerName " + (container == null ? null
             : container.getName()) + ", containerId " + (container == null ? null : container.getId()));
+  }
 
+  @Override
+  public void onBlobReplicated(String blobId, String serviceId, Account account, Container container,
+      DataNodeId sourceHost) {
+    logger.debug("onBlobReplicated {}", blobId,
+        ", " + serviceId + ", accountName " + (account == null ? null : account.getName()) + ", accountId" + (
+            account == null ? null : account.getId()) + ", containerName " + (container == null ? null
+            : container.getName()) + ", containerId " + (container == null ? null : container.getId())
+            + ", sourceHostName " + sourceHost.getHostname() + ", sourceHostPort " + sourceHost.getPort());
   }
 
   @Override
--- a/ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobManager.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobManager.java
@@ -1,0 +1,210 @@
+/**
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.router;
+
+import com.github.ambry.account.Account;
+import com.github.ambry.account.AccountService;
+import com.github.ambry.account.Container;
+import com.github.ambry.clustermap.ClusterMap;
+import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.commons.BlobId;
+import com.github.ambry.commons.Callback;
+import com.github.ambry.commons.ResponseHandler;
+import com.github.ambry.config.RouterConfig;
+import com.github.ambry.network.RequestInfo;
+import com.github.ambry.network.ResponseInfo;
+import com.github.ambry.notification.NotificationSystem;
+import com.github.ambry.protocol.ReplicateBlobResponse;
+import com.github.ambry.quota.QuotaChargeCallback;
+import com.github.ambry.utils.Pair;
+import com.github.ambry.utils.Time;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Handles {@link ReplicateBlobOperation}. A {@code ReplicateBlobManager} keeps track of all the ReplicateBlob
+ * operations that are assigned to it, and manages their states and life cycles.
+ */
+class ReplicateBlobManager {
+  private final Set<ReplicateBlobOperation> replicateBlobOperations;
+  private final HashMap<Integer, ReplicateBlobOperation> correlationIdToReplicateBlobOperation;
+  private final NotificationSystem notificationSystem;
+  private final Time time;
+  private final ResponseHandler responseHandler;
+  private final AccountService accountService;
+  private final NonBlockingRouterMetrics routerMetrics;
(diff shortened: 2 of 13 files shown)


ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -248,6 +248,18 @@ class SimpleOperationTracker implements OperationTracker {
         // Here set the success target to the number of eligible replicas.
         replicaSuccessTarget = eligibleReplicas.size();
         break;
+      case ReplicateBlobOperation:
+        // Replicate one blob. Use the same "success target" and "request parallelism" config settings as PutOperation.
+        // Only difference is for ReplicateBlobOperation, crossColoEnabled is true.
+        // ON_DEMAND_REPLICATION_TODO: we may consider add its own configure setting if needed.
+        eligibleReplicas = getEligibleReplicas(datacenterName, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER));
```
justinlin-linkedin, line 255: if you want to enable cross-colo, then getEligibleReplica's first argument has to be null. You can't specify a datacenter name as the first argument since it will only get the replicas from that specific datacenter.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -248,6 +248,18 @@ class SimpleOperationTracker implements OperationTracker {
...
+        replicaSuccessTarget =
```
justinlin-linkedin, line 256: the sucessTarget is too large for ReplicateBlob. If we only have 1 source datanode and 6 remote colo nodes that are missing data and eligiable, then the successTarget is 5 from the code. We don't need that many replicas to have this blob.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -248,6 +248,18 @@ class SimpleOperationTracker implements OperationTracker {
...
+            routerConfig.routerGetEligibleReplicasByStateEnabled ? Math.max(eligibleReplicas.size() - 1,
```
justinlin-linkedin, line 257: We should treat ReplicateBlob as a special case for replica pool. If we are here, then it's likely that two replicas in colo datacenter are down. We should probably prioritize remote datacenter in the replica pool.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
```diff
@@ -0,0 +1,490 @@
...
+  private final FutureResult<Void> futureResult;
+  private final Callback<Void> callback;
+  private final Time time;
+  private final NonBlockingRouterMetrics routerMetrics;
+  private final long submissionTimeMs;
+
+  // The operation tracker that tracks the state of this operation.
+  private final OperationTracker operationTracker;
+  // A map used to find inflight requests using a correlation id.
+  private final Map<Integer, RequestInfo> ReplicateBlobRequestInfos;
```
justinlin-linkedin, line 61: variable name should start with lower case letter.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
```diff
@@ -0,0 +1,490 @@
...
+  // The result of this operation to be set into FutureResult.
+  private final Void operationResult = null;
+  private final QuotaChargeCallback quotaChargeCallback;
```
justinlin-linkedin, line 64: I have different opinion on this. ReplicateBlob is an internal operation, it should not subject to quota charge.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
```diff
@@ -0,0 +1,490 @@
...
+      if (operationTracker.hasSucceeded()) {
+        operationException.set(null);
+      } else if (operationTracker.maybeFailedDueToOfflineReplicas()) {
+        operationException.set(
+            new RouterException("ReplicateBlobOperation failed possibly because some replicas are unavailable",
+                RouterErrorCode.AmbryUnavailable));
+      } else if (operationTracker.hasFailedOnNotFound()) {
+        operationException.set(new RouterException("ReplicateBlobOperation failed because of BlobNotFound",
+            RouterErrorCode.BlobDoesNotExist));
+      }
```
justinlin-linkedin, line 367: ReplicateBlobOperation should not care about the offline replica and should not care about the not found error.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: Good catch. Thank you.
Due to NonBlockRouterTest set routerConfig.routerGetEligibleReplicasByStateEnabled to false, so all my test works as expected. It uses all the 9 replicas and parallelism is 2.

Will fix it and add test coverage of NonBlockingRouterTest.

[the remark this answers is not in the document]
JingQianCloud: Got it. Fixed.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: Intended to use successTarget as 2.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
JingQianCloud: Just want to know the code a little bit more.
ReplicateBlob is possible to hit Not Found or unavailable as well. 
Although the operationTracker.operationException should have the corresponding error already, looks like to me, there is no harm to have these two conditions? Can you explain it a little bit more?

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
JingQianCloud: Sorry. Accident caused by replacement. Fixed.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
JingQianCloud: Removed Quota code.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: Can we fine tune it later? For simplicity, for now, we pick the local first and then remote. I've added TODO in the second revision.

ambry-commons/src/main/java/com/github/ambry/commons/LoggingNotificationSystem.java
```diff
@@ -69,7 +70,16 @@ public void onBlobUndeleted(String blobId, String serviceId, Account account, Co
         ", " + serviceId + ", accountName " + (account == null ? null : account.getName()) + ", accountId" + (
             account == null ? null : account.getId()) + ", containerName " + (container == null ? null
             : container.getName()) + ", containerId " + (container == null ? null : container.getId()));
+  }
 
+  @Override
+  public void onBlobReplicated(String blobId, String serviceId, Account account, Container container,
+      DataNodeId sourceHost) {
+    logger.debug("onBlobReplicated {}", blobId,
```
justinlin-linkedin, line 78: this string format is super weird. Could you please fix it, just like the string format in onBlobCreated. And BTW, we should probably fix the onBlobDeleted and onBlobUndeleted.

JingQianCloud: Sure. Fixed the code format of these functions.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -248,6 +248,19 @@ class SimpleOperationTracker implements OperationTracker {
         // Here set the success target to the number of eligible replicas.
         replicaSuccessTarget = eligibleReplicas.size();
         break;
+      case ReplicateBlobOperation:
+        // Replicate one blob. Use the same "success target" and "request parallelism" config settings as PutOperation.
+        // But ReplicateBlobOperation sets crossColoEnabled to true.
+        // ON_DEMAND_REPLICATION_TODO: we may consider add its own configure setting if needed.
+        // ON_DEMAND_REPLICATION_TODO: may tune the order of the replica pool. Currently still local first and then remote.
+        // a. right now source HostNode is included. Probably will exclude it.
```
justinlin-linkedin, line 256: The source host should be excluded, otherwise, we set up to at least have one success from the beginning.
If the source host is not excluded, then when the SimpleOperationTracker hits the source host, we know it's going to be a success. This would be equivalent to subtracting 1 from success target.

justinlin-linkedin:
BTW, in AmbryServerRequests, when handling ReplicateBlobRequest, we should check if the local store already has the blob first, before sending request to remote server to copy the blob. But that would be another PR.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobManager.java
```diff
@@ -0,0 +1,210 @@
...
+import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.commons.BlobId;
+import com.github.ambry.commons.Callback;
+import com.github.ambry.commons.ResponseHandler;
+import com.github.ambry.config.RouterConfig;
+import com.github.ambry.network.RequestInfo;
+import com.github.ambry.network.ResponseInfo;
+import com.github.ambry.notification.NotificationSystem;
+import com.github.ambry.protocol.ReplicateBlobResponse;
+import com.github.ambry.quota.QuotaChargeCallback;
```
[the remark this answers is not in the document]
JingQianCloud: Removed.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobManager.java
```diff
@@ -0,0 +1,210 @@
...
+  private final NotificationSystem notificationSystem;
+  private final Time time;
+  private final ResponseHandler responseHandler;
+  private final AccountService accountService;
+  private final NonBlockingRouterMetrics routerMetrics;
+  private final ClusterMap clusterMap;
+  private final RouterConfig routerConfig;
+  private final NonBlockingRouter nonBlockingRouter;
+
+  private static final Logger logger = LoggerFactory.getLogger(ReplicateBlobManager.class);
```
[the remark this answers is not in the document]
JingQianCloud: Removed.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: I use the same parallelism and success target as Put. So parallelism is 3 and success target is 2. Basically it's retrying the Put Request and meet the Put quorum requirement. 
So even source host is included, we still will reach at least 2. 

But as you suggested, another way I can do is, exclude source host but set its configurations like success target as 1.

I'm ok with both but I prefer the current solution a little more. Because it's synced with Put requirement.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
```diff
@@ -0,0 +1,449 @@
...
+
+
+/**
+ * This class manages the internal state of a {@code ReplicateBlobOperation} during its life cycle.
+ */
+class ReplicateBlobOperation {
+  private static final Logger logger = LoggerFactory.getLogger(ReplicateBlobOperation.class);
+  //Operation arguments
+  private final RouterConfig routerConfig;
+  private final ResponseHandler responseHandler;
```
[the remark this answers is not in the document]
JingQianCloud: > BTW, in AmbryServerRequests, when handling ReplicateBlobRequest, we should check if the local store already has the blob first, before sending request to remote server to copy the blob. But that would be another PR.

Good suggestion. I created one spring task to keep track it.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
justinlin-linkedin: ok, we can try this out and come back later.

justinlin-linkedin (approved):

ambry-router/src/test/java/com/github/ambry/router/MockServer.java
```diff
@@ -445,6 +471,25 @@ private UndeleteResponse makeUndeleteResponse(UndeleteRequest undeleteRequest, S
...
+   * *
+   * @return the constructed {@link ReplicateBlobResponse}
+   */
+  ReplicateBlobResponse makeReplicateBlobResponse(ReplicateBlobRequest replicateBlobRequest,
+      ServerErrorCode replicateBlobError) {
+    if (replicateBlobError == ServerErrorCode.No_Error) {
+      replicateBlobError =
+          errorCodeForBlobs.getOrDefault(replicateBlobRequest.getBlobId().getID(), ServerErrorCode.No_Error);
+    }
+    if (replicateBlobError == ServerErrorCode.No_Error) {
```
SophieGuo410, line 486: Can we combine them into one if condition?

ambry-router/src/test/java/com/github/ambry/router/MockServer.java
```diff
@@ -445,6 +471,25 @@ private UndeleteResponse makeUndeleteResponse(UndeleteRequest undeleteRequest, S
...
+      updateBlobMap(replicateBlobRequest);
```
SophieGuo410, line 487: Can we combine them together?

SophieGuo410 (approved):
lgtm. Just left one minor comment.

ambry-router/src/test/java/com/github/ambry/router/MockServer.java
JingQianCloud: Inside the first if block, we'll re-assign the replicateBlobError again. Since replicateBlobError is changed, if I combine these two, I still need add the inner if condition to test the replicateBlobError again.
Seems it won't have much difference. I hope it's ok that I don't change it.



if (replicateBlobError == ServerErrorCode.No_Error) {
      replicateBlobError =
          errorCodeForBlobs.getOrDefault(replicateBlobRequest.getBlobId().getID(), ServerErrorCode.No_Error);
}
if (replicateBlobError == ServerErrorCode.No_Error) {
      updateBlobMap(replicateBlobRequest);
}

PR #2247, opened 2022-10-05
Named blob update api implementation

mikeyang923:
Named blob update api implementation:
1. Create a new blob with temp TTL
2. Insert a row into MySQL named_blobs_v2 db with state='IN_PROGRESS' 
3. Update Blob TTL to infinity
4. Insert a MySQL row with state to be 'READY'

Will schedule a repeating job (eg daily) to 1. try update Blob TTL for active 'IN_PROGRESS' rows and 2. clean up the outdated 'IN_PROGRESS' mysql rows.


ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -55,14 +56,22 @@ public class MySqlNamedBlobDbConfig {
    * A flag on whether to turn on DB Transition run
    */
   @Config(DB_TRANSITION)
-  @Default("false")
+  @Default("true")
```
panchow, line 59: You really don't need to change the @Default. it is not used at all.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -55,14 +56,22 @@ public class MySqlNamedBlobDbConfig {
...
   public final boolean dbTransition;
 
+  /**
+   * A flag on whether to use new table data
+   */
+  @Config(DB_NEW_TABLE)
+  @Default("false")
```
panchow, line 66: Remove the default.  It is not used at all.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -55,14 +56,22 @@ public class MySqlNamedBlobDbConfig {
...
+  public final boolean dbNewtable;
+
   public MySqlNamedBlobDbConfig(VerifiableProperties verifiableProperties) {
     this.dbInfo = verifiableProperties.getString(DB_INFO);
     this.localPoolSize = verifiableProperties.getIntInRange(LOCAL_POOL_SIZE, 5, 1, Integer.MAX_VALUE);
     this.remotePoolSize = verifiableProperties.getIntInRange(REMOTE_POOL_SIZE, 1, 1, Integer.MAX_VALUE);
     this.listMaxResults = verifiableProperties.getIntInRange(LIST_MAX_RESULTS, 100, 1, Integer.MAX_VALUE);
     this.dbTransition = verifiableProperties.getBoolean(DB_TRANSITION, false);
```
panchow, line 74: Do you intend to change the default to true (based on the @Default above).

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -124,6 +124,11 @@ public static final class Headers {
      * has been modified after the value set for this header.
      */
     public static final String IF_MODIFIED_SINCE = "If-Modified-Since";
+    /**
+     * Header to be set by clients during a Put blob call to clearly specify that the client is aware of the
+     * UPDATE feature of named blob, and client wants to update when the named blob already exist.
+     */
+    public static final String PUT_UPDATE = "x-ambry-update";
```
panchow, line 131: I think it would help to be more specific.  There's already HTTP PUT verb.  As a user, I wouldn't know when to set this header and what value to set to.  Is this the name of the named blob?  If so, consider "x-ambry-blob-name"?

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -806,6 +811,16 @@ public static boolean isChunkUpload(Map<String, Object> args) throws RestService
     return getBooleanHeader(args, Headers.CHUNK_UPLOAD, false);
   }
 
+  /**
+   * Determine if {@link Headers#PUT_UPDATE} is set in the request args.
+   * @param args The request arguments.
+   * @return {@code true} if {@link Headers#PUT_UPDATE} is set.
+   * @throws RestServiceException if exception occurs during parsing the arg.
+   */
+  public static boolean isUpsert(Map<String, Object> args) throws RestServiceException {
```
panchow, line 820: BTW, RESTful PUT is supposed to be insert-if-missing and update-if-exists. That means PUT itself means upsert.  Why do we need an extra header to indicate that?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -256,9 +257,9 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
         new GetTransactionStateTracker(remoteDatacenters, localDatacenter);
     return executeTransactionAsync(accountName, containerName, true, (accountId, containerId, connection) -> {
       long startTime = System.currentTimeMillis();
-      NamedBlobRecord record, recordV2;
+      NamedBlobRecord record, recordV2 = null;
       record = run_get(accountName, containerName, blobName, option, accountId, containerId, connection);
-      if (config.dbTransition) {
+      if (config.dbTransition || config.dbNewtable) {
```
panchow, line 262: nit: can we use dbNewTable?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -256,9 +257,9 @@ public CompletableFuture<NamedBlobRecord> get(String accountName, String contain
         new GetTransactionStateTracker(remoteDatacenters, localDatacenter);
     return executeTransactionAsync(accountName, containerName, true, (accountId, containerId, connection) -> {
       long startTime = System.currentTimeMillis();
-      NamedBlobRecord record, recordV2;
+      NamedBlobRecord record, recordV2 = null;
       record = run_get(accountName, containerName, blobName, option, accountId, containerId, connection);
```
panchow, line 261: nit: NamedBlobRecord record = (in one line)

panchow (changes requested):

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
mikeyang923: Sounds good, removed.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
mikeyang923: Updated, so by default the update for named-blob will be upsert. User can specify a header 'x-ambry-named-not-upsert' to disable this default upsert (see the latest commit).

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
mikeyang923: Yeah, updated. By default the update for named-blob will be upsert. User can specify a header 'x-ambry-named-not-upsert' to disable this default upsert (see the latest commit).

panchow (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -175,7 +176,13 @@ private CompletionStage<String> convertId(String input, RestRequest restRequest,
...
         NamedBlobRecord record = new NamedBlobRecord(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
             namedBlobPath.getBlobName(), blobId, expirationTimeMs);
-        conversionFuture = getNamedBlobDb().put(record).thenApply(result -> result.getInsertedRecord().getBlobId());
+        NamedBlobState state;
+        if (RestUtils.isUpsertForNamedBlob(restRequest.getArgs()) && properties.getTimeToLiveInSeconds() == Utils.Infinite_Time) {
+          state = NamedBlobState.IN_PROGRESS;
+        } else {
+          state = NamedBlobState.READY;
+        }
+        conversionFuture = getNamedBlobDb().put(record, state).thenApply(result -> result.getInsertedRecord().getBlobId());
```
snalli, line 185:
```
        NamedBlobState state = NamedBlobState.READY;
        if (RestUtils.isUpsertForNamedBlob(restRequest.getArgs()) && properties.getTimeToLiveInSeconds() == Utils.Infinite_Time) {
          state = NamedBlobState.IN_PROGRESS;
        }
```

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -276,10 +282,19 @@ private boolean isRetriable(Throwable throwable) {
...
      * request time security checks that rely on the request being fully parsed and any additional arguments set.
      * @param blobInfo the {@link BlobInfo} to use for security checks.
+     * @param blobId the {@link String} to use for blob id.
      * @return a {@link Callback} to be used with {@link Router#updateBlobTtl(String, String, long)}.
      */
-    private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo) {
-      return buildCallback(frontendMetrics.updateBlobTtlRouterMetrics, blobId -> {
+    private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo, String blobId) {
+      return buildCallback(frontendMetrics.updateBlobTtlRouterMetrics, convertedBlobId -> {
+        if (RestUtils.isUpsertForNamedBlob(restRequest.getArgs()) &&
```
snalli, line 290: just make a helper since this pattern appears twice
```
boolean isUpsertForPermNamedBlob() {
 return RestUtils.isUpsertForNamedBlob(restRequest.getArgs()) &&
            blobInfo.getBlobProperties().getTimeToLiveInSeconds() == Utils.Infinite_Time;
}
```

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
mikeyang923: Good point, updated.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -124,6 +124,11 @@ public static final class Headers {
      * has been modified after the value set for this header.
      */
     public static final String IF_MODIFIED_SINCE = "If-Modified-Since";
+    /**
+     * Header to be set by clients during a Put blob call to clearly specify that the client is aware of the
+     * UPDATE feature of named blob, and client does NOT wants to update when the named blob already exist.
+     */
+    public static final String NAMED_NOT_UPSERT = "x-ambry-named-not-upsert";
```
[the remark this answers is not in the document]
mikeyang923: We agreed to default to be upsert. Since the '`getBooleanHeader`' method was defaulting to 'false', so I used this 'NAMED_NOT_UPSERT' flag. I agree that double negative is confusing, so I just updated it to be positive flag, and also updated '`getBooleanHeader`' method to accept a 'defaultVal' (in the isUpsertForNamedBlob case, the defaultVal is true).

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -71,7 +70,7 @@ public MySqlNamedBlobDbConfig(VerifiableProperties verifiableProperties) {
     this.localPoolSize = verifiableProperties.getIntInRange(LOCAL_POOL_SIZE, 5, 1, Integer.MAX_VALUE);
     this.remotePoolSize = verifiableProperties.getIntInRange(REMOTE_POOL_SIZE, 1, 1, Integer.MAX_VALUE);
     this.listMaxResults = verifiableProperties.getIntInRange(LIST_MAX_RESULTS, 100, 1, Integer.MAX_VALUE);
-    this.dbTransition = verifiableProperties.getBoolean(DB_TRANSITION, false);
+    this.dbTransition = verifiableProperties.getBoolean(DB_TRANSITION, true);
```
snalli, line 73: is this used somewhere in this PR ? I'd suggest putting this in a separate PR if not used just for easy reverts.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -648,7 +653,7 @@ public static GetBlobOptions buildGetBlobOptions(Map<String, Object> args, SubRe
       throw new RestServiceException("Ranges not supported for sub-resources that aren't Segment.",
           RestServiceErrorCode.InvalidArgs);
     }
-    boolean resolveRangeOnEmptyBlob = getBooleanHeader(args, Headers.RESOLVE_RANGE_ON_EMPTY_BLOB, false);
+    boolean resolveRangeOnEmptyBlob = getBooleanHeader(args, Headers.RESOLVE_RANGE_ON_EMPTY_BLOB, false, false);
```
snalli, line 656: how is this related to named blob upsert ? curious

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
mikeyang923: Ok, I reverted this one.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
mikeyang923: So I changed 'getBooleanHeader' to accept a defaultVal, and need to pass false here to be backward-compatible. Also see https://github.com/linkedin/ambry/pull/2247#discussion_r1018249962

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -56,7 +56,6 @@ public class MySqlNamedBlobDbConfig {
    * A flag on whether to turn on DB Transition run
    */
   @Config(DB_TRANSITION)
-  @Default("false")
```
snalli, line 59: revert

mikeyang923: Those Default are not used at all. It's ok to clean up. See
```
  public MySqlNamedBlobDbConfig(VerifiableProperties verifiableProperties) {
    this.dbInfo = verifiableProperties.getString(DB_INFO);
    this.localPoolSize = verifiableProperties.getIntInRange(LOCAL_POOL_SIZE, 5, 1, Integer.MAX_VALUE);
    this.remotePoolSize = verifiableProperties.getIntInRange(REMOTE_POOL_SIZE, 1, 1, Integer.MAX_VALUE);
    this.listMaxResults = verifiableProperties.getIntInRange(LIST_MAX_RESULTS, 100, 1, Integer.MAX_VALUE);
    this.dbTransition = verifiableProperties.getBoolean(DB_TRANSITION, false);
    this.dbRelyOnNewTable = verifiableProperties.getBoolean(DB_RELY_ON_NEW_TABLE, false);
  }
```

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
snalli: nvm, i see this is a side effect of the named-upsert flag

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -803,7 +808,29 @@ public static boolean isPrivate(Map<String, Object> args) throws RestServiceExce
...
+    return getBooleanHeader(args, Headers.CHUNK_UPLOAD, false, false);
+  }
+
+  /**
+   * Determine if {@link Headers#NAMED_UPSERT} is set in the request args.
+   * @param args The request arguments.
+   * @return {@code false} if {@link Headers#NAMED_UPSERT} is set to false else return true.
+   * @throws RestServiceException if exception occurs during parsing the arg.
+   */
+  public static boolean isUpsertForNamedBlob(final Map<String, Object> args) throws RestServiceException {
```
snalli, line 820: private

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -803,7 +808,29 @@ public static boolean isPrivate(Map<String, Object> args) throws RestServiceExce
...
+    return getBooleanHeader(args, Headers.NAMED_UPSERT, false, true);
```
snalli, line 821: So the side effect of assuming the default as true is that even if the user does not pass it, we end up overwriting their old named blob. Is this the agreed upon contract ?

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -803,7 +808,29 @@ public static boolean isPrivate(Map<String, Object> args) throws RestServiceExce
...
+  }
```
snalli, line 822: MAke a new method to avoid this code cascade

```
  public static boolean getOptionalBooleanHeader(Map<String, Object> args, String header, boolean defaultVal)
      throws RestServiceException {
    String stringValue = getHeader(args, header, false);
    if (stringValue == null) {
      return defaultVal;
    } else if ("false".equalsIgnoreCase(stringValue)) {
      return false;
    } else if ("true".equalsIgnoreCase(stringValue)) {
      return true;
    } else {
      throw new RestServiceException(
          header + "[" + stringValue + "] has an invalid value (allowed values: true, false)",
          RestServiceErrorCode.InvalidArgs);
    }
  }
```

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
snalli: or get rid of it since there is only 1 usage

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
```diff
@@ -125,7 +125,7 @@ public String getSignedUrl(RestRequest restRequest) throws RestServiceException
             maxUploadSize = RestUtils.getLongHeader(args, RestUtils.Headers.MAX_UPLOAD_SIZE, true);
             break;
           case RestUtils.Headers.CHUNK_UPLOAD:
-            chunkUpload = RestUtils.getBooleanHeader(args, RestUtils.Headers.CHUNK_UPLOAD, true);
+            chunkUpload = RestUtils.getBooleanHeader(args, RestUtils.Headers.CHUNK_UPLOAD, true, false);
```
snalli, line 128: ifu make a new method getOptionalBooleanHeader, we can avoid this cascade

snalli (changes requested):
requesting changes

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
snalli: most of the cascade makes it hard to read and unrelated to namedblobs

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
mikeyang923: It's created per your previous comment: https://github.com/linkedin/ambry/pull/2247#discussion_r1017462532
Since it was used in two different scripts, it has to to public?

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
mikeyang923: Yes, that's what the default to 'upsert' mean.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
mikeyang923: Ok, i removed it in latest commit.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
mikeyang923: I think my current implementation is better due to:
1. Better to be explicit: the previous way assume the defaultVal is False (when the header is not set). My current way make it explicit for that. And allow user to set defaultVal to true when needed (for upsert case).
2. Avoid code/logic duplication: With your proposed new method, there will be duplicate code.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
mikeyang923: I see this as more about being-explicit, not about code cascade. This method is simple enough (just added one more parameter to be explicit about defaultVal), and not that hard to read comparing to others.

mikeyang923: I updated this in latest commit. please take another look.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -919,6 +934,37 @@ public static boolean getBooleanHeader(Map<String, Object> args, String header,
...
+   * @param required if {@code true}, {@link RestServiceException} will be thrown if {@code header} is not present
+   *                 in {@code args}.
+   * @param defaultVal the default value for this header
+   * @return {@code true} if the header's value is {@code "true"} (case-insensitive), or {@code false} if the header's
+   *         value is {@code "false} (case-insensitive) or the header is not present and {@code required} is
+   *         {@code false}.
+   * @throws RestServiceException same as cases of {@link #getHeader(Map, String, boolean)} and if the value cannot be
+   *                              converted to a {@code boolean}.
+   */
+  public static boolean getBooleanHeaderWithDefaultValue(Map<String, Object> args, String header, boolean required, boolean defaultVal)
```
snalli, line 950: My point of making a new function was to 1) not cascade your changes into existing code path 2) also reduce the number of arguments to the function itself to make the code readable. 
I don't prefer names like XXwithYY because later on someone will add a new arg and call it XXwithYYandZZ etc. Once you name it `getOptionalBooleanHeader`, you automatically get rid of the redundant `required` arg and then passing a `default` value follows from the name. I hope going forward people will use it. The original method should have been called `getRequiredBooleanHeader` but that's history. But going forward, the names and function signatures should be short, meaningful and descriptive at a glance.

mikeyang923: Sounds good, updated it in latest commit, can you take another look?

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -806,6 +811,16 @@ public static boolean isChunkUpload(Map<String, Object> args) throws RestService
...
 
+  /**
+   * Determine if {@link Headers#NAMED_UPSERT} is set in the request args.
+   * @param args The request arguments.
+   * @return {@code false} if {@link Headers#NAMED_UPSERT} is set to false else return true.
+   * @throws RestServiceException if exception occurs during parsing the arg.
+   */
+  public static boolean isUpsertForNamedBlob(final Map<String, Object> args) throws RestServiceException {
+    return getOptionalBooleanHeader(args, Headers.NAMED_UPSERT, true);
+  }
```
snalli, line 822: The PR looks much better now that i can focus on your changes and not be distracted by all the other other irrelevant code.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -241,21 +241,23 @@ public CompletableFuture<Page<NamedBlobRecord>> list(String accountName, String
...
-            recordCurrent = run_get_v2(record.getAccountName(), record.getContainerName(), record.getBlobName(),
-                GetOption.None, accountId, containerId, connection);
-          } catch (RestServiceException e) {
-            logger.trace("Skip exception in pulling data from db: accountId='{}', containerId='{}', blobName='{}': {}",
-                accountId, containerId, record.getBlobName(), e);
-          }
-          if (recordCurrent != null) {
-            throw buildException("PUT: Blob still alive", RestServiceErrorCode.Conflict, record.getAccountName(),
-                record.getContainerName(), record.getBlobName());
+          if (!(config.dbRelyOnNewTable && isUpsert)) {
```
snalli, line 248: What is the newTable ? Is this a new mysql table with new schema for named-blob-update ?

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -276,10 +283,17 @@ private boolean isRetriable(Throwable throwable) {
      * After TTL update finishes, call {@link SecurityService#postProcessRequest} to perform
      * request time security checks that rely on the request being fully parsed and any additional arguments set.
      * @param blobInfo the {@link BlobInfo} to use for security checks.
+     * @param blobId the {@link String} to use for blob id.
      * @return a {@link Callback} to be used with {@link Router#updateBlobTtl(String, String, long)}.
      */
-    private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo) {
-      return buildCallback(frontendMetrics.updateBlobTtlRouterMetrics, blobId -> {
+    private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo, String blobId) {
+      return buildCallback(frontendMetrics.updateBlobTtlRouterMetrics, convertedBlobId -> {
```
snalli, line 290: where is convertedBlobId used in this callback ?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli, line 248: Also some comments in your code would help understand what is happening. For eg. this path is to be executed for regular named-blob puts or named-blob updates that use old table.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -276,10 +283,17 @@ private boolean isRetriable(Throwable throwable) {
...
+        String blobIdClean = RestUtils.stripSlashAndExtensionFromId(blobId);
+        NamedBlobPath namedBlobPath = NamedBlobPath.parse(RestUtils.getRequestPath(restRequest), restRequest.getArgs());
+        NamedBlobRecord record = new NamedBlobRecord(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
+            namedBlobPath.getBlobName(), blobIdClean, Utils.Infinite_Time);
+        namedBlobDb.put(record, NamedBlobState.READY, RestUtils.isUpsertForNamedBlob(restRequest.getArgs())).get();
```
snalli, line 295: comments

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -175,7 +176,11 @@ private CompletionStage<String> convertId(String input, RestRequest restRequest,
             Utils.addSecondsToEpochTime(properties.getCreationTimeInMs(), properties.getTimeToLiveInSeconds());
         NamedBlobRecord record = new NamedBlobRecord(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
             namedBlobPath.getBlobName(), blobId, expirationTimeMs);
-        conversionFuture = getNamedBlobDb().put(record).thenApply(result -> result.getInsertedRecord().getBlobId());
+        NamedBlobState state = NamedBlobState.READY;
+        if (properties.getTimeToLiveInSeconds() == Utils.Infinite_Time) {
+          state = NamedBlobState.IN_PROGRESS;
```
snalli, line 181: so when does the record go from in_progress to ready ? after the ttl-update ?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Yes, we created a new table with new schema (added 'version' and 'state' fields).

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
mikeyang923: Think this is a placeholder for the input parameter.
Look at the previous code, it was below. It use 'blobId' as the input parameter which is not used here as well. I just changed that 'blobId' to 'convertedBlobId' to avoid the name conflicts.
```
private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo) {
      return buildCallback(frontendMetrics.updateBlobTtlRouterMetrics, blobId -> {
   securityService.processResponse(restRequest, restResponseChannel, blobInfo, securityProcessResponseCallback());
      }, uri, LOGGER, finalCallback);
    }
```

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done in latest commit, pls take another look.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
mikeyang923: Done in latest commit, pls take another look.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
mikeyang923: In the ttl-update success callback, added comment about it in latest commit.

snalli (approved):

PR #2248, opened 2022-10-06
Changes to use Helix aggregated view cluster

Arun-LinkedIn:
This PR contains changes needed to use Helix aggregated view cluster (which should be available locally in each data center) for getting global cluster information instead of fetching individual data center information from each of the helix/ZK clusters. 

This should speed up the start time frontends and servers since cluster information would be fetched from local Zookeeper servers instead of fetching from multiple remote Zookeepers.


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -646,24 +679,37 @@ public String getResourceNameForPartition(AmbryPartition partition) {
...
+            String resourceName = partitionToResourceNameByDc.get(dc).get(partition.toPathString());
+            dcToRoutingTableSnapshotRef.get(dcInfo.dcName)
+                .get()
+                .getInstancesForResource(resourceName, partition.toPathString(), state.name())
+                .stream()
+                .map(instanceConfig -> instanceNameToAmbryDataNode.get(instanceConfig.getInstanceName()))
+                .map(dataNode -> ambryDataNodeToAmbryReplicas.get(dataNode).get(partition.toPathString()))
+                .filter(Objects::nonNull)
+                .forEach(replicas::add);
+          }
```
justinlin-linkedin, line 712: should probably merge these two method. 
This method can be
```
Map<ReplicaState, List<AmbryReplica>> result = new HashMap<>();
getReplicaIdByStates(result, partition, Collections.singleton(state), dcName);
return result.get(state);
```

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -646,24 +679,37 @@ public String getResourceNameForPartition(AmbryPartition partition) {
...
-          dcToRoutingTableSnapshotRef.get(dcInfo.dcName)
-              .get()
-              .getInstancesForResource(resourceName, partition.toPathString(), state.name())
-              .stream()
-              .map(instanceConfig -> instanceNameToAmbryDataNode.get(instanceConfig.getInstanceName()))
-              .map(dataNode -> ambryDataNodeToAmbryReplicas.get(dataNode).get(partition.toPathString()))
-              .filter(Objects::nonNull)
-              .forEach(replicas::add);
+      if (clusterMapConfig.clusterMapUseAggregatedView) {
+        String resourceName = dcName != null ? partitionToResourceNameByDc.get(dcName).get(partition.toPathString())
```
justinlin-linkedin, line 689: does this mean that all datacenters have the same resourceName for a partition. I suppose they should be the same, we only get on resourceName for each partition in clustermap.

justinlin-linkedin (commented):
lgtm, is there anyway we can write some tests?

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
```diff
@@ -23,6 +23,8 @@
 public class ClusterMapConfig {
 
   public static final String CLUSTERMAP_CLUSTER_NAME = "clustermap.cluster.name";
+  public static final String CLUSTERMAP_AGGREGATED_VIEW_CLUSTER_NAME = "clustermap.aggregated.view.cluster.name";
+  public static final String CLUSTERMAP_USE_AGGREGATED_VIEW = "clustermap.use.aggregated.view";
```
panchow, line 27: Can we switch to use the aggregated view only?  Is there reason why we want to keep the old way?

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
```diff
@@ -154,6 +156,20 @@ public class ClusterMapConfig {
   @Config(CLUSTERMAP_CLUSTER_NAME)
   public final String clusterMapClusterName;
 
+  /**
+   * The name of the associated aggregated view cluster for this node.
+   */
+  @Config(CLUSTERMAP_AGGREGATED_VIEW_CLUSTER_NAME)
+  @Default("")
```
panchow, line 163: You don't need to set the default.  It's not used.  Same for the default below.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
```diff
@@ -0,0 +1,198 @@
...
+
+    // Ambry currently needs and listens to following changes in Helix/ZKs in all datacenters:
+    // 1. External view (indirectly via using RoutingTableProvider)
+    // 2. Live instances (List of live hosts in cluster)
+    // 3. HelixPropertyStore. This is where Ambry data node configuration like disk information, disk to replica mapping,
+    //    replica sealed states, etc are stored.
+    // 4. Helix ideal states. This is needed to keep track of Helix Resource (Helix terminology) to Ambry partitions
+    //    (Multiple Ambry partitions are grouped under one helix resource).
+
+    // As we can see above, since #3 and #4 are not aggregated by Helix currently, we get these information by talking
```
panchow, line 144: Can we ask the Helix team to bring #3 and #4 to the aggregate view too?  If we still need to talk to all ZK in all DC, how much improvement are we getting?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
```diff
@@ -0,0 +1,198 @@
...
+    // 2. Instance configs (These mostly contain fields set and used by Helix like host name and port).
+    // 3. Live instances (List of live hosts in cluster).
+
+    // Ambry currently needs and listens to following changes in Helix/ZKs in all datacenters:
+    // 1. External view (indirectly via using RoutingTableProvider)
+    // 2. Live instances (List of live hosts in cluster)
+    // 3. HelixPropertyStore. This is where Ambry data node configuration like disk information, disk to replica mapping,
+    //    replica sealed states, etc are stored.
+    // 4. Helix ideal states. This is needed to keep track of Helix Resource (Helix terminology) to Ambry partitions
+    //    (Multiple Ambry partitions are grouped under one helix resource).
```
panchow, line 142: Good documentation.  I like it.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
```diff
@@ -0,0 +1,198 @@
...
+   */
+  public HelixAggregatedViewClusterInfo initializeHelixAggregatedViewCluster() throws Exception {
+    // We can assume that Helix Aggregated view cluster will be present in same zookeeper hosting regular Ambry cluster
+    // information. Get the HelixManager to talk to the local zk service.
+    DcZkInfo localDcZkInfo = dataCenterToZkAddress.get(clusterMapConfig.clusterMapDatacenterName);
+
+    // For now, the first ZK endpoint (if there are more than one endpoints) will be adopted by default for initialization.
+    // Note that, Ambry currently doesn't support multiple spectators, because there should be only one source of truth.
+    String localZkConnectStr = localDcZkInfo.getZkConnectStrs().get(0);
+    HelixManager aggregatedViewManager =
```
panchow, line 122: nit: can we just call this helixManager?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
```diff
@@ -0,0 +1,198 @@
...
+    this.selfInstanceName = selfInstanceName;
+    this.dataNodeConfigSourceMetrics = dataNodeConfigSourceMetrics;
+    this.helixClusterManager = helixClusterManager;
+  }
+
+  /**
+   * Begin the startup procedure in a background thread.
+   */
+  public void start() {
+    Utils.newThread(() -> {
```
panchow, line 77: There is only 1 instance of HelixAggregatedViewClusterInitializer, no need to create another thread.  Old approach created multiple initializers and ran them in parallel, then joined the threads.  That's why the old approach used threads.  For this new approach, it creates 1 thread and joins, there is no benefit for using a thread so just remove it.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -461,10 +486,14 @@ public void registerClusterMapListener(ClusterMapChangeListener clusterMapChange
    */
   @Override
   public void close() {
-    for (DcInfo dcInfo : dcToDcInfo.values()) {
-      dcInfo.close();
+    if (clusterMapConfig.clusterMapUseAggregatedView) {
+      helixAggregatedViewClusterInfo.close();
+    } else {
+      for (DcInfo helixDcInfo : dcToDcInfo.values()) {
+        helixDcInfo.close();
```
panchow, line 493: Even clusterMapUseAggregatedView is true, dcToDCInfo map is still populated.  Do you still need to loop through DCInfo and call close?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -451,8 +472,12 @@ public ReplicaId getBootstrapReplica(String partitionIdStr, DataNodeId dataNodeI
 
   @Override
   public void registerClusterMapListener(ClusterMapChangeListener clusterMapChangeListener) {
-    for (DcInfo dcInfo : dcToDcInfo.values()) {
-      dcInfo.clusterChangeHandler.registerClusterMapListener(clusterMapChangeListener);
+    if (clusterMapConfig.clusterMapUseAggregatedView) {
+      helixAggregatedViewClusterInfo.clusterChangeHandler.registerClusterMapListener(clusterMapChangeListener);
+    } else {
+      for (DcInfo helixDcInfo : dcToDcInfo.values()) {
+        helixDcInfo.clusterChangeHandler.registerClusterMapListener(clusterMapChangeListener);
```
panchow, line 479: What does registerClusterMapListener listen to?  In the above #1 to #4, if it listens to #3 and #4, does it needs to listen to changes in each DC?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -646,24 +679,37 @@ public String getResourceNameForPartition(AmbryPartition partition) {
...
-              .forEach(replicas::add);
+      if (clusterMapConfig.clusterMapUseAggregatedView) {
+        String resourceName = dcName != null ? partitionToResourceNameByDc.get(dcName).get(partition.toPathString())
+            : getResourceNameForPartition(partition);
+        globalRoutingTableSnapshotRef.get()
+            .getInstancesForResource(resourceName, partition.toPathString(), state.name())
+            .stream()
+            .map(instanceConfig -> instanceNameToAmbryDataNode.get(instanceConfig.getInstanceName()))
+            .filter(dataNode -> dcName == null || dataNode.getDatacenterName().equals(dcName))
+            .map(dataNode -> ambryDataNodeToAmbryReplicas.get(dataNode).get(partition.toPathString()))
```
panchow, line 696: Cache partition.toPathString() in variable.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
panchow, line 689: Just note that this logic is slightly different from old logic.  It may be correct according to new logic, but just calling it out.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
panchow: +1 in combining.  The method below can for(ReplicateState : states) call above method.

panchow (changes requested):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: Thank you for pointing it out. Actually it looks like it is not guaranteed that partitions would be mapped to same resource in each data center (found some examples in zookeeper). Made changes to go through resource names of all DCs and get replicas accordingly.

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
Arun-LinkedIn: Sure. I was thinking that we could have this config for now as an option to rollback in case we see any issues. We can probably remove it eventually?

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
Arun-LinkedIn: Sure. Updated

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
Arun-LinkedIn: Thank you for pointing it out. I will remove the thread. Had it as a TODO item but forgot about it.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
Arun-LinkedIn: Sure, will check with them. Based on earlier discussion with them, it looks like one of the reasons property store (#3) is not aggregated right now is that it has custom application data (and not helix data). So the folder structure/fields is probably not guaranteed to be same across all data centers. And helix team wanted to know on why we needed to aggregate ideal states (#4).

Right now, 70% of time is spent in fetching item routing table #1. So, we are still getting 70% improvement.

panchow (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
justinlin-linkedin: I am quite surprised that 70% of the time is spent on fetching routing table. Because routing table is the external view and external view has same amount of data to fetch comparing to IdealState. Why is that fetching IdealState is fast, but external view is slow?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -911,18 +968,31 @@ public void onDataNodeConfigChange(Iterable<DataNodeConfig> configs) {
...
+      String dcName;
+      if (isAggregatedViewHandler) {
+        // Since helix aggregated view service doesn't aggregate ideal states across colos, we register ideal state
+        // change listener with helix service in each colo separately. So, this callback would be invoked for each
+        // data center separately. In order to know which data center, this callback corresponds to, we try to get one
+        // of the host names from Ideal state and get its data center name.
+        IdealState state = idealState.iterator().next();
+        String partition = state.getPartitionSet().iterator().next();
+        String instanceName = state.getInstanceSet(partition).iterator().next();
+        dcName = instanceNameToAmbryDataNode.get(instanceName).getDatacenterName();
```
justinlin-linkedin, line 992: instanceNameToAmbryDataNode map is populated by DataNodeConfigSource, is there a way to guarantee the DataNodeConfigSource callback would be invoked before IdealState callback?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixAggregatedViewClusterInitializer.java
Arun-LinkedIn: My understanding is that for getting most up-to-date external view, helix would still have to go from host to host, check list of live hosts, etc and aggregate the current states of partitions in each host to generate external view. Due to that generation of external view takes time. 

Also, based on my discussion with helix team, it seems like there are two modes of instantiating a Routing Table. One is EXTERNAL_VIEW based and other is CURRENT_STATES based. For the former one, calculation of external view is done by helix controller and we wait until the most up-to-date view is generated. But, it has less read traffic (almost same data as ideal states) but has dependency on helix controller and is said to take longer time. For the latter one, calculation of external view (by going from host to host and getting partition states, etc) is done at the helix agent. Due to that, we avoid dependency on helix controller (and is supposed to be faster) but has more read traffic. We seem to use CURRENT_STATES based Routing Table. Even this takes time since we need to read all the data from remote Zookeepers currently.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: I think we make sure that DataNodeConfigSource callback is invoked in the same context of thread that adds listener by [waiting](https://github.com/linkedin/ambry/blob/a6422967b7f01255d208da9f5c5cdb503451b257/ambry-clustermap/src/main/java/com/github/ambry/clustermap/PropertyStoreToDataNodeConfigAdapter.java#L125) and fetching the information. Due to that, I think we can guarantee that `instanceNameToAmbryDataNode` is populated.

justinlin-linkedin: that's right, that makes sense

justinlin-linkedin (approved):

justinlin-linkedin (approved):

panchow (approved):

PR #2252, opened 2022-10-10
GetBlob Operations should call ResponseHandler on partition response error

justinlin-linkedin:
We are only calling ResponseHandler.onEvent on ResponseInfo's network error and the Response's server error code, but we didn't call onEvent on Partition response error for GetBlob operations.
This PR would start doing that.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
@@ -94,10 +94,10 @@ class GetBlobInfoOperation extends GetOperation {
    * @param nonBlockingRouter The non-blocking router object
    */
   GetBlobInfoOperation(RouterConfig routerConfig, NonBlockingRouterMetrics routerMetrics, ClusterMap clusterMap,
-      ResponseHandler responseHandler, BlobId blobId, GetBlobOptionsInternal options,
-      Callback<GetBlobResult> callback, RouterCallback routerCallback, KeyManagementService kms,
-      CryptoService cryptoService, CryptoJobHandler cryptoJobHandler, Time time, boolean isEncrypted,
-      QuotaChargeCallback quotaChargeCallback, NonBlockingRouter nonBlockingRouter) {
+      ResponseHandler responseHandler, BlobId blobId, GetBlobOptionsInternal options, Callback<GetBlobResult> callback,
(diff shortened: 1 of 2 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetBlobInfoOperation.java
```diff
@@ -300,6 +300,7 @@ void handleResponse(ResponseInfo responseInfo, GetResponse getResponse) {
   private void processGetBlobInfoResponse(RequestInfo getRequestInfo, GetResponse getResponse)
       throws IOException, MessageFormatException {
     ServerErrorCode getError = getResponse.getError();
+    responseHandler.onEvent(getRequestInfo.getReplicaId(), getError);
```
Arun-LinkedIn, line 303: Just wondering if we might be notifying response handler of server or partition error in [RouterUtils.extractResponseAndNotifyResponseHandler](https://github.com/linkedin/ambry/blob/1339c995283f00555173f820ed22e47ab15fbf1b/ambry-router/src/main/java/com/github/ambry/router/RouterUtils.java#L226).

justinlin-linkedin: That's the line for Response's ServerErrorCode. For example. DeleteResponse has a ServerErrorCode. After we deserialized a DeleteResponse, we will notify ResponseHandler with DeleteResponse.getError. The logic applies to all other Responses.
However, GetResponse is different, because it carries multiple layer of errors. First we have a GetResponse.getError, and for each partition, we have another error https://github.com/linkedin/ambry/blob/master/ambry-protocol/src/main/java/com/github/ambry/protocol/PartitionResponseInfo.java#L38. 
An in server, we already return an NoError in GetResponse. The real error for each partition is buried in the PartitionResponseInfo.

Arun-LinkedIn: Hmm, may be we might be taking care of error from PartitionResponseInfo as well by we call `RouterUtils.extractResponseAndNotifyResponseHandler` from [GetManager.handleResponse](https://github.com/linkedin/ambry/blob/1339c995283f00555173f820ed22e47ab15fbf1b/ambry-router/src/main/java/com/github/ambry/router/GetManager.java#L226)?

justinlin-linkedin: that's right, then I can close this one.

PR #2253, opened 2022-10-10
Add jackson suppport for MessageInfo

justinlin-linkedin:
Add jackson support for MessageInfo so we can serialize it to json format.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/store/MessageInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/MessageInfo.java
@@ -13,6 +13,10 @@
  */
 package com.github.ambry.store;
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonPOJOBuilder;
 import com.github.ambry.utils.Utils;
 import java.util.Objects;
 
@@ -20,6 +24,7 @@ import java.util.Objects;
 /**
  * A message info class that contains basic info about a message
  */
+@JsonDeserialize(builder=MessageInfo.Builder.class)
 public class MessageInfo {
 
   // The life version when the operation is triggered by the requests from frontend.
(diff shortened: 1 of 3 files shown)


ambry-api/src/main/java/com/github/ambry/store/MessageInfo.java
```diff
@@ -429,6 +442,7 @@ public Builder expirationTimeInMs(long expirationTimeInMs) {
      * @param isDeleted the isDeleted to set.
      * @return This builder.
      */
+    @JsonProperty("deleted")
```
JingQianCloud, line 445: I guess by default, Jackson renames primitive boolean field by removing 'is' and It'll become "deleted". Right?
Are you intended to use @JsonProperty("isDeleted")?

JingQianCloud (commented):
Overall looks good to me. A question embedded.

ambry-api/src/main/java/com/github/ambry/store/MessageInfo.java
justinlin-linkedin: yes. it's one way or another. we either annotate on MessageInfo or MessageInfo.Builder. I don't see there is a difference.

ambry-api/src/main/java/com/github/ambry/store/MessageInfo.java
```diff
@@ -439,6 +453,7 @@ public Builder isDeleted(boolean isDeleted) {
      * @param isTtlUpdated the isTtlUpdated to set.
      * @return This builder.
      */
+    @JsonProperty("ttlUpdated")
```
ankagrawal, line 456: The code changes look generally ok to me. But do we think that serializing and deserializing in JSON format is the optimal thing to do?
Currently, we are serializing only values and using the serialization order to deserialize. Using JSON as the serialization and deserialization format will add the field names as the key in the serialized format. These field names will be strings in all cases. This will lead to larger serialized data, and potentially need larger network bandwidth to transfer the same amount of information.

JingQianCloud (approved):

ambry-api/src/main/java/com/github/ambry/store/MessageInfo.java
justinlin-linkedin: this isn't going on critical read and write path. The purpose of this is for an admin command to show message info of a given blob id in each host.

ankagrawal (approved):

PR #2254, opened 2022-10-11
Apply compression at Frontend router.

panchow:
Apply compression at PutRequest per chunk.
Apply decompression at GetBlob per chunk.
Introduced a new version of Blob format V3 that add compression flag. Update other unit tests affected by this test.

### DIFF
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/BlobStoreHardDelete.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/BlobStoreHardDelete.java
@@ -233,7 +233,7 @@ class HardDeleteRecoveryMetadata {
     userMetadataVersion = stream.readShort();
     userMetadataSize = stream.readInt();
     blobRecordVersion = stream.readShort();
-    if (blobRecordVersion == Blob_Version_V2) {
+    if (blobRecordVersion >= Blob_Version_V2) {
       blobType = BlobType.values()[stream.readShort()];
     } else {
       blobType = BlobType.DataBlob;
@@ -280,7 +280,7 @@ class HardDeleteRecoveryMetadata {
     // create a byte array to hold the headerVersion + userMetadataVersion + userMetadataSize + blobRecordVersion +
     // blobType + blobRecordSize + storeKey.
     byte[] bytes = new byte[Version_Field_Size_In_Bytes + Version_Field_Size_In_Bytes + Integer.SIZE / 8
-        + Version_Field_Size_In_Bytes + (blobRecordVersion == Blob_Version_V2 ? (Short.SIZE / 8) : 0) + Long.SIZE / 8
+        + Version_Field_Size_In_Bytes + (blobRecordVersion >= Blob_Version_V2 ? (Short.SIZE / 8) : 0) + Long.SIZE / 8
         + storeKey.sizeInBytes()];
 
     ByteBuffer bufWrap = ByteBuffer.wrap(bytes);
@@ -288,7 +288,7 @@ class HardDeleteRecoveryMetadata {
     bufWrap.putShort(userMetadataVersion);
     bufWrap.putInt(userMetadataSize);
     bufWrap.putShort(blobRecordVersion);
-    if (blobRecordVersion == Blob_Version_V2) {
+    if (blobRecordVersion >= Blob_Version_V2) {
       bufWrap.putShort((short) blobType.ordinal());
     }
     bufWrap.putLong(blobStreamSize);
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
@@ -55,6 +55,7 @@ public class MessageFormatRecord {
   public static final short UserMetadata_Version_V1 = 1;
   public static final short Blob_Version_V1 = 1;
   public static final short Blob_Version_V2 = 2;
+  public static final short Blob_Version_V3 = 3;
   public static final short Metadata_Content_Version_V2 = 2;
   public static final short Metadata_Content_Version_V3 = 3;
   public static final int Message_Header_Invalid_Relative_Offset = -1;
@@ -232,20 +233,17 @@ public class MessageFormatRecord {
         return new DeserializedBlob(Blob_Version_V1, Blob_Format_V1.deserializeBlobRecord(crcStream));
       case Blob_Version_V2:
         return new DeserializedBlob(Blob_Version_V2, Blob_Format_V2.deserializeBlobRecord(crcStream));
+      case Blob_Version_V3:
+        return new DeserializedBlob(Blob_Version_V3, Blob_Format_V3.deserializeBlobRecord(crcStream));
       default:
         throw new MessageFormatException("data version not supported", MessageFormatErrorCodes.Unknown_Format_Version);
     }
   }
 
   static boolean isValidBlobRecordVersion(short blobRecordVersion) {
-    switch (blobRecordVersion) {
-      case Blob_Version_V1:
-        return true;
-      case Blob_Version_V2:
-        return true;
-      default:
-        return false;
-    }
+    return blobRecordVersion == Blob_Version_V1 ||
+        blobRecordVersion == Blob_Version_V2 ||
+        blobRecordVersion == Blob_Version_V3;
   }
 
   /**
@@ -1753,6 +1751,78 @@ public class MessageFormatRecord {
             MessageFormatErrorCodes.Data_Corrupt);
       }
       return new BlobData(blobContentType, dataSize, byteBuf);
+    }
+  }
+
+  /**
+   *  - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+   * |         |           |              |            |            |            |
+   * | version | blobType  | isCompressed |    size    |  content   |     Crc    |
+   * |(2 bytes)| (2 bytes) | (1 byte)     |  (8 bytes) |  (n bytes) |  (8 bytes) |
+   * |         |           |              |            |            |            |
+   *  - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+   *  version    - The version of the blob record
+   *
+   *  blobType   - The type of the blob
+   *
+   *  isCompressed - Whether the content is compressed.
+   *
+   *  size       - The size of the blob content
+   *
+   *  content    - The actual content that represents the blob
+   *
+   *  crc        - The crc of the blob record
+   *
+   */
+  public static class Blob_Format_V3 {
+    public static final int Blob_Size_Field_In_Bytes = 8;
+    public static final int Blob_Type_Field_In_Bytes = 2;
+    public static final int Is_Compressed_In_Bytes = 1;
+
+    private static Logger logger = LoggerFactory.getLogger(Blob_Format_V3.class);
+
+    public static long getBlobRecordSize(long blobSize) {
+      return Version_Field_Size_In_Bytes + Blob_Type_Field_In_Bytes + Blob_Size_Field_In_Bytes + blobSize + Crc_Size
+          + Is_Compressed_In_Bytes;
+    }
+
+    public static void serializePartialBlobRecord(ByteBuffer outputBuffer, long blobContentSize, BlobType blobType,
+        boolean isCompressed) {
+      outputBuffer.putShort(Blob_Version_V3);
+      outputBuffer.putShort((short) blobType.ordinal());
+      outputBuffer.put(isCompressed ? (byte) 1 : 0);
+      outputBuffer.putLong(blobContentSize);
+    }
+
+    public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IOException, MessageFormatException {
+      DataInputStream dataStream = new DataInputStream(crcStream);
+      // Get the BlobType.
+      short blobTypeOrdinal = dataStream.readShort();
+      if (blobTypeOrdinal > BlobType.values().length) {
+        logger.error("corrupt data while parsing blob content BlobContentType {}", blobTypeOrdinal);
+        throw new MessageFormatException("corrupt data while parsing blob content",
+            MessageFormatErrorCodes.Data_Corrupt);
+      }
+      BlobType blobContentType = BlobType.values()[blobTypeOrdinal];
+
+      // Get whether the blob is compressed.
+      boolean isCompressed = dataStream.readByte() == 1;
+
+      // Get the blob binary.
+      long dataSize = dataStream.readLong();
+      if (dataSize > Integer.MAX_VALUE) {
+        throw new IOException("We only support data of max size == MAX_INT. Error while reading blob from store");
+      }
+      ByteBuf byteBuf = Utils.readNettyByteBufFromCrcInputStream(crcStream, (int) dataSize);
+      long crc = crcStream.getValue();
+      long streamCrc = dataStream.readLong();
+      if (crc != streamCrc) {
+        logger.error("corrupt data while parsing blob content expectedcrc {} actualcrc {}", crc, streamCrc);
+        throw new MessageFormatException("corrupt data while parsing blob content",
+            MessageFormatErrorCodes.Data_Corrupt);
+      }
+
+      return new BlobData(blobContentType, dataSize, byteBuf, isCompressed);
     }
   }
 
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
@@ -1732,7 +1732,7 @@ public class MessageFormatRecord {
     public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IOException, MessageFormatException {
       DataInputStream dataStream = new DataInputStream(crcStream);
       short blobTypeOrdinal = dataStream.readShort();
-      if (blobTypeOrdinal > BlobType.values().length) {
+      if (blobTypeOrdinal >= BlobType.values().length) {
         logger.error("corrupt data while parsing blob content BlobContentType {}", blobTypeOrdinal);
         throw new MessageFormatException("corrupt data while parsing blob content",
             MessageFormatErrorCodes.Data_Corrupt);
@@ -1798,7 +1798,7 @@ public class MessageFormatRecord {
       DataInputStream dataStream = new DataInputStream(crcStream);
       // Get the BlobType.
       short blobTypeOrdinal = dataStream.readShort();
-      if (blobTypeOrdinal > BlobType.values().length) {
+      if (blobTypeOrdinal >= BlobType.values().length) {
         logger.error("corrupt data while parsing blob content BlobContentType {}", blobTypeOrdinal);
         throw new MessageFormatException("corrupt data while parsing blob content",
             MessageFormatErrorCodes.Data_Corrupt);
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
@@ -1779,7 +1779,7 @@ public class MessageFormatRecord {
     public static final int Blob_Type_Field_In_Bytes = 2;
     public static final int Is_Compressed_In_Bytes = 1;
 
-    private static Logger logger = LoggerFactory.getLogger(Blob_Format_V3.class);
+    private static final Logger logger = LoggerFactory.getLogger(Blob_Format_V3.class);
 
     public static long getBlobRecordSize(long blobSize) {
       return Version_Field_Size_In_Bytes + Blob_Type_Field_In_Bytes + Blob_Size_Field_In_Bytes + blobSize + Crc_Size
--- a/ambry-messageformat/src/test/java/com/github/ambry/messageformat/BlobDataTest.java
+++ b/ambry-messageformat/src/test/java/com/github/ambry/messageformat/BlobDataTest.java
@@ -1,0 +1,35 @@
+/**
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.messageformat;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class BlobDataTest {
+
+  @Test
+  public void testBlobDataConstructor() {
+    ByteBuf buf = Unpooled.wrappedBuffer(new byte[] { 1, 2, 3, 4});
+    BlobData data = new BlobData(BlobType.MetadataBlob, 100, buf);
+    Assert.assertEquals(BlobType.MetadataBlob, data.getBlobType());
+    Assert.assertEquals(100, data.getSize());
+    Assert.assertFalse(data.isCompressed());
+
+    // Test the other constructor.
+    data = new BlobData(BlobType.MetadataBlob, 100, buf, true);
+    Assert.assertTrue(data.isCompressed());
+  }
+}
--- a/ambry-messageformat/src/test/java/com/github/ambry/messageformat/BlobStoreHardDeleteTest.java
+++ b/ambry-messageformat/src/test/java/com/github/ambry/messageformat/BlobStoreHardDeleteTest.java
@@ -211,7 +211,7 @@ public class BlobStoreHardDeleteTest {
     private MessageFormatInputStream getPutMessage(StoreKey key, ByteBuffer encryptionKey,
         BlobProperties blobProperties, byte[] usermetadata, byte[] blob, int blobSize, short blobVersion,
         BlobType blobType) throws MessageFormatException {
-      if (blobVersion == MessageFormatRecord.Blob_Version_V2) {
+      if (blobVersion >= MessageFormatRecord.Blob_Version_V2) {
         return new PutMessageFormatInputStream(key, encryptionKey, blobProperties, ByteBuffer.wrap(usermetadata),
             new ByteBufferInputStream(ByteBuffer.wrap(blob)), blobSize, blobType);
       } else {
@@ -338,18 +338,18 @@ public class BlobStoreHardDeleteTest {
     short[] blobVersions = new short[5];
     BlobType[] blobTypes = new BlobType[5];
     for (int i = 0; i < 5; i++) {
-      blobVersions[i] = MessageFormatRecord.Blob_Version_V2;
+      blobVersions[i] = MessageFormatRecord.Blob_Version_V3;
       blobTypes[i] = BlobType.DataBlob;
     }
-    // all blobs V2 with Data blob
+    // all blobs V3 with Data blob
     blobStoreHardDeleteTestUtil(blobVersions, blobTypes);
     blobVersions = new short[5];
     blobTypes = new BlobType[5];
     for (int i = 0; i < 5; i++) {
-      blobVersions[i] = MessageFormatRecord.Blob_Version_V2;
+      blobVersions[i] = MessageFormatRecord.Blob_Version_V3;
       blobTypes[i] = BlobType.MetadataBlob;
     }
-    // all blobs V2 with Metadata blob
+    // all blobs V3 with Metadata blob
     blobStoreHardDeleteTestUtil(blobVersions, blobTypes);
   }
 
@@ -362,16 +362,16 @@ public class BlobStoreHardDeleteTest {
     blobVersions[0] = MessageFormatRecord.Blob_Version_V1;
     blobTypes[0] = BlobType.DataBlob;
 
-    blobVersions[1] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[1] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[1] = BlobType.MetadataBlob;
 
-    blobVersions[2] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[2] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[2] = BlobType.DataBlob;
 
-    blobVersions[3] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[3] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[3] = BlobType.MetadataBlob;
 
-    blobVersions[4] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[4] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[4] = BlobType.DataBlob;
 
     blobStoreHardDeleteTestUtil(blobVersions, blobTypes);
@@ -381,16 +381,16 @@ public class BlobStoreHardDeleteTest {
     blobVersions[0] = MessageFormatRecord.Blob_Version_V1;
     blobTypes[0] = BlobType.DataBlob;
 
-    blobVersions[1] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[1] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[1] = BlobType.DataBlob;
 
-    blobVersions[2] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[2] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[2] = BlobType.MetadataBlob;
 
-    blobVersions[3] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[3] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[3] = BlobType.DataBlob;
 
-    blobVersions[4] = MessageFormatRecord.Blob_Version_V2;
+    blobVersions[4] = MessageFormatRecord.Blob_Version_V3;
     blobTypes[4] = BlobType.MetadataBlob;
 
     blobStoreHardDeleteTestUtil(blobVersions, blobTypes);
--- a/ambry-messageformat/src/test/java/com/github/ambry/messageformat/HardDeleteMessageFormatInputStreamTest.java
+++ b/ambry-messageformat/src/test/java/com/github/ambry/messageformat/HardDeleteMessageFormatInputStreamTest.java
@@ -1,0 +1,30 @@
+/**
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.messageformat;
+
+import java.io.IOException;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class HardDeleteMessageFormatInputStreamTest {
+
+  @Test
+  public void testConstructor() throws MessageFormatException, IOException {
+    HardDeleteMessageFormatInputStream inputStream = new HardDeleteMessageFormatInputStream(1000,
+        MessageFormatRecord.UserMetadata_Version_V1, 100, MessageFormatRecord.Blob_Version_V3,
+        BlobType.DataBlob, 2000);
+
+    Assert.assertTrue(inputStream.streamLength > 0);
+  }
+}
--- a/ambry-messageformat/src/test/java/com/github/ambry/messageformat/HardDeleteMessageFormatInputStreamTest.java
+++ b/ambry-messageformat/src/test/java/com/github/ambry/messageformat/HardDeleteMessageFormatInputStreamTest.java
@@ -17,8 +17,15 @@ import java.io.IOException;
 import org.junit.Assert;
 import org.junit.Test;
 
+
+/**
+ * Test hard-delete input stream message format.
+ */
 public class HardDeleteMessageFormatInputStreamTest {
 
+  /**
+   * Majority of the logic resides in the constructor. Test constructor to make sure it does not throw.
+   */
   @Test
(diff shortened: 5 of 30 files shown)


ambry-messageformat/src/main/java/com/github/ambry/messageformat/BlobStoreHardDelete.java
```diff
@@ -233,7 +233,7 @@ class HardDeleteRecoveryMetadata {
     userMetadataVersion = stream.readShort();
     userMetadataSize = stream.readInt();
     blobRecordVersion = stream.readShort();
-    if (blobRecordVersion == Blob_Version_V2) {
+    if (blobRecordVersion >= Blob_Version_V2) {
       blobType = BlobType.values()[stream.readShort()];
```
mikeyang923, line 237: So this change imply that all future higher versions have to put blobType in the BlobType.values(), while currently we only do that for Blob_Version_V2. Is that a contract we want to obey?

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
```diff
@@ -1756,6 +1754,78 @@ public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IO
...
+   *  blobType   - The type of the blob
+   *
+   *  isCompressed - Whether the content is compressed.
+   *
+   *  size       - The size of the blob content
+   *
+   *  content    - The actual content that represents the blob
+   *
+   *  crc        - The crc of the blob record
+   *
```
mikeyang923, line 1775: For better readability, can you specify what CRC stands for here?

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
```diff
@@ -1756,6 +1754,78 @@ public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IO
...
+      outputBuffer.put(isCompressed ? (byte) 1 : 0);
+      outputBuffer.putLong(blobContentSize);
+    }
+
+    public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IOException, MessageFormatException {
+      DataInputStream dataStream = new DataInputStream(crcStream);
+      // Get the BlobType.
+      short blobTypeOrdinal = dataStream.readShort();
+      if (blobTypeOrdinal > BlobType.values().length) {
+        logger.error("corrupt data while parsing blob content BlobContentType {}", blobTypeOrdinal);
```
mikeyang923, line 1802: Should we have ">=" instead of ">" here?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -873,6 +882,35 @@ protected void maybeProcessCallbacks() {
...
+        return sourceBuffer;
+      }
+
+      // Note: decompress() already emit metrics to for decompress() failure.
+      try {
+        ByteBuf result = decompressionService.decompress(sourceBuffer, routerConfig.routerMaxPutChunkSizeBytes);
+        sourceBuffer.release();
+        return result;
+      }
+      catch (Exception ex) {
```
mikeyang923, line 905: nit: better to use below format to be consistent:
`} catch (...) {`

ambry-messageformat/src/main/java/com/github/ambry/messageformat/BlobStoreHardDelete.java
panchow: That is correct, BlobType is introduced in V2 and stays in V3.  That's why it is V2+.  In the future, if it gets removed in V4, then we'll need to modify the logic.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
panchow: CRC is Cyclic Redundancy Code.  It is better known by CRC than by the name, just like JPEG.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
panchow: You are right.  It should be >=, let me correct that.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: I just searched and fixed all them all in the project.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
```diff
@@ -1756,6 +1754,78 @@ public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IO
...
   }
 
+  /**
+   *  - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+   * |         |           |              |            |            |            |
+   * | version | blobType  | isCompressed |    size    |  content   |     Crc    |
+   * |(2 bytes)| (2 bytes) | (1 byte)     |  (8 bytes) |  (n bytes) |  (8 bytes) |
+   * |         |           |              |            |            |            |
+   *  - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+   *  version    - The version of the blob record
```
snalli, line 1764: why 1 byte for a boolean ?

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
```diff
@@ -1756,6 +1754,78 @@ public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IO
...
+   *
+   *  blobType   - The type of the blob
+   *
```
snalli, line 1767: just for my knowledge, this blob record is persisted along with the blob ?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -873,6 +882,34 @@ protected void maybeProcessCallbacks() {
...
+     * Decompress the compressed buffer specified.
+     * If the buffer is not compressed, based on blobCompressed property, then return the input buffer.
+     *
+     * @param sourceBuffer The source buffer to decompress.
+     * @return New buffer if decompressed successfully and the source buffer is released.
+     *         Original buffer is blob not compressed.  NULL if failed and source buffer is released.
+     */
+    protected ByteBuf decompressContent(ByteBuf sourceBuffer) {
+      if (!isChunkCompressed) {
+        logger.info("Blob is not compressed.  Return source buffer.");
```
snalli, line 895: This will overwhelm the log, plus its not very informative. It doesn't print out blob id

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -164,6 +166,8 @@ class GetBlobOperation extends GetOperation {
         shouldLookupMetadataCache() ? (BlobMetadata) blobMetadataCache.getObject(blobId.toString()) : null;
     firstChunk = (blobMetadata == null) ? new FirstGetChunk() : new CachedFirstChunk(blobMetadata);
     this.nonBlockingRouter = nonBlockingRouter;
+    decompressionService = new CompressionService(routerConfig.getCompressionConfig(),
```
snalli, line 169: So for each GET, we create new instance of compression service ?
Can we do better - like having single shared instance of this service and enqueueing requests to it like how encryption/decryption work ?

snalli: actually, nevermind. This is ok.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -120,6 +120,14 @@ class PutOperation {
   private boolean isEncryptionEnabled;
   private final QuotaChargeCallback quotaChargeCallback;
 
+  // Whether the entire blob (all chunks) are compressible.  Default is null for undecided.
+  // The value is set based on blobProperties and compression service configuration.
+  // For example, if BlobProperties.ContentEncoding specified, then the entire blob will
+  // not be compressed by setting this field to FALSE.
+  // This field is set once when inserting the first chunk based on blob properties.
+  // All chunk compression will check this field and skip compression if this value FALSE.
+  private Boolean isBlobCompressible;
```
snalli, line 129: Why can't the default be false ?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -852,8 +857,12 @@ protected void maybeProcessCallbacks() {
               routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
             }
             routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-            chunkIndexToBuf.put(chunkIndex, filterChunkToRange(result.getDecryptedBlobContent()));
-            numChunksRetrieved.incrementAndGet();
+            ByteBuf decryptedContent = result.getDecryptedBlobContent();
+            ByteBuf decompressedContent = decompressContent(decryptedContent);
```
snalli, line 861: so decompression is only for encrypted blobs ?

snalli: nevermind, i see it down below for un-encrypted blobs

ambry-protocol/src/test/java/com/github/ambry/protocol/PutRequestTest.java
```diff
@@ -0,0 +1,62 @@
...
+import com.github.ambry.messageformat.BlobType;
+import com.github.ambry.utils.NettyByteBufDataInputStream;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class PutRequestTest {
```
ankagrawal, line 30: nit: javadocs?
Also this is not PutRequestTest. This PutRequest v5 deserialization test.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1014,17 +1051,20 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
         BlobData blobData = MessageFormatRecord.deserializeBlob(payload);
         ByteBuffer encryptionKey = messageMetadata == null ? null : messageMetadata.getEncryptionKey();
         ByteBuf chunkBuf = blobData.content();
+        this.isChunkCompressed = blobData.isCompressed();
```
snalli, line 1054: this is not used in this method

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
ankagrawal, line 169: Do we create a new `CompressionService` per `GetOperation`? Can this be created at `OperationController` and be shared by all `GetOperation`s?

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
panchow: that's the pattern that is common used Ambry.  Use 1 byte for a boolean.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
panchow: Yes, these information is always persisted with the blob.  It was there since V1 and V2.  I created V3 and it is backward compatible.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: I'll remove this log in next revision.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Eventually, when we adopt dependency injection.  It should be a singleton.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: I either use 2 boolean variables or use this as a tri-state variable.  Null means undetermined.  false means determined and do not compress.  true means determined and compress.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Correct.  This is not used in this method.  It is used when generating the PutRequest.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1014,17 +1051,20 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
...
-            chunkBuf = filterChunkToRange(chunkBuf);
             if (chunkIndex == 0) {
               routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
             }
             routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-            chunkIndexToBuf.put(chunkIndex, chunkBuf.retainedDuplicate());
-            numChunksRetrieved.incrementAndGet();
+            ByteBuf decompressedContent = decompressContent(chunkBuf.retainedDuplicate());
+            if (decompressedContent != null) {
+              chunkIndexToBuf.put(chunkIndex, filterChunkToRange(decompressedContent));
```
snalli, line 1065: These two lines are common. Can they be pushed down to decompressContent ?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1291,6 +1305,35 @@ private void encryptionCallback(EncryptJob.EncryptJobResult result, Exception ex
...
+    /**
+     * Compress the buffer in "buf" field for data chunks.  It does not compress metadata chunks.
+     * It also applies rules specified in config.
+     * After compression, it replaces "buf" field and update isCompressed field in blob properties.
+     */
+    private void compressChunk() {
+      // Do not compress metadata chunk or already compressed chunk.
+      if (isMetadataChunk() || isChunkCompressed) {
+        return;
+      }
```
snalli, line 1317: will this be called more than once ?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -852,8 +857,12 @@ protected void maybeProcessCallbacks() {
               routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
             }
             routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-            chunkIndexToBuf.put(chunkIndex, filterChunkToRange(result.getDecryptedBlobContent()));
-            numChunksRetrieved.incrementAndGet();
+            ByteBuf decryptedContent = result.getDecryptedBlobContent();
```
ankagrawal, line 860: The decryption runs in a separate thread pool asynchronously, so that progress with other operations of the single-threaded `OperationController` is not delayed. Should we create a similar scheduler for decryption as well?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1482,8 +1522,11 @@ protected void maybeProcessCallbacks() {
               if (!resolveRange(totalSize)) {
                 routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
                 routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-                chunkIndexToBuf.put(0, filterChunkToRange(decryptedBlobContent));
-                numChunksRetrieved.set(1);
+                ByteBuf decompressedContent = decompressContent(decryptedBlobContent);
+                if (decompressedContent != null) {
+                  chunkIndexToBuf.put(0, filterChunkToRange(decompressedContent));
+                  numChunksRetrieved.set(1);
+                }
```
ankagrawal, line 1529: If decompression fails and `decompressedContent != null` is false, then what should we do?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1568,8 +1613,11 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
             dataChunks = null;
             chunkIndex = 0;
             numChunksTotal = 1;
-            chunkIndexToBuf.put(0, Unpooled.wrappedBuffer(rawPayloadBuffer));
-            numChunksRetrieved.set(1);
+            ByteBuf decompressedBuffer = decompressContent(Unpooled.wrappedBuffer(rawPayloadBuffer));
+            if (decompressedBuffer != null) {
```
ankagrawal, line 1617: If decompression fails and `decompressedContent != null` is false, then what should we do?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1751,11 +1799,14 @@ private void handleSimpleBlob(BlobData blobData, byte[] userMetadata, ByteBuffer
...
             boolean launchedJob = maybeLaunchCryptoJob(chunkBuf, userMetadata, encryptionKey, blobId);
             if (!launchedJob) {
-              chunkBuf = filterChunkToRange(chunkBuf);
               routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
               routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-              chunkIndexToBuf.put(0, chunkBuf.retainedDuplicate());
-              numChunksRetrieved.set(1);
+
+              ByteBuf decompressedContent = decompressContent(chunkBuf.retainedDuplicate());
+              if (decompressedContent != null) {
```
ankagrawal, line 1806: Same as above.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -305,6 +314,7 @@ private PutOperation(RouterConfig routerConfig, NonBlockingRouterMetrics routerM
     isEncryptionEnabled = passedInBlobProperties.isEncrypted();
     restRequest = options.getRestRequest();
     loggingContext = makeLoggingContext();
+    compressionService = new CompressionService(routerConfig.getCompressionConfig(), routerMetrics.compressionMetrics);
```
ankagrawal, line 317: One for each operation?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1482,8 +1522,11 @@ protected void maybeProcessCallbacks() {
               if (!resolveRange(totalSize)) {
                 routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
                 routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-                chunkIndexToBuf.put(0, filterChunkToRange(decryptedBlobContent));
-                numChunksRetrieved.set(1);
+                ByteBuf decompressedContent = decompressContent(decryptedBlobContent);
```
Arun-LinkedIn, line 1525: Do we need to decompress content before calling `resolveRange(totalSize)` above? Else, it seems like we might try to resolve the byte range against compressed content size which might be incorrect. For example, start offset of range request could be greater than compressed content size.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1751,11 +1799,14 @@ private void handleSimpleBlob(BlobData blobData, byte[] userMetadata, ByteBuffer
             numChunksTotal = 1;
             boolean launchedJob = maybeLaunchCryptoJob(chunkBuf, userMetadata, encryptionKey, blobId);
             if (!launchedJob) {
-              chunkBuf = filterChunkToRange(chunkBuf);
               routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
               routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-              chunkIndexToBuf.put(0, chunkBuf.retainedDuplicate());
-              numChunksRetrieved.set(1);
+
+              ByteBuf decompressedContent = decompressContent(chunkBuf.retainedDuplicate());
```
Arun-LinkedIn, line 1805: Same as above. Do we probably need to decompress content before trying to resolve byte range in line 1789?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1326,6 +1369,9 @@ void onFillComplete(boolean updateMetric) {
         routerMetrics.chunkFillTimeMs.update(time.milliseconds() - chunkFreeAtMs);
       }
 
+      // Attempt to apply compression on data chunk.
+      compressChunk();
```
Arun-LinkedIn, line 1373: Would we eventually make compression and decompression asynchronous as well?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
SophieGuo410, line 129: Can we also check the "x-ambry-content-encoding". Customer will set it to gzip if the blob has been pre-compressed. So we don't need to compress twice. But during decompress stage, we should not decompress it. Customer will handle the decompression.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
```diff
@@ -1756,6 +1754,78 @@ public static BlobData deserializeBlobRecord(CrcInputStream crcStream) throws IO
...
+   *  crc        - The crc of the blob record
+   *
+   */
+  public static class Blob_Format_V3 {
+    public static final int Blob_Size_Field_In_Bytes = 8;
+    public static final int Blob_Type_Field_In_Bytes = 2;
+    public static final int Is_Compressed_In_Bytes = 1;
+
+    private static Logger logger = LoggerFactory.getLogger(Blob_Format_V3.class);
```
SophieGuo410, line 1782: Could be final.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1555,6 +1598,8 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
           }
         }
         blobType = blobData.getBlobType();
+        isChunkCompressed = blobData.isCompressed();
+        logger.trace("In FirstGetChunk.handleBody(), is Chunk compressed = " + isChunkCompressed);
```
SophieGuo410, line 1602: nit: probably it can include some blobId info, otherwise it's hard to trace the log.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: No... decompressContent should just decompress content.  Applying range to a chunk is another operation.  If we apply filtering inside decompressContent(), it's considered hidden side-effects which is not desirable.  Also moving that line to the decompressContent() has no performance gain or readability improvement, so I suggest we keep it as is.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: This is called once per chunk.

ambry-messageformat/src/test/java/com/github/ambry/messageformat/BlobDataTest.java
```diff
@@ -0,0 +1,35 @@
...
+    BlobData data = new BlobData(BlobType.MetadataBlob, 100, buf);
+    Assert.assertEquals(BlobType.MetadataBlob, data.getBlobType());
+    Assert.assertEquals(100, data.getSize());
+    Assert.assertFalse(data.isCompressed());
+
+    // Test the other constructor.
+    data = new BlobData(BlobType.MetadataBlob, 100, buf, true);
+    Assert.assertTrue(data.isCompressed());
+  }
+}
```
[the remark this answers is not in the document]
panchow: Ok

ambry-messageformat/src/test/java/com/github/ambry/messageformat/BlobStoreHardDeleteTest.java
```diff
@@ -338,18 +338,18 @@ public void blobStoreHardDeleteTestBlobV2Simple() throws MessageFormatException,
     short[] blobVersions = new short[5];
     BlobType[] blobTypes = new BlobType[5];
     for (int i = 0; i < 5; i++) {
-      blobVersions[i] = MessageFormatRecord.Blob_Version_V2;
+      blobVersions[i] = MessageFormatRecord.Blob_Version_V3;
```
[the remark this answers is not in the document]
panchow: This test continues to test V1 and V2 de-serialization from stream, but this test does not test V2 serialization to stream.  There are other unit tests that test V2 serialization and de-serialization.

Those tests need to serialize a blob to stream, as part of the test flow.  The serialization should use the latest version only V3, just like all other record formats.  We should not serialize new put operations using an older blob versions.  If we support serializing put operations using old format version, which could lead to data lost, it will make unit testing very complex because every serialize call need to pass in a version.  Imagine the PutRequest needs to pass in header version, blob property version, user metadata version, and so on.  That would be messy and unnecessary.  So let's just use the latest version in the put operation, and support all previous versions in the get operation.

ambry-messageformat/src/test/java/com/github/ambry/messageformat/HardDeleteMessageFormatInputStreamTest.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.messageformat;
+
+import java.io.IOException;
+import org.junit.Assert;
+import org.junit.Test;
+
+public class HardDeleteMessageFormatInputStreamTest {
```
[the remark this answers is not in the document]
panchow: Added.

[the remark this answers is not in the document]
panchow: added

ambry-messageformat/src/test/java/com/github/ambry/messageformat/MessageFormatInputStreamTest.java
```diff
@@ -70,22 +70,22 @@ public void after() {
...
-        MessageFormatRecord.Message_Header_Version_V1);
-    messageFormatPutRecordsTest(MessageFormatRecord.Blob_Version_V2, BlobType.DataBlob,
-        MessageFormatRecord.Message_Header_Version_V2);
-    messageFormatPutRecordsTest(MessageFormatRecord.Blob_Version_V2, BlobType.MetadataBlob,
-        MessageFormatRecord.Message_Header_Version_V2);
-    messageFormatPutRecordsTest(MessageFormatRecord.Blob_Version_V2, BlobType.DataBlob,
-        MessageFormatRecord.Message_Header_Version_V3);
-    messageFormatPutRecordsTest(MessageFormatRecord.Blob_Version_V2, BlobType.MetadataBlob,
-        MessageFormatRecord.Message_Header_Version_V3);
+        MessageFormatRecord.Message_Header_Version_V1, false);
```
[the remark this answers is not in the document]
panchow: If we support invoking serialization using old versions, it will make the serialization unnecessarily complex.  First, we should never serialize using old versions because it could lead to data lost.  For example, the chunk is compressed and compression flag is introduced in V3.  If we serialize using V2, the compression flag will be lost and will result in data lost.  So we should always serialize using the latest version.

Coming back to the need to test serializing using older version.  If we add a parameter deep in the serialization, we will need to pass this serialization version through many layers of code.  Again, it's dangerous to pass old version so if someone mistakenly pass in an older version, it can get deployed without failing any tests.

Instead, I suggest we keep these tests serializing using the latest version V3, but have different unit test to make sure V2 specifically to make sure their serializer and de-serializer are compatible.

ambry-messageformat/src/test/java/com/github/ambry/messageformat/MessageFormatRecordTest.java
```diff
@@ -748,9 +748,31 @@ private BlobData getBlobRecordV2(int blobSize, BlobType blobType, ByteBuffer blo
...
+    crc.update(outputBuffer.array(), 0, outputBuffer.position());
+    outputBuffer.putLong(crc.getValue());
+    outputBuffer.flip();
+    return MessageFormatRecord.deserializeBlob(new ByteBufferInputStream(outputBuffer));
+  }
+
   @Test
   public void testBlobRecordWithMetadataContentV2() throws IOException, MessageFormatException {
-    // Test Blob V2 with actual metadata blob V2
+    // Test Blob V3 with actual metadata blob V2
```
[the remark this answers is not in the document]
panchow: In my opinion, we should remove testing V1 and V2.  Just stay with V3.  The additional logic written to serialize using V1 will never happen.  We will never go back to start serializing V1 because it will be lost of data.  So let's not keep code that don't make sense to keep.  Instead, we should clean it up to remove dangerous code like serializing in older versions.

ambry-messageformat/src/test/java/com/github/ambry/messageformat/MessageSievingInputStreamTest.java
```diff
@@ -125,10 +125,10 @@ public void testEmptyMsgInfoList() throws Exception {
   public void testValidBlobsAgainstCorruption() throws Exception {
     testValidBlobs(Blob_Version_V1, BlobType.DataBlob, Message_Header_Version_V1);
     testValidBlobs(Blob_Version_V1, BlobType.DataBlob, Message_Header_Version_V2);
-    testValidBlobs(Blob_Version_V2, BlobType.DataBlob, Message_Header_Version_V2);
-    testValidBlobs(Blob_Version_V2, BlobType.MetadataBlob, Message_Header_Version_V2);
-    testValidBlobs(Blob_Version_V2, BlobType.DataBlob, Message_Header_Version_V3);
-    testValidBlobs(Blob_Version_V2, BlobType.MetadataBlob, Message_Header_Version_V3);
+    testValidBlobs(Blob_Version_V3, BlobType.DataBlob, Message_Header_Version_V2);
```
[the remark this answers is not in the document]
panchow: same reason. Let's sync in person.

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
```diff
@@ -388,6 +440,14 @@ public Long getCrc() {
     return crcValue;
   }
 
+  /**
+   * Get whether the data blob is compressed.
+   * @return TRUE if compressed; FALSE if not.
```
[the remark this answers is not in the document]
panchow: done

ambry-protocol/src/test/java/com/github/ambry/protocol/PutRequestTest.java
panchow: I'll add the Java doc.  This class is for PutRequest testing.  It starts with only 1 unit test instead of testing the entire class.  In the future, anyone modifying the PutRequest class should add their unit test here.  The unit test inside this class make sure it can serialize a put request and de-serialize a put request.  Serialization always use the latest format version, so the de-serialized output should match the latest version.  Again, this unit test is not intended to test all versions.  That's the job of other unit tests that test each version.  This unit test should focus on testing PutRequest only.  BTW, we could add more unit tests as needed.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Eventually we want the compression service to be a singleton using Dependency Injection.  For this case, we should not create an instance in controller and pass in here because it breaks the encapsulation principal that the class itself should be self-contain and hold on to objects it needs to complete its operation.  Clients (operation class in this case) should not interfere or decide how the dependent class should behave.  This is also coding style thing.  That means class A should use class B by its APIs, not class A controls how class B implements its methods.  In this case, if Operation class pass in CompressionService to PutOperation, then Operation class forces compression service must be singleton, and put operation must have compression as its process.  In this pattern, classes know too much about other classes and interfere their implementation.  Instead, classes should be clean, self-contained, and abstracted so that they communicate only by API without knowing how the classes are implemented.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Encryption uses another thread.  It does not make the operation faster faster or higher throughput but the code is much more complicated.  Let me explain.

It's important to have 2 layers of threads.  The IO threads and the business logic threads.  The IO threads is managed by Netty.  So when Netty invokes a callback, it's calling using IO thread.  Ambry handler needs to respond to Netty callback ASAP so there is no IO blocking.  That means the callback handler (running in Netty IO thread) should pickup the object and pass it to Ambry business logic threads, and return right way to stop taking more IO thread time.  The business logic threads should do what the processing.  If business logic needs to do disk IO, it should be asynchronous.  If it needs to loop through the content, it should be synchronous.  Yes, synchronous!  It doesn't help if Thread A calls Thread B and wait for Thread B to complete.  Encryption and compression are effectively looping through the content.  It does not help for one business logic thread to pass that work to another business logic thread (in the same thread pool).

Our operation manager already created the business logic thread pool to handle Netty requests and responses, so when put operation executes, they are already executing it the business logic thread.  There is no need to create more threads.  Our router code can be much simpler.  Let's work on cleaning it up.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: If decompression failed, we set the operation status to failed with the exception.  Setting the operation status and exception is done inside the decompressContent().

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Same as above.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: Not ideal.  We should use Dependency injection so we create only one.  Let me create a ticket to come back to refactor this to use dependency injection.https://jira01.corp.linkedin.com:8443/browse/AMBRY-9407

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: That is correct, content must be decompressed first before applying the range.  ResolveRange() is calculated based on the chunk sizes, not compressed content size, so there should have no impact.  Compression is hidden to the entire put operation except serialize and de-serialize to blob content.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Yes, we must decompress first before applying range filtering.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: No, compression and decompression do not need to be asynchronous because we are already running in a worker thread, not Netty IO thread.  Make it async will increase complexity and slow down operation.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: Sure thing.  Let me make that change.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatRecord.java
panchow: Thank you.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
SophieGuo410: You can also have a follow up pr if you think this pr is huge.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Agreed.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: According to RestUtil.java, the BlobProperties.contentEncoding is obtained from the "x-ambry-content-encoding" header.  So I believe we are currently doing that already.

    public final static String AMBRY_CONTENT_ENCODING = "x-ambry-content-encoding";

public static BlobProperties buildBlobProperties(Map<String, Object> args) throws RestServiceException {
    Account account = getAccountFromArgs(args);
    Container container = getContainerFromArgs(args);
    String serviceId = getHeader(args, Headers.SERVICE_ID, true);
    String contentType = getHeader(args, Headers.AMBRY_CONTENT_TYPE, true);
    String ownerId = getHeader(args, Headers.OWNER_ID, false);
    String externalAssetTag = getHeader(args, Headers.EXTERNAL_ASSET_TAG, false);
    **String contentEncoding = getHeader(args, Headers.AMBRY_CONTENT_ENCODING, false);**
    String filename = getHeader(args, Headers.AMBRY_FILENAME, false);

    long ttl = getTtlFromRequestHeader(args);

    // This field should not matter on newly created blobs, because all privacy/cacheability decisions should be made
    // based on the container properties and ACLs. For now, BlobProperties still includes this field, though.
    boolean isPrivate = !container.isCacheable();
    return new BlobProperties(-1, serviceId, ownerId, contentType, isPrivate, ttl, account.getId(), container.getId(),
        container.isEncrypted(), externalAssetTag, contentEncoding, filename);
  }

SophieGuo410: yeah, you are right, I missed this part. I think it's good!
`if (isSkipWithContentEncoding && !Utils.isNullOrEmpty(contentEncoding))`

mikeyang923 (approved):

panchow:
I have converted this PR to DRAFT because of deployment concern.  I'm going to write a short doc to describe the deployment plan.

PR #2258, opened 2022-10-12
Change configuration to enable rebuild replication token

justinlin-linkedin:
Change the meaning of store.rebuild.token.based.on.compaction.history so that now if it's true, it will enable all the partitions in the host.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -253,8 +253,8 @@ public class BlobStore implements Store {
         index = new PersistentIndex(dataDir, storeId, taskScheduler, log, config, factory, recovery, hardDelete,
             diskIOScheduler, metrics, time, sessionId, storeDescriptor.getIncarnationId());
         compactor.initialize(index);
-        if (replicaId != null && (config.storeRebuildTokenBasedOnCompactionHistory
-            || config.storePartitionsToRebuildTokenBasedOnCompactionHistory.contains(
+        if (config.storeRebuildTokenBasedOnCompactionHistory || (replicaId != null
+            && config.storePartitionsToRebuildTokenBasedOnCompactionHistory.contains(
             replicaId.getPartitionId().getId()))) {
           compactor.enablePersistIndexSegmentOffsets();
           index.enableRebuildTokenBasedOnCompactionHistory();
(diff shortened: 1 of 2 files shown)


SophieGuo410 (approved):

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -253,9 +253,9 @@ public void start() throws StoreException {
         index = new PersistentIndex(dataDir, storeId, taskScheduler, log, config, factory, recovery, hardDelete,
             diskIOScheduler, metrics, time, sessionId, storeDescriptor.getIncarnationId());
         compactor.initialize(index);
-        if (config.storeRebuildTokenBasedOnCompactionHistory
-            && config.storePartitionsToRebuildTokenBasedOnCompactionHistory.contains(
-            replicaId.getPartitionId().getId())) {
```
panchow, line 258: should it be (config.storeRebuildTokenBasedOnCompactionHistory
|| (replicaId != null && config.storePartitionsToRebuildTokenBasedOnCompactionHistory.contains(replicaId.getPartitionId().getId()))

panchow (changes requested):

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
justinlin-linkedin: that's better

mikeyang923 (approved):

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -253,9 +253,9 @@ public void start() throws StoreException {
         index = new PersistentIndex(dataDir, storeId, taskScheduler, log, config, factory, recovery, hardDelete,
             diskIOScheduler, metrics, time, sessionId, storeDescriptor.getIncarnationId());
         compactor.initialize(index);
-        if (config.storeRebuildTokenBasedOnCompactionHistory
+        if (config.storeRebuildTokenBasedOnCompactionHistory || (replicaId != null
```
Arun-LinkedIn, line 256: QQ: This means we could enable this feature on an entire host or selected partitions?

Arun-LinkedIn (approved):

PR #2262, opened 2022-10-13
Change local server state only in response to primary helix callbacks

snalli:
Add debug logs to debug state changes

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
@@ -269,6 +269,8 @@ public class ReplicationManager extends ReplicationEngine {
 
     @Override
     public void onPartitionBecomeStandbyFromBootstrap(String partitionName) {
+      logger.info("Partition state change notification from Bootstrap to Standby received for partition {}",
+          partitionName);
       // if code arrives here, it means local replica has completed OFFLINE -> BOOTSTRAP transition. We don't have to
       // check if local replica exists or not.
       ReplicaId localReplica = storeManager.getReplica(partitionName);
@@ -278,15 +280,20 @@ public class ReplicationManager extends ReplicationEngine {
         throw new StateTransitionException(
             "Store " + partitionName + " is not started during Bootstrap-To-Standby transition", StoreNotStarted);
       }
+      logger.info("Partition = {}, replicaId = {}, storeId = {}", partitionName, localReplica, store);
       // 2. check if store is new added and needs to catch up with peer replicas.
+      ReplicaState currentState = store.getCurrentState();
       if (store.isBootstrapInProgress()) {
+        logger.info("Partition {} bootstrap in progress, current state = {}", partitionName, currentState);
         store.setCurrentState(ReplicaState.BOOTSTRAP);
         // store state will updated to STANDBY in ReplicaThread when bootstrap is complete
         replicaSyncUpManager.initiateBootstrap(localReplica);
       } else {
+        logger.info("Partition {} bootstrap complete, current state = {}", partitionName, currentState);
         // if this is existing replica, then directly set state to STANDBY
         store.setCurrentState(ReplicaState.STANDBY);
       }
+      logger.info("Partition {} has changed state from {} to {}", partitionName, currentState, store.getCurrentState());
     }
 
     @Override
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -562,7 +562,6 @@ public class StorageManager implements StoreManager {
               "Store " + partitionName + " didn't start correctly, replica should be set to ERROR state",
               StoreNotStarted);
         }
-
         // if store's used capacity is less than or equal to header size, we create a bootstrap_in_progress file and force
         // it to stay in BOOTSTRAP state when catching up with peers.
         long storeUsedCapacity = store.getSizeInBytes();
@@ -621,7 +620,6 @@ public class StorageManager implements StoreManager {
               "Store " + partitionName + " is already disabled due to I/O error or by " + "admin operation",
               ReplicaOperationFailure);
         }
-
         // 0. as long as local replica exists, we create a decommission file in its dir
         File decommissionFile = new File(replica.getReplicaPath(), BlobStore.DECOMMISSION_FILE_NAME);
         try {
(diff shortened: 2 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -269,24 +269,31 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
 
     @Override
     public void onPartitionBecomeStandbyFromBootstrap(String partitionName) {
+      logger.info("Partition state change notification from Bootstrap to Standby received for partition {}",
+          partitionName);
```
mikeyang923, line 273: Do you want to use lower level log (eg trace)? and run shell script to set the proper log level for the host?

mikeyang923 (approved):
lgtmbc

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
snalli: nope, this will not overwhelm the log and is critical to debug state transitions which happen only once during startup or rare thereafter

Arun-LinkedIn, line 273: Since we already print this in Base class, we may see duplicate prints of similar messages on bootstrap to standby transition.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -269,24 +269,31 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
...
       // if code arrives here, it means local replica has completed OFFLINE -> BOOTSTRAP transition. We don't have to
       // check if local replica exists or not.
       ReplicaId localReplica = storeManager.getReplica(partitionName);
       Store store = storeManager.getStore(localReplica.getPartitionId());
+      logger.info("Partition = {}, replicaId = {}, storeId = {}", partitionName, localReplica, store);
```
Arun-LinkedIn, line 278: Since`store` could be null here, may be log after the null check?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
snalli: its not printed in base class PartitionStateChangeListener

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -269,24 +269,31 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
...
+        logger.info("Partition {} bootstrap is in progress with current state as {}", partitionName, currentState);
         store.setCurrentState(ReplicaState.BOOTSTRAP);
         // store state will updated to STANDBY in ReplicaThread when bootstrap is complete
         replicaSyncUpManager.initiateBootstrap(localReplica);
       } else {
+        logger.info("Partition {} is not in bootstrap with current state as {}", partitionName, currentState);
         // if this is existing replica, then directly set state to STANDBY
         store.setCurrentState(ReplicaState.STANDBY);
       }
+      logger.info("Partition {} has changed state from {} to {}", partitionName, currentState, store.getCurrentState());
```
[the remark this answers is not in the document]
snalli: Printed in setCurrentState but there we don't have partitionName. Then while grepping logs we have to make a connection between paritionName and storeId.

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -478,17 +478,30 @@ private class PartitionStateChangeListenerImpl implements PartitionStateChangeLi
...
+     * Renamed from isPrimaryClusterParticipantListener to isPrimaryClusterManagerListener because
+     * the usage of "participant" is completely wrong in the code at this point. It is being used to refer to both
+     * ambry-server node and helix. We don't want to fix all usages retroactively at this point,
+     * but going forward please follow this convention below.
+     * <p>
+     * participant = ambry-server node that _participates_ in cluster management done by helix cluster-manager service
+     * participant = _NOT_ helix cluster-manager service
+     * <p>
+     * And PartitionStateChangeListener listens to state transition messages from helix cluster-manager service.
+     */
```
justinlin-linkedin, line 494: we don't need this explanation. I think this is confusing.

snalli: fixed

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -551,6 +564,11 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
               "Store " + partitionName + " didn't start correctly, replica should be set to ERROR state",
               StoreNotStarted);
         }
+
+        if (!isPrimaryHelixListener) {
```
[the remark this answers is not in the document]
justinlin-linkedin: Let me take another look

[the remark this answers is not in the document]
justinlin-linkedin: The purpose of listening on both helix servers is to make sure when one of the helix transitioned the state to ERROR, the second one would do the same. 

There are several cases in this PR, when we terminate early for the second helix, it will not ends up in ERROR state. We should let the code in the those callback executed for both helix.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -278,15 +280,20 @@ public void onPartitionBecomeStandbyFromBootstrap(String partitionName) {
         throw new StateTransitionException(
             "Store " + partitionName + " is not started during Bootstrap-To-Standby transition", StoreNotStarted);
       }
+      logger.info("Partition = {}, replicaId = {}, storeId = {}", partitionName, localReplica, store);
       // 2. check if store is new added and needs to catch up with peer replicas.
+      ReplicaState currentState = store.getCurrentState();
```
panchow, line 285: nit: This variable is used only once in either if() or else block.  There's no benefit for introducing a variable.

panchow (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
snalli: Its used in one more place L296

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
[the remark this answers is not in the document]
snalli: Where is this callback below the return statement ?
In both cases where it returns early, there is no callback below.
The code below the return just creates some files (or not) and sets/resets some state.

snalli:
@justinlin-linkedin can u please take another look and merge if ok ?

justinlin-linkedin (approved):

PR #2263, opened 2022-10-13
Make sure Store internal state is only updated by only primary helix participant listener

Arun-LinkedIn:
We could be occasionally participating in multiple helix/zk clusters during migration from one ZK cluster to another. Ideally, the actions in `StorageManager` like creation of Store, etc should be taken only for primary helix participant State changes (Ex: offline to bootstrap). But, in order to detect stores that are down and mark appropriate ERROR in helix state on both helix clusters, we let the listener logic in `StorageManager` (like creation of Store) to be executed for both participants (Reference [PR](https://github.com/linkedin/ambry/pull/1550)). However, we need to make sure the internal state of the `Store` is only updated by primary helix participant since it is also updated by `ReplicationManager` which only listens to helix primary participant.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -77,7 +77,7 @@ public class StorageManager implements StoreManager {
   private final MessageStoreRecovery recovery;
   private final MessageStoreHardDelete hardDelete;
   private final List<ClusterParticipant> clusterParticipants;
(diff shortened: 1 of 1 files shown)


mikeyang923 (approved):

justinlin-linkedin (approved):
LGTM

snalli:
please wait for me to go over and discuss this first before merging

snalli (changes requested):
please wait for me to go over your pr before merging

PR #2264, opened 2022-10-14
Implemented TtlUpdate retry with ReplicateBlob

JingQianCloud:
Implemented TtlUpdate retry with ReplicateBlob

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -124,6 +124,7 @@ public class RouterConfig {
       "router.update.op.metadata.reliance.timestamp.in.ms";
   public static final String ROUTER_UNAVAILABLE_DUE_TO_SUCCESS_COUNT_IS_NON_ZERO_FOR_DELETE =
       "router.unavailable.due.to.success.count.is.non.zero.for.delete";
+  public static final String ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED = "router.repair.with.replicate.blob.enabled";
 
   /**
    * Number of independent scaling units for the router.
@@ -638,6 +639,14 @@ public class RouterConfig {
   @Config(ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS)
   public final long routerUpdateOpMetadataRelianceTimestampInMs;
 
+  /**
+   * If this config is set to {@code true}, when operation fails because of not enough replicas having replicated the Blob,
+   * the operation will trigger the on-demand replication to replicate the Blob to more replicas and then retry the request.
+   */
+  @Config(ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED)
+  @Default("false")
+  public final boolean routerRepairWithReplicateBLobEnabled;
+
   // Group compression-related configs in the CompressConfig class.
   private final CompressionConfig compressionConfig;
 
@@ -780,6 +789,8 @@ public class RouterConfig {
         ROUTER_NOT_FOUND_CACHE_MAX_TTL_IN_MS);
     routerUpdateOpMetadataRelianceTimestampInMs = verifiableProperties.getLong(
         ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS, DEFAULT_ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS);
+    routerRepairWithReplicateBLobEnabled =
+        verifiableProperties.getBoolean(ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED, false);
 
     compressionConfig = new CompressionConfig(verifiableProperties);
   }
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -645,7 +645,7 @@ public class RouterConfig {
    */
   @Config(ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED)
   @Default("false")
-  public final boolean routerRepairWithReplicateBLobEnabled;
+  public final boolean routerRepairWithReplicateBlobEnabled;
 
   // Group compression-related configs in the CompressConfig class.
   private final CompressionConfig compressionConfig;
@@ -789,7 +789,7 @@ public class RouterConfig {
         ROUTER_NOT_FOUND_CACHE_MAX_TTL_IN_MS);
     routerUpdateOpMetadataRelianceTimestampInMs = verifiableProperties.getLong(
         ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS, DEFAULT_ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS);
-    routerRepairWithReplicateBLobEnabled =
+    routerRepairWithReplicateBlobEnabled =
         verifiableProperties.getBoolean(ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED, false);
 
     compressionConfig = new CompressionConfig(verifiableProperties);
--- a/ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobCallback.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobCallback.java
@@ -1,0 +1,91 @@
+/**
+ * Copyright 2022 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.router;
+
+import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.commons.BlobId;
+import com.github.ambry.commons.Callback;
+
+
+/**
+ * If some operation like {@link TtlUpdateOperation} fails due to some replicas don't have the Blob.
+ * We use {@link ReplicateBlobOperation} to replicate the blob to the NOT_FOUND replicas.
+ * After that, retry the failed operation for example the {@link TtlUpdateOperation}
+ * ReplicateBlobCallback is the callback function of the {@link ReplicateBlobOperation},
+ * also it stores the retry state.
+ */
+public class ReplicateBlobCallback implements Callback<Void> {
+
+  public enum RetryState {
+    REPLICATING_BLOB, // replicating the blob
+    REPLICATION_DONE, // blob replication is done
+    RETRYING          // retry the original failed operation
+  }
+
+  private final BlobId blobId;
+  private final DataNodeId sourceDataNode;
+  private RetryState retryState;
+  private Exception exception;
+
+  public ReplicateBlobCallback(BlobId blobId, DataNodeId sourceDataNode) {
+    this.blobId = blobId;
+    this.sourceDataNode = sourceDataNode;
+    this.retryState = RetryState.REPLICATING_BLOB;
+  }
+
+  /**
+   * @return the BlobId
+   */
+  public BlobId getBlobId() {
+    return blobId;
+  }
+
+  /**
+   * @return the source data node from which we get the Blob
+   */
+  public DataNodeId getSourceDataNode() {
+    return sourceDataNode;
+  }
+
+  /**
+   * @return the {@link RetryState}
+   */
+  public RetryState getState() {
+    return retryState;
+  }
+
+  /**
+   * Change the {@link RetryState}
+   */
+  public void setState(RetryState newState) {
+    this.retryState = newState;
+  }
+
+  /**
+   * @return the {@link Exception} of the ReplicateBlob operation
+   */
+  public Exception getException() {
+    return exception;
+  }
+
+  /**
+   * Will be called when the ReplicateBlob operation is done.
+   * @param exception The exception that was reported on execution of the request (if any).
+   */
+  @Override
+  public void onCompletion(Void v, Exception exception) {
+    this.exception = exception;
+    this.retryState = RetryState.REPLICATION_DONE;
+  }
+}
--- a/ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobCallback.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobCallback.java
@@ -27,7 +27,13 @@ import com.github.ambry.commons.Callback;
  */
 public class ReplicateBlobCallback implements Callback<Void> {
 
-  public enum RetryState {
+  /**
+   * There are three stages during the retry
+   * 1. Sends out the ReplicateBlob, waiting for the response
+   * 2. ReplicateBlob is finished. ReplicateBlobCallback is called.
+   * 3. Retrying the original operation like TtlUpdateOperation or DeleteOperation
+   */
+  public enum State {
     REPLICATING_BLOB, // replicating the blob
     REPLICATION_DONE, // blob replication is done
     RETRYING          // retry the original failed operation
@@ -35,13 +41,13 @@ public class ReplicateBlobCallback implements Callback<Void> {
 
   private final BlobId blobId;
   private final DataNodeId sourceDataNode;
-  private volatile RetryState retryState;
+  private volatile State retryState;
   private Exception exception;
 
   public ReplicateBlobCallback(BlobId blobId, DataNodeId sourceDataNode) {
     this.blobId = blobId;
     this.sourceDataNode = sourceDataNode;
-    this.retryState = RetryState.REPLICATING_BLOB;
+    this.retryState = State.REPLICATING_BLOB;
   }
 
   /**
@@ -59,16 +65,16 @@ public class ReplicateBlobCallback implements Callback<Void> {
   }
 
   /**
-   * @return the {@link RetryState}
+   * @return the {@link State}
    */
-  public RetryState getState() {
+  public State getState() {
     return retryState;
   }
 
   /**
-   * Change the {@link RetryState}
+   * Change the {@link State}
    */
-  public void setState(RetryState newState) {
+  public void setState(State newState) {
     this.retryState = newState;
   }
 
@@ -86,6 +92,6 @@ public class ReplicateBlobCallback implements Callback<Void> {
   @Override
   public void onCompletion(Void v, Exception exception) {
     this.exception = exception;
-    this.retryState = RetryState.REPLICATION_DONE;
+    this.retryState = State.REPLICATION_DONE;
   }
 }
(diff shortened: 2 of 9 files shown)


ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -355,6 +362,15 @@ private void onErrorResponse(ReplicaId replicaId, RouterException exception, boo
   private void checkAndMaybeComplete() {
     // operationCompleted is true if Blob_Authorization_Failure was received.
     if (operationTracker.isDone() || operationCompleted) {
+      // JING TODO next revision, will add configuration for this feature
+      if (!operationTracker.hasSucceeded()) {
+        // the operation is failed, try to recover with ReplicateBlob
+        // if retryWithReplicateBlob returns true, it's under retry. operation is not completed yet.
+        if (retryWithReplicateBlob()) {
```
justinlin-linkedin, line 369: Retry should be done asynchronous, right. We still return 503 back to client and at the same time, we run the replicate blob so next time client comes back with the same ttlupdate request, it would be a success.

panchow (changes requested):

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -388,6 +403,61 @@ private void checkAndMaybeComplete() {
...
+          // retry the TtlUpdate
+          operationException.set(null);
+          operationTracker =
+              new SimpleOperationTracker(routerConfig, RouterOperation.TtlUpdateOperation, blobId.getPartition(),
+                  originatingDcName, false, routerMetrics, blobId);
+          return true;
+        }
+      case RETRYING:
+      default:
+        // retry is done but still failed. Nothing else we can do.
```
SophieGuo410, line 456: Are we suppose to through exception when we hit this point? We only support three enum type right?

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
```diff
@@ -198,22 +199,24 @@ void handleResponse(ResponseInfo responseInfo, ReplicateBlobResponse replicateBl
           ServerErrorCode serverError = replicateBlobResponse.getError();
           if (serverError == ServerErrorCode.No_Error || serverError == ServerErrorCode.Blob_Already_Exists) {
             operationTracker.onResponse(replica, TrackedRequestFinalState.SUCCESS);
+            logger.info("ReplicateBlob {} successful for replica {} in {} with correlationId {} ",
+                replicateBlobRequest.getBlobId(), replica.getDataNodeId(), replica.getDataNodeId().getDatacenterName(),
+                replicateBlobRequest.getCorrelationId());
             if (RouterUtils.isRemoteReplica(routerConfig, replica)) {
-              logger.trace("Cross colo request successful for remote replica {} in {} ", replica.getDataNodeId(),
```
snalli, line 202: why remove ?

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -388,6 +403,61 @@ private void checkAndMaybeComplete() {
...
 
+  /**
+   * Try to replicate the blob and then retry the TtlUpdate Request
+   * @return true if it's under retry. return false if cannot do retry or retry is finished.
+   */
+  private boolean retryWithReplicateBlob() {
+    // the conditions to trigger ReplicateBlob operation:
+    // 1. at least one replica returned successful status
+    // 2. and at least one replica returned NOT_FOUND status.
+    if (!operationTracker.hasNotFound() || operationTracker.getSuccessReplica().size() == 0) {
```
snalli, line 414: double negatives are confusing
not hasNotFound == hasFound 

please make a helper hasFound()

snalli: Also why return when the operationTracker.getSuccessReplica().size() == 0 ?
Doesn't it mean no replicas were found to have the blob ?

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -388,6 +403,61 @@ private void checkAndMaybeComplete() {
...
+      case REPLICATING_BLOB:
+        LOGGER.trace("Running on-demand replication {} {}.", replicateBlobCallback.getBlobId(),
+            replicateBlobCallback.getSourceDataNode());
+        return true;
+      case REPLICATION_DONE:
+        if (replicateBlobCallback.getException() != null) {
+          LOGGER.error("on-demand replication {} {} failed.", replicateBlobCallback.getBlobId(),
+              replicateBlobCallback.getSourceDataNode(), replicateBlobCallback.getException());
+          return false;
+        } else {
```
snalli, line 442: this else statement can be removed

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -388,6 +403,61 @@ private void checkAndMaybeComplete() {
...
+          LOGGER.info("On-demand replication {} {} is successful.", replicateBlobCallback.getBlobId(),
+              replicateBlobCallback.getSourceDataNode());
+          replicateBlobCallback.setState(ReplicateBlobCallback.RetryState.RETRYING);
+
+          // retry the TtlUpdate
+          operationException.set(null);
+          operationTracker =
```
SophieGuo410, line 449: If we hit to this point, the operationTracker should been set already right?

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
JingQianCloud: Yes, as the comment, these two conditions are 
1. At least there is one replica returned NOT_FOUND. So it makes sense to do the ReplicateBlob.
2. At least here is one replica is successful. So we can get the Blob from that replica.

Change it to "hasFound" logically is not not we want to express.

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobOperation.java
```diff
@@ -171,25 +171,26 @@ void handleResponse(ResponseInfo responseInfo, ReplicateBlobResponse replicateBl
         requestLatencyMs);
     // Check the error code from NetworkClient.
     if (responseInfo.getError() != null) {
-      logger.trace("ReplicateBlobRequest with response correlationId {} timed out for replica {} ",
-          replicateBlobRequest.getCorrelationId(), replica.getDataNodeId());
+      logger.info("ReplicateBlobRequest {} with response correlationId {} timed out for replica {} ",
```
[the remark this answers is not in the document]
JingQianCloud: Changed.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -355,6 +362,14 @@ private void onErrorResponse(ReplicaId replicaId, RouterException exception, boo
   private void checkAndMaybeComplete() {
     // operationCompleted is true if Blob_Authorization_Failure was received.
     if (operationTracker.isDone() || operationCompleted) {
+      if (routerConfig.routerRepairWithReplicateBLobEnabled && !operationTracker.hasSucceeded()) {
+        // the operation is failed, try to recover with ReplicateBlob
+        // if retryWithReplicateBlob returns true, it's under retry. operation is not completed yet.
+        if (retryWithReplicateBlob()) {
```
justinlin-linkedin, line 368: I don't this is the best place to put this statement. The `checkAndMaybeComplete` should be restructured as such
```
if (operationTracker.isDone() || operationCompleted ) {
    if (operationTracker.hasSucceeded()) {
        operationException.set(null);
    } else if (operationTracker.getSuccessCount() > 0){
        // We have one success
        if (shouldRetryOnOneSuccess(operationException)) { // Exception that has precedency over AmbryUnavailable shouldn't be retried
            // Retry with Replicate Blob
        }
    } else if (operationTracker.maybeFailedDueToOfflineReplicas()) {
        operationException.set(new RouterException("TtlUpdateOperation failed possibly because of offline replicas", RouterErrorCode.AmbryUnavailable));
    } else if (operationTracker.hasFailedOnNotFound()) {
         operationException.set(new RouterException("TtlUpdateOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
    }
}
```

The reason to set the order as this, is because 
1. Success already take precedence over failures
2. As long as there is one success, and the exception is not Authorization Error, or Deleted Error, or UpdateNotAllow Error, we retry
3. other errors.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
justinlin-linkedin, line 414: we can put this if statement out side of retry method like my snippet above.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -388,6 +403,61 @@ private void checkAndMaybeComplete() {
...
+      replicateBlobCallback = new ReplicateBlobCallback(blobId, sourceDataNode);
+
+      nonBlockingRouter.replicateBlob(blobId.getID(), this.getClass().getSimpleName(), sourceDataNode,
+          replicateBlobCallback);
+      LOGGER.trace("Start the on-demand replication {} {}.", replicateBlobCallback.getBlobId(),
+          replicateBlobCallback.getSourceDataNode());
+      return true;
+    }
+
+    ReplicateBlobCallback.RetryState state = replicateBlobCallback.getState();
```
justinlin-linkedin, line 431: if we get rid of the State, we can change this to something like
```
if (replicateBlobCallback == null) { 
    // start replicateBlobCallback
} else {
    if (!replicateBlobCallback.isDone()) {
        // wait for this callback to be done
    } else {
        if (replicateBlobCallback.getException() != null) {
            // it's over
        } else {
            operationCompleted = false;
            operationException.set(null);
            operationTracker = new SimpleOperationTracker(....);
            ttlUpdateRequestInfos.clear() // Don't forget about this one.
        }
    }
}
```

ambry-router/src/main/java/com/github/ambry/router/ReplicateBlobCallback.java
```diff
@@ -0,0 +1,91 @@
...
+ * ReplicateBlobCallback is the callback function of the {@link ReplicateBlobOperation},
+ * also it stores the retry state.
+ */
+public class ReplicateBlobCallback implements Callback<Void> {
+
+  public enum RetryState {
+    REPLICATING_BLOB, // replicating the blob
+    REPLICATION_DONE, // blob replication is done
+    RETRYING          // retry the original failed operation
+  }
```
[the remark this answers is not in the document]
JingQianCloud: One boolean "isDone" is not enough. It can be used to indicate if the ReplicateBlob is done. But we also have retry state(retrying the TtlUpdate).
In checkAndMaybeComplete, we couldn't differentiate these two cases
1. first TtlUpdate failed, isDone is true
2. second TtlUpdate failed. isDone is true

For case 1, we should retry. 
For case 2, we should stop retrying and fail the request. 
Comparing with two boolean, isDone, hasRetried, I prefer using one single enum state.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
JingQianCloud: If TtlUpdate retry fails, we may end with a loop to retry again and again, right?

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
JingQianCloud: I think we have two approaches here,
1. One is the solution I'm currently using right now. Encapsulate everything into the retryWithReplicateBlob(). With that, we can easily port it to the DeleteBlob handling.

2. The other approach is as you suggested, add the code to the existing logic.
But the above "// Retry with Replicate Blob" is not a single step operation. Actually it has three stages,
a. Send ReplicateBlobRequest
b. Wait for it to be done.
c. recreate the SimpleOperationTracker and rerun the TtlUpdate.
For each sub stage, we need do different things. Also we need make sure if the TtlUpdate retry fail, don't retry again. So we need another boolean to indicate retry has been done.

if (operationTracker.isDone() || operationCompleted ) {
    if (operationTracker.hasSucceeded()) {
        operationException.set(null);
    } else if (isRunningReplicateBlob()) { // currently is executing ReplicateBlob
       // do nothing
       return;
    } else if (shouldReplicateBlob()) { // haven't do replicateBlob yet but we should call ReplicateBlob)
       // send replicate blob request
       return;
    } else if (replicateBlobCallback.isDone) {
       if (replicateBlobCallback.getException() != null) {
         // this will be the dirty part. 
         // Actually we should run the below "else if" condition "maybeFailedDueToOfflineReplicas or operationTracker.hasFailedOnNotFound" now
        // so have to move replicateBlobCallback.getException out side of this if-else-if-else chain.
        } else {
         // retry the Ttlupdate by recreating the SimpleOperationTracker etc.
          operationTracker =
              new SimpleOperationTracker(routerConfig, RouterOperation.TtlUpdateOperation, blobId.getPartition(),
                  originatingDcName, false, routerMetrics, blobId);
          ttlUpdateRequestInfos.clear();
          return true;
        }
    } else if (operationTracker.maybeFailedDueToOfflineReplicas()) {
        operationException.set(new RouterException("TtlUpdateOperation failed possibly because of offline replicas", RouterErrorCode.AmbryUnavailable));
    } else if (operationTracker.hasFailedOnNotFound()) {
         operationException.set(new RouterException("TtlUpdateOperation failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist));
    }
}

I feel embedding the retry logic into the current code logic is not clean.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
snalli: 1. So the first part of your comment explains the first part of your boolean and the second part explains the second part of the boolean ?
2. A negation of AND expression instead of OR makes more sense - we want to replicate if at least one replica returned NOT_FOUND && another one returned success  otherwise we don't replicate ?

```
if (! (operationTracker.hasNotFound() && operationTracker.getSuccessReplica().size() > 0)) {
  return false
}
```

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
snalli: No i do not think its readable, but up to you.
The if-statement here is a guard code preventing the main code below in case of an error and that should be evident in the way its written. Further this style of write adds more indents burying the main code deeper.

```
case REPLICATION_DONE:

    if (replicateBlobCallback.getException() != null) {
      LOGGER.error("on-demand replication {} {} failed.", replicateBlobCallback.getBlobId(),
          replicateBlobCallback.getSourceDataNode(), replicateBlobCallback.getException());
      return false;
    }

  LOGGER.info("On-demand replication {} {} is successful.", replicateBlobCallback.getBlobId(),
      replicateBlobCallback.getSourceDataNode());
  replicateBlobCallback.setState(ReplicateBlobCallback.RetryState.RETRYING);

  // retry the TtlUpdate
  operationException.set(null);
  operationTracker =
      new SimpleOperationTracker(routerConfig, RouterOperation.TtlUpdateOperation, blobId.getPartition(),
          originatingDcName, false, routerMetrics, blobId);
  ttlUpdateRequestInfos.clear();
  return true;

```

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
justinlin-linkedin: the retry logic can still be encapsulated in a single method, so we would merge `isRunningReplicateBlob` `shouldReplicaBlob` and `replicateBlobCallback.isDone`, just like the method you have now in the PR. My point is that we have to carefully structure the if statements.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
JingQianCloud: Moved this condition out to the caller.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
JingQianCloud: Done.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
JingQianCloud: Refined.

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
```diff
@@ -349,34 +356,60 @@ private void onErrorResponse(ReplicaId replicaId, RouterException exception, boo
...
+   * Check if we should replicate the blob and retry the operation.
+   * @return true if need to do the retry.
+   */
+  private boolean shouldReplicateBlobAndRetry() {
+    // the conditions to trigger ReplicateBlob and retry the original operation:
+    // 1. the feature is enabled.
+    // 2. and at least one replica returned NOT_FOUND status.
+    // 3. and at least one replica returned successful status.
+    return (routerConfig.routerRepairWithReplicateBlobEnabled && operationTracker.hasNotFound()
+        && operationTracker.getSuccessCount() > 0);
```
justinlin-linkedin, line 369: I don't think this is the right condition. If you have one success and one one delete, this is probably not retry-able anymore. I think the right condition should be that we don't an error code that has precedence over AmbryUnavailable. Something like
```
RouterErrorCode errorCode = ((RouterException)operationException.get()).getErrorCode();
return routerConfig.ReplicateBlobEnabled && operationTracker.getSuccessCount() > 0 && getPrecedence(errorCode) >= getPrecedence(RouterErrorCode.AmbryUnavailable);
```

justinlin-linkedin (commented):
the rest looks fine to me, just one minor comment

ambry-router/src/main/java/com/github/ambry/router/TtlUpdateOperation.java
JingQianCloud: You are right. Thanks for catching this.

Added the error code precedence condition. (Sorry I didn't get it from your first comment about it.)
I think we should still keep hasNotFound condition because ReplicateBlob only helps if there is some replica returned NOT_FOUND. If 8 replicas are unavailable, won't help either.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -638,6 +639,14 @@ public class RouterConfig {
   @Config(ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS)
   public final long routerUpdateOpMetadataRelianceTimestampInMs;
 
+  /**
+   * If this config is set to {@code true}, when operation fails because of not enough replicas having replicated the Blob,
+   * the operation will trigger the on-demand replication to replicate the Blob to more replicas and then retry the request.
+   */
+  @Config(ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED)
+  @Default("false")
```
panchow, line 647: You can remove the @Default annotation.  It is not used at all.  It only causes confusion.

panchow (approved):

SophieGuo410 (approved):

justinlin-linkedin (approved):
LGTM

PR #2265, opened 2022-10-17
Do not catch exception when there is error in put/delete for old table

mikeyang923:
Throw exception when there is conflict error in put for old table


ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -337,8 +337,8 @@ public CompletableFuture<PutResult> put(NamedBlobRecord record, NamedBlobState s
           PutResult putResult = null, putResultV2 = null;
           try {
             putResult = run_put(record, accountId, containerId, connection);
-          } catch (Exception eOld) {
-            if (!config.dbRelyOnNewTable) {
+          } catch (RestServiceException eOld) {
+            if (!config.dbRelyOnNewTable || eOld.getErrorCode() == RestServiceErrorCode.Conflict) {
```
snalli, line 341: whenever the boolean grows long, make a helper. Its likely it'll grow.
i don't know what this expression means but a readable helper would instantly convey the meaning.
I can think of the following:
```
void onConflictError(eOld) throws RestServiceException {
  // your logic to throw here
}
```

snalli: And please make a new catch block for this exception.

snalli:
```
catch (NewException) {
 onNewException()
}
```
something like this

justinlin-linkedin:
what is the best way going forward? If we want to compare content fetched from both old and new tables, maybe we should always stop the transaction as long as one of the table throws exception, regardless what exception they are throwing.

mikeyang923:
> what is the best way going forward? If we want to compare content fetched from both old and new tables, maybe we should always stop the transaction as long as one of the table throws exception, regardless what exception they are throwing.

Yeah, I agree, I just updated, can you take another look?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: This is outdated, I updated per Justin's comment in 
2e39a00. Can you take another look?

justinlin-linkedin (approved):
LGTM

panchow (approved):

snalli (approved):
if this is the only change, then lgtm

snalli:
actually mike, you may wanna change the title of the PR now since we are not throwing any exceptions

snalli (changes requested):
actually mike, you may wanna change the title of the PR now since we are not throwing any exceptions

mikeyang923:
> actually mike, you may wanna change the title of the PR now since we are not throwing any exceptions

Good point, just updated the PR title. Can you take another look?

PR #2267, opened 2022-10-17
Add isCompressed to PutRequest V5

panchow:
A new version of PutRequest V5 is introduced but disabled, so it continues to use V4 as the default.
The new version has aded the "isCompressed" field, serialization and de-serialization would apply PutRequest V5 only if the version is set to V5.

Added unit test to verify behavior.


ambry-protocol/src/test/java/com/github/ambry/protocol/PutRequestTest.java
```diff
@@ -0,0 +1,77 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import org.junit.Assert;
+import org.junit.Test;
+
+
+/**
+ * PutRequest testing.  Add more unit tests here to test all public methods and code changes to the PutRequest.
+ */
+public class PutRequestTest {
```
justinlin-linkedin, line 34: we have a test class RequestResponseTest where we do all the Request/Response test, please move this test case to it.

justinlin-linkedin (commented):
The rest looks good to me, just one minor comments

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
```diff
@@ -44,6 +44,7 @@ public class PutRequest extends RequestOrResponse {
   protected final BlobProperties properties;
   protected final BlobType blobType;
   protected final ByteBuffer blobEncryptionKey;
+  protected boolean isCompressed;
```
snalli, line 47: final

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
```diff
@@ -213,6 +268,9 @@ protected void prepareBuffer() {
     if (keyLength > 0) {
       bufferToSend.writeBytes(blobEncryptionKey);
     }
+    if (versionId >= PUT_REQUEST_VERSION_V5) {
```
snalli, line 271:
```
if (versionId >= PUT_REQUEST_VERSION_V5 && isCompressed) {
 bufferToSend.writeByte(1);
}
```

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
```diff
@@ -213,6 +268,9 @@ protected void prepareBuffer() {
...
+      bufferToSend.writeByte(isCompressed ? 1 : 0);
```
snalli, line 272:
```
    if (versionId >= PUT_REQUEST_VERSION_V5) {
      bufferToSend.writeByte(Boolean.compare(isCompressed, true));
    }
```

snalli:
Added unit to verify behavior -> Added unit test to verify behavior.

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
panchow: Isn't it easier to read using (isCompressed ? 1 : 0)? This pattern is used in the rest of the Ambry code too.

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
panchow: thank you

ambry-protocol/src/test/java/com/github/ambry/protocol/PutRequestTest.java
panchow: Moved the test to the RequestResponseTest class.

justinlin-linkedin (approved):
lgtm

ambry-protocol/src/main/java/com/github/ambry/protocol/PutRequest.java
snalli: Both are easy enough but if Java provides a method, I think it's better to use it. 
Agreed this pattern is in rest of Ambry, but we can try changing that henceforth.

snalli: But nvm, the pr lgtm.

PR #2268, opened 2022-10-18
Added request parallelism and success target config for ReplicateBlob

JingQianCloud:
Added ReplicateBlobRequestParallelism and ReplicateBlobSuccessTarget

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -59,6 +59,8 @@ public class RouterConfig {
   public static final String ROUTER_MAX_PUT_CHUNK_SIZE_BYTES = "router.max.put.chunk.size.bytes";
   public static final String ROUTER_PUT_REQUEST_PARALLELISM = "router.put.request.parallelism";
   public static final String ROUTER_PUT_SUCCESS_TARGET = "router.put.success.target";
+  public static final String ROUTER_REPLICATE_BLOB_REQUEST_PARALLELISM = "router.replicate.blob.request.parallelism";
+  public static final String ROUTER_REPLICATE_BLOB_SUCCESS_TARGET = "router.replicate.blob.success.target";
   public static final String ROUTER_MAX_SLIPPED_PUT_ATTEMPTS = "router.max.slipped.put.attempts";
   public static final String ROUTER_DELETE_REQUEST_PARALLELISM = "router.delete.request.parallelism";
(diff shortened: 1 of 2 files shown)


SophieGuo410 (approved):

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -234,6 +236,20 @@ public class RouterConfig {
   @Default("2")
   public final int routerPutSuccessTarget;
 
+  /**
+   * The maximum number of parallel requests issued at a time by the ReplicateBlob manager.
+   */
+  @Config(ROUTER_REPLICATE_BLOB_REQUEST_PARALLELISM)
+  @Default("3")
```
snalli, line 243: confused. Why are we issuing 3 replicateBlob requests ? 
If a ttl-update fails due to 1 missing replica , then shdn't we just issue that. Basically, we shd issue the minimum number of replicateBlob requests just enough for the ttl-update or whatever other mutation to succeed.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -234,6 +236,20 @@ public class RouterConfig {
...
+  public final int routerReplicateBlobRequestParallelism;
+
+  /**
+   * The minimum number of successful responses required for a ReplicateBlob operation.
+   */
+  @Config(ROUTER_REPLICATE_BLOB_SUCCESS_TARGET)
+  @Default("2")
```
justinlin-linkedin, line 250: Should we make this 3?
It's likely that we are going to hit the source data node in ReplicateBlob operations, which is a guarantee success (if it's still available). In this case, if we set the success target to 3, then we would probably have one cross-colo replication. This would greatly increase the success of the following TTL_UPDATE operation.

JingQianCloud: About excluding the source host, definitely we can do it. I'll clean up all the ON_DEMAND_TO_DO before production. 

Changing the success number from 2 to 3 may have one side effect.
If we only have two replicas left to replicate the Blob, (may include the source), suppose we can return success status and retry the TtlUpdate successful. But now we think ReplicateBlob failed, so we don't retry the TtlUpdate in the TtlUpdateOperation.
Although client retry may still be successful, I guess it's not our intention.

justinlin-linkedin: that makes sense, given that we might have to use this feature in a local cluster with any other colos.

justinlin-linkedin (approved):

PR #2269, opened 2022-10-19
Remove unnecessary static key in Account and Container

justinlin-linkedin:
Remove the static variables in Account and Containers, which serve as keys for json object. We are already using Jackson library, we don't need those keys anymore.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/account/AccountBuilder.java
+++ b/ambry-api/src/main/java/com/github/ambry/account/AccountBuilder.java
@@ -228,8 +228,10 @@ public class AccountBuilder {
     }
     // Did we check the container parent account id here?
     for (Map.Entry<Short, Container> entry : idToContainerMetadataMap.entrySet()) {
-      if (entry.getValue().getParentAccountId() == Container.UNKNOWN_CONTAINER_PARENT_ACCOUNT_ID) {
+      if (entry.getValue().getParentAccountId() != id) {
         entry.setValue(new ContainerBuilder(entry.getValue()).setParentAccountId(id).build());
+      } else {
+        entry.setValue(new ContainerBuilder(entry.getValue()).build());
       }
     }
     return new Account(id, name, status, aclInheritedByContainer, snapshotVersion, idToContainerMetadataMap.values(),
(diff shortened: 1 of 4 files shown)


ambry-api/src/test/java/com/github/ambry/account/AccountContainerTest.java
```diff
@@ -164,33 +137,31 @@ public void testConstructAccountAndContainerFromArguments() throws JSONException
...
-            .setCacheable(refContainerCachingValues.get(0))
-            .setBackupEnabled(refContainerBackupEnabledValues.get(0))
-            .setMediaScanDisabled(refContainerMediaScanDisabledValues.get(0))
-            .setReplicationPolicy(refContainerReplicationPolicyValues.get(0))
-            .setTtlRequired(refContainerTtlRequiredValues.get(0))
-            .setContentTypeWhitelistForFilenamesOnDownload(
-                refContainerContentTypeAllowListForFilenamesOnDownloadValues.get(0))
-            .setDeleteTriggerTime(refContainerDeleteTriggerTime.get(0))
-            .build());
+    containers.add(new ContainerBuilder((short) 0, "0", refContainerStatuses.get(0), refContainerDescriptions.get(0),
```
panchow, line 140: This pattern has repeated multiple times.  Would it make be simpler to create a function that takes an Id and name?

panchow (changes requested):

ambry-api/src/test/java/com/github/ambry/account/AccountContainerTest.java
justinlin-linkedin: I have another PR ready to be published. This would be done in that PR.

SophieGuo410 (approved):

ambry-api/src/main/java/com/github/ambry/account/AccountBuilder.java
```diff
@@ -228,8 +228,10 @@ public Account build() {
     }
     // Did we check the container parent account id here?
     for (Map.Entry<Short, Container> entry : idToContainerMetadataMap.entrySet()) {
-      if (entry.getValue().getParentAccountId() == Container.UNKNOWN_CONTAINER_PARENT_ACCOUNT_ID) {
+      if (entry.getValue().getParentAccountId() == Container.UNKNOWN_CONTAINER_ID) {
```
ankagrawal, line 231: nit: In terms of value this might work fine, because we map both `UNKNOWN_CONTAINER_ID` and `UNKNOWN_CONTAINER_PARENT_ACCOUNT_ID` to same values. But logically, they can be different.

Since we aren't getting rid of `UNKNOWN_CONTAINER_PARENT_ACCOUNT_ID` , we don't need to change here.

ankagrawal (approved):
Looks good to me, with one nitpick.

PR #2271, opened 2022-10-19
Support Blob Format V3 in PutMessage stream.

panchow:
- Blob Data format V3 is introduced to include compression information.  By default, it is disabled.  Future PR will enable it.  The unit tests continue to test V2 plus the newly introduced V3.

This change seems big but probably 95% of the changes are unit test and all related to BlobFormat version selection.  It's not appropriate to sub-divide this PR.   As shown in this PR, we continue to unit test V1 and V2 Blob Data format serialization.  However, we should remove V1 and V2 serialization code and unit tests since they should no longer be used after V3 is deployed.  When I submit the clean PR in the future, I plan to remove V1 and V2 serialization.


mikeyang923 (approved):
lgtm

justinlin-linkedin (approved):
lgtm

snalli:
merging since it has two approvals already

PR #2272, opened 2022-10-20
Fix checking of error counts in SimpleOperationTracker for offline/not-found responses

Arun-LinkedIn:
As seen in this [code](https://github.com/linkedin/ambry/blob/aeb6b93e87d0076a7243702b0e2f13acfe3c6d31/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java#L274), when [routerGetEligibleReplicasByStateEnabled](https://github.com/linkedin/ambry/blob/aeb6b93e87d0076a7243702b0e2f13acfe3c6d31/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java#L512) is enabled, the list of replicas for a given partition is fetched by calling [partition.getReplicaIdsByState](https://github.com/linkedin/ambry/blob/aeb6b93e87d0076a7243702b0e2f13acfe3c6d31/ambry-api/src/main/java/com/github/ambry/clustermap/PartitionId.java#L51). Else, they are fetched by calling [partition.getReplicaIds](https://github.com/linkedin/ambry/blob/aeb6b93e87d0076a7243702b0e2f13acfe3c6d31/ambry-api/src/main/java/com/github/ambry/clustermap/PartitionId.java#L43). The former fetches replicas (and their states) from Helix external view API ([RoutingTableSnapshot](https://github.com/linkedin/ambry/blob/aeb6b93e87d0076a7243702b0e2f13acfe3c6d31/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java#L679)). The latter fetches the replicas from Ambry hosts/disk/partition/replica information stored in [property store](https://github.com/linkedin/ambry/blob/aeb6b93e87d0076a7243702b0e2f13acfe3c6d31/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java#L630). The replica information in both these sources can be inconsistent during replica movement. For example, for a partition, when we moved its replicas {R1, R2, R3} to {R4, R5, R6}, `partition.getReplicaIds()` could return {R4, R5, R6} but `partition.getReplicaIdsByState()` could return {R1, R2, R3, R4, R5, R6} if helix view is old.

When keeping track of counts for total replicas, offline replicas, etc in SimpleOperationTracker, this PR makes sure to use consistent source for replica information.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -349,32 +349,32 @@ class SimpleOperationTracker implements OperationTracker {
     // Count of total and offline replicas in all data centers. This will be used later to check if we
     // should return 503 unavailable instead of 404 not found.
     if (routerConfig.routerGetEligibleReplicasByStateEnabled) {
-      for (List<ReplicaId> replicaIds : replicasByStateSnapshot.values()) {
-        totalReplicaCount += replicaIds.size();
-      }
+      replicasByStateSnapshot.values().forEach(replicaIds -> totalReplicaCount += replicaIds.size());
+      totalOfflineReplicaCount =
+          replicasByStateSnapshot.getOrDefault(ReplicaState.OFFLINE, Collections.emptyList()).size();
     } else {
       totalReplicaCount = partitionId.getReplicaIds().size();
+      // Since we are not looking into replica states, consider offline replica count as 0.
+      totalOfflineReplicaCount = 0;
     }
-    totalOfflineReplicaCount =
-        getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE)).getOrDefault(ReplicaState.OFFLINE,
-            Collections.emptyList()).size();
 
     // Count of total and offline replicas in originating data center. This will be used later to check if we
     // should return 503 unavailable instead of 404 not found.
     if (routerConfig.routerGetEligibleReplicasByStateEnabled) {
-      for (List<ReplicaId> replicaIds : getReplicasByState(originatingDcName,
-          EnumSet.allOf(ReplicaState.class)).values()) {
-        originatingDcTotalReplicaCount += replicaIds.size();
-      }
+      Map<ReplicaState, List<ReplicaId>> replicasByStateInOrigDC =
+          getReplicasByState(originatingDcName, EnumSet.allOf(ReplicaState.class));
+      replicasByStateInOrigDC.values().forEach(replicaIds -> originatingDcTotalReplicaCount += replicaIds.size());
+      originatingDcOfflineReplicaCount =
+          replicasByStateInOrigDC.getOrDefault(ReplicaState.OFFLINE, Collections.emptyList()).size();
     } else {
       for (ReplicaId replicaId : partitionId.getReplicaIds()) {
         if (replicaId.getDataNodeId().getDatacenterName().equals(originatingDcName)) {
           originatingDcTotalReplicaCount++;
         }
       }
+      // Since we are not looking into replica states, consider offline replica count as 0.
+      originatingDcOfflineReplicaCount = 0;
     }
-    originatingDcOfflineReplicaCount =
-        getReplicasByState(originatingDcName, EnumSet.of(ReplicaState.OFFLINE)).values().size();
 
     // MockPartitionId.getReplicaIds() is returning a shared reference which may cause race condition.
     // Please report the test failure if you run into this exception.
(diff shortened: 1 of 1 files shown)


mikeyang923 (approved):
lgtm

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -190,15 +189,10 @@ class SimpleOperationTracker implements OperationTracker {
...
     // to avoid the case where a replica state might change in between an operation.
-    allDcReplicasByState =
+    replicasByStateSnapshot =
         (Map<ReplicaState, List<ReplicaId>>) partitionId.getReplicaIdsByStates(EnumSet.allOf(ReplicaState.class), null);
     List<ReplicaId> eligibleReplicas;
     List<ReplicaId> offlineReplicas = new ArrayList<>();
-    totalOfflineReplicaCount =
-        getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE)).getOrDefault(ReplicaState.OFFLINE,
-            Collections.emptyList()).size();
-    allReplicas = partitionId.getReplicaIds().stream().collect(Collectors.toList());
```
Arun-LinkedIn, line 200: When `routerConfig.routerGetEligibleReplicasByStateEnabled`, list of eligible replicas for this operation comes from `partitionId.getReplicaIdsByState` which in turn fetches from helix `RoutingTableSnapshot`. On the other hand, `partitionId.getReplicaIds()` fetches from replica information updated by Ambry in helix property store. I think information in both places could be inconsistent. Due to that, we might need to make sure same source of information is used in all places when calculating counts. This is the main change in this PR.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -372,12 +391,24 @@ class SimpleOperationTracker implements OperationTracker {
...
+      // Count the number of replicas from originating dc that will be used for this operation and set the failure
+      // threshold for NOT_FOUND errors based on it.
+      int originatingDcOperationReplicaCount = 0;
+      for (ReplicaId replicaId : replicaPool) {
+        if (replicaId.getDataNodeId().getDatacenterName().equals(originatingDcName)) {
+          originatingDcOperationReplicaCount++;
+        }
+      }
+      originatingDcNotFoundFailureThreshold =
+          originatingDcOperationReplicaCount - routerConfig.routerPutSuccessTarget + 1;
```
JingQianCloud, line 411: I feel we should use "original DC total replica count" instead of "original DC operation replica count".
This threshold means that we have at least 2 Put in the local DC. So if local DC returns 2 NOT_FOUND, we think the blob doesn't exist.
After the change, in the local DC, if one replica is not in the operational state.  I assume the threshold will be 2-2+1 = 1. 
So as long as there is one local DC returned NOT_FOUND, we'll terminate it as NOT_FOUND. But it's possible the "not operational replica" has the Blob, just somehow it's not available at this moment, right? In this case, I think we shouldn't terminate it. We need check with remote replicas, right?

JingQianCloud (commented):
That's great that you found and fixed the helix inconsistent view in the PR.

Not necessary related to this PR. I'm a little bit concern about the following ticket. 
https://jira01.corp.linkedin.com:8443/browse/HELIX-4438
Helix may remove offline replica from the view. So totalOfflineReplicaCount may not be accurate based on the Helix External View.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -75,7 +75,7 @@ class SimpleOperationTracker implements OperationTracker {
   // How many NotFound responses from originating dc will terminate the operation.
   // It is set to tolerate one random failure in the originating dc if all other responses are not found.
   protected final int originatingDcNotFoundFailureThreshold;
-  protected final int totalReplicaCount;
+  protected final int operationReplicaCount;
```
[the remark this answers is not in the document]
JingQianCloud: I like the rename.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -349,20 +343,45 @@ class SimpleOperationTracker implements OperationTracker {
...
+          replicasByStateSnapshot.getOrDefault(ReplicaState.OFFLINE, Collections.emptyList()).size();
+    } else {
+      totalReplicaCount = partitionId.getReplicaIds().size();
+      // Since we are not looking into replica states, consider offline replica count as 0.
+      totalOfflineReplicaCount = 0;
+    }
+
+    // Count of total and offline replicas in originating data center. This will be used later to check if we
+    // should return 503 unavailable instead of 404 not found.
+    if (routerConfig.routerGetEligibleReplicasByStateEnabled) {
```
panchow, line 363: I wonder why we have 2 implementations and select based on config.  Is there plan to delete the else case?  Looking at AmbryLI, this config value is true for all frontend hosts except one machine "machine-ltx1-app56003.prod.linkedin.com.src" that has a value of false.  Anyone know why this host has special value?  I think we could simplify our code by removing deprecated old code.

panchow (approved):

Arun-LinkedIn:
Discarding this rb. Will revisit later

PR #2273, opened 2022-10-20
Adding two new fields to container

justinlin-linkedin:
Adding two new fields to container
1. CacheTtlInSecond: This would dictate how long the blob of this container can be cached by the cdn and the browser. This would be null and if it's null, the user of this field can set it a value that the user deems reasonable.
2. UserMetadataKeysToNotPrefixInResponse: For each user metadata, when we return it in the response, we already prefix the key with x-ambry-um-. For some applications, they want to return the user metadata without prefix. This new fields would be a set of user metadata keys that we would not prefix when returning in response.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
+++ b/ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
@@ -54,6 +54,9 @@ public class ContainerBuilder {
   private long lastModifiedTime = LAST_MODIFIED_TIME_DEFAULT_VALUE;
   private int snapshotVersion = SNAPSHOT_VERSION_DEFAULT_VALUE;
   private String accessControlAllowOrigin = ACCESS_CONTROL_ALLOW_ORIGIN_DEFAULT_VALUE;
+  private Long cacheTtlInSecond = CACHE_TTL_IN_SECOND_DEFAULT_VALUE;
+  private Set<String> userMetadataKeysToNotPrefixInResponse =
+      USER_METADATA_KEYS_TO_NOT_PREFIX_IN_RESPONSE_DEFAULT_VALUE;
 
   /**
    * Constructor. This will allow building a new {@link Container} from an existing {@link Container}. The builder will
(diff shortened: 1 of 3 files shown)


ambry-api/src/test/java/com/github/ambry/account/AccountContainerTest.java
```diff
@@ -888,6 +837,11 @@ private void assertContainer(Container container, int index) {
             container.getSnapshotVersion());
         assertEquals("Wrong accessControlAllowOrigin", refAccessControlAllowOriginValues.get(index),
             container.getAccessControlAllowOrigin());
+        assertEquals("Wrong cacheTtlInSeconds", refContainerCacheTtlInSeconds.get(index),
+            container.getCacheTtlInSecond());
+        assertEquals("Wrong user metadata keys to not prefix in response",
+            refContainerUserMetadataKeysToNotPrefixInResponses.get(index),
```
SophieGuo410, line 843: Same as refContainerContentTypeAllowListForFilenamesOnDownloadValues, if it's null, should return an empty set.

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
```diff
@@ -311,6 +316,26 @@ public ContainerBuilder setSnapshotVersion(int snapshotVersion) {
     return this;
   }
 
+  /**
+   * Sets the cacheTtlInSecond of the {@link Container} to build.
+   * @param cacheTtlInSecond The value to set.
+   * @return This builder.
+   */
+  public ContainerBuilder setCacheTtlInSecond(Long cacheTtlInSecond) {
+    this.cacheTtlInSecond = cacheTtlInSecond;
```
SophieGuo410, line 325: Probably this can only been set when Cacheable is true?

ambry-api/src/test/java/com/github/ambry/account/AccountContainerTest.java
justinlin-linkedin: you are right, updated.

ambry-api/src/main/java/com/github/ambry/account/ContainerBuilder.java
justinlin-linkedin: I want to make this builder as simple as possible. When we use these values, we would check if the container is cacheable or not first.

SophieGuo410: nit: probably the java doc can emphasize that in case someone don't aware of it.

SophieGuo410 (approved):

ankagrawal (approved):
Lgtm. I am assuming that the code to save the container settings for cache TTL and metadata keys with no prefix will come later.

PR #2276, opened 2022-10-22
Use config.CompressionEnabled to control new Blob Record V3 and PutRequest V5.

panchow:
Plus a few missed cleanup.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
@@ -127,8 +127,7 @@ public class CompressionService {
     }
     defaultCompressor = compressor;
 
-    // Temporary deployment control code that will be removed after testing and deployment.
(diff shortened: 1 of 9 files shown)


ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -124,6 +126,13 @@ public CompressionService(CompressionConfig config, CompressionMetrics metrics)
           + ", specified in config does not exist.  This default has changed to " + compressor.getAlgorithmName());
     }
     defaultCompressor = compressor;
+
+    // Temporary deployment control code that will be removed after testing and deployment.
+    // https://jira01.corp.linkedin.com:8443/browse/AMBRY-9410
+    if (config.isCompressionEnabled) {
```
mikeyang923, line 132: nit: can use 'isCompressionEnabled' directly here.

mikeyang923 (approved):

justinlin-linkedin (approved):
LGTM

PR #2278, opened 2022-10-25
Integrate compression with Put and Get Request

panchow:


ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -873,6 +883,34 @@ protected void maybeProcessCallbacks() {
       }
     }
 
+    /**
+     * Decompress the compressed buffer specified.
+     * If the buffer is not compressed, based on blobCompressed property, then return the input buffer.
+     *
+     * @param sourceBuffer The source buffer to decompress.
+     * @return New buffer if decompressed successfully and the source buffer is released.
+     *         Original buffer is blob not compressed.  NULL if failed and source buffer is released.
```
mikeyang923, line 892: typo 'riginal buffer is blob not compressed' ---> "riginal buffer if blob not compressed"?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -873,6 +883,34 @@ protected void maybeProcessCallbacks() {
...
+     */
+    protected ByteBuf decompressContent(ByteBuf sourceBuffer) {
+      if (!isChunkCompressed) {
+        // Blob is not compressed.  Return source buffer.
+        return sourceBuffer;
+      }
+
+      // Note: decompress() already emit metrics to for decompress() failure.
```
mikeyang923, line 900: Remove 'to'?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -873,6 +883,34 @@ protected void maybeProcessCallbacks() {
...
+        ByteBuf result = decompressionService.decompress(sourceBuffer, routerConfig.routerMaxPutChunkSizeBytes);
+        sourceBuffer.release();
+        return result;
+      } catch (Exception ex) {
+        logger.error("Failed to decompress chunk data for chunkBlobId " + chunkBlobId + " due to exception.", ex);
+        setOperationException(buildChunkException("Exception thrown on decompressing the data chunk", ex,
+            RouterErrorCode.UnexpectedInternalError));
+        sourceBuffer.release();
+        return null;
+      }
```
mikeyang923, line 911: Maybe better to do it in below way since looks we want to catch the exception in 'decompress' only, and avoid some duplicate code:
```
ByteBuf result = null;
try {
     result = decompressionService.decompress(sourceBuffer, routerConfig.routerMaxPutChunkSizeBytes);
 } catch (Exception ex) {
    log and set operation exception 
}
sourceBuffer.release();
return result;
```

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -852,8 +858,12 @@ protected void maybeProcessCallbacks() {
               routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
             }
             routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-            chunkIndexToBuf.put(chunkIndex, filterChunkToRange(result.getDecryptedBlobContent()));
-            numChunksRetrieved.incrementAndGet();
+            ByteBuf decryptedContent = result.getDecryptedBlobContent();
+            ByteBuf decompressedContent = decompressContent(decryptedContent);
+            if (decompressedContent != null) {
+              chunkIndexToBuf.put(chunkIndex, filterChunkToRange(decompressedContent));
+              numChunksRetrieved.incrementAndGet();
```
mikeyang923, line 865: Curious, in case of exception in decompress (therefore decompressedContent=null), do we still need to run `numChunksRetrieved.incrementAndGet()` or not?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -120,6 +120,14 @@ class PutOperation {
   private boolean isEncryptionEnabled;
   private final QuotaChargeCallback quotaChargeCallback;
 
+  // Whether the entire blob (all chunks) are compressible.  Default is null for undecided.
+  // The value is set based on blobProperties and compression service configuration.
+  // For example, if BlobProperties.ContentEncoding specified, then the entire blob will
+  // not be compressed by setting this field to FALSE.
+  // This field is set once when inserting the first chunk based on blob properties.
+  // All chunk compression will check this field and skip compression if this value FALSE.
+  private Boolean isBlobCompressible;
```
justinlin-linkedin, line 129: could just assign the value to this variable in the constructor, then you don't need to check if it's null for each chunk.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -305,6 +314,9 @@ private PutOperation(RouterConfig routerConfig, NonBlockingRouterMetrics routerM
     isEncryptionEnabled = passedInBlobProperties.isEncrypted();
     restRequest = options.getRestRequest();
     loggingContext = makeLoggingContext();
+
+    // TODO - use dependency injection when available.
+    compressionService = new CompressionService(routerConfig.getCompressionConfig(), routerMetrics.compressionMetrics);
```
justinlin-linkedin, line 319: we should create one CompressionService object and pass it PutOperation. CompressionService's constructor has some logics that will be annoy to see them run every single time when creating a PutOperation, like logging an error about default Compression, and setting PutRequest.Version.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1049,6 +1061,10 @@ class PutChunk {
     // Tracks quota charging for this chunk.
     private OperationQuotaCharger operationQuotaCharger;
 
+    // Whether this chunk is compressed.  Default is false for not compressed.
+    // This value is set after compression has completed.  It is used to create PutRequest.
+    private boolean isChunkCompressed;
```
justinlin-linkedin, line 1066: please reset this variable back to false in clear method.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1291,6 +1307,35 @@ private void encryptionCallback(EncryptJob.EncryptJobResult result, Exception ex
...
+      if (isBlobCompressible == null) {
+        isBlobCompressible = compressionService.isBlobCompressible(passedInBlobProperties);
+      }
+      if (!isBlobCompressible) {
+        return;
+      }
+
+      // Note: compress() returns null if it failed.
+      boolean isFullChunk = (buf.readableBytes() == routerConfig.routerMaxPutChunkSizeBytes);
+      ByteBuf newBuffer = compressionService.compressChunk(buf, isFullChunk);
```
justinlin-linkedin, line 1331: Now is probably the time to talk about ByteBuf management in compression.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
justinlin-linkedin, line 911: Handling decompression exceptions needs more thoughts. We are not going to serve data back to client when the decompression fails, so we have to set an exception for the entire operation. We can also set the operationCompleted to true so the operation would be stopped right away.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Thank you

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Thank you.  let me introduce finally to the try clause.  It is an alternative to this good suggestion.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: numChunksRetrieved is used to send back to client, not for statistic purpose.  So unless the content is available, we should not increment numChunksRetrieved.  Decrypt and decompress is part of the retrieving steps, so unless both steps succeeded, any failure should be considered as retrieval failure.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: CompressionService should be a singleton.  If we create it in operation manager and pass it to every PutOperation, that implies it's a singleton.  In my opinion, the best place for singleton is Dependency Injection which we don't have yet, but not through parameter passing.  In this case, passing a singleton to PutOperation constructor does not add "useful value" to the logic.  Changing many unit tests to match the new constructor also has no value.  In here, I define "value" as additional feature or improvements.  It is not a good idea to write/change code that has no value just to meet the compiler syntax requirement.  It will just occupy developer and reviewers time in the long run.  When we have changes the has wide impact, it is a signal that the coding structure might need improvement.

I will make the suggested change here until we have Dependency Injection. Together, let's continue to make improvements.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: ok thank you.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: Good idea.  Thank you.

mikeyang923 (approved):

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -873,6 +882,36 @@ protected void maybeProcessCallbacks() {
...
+      }
+
+      try {
+        return decompressionService.decompress(sourceBuffer, routerConfig.routerMaxPutChunkSizeBytes);
+      } catch (Exception ex) {
+        // Note: No need to emit metrics here because decompress() already emits failure metrics.
+        logger.error("Failed to decompress chunk data for chunkBlobId " + chunkBlobId + " due to exception.", ex);
+        setOperationException(buildChunkException("Exception thrown on decompressing the data chunk", ex,
+            RouterErrorCode.UnexpectedInternalError));
+        setOperationCompleted();
```
justinlin-linkedin, line 907: we should just set the operationCompleted=true without calling this method. This method would also release the previous chunk data. We don't have to call it here.

justinlin-linkedin (commented):
some minor comments

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
panchow: I agree.  Let me make that change.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
panchow: In this case, it should release other data chunks right?  If one chunk failed, the entire blob should fail.  Is there a reason why we want to keep the previous chunks of data?

justinlin-linkedin (approved):

PR #2279, opened 2022-10-25
Return correct cache TTL and user metadata based on container

justinlin-linkedin:
This PR start using two new fields in container to
1. Return custom ttl for cache in response header
2. Return user metadata in response header with prefix if container has a set of keys that requires no prefix.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
@@ -997,13 +997,7 @@ public class RestUtils {
    * @throws RestServiceException if there are any problems setting the header.
    */
   public static boolean setUserMetadataHeaders(byte[] userMetadata, RestResponseChannel restResponseChannel) {
-    Map<String, String> userMetadataMap = buildUserMetadata(userMetadata);
-    if (userMetadataMap != null) {
-      setUserMetadataHeaders(userMetadataMap, restResponseChannel);
-      return true;
-    } else {
-      return false;
-    }
+    return setUserMetadataHeaders(userMetadata, restResponseChannel, Collections.emptySet());
   }
 
   /**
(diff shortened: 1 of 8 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -378,9 +379,11 @@ private void setCacheHeaders(RestRequest restRequest, RestResponseChannel restRe
       throws RestServiceException {
     Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
     if (container.isCacheable()) {
+      long cacheTtl = container.getCacheTtlInSecond() != null ? container.getCacheTtlInSecond().longValue()
```
Arun-LinkedIn, line 382: QQ: Would it be safer to check for some minimum cache TTL value set by user? Since the requests could be "internet" facing urls, if a cache value has been mistakenly set to a very short value by the user, it could result in lot of requests coming to Ambry.

Arun-LinkedIn (approved):
Looks good to me. Just had one comment.

SophieGuo410 (approved):

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -1010,18 +1034,24 @@ public static boolean setUserMetadataHeaders(byte[] userMetadata, RestResponseCh
...
     for (Map.Entry<String, String> entry : userMetadataMap.entrySet()) {
       String headerName = entry.getKey();
       String headerValue = entry.getValue();
       // Ensure um prefix is prepended to header name
       if (!headerName.startsWith(Headers.USER_META_DATA_HEADER_PREFIX)) {
         headerName = Headers.USER_META_DATA_HEADER_PREFIX + headerName;
       }
+      String headerNameWithoutPrefix = headerName.substring(Headers.USER_META_DATA_HEADER_PREFIX.length());
+      if (keysToNotPrefix.contains(headerNameWithoutPrefix)) {
+        headerName = headerNameWithoutPrefix;
```
JingQianCloud, line 1053: Is there any possibility that after removing the prefix, the headName is same as some Ambry using standard header?
Also also does that mean customer can expose any headerName as they want with the container_without_prefix? Or probably we have some constraint for the container_without_prefix?

JingQianCloud (commented):
Some questions embedded.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -1010,18 +1034,24 @@ public static boolean setUserMetadataHeaders(byte[] userMetadata, RestResponseCh
...
     Objects.requireNonNull(userMetadataMap, "user metadata must be supplied");
+    Objects.requireNonNull(keysToNotPrefix, "keys to not prefix must be supplied");
     for (Map.Entry<String, String> entry : userMetadataMap.entrySet()) {
       String headerName = entry.getKey();
       String headerValue = entry.getValue();
       // Ensure um prefix is prepended to header name
       if (!headerName.startsWith(Headers.USER_META_DATA_HEADER_PREFIX)) {
         headerName = Headers.USER_META_DATA_HEADER_PREFIX + headerName;
       }
+      String headerNameWithoutPrefix = headerName.substring(Headers.USER_META_DATA_HEADER_PREFIX.length());
```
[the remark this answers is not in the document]
justinlin-linkedin: if it's not, the if statement above would add the prefix.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
justinlin-linkedin: That's a solid concern, however, it's probably fine. Since what goes to this set is controlled us in the end. We will manually examine the keys and make sure it doesn't conflict with standard http response header keys.

JingQianCloud: Yes, as long as we exam the list well in the container setting, it should be fine.

JingQianCloud (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
justinlin-linkedin: changing cache ttl for container will be an admin op. we should be good.

PR #2281, opened 2022-10-28
Catch ArrayIndexOutOfBoundsException

snalli:
Catch ArrayIndexOutOfBoundsException when parsing request path

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
@@ -130,6 +130,10 @@ public class RequestPath {
           throw new RestServiceException("Segment index given is not an integer", RestServiceErrorCode.BadRequest);
         }
         //otherwise, nothing to do
+      } catch (ArrayIndexOutOfBoundsException e) {
+        // For eg., errorMessage = "java.lang.ArrayIndexOutOfBoundsException: -1"
+        String errorMessage = e.getClass().getCanonicalName() + ": " + e.getMessage();
+        throw new RestServiceException(errorMessage, RestServiceErrorCode.BadRequest);
       }
     }
 
--- a/ambry-api/src/test/java/com/github/ambry/rest/RequestPathTest.java
+++ b/ambry-api/src/test/java/com/github/ambry/rest/RequestPathTest.java
@@ -201,4 +201,21 @@ public class RequestPathTest {
       throw new RuntimeException(e);
     }
   }
+
+  /**
+   * Tests for invalid request path
+   */
+  @Test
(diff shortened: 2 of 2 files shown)


ankagrawal (changes requested):
The code Chan free looks good. Please add unit test and check if this can be quickly added to an existing integration test (to confirm that in the user request path, everything will be able to pass on the correct exception to the user.

ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
```diff
@@ -130,6 +130,8 @@ public static RequestPath parse(String path, Map<String, Object> args, List<Stri
           throw new RestServiceException("Segment index given is not an integer", RestServiceErrorCode.BadRequest);
         }
         //otherwise, nothing to do
+      } catch (ArrayIndexOutOfBoundsException e) {
+        throw new RestServiceException("Bad request path", RestServiceErrorCode.BadRequest);
```
justinlin-linkedin, line 134: let's not do this, let's find the code that use out of bound index and check the length of the array,  throw a RestServiceException.

snalli: The code that used out-of-bound index is L120
```
if (SEGMENT.equals(fields[fields.length - 2])) {
```

What's the difference between the approach in this PR and the one you are suggesting ? The end result in either case is a RestServiceException.

snalli:
> The code Chan free looks good. Please add unit test and check if this can be quickly added to an existing integration test (to confirm that in the user request path, everything will be able to pass on the correct exception to the user.

added test

ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
panchow: Let's lost the root cause exception.  Normally, we if a wrap an other exception, we capture it as the cause.  I notice the other new RestServiceException() doesn't capture the cause exception.  It's a good practice to capture the cause exception.

panchow (approved):

ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
snalli: good point !

justinlin-linkedin: `ArrayIndexOutOfBoundsException` is a RuntimeException. It's a bit like `NullPointerException`. If you already know which logic would yield this exception, then we rely on if-else statement, not try-catch statement. It's faster with if-else.

ankagrawal (approved):

PR #2282, opened 2022-11-01
Make ReplicateBlob compatible with QuotaAwareOperationController

JingQianCloud:
Make ReplicateBlob compatible with QuotaAwareOperationController


justinlin-linkedin (approved):
LGTM

ankagrawal (approved):
I guess it would be best to define a system quota resource, and map everything in the "ReplicateBlobOperation" to this quota resource. But that doesn't need to happen in the current PR. I can take that up as a follow-up action item.

PR #2284, opened 2022-11-02
Compression library accept ByteBuffer.

panchow:
Previous implementation accepts byte[] or Heap.  It relies on JVM byte array pinning to avoid memory copy when calling native compression library.  However, pinning is not reliable and unpredictable.  A better alternative is to use ByteBuffer that can hold either Heap or Direct memory.  This change adds ByteBuffer to API.  After ByteBuffer API code changes are mere, the byte[] will be deleted.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/compression/Compression.java
+++ b/ambry-api/src/main/java/com/github/ambry/compression/Compression.java
@@ -15,6 +15,8 @@ package com.github.ambry.compression;
 
 import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Utils;
+import java.nio.ByteBuffer;
+
 
 /**
  * Compression interface that provides compression and decompression feature.
@@ -25,17 +27,17 @@ import com.github.ambry.utils.Utils;
  * Shared buffer APIs (caller management memory).
  *   Compression example:
  *   {@code
- *     byte[] originalData = ...;
- *     int compressedBufferSize = Compression.getCompressBufferSize(originalData.length);
- *     byte[] compressedBuffer = new byte[compressedBufferSize];
+ *     ByteBuffer originalData = ...;
+ *     int compressedBufferSize = Compression.getCompressBufferSize(originalData.remaining());
+ *     ByteBuffer compressedBuffer = ByteBuffer.allocateDirect(compressedBufferSize);
  *     int compressedSize = Compression.compress(originalData, 0, originalData.length,
- *                                   compressedBuffer, 0, compressedBuffer.length);
+ *                                   compressedBuffer, 0, compressedBuffer.capacity());
  *   }
  *
  *   Decompression example:
  *   {@code
- *     byte[] compressedBuffer = ...;
- *     int originalDataSize = Compression.getDecompressBufferSize(compressedBuffer, 0, compressedBuffer.length);
+ *     ByteBuffer compressedBuffer = ...;
+ *     int originalDataSize = Compression.getDecompressBufferSize(compressedBuffer, compressedBuffer.position(), compressedBuffer.remaining());
  *     byte[] originalData = new byte[originalDataSize];
  *     int decompressedSize = Compression.decompress(compressedBuffer, 0, compressedBuffer.length,
  *                                   originalData, 0, originalData.length);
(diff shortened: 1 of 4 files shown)


ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -222,6 +236,9 @@ public int compress(byte[] sourceBuffer, int sourceBufferOffset, int sourceDataS
     compressedBuffer[offset] = 1;
     offset++;
 
+    compressedBuffer[offset] = (byte) algorithmNameBinary.length;
+    offset++;
+
```
justinlin-linkedin, line 241: should move these two line after line 243 to make sure that we called getAlgorithmNameBinary first.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -306,4 +323,146 @@ public int decompress(byte[] compressedBuffer, int compressedBufferOffset, int c
...
+    // Write version, algorithm name, and source data size.
+    int compressedBufferStartPosition = compressedBuffer.position();
+    int sourceBufferStartPosition = sourceBuffer.position();
+    try {
+      byte[] algorithmNameBinary = getAlgorithmNameBinary();
+      compressedBuffer.put((byte) 1);
+      compressedBuffer.put((byte) algorithmNameBinary.length);
+      compressedBuffer.put(algorithmNameBinary);
+      compressedBuffer.putInt(sourceBuffer.remaining());
```
justinlin-linkedin, line 397: I vaguely remember that Java ByteBuffer's default byte order is big endian. But the method of compress (with byte array) is putting length in little endian order. This seems to be something concerning. Do we have a test for this?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -306,4 +323,146 @@ public int decompress(byte[] compressedBuffer, int compressedBufferOffset, int c
...
+      int sourceBufferLimit = sourceBuffer.limit();
+      int compressedDataLength =
+          compressNative(sourceBuffer, sourceBuffer.position(), sourceBuffer.remaining(), compressedBuffer,
+              compressedBuffer.position(), compressedBuffer.capacity() - compressedBuffer.position());
+
+      // compressNative() may or may not alter the position indexes, so set them just in case.
+      sourceBuffer.position(sourceBufferLimit);  // Advance the sourceBuffer to limit to indicate entire buffer read.
+      compressedBuffer.position(compressedBufferStartPosition + overheadSize + compressedDataLength);
+      return (overheadSize + compressedDataLength);
+    } catch (Exception ex) {
```
justinlin-linkedin, line 410: any other exception rather than CompressionException?

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -25,17 +27,17 @@
  * Shared buffer APIs (caller management memory).
  *   Compression example:
  *   {@code
- *     byte[] originalData = ...;
- *     int compressedBufferSize = Compression.getCompressBufferSize(originalData.length);
- *     byte[] compressedBuffer = new byte[compressedBufferSize];
+ *     ByteBuffer originalData = ...;
```
snalli, line 30: What was the motivation to go from byte array to a ByteBuffer ?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -30,16 +31,22 @@ public abstract class BaseCompression implements Compression {
...
    * Excluding the name chars, the minimal size is 6 bytes.
    */
+  private static final int VERSION_AND_ORIGINAL_SIZE_SIZE = 5;
+  private static final int ALGORITHM_NAME_LENGTH_SIZE = 1;
+
+  // Will be removed.
   private static final int SIZE_OF_VERSION_AND_ORIGINAL_SIZE = 5;
   static final int MINIMUM_OVERHEAD_SIZE = SIZE_OF_VERSION_AND_ORIGINAL_SIZE + 1;
 
+
```
snalli, line 49: intended new line or mistake ?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -108,17 +129,13 @@ private byte[] getAlgorithmNameBinary() {
...
           throw new IllegalArgumentException("algorithmName " + algorithmName + " exceeds the size of " +
                   MAX_ALGORITHM_NAME_LENGTH + " bytes.");
         }
-
-        byte[] binaryWithSize = new byte[binary.length + 1];
-        binaryWithSize[0] = (byte) binary.length;
-        System.arraycopy(binary, 0, binaryWithSize, 1, binary.length);
-
-        algorithmNameBinary = binaryWithSize;
+        algorithmNameBinary = binary;
```
snalli, line 138: return algorithmNameBinary ?

snalli: nvm, its down below.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -226,4 +228,86 @@ default byte[] decompress(byte[] compressedBuffer, int compressedBufferOffset, i
...
+   *
+   * @param sourceBuffer The source uncompressed data to compress.  It cannot be null or empty.
+   * @param allocateDirectBuffer Whether to allocate direct buffer or heap buffer to store compressed data.
+   * @return The compressed buffer.
+   * @throws CompressionException Throws this exception if compression has internal failure.
+   */
+  default ByteBuffer compress(ByteBuffer sourceBuffer, boolean allocateDirectBuffer) throws CompressionException {
+    Utils.checkNotNullOrEmpty(sourceBuffer, "sourceBuffer cannot be null or empty.");
+    int compressedBufferSize = getCompressBufferSize(sourceBuffer.remaining());
+    ByteBuffer compressedBuffer = allocateDirectBuffer ? ByteBuffer.allocateDirect(compressedBufferSize) :
```
snalli, line 273: when do we use allocateDirect ? and when do we don't ?

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -226,4 +228,86 @@ default byte[] decompress(byte[] compressedBuffer, int compressedBufferOffset, i
...
+   *                       at least the size return from getCompressBufferSize() or it may fail due to buffer size.
+   * @return The actual compressed data size in bytes.
+   * @throws CompressionException Throws this exception if compression has internal failure.
+   */
+  int compress(ByteBuffer sourceBuffer, ByteBuffer compressedBuffer) throws CompressionException;
+
+  /**
+   * Compress the buffer specified in {@code sourceBuffer}.  This method allocates either direct memory or heap memory,
+   * based on the allocateDirectBuffer parameter, to store the compressed data.  The output compressed buffer capacity
+   * may be bigger than necessary in case sourceBuffer is incompressible.  The index of sourceBuffer will be updated
```
snalli, line 262: why would the compressed buffer be bigger than the sourceBuffer if its not compressed ?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
snalli, line 241: why change any of this if it will be removed ?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -252,9 +269,9 @@ public int compress(byte[] sourceBuffer, int sourceBufferOffset, int sourceDataS
   private void verifySourceData(byte[] sourceData, int sourceDataOffset, int sourceDataSize) {
     Utils.checkNotNullOrEmpty(sourceData, "sourceData cannot be null");
     Utils.checkValueInRange(sourceDataOffset, 0, sourceData.length - 1,
-      "sourceDataOffset is outside of sourceData.");
+        "sourceDataOffset is outside of sourceData.");
```
snalli, line 272: why change any of this if it will be removed ?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -306,4 +323,146 @@ public int decompress(byte[] compressedBuffer, int compressedBufferOffset, int c
...
+      compressedBuffer.put((byte) algorithmNameBinary.length);
+      compressedBuffer.put(algorithmNameBinary);
+      compressedBuffer.putInt(sourceBuffer.remaining());
+
+      // Apply compression and store the output in the compressed buffer.
+      // Note: compressNative() uses less memory than buffer size and that's why it returns the actual compressed size.
+      int sourceBufferLimit = sourceBuffer.limit();
+      int compressedDataLength =
+          compressNative(sourceBuffer, sourceBuffer.position(), sourceBuffer.remaining(), compressedBuffer,
+              compressedBuffer.position(), compressedBuffer.capacity() - compressedBuffer.position());
```
snalli, line 404: whats the difference between capacity and limit ?

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
```diff
@@ -30,16 +31,22 @@ public abstract class BaseCompression implements Compression {
...
-   * - 4 (original size)
+   * The compressed data structure contains:
+   * - 1 byte version
+   * - 1 byte name size
+   * - N bytes name characters
+   * - 4 bytes original size.
+   * - N bytes compressed data.
    * Excluding the name chars, the minimal size is 6 bytes.
    */
+  private static final int VERSION_AND_ORIGINAL_SIZE_SIZE = 5;
```
snalli, line 42: better, less confusing name VERSION_FIELD_SIZE.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
snalli: nvm, i read the commit description

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
panchow: capacity is always the size of the buffer.  limit() is the reading limit the upper limit of reading.  For example, a 10-bytes buffer will have capacity of 10, but it may only have 3 bytes of data so limit() would be 3.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
panchow: IntelliJ auto-styling made the change.  It will be removed.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
panchow: This change is necessary to make keep the current unit tests pass.  In previous implementation, the getAlgorithmNameBinary() returns both the size and the name.  I changed it to better match the method name to return just the name binary (no size).  That's why this change is necessary to keep existing unit tests pass.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
panchow: This is algorithm specific thing.   For example, if compressing 1-byte, the compressed output is likely be bigger than 1 byte due to structural overhead and internal compression dictionary, depending algorithm.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
panchow: It would be more obvious in later PR because I document it in later PR.  Basically, if compression output is sent to encryption, the output should be heap because encryption takes a heap (not direct memory).  If compression output is sent to Netty, it should be direct memory because Netty prefers direct memory.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
panchow: Thank you.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
panchow: ByteBuffer is more generic than byte[].  Byte[] is always Heap while ByteBuffer supports both Heap and Direct memory.  The output of Netty (incoming request) is Direct memory.  If compression takes Heap, Direct memory must be copied to Heap first.  Similarly, Netty prefers Direct memory when sending responses.  If Heap is sent to Netty, Netty makes a copy of the buffer before sending.  Also, when calling native code (compression in case this), native code only works with direct memory.  If we call with Heap like byte[], JVM either pin or copy the byte[] to direct memory.  Pinning is unpredictable and uncontrollable by ambry, so if ambry passes direct memory, we know it does not copy.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
panchow: Yes, to comply with the documentation "if compress() failed, the indexes will not change".  The indexes advances only if succeeded.  The goal is to avoid leaving bytebuffer in an unpredictable position when exception is thrown.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
panchow: ByteBuffer default order is always big-endian order no matter what OS it is running in (OS independent).    Only calling order(newOrder) could alter that instance of ByteBuffer.  None of our code calls that.  You were right that the old implementation used little-endian by taking encoding in own hand.  To simplify the code, I decided to use big-endian which is ByteBuffer default and Java JVM default.

ambry-commons/src/main/java/com/github/ambry/compression/BaseCompression.java
panchow: You are right.  Let me do that.

ambry-api/src/main/java/com/github/ambry/compression/Compression.java
```diff
@@ -25,17 +27,17 @@
  * Shared buffer APIs (caller management memory).
  *   Compression example:
  *   {@code
- *     byte[] originalData = ...;
- *     int compressedBufferSize = Compression.getCompressBufferSize(originalData.length);
- *     byte[] compressedBuffer = new byte[compressedBufferSize];
+ *     ByteBuffer originalData = ...;
+ *     int compressedBufferSize = Compression.getCompressBufferSize(originalData.remaining());
+ *     ByteBuffer compressedBuffer = ByteBuffer.allocateDirect(compressedBufferSize);
  *     int compressedSize = Compression.compress(originalData, 0, originalData.length,
```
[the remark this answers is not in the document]
panchow: Thank you.  I will update this java doc.

justinlin-linkedin (approved):
LGTM

mikeyang923 (approved):

PR #2285, opened 2022-11-02
Only create one metric for IndexSegmentFilesDirectMemoryUsage

justinlin-linkedin:
Before this PR, we are using store id prefix when creating metric for index segment files' direct memory usage metric. This would end up creating multiple metrics with the same value.

This PR fixes that.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
@@ -277,17 +277,9 @@ public class StoreMetrics {
     }
     if (enableStoreIndexDirectMemoryUsageMetric) {
       indexes.putIfAbsent(storeId, index);
-      Gauge<Long> indexDirectMemoryUsage = () -> {
-        long start = System.nanoTime();
-        long usage = indexes.values().stream().mapToLong(PersistentIndex::getDirectMemoryUsage).sum();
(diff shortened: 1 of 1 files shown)


ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -277,17 +277,9 @@ void initializeIndexGauges(String storeId, final PersistentIndex index, final lo
...
-          logger.trace("Time to get direct memory usage from persistent index is: {} nano seconds",
-              System.nanoTime() - start);
-        }
-        return usage;
-      };
-      registry.gauge(MetricRegistry.name(BlobStore.class, prefix + "IndexDirectMemoryUsage"),
-          () -> indexDirectMemoryUsage);
+      Gauge<Long> indexDirectMemoryUsage =
+          () -> indexes.values().stream().mapToLong(PersistentIndex::getDirectMemoryUsage).sum();
+      registry.gauge(MetricRegistry.name(BlobStore.class, "IndexDirectMemoryUsage"), () -> indexDirectMemoryUsage);
```
SophieGuo410, line 282: QQ: Does that mean for all blobstores, they will have same IndexDirectMemoryUsage? What if some blobStore has higher usage?

SophieGuo410 (approved):

PR #2293, opened 2022-11-16
ReplicateBlob: call keyConverter before store.findKey

JingQianCloud:


ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -740,10 +740,10 @@ private boolean localStoreHasTheKey(ReplicateBlobRequest replicateBlobRequest) {
 
     // Currently we don't enable the write repair. As long as the local store has the Blob, return success immediately.
     // check if local store has the key already
-    // we don't use the keyConverter to convert the key here since it's a valid Blob ID coming from the frontend.
+    List<StoreKey> convertedStoreKeys = getConvertedStoreKeys(Collections.singletonList(blobId));
     Store store = storeManager.getStore(blobId.getPartition());
     try {
-      store.findKey(blobId);
+      store.findKey(convertedStoreKeys.get(0));
```
justinlin-linkedin, line 746: we can do ```
StoreKey convertedKey = getConvertedStoreKeys(Collections.singletonList(blobId)).get(0);
Store store = storeManager.getStore(convertedKey.getPartition());
```

JingQianCloud: It's better to do 
     StoreKey convertedKey = getConvertedStoreKeys(Collections.singletonList(blobId)).get(0);
For the convertedKey.getPartition, looks like we don't have getPartition for StoreKey?

JingQianCloud: I submit the second revision.

justinlin-linkedin: Doesn't seems like StoreKeyConverter would guarantee the partition would stay then same before and after conversion. We better use convertedKey's partition here. We can type case StoreKey to BlobId.

JingQianCloud: I see. Thank you for bringing it up.
Changed.

SophieGuo410 (approved):

justinlin-linkedin (approved):

PR #2295, opened 2022-11-29
ReplicateBlob: fixed the KeyConvert Cache miss.

JingQianCloud:
AmbryRequests was shared among all the requestHandlers. Now assign one AmbryRequests for each handler.
Otherwise, the keyConvert cache may be reset by other thread accidently.


ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -95,7 +95,7 @@ public class AmbryServerRequests extends AmbryRequests {
   private final DiskManagerConfig diskManagerConfig;
   private final StatsManager statsManager;
   private final ClusterParticipant clusterParticipant;
-  private final ConcurrentHashMap<RequestOrResponseType, Set<PartitionId>> requestsDisableInfo =
+  private static final ConcurrentHashMap<RequestOrResponseType, Set<PartitionId>> requestsDisableInfo =
```
mikeyang923, line 98: Why do we need to change this to 'static'?

JingQianCloud: Before we use the single AmbryServerRequests object, so whenever there is a adminCommand to change the requestsDisableInfo, it applies to this single object. And all the handlers are aware of that.
After we change to use one object for each handler, all the handler should be aware of it as well. If we don't use static, only one thread is aware of the requestsDisableInfo.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -134,15 +134,20 @@ public AmbryRequests(StoreManager storeManager, RequestResponseChannel requestRe
...
-      try {
-        StoreKeyConverter keyConverter = storeKeyConverterFactory.getStoreKeyConverter();
-        transformer = Utils.getObj(serverConfig.serverMessageTransformer, storeKeyFactory, keyConverter);
-      } catch (Exception e) {
-        logger.error("Failed to create transformer", e);
+    /* All the request handlers share one single AmbryRequests object.
+     * But the StoreKeyConverter of the Transformer has a cache which shouldn't be shared among handlers.
+     * So use ThreadLocal transformer. */
+    this.transformer = ThreadLocal.withInitial(() -> {
+      if (serverConfig != null) {
```
justinlin-linkedin, line 141: can we just require serverConfig to be not null?

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -134,15 +134,20 @@ public AmbryRequests(StoreManager storeManager, RequestResponseChannel requestRe
...
-    if (serverConfig != null) {
-      try {
-        StoreKeyConverter keyConverter = storeKeyConverterFactory.getStoreKeyConverter();
-        transformer = Utils.getObj(serverConfig.serverMessageTransformer, storeKeyFactory, keyConverter);
-      } catch (Exception e) {
-        logger.error("Failed to create transformer", e);
+    /* All the request handlers share one single AmbryRequests object.
+     * But the StoreKeyConverter of the Transformer has a cache which shouldn't be shared among handlers.
+     * So use ThreadLocal transformer. */
+    this.transformer = ThreadLocal.withInitial(() -> {
```
justinlin-linkedin, line 140: This works too, but I think the storeKeyConverter should be the thread local variable, not the transformer. Because the storekeyconverter is also used in other handlexxx methods and it's the one that has a local state that requires synchronization.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: It's very possible there is some legacy code or test code doesn't provide it. So I added this check.
Our production code always provide it.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
JingQianCloud: In the AmbryRequests.java, we don't explicitly use keyConverter. 
We only provide the Transformer to the MessageSievingInputStream.transferInputStream. And MessageSievingInputStream.transferInputStream calls transformer.warmup only. Along the path, we don't touch keyConverter. 
Only the transformer LI implementation itself will use the keyConverter.


storeKeyconverter is not used in any other handleXXX.

JingQianCloud: You are right.
Make keyConverter as well to avoid we create keyConverter every time when we handle one request.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -130,19 +130,32 @@ public AmbryRequests(StoreManager storeManager, RequestResponseChannel requestRe
...
       }
-    }
+      return null;
+    });
+
+    /* All the request handlers share one single AmbryRequests object.
+     * But the StoreKeyConverter of the Transformer has a cache which shouldn't be shared among handlers.
+     * So use ThreadLocal transformer. */
+    this.transformer = ThreadLocal.withInitial(() -> {
+      if (serverConfig != null && keyConverter != null && keyConverter.get() != null) {
```
justinlin-linkedin, line 150: we don't have to check if the keyConverter is null, we know it's not going to be null.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -753,7 +766,7 @@ private boolean localStoreHasTheKey(ReplicateBlobRequest replicateBlobRequest) t
 
   @Override
   public void handleReplicateBlobRequest(NetworkRequest request) throws IOException, InterruptedException {
-    if (connectionPool == null || transformer == null) {
+    if (connectionPool == null || transformer == null || transformer.get() == null) {
```
justinlin-linkedin, line 769: same here, we don't have to check if the transformer is null.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -1179,8 +1192,8 @@ protected ServerErrorCode validateRequest(PartitionId partition, RequestOrRespon
    * @return A list of converted storeKeys.
    */
   protected List<StoreKey> getConvertedStoreKeys(List<? extends StoreKey> storeKeys) throws Exception {
-    storeKeyConverterFactory.getStoreKeyConverter().dropCache();
-    Map<StoreKey, StoreKey> conversionMap = storeKeyConverterFactory.getStoreKeyConverter().convert(storeKeys);
+    keyConverter.get().dropCache();
```
justinlin-linkedin, line 1195: thanks for fixing this.

justinlin-linkedin (approved):
lgtm, just some minor comments

justinlin-linkedin (approved):

PR #2296, opened 2022-11-30
Support building user metadata for url encoded metadata

SophieGuo410:
When customer send out some metadata which can't be encoded by ASCII, we would set the key with "x-ambry-enc-um-" prefix, and encode value by UrlEncode. In that way,  since url has to have ASCII, http encode and decode won't mess up the value. Since that, when we build the metadata, we should decode by calling UrlDecode so internally we can get the correct userMetadataMap.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
@@ -534,8 +534,7 @@ public class RestUtils {
    * @return the user metadata extracted from arguments.
    * @throws RestServiceException if usermetadata arguments have null values.
    */
-  public static byte[] buildUserMetadata(Map<String, Object> args)
-      throws RestServiceException, UnsupportedEncodingException {
+  public static byte[] buildUserMetadata(Map<String, Object> args) throws RestServiceException {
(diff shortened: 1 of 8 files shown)


mikeyang923 (approved):
lgtm

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -532,7 +534,8 @@ public static long getTtlFromRequestHeader(Map<String, Object> args) throws Rest
    * @return the user metadata extracted from arguments.
    * @throws RestServiceException if usermetadata arguments have null values.
    */
-  public static byte[] buildUserMetadata(Map<String, Object> args) throws RestServiceException {
+  public static byte[] buildUserMetadata(Map<String, Object> args)
+      throws RestServiceException, UnsupportedEncodingException {
```
justinlin-linkedin, line 538: can we convert the UnsupportedEncodingException to RestServiceException so we don't have to change the method signature?

SophieGuo410: Sure, updated.

justinlin-linkedin (approved):

PR #2297, opened 2022-11-30
Replication Optimization: Treating deleted put record's size same as delete 

justinlin-linkedin:
This is an optimization for replication. 
In replication, when replicator sends a ReplicaMetadataRequest to remote server, remote server would scan through the index segment files. It will stop when the number of scanned bytes are larger than the max number of byte in the ReplicaMetadataRequest. And the max number of bytes is usually 4MB (one PutBlob size).

In this case, when the remote server sees a Put record in the index segment file, it will stop scanning right away. However, if this Put record is already deleted (there is a Delete record after this Put), the size in the MessageInfo would be changed to size of Delete.  When the replicator sees this id, it knows that this id is deleted in the remote server so it does nothing. This is a terrible waste of roundtrip since we are spending one round trip to deal with one blob id. 

In this PR, we would keep scanning if this put Record is already deleted. In this way, we can return more ids back to the replicator in one request. We only do this for IndexBased replication token.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -521,7 +521,7 @@ public class StoreConfig {
   /**
    * If true, then in findEntriesSince method, use Delete entry's size for a deleted Put entry. This is useful for replication.
    * In replication, when handling ReplicaMetadataRequest, server would scan through index segments. One Put entry could be
-   * as big as 4MB, but is this Put entry is already deleted, then in the requester server side it will do nothing. This is bad
+   * as big as 4MB, but if this Put entry is already deleted, then in the requester server side it will do nothing. This is bad
    * because we are using one roundtrip to bypass on blob id.
    * Set this to true would help mitigate this issue. We know the put's final state is delete and the requester server won't
    * do anything, then we just move on to scan more entries.
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -518,6 +518,18 @@ public class StoreConfig {
   public final boolean storeUseBlobFormatV3;
   public static final String storeUseBlobFormatV3Name = "store.use.blob.format.v3";
 
+  /**
+   * If true, then in findEntriesSince method, use Delete entry's size for a deleted Put entry. This is useful for replication.
+   * In replication, when handling ReplicaMetadataRequest, server would scan through index segments. One Put entry could be
+   * as big as 4MB, but is this Put entry is already deleted, then in the requester server side it will do nothing. This is bad
+   * because we are using one roundtrip to bypass on blob id.
+   * Set this to true would help mitigate this issue. We know the put's final state is delete and the requester server won't
+   * do anything, then we just move on to scan more entries.
+   */
+  @Config(storeDeletedPutAsDeleteInFindEntriesName)
+  public final boolean storeDeletedPutAsDeleteInFindEntries;
+  public static final String storeDeletedPutAsDeleteInFindEntriesName = "store.deleted.put.as.delete.in.find.entries";
+
   public StoreConfig(VerifiableProperties verifiableProperties) {
 
     storeKeyFactory = verifiableProperties.getString("store.key.factory", "com.github.ambry.commons.BlobIdFactory");
@@ -646,5 +658,7 @@ public class StoreConfig {
     }
 
     storeUseBlobFormatV3 = verifiableProperties.getBoolean(storeUseBlobFormatV3Name, false);
+    storeDeletedPutAsDeleteInFindEntries =
+        verifiableProperties.getBoolean(storeDeletedPutAsDeleteInFindEntriesName, false);
   }
 }
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/DeleteMessageFormatInputStream.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/DeleteMessageFormatInputStream.java
@@ -69,7 +69,7 @@ public class DeleteMessageFormatInputStream extends MessageFormatInputStream {
   }
 
   /**
-   * Return the {@link DeleteMessageFormatInputStream} size for the given {@code key}.
+   * Only used in test. Return the {@link DeleteMessageFormatInputStream} size for the given {@code key}.
    * @param key The {@link StoreKey}
    * @return The size of {@link DeleteMessageFormatInputStream}
    */
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/DeleteMessageFormatInputStream.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/DeleteMessageFormatInputStream.java
@@ -31,18 +31,6 @@ import java.nio.ByteBuffer;
  *
  */
 public class DeleteMessageFormatInputStream extends MessageFormatInputStream {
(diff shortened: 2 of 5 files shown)


SophieGuo410 (commented):
Let some minor comments.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/DeleteMessageFormatInputStream.java
```diff
@@ -67,4 +67,19 @@ public DeleteMessageFormatInputStream(StoreKey key, short accountId, short conta
     messageLength = buffer.capacity();
     buffer.flip();
   }
+
+  /**
+   * Return the {@link DeleteMessageFormatInputStream} size for the given {@code key}.
+   * @param key The {@link StoreKey}
+   * @return The size of {@link DeleteMessageFormatInputStream}
+   */
+  public static long getDeleteMessageFormatInputStreamSize(StoreKey key) {
```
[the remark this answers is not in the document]
justinlin-linkedin: This one is only used in test and the exception is caught in the method. I think it's fine to not use this method.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1857,6 +1870,31 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
...
+  /**
+   * Return true to stop scanning the index segment files. We stop scanning when
+   * 1. storeDeletedPutAsDeleteInFindEntries is false or
+   * 2. No message entries was found in the last scan or
+   * 3. new Token is not IndexBased anymore or
+   * 4. the total size of returned message infos are already larger than the max totals zie of entries.
+   *
+   * If we are comparing the size of returned message infos, then we are treating a deleted Put Message
+   * as a Delete message.
+   * @param messageEntries
```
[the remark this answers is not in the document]
justinlin-linkedin: updated

ambry-store/src/test/java/com/github/ambry/store/IndexTest.java
```diff
@@ -1250,6 +1251,94 @@ public void findEntriesSinceTest() throws StoreException {
...
+  public void findEntriesSinceDeletedPutAsDeleteTest() throws StoreException {
+    // first we have to change PUT and DELETE size, they are not the same as the real size.
+    assumeTrue(isLogSegmented && persistentIndexVersion == PersistentIndex.VERSION_4 && !addUndeletes);
+    state.closeAndClearIndex();
+
+    long oldPutSize = PUT_RECORD_SIZE;
+    long oldDeleteSize = DELETE_RECORD_SIZE;
+    try {
+      MockId id = state.getUniqueId();
+      long deleteSize = DeleteMessageFormatInputStream.getDeleteMessageFormatInputStreamSize(id);
```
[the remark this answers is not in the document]
justinlin-linkedin: updated

SophieGuo410 (approved):

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -518,6 +518,18 @@ public class StoreConfig {
   public final boolean storeUseBlobFormatV3;
   public static final String storeUseBlobFormatV3Name = "store.use.blob.format.v3";
 
+  /**
+   * If true, then in findEntriesSince method, use Delete entry's size for a deleted Put entry. This is useful for replication.
+   * In replication, when handling ReplicaMetadataRequest, server would scan through index segments. One Put entry could be
+   * as big as 4MB, but is this Put entry is already deleted, then in the requester server side it will do nothing. This is bad
```
JingQianCloud, line 524: nit: "is this Put entry" -> "if this Put entry"

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1825,25 +1825,38 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
...
 
-        startTimeInMs = time.milliseconds();
-        eliminateDuplicates(messageEntries);
-        logger.trace("Segment based token, Time used to eliminate duplicates: {}",
-            (time.milliseconds() - startTimeInMs));
+          startTimeInMs = time.milliseconds();
+          eliminateDuplicates(messageEntries);
+          logger.trace("Segment based token, Time used to eliminate duplicates: {}",
+              (time.milliseconds() - startTimeInMs));
+          sizeSum = messageEntries.stream().mapToLong(MessageInfo::getSize).sum();
```
JingQianCloud, line 1851: I don't see the code to determine the Blob is deleted or not. Or is that the MessageInfo::getSize different if the PutBlob is deleted?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -1857,6 +1870,32 @@ FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries) throws St
     }
   }
 
+  /**
+   * Return true to stop scanning the index segment files. We stop scanning when
+   * 1. storeDeletedPutAsDeleteInFindEntries is false or
+   * 2. No message entries was found in the last scan or
+   * 3. new Token is not IndexBased anymore or
+   * 4. the total size of returned message infos are already larger than the max total size of entries.
```
JingQianCloud, line 1878: Does findEntriesFromSegmentStartOffset faithfully make sure the entry size is less or equal than required size?
If yes, so for the last loop, if the required size is very small like 10 bytes and findEntriesFromSegmentStartOffset will return nothing. But we are good since the shouldStopScanning will stop if messageEntries.isEmpty().
In this case, messageInfoSizeSum only can be equal to maxTotalSizeOfEntries but never exceed it, right?

ambry-messageformat/src/main/java/com/github/ambry/messageformat/DeleteMessageFormatInputStream.java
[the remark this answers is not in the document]
JingQianCloud: If so, probably add some comment for it to avoid people accidentally use it in production code. It doesn't cover all the Format versions.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: In findEntriesSince, there is a method call `updateStateForMessages`, it will update all the state for the put if the put is already deleted.

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
justinlin-linkedin: not really, the sum of message info size can be greater than the maxTotalSize. The way it works it 
```
for each indexvalue:
    if currentSize < max size: 
        add index value to the list
    else:
        break
    currentSize += index value.size
```
as you can see if the max size is 10, but the first index value's size is 100, we still add the first index value to the list, but we will not add the second one.

justinlin-linkedin: not just double. Imagine you set the max total size to be 100 but one PutBlob is 4MB, we would still send this PutBlob back to requester.

JingQianCloud: Yes, this is what I mean. Currently maxTotalSizeOfEntries is 4MB. We may send back 8MB entries.

JingQianCloud (approved):

PR #2298, opened 2022-12-02
Debug resolving sealed-unsealed state in ZK

snalli:


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -447,8 +453,7 @@ private void checkCapacityAndUpdateReplicaStatusDelegate() {
         }
       }
       // During startup, we also need to reconcile the replica state from both ZK clusters.
-      if (!started && replicaStatusDelegates.size() > 1 && thresholdBytesLow < index.getLogUsedCapacity()
-          && index.getLogUsedCapacity() <= thresholdBytesHigh) {
+      if (!started && replicaStatusDelegates.size() > 1) {
```
SophieGuo410, line 456: Did you find any case where the current logUsedCapacity < thresholdBytesLow(readwrite threshold) but the state is sealed and need to be unseal during startup? or any case logUsedCapacity > thresholdBytesHigh(readonly threshold) but state should change to unseal state during startup? Is this only used for zk migration since you want to keep the partition state same? If so, I would suggest you could use config to enable and disable this method once the migration finished. Since for each put and compaction we will tuning this state anyway.

snalli: This code is executed only at startup. There is a node in ei4 atm that is not reconciling the states in zookeeper and I suspect its because of this threshold boolean. There is no easy way to test it unless we merge it. I am looking at some hacky ways of testing atm.

panchow (approved):

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
JingQianCloud: Since it's for debugging purpose, can you not change the if condition to avoid logic change?
Can we only add debug message?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -196,7 +196,13 @@ public BlobStore(ReplicaId replicaId, StoreConfig config, ScheduledExecutorServi
     this.recovery = recovery;
     this.hardDelete = hardDelete;
     this.accountService = accountService;
-    this.replicaStatusDelegates = config.storeReplicaStatusDelegateEnable ? replicaStatusDelegates : null;
+    if (config.storeReplicaStatusDelegateEnable && replicaStatusDelegates != null) {
```
JingQianCloud, line 199: Double check it's only print once in the bootup time.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -207,7 +213,7 @@ public BlobStore(ReplicaId replicaId, StoreConfig config, ScheduledExecutorServi
     errorCount = new AtomicInteger(0);
     currentState = ReplicaState.OFFLINE;
     remoteTokenTracker = replicaId == null ? null : new RemoteTokenTracker(replicaId);
-    logger.debug(
+    logger.info(
```
JingQianCloud, line 216: Double check if it only print once during the bootup time.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
snalli: printed once per partition but only during server startup

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
snalli: printed once per partition but only during server startup

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
snalli: i want the code to enter the if condition, i suspect its not entering this, hence i changed the logic

snalli: and don't worry, this code only executes when there are two ZKs and we don't have two ZKs in corp or prod.

JingQianCloud (approved):
LGTM.
Please clean up it after the issue is verified.

snalli:
I am discarding it. I debugged the issue. The code never reaches this point.

PR #2301, opened 2022-12-12
Add additional queue choices and request dropping at Ambry Server

Arun-LinkedIn:
This PR is mostly copied from @cgtz work in this [Pull request](https://github.com/linkedin/ambry/commit/d19707ed62e24da08ca5e01b19aa94e79a3bdef5) which adds following changes at Ambry server. 

- Add support for dropping requests that have exceeded a queueing timeout.
- Add an implementation of the Adaptive-LIFO co-del queue described in https://queue.acm.org/detail.cfm?id=2839461 that supports request dropping and based on an HBase java implementation. However, keeping this implementation disabled until we do few more tests to confirm on default config parameters for co-del algorithm.


ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -80,30 +102,65 @@ public void closeConnection(NetworkRequest originalRequest) throws InterruptedEx
...
+  public NetworkRequestBundle receiveRequest() throws InterruptedException {
+
+    NetworkRequestBundle networkRequestBundle;
+    NetworkRequest requestToServe;
+    Collection<NetworkRequest> requestsToDrop;
+    do {
+      networkRequestBundle = networkRequestQueue.take();
+      NetworkRequest unqueuedRequest;
+      while ((unqueuedRequest = unqueuedRequests.poll()) != null) {
+        networkRequestBundle.getRequestsToDrop().add(unqueuedRequest);
```
litingulfs, line 117: It seems counter-intuitive that you're immediately adding new requests to the queue for requests that may get dropped. Why?

litingulfs (changes requested):
This is very nice, I love the idea of making Ambry more resilient by adaptively dropping requests!

I'm new to this part of the Ambry code, so I might have misunderstood what is going on here. However, it seems to me that replication related requests may be dropped as well as client requests - is that correct?

I didn't see any metrics related to dropping requests or switching between queue policies. Unless I missed something here, would it be possible to add metrics for this? I suspect that we may want to alert if we drop a large number of requests for a significant duration.

Finally, is there any way in which we could roll out this implementation based on a lix? Even if it is just the dropping of requests that is gated by the lix?

ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
```diff
@@ -25,77 +25,88 @@ public interface RequestAPI {
   /**
    * Handle a request.
    * @param request The request to handle.
+   * @param shouldDrop indicates if request should be dropped.
    * @throws InterruptedException if request processing is interrupted.
    */
-  void handleRequests(NetworkRequest request) throws InterruptedException;
+  void handleRequests(NetworkRequest request, boolean shouldDrop) throws InterruptedException;
```
justinlin-linkedin, line 31: These interfaces look weird to me for two reasons.
1. Dropping request should be done in netty io thread, it shouldn't invoke request handling requests.
2. Even if we want to do this in request handling request, can we have a new method like handleDropppedRequest?

JingQianCloud: +1
Not sure how easy or hard to reject the requests in netty io thread since it doesn't deal with AmbryRequest before. But if it's doable, we will be able to reject these request at the earliest convenience. It's more efficient.

And with either 1 or 2, it can avoid the massive interface change.

ambry-network/src/main/java/com/github/ambry/network/AdaptiveLifoCoDelNetworkRequestQueue.java
```diff
@@ -0,0 +1,169 @@
...
+
+
+/**
+ * Adaptive LIFO blocking queue utilizing CoDel algorithm to prevent queue overloading. This queue is based on
+ * (and borrows most of its code from) {@code AdaptiveLifoCoDelCallQueue} in HBase.
+ *
+ * @see <a href="http://queue.acm.org/detail.cfm?id=2839461">Fail at Scale paper</a>
+ * @see <a href="https://github.com/apache/hbase/blob/master/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java">HBase implementation</a>
+ */
+public class AdaptiveLifoCoDelNetworkRequestQueue implements NetworkRequestQueue {
```
JingQianCloud, line 31: Not sure how well the parameters are tuned for the adaptive queue and if it's suitable for our case.
To my own opinion, probably we can just start with some simple LIFO. But it's just my 2 cents.

ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
snalli: Would the netty thread print out any logs indicating that requests are being rejected ? or we might be in a limbo.

tbingol: I think this cannot be achieved in the netty io thread since the decision to drop a request cannot be reached during enqueuing. Requests can decay while they are in the queue.

OTOH, +1 on having a generic request drop routing rather than delaying it to specific request handlers, e.g., handlePutRequest.

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -80,30 +102,65 @@ public void closeConnection(NetworkRequest originalRequest) throws InterruptedEx
...
+        if (!consumeStream(requestToServe)) {
+          requestToServe = null;
+        }
+      }
+
+      Set<NetworkRequest> errorRequests = new HashSet<>();
+      for (NetworkRequest requestToDrop : requestsToDrop) {
+        http2ServerMetrics.requestQueuingTime.update(System.currentTimeMillis() - requestToDrop.getStartTimeInMs());
+        if (!consumeStream(requestToDrop)) {
+          errorRequests.add(requestToDrop);
```
tbingol, line 133: Log the errors requests?

ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
Arun-LinkedIn: Sure. Made changes to have a separate method in RequestAPI to drop requests instead of modifying all `handlexxxRequest` methods. Also, moved metrics also to a common place so that they can be reused for both 'handled' and 'dropped' requests.

Even though it is ideal to send responses 'unqueued' requests immediately in Netty Thread, I was thinking it might be better to have same request-response flow path (i.e. requests queued by Netty thread and dequeued, processed and handed over back to netty outbound queues by request handler thread) for both 'unqueued' and 'delayed' requests. It was also easier to have response handling code (like updating metrics, etc) in one place in that way.

ambry-network/src/main/java/com/github/ambry/network/AdaptiveLifoCoDelNetworkRequestQueue.java
Arun-LinkedIn: Sure. Yeah, true. I was also thinking if we can start with simple Fifo queue first and later do more tests on adaptive queue to tune the config parameters.

ambry-network/src/main/java/com/github/ambry/network/AdaptiveLifoCoDelNetworkRequestQueue.java
```diff
@@ -0,0 +1,169 @@
...
+   * @throws InterruptedException if interrupted while waiting
+   */
+  @Override
+  public NetworkRequestBundle take() throws InterruptedException {
+    NetworkRequest requestToServe = null;
+    List<NetworkRequest> requestsToDrop = new ArrayList<>();
+    while (true) {
+      NetworkRequest nextRequest;
+      if (useLifoMode()) {
+        nextRequest = deque.pollLast();
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Added log statement to print when mode changes in `useLifoMode` method. For metrics, actually added a set of metrics at higher level to track dropped request rate for various types of requests. Since changing of queue mode is an implementation detail, wondering if metric may not be needed here. What do you say?

ambry-network/src/main/java/com/github/ambry/network/AdaptiveLifoCoDelNetworkRequestQueue.java
```diff
@@ -0,0 +1,169 @@
...
+  public String toString() {
+    return deque.toString();
+  }
+
+  /**
+   * Based on queue size, decide whether the queue should operate in LIFO or FIFO mode.
+   * @return {@code true} if the queue should operate in LIFO mode.
+   */
+  private boolean useLifoMode() {
+    double loadFactor = (double) deque.size() / maxCapacity;
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Done

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -30,18 +36,34 @@
...
+    }
     this.http2ServerMetrics = http2ServerMetrics;
   }
 
   /** Send a request to be handled, potentially blocking until there is room in the queue for the request */
   @Override
   public void sendRequest(NetworkRequest request) throws InterruptedException {
-    requestQueue.put(request);
+    if (!networkRequestQueue.offer(request)) {
+      logger.debug("Request queue is full, dropping incoming request: {}", request);
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Updated as warning message.

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
Arun-LinkedIn: Sure, done.

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
```diff
@@ -119,6 +122,35 @@ public class NetworkConfig {
...
+  @Config("adaptive.lifo.queue.codel.target.delay.ms")
+  @Default("100")
+  public final int adaptiveLifoQueueCodelTargetDelayMs;
+
+  /**
+   * The threshold over which we switch the request queue from FIFO to LIFO mode.
+   */
+  @Config("adaptive.lifo.queue.threshold")
+  @Default("0.7")
+  public final double adaptiveLifoQueueThreshold;
```
mikeyang923, line 152: Is those `@Default `actually in-use/needed? Since i see we explicitly set the defaultValue in constructor like below:
` adaptiveLifoQueueThreshold = verifiableProperties.getDoubleInRange("adaptive.queue.lifo.threshold", 0.7, 0.0, 1.0);`

ambry-network/src/main/java/com/github/ambry/network/FifoNetworkRequestQueue.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  public NetworkRequestBundle take() throws InterruptedException {
+    NetworkRequest requestToServe = null;
+    List<NetworkRequest> requestsToDrop = new ArrayList<>();
+    NetworkRequest nextRequest;
+    while ((nextRequest = queue.poll()) != null) {
+      if (needToDrop(nextRequest)) {
+        logger.warn("Request timed out waiting in queue, dropping it: {}", nextRequest);
+        requestsToDrop.add(nextRequest);
```
JingQianCloud, line 50: When "needToDrop" is true, shall we also need return a request to server?
Looks like we either return one request to serve or one request to drop.

But for the definition of NetworkRequestBundle, it can have 1 request to serve and multiple requests to drop.

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -30,18 +36,34 @@
...
+      default:
+        throw new IllegalArgumentException("Queue type not supported by channel: " + config.requestQueueType);
+    }
     this.http2ServerMetrics = http2ServerMetrics;
   }
 
   /** Send a request to be handled, potentially blocking until there is room in the queue for the request */
   @Override
   public void sendRequest(NetworkRequest request) throws InterruptedException {
-    requestQueue.put(request);
```
JingQianCloud, line 44: Looks like we have two places to decide drop request.
1. NetworkRequestQueue has a blocking queue. It may drop request as well.
2. RequestResponseChannel has a unqueuedRequests queue if NetworkRequestQueue blocking queue is full. 

Is that possible to let RequestResponseChannel manage all the requests and decide the drop policy? Don't split it to two places. NetworkRequestQueue can have its own queue instead of the blocking queue.

ambry-network/src/main/java/com/github/ambry/network/SocketRequestResponseChannel.java
```diff
@@ -145,7 +162,10 @@ public SocketRequestResponseChannel(int numProcessors, int queueSize) {
   /** Send a request to be handled, potentially blocking until there is room in the queue for the request */
   @Override
   public void sendRequest(NetworkRequest request) throws InterruptedException {
-    requestQueue.put(request);
+    if (!networkRequestQueue.offer(request)) {
```
JingQianCloud, line 165: Same here. 
If NetworkRequestQueue hold all the requests, we don't need unqueuedRequest queue for each RequestResponseChannel implementation.

Arun-LinkedIn:
Just wanted to add this PR has 3 changes. 

1. Adding fifo and adaptive queues to support request dropping as originally described in description.
2. Added metrics to track drop rate for various types of requests like put, get, etc. Also, moved the updating of metrics (like put request rate, get request rate, etc) to a separate method since it was turning out to a lot of duplicate code
3. Added a background thread (request-dropper thread) to do function of only dropping requests to avoid memory build since request-handler threads may be busy and cannot catch up with drop rate.

Please let me know if you want to split 2 and 3 points into separate PRs.

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
Arun-LinkedIn: Will check once. I copied from other configs but like you said`@Default` annotation may not be in use.

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
Arun-LinkedIn: Yeah, true. Like you said, it is probably good to have `unqueuedRequests` queue as well in `NetworkRequestQueue`. But, with the latest commit, since I am fetching unqueued requests separately from `RequestResponseChannel`, I thought I will leave the `unqueuedRequests` queue in RequestResponseChannel. Please let me know what you think.

ambry-network/src/main/java/com/github/ambry/network/FifoNetworkRequestQueue.java
Arun-LinkedIn: Hmm, in this case, if we find that the request needs to be dropped, we are not actually breaking from loop and are continuing to iterate until we find a valid request or the queue is empty. So, the `NetworkRequestBundle` could return 1  request to serve and multiple requests to drop.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -196,6 +196,77 @@ public void handleRequests(NetworkRequest networkRequest) throws InterruptedExce
...
+  public void dropRequest(NetworkRequest request) throws InterruptedException {
+    try {
+      InputStream is = request.getInputStream();
+      DataInputStream dis = is instanceof DataInputStream ? (DataInputStream) is : new DataInputStream(is);
+      RequestOrResponseType type = RequestOrResponseType.values()[dis.readShort()];
+      Response response;
+      long requestQueueTime = SystemTime.getInstance().milliseconds() - request.getStartTimeInMs();
+      RequestMetricsUpdater metricsUpdater = new RequestMetricsUpdater(requestQueueTime, 0, 0, 0, true);
+      switch (type) {
+        case PutRequest:
```
justinlin-linkedin, line 209: nit: we don't have to deserialize the entire put request, we can just get the correlation id and the client id. we can probably do this in another PR.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -1246,4 +1253,322 @@ private void applyDelete(MessageInfo messageInfo, ReplicateBlobRequest replicate
       }
     }
   }
+
+  /**
+   * Used to update common metrics for the incoming request.
+   */
+  public class RequestMetricsUpdater implements RequestVisitor {
```
justinlin-linkedin, line 1260: I don't think we need to use visitor pattern given that we already have lots of code branches that do a switch statement based on RequestType. It seems like the handleXXXRequest is more suitable to be converted to a visitor. But, anyway, it's fine to use it.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -196,6 +196,77 @@ public void handleRequests(NetworkRequest networkRequest) throws InterruptedExce
...
+  @Override
+  public void dropRequest(NetworkRequest request) throws InterruptedException {
+    try {
+      InputStream is = request.getInputStream();
+      DataInputStream dis = is instanceof DataInputStream ? (DataInputStream) is : new DataInputStream(is);
+      RequestOrResponseType type = RequestOrResponseType.values()[dis.readShort()];
+      Response response;
+      long requestQueueTime = SystemTime.getInstance().milliseconds() - request.getStartTimeInMs();
+      RequestMetricsUpdater metricsUpdater = new RequestMetricsUpdater(requestQueueTime, 0, 0, 0, true);
+      switch (type) {
```
tbingol, line 208: Thanks for creating a function to drop all request types. However, I was wondering if we could return a generic throttling error rather than a request-specific one. The reason is that we can use other frameworks for load shedding, such as Hodor, in the future. Such frameworks will be application-unaware and will likely drop the requests before they make it even this far.

ambry-network/src/main/java/com/github/ambry/network/AdaptiveLifoCoDelNetworkRequestQueue.java
```diff
@@ -0,0 +1,187 @@
...
+   * @throws InterruptedException if interrupted while waiting
+   */
+  @Override
+  public NetworkRequestBundle take() throws InterruptedException {
+    NetworkRequest requestToServe = null;
+    List<NetworkRequest> requestsToDrop = new ArrayList<>();
+    // Dequeue next request. If the request timed out, continue to dequeue until we find a valid request to serve
+    // or queue is empty.
+    NetworkRequest nextRequest;
+    while (true) {
```
snalli, line 93: while (deque.size() > 0)

ambry-network/src/main/java/com/github/ambry/network/AdaptiveLifoCoDelNetworkRequestQueue.java
```diff
@@ -0,0 +1,187 @@
...
+      if (useLifoMode()) {
+        nextRequest = deque.pollLast();
+      } else {
+        nextRequest = deque.pollFirst();
+      }
+      if (nextRequest == null) {
```
snalli, line 99: if the loops condition is on size, then we can get rid of this break ?

ambry-network/src/main/java/com/github/ambry/network/FifoNetworkRequestQueue.java
```diff
@@ -0,0 +1,97 @@
...
+      } else {
+        requestToServe = nextRequest;
+        break;
+      }
+    }
+    // If there are no requests to drop and no requests to serve currently in the queue, block until a new request comes
+    // so we can give the consumer something to do.
+    if (requestToServe == null && requestsToDrop.isEmpty()) {
+      requestToServe = queue.take();
```
snalli, line 73: i don't understand what's happening here. This call blocks or returns immediately if the queue is empty ?

ambry-network/src/main/java/com/github/ambry/network/FifoNetworkRequestQueue.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * A FIFO queue to hold network requests. It Internally, it maintains two queues, a bounded blocking queue to hold immediate
+ * set of next requests that can be handled by Ambry and unbounded queue to hold overflow requests.
```
snalli, line 27: where is this unbounded queue ? is it requestsToDrop ? is it a good idea to have an unbounded queue ?

ambry-network/src/main/java/com/github/ambry/network/http2/AmbryNetworkRequestHandler.java
```diff
@@ -41,7 +41,7 @@ public AmbryNetworkRequestHandler(RequestResponseChannel requestResponseChannel,
 
   @Override
   protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest msg) {
-
+    // I received a message
```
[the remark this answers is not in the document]
snalli: +1

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -30,18 +38,34 @@
...
+            SystemTime.getInstance());
+        break;
+      default:
+        throw new IllegalArgumentException("Queue type not supported by channel: " + config.requestQueueType);
+    }
     this.http2ServerMetrics = http2ServerMetrics;
   }
 
-  /** Send a request to be handled, potentially blocking until there is room in the queue for the request */
+  /** Send a request to be handled */
```
snalli, line 62: send to the server ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -80,30 +104,91 @@ public void closeConnection(NetworkRequest originalRequest) throws InterruptedEx
     }
   }
 
-  /** Get the next request or block until there is one */
+  /** Get the next request or block until there is one.
+   * @return*/
+  @Override
+  public NetworkRequestBundle receiveRequest() throws InterruptedException {
```
snalli, line 110: this receives requests from the client ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -80,30 +104,91 @@ public void closeConnection(NetworkRequest originalRequest) throws InterruptedEx
...
+      if (requestsToDrop.isEmpty()) {
+        // If there are no un-queued requests, wait on the queue.
+        unqueuedRequest = unqueuedRequests.take();
+        requestsToDrop.add(unqueuedRequest);
+      }
+      Set<NetworkRequest> errorRequests = new HashSet<>();
+      for (NetworkRequest requestToDrop : requestsToDrop) {
+        if (!consumeStream(requestToDrop)) {
+          errorRequests.add(requestToDrop);
+          logger.error("Error reading the stream, dropping the request: {}", requestToDrop);
```
snalli, line 162: i feel like this loop can be simplified. 

`  @Override
  public List<NetworkRequest> getUnqueuedRequestsX() throws InterruptedException {
    List<NetworkRequest> requestsToDrop = new ArrayList<>();
    NetworkRequest unqueuedRequest;
    while (requestsToDrop.isEmpty()) {
      while ((unqueuedRequest = unqueuedRequests.poll()) != null) {
        if (consumeStream(requestToDrop)) {
          requestsToDrop.add(unqueuedRequest);
        }
      }
      return requestsToDrop;
    }
  }`

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -80,30 +104,91 @@ public void closeConnection(NetworkRequest originalRequest) throws InterruptedEx
...
+        requestsToDrop.add(unqueuedRequest);
+      }
+      if (requestsToDrop.isEmpty()) {
+        // If there are no un-queued requests, wait on the queue.
+        unqueuedRequest = unqueuedRequests.take();
+        requestsToDrop.add(unqueuedRequest);
+      }
+      Set<NetworkRequest> errorRequests = new HashSet<>();
+      for (NetworkRequest requestToDrop : requestsToDrop) {
+        if (!consumeStream(requestToDrop)) {
```
snalli, line 160: 1/ why do we consumeStream if we have decided to drop ?
2/ and what do we do with the memory occupied by a request whose stream we cannot consume ? that doesn't seem to get added to the requeststoDrop queue ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -80,30 +104,91 @@ public void closeConnection(NetworkRequest originalRequest) throws InterruptedEx
...
-          stream.readLong();
-        } catch (IOException e) {
-          logger.error("Encountered an error while reading length out, close the connection", e);
-          closeConnection(request);
-          request = null;
+  public List<NetworkRequest> getUnqueuedRequests() throws InterruptedException {
+    List<NetworkRequest> requestsToDrop = new ArrayList<>();
+    NetworkRequest unqueuedRequest;
+    do {
+      while ((unqueuedRequest = unqueuedRequests.poll()) != null) {
```
snalli, line 150: hope the poll is blocking else this would be an infinite spinning loop if the drop queue is always empty..

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
snalli: can we explicitly release the mem for those erroneous requests ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -80,30 +104,91 @@ public void closeConnection(NetworkRequest originalRequest) throws InterruptedEx
...
+    return requestsToDrop;
+  }
+
+  /**
+   * Consumes the first 8 bytes of the input stream in the incoming request since TCP implementation uses this size to
+   * allocated Buffer.
+   * @param request incoming network request.
+   * @return false if there was any error while reading the first 8 bytes. Else, return true.
+   */
+  private boolean consumeStream(NetworkRequest request) throws InterruptedException {
```
snalli, line 176: this just seems like some verification method right ?

ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
```diff
@@ -45,13 +46,18 @@ public void sendRequest(NetworkRequest request) {
   }
 
   @Override
-  public NetworkRequest receiveRequest() throws InterruptedException {
+  public NetworkRequestBundle receiveRequest() throws InterruptedException {
```
snalli, line 49: is the local channel in use ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
snalli: or is this just a method used by a producer to put something into the queue ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
snalli: or is this just a method used by a consumer to get something frm the queue ?

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -196,6 +196,77 @@ public void handleRequests(NetworkRequest networkRequest) throws InterruptedExce
...
+        case PutRequest:
+          PutRequest putRequest = PutRequest.readFrom(dis, clusterMap);
+          putRequest.accept(metricsUpdater);
```
snalli, line 211: why not just metricsUpdater.accept here ? 
also, the accept doesn't really accept, kind of a misnomer

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
```diff
@@ -196,6 +196,77 @@ public void handleRequests(NetworkRequest networkRequest) throws InterruptedExce
...
+          replicaMetadataRequest.accept(metricsUpdater);
+          response = new ReplicaMetadataResponse(replicaMetadataRequest.getCorrelationId(),
+              replicaMetadataRequest.getClientId(), ServerErrorCode.Retry_After_Backoff,
+              ReplicaMetadataResponse.getCompatibleResponseVersion(replicaMetadataRequest.getVersionId()));
+          break;
+        case AdminRequest:
+          AdminRequest adminRequest = AdminRequest.readFrom(dis, clusterMap);
+          adminRequest.accept(metricsUpdater);
+          response = new AdminResponse(adminRequest.getCorrelationId(), adminRequest.getClientId(),
+              ServerErrorCode.Retry_After_Backoff);
```
snalli, line 252: Retry_After_Backoff == 429 ?

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestDropper.java
```diff
@@ -0,0 +1,69 @@
...
+
+  public void run() {
+    Collection<NetworkRequest> requestsToDrop = null;
+    while (true) {
+      try {
+        requestsToDrop = requestChannel.getUnqueuedRequests();
+        for (NetworkRequest requestToDrop : requestsToDrop) {
+          if (requestToDrop.equals(EmptyRequest.getInstance())) {
+            logger.debug("Request dropper received shut down command");
+            return;
```
snalli, line 46: where do we actually drop the request and release all the mem occupied by it ?

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestDropper.java
```diff
@@ -0,0 +1,69 @@
...
+  private final RequestResponseChannel requestChannel;
+  private final RequestAPI requests;
+  private static final Logger logger = LoggerFactory.getLogger(RequestDropper.class);
+
+  public RequestDropper(RequestResponseChannel requestChannel, RequestAPI requests) {
+    this.requestChannel = requestChannel;
+    this.requests = requests;
+  }
+
+  public void run() {
```
snalli, line 38: can this thread get blocked on sending a response ? if so, we might be blocked for a while.

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestDropper.java
```diff
@@ -0,0 +1,69 @@
...
+      } finally {
+        // Release any resources associated with requests.
+        if (requestsToDrop != null) {
+          requestsToDrop.forEach(NetworkRequest::release);
+        }
+      }
+    }
+  }
+
+  public void shutdown() throws InterruptedException {
```
snalli, line 65: can we have a more direct way of shutting down the thread ? like set a boolean and have the run method above just return when that boolean is true ?

snalli: like while(shutdown == false)

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
snalli: +1, these visit methods are just adding more unnecessary code

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
```diff
@@ -119,6 +122,35 @@ public class NetworkConfig {
   @Default("false")
   public final boolean selectorUseDirectBuffers;
 
+  /**
+   * The type of queue to use to hold requests waiting to be processed.
+   * TODO: Do tests to tune controlled delay queue parameters before switching the request queue type to {@link RequestQueueType#ADAPTIVE_LIFO_CO_DEL}
+   */
+  @Config("request.queue.type")
+  @Default("BASIC_FIFO")
```
snalli, line 130: Nit: just fifo would do unless you have an advance fifo and a basic fifo

snalli: Also these defaults are not useful anywhere

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
snalli: better name would verifyXXXHeader ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
snalli: if requestsToDrop is empty, will a call to this method block ?

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
```diff
@@ -119,6 +122,35 @@ public class NetworkConfig {
...
+  public final RequestQueueType requestQueueType;
+
+  /**
+   * The maximum time in milliseconds a request is allowed to spend in the queue waiting to be processed.
+   */
+  @Config("request.queue.timeout.ms")
+  @Default("2000")
```
snalli, line 137: how do u decide these defaults ?

justinlin-linkedin (approved):
lgtm

tbingol (approved):
LGTM. Let's return a generic throttling error rather than a request-specific one in the following iterations

JingQianCloud (commented):
As we discussed, if we drop request whenever the 64 entry queue is full is probably too aggressive. We have 10 disks. So each disk has queue depth as 6. It's not very heavy.
Without retry mechanism in the frontend, we may return too many 503.

I think we shouldn't drop request only based on the queue size. We probably should queue many more requests but start to drop request when the response time or the queue time is too long.

snalli:
@Arun-LinkedIn Can you please address my comments before merging ?

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
Arun-LinkedIn: > 1/ why do we consumeStream if we have decided to drop ?

When we say drop, we actually mean to reject the reject without handling it. Due to that, we still need to consume the first 8 bytes of the stream. Added more comments in javadoc comments on why we are consuming the first 8 bytes.

> can we explicitly release the mem for those erroneous requests ?

Thank you for pointing it out. Released the memory for erroneous requests

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
Arun-LinkedIn: Thank you for suggestion. Simplified the loop.

> or why not let just dropRequest handle what to do with these errorneous requests instead of branching here ?

I think we need to keep code to read first 8 bytes at this layer so that it can be applicable to both valid and dropped requests. 

> if requestsToDrop is empty, will a call to this method block ?

Yeah, this method will block if requestsToDrop is empty. Added it as comments in javadoc.

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestDropper.java
```diff
@@ -0,0 +1,69 @@
+/**
+ * Copyright 2016 LinkedIn Corp. All rights reserved.
```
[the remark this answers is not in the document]
Arun-LinkedIn: Done.

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
Arun-LinkedIn: Actually meant to name as basic queue of type FIFO. Renamed it. Hopefully it is less confusing now.

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
Arun-LinkedIn: Discussed with @justinlin-linkedin offline on coming up with this defaults. It is related to our request timeout at frontends and request processing time at server. Since our frontends mostly have request time out of 4 seconds and if we assume request processing time around 1.5 second (based on P95 values in our graphs) and give another 0.5 seconds to network trip, we can probably allow 2 seconds for the requests in the queue before declaring them stale.

ambry-protocol/src/main/java/com/github/ambry/protocol/AmbryRequests.java
Arun-LinkedIn: Yeah, it is an error code understood by Ambry frontends. We can treat it as similar to 429 in frontends and retry the requests.

ambry-network/src/main/java/com/github/ambry/network/LocalRequestResponseChannel.java
Arun-LinkedIn: Yeah, local channel was actually used for Cloud requests. I will remove it in a separate pr.

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
Arun-LinkedIn: These are actually to consume/discard the first 8 bytes. Added few java docs comments and renamed it to `consumeSizeHeader`. Hopefully it is less confusing now.

justinlin-linkedin (approved):

mikeyang923:
Can you address the unit test failure?
```
com.github.ambry.frontend.AmbrySecurityServiceTest > processResponseTest FAILED
    java.lang.AssertionError: Value of x-ambry-enc-um-filename not as expected expected:<%E6%B5%8B%E8%AF%95> but was:<null>

```

JingQianCloud (approved):
The unbounded queue solution which drops request only by timeout looks good to me.

PR #2306, opened 2022-12-21
Fix Leak in ReplicateBlob

justinlin-linkedin:
This PR would fix a memory in ReplicateBlob request.

In ReplicateBlobRequest, ambry-server would issue GetRequest to remote host and get a GetResponse back. After that, ambry-server would go through a transformer to transform PutBlob. In BlobIdTransformer, we would increase the ref counter for the GetResponse's InputStream when it detects that the InputStream is a NettyByteBufDataInputStream. We didn't decrease the counter any where so there is a memory leak.

This PR fixes that, by creating a new InputStream and pass this new InputStream to transformer. Since this new InputStream is not a NettyByteBufDataInputStream, the ref counter is not going to be increased. Thus, we don't have memory leak issue.


JingQianCloud (approved):
Hi Justin,
Thank you for identifying the bug and fixing it.

SophieGuo410 (approved):
lgtm!

PR #2307, opened 2022-12-21
Enable PutRequest format V5 and enable config by default.

panchow:
According to Blob Compression deployment plan, the last step is enable PutRequest V5 in frontend and enable compression by default.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
@@ -21,9 +21,9 @@ package com.github.ambry.config;
 public class CompressionConfig {
 
   // Boolean whether compression is enabled in PUT operation.
-  // TODO - Compression is disabled by default.  Need to enable after testing and verification.
+  // Compression is enabled by default.  It can be disabled using config.
   static final String COMPRESSION_ENABLED = "router.compression.enabled";
-  static final boolean DEFAULT_COMPRESSION_ENABLED = false;
+  static final boolean DEFAULT_COMPRESSION_ENABLED = true;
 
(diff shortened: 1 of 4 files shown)


ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -21,9 +21,9 @@
 public class CompressionConfig {
 
   // Boolean whether compression is enabled in PUT operation.
-  // TODO - Compression is disabled by default.  Need to enable after testing and verification.
+  // Compression is enabled by default.  It can be disabled using config.
   static final String COMPRESSION_ENABLED = "router.compression.enabled";
-  static final boolean DEFAULT_COMPRESSION_ENABLED = false;
+  static final boolean DEFAULT_COMPRESSION_ENABLED = true;
```
justinlin-linkedin, line 26: I recommend doing this in two steps
First step is to enable V5 on PutRequest without enabling compression
Second step is to enable compression.

We can keep the most of the this PR but still make this default value to false.

snalli (approved):
once the coments are addressed, this can be merged.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
mikeyang923: +1

panchow: I have switch this back to "false".  It's okay to separate them to separate PRs.

justinlin-linkedin (approved):
LGTM

mikeyang923 (approved):

PR #2311, opened 2023-01-03
Enable compression by default.

panchow:
Compression structure changes and implementation have been tested. Compression end-to-end also tested in Perf environment. This change enable compression by default.  In case something goes wrong, it can be disabled by ambry-frontend-config entry <property name="ambry.routerCompressionEnabled" value="false" />


mikeyang923 (approved):
LGTM

PR #2318, opened 2023-01-06
Frontend: Fix size mismatch error for compressed blobs in GetBlobOperations

justinlin-linkedin:
After enabling compression, we saw lots of size mismatch error log in frontend. Turns out it's because we are setting the blob size in blob properties to the size of compressed content. This PR fixes it.

Along with the fix, this PR also does:
1. Add more test cases for compression in GetBlobOperation
2. Set the default value of compression to false. We should use configuration file to change the value.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
@@ -21,8 +21,8 @@ package com.github.ambry.config;
 public class CompressionConfig {
 
   // Boolean whether compression is enabled in PUT operation.
-  static final String COMPRESSION_ENABLED = "router.compression.enabled";
-  static final boolean DEFAULT_COMPRESSION_ENABLED = true;
+  public static final String COMPRESSION_ENABLED = "router.compression.enabled";
+  static final boolean DEFAULT_COMPRESSION_ENABLED = false;
 
   // Whether to skip compression if content-encoding present.
   static final String SKIP_IF_CONTENT_ENCODED = "router.compression.skip.if.content.encoded";
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -1518,20 +1518,24 @@ class GetBlobOperation extends GetOperation {
                 blobInfo = new BlobInfo(serverBlobProperties, result.getDecryptedUserMetadata().array(), lifeVersion);
               }
               ByteBuf decryptedBlobContent = result.getDecryptedBlobContent();
-              totalSize = decryptedBlobContent.readableBytes();
-              if (!resolveRange(totalSize)) {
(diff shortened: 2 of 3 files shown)


ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
```diff
@@ -21,8 +21,8 @@
 public class CompressionConfig {
 
   // Boolean whether compression is enabled in PUT operation.
-  static final String COMPRESSION_ENABLED = "router.compression.enabled";
-  static final boolean DEFAULT_COMPRESSION_ENABLED = true;
+  public static final String COMPRESSION_ENABLED = "router.compression.enabled";
+  static final boolean DEFAULT_COMPRESSION_ENABLED = false;
```
panchow, line 25: I think we can leave compression enabled.  The code changes below fixes decompression.  The issue is in decompression, not in the compression.  I suggest we enable compression so that we can test the decompression fix below.

snalli, line 25: why disable it if the bug is fixed ?

justinlin-linkedin: we should make the default as false, so if we want to enable it, we have to turn it on in configuration file. In this way, we will be more aware of what is happening.

panchow (changes requested):

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1518,20 +1518,24 @@ protected void maybeProcessCallbacks() {
...
-              totalSize = decryptedBlobContent.readableBytes();
-              if (!resolveRange(totalSize)) {
-                routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
-                routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
-                ByteBuf decompressedContent = decompressContent(decryptedBlobContent);
-                if (decompressedContent != null) {
+              routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
+              routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
+              ByteBuf decompressedContent = decompressContent(decryptedBlobContent);
+              if (decompressedContent == null) {
```
snalli, line 1524: nit: Can u put the most likely case first to avoid a code jump ? decompressedContent != null first.

ambry-api/src/main/java/com/github/ambry/config/CompressionConfig.java
panchow: Par discussion, we can leave the feature off by default.  When deploying this feature, we need to create RB to enable this feature in each cluster and each fabric as deployment progresses.

panchow: The rest LGTM

panchow (approved):

snalli:
lgtm

snalli (approved):

PR #2319, opened 2023-01-08
Remove some unused server configurations wrt storage stats

justinlin-linkedin:
When switching from helix to mysql database for stats publishing, we added lots of configurations that are not used any where. This PR would remove those configurations and make thing simpler.


mikeyang923 (approved):

SophieGuo410 (approved):
lgtm! It seems like some config value set at tag level (for example ambry-backup-checker.src) are different as what we set in application.src. It would be better to remove it and use the application level value to keep it clear.

PR #2321, opened 2023-01-12
Fix compression ratio metric calculation.

panchow:
The there is a bug in the compression ratio calculation that affects a metric only. This fixes the calculation formula and emit compression metrics only if accepted, to be consistent with decompression.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
@@ -249,6 +249,7 @@ public class CompressionService {
     ByteBuf newChunkByteBuf = combineBuffer(chunkByteBuf, outputDirectBuffer, defaultCompressor.requireMatchingBufferType());
     ByteBuf compressedByteBuf = null;
     int actualCompressedByteBufferSize;
+    long durationMicroseconds;
     try {
       algorithmMetrics.compressRate.mark();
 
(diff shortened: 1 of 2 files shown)


ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
```diff
@@ -306,6 +296,17 @@ public ByteBuf compressChunk(ByteBuf chunkByteBuf, boolean isFullChunk, boolean
       return null;
     }
 
+    // Compress accepted, emit metrics. MB/sec = (#bytes/#microseconds) x BytesPerMicroSecondsToMBToSec
```
snalli, line 299: LGTM, just wanna know what will be the value of these metrics if we never hit this point and return from the catch-block ?

snalli (approved):

mikeyang923 (approved):

ambry-router/src/main/java/com/github/ambry/router/CompressionService.java
panchow: If not hitting this point, we don't have emit metrics.  So the metrics in the compressed blob requests.

snalli: So the value is 0 or just a gap in the graph ?

PR #2322, opened 2023-01-14
[vcr] Reduce noise in logging exceptions

snalli:
We have been getting many tickets from VCR for errors including throttling errors and blob-already-exists cases. We get tickets because we log these errors. This just causes a lot of noise. This patch converts those log messages to debug messages.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -75,6 +75,7 @@ class CloudBlobStore implements Store {
   private static final int cacheInitialCapacity = 1000;
   private static final float cacheLoadFactor = 0.75f;
   static final int STATUS_NOT_FOUND = 404;
+  static final int TOO_MANY_REQUESTS = 429;
   private static final short IGNORE_LIFE_VERSION = -2;
   private final PartitionId partitionId;
   private final CloudDestination cloudDestination;
@@ -637,6 +638,11 @@ class CloudBlobStore implements Store {
             "Delete", partitionId.toPathString());
       }
     } catch (CloudStorageException ex) {
+      if (ex.getStatusCode() == TOO_MANY_REQUESTS) {
+        // Throw no exception for throttling errors
+        logger.debug(ex.toString());
+        return;
+      }
       if (ex.getCause() instanceof StoreException) {
         throw (StoreException) ex.getCause();
       }
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -75,7 +75,6 @@ class CloudBlobStore implements Store {
   private static final int cacheInitialCapacity = 1000;
   private static final float cacheLoadFactor = 0.75f;
   static final int STATUS_NOT_FOUND = 404;
-  static final int TOO_MANY_REQUESTS = 429;
   private static final short IGNORE_LIFE_VERSION = -2;
   private final PartitionId partitionId;
   private final CloudDestination cloudDestination;
@@ -638,11 +637,6 @@ class CloudBlobStore implements Store {
             "Delete", partitionId.toPathString());
       }
     } catch (CloudStorageException ex) {
-      if (ex.getStatusCode() == TOO_MANY_REQUESTS) {
-        // Throw no exception for throttling errors
-        logger.debug(ex.toString());
-        return;
-      }
       if (ex.getCause() instanceof StoreException) {
         throw (StoreException) ex.getCause();
       }
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
@@ -426,9 +426,7 @@ public abstract class StorageClient implements AzureStorageClient {
     if (T instanceof BlobStorageException
         && ((BlobStorageException) T).getErrorCode() == BlobErrorCode.BLOB_ALREADY_EXISTS) {
       // Common case first
-      if (logger.isDebugEnabled()) {
         logger.debug("Blob store operation failed due to exception: " + T);
-      }
     } else {
       logger.error("Blob store operation failed due to exception: " + T.toString());
       azureMetrics.storageClientOperationExceptionCount.inc();
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
@@ -58,6 +58,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Predicate;
+import java.util.logging.Level;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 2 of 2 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
```diff
@@ -423,10 +424,20 @@ protected abstract BlobServiceAsyncClient buildBlobServiceAsyncClient(HttpClient
    * @param throwable associated with this error.
    */
   private void onError(Throwable throwable) {
-    logger.error("Blob store operation failed due to exception: " + throwable.toString());
-    azureMetrics.storageClientOperationExceptionCount.inc();
-    // All retries must have been completed internally by now.
-    azureMetrics.storageClientFailureAfterRetryCount.inc();
+    boolean logError = true;
```
SophieGuo410, line 427: Would it be simplify by:

> if (throwable instanceof BlobStorageException && ((BlobStorageException) throwable).getStatusCode() == BLOB_ALREADY_EXISTS) 
   {
      logger.debug("Blob store operation failed due to exception: " + throwable.toString());
    } else {
      logger.error("Blob store operation failed due to exception: " + throwable.toString());
      azureMetrics.storageClientOperationExceptionCount.inc();
      // All retries must have been completed internally by now.
      azureMetrics.storageClientFailureAfterRetryCount.inc();
    }

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
```diff
@@ -423,10 +424,20 @@ protected abstract BlobServiceAsyncClient buildBlobServiceAsyncClient(HttpClient
...
+      // Do not log this error if status = 409 and blob-exists, it's just noise
+      logError = !(bse.getStatusCode() == BLOB_ALREADY_EXISTS && bse.getMessage().equals("The specified blob already exists"));
+    }
+    if (logError) {
+      logger.error("Blob store operation failed due to exception: " + throwable.toString());
+      azureMetrics.storageClientOperationExceptionCount.inc();
+      // All retries must have been completed internally by now.
+      azureMetrics.storageClientFailureAfterRetryCount.inc();
+    } else {
+      logger.debug("Blob store operation failed due to exception: " + throwable);
```
SophieGuo410, line 439: throwable.toString()

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -637,6 +638,11 @@ public void delete(List<MessageInfo> infos) throws StoreException {
             "Delete", partitionId.toPathString());
       }
     } catch (CloudStorageException ex) {
+      if (ex.getStatusCode() == TOO_MANY_REQUESTS) {
```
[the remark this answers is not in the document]
snalli: This is based on your recommendation from your document to deal with the vast number of cosmos exceptions coming in. Additionally, my idea is to silence these 429's so that real errors become visible. 429's aren't really harmful. The impact would be slower backups and with DR efforts, that can be identified and rectified if necessary. We can enable debug level when we want to debug. Besides, we can see 429s and other errors in azure portal as a first response. The current logging isn't really adding any value.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: probably, but the boolean expression is too long. I can eliminate the logError variable at most.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: Yeah, intellij recommended its not necessary.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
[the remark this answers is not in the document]
snalli: Wait, based on your investigation, isn't the vast majority of errors 429 ? However, i didn't find many of those in the logs when i simply tailed the logs. The number of 429s in prod is 0 at the moment.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: Also 409 can have multiple error messages. So we should check the error msg also.
https://learn.microsoft.com/en-us/rest/api/storageservices/blob-service-error-codes

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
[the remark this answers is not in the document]
SophieGuo410: I suggest we can add metrics for 429 errors, and if the 429 percentage is less then 5% which mentioned in the wiki, we can just ignore this issue. But if it's above that threshold, we need to think about increasing the RU. Azure has it's own metric but we can't set alert on it.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: shortened it by checked error code instead of status code

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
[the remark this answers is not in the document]
snalli: removed

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
```diff
@@ -420,13 +420,19 @@ protected abstract BlobServiceAsyncClient buildBlobServiceAsyncClient(HttpClient
...
-    azureMetrics.storageClientOperationExceptionCount.inc();
-    // All retries must have been completed internally by now.
-    azureMetrics.storageClientFailureAfterRetryCount.inc();
+  private void onError(Throwable T) {
+    if (T instanceof BlobStorageException
+        && ((BlobStorageException) T).getErrorCode() == BlobErrorCode.BLOB_ALREADY_EXISTS) {
+      // Common case first
+      logger.debug("Blob store operation failed due to exception: " + T);
+    } else {
+      logger.error("Blob store operation failed due to exception: " + T.toString());
```
panchow, line 431: .toString() is redundant, but minor.

panchow (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
```diff
@@ -420,13 +420,19 @@ protected abstract BlobServiceAsyncClient buildBlobServiceAsyncClient(HttpClient
...
    * Handle any errors. For now, we update the error metrics.
-   * @param throwable associated with this error.
+   * @param T associated with this error.
    */
-  private void onError(Throwable throwable) {
-    logger.error("Blob store operation failed due to exception: " + throwable.toString());
-    azureMetrics.storageClientOperationExceptionCount.inc();
-    // All retries must have been completed internally by now.
-    azureMetrics.storageClientFailureAfterRetryCount.inc();
+  private void onError(Throwable T) {
```
SophieGuo410, line 425: it's minor, but please use lower case for variable names.

SophieGuo410 (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: yeah i noticed it too, strangely intellij says its redundant on L429 in the if-case but not in else-case.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: ok

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: done

PR #2324, opened 2023-01-17
Support replication with nonblocking network client

justinlin-linkedin:
Start using nonblocking network client in replica thread to boost up replication speed for cross-colo replication.

This PR adds nonblocking network client support in ReplicaThread. It mimics the logic/code in OperationControllers in router package. It first divides replicas to replicate into several groups and for each group, it sends the request to the remote servers and feed the response back to the group.

This PR also adds nonblocking network client to the existing test cases., and add a new test case to test replication for multiple remote hosts at the same time.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -224,22 +224,35 @@ public class ReplicationConfig {
   @Default("120")
   public final int replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds;
 
-  @Config(REPLICATION_REQUEST_NETWORK_TIMEOUT_MS)
-  @Default("10 * 1000")
-  public final long replicationRequestNetworkTimeoutMs;
-  public final static String REPLICATION_REQUEST_NETWORK_TIMEOUT_MS = "replication.request.network.timeout.ms";
-
-  @Config(REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS)
-  @Default("40")
-  public final long replicationRequestNetworkPollTimeoutMs;
-  public final static String REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS =
-      "replication.request.network.poll.timeout.ms";
-
+  /**
+   * True to start using nonblocking network client for remote colo replication. The remote colo replication is a lot
+   * slower than the intra-colo replication so using a nonblocking network client should greatly improve the speed of
+   * replication. When doing so, also set the number of inter-colo replication thread to 1 since we only need one thread
+   * when using a nonblocking network client.
+   */
   @Config(REPLICATION_USING_NONBLOCKING_NETWORK_CLIENT_FOR_REMOTE_COLO)
   @Default("false")
   public final boolean replicationUsingNonblockingNetworkClientForRemoteColo;
   public final static String REPLICATION_USING_NONBLOCKING_NETWORK_CLIENT_FOR_REMOTE_COLO =
       "replication.using.nonblocking.network.client.for.remote.colo";
+  /**
+   * Timeout in milliseconds for requests in nonblocking network client. This configuration is only useful when the
+   * nonblocking network client is enabled.
+   */
+  @Config(REPLICATION_REQUEST_NETWORK_TIMEOUT_MS)
+  @Default("10 * 1000")
+  public final long replicationRequestNetworkTimeoutMs;
+  public final static String REPLICATION_REQUEST_NETWORK_TIMEOUT_MS = "replication.request.network.timeout.ms";
+
+  /**
+   * Timeout in milliseconds for polling the nonblocking network client to fetch more responses from remote hosts. This
+   * configuration is only useful when the nonblocking network client is enabled.
+   */
+  @Config(REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS)
+  @Default("40")
+  public final long replicationRequestNetworkPollTimeoutMs;
+  public final static String REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS =
+      "replication.request.network.poll.timeout.ms";
 
   public ReplicationConfig(VerifiableProperties verifiableProperties) {
 
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -224,6 +224,23 @@ public class ReplicationConfig {
   @Default("120")
   public final int replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds;
 
+  @Config(REPLICATION_REQUEST_NETWORK_TIMEOUT_MS)
+  @Default("10 * 1000")
+  public final long replicationRequestNetworkTimeoutMs;
+  public final static String REPLICATION_REQUEST_NETWORK_TIMEOUT_MS = "replication.request.network.timeout.ms";
+
+  @Config(REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS)
+  @Default("40")
+  public final long replicationRequestNetworkPollTimeoutMs;
+  public final static String REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS =
+      "replication.request.network.poll.timeout.ms";
+
+  @Config(REPLICATION_USING_NONBLOCKING_NETWORK_CLIENT_FOR_REMOTE_COLO)
+  @Default("false")
+  public final boolean replicationUsingNonblockingNetworkClientForRemoteColo;
+  public final static String REPLICATION_USING_NONBLOCKING_NETWORK_CLIENT_FOR_REMOTE_COLO =
+      "replication.using.nonblocking.network.client.for.remote.colo";
+
   public ReplicationConfig(VerifiableProperties verifiableProperties) {
 
     replicationStoreTokenFactory =
@@ -275,5 +292,10 @@ public class ReplicationConfig {
     replicationStandbyWaitTimeoutToTriggerCrossColoFetchSeconds =
         verifiableProperties.getIntInRange(REPLICATION_STANDBY_WAIT_TIMEOUT_TO_TRIGGER_CROSS_COLO_FETCH_SECONDS, 120,
             -1, Integer.MAX_VALUE);
+    replicationRequestNetworkTimeoutMs = verifiableProperties.getLong(REPLICATION_REQUEST_NETWORK_TIMEOUT_MS, 10000);
+    replicationRequestNetworkPollTimeoutMs =
+        verifiableProperties.getLong(REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS, 40);
+    replicationUsingNonblockingNetworkClientForRemoteColo =
+        verifiableProperties.getBoolean(REPLICATION_USING_NONBLOCKING_NETWORK_CLIENT_FOR_REMOTE_COLO, false);
   }
 }
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -2094,30 +2094,22 @@ public class ReplicaThread implements Runnable {
       }
       List<RequestInfo> requestInfos =
           pollRemoteReplicaGroups(remoteReplicaGroups, correlationIdToRequestInfo, correlationIdToReplicaGroup);
-      List<ResponseInfo> timedOutResponses =
-          filterTimedOutRequests(correlationIdToRequestInfo, correlationIdToReplicaGroup);
-      Set<Integer> requestsToDrop = timedOutResponses.stream()
-          .map(resp -> resp.getRequestInfo().getRequest().getCorrelationId())
-          .collect(Collectors.toSet());
+      Set<Integer> requestsToDrop = filterTimedOutRequests(correlationIdToRequestInfo, correlationIdToReplicaGroup);
 
       final int pollTimeoutMs = (int) replicationConfig.replicationRequestNetworkPollTimeoutMs;
       List<ResponseInfo> responseInfos = networkClient.sendAndPoll(requestInfos, requestsToDrop, pollTimeoutMs);
-      // Adding the timed out responses after network client, since a timed out request might have its response coming
-      // back in the sendAndPoll call. We want to handle the responses from the network client first if they are present.
-      responseInfos.addAll(timedOutResponses);
       onResponses(responseInfos, correlationIdToReplicaGroup);
     }
     logger.trace("Thread name: {} Finish all RemoteReplicaGroup replication", threadName);
-    remoteReplicaGroups.stream()
-        .forEach(
-            g -> exchangeMetadataResponsesInEachCycle.computeIfAbsent(g.getRemoteDataNode(), k -> new ArrayList<>())
-                .addAll(g.getExchangeMetadataResponseList()));
+    remoteReplicaGroups.forEach(
+        g -> exchangeMetadataResponsesInEachCycle.computeIfAbsent(g.getRemoteDataNode(), k -> new ArrayList<>())
+            .addAll(g.getExchangeMetadataResponseList()));
 
     // Print out the exceptions.
-    remoteReplicaGroups.stream().filter(g -> g.isDone() && g.getException() != null).forEach(g -> {
-      logger.error("Remote node: {} Thread name: {} RemoteReplicaGroup {} has exception {}", g.getRemoteDataNode(),
-          threadName, g.getRemoteReplicaInfos(), g.getException());
-    });
+    remoteReplicaGroups.stream()
+        .filter(g -> g.isDone() && g.getException() != null)
+        .forEach(g -> logger.error("Remote node: {} Thread name: {} RemoteReplicaGroup {} has exception {}",
+            g.getRemoteDataNode(), threadName, g.getRemoteReplicaInfos(), g.getException()));
 
     replicationMetrics.updateOneCycleReplicationTime(time.milliseconds() - oneRoundStartTimeMs,
         replicatingFromRemoteColo, datacenterName);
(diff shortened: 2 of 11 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1754,482 @@ void shutdown() throws InterruptedException {
...
+      if (requestInfo == null) {
+        // Only possible when warming up the network client
+        DataNodeId dataNodeId = responseInfo.getDataNode();
+        responseHandler.onConnectionTimeout(dataNodeId);
+      } else {
+        RemoteReplicaGroup remoteReplicaGroup =
+            correlationIdToReplicaGroup.remove(requestInfo.getRequest().getCorrelationId());
+        // This correlation id might be removed because the corresponding request timed out. But later we get the response
+        // back from the network client.
+        if (remoteReplicaGroup != null) {
```
JingQianCloud, line 2212: Can we add trace if the remoteReplicaGroup is null?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1754,482 @@ void shutdown() throws InterruptedException {
...
+              ReplicaThread.this.handleReplicaMetadataResponse(response, remoteReplicaInfos, remoteDataNode);
+          state = ReplicaGroupState.REPLICA_METADATA_RESPONSE_HANDLED;
+          logger.trace(
+              "Remote node: {} Thread name: {} RemoteReplicaGroup {} Handled ReplicaMetadataResponse for correlation id {}, set state to {}",
+              dataNodeId, threadName, id, responseInfo.getRequestInfo().getRequest().getCorrelationId(), state);
+          long exchangeMetadataTimeInMs = time.milliseconds() - exchangeMetadataStartTimeInMs;
+          replicationMetrics.updateExchangeMetadataTime(exchangeMetadataTimeInMs, replicatingFromRemoteColo,
+              replicatingOverSsl, datacenterName);
+        } catch (Exception e) {
+          setException(e, "ReplicaMetadataResponse deserialization received unexpected error");
```
JingQianCloud, line 1975: After we set the exception, shall we set state = ReplicaGroupState.DONE?
Same in the following "else" condition.

JingQianCloud: Never mind, the setException method has done that.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -391,6 +418,8 @@ public void replicate() {
             List<ExchangeMetadataResponse> exchangeMetadataResponseList =
                 exchangeMetadata(connectedChannel, replicaSubList);
             exchangeMetadataTimeInMs = time.milliseconds() - startTimeInMs;
+            exchangeMetadataResponsesInEachCycle.computeIfAbsent(remoteNode, k -> new ArrayList<>())
```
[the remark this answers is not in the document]
justinlin-linkedin: yes, this is used in the test cases, to verify the missing blobs in each replication cycle. I will add some comments.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1754,482 @@ void shutdown() throws InterruptedException {
...
+    }
     shutdownLatch.await();
   }
+
+  /**
+   * ReplicaGroupState indicates different state of RemoteReplicaGroup in one replication cycle.
+   * Each replication cycle, we will break all the replicas in order different groups and each group
+   * has to go through these states to finish replication with nonblocking network client.
+   */
+  enum ReplicaGroupState {
```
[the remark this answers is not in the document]
justinlin-linkedin: sure, if this causes confusion.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1754,482 @@ void shutdown() throws InterruptedException {
...
+     * Constructor of {@link RemoteReplicaGroup}.
+     * @param remoteReplicaInfos The list {@link RemoteReplicaInfo} that point to the list of replicas.
+     * @param dataNodeId The remote data no
+     * @param standbyCatchingUp True if this is standby replicas trying to catch up with the missing blob content. If true,
+     *                          this group should start with REPLICA_METADATA_RESPONSE_HANDLED, since it already exchanged
+     *                          ReplicaMetadata before and now wants to download missing blobs.
+     * @param id The id of this RemoteReplicaGroup. This id should be unique at each cycle.
+     */
+    public RemoteReplicaGroup(List<RemoteReplicaInfo> remoteReplicaInfos, DataNodeId dataNodeId,
+        boolean standbyCatchingUp, int id) {
```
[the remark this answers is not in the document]
justinlin-linkedin: this boolean in used in some other place. I think we should keep it. I can rename it to something else if this particular name causes any confusion.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1754,482 @@ void shutdown() throws InterruptedException {
...
+      this.remoteReplicaInfos = remoteReplicaInfos;
+      this.remoteDataNode = dataNodeId;
+      this.standbyCatchingUp = standbyCatchingUp;
+      this.id = id;
+      if (standbyCatchingUp) {
+        exchangeMetadataResponseList = remoteReplicaInfos.stream()
```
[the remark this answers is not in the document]
justinlin-linkedin: Poll should just be doing the requests, for state transition, when the responses come back ,we have to pass the response to whatever function is doing state transition. It's more clear to separate the request creation and the response handling methods.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1759,484 @@ void shutdown() throws InterruptedException {
...
+        DataNodeId dataNodeId = responseInfo.getDataNode();
+        responseHandler.onConnectionTimeout(dataNodeId);
+      } else {
+        int correlationId = requestInfo.getRequest().getCorrelationId();
+        RemoteReplicaGroup remoteReplicaGroup = correlationIdToReplicaGroup.remove(correlationId);
+        // This correlation id might be removed because the corresponding request timed out. But later we get the response
+        // back from the network client.
+        if (remoteReplicaGroup != null) {
+          RequestOrResponseType type = ((RequestOrResponse) requestInfo.getRequest()).getRequestType();
+          logger.debug("Remote node: {} Thread name: RemoteReplicaGroup {} Handling response of type {} for {}",
```
SophieGuo410, line 2219: You have 5 arguments need to be printed out, but only four "{}"

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1759,484 @@ void shutdown() throws InterruptedException {
...
+            case GetRequest:
+              remoteReplicaGroup.handleGetResponse(responseInfo);
+              break;
+            default:
+              logger.error(
+                  "Remote node: {} Thread name: {} RemoteReplicaGroup {} Unexpected type: {} received, discarding",
+                  remoteReplicaGroup.getRemoteDataNode(), threadName, remoteReplicaGroup.getId(), type);
+          }
+        } else {
+          logger.trace("Thread Name: {} No RemoteReplicaGroup found for correlation Id {}, it might already time out",
```
SophieGuo410, line 2234: Same here.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1759,484 @@ void shutdown() throws InterruptedException {
...
+        inFlightRequestIterator.remove();
+        responseInfos.add(new ResponseInfo(requestInfo, NetworkClientErrorCode.TimeoutError, null));
+        RemoteReplicaGroup group = correlationIdToRemoteReplicaGroup.remove(entry.getKey());
+        if (group != null) {
+          logger.trace(
+              "Remote node: {} Thread name: {} RemoteReplicaGroup {} Request {} timed out, it was issued at {}",
+              group.getRemoteDataNode(), threadName, group.getId(), entry.getKey(), requestInfo.getRequestCreateTime());
+        } else {
+          // This shouldn't happen
+          logger.trace("Thread name: {} Request {} timed out", threadName, group.getId(), entry.getKey());
```
SophieGuo410, line 2185: You have three arguments want to be printed here and only two "{}"

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -239,6 +241,8 @@ public ReplicationMetrics(MetricRegistry registry, List<? extends ReplicaId> rep
         registry.histogram(MetricRegistry.name(ReplicaThread.class, "SslIntraColoBatchStoreWriteTime"));
     intraColoTotalReplicationTime =
         registry.histogram(MetricRegistry.name(ReplicaThread.class, "IntraColoTotalReplicationTime"));
+    intraColoOneCycleReplicationTime =
+        registry.histogram(MetricRegistry.name(ReplicaThread.class, "IntraColoOneRoundReplicationTime"));
```
SophieGuo410, line 245: OneCycle -> OneRound? I feel like using different name is hard to debug.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1759,484 @@ void shutdown() throws InterruptedException {
...
+
+    // Before each cycle of replication, we clean up the cache in key converter.
+    storeKeyConverter.dropCache();
+    Map<DataNodeId, List<RemoteReplicaInfo>> dataNodeToRemoteReplicaInfo = getRemoteReplicaInfos();
+    for (Map.Entry<DataNodeId, List<RemoteReplicaInfo>> entry : dataNodeToRemoteReplicaInfo.entrySet()) {
+      DataNodeId remoteNode = entry.getKey();
+      List<RemoteReplicaInfo> replicasToReplicatePerNode = entry.getValue();
+      List<RemoteReplicaInfo> activeReplicasPerNode = new ArrayList<>();
+      List<RemoteReplicaInfo> standbyReplicasWithNoProgress = new ArrayList<>();
+      filterRemoteReplicasToReplicate(replicasToReplicatePerNode, activeReplicasPerNode, standbyReplicasWithNoProgress);
```
JingQianCloud, line 2068: Probably it's not introduced by this PR.
Looks like for each replication cycle, we have to check replica status and filter active, standbyReplicas and setup groups. It introduced extra overhead.
Ideally only when there is some cluster map change, we may need do that.
Not necessary for this PR. Just something if we can improve.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1759,484 @@ void shutdown() throws InterruptedException {
...
+      Port port = remoteReplicaInfos.get(0).getPort();
+      long timeout = replicationConfig.replicationRequestNetworkTimeoutMs;
+      if (state == ReplicaGroupReplicationState.STARTED) {
+        // When the state is STARTED, we will send the ReplicaMetadataRequest out.
+        replicaMetadataRequestStartTimeMs = time.milliseconds();
+        exchangeMetadataStartTimeInMs = replicaMetadataRequestStartTimeMs;
+        ReplicaMetadataRequest request = createReplicaMetadataRequest(remoteReplicaInfos, remoteDataNode);
+        RequestInfo requestInfo =
+            new RequestInfo(remoteDataNode.getHostname(), port, request, remoteReplicaInfos.get(0).getReplicaId(), null,
+                time.milliseconds(), timeout, timeout);
```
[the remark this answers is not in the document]
justinlin-linkedin: we don't have a quota in replica thread, so we can just ignore the time requests are queued in before network client.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -577,6 +617,11 @@ List<ExchangeMetadataResponse> exchangeMetadata(ConnectedChannel connectedChanne
       ReplicaMetadataResponse response =
           getReplicaMetadataResponse(replicasToReplicatePerNode, connectedChannel, remoteNode);
       return handleReplicaMetadataResponse(response, replicasToReplicatePerNode, remoteNode);
+    } catch (Exception e) {
+      if (e instanceof ReplicationException) {
```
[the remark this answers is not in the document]
justinlin-linkedin: not really, for ReplicationException, we just want to rethrow it, but if it's not ReplicationException, we want to wrap it up in a ReplicationException.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -342,9 +356,23 @@ Map<DataNodeId, List<RemoteReplicaInfo>> getRemoteReplicaInfos() {
...
+  /**
+   * Do replication for replicas grouped by {@link DataNodeId} with blocking channel.
+   */
+  public void replicateWithBlockingChannel() {
+    exchangeMetadataResponsesInEachCycle = new HashMap<>();
     boolean allCaughtUp = true;
+    long oneRoundStartTimeMs = time.milliseconds();
     Map<DataNodeId, List<RemoteReplicaInfo>> dataNodeToRemoteReplicaInfo = getRemoteReplicaInfos();
 
+    storeKeyConverter.dropCache();
```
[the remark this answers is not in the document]
justinlin-linkedin: We don't have to drop caches in storekeyconverter, everytime we deal with one sublist of replicas, we can drop the cache all together after we finish one cycle of replication.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
justinlin-linkedin: It's not that. We have different logic for standby replicas replicating from remote datacenter. If the local replica is a standby replica, then it doesn't replicate blob content from any replicas in remote datacenter.

For example, in DC1, we have replica A, B, C and in DC2, we have replica X, Y, Z. Replica A and X are the leaders for each DC. If this host has replica B, and it will try to replicate from A, C and X, Y, Z. However, copying blob content cross datacenter is expensive, so we don't allow B to copy blob content, we only allow B to replicate metadata cross datacenter. So B would replicate metadata and data from A and C, but only metadata from X, Y, Z.

This filter logic happens before we send out the GetRequest to replicate data, so we will filter out the RemoteReplica X, Y, Z.

SophieGuo410 (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
JingQianCloud: My point is that theoretically we don't need do this for every cycle. 
For this host and this partition, we always replicate data from A and C and always filter out X,Y,Z. It never change unless A or C is down and B become the leader.

JingQianCloud: But I guess the code path is not the performance bottleneck yet, it's ok.

JingQianCloud (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1759,484 @@ void shutdown() throws InterruptedException {
...
+        RemoteReplicaGroup group =
+            new RemoteReplicaGroup(standbyReplicasTimedOutOnNoProgress, remoteNode, true, remoteReplicaGroupId++);
+        remoteReplicaGroups.add(group);
+      }
+    }
+    // A map from correlation id to RemoteReplicaGroup. This is used to find the group when response comes back.
+    Map<Integer, RemoteReplicaGroup> correlationIdToReplicaGroup = new HashMap<>();
+    // A map from correlation id to RequestInfo. This is used to find timed out RequestInfos.
+    Map<Integer, RequestInfo> correlationIdToRequestInfo = new LinkedHashMap<>();
+    while (!remoteReplicaGroups.stream().allMatch(RemoteReplicaGroup::isDone)) {
```
Arun-LinkedIn, line 2091: Just a question. Since we are waiting for replication to complete across all hosts for this cycle, if one of the remote host is slow or has connection problems, would it slow replication for all hosts. Just thinking if this similar to what we see during Threads.join().

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
```diff
@@ -224,6 +224,36 @@ public class ReplicationConfig {
...
+  @Default("10 * 1000")
+  public final long replicationRequestNetworkTimeoutMs;
+  public final static String REPLICATION_REQUEST_NETWORK_TIMEOUT_MS = "replication.request.network.timeout.ms";
+
+  /**
+   * Timeout in milliseconds for polling the nonblocking network client to fetch more responses from remote hosts. This
+   * configuration is only useful when the nonblocking network client is enabled.
+   */
+  @Config(REPLICATION_REQUEST_NETWORK_POLL_TIMEOUT_MS)
+  @Default("40")
```
Arun-LinkedIn, line 252: Just an observation. When using network client in router, we seem to have poll timeout around 1 second.

Arun-LinkedIn (approved):
Mostly looks good. Just a few comments.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1721,6 +1759,484 @@ void shutdown() throws InterruptedException {
...
+    logger.trace("Thread Name: {} Trying to filter out timed out requests", threadName);
+    List<ResponseInfo> responseInfos = new ArrayList<>();
+    Iterator<Map.Entry<Integer, RequestInfo>> inFlightRequestIterator = correlationIdToRequest.entrySet().iterator();
+    while (inFlightRequestIterator.hasNext()) {
+      Map.Entry<Integer, RequestInfo> entry = inFlightRequestIterator.next();
+      RequestInfo requestInfo = entry.getValue();
+      long currentTimeInMs = time.milliseconds();
+      if (currentTimeInMs > requestInfo.getRequestCreateTime() + replicationConfig.replicationRequestNetworkTimeoutMs) {
+        inFlightRequestIterator.remove();
+        responseInfos.add(new ResponseInfo(requestInfo, NetworkClientErrorCode.TimeoutError, null));
```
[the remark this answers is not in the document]
justinlin-linkedin: oh, you are right, I don't have to create ResponseInfo for the dropped requests in ReplicaThread, NetworkClient would do that for me.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
justinlin-linkedin: That's right, and that's the design for now.  We are having two implementation for replication, one with blocking channel and the other with nonblocking network client. I want to make sure that we can keep these two implementations during testing for nonblocking network client. So I have to do one cycle replication at a time.

After we remove the blocking channel and only use nonblocking network client, we can then replicate same RemoteReplicaGroup at different cycle without joining.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
justinlin-linkedin: I want to be a bit faster on the polling and see if it makes replication faster.

PR #2325, opened 2023-01-18
Add Dataset and Dataset Version table in Account metadata DB

SophieGuo410:
Adding Datasets and DatasetVersions table in order to support PROML migration.
Will ask MYSQL team to review it again before it gets merged.
In order to support auto purge, the naming convention for delete_ts has to follow this exact format with "_"
And to keep consistent with other table, for non auto-purge related field, I follow what we have in Accounts and Containers table.

### DIFF
--- a/ambry-account/src/main/resources/AccountSchema.ddl
+++ b/ambry-account/src/main/resources/AccountSchema.ddl
@@ -44,3 +44,27 @@ CREATE TABLE IF NOT EXISTS Containers
     INDEX statusIndex (status)
 )
 CHARACTER SET utf8 COLLATE utf8_bin;
+
+CREATE TABLE IF NOT EXISTS Datasets (
+    accountId INT NOT NULL,
+    containerId INT NOT NULL,
+    datasetName VARCHAR(255) NOT NULL,
+    versionSchema INT NOT NULL,
+    retentionCount INT DEFAULT NULL,
+    userTags JSON DEFAULT NULL,
+    lastModifiedTime DATETIME(3) NOT NULL,
+    delete_ts datetime(6) DEFAULT NULL,
+    PRIMARY KEY (accountId, containerId, datasetName)
+)
+CHARACTER SET utf8 COLLATE utf8_bin;
+
+CREATE TABLE IF NOT EXISTS DatasetVersions (
+    accountId INT NOT NULL,
+    containerId INT NOT NULL,
+    datasetName VARCHAR(255) NOT NULL,
+    version bigint NOT NULL,
+    lastModifiedTime DATETIME(3) NOT NULL,
+    delete_ts datetime(6) DEFAULT NULL,
+    PRIMARY KEY (accountId, containerId, datasetName, version)
+)
+CHARACTER SET utf8 COLLATE utf8_bin;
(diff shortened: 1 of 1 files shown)


ambry-account/src/main/resources/AccountSchema.ddl
```diff
@@ -44,3 +44,27 @@ CREATE TABLE IF NOT EXISTS Containers
     INDEX statusIndex (status)
 )
 CHARACTER SET utf8 COLLATE utf8_bin;
+
+CREATE TABLE IF NOT EXISTS Datasets (
+    accountId INT NOT NULL,
+    containerId INT NOT NULL,
+    datasetName VARCHAR(255) NOT NULL,
```
tbingol, line 51: Is the 255 character limit consistent with the Wormhole?

ambry-account/src/main/resources/AccountSchema.ddl
```diff
@@ -44,3 +44,27 @@ CREATE TABLE IF NOT EXISTS Containers
...
+    lastModifiedTime DATETIME(3) NOT NULL,
+    delete_ts datetime(6) DEFAULT NULL,
+    PRIMARY KEY (accountId, containerId, datasetName)
+)
+CHARACTER SET utf8 COLLATE utf8_bin;
+
+CREATE TABLE IF NOT EXISTS DatasetVersions (
+    accountId INT NOT NULL,
+    containerId INT NOT NULL,
+    datasetName VARCHAR(255) NOT NULL,
```
tbingol, line 64: Same as above: Is the 255 character limit consistent with the Wormhole?

ambry-account/src/main/resources/AccountSchema.ddl
```diff
@@ -44,3 +44,27 @@ CREATE TABLE IF NOT EXISTS Containers
...
+    version bigint NOT NULL,
```
tbingol, line 65: For consistency reasons, s/bigint/BIGINT

tbingol (approved):

ambry-account/src/main/resources/AccountSchema.ddl
```diff
@@ -44,3 +44,27 @@ CREATE TABLE IF NOT EXISTS Containers
...
+
+CREATE TABLE IF NOT EXISTS Datasets (
+    accountId INT NOT NULL,
+    containerId INT NOT NULL,
+    datasetName VARCHAR(255) NOT NULL,
+    versionSchema INT NOT NULL,
+    retentionCount INT DEFAULT NULL,
+    userTags JSON DEFAULT NULL,
+    lastModifiedTime DATETIME(3) NOT NULL,
+    delete_ts datetime(6) DEFAULT NULL,
```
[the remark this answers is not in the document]
SophieGuo410: I can change the DATETIME, but for delete_ts, as I described in the comments, mysql need the exact name with "_" to support auto purge, so I can't change it.

ambry-account/src/main/resources/AccountSchema.ddl
SophieGuo410: Wormhole can only support 100 characters. Do we need to keep it same as Wormhole? So far I just keep it same as named blob DB, probably I can decrease a little bit since the blob Name = datasetName+version.

tbingol: It is good as long as we support at least as much as Wormhole supports

PR #2326, opened 2023-01-19
Handle QuotaReject error in operations, not SimpleOperationTracker

justinlin-linkedin:
Simplify the SimpleOperationTracker to move the logic of handling QuotaRejection from it to different Operations.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -1221,6 +1221,7 @@ class PutOperation {
             passedInBlobProperties.getExternalAssetTag(), passedInBlobProperties.getContentEncoding(),
             passedInBlobProperties.getFilename());
         operationTracker = getOperationTracker();
+        chunkException = null;
         correlationIdToChunkPutRequestInfo.clear();
         logger.trace("{}: Chunk {} is ready for sending out to server", loggingContext, chunkIndex);
         state = ChunkState.Ready;
@@ -1424,7 +1425,8 @@ class PutOperation {
     void checkAndMaybeComplete() {
       boolean done = false;
       // Now, check if this chunk is done.
-      if (operationTracker.isDone()) {
+      if (operationTracker.isDone() || (chunkException != null
+          && chunkException.getErrorCode() == RouterErrorCode.TooManyRequests)) {
         if (!operationTracker.hasSucceeded()) {
           failedAttempts++;
           appendSlippedPutBlobId(chunkBlobId);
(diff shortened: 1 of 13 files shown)


ambry-router/src/test/java/com/github/ambry/router/TtlUpdateManagerTest.java
```diff
@@ -177,20 +179,28 @@ public void metadataChunkUpdatedLastTest() throws Exception {
...
         .forEach(mockServer -> mockServer.setErrorCodeForBlob(blobIds.get(0), ServerErrorCode.Unknown_Error));
-    executeOpAndVerify(blobIds, RouterErrorCode.UnexpectedInternalError, false, false, false,
+    executeOpAndVerifyWithInfinity(blobIds, RouterErrorCode.UnexpectedInternalError, false, false, false,
         blobIds.subList(1, blobIds.size()));
   }
 
+  /**
+   * Test for the case when delete operation fails due to quota compliance.
+   */
+  @Test
```
ankagrawal, line 189: Is TtlUpdateManagerTest the only place where we are testing if operation fails due to quota compliance? Do we have such tests for Put, Delete, Undelete and Get as well?

SophieGuo410 (approved):
LGTM overall. Once you address all comments and make all tests passed, it's ok to merge.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1424,7 +1425,8 @@ int fillFrom(ByteBuf channelReadBuf) {
     void checkAndMaybeComplete() {
       boolean done = false;
       // Now, check if this chunk is done.
-      if (operationTracker.isDone()) {
+      if (operationTracker.isDone() || (chunkException != null
```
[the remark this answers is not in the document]
justinlin-linkedin: PutOperation is different from other operations, it doesn't have a `operationCompleted` or `chunkCompleated` boolean variable to terminate operation. Other operations has this variable, so when the quota rejection happens, I can just set this variable to true and the operation would be terminated. But since PutOperation doesn't this variable, I have to check the chunkException and terminate it when quota rejected.

ambry-router/src/test/java/com/github/ambry/router/TtlUpdateManagerTest.java
justinlin-linkedin: We do, it seems like only TtlUpdateOperation doesn't have such test. Please search on GetBlobOperationTest, UndeleteManagerTest, DeleteManagerTest and PutOperationTest file to search QuotaReject. you will find all these test files have some test cases that cover quota rejection error.

ankagrawal (approved):

PR #2327, opened 2023-01-19
Setup up a process to clean up stale named blob related data systematically

mikeyang923:
Setup up a process to clean up stale named blob related data systematically
Stale named blob refer to any old entries that are created more than 2 months ago and meet one of below conditions: 
1.  Has blob_state=0 (In_Progress). or
2. Has version lower than the latest blob_state=1 version (which mean the named blob was updated, and the previous versions can be cleaned up).


ambry-api/src/main/java/com/github/ambry/named/StaleNamedResult.java
```diff
@@ -0,0 +1,111 @@
...
+    StaleNamedResult record = (StaleNamedResult) o;
+    return accountId == record.accountId && containerId == record.containerId && Objects.equals(blobName, record.blobName)
+        && Objects.equals(blobId, record.blobId) && version == record.version && deleteTs.equals(record.deleteTs);
+  }
+
+  @Override
+  public String toString() {
+    return "StaleNamedResult[accountId=" + accountId + ",containerId=" + containerId + ",blobName=" + blobName +
+        ",blobId=" + blobId + ",version=" + version + ",deleteTs=" + deleteTs + "]";
+  }
```
justinlin-linkedin, line 110: please format this file.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobsCleanupRunner.java
```diff
@@ -0,0 +1,50 @@
...
+  @Override
+  public void run() {
+    logger.info("Named Blobs Cleanup Runner is initiated");
+    try {
+      List<StaleNamedResult> staleResultList = namedBlobDb.pullStaleBlobIds().get();
+      for (StaleNamedResult staleResult : staleResultList) {
+        router.deleteBlob(staleResult.getBlobId(), "ambry-named-blobs-cleanup-runner").get();
+      }
+      namedBlobDb.cleanupStaleData(staleResultList);
+      logger.info("Named Blobs Cleanup Runner is completed for {} stale cases", staleResultList.size());
```
justinlin-linkedin, line 45: running this in a serial order is fine and probably preferred. However, I think we should not just throw out an exception when there is one stale result failed on delete. We should filter the failed result out and keep move on to the cleanup.

That being said, we should probably limit the number of stale results for each query from named blob DB. And only deal with say 10 at a time. If we get 1M stale result from one query and we will be stuck in delete for a long time. And if one of them failed, we will fail the clean up as well.

so we can probably do
```
List<StaleNamedResult> staleResultList = namedBlob.pullStaleBlobIds().get();
List<StaleNamedResult> resultsToCleanup = new ArrayList<>();
List<StaleNamedResult> failedResults = new ArrayList<>();
while(!staleResultList.isEmpty()) {
   resultsToCleanup.clear();
   for (StaleNameResult result: staleResultList) {
        try {
            router.deleteBlob(result.getBlobId(), "ambry-named-blobs-cleanup-runner").get();
            resultsToCleanup.add(result);
        } catch (Exception e) {
            logger.error("")
            failedResults.add(result);
        }
    }
    namedBlobDb.cleaupStaleData(resultsToCleanup);
    staleResultList = namedBlob.pullStaleBlobIds().get();
    staleResultList.removeAll(failedResults);
}
```

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -135,6 +135,23 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
+          + "ON (t1.account_id,t1.container_id,t1.blob_name,t1.version) = (t2.account_id,t2.container_id,t2.blob_name,t2.version) "
+          + "WHERE t1.gg_modi_ts<(CURRENT_TIMESTAMP(6) - INTERVAL 60 DAY) and t2.account_id IS NULL and t2.container_id IS NULL and t2.blob_name IS NULL and t2.version IS NULL",
+      ACCOUNT_ID,
+      CONTAINER_ID,
+      BLOB_NAME,
+      BLOB_ID,
+      VERSION,
+      CURRENT_TIME,
+      NAMED_BLOBS_V2,
+      NAMED_BLOBS_V2);
```
justinlin-linkedin, line 153: Can you turn off the formatter and reformat this query statement just like list_query_v2?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -213,6 +232,9 @@ public Metrics(MetricRegistry metricRegistry) {
           MetricRegistry.name(MySqlNamedBlobDb.class, "NamedBlobPutTimeInMs"));
       namedBlobDeleteTimeInMs = metricRegistry.histogram(
           MetricRegistry.name(MySqlNamedBlobDb.class, "NamedBlobDeleteTimeInMs"));
+
+      namedBlobPullStaleTimeInMs = metricRegistry.histogram(
+          MetricRegistry.name(MySqlNamedBlobDb.class, "NamedBlobPullStaleTimeInMs"));
```
justinlin-linkedin, line 237: metric for cleanup?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -491,19 +585,43 @@ private DeleteResult run_delete_v2(String accountName, String containerName, Str
...
+        return resultList;
+      }
+    }
+  }
+
+  private void applySoftDelete(short accountId, short containerId, String blobName, long version, Timestamp currentTime,
+      Connection connection) throws Exception {
+    try (PreparedStatement statement = connection.prepareStatement(SOFT_DELETE_QUERY_V2)) {
+      // use the current time
+      statement.setTimestamp(1, currentTime);
```
SophieGuo410, line 616: I would suggest to renaming currentTime to deleteTs. Since this is a common method for different use cases, it's better to name it in a more generic way to avoid confusion.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -491,19 +585,43 @@ private DeleteResult run_delete_v2(String accountName, String containerName, Str
...
-      try (PreparedStatement statement = connection.prepareStatement(SOFT_DELETE_QUERY_V2)) {
-        // use the current time
-        statement.setTimestamp(1, currentTime);
-        statement.setInt(2, accountId);
-        statement.setInt(3, containerId);
-        statement.setString(4, blobName);
-        statement.setLong(5, version);
-        statement.executeUpdate();
-      }
+      applySoftDelete(accountId, containerId, blobName, version, currentTime, connection);
```
SophieGuo410, line 588: nit: I suggest probably we can rename deletionTime to originalDeletionTime and rename currentTime to currentDeleteTime. At the first glance, this currentTime is a little bit confused to me.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -491,19 +585,43 @@ private DeleteResult run_delete_v2(String accountName, String containerName, Str
...
+          resultList.add(result);
+        }
+        return resultList;
+      }
+    }
+  }
+
+  private void applySoftDelete(short accountId, short containerId, String blobName, long version, Timestamp currentTime,
+      Connection connection) throws Exception {
+    try (PreparedStatement statement = connection.prepareStatement(SOFT_DELETE_QUERY_V2)) {
```
SophieGuo410, line 614: nit: You don't have to use try logic since you throw the exception right away.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobsCleanupRunner.java
SophieGuo410: +1, we should add some constraint to stop the while loop in runPullStaleBlobIds in case there's a lot in one transaction.

ambry-api/src/main/java/com/github/ambry/named/StaleNamedResult.java
mikeyang923: Done

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobsCleanupRunner.java
mikeyang923: Updated with below changes:
1. Limit each staleResultList to be at most 1000 (done via the MySqlNamedBlobDb query side)
2. Catch up exception in each deleteBlob request, and remove those cases in the cleanupStaleData call later.
3. Update the scheduler to run every 10 minutes instead of once per day, to spread out the work more evenly.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Sure, done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Added new metric for cleanup timing

ambry-frontend/src/test/java/com/github/ambry/frontend/TestNamedBlobDb.java
```diff
@@ -148,27 +153,98 @@ public CompletableFuture<DeleteResult> delete(String accountName, String contain
...
     future.complete(new DeleteResult(recordWithDelete.getFirst().getBlobId(), alreadyDeleted));
     return future;
   }
 
+  @Override
+  public CompletableFuture<List<StaleNamedResult>> pullStaleBlobIds() {
+    CompletableFuture<List<StaleNamedResult>> future = new CompletableFuture<>();
+    List<StaleNamedResult> resultList = new ArrayList<>();
+    for (String accountName : allRecords.keySet()) {
+      Map<String, TreeMap<String, List<Pair<NamedBlobRecord, Pair<NamedBlobState, Long>>>>> map2 = allRecords.get(accountName);
```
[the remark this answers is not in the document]
mikeyang923: Good point, updated.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -190,6 +197,11 @@ public void start() throws InstantiationException {
     getAccountsHandler = new GetAccountsHandler(securityService, accountService, frontendMetrics);
     getStatsReportHandler = new GetStatsReportHandler(securityService, frontendMetrics, accountStatsStore);
     postAccountsHandler = new PostAccountsHandler(securityService, accountService, frontendConfig, frontendMetrics);
+
+    namedBlobsCleanupRunner = new NamedBlobsCleanupRunner(router, namedBlobDb);
+    namedBlobsCleanupScheduler = Utils.newScheduler(1, "named-blobs-cleanup-", false);
+    namedBlobsCleanupTask = namedBlobsCleanupScheduler.scheduleAtFixedRate(namedBlobsCleanupRunner, 60*60*1, 60*60*24, TimeUnit.SECONDS);
```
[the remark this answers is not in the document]
mikeyang923: I thought about this. I feel there is not much value in setting up control via config. And config sometimes make it harder to check what's actually doing on the host. Since it's a background job supposedly running with light load, a universal scheduling (every 10 minutes) is preferred.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done, thanks!

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: I am putting it into the try () so the resource can be closed/released no matter what: https://www.baeldung.com/java-try-with-resources

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
```diff
@@ -133,6 +133,14 @@ public void testQueryExecutionFailure() throws Exception {
         e -> Assert.assertEquals(sqlException, e.getCause()));
   }
 
+  @Test
+  public void testPullAndCleanStaleNamedBlobs() throws Exception {
+    dataSourceFactory.setLocalDatacenter(localDatacenter);
+    dataSourceFactory.triggerEmptyResultSetForLocalDataCenter(datacenters);
+    List<StaleNamedResult> staleNamedResults = namedBlobDb.pullStaleBlobIds().get();
+    namedBlobDb.cleanupStaleData(staleNamedResults);
+  }
```
justinlin-linkedin, line 142: this test is weird, it doesn't check the state of database. Can you write a bunch of stale data to the database and then verify pullStaleBlodIds get those data out. And can you also verify that after cleanupStateData, pullStaleBlobIds won't pull them out again.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -135,6 +135,30 @@ class MySqlNamedBlobDb implements NamedBlobDb {
   private static final String SOFT_DELETE_QUERY_V2 =
       String.format("UPDATE %s SET %s = ? WHERE %s", NAMED_BLOBS_V2, DELETED_TS, PK_MATCH_VERSION);
 
+  /**
+   * Pull the stale blobs that need to be cleaned up
+   */
+  private static final String GET_STALE_QUERY = String.format(""
```
justinlin-linkedin, line 141: you probably have to turn off formatter with //@formatter:off and turn it back on after this statement.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -135,6 +135,30 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
+          + "SELECT %s, %s, %s, %s, %s, %s "
+          + "FROM %s t1 "
+          + "LEFT JOIN "
+          + "(SELECT account_id, container_id, blob_name, max(version) as version "
+          + "FROM %s "
+          + "WHERE blob_state=1 "
+          + "      GROUP BY account_id, container_id, blob_name) t2 "
+          + "ON (t1.account_id,t1.container_id,t1.blob_name,t1.version) = (t2.account_id,t2.container_id,t2.blob_name,t2.version) "
+          + "WHERE t1.gg_modi_ts<(%s - INTERVAL 60 DAY) and t2.account_id IS NULL and t2.container_id IS NULL and t2.blob_name IS NULL and t2.version IS NULL "
+          + "LIMIT 1000",
```
justinlin-linkedin, line 151: I suggest we make limit configurable. So we don't have to release another version when this number is too big.

justinlin-linkedin: could you also add comment to this query to indicate what you are trying to do with this query. Basically what data will be returned by this query? Any version rather than the max version? Any version that has blob_state=1?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
SophieGuo410, line 141: Can you please add more comments to this query, this query is pretty complex.

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
mikeyang923: Added the proper tests in MySqlNamedBlobDbIntegrationTest in latest commit [1594fb9](https://github.com/linkedin/ambry/pull/2327/commits/1594fb9f2aa93ecac7c1d4f7a2b3109d196ec289)

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Sounds good, done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Good points. Updated for both: 1. make limit value driven by config, 2. Added comment to the query

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Sure, added.

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
```diff
@@ -6,9 +6,11 @@ CREATE TABLE IF NOT EXISTS named_blobs_v2 (
     version bigint NOT NULL,
     blob_id varbinary(50) NOT NULL,
     deleted_ts datetime(6) DEFAULT NULL,
+    gg_modi_ts datetime(6) DEFAULT (CURRENT_TIMESTAMP(6) - INTERVAL 61 DAY),
```
SophieGuo410, line 9: QQ, May I know why do we made this change? Will this affect the GG replication since it leverage this value to identify the replication priority. And can you please remind me why we don't have gg_modi_ts and gg_status in the schema before? And the "GG triggers" like what we have in "database/AMBRYBLOBS/schema/ddl/00-table.sql". I remember MYSQL team told me that we have to have this in schema to enable GG replication, but it looks like currently we don't have it but the GG replication are still been enabled.

mikeyang923: @SophieGuo410 you are right, we need this 'gg_modi_ts' field so GG replication can work. What we did is provided the table creation schema to sql team (and create a RB for internal source), then the sql team will apply that change. So the internal and external kind of going out sync for this schema. Even though it doesn't impact the deployed service. I am adding it back so they are in sync.

SophieGuo410: I see, so in order to keep it in sync I guess we need the gg_status and the GG trigger condition as well right? Just like what we have in "database/AMBRYBLOBS/schema/ddl/00-table.sql"

mikeyang923: I will add the gg_status. The GG replication setting needs some work at mysql team side as well. So we can't copy the triggers over. Otherwise it will get error eg ```Unknown column 'gg_status' in 'NEW'```

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
```diff
@@ -79,6 +85,7 @@ public MySqlNamedBlobDbIntegrationTest() throws Exception {
   public void testPutGetListDeleteSequence() throws Exception {
     int blobsPerContainer = 5;
 
+    time.setCurrentMilliseconds(System.currentTimeMillis());
```
SophieGuo410, line 88: I kind of feel that it's not scalable that people have to update the currentTime each time when we access the DB. Is there any way to avoid this?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -135,6 +142,40 @@ class MySqlNamedBlobDb implements NamedBlobDb {
   private static final String SOFT_DELETE_QUERY_V2 =
       String.format("UPDATE %s SET %s = ? WHERE %s", NAMED_BLOBS_V2, DELETED_TS, PK_MATCH_VERSION);
 
+  /**
+   * Pull the stale blobs that need to be cleaned up
```
SophieGuo410, line 146: Thanks for adding the comments, I'm wondering is there any easy way to filter the staled named blob by only use where? Left join is always used to join two tables.

SophieGuo410 (commented):
lgtm overall. just left two comments.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
mikeyang923: We only set this in the testing code. In actual service code, we don't need to update the currentTime at all. Here we are updating the MockTime's currentTime to do the time travel to mimic and test out the actual use case.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Yeah, I thought about this, and can't think a better way than current query. I timed the query and it perform well. And since it's a background batch job (not used in user requests), it should be ok

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -52,6 +54,22 @@ public class MySqlNamedBlobDbConfig {
...
+  @Config(STALE_MAX_RESULTS)
+  @Default("1000")
+  public final int staleMaxResults;
+
+  /**
+   * The maximum number of days for a stale blob to say uncleaned.
+   */
+  @Config(STALE_MAX_DAYS)
+  @Default("5")
+  public final int staleMaxDays;
```
justinlin-linkedin, line 69: nit: Honestly I think the variable names and the config names should be more descriptive. Something like
`query.stale.data.max.results` and `stale.data.retention.day`

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -484,35 +613,66 @@ private DeleteResult run_delete_v2(String accountName, String containerName, Str
...
-      }
+      applySoftDelete(accountId, containerId, blobName, version, currentDeleteTime, connection);
     }
     return new DeleteResult(blobId, alreadyDeleted);
   }
 
+  private List<StaleNamedResult> runPullStaleBlobs(final Connection connection) throws Exception {
+    try (PreparedStatement statement = connection.prepareStatement(GET_STALE_QUERY)) {
+      Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+      calendar.add(Calendar.DATE, -config.staleMaxDays);
```
justinlin-linkedin, line 631: Calendar.DAY_OF_MONTH?

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
```diff
@@ -5,10 +5,13 @@ CREATE TABLE IF NOT EXISTS named_blobs_v2 (
     blob_name varchar(191) NOT NULL,
     version bigint NOT NULL,
     blob_id varbinary(50) NOT NULL,
-    deleted_ts datetime(6) DEFAULT NULL,
     blob_state smallint NOT NULL,
+    deleted_ts datetime(6) DEFAULT NULL,
+    gg_modi_ts datetime(6) DEFAULT CURRENT_TIMESTAMP(6),
+    gg_status varchar(1) DEFAULT 'o',
```
justinlin-linkedin, line 11: we don't need this, do we.

ambry-api/src/main/java/com/github/ambry/named/StaleNamedResult.java
```diff
@@ -0,0 +1,112 @@
...
+package com.github.ambry.named;
+
+import java.sql.Timestamp;
+import java.util.Objects;
+
+
+/**
+ * Class to convey information about a successful stale named blob pulling {@link NamedBlobDb}.
+ */
+public class StaleNamedResult {
```
justinlin-linkedin, line 24: nit: can you rename this to StaleNamedBlob

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
justinlin-linkedin: you can create a MockTime with initial value as System.currentTimeMillis().

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
```diff
@@ -127,10 +139,12 @@ public void testPutGetListDeleteSequence() throws Exception {
 
     // delete the records and check that they cannot be fetched with a get call.
     for (NamedBlobRecord record : records) {
+      time.setCurrentMilliseconds(System.currentTimeMillis());
       DeleteResult deleteResult =
           namedBlobDb.delete(record.getAccountName(), record.getContainerName(), record.getBlobName()).get();
       assertEquals("Unexpected deleted ID", record.getBlobId(), deleteResult.getBlobId());
       assertFalse("Unexpected alreadyDeleted value", deleteResult.isAlreadyDeleted());
+      time.setCurrentMilliseconds(System.currentTimeMillis());
```
justinlin-linkedin, line 147: why do you set timestamp in the same method? Once you set it, it will not change.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
```diff
@@ -210,6 +230,79 @@ public void testExpiredBlobs() throws Exception {
         namedBlobDb.get(account.getName(), container.getName(), blobName).get());
   }
 
+  /**
+   * Test behavior with stale blob cleanup
+   */
+  @Test
+  public void testCleanupStaleBlobs() throws Exception {
+    Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
+    calendar.add(Calendar.MONTH, -2);
```
justinlin-linkedin, line 239: this should be in the configuration. It's better to get the retention time from the configuration so that you it's the same in the named blob db.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
```diff
@@ -210,6 +230,79 @@ public void testExpiredBlobs() throws Exception {
...
+    long twoMonthAgo = calendar.getTimeInMillis();
+    long twoMonthAgoABitLater = twoMonthAgo + 1;
```
justinlin-linkedin, line 241: please add more time to this. 1 millisecond is too short sometimes. could you make it 1 minute.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -135,6 +142,40 @@ class MySqlNamedBlobDb implements NamedBlobDb {
   private static final String SOFT_DELETE_QUERY_V2 =
       String.format("UPDATE %s SET %s = ? WHERE %s", NAMED_BLOBS_V2, DELETED_TS, PK_MATCH_VERSION);
 
+  /**
+   * Pull the stale blobs that need to be cleaned up
+   * It will pull out any stale record (limit to be config.staleMaxResults [Default 1000] records at most) meeting below conditions:
+   * 1. created more than 60 days ago, and
```
justinlin-linkedin, line 148: comments should be updated.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
```diff
@@ -210,6 +230,79 @@ public void testExpiredBlobs() throws Exception {
...
+    int staleCount = 10;
+    int needCleanupCount = 0;
+    List<NamedBlobRecord> records = new ArrayList<>();
+
+    // Create stale named blob records
+    for (int i = 0; i < staleCount; i++) {
+      String blobId = getBlobId(account, container);
+      String blobName = "stale/" + i + "/more path segments--";
+      long expirationTime = i % 2 == 0 ? Utils.Infinite_Time : twoMonthAgoABitLater;
+      NamedBlobRecord record =
```
justinlin-linkedin, line 255: we are creating stale named blob for Update API for now. These should be considered stale. Assume we give 5 days of retention.
Case 1. created 10 days ago, InProgress, is Largest (Don't care about the ttl and delete)
Case 2. created 10 days ago, Ready, not Largest and bigger version is ready

And these should considered good
Case 1, created 2 days ago, InProgress, is Largest (Don't care about ttl and delete)
Case 2. created 2 days ago, Ready, not largest, and bigger version is ready
Case 3. created 10 days ago, Ready, is Largest
Case 4. created 10 days ago, Ready, not Largest, but bigger version is InProgress.


Can you at least write all those cases and make sure that are treated as expected.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
mikeyang923: Done.

ambry-api/src/main/java/com/github/ambry/config/FrontendConfig.java
```diff
@@ -73,6 +73,13 @@ public class FrontendConfig {
   @Default("24 * 60 * 60")
   public final long optionsValiditySeconds;
 
+  /**
+   * Whether to enable named blob stale data cleanup process
+   */
+  @Config("named.blob.cleanup.enabled")
```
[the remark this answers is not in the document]
mikeyang923: Done

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
mikeyang923: Done

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: I see DAY_OF_MONTH is the same thing as DATE. And I feel 'DAY_OF_MONTH' is a bit confusion here since we really want a unit of DAYS. Not which day of the month.

ambry-named-mysql/src/main/resources/NamedBlobsSchema.ddl
mikeyang923: Removed

ambry-api/src/main/java/com/github/ambry/named/StaleNamedResult.java
mikeyang923: Done

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
mikeyang923: Bc we need to move the time along. As you mentioned, the time once is set, it will NOT change. That will a problem. Since we are testing the operations per sequence of time. If we don't set here, it's as if we are doing all those operations at the SAME time, which will fail.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
mikeyang923: Done

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
mikeyang923: That's the beauty of math, here 1 millisecond is enough. I updated at other places when needed and renamed the variables

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
mikeyang923: Added.

SophieGuo410 (approved):

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
justinlin-linkedin: if you want to make sure that the time moved, you should use `time.sleep` method. otherwise, there is no guarantee that these two `System.currentTimeMillis()` won't return the same result. Those statements can be executed within the same millisecond.

justinlin-linkedin (approved):

PR #2329, opened 2023-01-21
ReplicaThread: Fix error handling logic and wrong data node id in the log messages

justinlin-linkedin:
This PR fixes a bug in the error handing log in nonblocking network client.

When there is a network error in Response for ReplicaMetadataRequest, the exchangeMetadataResponseList for each RemoteReplicaGroup would be null, and a null can't be added to an ArrayList when trying to add all the exchangeMetadataResponseList to the map exchangeMetadateaResponsesInEachCycle. This PR fixed it.

This PR also fix the datanodeid error in log messages.


JingQianCloud (approved):
LGTM

Arun-LinkedIn (approved):

PR #2331, opened 2023-01-25
FE: Add retry logic to router for get blob

justinlin-linkedin:
This PR adds retry logic for GetBlobOperation in router.

Why are we doing this:
Router might fail client's get requests due to temporary network error, or server being overloaded. If frontend returns 503 right away, client would would probably retry the entire blob. However, if this is a composite blob made of hundreds of chunks, then client resending get request would end up wasting lots of resources to refetch those already fetched blob in router.

This PR adds retry logic to prevent resource from wasting. It will retry on GetBlobOperation. But there is a limit to which an GetBlobOperation can retry. This limit is based on both count and duration. We can configure the maximum number of retries and the longest duration for retries. A retry can't exceed either limit.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -656,6 +656,25 @@ public class RouterConfig {
   @Default("false")
   public final boolean routerRepairWithReplicateBlobEnabled;
 
+  /**
+   * The maximum duration in seconds to retry. If the get blob operation takes more than this duration, we would not retry.
+   */
+  @Config(ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC)
+  @Default("0")
+  public final int routerGetBlobRetryLimitInSec;
+  public static final String ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC = "router.get.blob.retry.limit.in.sec";
+  public static final int ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC_MAX = 30;
+
+  /**
+   * The maximum number of retries for get blob. The default value is 0, which means no retry for get blob.
+   * The maximum number for this is 10. We don't want to retry more than 10 times.
+   */
+  @Config(ROUTER_GET_BLOB_RETRY_LIMIT_COUNT)
+  @Default("0")
+  public final int routerGetBlobRetryLimitCount;
+  public static final String ROUTER_GET_BLOB_RETRY_LIMIT_COUNT = "router.get.blob.retry.limit.count";
+  public static final int ROUTER_GET_BLOB_RETRY_LIMIT_COUNT_MAX = 10;
+
   // Group compression-related configs in the CompressConfig class.
   private final CompressionConfig compressionConfig;
 
@@ -798,10 +817,15 @@ public class RouterConfig {
         verifiableProperties.getBoolean(ROUTER_UNAVAILABLE_DUE_TO_SUCCESS_COUNT_IS_NON_ZERO_FOR_DELETE, true);
     routerNotFoundCacheTtlInMs = verifiableProperties.getLongInRange(ROUTER_NOT_FOUND_CACHE_TTL_IN_MS, 15 * 1000L, 0,
         ROUTER_NOT_FOUND_CACHE_MAX_TTL_IN_MS);
-    routerUpdateOpMetadataRelianceTimestampInMs = verifiableProperties.getLong(
-        ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS, DEFAULT_ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS);
+    routerUpdateOpMetadataRelianceTimestampInMs =
+        verifiableProperties.getLong(ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS,
+            DEFAULT_ROUTER_UPDATE_OP_METADATA_RELIANCE_TIMESTAMP_IN_MS);
     routerRepairWithReplicateBlobEnabled =
         verifiableProperties.getBoolean(ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED, false);
+    routerGetBlobRetryLimitInSec = verifiableProperties.getIntInRange(ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC, 0, 0,
+        ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC_MAX);
+    routerGetBlobRetryLimitCount = verifiableProperties.getIntInRange(ROUTER_GET_BLOB_RETRY_LIMIT_COUNT, 0, 0,
+        ROUTER_GET_BLOB_RETRY_LIMIT_COUNT_MAX);
 
     compressionConfig = new CompressionConfig(verifiableProperties);
   }
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -663,17 +663,16 @@ public class RouterConfig {
   @Default("0")
   public final int routerGetBlobRetryLimitInSec;
   public static final String ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC = "router.get.blob.retry.limit.in.sec";
-  public static final int ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC_MAX = 30;
+  public static final int ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC_MAX = 300;
 
   /**
    * The maximum number of retries for get blob. The default value is 0, which means no retry for get blob.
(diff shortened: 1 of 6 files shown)


mikeyang923 (approved):

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1434,6 +1433,7 @@ void checkAndMaybeComplete() {
             logger.trace("{}: Attempt to put chunk with id: {} failed, attempting slipped put ", loggingContext,
                 chunkBlobId);
             routerMetrics.slippedPutAttemptCount.inc();
+            chunkException = null;
```
JingQianCloud, line 1436: Do we have issue here? Looks like either way, chunkException will be reset before the PutBlob retry, right?
Somehow for PutBlob, I like the current implementation more since we consolidate all the initial setting all in  prepareForSending.
But your call.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -1038,6 +1038,143 @@ public void testNetworkClientTimeoutAllFailure() throws Exception {
...
+      Assert.assertNull("Operation should succeed now", exceptionRef.get());
+      Assert.assertNotNull("Operation should succeed now", resultRef.get());
+      resultRef.get().getBlobDataChannel().close();
+      assertEquals(1, routerMetrics.getBlobRetryCount.getCount() - retryCountBefore);
+    } finally {
+      routerConfig = originalConfig;
+    }
+  }
+
   /**
```
JingQianCloud, line 1178: Can we add unit test for composite blob and test retry on multiple chunks?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -792,7 +795,22 @@ void initialize(int index, CompositeBlobInfo.ChunkMetadata chunkMetadata) {
...
+    void resetForRetry() {
+      logger.trace("BlobId {}: Retry for chunk Id: {}, failed attempts: {}, initializedTime: {}", blobId, chunkBlobId,
+          failedAttempts, initializedTimeMs);
+      chunkException = null;
+      chunkOperationTracker = getOperationTracker(chunkBlobId.getPartition(), chunkBlobId.getDatacenterId(),
+          RouterOperation.GetBlobOperation, chunkBlobId);
+      progressTracker = new ProgressTracker(chunkOperationTracker);
+      state = ChunkState.Ready;
+      failedAttempts++;
+      routerMetrics.getBlobRetryCount.inc();
```
JingQianCloud, line 813: If for some reason, the servers returns the server unavailable very quickly and there are not many replicas online for this partition, our frontend retry may be too stressful to these remaining replicas.

I was thinking about some simple backoff.
But we cannot do 
"sleep(10ms * failedAttempts);"
It'll block the RequestResponseHandlerThread which is shared by different Operations.

Do we have some other way for some simple backoff? Probably record the time and when it reaches that time + delay time, start the next retry. Not sure if the code will be too complicated.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -656,6 +656,24 @@ public class RouterConfig {
   @Default("false")
   public final boolean routerRepairWithReplicateBlobEnabled;
 
+  /**
+   * The maximum duration in seconds to retry. If the get blob operation takes more than this duration, we would not retry.
+   */
+  @Config(ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC)
+  @Default("0")
```
snalli, line 663: lets get rid of the Defaults. they are not used anywhere and are jsut confusing

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -656,6 +656,24 @@ public class RouterConfig {
...
+  public static final int ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC_MAX = 300;
+
+  /**
+   * The maximum number of retries for get blob. The default value is 0, which means no retry for get blob.
+   */
+  @Config(ROUTER_GET_BLOB_RETRY_LIMIT_COUNT)
+  @Default("0")
+  public final int routerGetBlobRetryLimitCount;
+  public static final String ROUTER_GET_BLOB_RETRY_LIMIT_COUNT = "router.get.blob.retry.limit.count";
+  public static final int ROUTER_GET_BLOB_RETRY_LIMIT_COUNT_MAX = 100;
```
snalli, line 675: 100 seems high

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1000,6 +1018,35 @@ private void fetchRequests(RequestRegistrationCallback<GetOperation> requestRegi
...
+     */
+    boolean shouldRetry(RouterException exception) {
+      if (exception == null) {
+        return false;
+      }
+      switch (exception.getErrorCode()) {
+        case AmbryUnavailable:
+        case OperationTimedOut:
+        case UnexpectedInternalError:
+          return failedAttempts < routerConfig.routerGetBlobRetryLimitCount
```
snalli, line 1042: Nice ! We are retrying with num failed attempts and a timeout

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1000,6 +1018,35 @@ private void fetchRequests(RequestRegistrationCallback<GetOperation> requestRegi
...
+     *   </ol>
+     * </p>
+     * @param exception The exception for this chunk operation.
+     * @return {@code True} to retry this chunk. otherwise, return false.
+     */
+    boolean shouldRetry(RouterException exception) {
+      if (exception == null) {
+        return false;
+      }
+      switch (exception.getErrorCode()) {
```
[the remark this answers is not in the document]
justinlin-linkedin: the method processServerError should do it. Basically, the RouterException for Retry_After_Backoff is UnexpectedInternalError for short.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
justinlin-linkedin: right, it's just prepareForSending should not reset the chunkException. The exception should be reset in the checkAndMaybeComplete method. It's easier to understand that way.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
justinlin-linkedin: will do

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
justinlin-linkedin: kinda hard to do this, we have to add another configuration for backoff, the time to sleep etc... we will be introducing too many new configurations.

JingQianCloud (approved):

snalli:
can u pls address my comments before merging ?

snalli:
can u pls address my comments before merging ?

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
ankagrawal: +1 Maybe 3 or 5 is good enough?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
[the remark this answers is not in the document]
ankagrawal: Should all types of `UnexepectedInternalError` lead to retry? Like if somehow the server returns `BlobDeleted` for data chunks, then we should not retry them here.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1000,6 +1018,35 @@ private void fetchRequests(RequestRegistrationCallback<GetOperation> requestRegi
...
+        case AmbryUnavailable:
+        case OperationTimedOut:
+        case UnexpectedInternalError:
+          return failedAttempts < routerConfig.routerGetBlobRetryLimitCount
+              && time.milliseconds() < initializedTimeMs + TimeUnit.SECONDS.toMillis(
+              routerConfig.routerGetBlobRetryLimitInSec);
```
ankagrawal, line 1044: nit: Maybe calculate `TimeUnit.SECONDS.toMillis(routerConfig.routerGetBlobRetryLimitInSec)` once in the constructor.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
justinlin-linkedin: this is the maximum number, I don't see we are setting the value to this number. But I don't want to limit it under 100.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
justinlin-linkedin: done

PR #2333, opened 2023-01-25
Update logic for sending 404 and 503

Arun-LinkedIn:
This PR refactors OperationTracker code to use originating data center as the source of truth when sending 404 responses. 

1. If all the replicas in originating data center are up and return that blob is not present, we send 404. Config `routerOperationTrackerTerminateOnNotFoundEnabled` is used to terminate the operation immediately. 
2. If any of the originating DC replicas are not up (i.e. in leader or standby state) and the success count doesn't meet the quorum, we send 503 if there are no other errors with higher precedence (such as Authorization failure, Blob deleted, etc).

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -127,6 +127,8 @@ public class RouterConfig {
   public static final String ROUTER_UNAVAILABLE_DUE_TO_SUCCESS_COUNT_IS_NON_ZERO_FOR_DELETE =
       "router.unavailable.due.to.success.count.is.non.zero.for.delete";
   public static final String ROUTER_REPAIR_WITH_REPLICATE_BLOB_ENABLED = "router.repair.with.replicate.blob.enabled";
+  public static final String ROUTER_OPERATION_TRACKER_CHECK_ALL_ORIGINATING_REPLICAS_FOR_NOT_FOUND =
+      "router.operation.tracker.check.all.originating.replicas.for.not.found";
 
   /**
    * Number of independent scaling units for the router.
@@ -672,6 +674,16 @@ public class RouterConfig {
   public static final String ROUTER_GET_BLOB_RETRY_LIMIT_COUNT = "router.get.blob.retry.limit.count";
   public static final int ROUTER_GET_BLOB_RETRY_LIMIT_COUNT_MAX = 100;
 
+  /*
+   * If this config is set to {@code true} the operation tracker would make sure all replicas in originating data center
+   * are up and respond with BLOB_NOT_FOUND before concluding that blob is not present. Else, it would check in
+   * total_originating_dc_replicas - put_success_target + 1 replicas.
+   */
+  @Config(ROUTER_OPERATION_TRACKER_CHECK_ALL_ORIGINATING_REPLICAS_FOR_NOT_FOUND)
+  @Default("true")
+  public final boolean routerOperationTrackerCheckAllOriginatingReplicasForNotFound;
+
+
   // Group compression-related configs in the CompressConfig class.
   private final CompressionConfig compressionConfig;
 
@@ -825,6 +837,8 @@ public class RouterConfig {
         ROUTER_GET_BLOB_RETRY_LIMIT_COUNT_MAX);
 
     compressionConfig = new CompressionConfig(verifiableProperties);
+    routerOperationTrackerCheckAllOriginatingReplicasForNotFound =
+        verifiableProperties.getBoolean(ROUTER_OPERATION_TRACKER_CHECK_ALL_ORIGINATING_REPLICAS_FOR_NOT_FOUND, true);
   }
 
   /**
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -1056,12 +1056,14 @@ class GetBlobOperation extends GetOperation {
       if (progressTracker.isDone()) {
         if (progressTracker.hasSucceeded() && !retainChunkExceptionOnSuccess) {
           chunkException = null;
-        } else if (chunkOperationTracker.maybeFailedDueToOfflineReplicas()) {
-          chunkException =
-              buildChunkException("Get Chunk failed because of offline replicas", RouterErrorCode.AmbryUnavailable);
-        } else if (chunkOperationTracker.hasFailedOnNotFound()) {
-          chunkException =
-              buildChunkException("Get Chunk failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist);
+        } else {
+          if (chunkOperationTracker.hasFailedOnNotFound()) {
+            chunkException =
+                buildChunkException("Get Chunk failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist);
+          } else if (chunkOperationTracker.hasSomeUnavailability()) {
+            setChunkException(chunkException =
+                buildChunkException("Get Chunk failed because of offline replicas", RouterErrorCode.AmbryUnavailable));
+          }
         }
         if (shouldRetry(chunkException)) {
           resetForRetry();
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -22,11 +22,13 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
+import java.util.Set;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 import org.slf4j.Logger;
@@ -68,13 +70,13 @@ import org.slf4j.LoggerFactory;
  */
 class SimpleOperationTracker implements OperationTracker {
   private static final Logger logger = LoggerFactory.getLogger(SimpleOperationTracker.class);
+  private static final int NOT_FOUND_THRESHOLD_INVALID_VALUE = -1;
   protected final String datacenterName;
   protected final String originatingDcName;
   protected final int replicaSuccessTarget;
   protected final int replicaParallelism;
   // How many NotFound responses from originating dc will terminate the operation.
-  // It is set to tolerate one random failure in the originating dc if all other responses are not found.
-  protected final int originatingDcNotFoundFailureThreshold;
+  protected int originatingDcNotFoundFailureThreshold = NOT_FOUND_THRESHOLD_INVALID_VALUE;
   protected final int totalReplicaCount;
   protected final LinkedList<ReplicaId> replicaPool = new LinkedList<>();
   protected final NonBlockingRouterMetrics routerMetrics;
@@ -82,7 +84,6 @@ class SimpleOperationTracker implements OperationTracker {
   private final RouterOperation routerOperation;
   private final PartitionId partitionId;
   private final RouterConfig routerConfig;
-  private final boolean crossColoEnabled;
   protected int inflightCount = 0;
   protected int replicaSuccessCount = 0;
   protected List<ReplicaId> successReplica = new ArrayList<>();
@@ -91,17 +92,11 @@ class SimpleOperationTracker implements OperationTracker {
   protected int disabledCount = 0;
   protected int originatingDcNotFoundCount = 0;
   protected int totalNotFoundCount = 0;
-  protected int diskDownCount = 0;
   protected ReplicaId lastReturnedByIterator = null;
   private Iterator<ReplicaId> replicaIterator;
-  private String reassignedOriginDc = null;
-  private int originatingDcOfflineReplicaCount = 0;
-  private int originatingDcEligibleReplicaCount = 0;
-  private int originatingDcTotalReplicaCount = 0;
-  private int totalOfflineReplicaCount = 0;
-  private int allReplicaCount = 0;
+  private final Set<ReplicaId> originatingDcLeaderOrStandbyReplicas;
+  private final int originatingDcTotalReplicaCount;
   private final Map<ReplicaState, List<ReplicaId>> allDcReplicasByState;
-  private final List<ReplicaId> allReplicas;
   private final BlobId blobId;
 
   /**
@@ -195,12 +190,11 @@ class SimpleOperationTracker implements OperationTracker {
         (Map<ReplicaState, List<ReplicaId>>) partitionId.getReplicaIdsByStates(EnumSet.allOf(ReplicaState.class), null);
     List<ReplicaId> eligibleReplicas;
     List<ReplicaId> offlineReplicas = new ArrayList<>();
-    totalOfflineReplicaCount =
-        getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE)).getOrDefault(ReplicaState.OFFLINE,
-            Collections.emptyList()).size();
-    allReplicas = partitionId.getReplicaIds().stream().collect(Collectors.toList());
-    allReplicaCount = allReplicas.size();
+    // This is a list of all replicas for this partition. This comes from replica information updated by us in helix
+    // Data node configs via bootstrap tool and not external view. This would be same as ideal view.
+    List<ReplicaId> allReplicas = new ArrayList<>(partitionId.getReplicaIds());
 
+    boolean crossColoEnabled;
     switch (routerOperation) {
       case GetBlobOperation:
       case GetBlobInfoOperation:
@@ -278,6 +272,7 @@ class SimpleOperationTracker implements OperationTracker {
     // rate of cross-colo requests(GET/DELETE/TTLUpdate), operation tracker should be allowed to try remote dc with most
     // replicas first. This is useful in cluster with "unbalanced" replica distribution (i.e. 3 replicas in local dc and
     // 1 replica per remote dc)
+    String reassignedOriginDc = null;
     if (originatingDcName == null && routerConfig.routerCrossColoRequestToDcWithMostReplicas) {
       Map<String, Long> dcToReplicaCnt = replicas.stream()
           .collect(Collectors.groupingBy(e -> e.getDataNodeId().getDatacenterName(), Collectors.counting()));
@@ -350,16 +345,6 @@ class SimpleOperationTracker implements OperationTracker {
       }
     }
     totalReplicaCount = replicaPool.size();
-    originatingDcOfflineReplicaCount =
-        getReplicasByState(originatingDcName, EnumSet.of(ReplicaState.OFFLINE)).values().size();
-    originatingDcEligibleReplicaCount = replicaPool.stream()
-        .filter(replicaId -> replicaId.getDataNodeId().getDatacenterName().equals(this.originatingDcName))
-        .collect(Collectors.toList())
-        .size();
-    originatingDcTotalReplicaCount = allReplicas.stream()
-        .filter(replicaId -> replicaId.getDataNodeId().getDatacenterName().equals(this.originatingDcName))
-        .collect(Collectors.toList())
-        .size();
 
     // MockPartitionId.getReplicaIds() is returning a shared reference which may cause race condition.
     // Please report the test failure if you run into this exception.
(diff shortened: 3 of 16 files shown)


ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -419,115 +402,28 @@ public boolean hasSucceeded() {
...
 
-    // For Delete and TtlUpdate
-    // We mark the failure as due to offline replicas when we know that we couldn't find the blob in eligible replicas,
-    // and remaining replicas are enough to make the request successful, and there is atleast one offline replica in the
-    // remaining set of replicas. The offline replicas can come back up in future to make the request successful, and
-    // hence such a failure should be deemed as retryable.
-    if (hasFailedOnOriginatingDcNotFound()
-        && originatingDcTotalReplicaCount - originatingDcNotFoundCount >= replicaSuccessTarget
-        && originatingDcOfflineReplicaCount > 0) {
+    if (originatingDcNotFoundCount >= originatingDcNotFoundFailureThreshold) {
```
justinlin-linkedin, line 415: let's add another condition before this, as long as there is a success response, we return False here.
There is some corner case here.
If we don't check all replicas in originating dc, then the threshold here is 2.
When we upload a blob to originating DC, assume there is only replica 1 and 2 get the blob and 3 didn't get it. And when we try to delete this blob, 2 become bootstrap. Now we have 1 as leader, 2 as bootstrap and 3 as standby. We have 2 replicas as leader or standby so the threshold is not -1. However we will get not found from 2 and 3. This way we would return NOT_FOUND,even when we get success from 1.

justinlin-linkedin: or we should always check all the replicas in originating DC but this way, we would return more 503 than we should.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -419,115 +402,28 @@ public boolean hasSucceeded() {
...
-                + "originatingDcEligibleReplicaCount {} {} "
-                + "replicasByState = {}",
-            routerOperation, partitionId, crossColoEnabled, diskDownCount, totalNotFoundCount, totalReplicaCount,
-            replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, originatingDcEligibleReplicaCount,
-            getBlobIdLog(), allDcReplicasByState);
-        routerMetrics.failedMaybeDueToTotalOfflineReplicasCount.inc();
-        return true;
-      }
-      return false;
+    if (originatingDcNotFoundFailureThreshold == NOT_FOUND_THRESHOLD_INVALID_VALUE) {
```
justinlin-linkedin, line 410: Threshold being valid doesn't mean the result should be inconclusive. There could be a case where the operation is terminated before any of the request is sent out, like quota rejection. In this case, we have no NOT_FOUND error returned from server and the threshold could be -1 as well. In this way, we should return false.

I guess it's better in this way
```
if (originatingDcNotFoundFailureThreshold != NOT_FOUND_THRESHOLD_INVALID_ALUE) {
    if (originatingDcNotFoundCount >= originatingDcNotFoundFailureThreshold) {
        return True;
    }
    return False;
} else {
    if (originatingDcNotFoundCount == availableReplicasInOrginatingDC) {
        return INCONCLUSIVE; // This means even if all the available replicas returns NOT_FOUND, we still can't say it's not FOUND.
    }
    return False; // As long as there is no replica not returning NOT_FOUND
}
```

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -195,12 +192,9 @@ class SimpleOperationTracker implements OperationTracker {
         (Map<ReplicaState, List<ReplicaId>>) partitionId.getReplicaIdsByStates(EnumSet.allOf(ReplicaState.class), null);
     List<ReplicaId> eligibleReplicas;
     List<ReplicaId> offlineReplicas = new ArrayList<>();
-    totalOfflineReplicaCount =
-        getReplicasByState(null, EnumSet.of(ReplicaState.OFFLINE)).getOrDefault(ReplicaState.OFFLINE,
-            Collections.emptyList()).size();
-    allReplicas = partitionId.getReplicaIds().stream().collect(Collectors.toList());
-    allReplicaCount = allReplicas.size();
+    List<ReplicaId> allReplicas = new ArrayList<>(partitionId.getReplicaIds());
```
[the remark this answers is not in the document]
justinlin-linkedin: it doesn't necessarily have 9 replicas, in video, it only has 5.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: Justin, I think the difference between your solution and Arun's solution is the case when "originatingDcNotFoundFailureThreshold == NOT_FOUND_THRESHOLD_INVALID_ALUE".

In Arun's implementation, the INCONCLUSIVE stands for there is either some offline or bootstrapping originating replicas. Although for the offline replica, we didn't send any request but we should consider there is some "virtual offline status" from these replicas not existing in the external view. 
Because of that, in the Operation code, it's a "setOperationException(AmbryUnavailable)" to compensate the missing offline we should have but we didn't receive. It's not operationException.set(AmbryUnavailable) which will overwrite the final status.
 If there is any higher precedence error, the high precedence error still take the precedence and won't change anything. So in your case, it'll still return quota_rejected.


We don't have maybeFailedDueToOfflineReplicas any more in the checkAndMaybeComplete. So if we change to below, in the case for PutBlob, that we have 1 originating replica returns success and 8 replicas are offline, not sure what the status we will return. 
    if (originatingDcNotFoundCount == availableReplicasInOrginatingDC) {
        return INCONCLUSIVE; // This means even if all the available replicas returns NOT_FOUND, we still can't say it's not FOUND.
    }

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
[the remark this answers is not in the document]
JingQianCloud: Yes, just want to mention it's the ideal view.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
justinlin-linkedin: I don't think this is true. Check out the DeleteOperation::checkAndMaybeComplete method. when it's quota rejected, and not enough leader and standby in originating dc, it will return 503. But we want it to return  429.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
Arun-LinkedIn: Good point. Thank you for mentioning the case. Updated the code.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -656,6 +658,14 @@ public class RouterConfig {
   @Default("false")
   public final boolean routerRepairWithReplicateBlobEnabled;
 
+  /**
+   * If this config is set to {@code true} the operation tracker would check all replicas in originating data center
+   * before concluding that blob is not present. Else, it would check in total_replicas - put_success_target + 1 replicas.
+   */
+  @Config(ROUTER_OPERATION_TRACKER_CHECK_ALL_REPLICAS_FOR_NOT_FOUND)
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Updated the name to make it less confusing.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Added comments to mention it is from data node configs and is same as ideal view.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
Arun-LinkedIn: Yeah, actually like Jing was mentioning, I actually set 503 `setOperationException` so that errors which have more precedence are not replaced. Please let me know if this is okay.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -419,115 +402,28 @@ public boolean hasSucceeded() {
...
-                + "replicasByState = {}",
-            routerOperation, partitionId, crossColoEnabled, diskDownCount, totalNotFoundCount, totalReplicaCount,
-            replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount, originatingDcEligibleReplicaCount,
-            getBlobIdLog(), allDcReplicasByState);
-        routerMetrics.failedMaybeDueToTotalOfflineReplicasCount.inc();
-        return true;
-      }
-      return false;
+    if (originatingDcNotFoundFailureThreshold == NOT_FOUND_THRESHOLD_INVALID_VALUE) {
+      // Since some originating DC replicas are down or in bootstrap state, we can't conclude if blob is missing.
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Added.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: I think probably Justin brought a generic question which always exist, 
In the checkAndMaybeComplete function, for NOT_FOUND, if hasFailedOnNotFound is true, we set the exception to NOT_FOUND no matter what the previous responses we ever received are. 

```
if (operationTracker.hasFailedOnNotFound()) {
    operationException.set(BlobDoesNotExist))
}
```

In Justin's case, replica 1 returns success status. 
What if replica 1 returns BlobAuthorizationFailure/BlobExpired etc? We are supposed to return these high precedence errors but we return NOT_FOUND.

Probably something like

```
if (operationTracker.hasFailedOnNotFound() && no_exception_received_higher_than_AmbryUnavailable) {
    operationException.set(BlobDoesNotExist))
}
```

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: For me, I kind of feel it's easier to understand the code logic when we have two separated function, hasFailedOnNotFound and hasOfflineReplicas.
FailedOnNotFoundResult.INCONCLUSIVE basically means we have offline replica(or bootstrapping). We can use standalone hasOfflineReplicas() function. 
When hasOfflineReplicas is true and  hasFailedOnNotFound is false, we just fake one unavailable exception. But we honor the exception precedence and honor all the response we received.
Just want to know your opinion of it. 


```
if (operationTracker.hasFailedOnNotFound()) {
    if (originatingDcNotFoundFailureThreshold == NOT_FOUND_THRESHOLD_INVALID_VALUE) {
        operationException.set(RouterErrorCode.BlobDoesNotExist));
} else if (operationTracker.hasOfflineReplica()) {
      // compensate the offline status we should have but we don't offline is not in the clustermap
       setOperationException(RouterErrorCode.AmbryUnavailable):
}
```

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -349,42 +348,31 @@ class SimpleOperationTracker implements OperationTracker {
...
-      // 2. We need enough active replicas in the originating DC
-      // How many active replicas is considered as enough? We have to consider some extreme cases.
-      // When the put success target is 2, we at least need 2 replicas in active state so we know the replicas are up to
-      // date. If we have only one replica in active state, we might run into a scenario that we are getting NotFound
-      // from the other two inactive replicas and we should not ignore the active one.
-      originatingDcNotFoundFailureThreshold = originatingDcTotalReplicaCount - routerConfig.routerPutSuccessTarget + 1;
-    } else {
-      originatingDcNotFoundFailureThreshold = 0;
+    originatingDcTotalReplicaCount = (int) allReplicas.stream()
+        .filter(replicaId -> replicaId.getDataNodeId().getDatacenterName().equals(this.originatingDcName))
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. Since it seems like .equals(null) would return false, `originatingDcTotalReplicaCount` would be 0. Added check that `originatingDcNotFoundFailureThreshold` should be > 0 in `hasFailedNotFound`.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -576,10 +476,12 @@ public void onResponse(ReplicaId replicaId, TrackedRequestFinalState trackedRequ
         if (trackedRequestFinalState == TrackedRequestFinalState.NOT_FOUND) {
           totalNotFoundCount++;
           if (replicaId.getDataNodeId().getDatacenterName().equals(originatingDcName)) {
-            originatingDcNotFoundCount++;
+            if (originatingDcLeaderOrStandbyReplicas.contains(replicaId)) {
+              // Since bootstrap replicas could still be catching up, treat NotFound responses only from leader or
+              // standby replicas as valid.
+              originatingDcNotFoundCount++;
```
Arun-LinkedIn, line 482: Thanks @justinlin-linkedin for pointing offline that we should increment not found only if they are from Originating DC.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -344,28 +348,29 @@ public enum FailedOnNotFoundResult {TRUE, FALSE, INCONCLUSIVE}
...
-            : originatingDcTotalReplicaCount - routerConfig.routerPutSuccessTarget + 1;
-    if (getEligibleReplicas(originatingDcName, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER)).size()
-        < originatingDcNotFoundFailureThreshold) {
-      // If sufficient replicas are not in leader or stand by state, we cannot conclude that blob is missing. Reset the
-      // NotFoundFailureThreshold to invalid value.
-      originatingDcNotFoundFailureThreshold = NOT_FOUND_THRESHOLD_INVALID_VALUE;
+    int minNumOriginatingDcReplicasToSearch =
+        routerConfig.routerOperationTrackerCheckAllOriginatingReplicasUpForNotFound ? originatingDcTotalReplicaCount
+            : Math.max(originatingDcTotalReplicaCount - routerConfig.routerPutSuccessTarget + 1, 0);
+    if (minNumOriginatingDcReplicasToSearch > 0
```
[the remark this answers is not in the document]
Arun-LinkedIn: I think we could have a case where originatingDC might be null. For example, we had a data center in Singapore earlier which was decommissioned. If we try to fetch a blob from that time, the originatingDC could be null. In that case, this count might be 0. This is probably very corner case though.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -674,12 +556,15 @@ public boolean hasFailed() {
       return totalReplicaCount - failedCount < Math.max(totalReplicaCount - 1,
           routerConfig.routerPutSuccessTarget + disabledCount);
     } else {
-      // if there is no possible way to use the remaining replicas to meet the success target,
-      // deem the operation a failure.
-      if (replicaInPoolOrFlightCount + replicaSuccessCount < replicaSuccessTarget) {
+      // If blob is not found in originating DC, we can terminate the operation early. Adding
+      // routerConfig.routerOperationTrackerTerminateOnNotFoundEnabled as a safe guard to search replicas in all colos
+      // before terminating the operation.
+      if (routerConfig.routerOperationTrackerTerminateOnNotFoundEnabled && hasFailedOnNotFound()) {
```
JingQianCloud, line 562: Do we plan to enable routerOperationTrackerTerminateOnNotFoundEnabled on delivery?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -419,115 +398,36 @@ public boolean hasSucceeded() {
...
-          "Terminating {} on {} due to disk down count and total Not_Found. CrossColoEnabled: {}, DiskDownCount: {},"
-              + "TotalNotFoundCount: {}, TotalReplicaCount: {}, replicaSuccessTarget: {}, OfflineReplicaCount: {},"
-              + "allReplicaCount: {} {} replicasByState = {}", routerOperation, partitionId, crossColoEnabled, diskDownCount,
-          totalNotFoundCount, totalReplicaCount, replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount,
-          getBlobIdLog(), allDcReplicasByState);
-      routerMetrics.failedOnTotalNotFoundCount.inc();
-      return true;
-    }
-    return false;
+  public boolean hasFailedOnUnavailability() {
```
JingQianCloud, line 426: I think we need change this name "hasFailedOnUnavailability".
The operation is not necessary failed on unavailability. Just we have some replicas are un-available. 
Does "hasSomeUnavailability" or something like that less confusing?

Three originating replicas, replica 1 is leader replica and returns deleted, replica 2 is standby and returns deleted, replica 3 is bootstrap, when we enable check_all_originating_replica, this function will return true.
But we don't fail on unavailability. The error status should be deleted.

Also we may have an issue in the GetOperation.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1007,12 +1007,16 @@ void checkAndMaybeComplete() {
...
         if (progressTracker.hasSucceeded() && !retainChunkExceptionOnSuccess) {
           chunkException = null;
-        } else if (chunkOperationTracker.maybeFailedDueToOfflineReplicas()) {
-          chunkException =
-              buildChunkException("Get Chunk failed because of offline replicas", RouterErrorCode.AmbryUnavailable);
-        } else if (chunkOperationTracker.hasFailedOnNotFound()){
-          chunkException =
-              buildChunkException("Get Chunk failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist);
+        } else {
+          if (chunkOperationTracker.hasFailedOnNotFound()) {
```
justinlin-linkedin, line 1011: GetBlobInfoOperation should have the same logic.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -419,115 +398,36 @@ public boolean hasSucceeded() {
...
+    // If sufficient number of originating DC replicas are not available in leader or standby state, it means Ambry is
+    // unavailable.
+    return originatingDcName != null
+        && originatingDcLeaderOrStandbyReplicas.size() < originatingDcNotFoundFailureThreshold;
```
justinlin-linkedin, line 430: please also check if there is any success count here.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1007,12 +1007,16 @@ void checkAndMaybeComplete() {
...
+            // If operation failed due to blob not being present, send 404.
+            chunkException =
+                buildChunkException("Get Chunk failed because of BlobNotFound", RouterErrorCode.BlobDoesNotExist);
+          } else if (chunkOperationTracker.hasFailedOnUnavailability()) {
+            // If some originating DC replicas are down, and we can't conclude if blob is missing, send 503.
+            chunkException =
```
JingQianCloud, line 1017: hasFailedOnUnavailability only means there is some replica unavailable, not necessary mean we fail on unavailable.
Seems chunkException is overwriting the previous exception. It'll overwrite "deleted" or "expired" status.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
Arun-LinkedIn: Thank you for pointing it. Added the same logic.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
Arun-LinkedIn: Yeah, was thinking that we could enable it.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
Arun-LinkedIn: Sure. Changed the name to `hasSomeUnavailability` and updated to call `setChunkException` instead of `exception = e `for unavailability error.

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -419,112 +398,38 @@ public boolean hasSucceeded() {
...
-          totalNotFoundCount, totalReplicaCount, replicaSuccessTarget, totalOfflineReplicaCount, allReplicaCount,
-          getBlobIdLog(), allDcReplicasByState);
-      routerMetrics.failedOnTotalNotFoundCount.inc();
+  public boolean hasSomeUnavailability() {
+    if (originatingDcName != null && originatingDcLeaderOrStandbyReplicas.size() < originatingDcTotalReplicaCount) {
+      logger.info("{} on {} has some replicas unavailable in originating DC {}, TotalReplicaCount: {},"
+              + "LeaderStandbyCount: {}, ReplicasByState: {}, {}", routerOperation.name(), partitionId, originatingDcName,
+          originatingDcTotalReplicaCount, originatingDcLeaderOrStandbyReplicas.size(), allDcReplicasByState,
+          getBlobIdLog());
+      routerMetrics.failedMaybeDueToUnavailableReplicasCount.inc();
```
JingQianCloud, line 432: This definition of this metrics is different from before now.

JingQianCloud (approved):
Can we add test cases to cover all the corner cases we discussed in the previous PRs if you haven't added them yet?
Thank you.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
Arun-LinkedIn: Sure. As we discussed offline, instead of checking success count, I updated `hasSomeUnavailability()` to return true if there is any replica in originating DC which is not in leader and standby state. Also added comments in java doc that this method can be used when operation has failed.

PR #2336, opened 2023-01-29
Update clustermap and state transition factory to deal with duplicate partition ids

justinlin-linkedin:
Ambry is currently going through the process of reconstructing resource names. We used to assign partition ids to several resources, even if those partition ids belong to the same clique. Now we are trying to gather all the partition ids of the same clique to one resource.

Since in server we honor new resource over the old resource, we have to do the same thing in frontend. This is easy since we can compare the resource names and know that higher number resource name would win.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
@@ -26,6 +26,38 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 
+/**
+ * Ambry is currently going through the process of reconstructing resources. We used to assign partition ids to several
+ * resources, even if those partition ids belong to the same clique. Now we are trying to gather all the partition ids
+ * of the same clique to one resource.
+ *
+ * In doing so, we should inevitably 1. duplicate all the partition ids with a different resource 2. remove the old
+ * resources.  The issue is that resource doesn't really matter in Ambry. Resource is a structure required by Helix.
+ * Ambry has a flatter structure. This structure starts with a cluster. Under this cluster, it has partitions. If we
+ * duplicate all the partition ids with different resources, we would force these partitions to go through state
+ * transition twice. And by removing the old resource, we would effectively force all the partitions to go through
+ * downward transition to DROPPED state. We have to make sure that frontend and server would be able to deal with both
+ * situations.
+ *
+ * Current resource names are all numeric numbers starting from 0. The new resource names will be numeric numbers as
+ * well, but it will be starting from a much higher number, like 1000. So we can easily find the new resource by just
+ * comparing the resource names.
+ *
+ * When we eventually remove old resources, all the partition ids under those resourced would have to go through state
+ * transition to become DROPPED in the end. A dropped partition means all the data will be removed, we don't want any
+ * of the partition to remove data. So in state transition, we have to make sure after we add new resources, the state
+ * transition messages from the old resources would be ignored and won't take effect anymore.
+ *
+ * Since in server we honor new resource over the old resource, we have to do the same thing in the clustermap.
+ *
+ * We have three implementations of {@link PartitionStateChangeListener}, StorageManager's implementation, ReplicationManager's
+ * implementation and StatsManager's implementation. Most of the those implementations are fine with going through state
+ * transition multiple times and generate the same result, except for ReplicationManager. When transitioning from standby
+ * to leader and from leader to standby, replication manager would change the leader pair. If different replicas are chosen
+ * to be leader in different transition, then we would have multiple leaders. But this is fine, since replication manager
+ * use this leader pair information for cross-col replication. The worst case is that we have more than one replicas doing
+ * cross-colo data replication.
+ */
 @StateModelInfo(initialState = "OFFLINE", states = {"BOOTSTRAP", "LEADER", "STANDBY", "INACTIVE"})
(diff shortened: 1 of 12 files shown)


justinlin-linkedin:
But this is one corner cases that we have to deal with.
When we add new resources to server, all the partitions would go through state transition all over again, from offline to bootstrap to standby to leader. There are two issues to deal with.
1. Will second state transition cause any trouble?
2. Second state transition might leader instances to different state, like standby instance would become leader or leader instance would become standby, will this be a problem.

I check the three partition state transition implementation from BlobStats, StorageManger and ReplicationManager. 
For BlobStats, it's ok since it only care about if the replica is in active mode or not.
For storage manager, there are more steps, but those steps are idempotent.
For ReplicationManager, there is a difference.

snalli:
Is this related to ATT ?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
```diff
@@ -25,104 +26,188 @@
...
   @Transition(to = "BOOTSTRAP", from = "OFFLINE")
   public void onBecomeBootstrapFromOffline(Message message, NotificationContext context) {
-    logger.info("Partition {} in resource {} is becoming BOOTSTRAP from OFFLINE", message.getPartitionName(),
-        message.getResourceName());
-    if (clusterMapConfig.clustermapEnableStateModelListener) {
-      partitionStateChangeListener.onPartitionBecomeBootstrapFromOffline(message.getPartitionName());
+    boolean shouldTransition = shouldTransition(message);
+    String partitionName = message.getPartitionName();
+    logger.info("Partition {} in resource {} is becoming BOOTSTRAP from OFFLINE, Should Transition? {}",
+        message.getPartitionName(), message.getResourceName(), shouldTransition);
```
SophieGuo410, line 87: nit: could use partitionName instead of message.getPartitionName().

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
```diff
@@ -25,104 +26,188 @@
...
+ * transition messages from the old resources would be ignored and won't take effect anymore.
+ *
+ * Since in server we honor new resource over the old resource, we have to do the same thing in the clustermap.
+ *
+ * We have three implementations of {@link PartitionStateChangeListener}, StorageManager's implementation, ReplicationManager's
+ * implementation and StatsManager's implementation. Most of the those implementations are fine with going through state
+ * transition multiple times and generate the same result, except for ReplicationManager. When transitioning from standby
+ * to leader and from leader to standby, replication manager would change the leader pair. If different replicas are chosen
+ * to be leader in different transition, then we would have multiple leaders. But this is fine, since replication manager
+ * use this leader pair information for cross-col replication. The worst case is that we have more than one replicas doing
```
SophieGuo410, line 58: nit: cross-colo

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
```diff
@@ -25,104 +26,188 @@
...
+      if (v == null || v.equals(resourceName)) {
+        return resourceName;
+      }
+      try {
+        int oldResourceId = Integer.valueOf(v);
+        int newResourceId = Integer.valueOf(resourceName);
+        return newResourceId > oldResourceId ? resourceName : v;
+      } catch (Exception e) {
+        logger.error("Failed to parse resource name to an integer", e);
+        throw new StateTransitionException("Failed to parse resource name",
```
SophieGuo410, line 207: You might want to add the resourceName to see which resouceName failed to be parsed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1152,19 +1155,53 @@ private void updatePartitionResourcesMappingFromExternalView(Collection<External
         logger.warn(
             "Partition to resource mapping for individual dc (non-aggregated view) would be built from ideal states");
       } else {
+        Map<String, Map<String, String>> partitionToDCToResourceNames = new HashMap<>();
```
SophieGuo410, line 1158: Just curious why we switch the ConcurrentHashMap to HashMap?

SophieGuo410 (approved):
lgtm! Just left some nit-picking.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: I mistakenly used ConcurrentHashMap earlier here. I think HashMap is probably sufficient here since we are serially adding all entries to it.

Arun-LinkedIn (approved):
lgtm!

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
justinlin-linkedin: The exception is NumberFormatException, which already have the input.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
justinlin-linkedin: I will keep it this way and update it later.

PR #2337, opened 2023-01-30
Fix statsManager config names

snalli:
Caused by: java.lang.NoSuchFieldError: outputFilePath
        at com.github.ambry.server.StatsManager.<init>(StatsManager.java:107) ~[ambry-server-0.3.962.jar:?]
        at com.github.ambry.server.AmbryServer.startup(AmbryServer.java:274) ~[ambry-server-0.3.962.jar:?]
        ... 28 more

Encountered this error during testing. Not sure how the server has been deployed so far. But the config variables in open source code do not match the closed-source version. This patch fixes that.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
@@ -37,14 +37,14 @@ public class StatsManagerConfig {
    */
   @Config(STATS_OUTPUT_FILE_PATH)
   @Default("/tmp/stats_output.json")
-  public final String outputFilePath;
+  public final String statsOutputFilePath;
 
   /**
    * The time period in seconds that configures how often stats are published.
    */
   @Config(STATS_PUBLISH_PERIOD_IN_SECS)
   @Default("7200")
-  public final long publishPeriodInSecs;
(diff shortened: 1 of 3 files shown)


mikeyang923 (approved):

ambry-api/src/main/java/com/github/ambry/config/StatsManagerConfig.java
```diff
@@ -37,70 +37,70 @@ public class StatsManagerConfig {
    */
   @Config(STATS_OUTPUT_FILE_PATH)
   @Default("/tmp/stats_output.json")
-  public final String outputFilePath;
+  public final String statsOutputFilePath;
```
SophieGuo410, line 40: The config name in close source should match with "stats.enable.mysql.report" and right now we should enable it. What you changed is just the variable name and used in open source only. I feel like this won't cause the issue.

snalli: Tested this and changing the variable name fixed the issue. I was able to deploy the server on DR node.

snalli: Will rebase and retest

PR #2338, opened 2023-01-30
Fix thread leak in unit tests

justinlin-linkedin:
There are some thread leak in our unit tests. Most of them are coming from zookeeper client. The reason why zookeeper client leaks threads is that we are not closing zookeeper client every time we create a PropertyStore with methods in CommonUtils class. 

In these methods, we create a ZkBaseDataAccessor and then pass this accessor to HelixPropertyStore. When we stop HelixPropertyStore, it doesn't close the ZkBaseDataAccessor since HelixPropertyStore thinks that ZkBaseDataAccessor is owned by something else. 

We have to return both HelixPropertyStore and ZkBaseDataAccessor to the caller of these methods and let the caller close both objects.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/commons/Notifier.java
+++ b/ambry-api/src/main/java/com/github/ambry/commons/Notifier.java
@@ -47,4 +47,10 @@ public interface Notifier<T> {
    * @param listener The {@link TopicListener} who unsubscribes the topic.
    */
   public void unsubscribe(String topic, TopicListener<T> listener);
+
+  /**
+   * Close this notifier.
+   */
+  default void close() {
+  }
 }
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
@@ -17,6 +17,7 @@ import com.github.ambry.commons.CommonUtils;
 import com.github.ambry.config.ClusterMapConfig;
 import com.github.ambry.config.HelixPropertyStoreConfig;
 import com.github.ambry.config.VerifiableProperties;
+import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Utils;
 import java.io.File;
 import java.io.IOException;
@@ -46,6 +47,7 @@ import org.apache.helix.HelixAdmin;
 import org.apache.helix.manager.zk.ZKHelixAdmin;
 import org.apache.helix.manager.zk.ZKUtil;
 import org.apache.helix.manager.zk.ZNRecordSerializer;
+import org.apache.helix.manager.zk.ZkBaseDataAccessor;
 import org.apache.helix.model.IdealState;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.model.LeaderStandbySMD;
@@ -236,6 +238,7 @@ public class HelixBootstrapUpgradeUtil {
           clusterMapToHelixMapper.clusterName);
     }
     clusterMapToHelixMapper.updateClusterMapInHelix(startValidatingClusterManager);
+
     if (!dryRun) {
       clusterMapToHelixMapper.validateAndClose();
     }
@@ -285,7 +288,8 @@ public class HelixBootstrapUpgradeUtil {
       String clusterNamePrefix, String dcs, String stateModelDef) throws Exception {
     HelixBootstrapUpgradeUtil clusterMapToHelixMapper =
         new HelixBootstrapUpgradeUtil(hardwareLayoutPath, partitionLayoutPath, zkLayoutPath, clusterNamePrefix, dcs,
-            DEFAULT_MAX_PARTITIONS_PER_RESOURCE, false, false, null, stateModelDef, null, null, null, null, null, false);
+            DEFAULT_MAX_PARTITIONS_PER_RESOURCE, false, false, null, stateModelDef, null, null, null, null, null,
+            false);
     clusterMapToHelixMapper.addStateModelDef();
     info("State model def is successfully added");
   }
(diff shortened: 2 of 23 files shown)


ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
```diff
@@ -15,50 +15,60 @@
...
+    ZkBaseDataAccessor<ZNRecord> baseDataAccessor =
+        new ZkBaseDataAccessor.Builder<ZNRecord>().setZkClientType(ZkClientType.DEDICATED)
+            .setZkAddress(zkServers)
+            .setRealmAwareZkClientConfig(new RealmAwareZkClient.RealmAwareZkClientConfig().setConnectInitTimeout(
+                propertyStoreConfig.zkClientConnectionTimeoutMs).setZkSerializer(new ZNRecordSerializer()))
+            .setRealmAwareZkConnectionConfig(
+                new RealmAwareZkClient.RealmAwareZkConnectionConfig.Builder().setSessionTimeout(
+                    propertyStoreConfig.zkClientSessionTimeoutMs).build())
+            .build();
+    return new Pair<>(new ZkHelixPropertyStore<>(baseDataAccessor, propertyStoreConfig.rootPath, subscribedPaths),
```
SophieGuo410, line 52: QQ: Should helix be responsible for closing the baseDataAccessor inside their stop() method since they have baseDataAccessor in HelixPropertyStore?

SophieGuo410 (commented):
thanks for fixing the thread leak issue for our test. lgtm, just left one question regarding this change.

ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
justinlin-linkedin: I've tried a couple different ways of fixing this, unfortunately, this seems to be the only way out.

The reason why ZkHelixPropertyStore doesn't close the zookeeper client that we create in the same method, is that ZkHelixPropertyStore doesn't think it owns zookeeper client. We are passing a BaseDataAccessor to ZkHelixPropertyStore, ZkHelixPropertyStore thinks BaseDataAccessor owns the zookeeper client, so it will let BaseDataAccessor close the zookeeper client.

Well, can we make ZkHelixPropertyStore create zookeeper client internally so it would own the zookeeper client? Unfortunately, we can't. Through there are 3 more constructors for ZkHelixPropertyStore that would let it create a zookeeper client internally, none of those 3 constructors meets our requirements. They either fail to let us specify ZkClientType, or watched paths. Unless we use a shared zookeeper client, or we don't cache those znodes, we can't use current constructors. If helix can expose more constructors to enable more flexible object creation, that would solve the issue.

Then, can we just use ZkCacheDataAccessor? It seems to implement HelixPropertyStore and has a rich constructor set. Unfortunately, we can't. Because some HelixManager, as a common interface, is expose method to return concrete class ZkHelixPropertyStore instead of a more general interface HelixPropertyStore. That locks us with ZkHelixPropertyStore instead of more general interface HelixPropertyStore.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -927,6 +936,7 @@ private void updateClusterMapInHelix(boolean startValidatingClusterManager) thro
           info(
               "[{}] Done adding all instances in {}, now scanning resources in Helix and ensuring instance set for partitions are the same.",
               dc.getName().toUpperCase(), dc.getName());
+
```
snalli, line 939: rm new line

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -1693,7 +1709,8 @@ private void verifyDataNodeAndDiskEquivalencyInDc(Datacenter dc, String clusterN
         DataNode dataNode = (DataNode) dataNodeId;
         String instanceName = getInstanceName(dataNode);
         ensureOrThrow(allInstancesInHelix.remove(instanceName), "Instance not present in Helix " + instanceName);
-        DataNodeConfig dataNodeConfig = getDataNodeConfigFromHelix(dcName, instanceName, propertyStoreAdapter, instanceConfigConverter);
+        DataNodeConfig dataNodeConfig =
+            getDataNodeConfigFromHelix(dcName, instanceName, propertyStoreAdapter, instanceConfigConverter);
```
snalli, line 1713: can u pls reduce these changes ? these are a bit distracting from the real changes.

ambry-api/src/main/java/com/github/ambry/commons/Notifier.java
```diff
@@ -47,4 +47,10 @@
    * @param listener The {@link TopicListener} who unsubscribes the topic.
    */
   public void unsubscribe(String topic, TopicListener<T> listener);
+
+  /**
+   * Close this notifier.
+   */
+  default void close() {
+  }
```
snalli, line 55: why is this empty ?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -236,6 +238,7 @@ public static void bootstrapOrUpgrade(String hardwareLayoutPath, String partitio
           clusterMapToHelixMapper.clusterName);
     }
     clusterMapToHelixMapper.updateClusterMapInHelix(startValidatingClusterManager);
+
```
snalli, line 241: if u cud please minimize these newlines and formatting chagnes, then i can focus on the actual changes.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -721,6 +727,7 @@ private void uploadClusterAdminInfos(Map<String, Map<String, Map<String, String>
         }
       } finally {
         helixPropertyStore.stop();
+        pair.getSecond().close();
```
Arun-LinkedIn, line 730: nit: May be use temporary variable ZkBaseDataAccessor for readability?

Arun-LinkedIn (approved):

Arun-LinkedIn (commented):
Thank you for fixing the thread leak issue!

justinlin-linkedin:
A better solution is proposed here https://github.com/linkedin/ambry/pull/2339

PR #2339, opened 2023-01-31
Fix zk client thread leak in unit test

justinlin-linkedin:
There are some thread leak in our unit tests. Most of them are coming from zookeeper client. The reason why zookeeper client leaks threads is that we are not closing zookeeper client every time we create a ZkHelixPropertyStore with methods in CommonUtils class.

In these methods, we create a ZkBaseDataAccessor and then pass this accessor to HelixPropertyStore. When we stop HelixPropertyStore, it doesn't close the ZkBaseDataAccessor since HelixPropertyStore thinks that ZkBaseDataAccessor is owned by something else.

We have use different constructors from ZkHelixPropertyStore so that the zk client is created internally in the PropertyStore, so that when we call ZkHelixPropertyStore.stop, it will close the zk client for us.

However, by doing so, we are choosing to create this zk client with a shared connection, instead of a dedicated connection. A shared connection would potentially damage our zk client's performance but it should be fine. Since we only use this shared connection in HelixNotifier and PropertyStoreDataNodeConfigAdapter. We are not using it much.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
@@ -306,6 +306,9 @@ public class HelixAccountService extends AbstractAccountService {
         shutDownExecutorService(scheduler, config.updaterShutDownTimeoutMs, TimeUnit.MILLISECONDS);
       }
       helixStore.stop();
+      if (notifier != null) {
+        notifier.close();
+      }
     }
   }
 
--- a/ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
+++ b/ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
@@ -35,13 +35,11 @@ public class CommonUtils {
     if (zkServers == null || zkServers.isEmpty() || propertyStoreConfig == null) {
       throw new IllegalArgumentException("Invalid arguments, cannot create HelixPropertyStore");
     }
-    return new ZkHelixPropertyStore<ZNRecord>(zkServers, new ZNRecordSerializer(), propertyStoreConfig.rootPath,
-        subscribedPaths);
+    return createHelixPropertyStore(zkServers, propertyStoreConfig.rootPath, subscribedPaths);
   }
 
   /**
    * Create an instance of {@link HelixPropertyStore}, using non-deprecated APIs.
-   * TODO replace usages of the other method with this one once verified to be stable.
    * @param zkAddress the ZooKeeper server address.
    * @param rootPath the root path for this {@link HelixPropertyStore}.
    * @param subscribedPaths the paths to subscribe to for change notification. Note that these should be absolute paths,
--- a/ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
+++ b/ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
@@ -16,13 +16,10 @@ package com.github.ambry.commons;
 
 import com.github.ambry.config.HelixPropertyStoreConfig;
 import java.util.List;
-import org.apache.helix.manager.zk.ZNRecordSerializer;
-import org.apache.helix.manager.zk.ZkBaseDataAccessor;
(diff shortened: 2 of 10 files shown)


SophieGuo410 (approved):

ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
```diff
@@ -38,9 +35,7 @@ public static HelixPropertyStore<ZNRecord> createHelixPropertyStore(String zkSer
     if (zkServers == null || zkServers.isEmpty() || propertyStoreConfig == null) {
       throw new IllegalArgumentException("Invalid arguments, cannot create HelixPropertyStore");
     }
-    ZkClient zkClient = new ZkClient(zkServers, propertyStoreConfig.zkClientSessionTimeoutMs,
-        propertyStoreConfig.zkClientConnectionTimeoutMs, new ZNRecordSerializer());
-    return new ZkHelixPropertyStore<>(new ZkBaseDataAccessor<>(zkClient), propertyStoreConfig.rootPath,
+    return new ZkHelixPropertyStore<ZNRecord>(zkServers, new ZNRecordSerializer(), propertyStoreConfig.rootPath,
```
snalli, line 38: Overall lgtm, but can this line be replaced with this since the code in the two methods is same:

>     return createHelixPropertyStore(zkServers, propertyStoreConfig.rootPath, subscribedPaths);

ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
```diff
@@ -55,10 +50,6 @@ public static HelixPropertyStore<ZNRecord> createHelixPropertyStore(String zkSer
    */
   public static HelixPropertyStore<ZNRecord> createHelixPropertyStore(String zkAddress, String rootPath,
       List<String> subscribedPaths) {
```
snalli, line 52: And a check here

>     if (zkAddress == null || zkAddress.isEmpty() || rootPath == null) {
      throw new IllegalArgumentException("Invalid arguments, cannot create HelixPropertyStore");
    }

ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
Arun-LinkedIn, line 38: Nit: May be we could remove `zkClientConnectionTimeoutMs` and `zkClientSessionTimeoutMs` in `HelixPropertyStoreConfig` since we no longer create zkClient ourselves?

ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
justinlin-linkedin: Chroot path can be null.

Arun-LinkedIn (approved):

ambry-commons/src/main/java/com/github/ambry/commons/CommonUtils.java
snalli: can be or cannot be ?

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
```diff
@@ -306,6 +306,9 @@ public void close() {
         shutDownExecutorService(scheduler, config.updaterShutDownTimeoutMs, TimeUnit.MILLISECONDS);
       }
       helixStore.stop();
+      if (notifier != null) {
```
snalli, line 309: does calling stop on helixStore twice have any issues ?
the stop at L308 may be removed since we are closing it in the notifier

justinlin-linkedin:
> How did you notice the thread leak? Does unit test report the thread leak?

Unit test was running really slow and hangs sometimes. I run it in Intellij and it shows lots of zookeeper client threads.

ambry-account/src/main/java/com/github/ambry/account/HelixAccountService.java
justinlin-linkedin: notifier has a different helixstore.

PR #2344, opened 2023-02-02
Advertise DR partitions as they DR partitions

snalli:
When DR tries to replicate from servers, the servers reject the requests because they do not recognize DR as a peer.


ankagrawal (approved):
lgtm

PR #2345, opened 2023-02-02
Fix several NPEs in ReplicaThread

justinlin-linkedin:
After enabling Nonblocking network client we found some NPEs in ReplicaThread.
1. In filterTimedoutRequests method, when the group doesn't exist, we try to call group.getId() and that's NPE.
2. In handleReplicaMetadataResponse method, when the response has an error, we try to process it anyway and that's NPE.
3. In handleGetResponse method, when the response has an error, we try to process it anyway and that's NPE.

This PR fixed all of them

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -632,12 +632,6 @@ public class ReplicaThread implements Runnable {
       DataNodeId remoteNode = replicasToReplicatePerNode.get(0).getReplicaId().getDataNodeId();
       ReplicaMetadataResponse response =
           getReplicaMetadataResponse(replicasToReplicatePerNode, connectedChannel, remoteNode);
-      if (response.getError() != ServerErrorCode.No_Error) {
-        // Getting error for entire response, something unexpected happened.
-        logger.error("Remote node: {} Thread name: {} ServerError for ReplicaMetadataResponse {}", remoteNode,
-            threadName, response.getError());
-        throw new ReplicationException("ReplicaMetadataResponse unexpected error " + response.getError());
-      }
       return handleReplicaMetadataResponse(response, replicasToReplicatePerNode, remoteNode);
     } catch (Exception e) {
       if (e instanceof ReplicationException) {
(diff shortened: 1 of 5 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -631,6 +632,12 @@ List<ExchangeMetadataResponse> exchangeMetadata(ConnectedChannel connectedChanne
       DataNodeId remoteNode = replicasToReplicatePerNode.get(0).getReplicaId().getDataNodeId();
       ReplicaMetadataResponse response =
           getReplicaMetadataResponse(replicasToReplicatePerNode, connectedChannel, remoteNode);
+      if (response.getError() != ServerErrorCode.No_Error) {
```
Arun-LinkedIn, line 635: nit: looks like we are already checking for this error in `getReplicaMetadataResponse`.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -808,6 +815,12 @@ protected void fixMissingStoreKeys(ConnectedChannel connectedChannel, List<Remot
           getMessagesForMissingKeys(connectedChannel, exchangeMetadataResponseList, replicasToReplicatePerNode,
               remoteNode, remoteColoGetRequestForStandby);
       if (getResponse != null) {
+        if (getResponse.getError() != ServerErrorCode.No_Error) {
```
Arun-LinkedIn, line 818: nit: Similar as above. Looks like we are already checking for this error in `getMessagesForMissingKeys`.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -2050,6 +2071,12 @@ public void handleGetResponse(ResponseInfo responseInfo) {
           // Deserialize the request from the given ResponseInfo
           DataInputStream dis = new NettyByteBufDataInputStream(responseInfo.content());
           GetResponse response = GetResponse.readFrom(dis, clusterMap);
+          if (response.getError() != ServerErrorCode.No_Error) {
+            // Getting error for entire response, something unexpected happened.
+            logger.error("Remote node: {} Thread name: {} RemoteReplicaGroup {} ServerError for GetResponse {}",
+                remoteDataNode, threadName, id, response.getError());
```
Arun-LinkedIn, line 2077: nit: Similar to error message in `getReplicaMetadataResponse`, would it be helpful if we print the list of remote replicas too.

Arun-LinkedIn (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
justinlin-linkedin: that's right, the blocking channel doesn't have this issue.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
justinlin-linkedin: the replicas will be printed out later

SophieGuo410 (approved):

PR #2346, opened 2023-02-06
ReplicateBlob on deletion: add the unit test for compacted blob.

JingQianCloud:
Add the unit test that the source blob is deleted and compacted.
ReplicateBlob will fail with Blob deleted.
The DeleteOperation fails with its original error.

### DIFF
--- a/ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
@@ -3158,7 +3158,7 @@ public class NonBlockingRouterTest extends NonBlockingRouterTestBase {
     serverErrors.add(ServerErrorCode.Blob_Deleted); // source PutBlob is compacted, replication fails with Blob_Deleted.
     for (MockServer server : layout.getMockServers()) {
       if (server.getDataCenter().equals(localDcName)) {
-        // local DC, return NO_ERROR for one replica,
+        // local DC, return NO_ERROR for one replica. This replica will be the source replica of the ReplicateBlob.
         if (sourceDataNode == null) {
           sourceDataNode = mockClusterMap.getDataNodeId(server.getHostName(), server.getHostPort());
         } else {
--- a/ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
@@ -3137,6 +3137,53 @@ public class NonBlockingRouterTest extends NonBlockingRouterTestBase {
   }
 
   /**
+   * Test Deletion first hits Not_FOUND. Then run ReplicateBlob.
+   * If the PutBlob is compacted already, ReplicateBlob may fail with ID_Deleted.
+   * @throws Exception
+   */
+  @Test
+  public void testReplicateBlobOnDeleteButPutBlobCompacted() throws Exception {
+    String serviceID = "delete-service";
+    MockServerLayout layout = new MockServerLayout(mockClusterMap);
+    String localDcName = "DC3";
+    setRouter(getNonBlockingRouterProperties(localDcName), layout, new LoggingNotificationSystem());
+    setOperationParams();
+    String blobId = router.putBlob(putBlobProperties, putUserMetadata, putChannel, new PutBlobOptionsBuilder().build())
+        .get(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+
+    // two local replicas are unavailable. Remote replica doesn't have the Blob.
+    DataNodeId sourceDataNode = null;
+    List<ServerErrorCode> serverErrors = new ArrayList<>();
+    serverErrors.add(ServerErrorCode.Blob_Not_Found); // return NOT_FOUND for the first Delete Request
+    serverErrors.add(ServerErrorCode.Blob_Deleted); // source PutBlob is compacted, replication fails with Blob_Deleted.
+    for (MockServer server : layout.getMockServers()) {
+      if (server.getDataCenter().equals(localDcName)) {
+        // local DC, return NO_ERROR for one replica,
+        if (sourceDataNode == null) {
+          sourceDataNode = mockClusterMap.getDataNodeId(server.getHostName(), server.getHostPort());
+        } else {
+          server.setServerErrorForAllRequests(ServerErrorCode.Replica_Unavailable);
+        }
+      } else {
+        server.setServerErrors(serverErrors);
+      }
+    }
+
+    // ReplicateBlob fails, it returns the error of the delete operation.
+    try {
+      router.deleteBlob(blobId, serviceID).get();
+      fail("deleteBlob should fail since ReplicateBlob fails.");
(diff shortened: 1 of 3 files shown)


snalli:
will take a look and get back

ankagrawal (approved):

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -3136,6 +3136,55 @@ public void testBlobDelete503NoRetrySinceNoSuccessReplica() throws Exception {
...
+    String serviceID = "delete-service";
+    MockServerLayout layout = new MockServerLayout(mockClusterMap);
+    String localDcName = "DC3";
+    setRouter(getNonBlockingRouterProperties(localDcName), layout, new LoggingNotificationSystem());
+    setOperationParams();
+    String blobId = router.putBlob(putBlobProperties, putUserMetadata, putChannel, new PutBlobOptionsBuilder().build())
+        .get(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+
+    // two local replicas are unavailable. Remote replica doesn't have the Blob.
+    DataNodeId sourceDataNode = null;
```
snalli, line 3155: sourceDataNode is not used except for assignment at L3163

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -3136,6 +3136,55 @@ public void testBlobDelete503NoRetrySinceNoSuccessReplica() throws Exception {
...
+      router.deleteBlob(blobId, serviceID).get();
+      fail("deleteBlob should fail since ReplicateBlob fails.");
+    } catch (Exception e) {
+      RouterException r = (RouterException) e.getCause();
+      Assert.assertEquals("AmbryUnavailable error is expected", RouterErrorCode.AmbryUnavailable, r.getErrorCode());
+    } finally {
+      layout.getMockServers().forEach(mockServer -> mockServer.setServerErrorForAllRequests(null));
+      router.getNotFoundCache().invalidateAll();
+      if (router != null) {
+        router.close();
```
snalli, line 3183: this null check is redundant because u r using router above and if it were null, then this test wouldn't run

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -3136,6 +3136,55 @@ public void testBlobDelete503NoRetrySinceNoSuccessReplica() throws Exception {
...
+    String localDcName = "DC3";
+    setRouter(getNonBlockingRouterProperties(localDcName), layout, new LoggingNotificationSystem());
+    setOperationParams();
+    String blobId = router.putBlob(putBlobProperties, putUserMetadata, putChannel, new PutBlobOptionsBuilder().build())
+        .get(AWAIT_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+
+    // two local replicas are unavailable. Remote replica doesn't have the Blob.
+    DataNodeId sourceDataNode = null;
+    List<ServerErrorCode> serverErrors = new ArrayList<>();
+    serverErrors.add(ServerErrorCode.Blob_Not_Found); // return NOT_FOUND for the first Delete Request
```
snalli, line 3157: so does the delete op fail with not_found or ambry_unavlb error ?

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
snalli: test passes even if i remove it

JingQianCloud: ReplicateBlob needs a source replica to replicate the Blob from. So at least we should have one replica available.

Removing the code won't have on-demand replication on-going.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
JingQianCloud: At server level, the single delete request fails with NOT_FOUND on the remote replica.
At frontend level, router checks all the responses from both the local and remote replica, the final status for deletion is  unavailable.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
JingQianCloud: removed.

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
snalli: sourceDataNode is a local variable here and not passed as an arg anywhere. How is replicateBlob using it internally ?

snalli: also, how do we know replicateBlob actually took place ?

snalli: oh i see, what u mean. By falling into the if-case u r ensuring not all servers throw an error.

snalli: so by removing it, the test passed because on-demand repl was not happening. Hmm, this is confusing. The test shd check if on-demand repl actually happened. Is there a way to do that ?

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
```diff
@@ -3136,6 +3136,53 @@ public void testBlobDelete503NoRetrySinceNoSuccessReplica() throws Exception {
...
+    serverErrors.add(ServerErrorCode.Blob_Deleted); // source PutBlob is compacted, replication fails with Blob_Deleted.
+    for (MockServer server : layout.getMockServers()) {
+      if (server.getDataCenter().equals(localDcName)) {
+        // local DC, return NO_ERROR for one replica,
+        if (sourceDataNode == null) {
```
snalli, line 3162: instead of sourceDataNode, I recommend boolean skipServer and comment explaining why we skip a server

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
JingQianCloud: Yes

ambry-router/src/test/java/com/github/ambry/router/NonBlockingRouterTest.java
JingQianCloud: For ReplicateBlob, we have the concept of "Source replicate". It's probably easier to understand with sourceDataNode name.
Also for replicationBlob test, we need specifically provide the sourceDataNode as below, 
// since there is only one local available, router will replicate the blob to at least 1 remote replicas.
router.replicateBlob(blobIdStr, serviceId, sourceDataNode).get();

justinlin-linkedin (approved):

PR #2349, opened 2023-02-07
[DR] Advance token in case of missing keys

snalli:
The replication loop waits for missing keys to be sent by remote replica so that it can apply them locally. The case for BackupChecker is different. It needs to report those missing keys and just move on.


JingQianCloud (commented):
Overall LGTM.
There is one binary file with weird name in the PR. I assume it was by accident.

snalli:
> Overall LGTM. There is one binary file with weird name in the PR. I assume it was by accident.

yeah just an error

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -168,23 +185,23 @@ protected void fixMissingStoreKeys(
    * @param acceptableStoreErrorCodes Acceptable error codes when retrieving the blob from local-store
```
ankagrawal, line 185: nit: you need to edit javadocs.

ankagrawal: Why do you need `MessageInfoType` when `MessageInfo` is already passed as argument.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -150,16 +150,33 @@ protected void fixMissingStoreKeys(
...
+    for (int i = 0; i < exchangeMetadataResponseList.size(); i++) {
+      ExchangeMetadataResponse exchangeMetadataResponse = exchangeMetadataResponseList.get(i);
+      RemoteReplicaInfo remoteReplicaInfo = replicasToReplicatePerNode.get(i);
       if (exchangeMetadataResponse.serverErrorCode == ServerErrorCode.No_Error) {
         for (MessageInfo messageInfo: exchangeMetadataResponse.getMissingStoreMessages()) {
           // Check local store once before logging an error
-          checkLocalStore(messageInfo, replicasToReplicatePerNode.get(0), acceptableLocalBlobStates, acceptableStoreErrorCodes);
+          checkLocalStore(messageInfo, replicasToReplicatePerNode.get(0), acceptableLocalBlobStates, acceptableStoreErrorCodes, MessageInfoType.PUT);
         }
+        // Advance token so that we make progress in spite of missing keys,
```
ankagrawal, line 161: Thanks for adding the comment here. This is very helpful

ankagrawal (commented):
looks good overall. A couple of nitpicks.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: 1/ what shd i add to javadocs ?
2/ yeah, removed.

ankagrawal (approved):

PR #2352, opened 2023-02-07
[Efficient Metadata Operations] Introduce partially sealed state in ReplicaId interface.

ankagrawal:
Introducing a partially sealed state in ReplicaID will enable BlobStore implementations to make replicas partially sealed. 
In order to manage three possible states of a replica this PR also introduces the ReplicaSealStatus enum.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
@@ -28,7 +28,7 @@ import static com.github.ambry.clustermap.ClusterMapUtils.*;
 public abstract class AmbryReplica implements ReplicaId {
   private final AmbryPartition partition;
   private final long capacityBytes;
-  private volatile boolean isSealed;
+  private volatile ReplicaSealStatus replicaSealStatus;
   volatile boolean isStopped;
   final ResourceStatePolicy resourceStatePolicy;
 
@@ -38,13 +38,13 @@ public abstract class AmbryReplica implements ReplicaId {
    * @param partition the {@link AmbryPartition} of which this is a replica.
    * @param isReplicaStopped whether this replica is stopped or not.
    * @param capacityBytes the capacity in bytes for this replica.
-   * @param isSealed whether this replica is in sealed state.
+   * @param replicaSealStatus {@link ReplicaSealStatus} of this replica.
    */
   AmbryReplica(ClusterMapConfig clusterMapConfig, AmbryPartition partition, boolean isReplicaStopped,
-      long capacityBytes, boolean isSealed) throws Exception {
+      long capacityBytes, ReplicaSealStatus replicaSealStatus) throws Exception {
     this.partition = Objects.requireNonNull(partition, "null partition");
     this.capacityBytes = capacityBytes;
-    this.isSealed = isSealed;
+    this.replicaSealStatus = replicaSealStatus;
     isStopped = isReplicaStopped;
     validateReplicaCapacityInBytes(capacityBytes);
     ResourceStatePolicyFactory resourceStatePolicyFactory =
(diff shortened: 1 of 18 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplica.java
```diff
@@ -79,10 +80,19 @@ public boolean isSealed() {
     return isSealed;
   }
 
+  @Override
```
JingQianCloud, line 83: isSealed and isPartiallySealed has a little bit overlap.
When isSealed, is isPartiallySealed true or false? They are mutual exclusive or has some overlap?

JingQianCloud: If they are mutually exclusive, when we set isPartiallySealed, we may set the opposite to isSealed. Vise verse. Otherwise, we will have some inconsistency between the variables.

Is one variable with enum type better,
NOT_SEALED,
PARTIALLY_SEALED,
SEALED.

ankagrawal: I agree that enum names being NOT_SEALED, PARTIALLY_SEALED and SEALED would be better. But we are so used to calling partition state ReadOnly and ReadWrite that I don't want to change it.

ankagrawal: To answer the first question -- there is definitely overlap between `isSealed` and `isPartiallySealed`. For example, every replica that is sealed is by definition also partially sealed. But in the code, for simplicity, we are treating it as separate states to avoid confusion.

JingQianCloud: It'll cause some trouble if we don't handle the "overlap" part correctly.
For example, when we call "setSealedState" to set isSealed to true, should we call "setPartiallySealedState" to set isPartiallySealed to true.

Internally we should have only one single variable which is an enum. But we can keep the interface/functions not changed.

Arun-LinkedIn: Yeah, I am also wondering if having one internal variable would be better for reasons that Jing mentioned

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1359,7 +1360,7 @@ private void updateReplicaStateAndOverrideIfNeeded(AmbryReplica replica, Collect
       } else {
         isSealed = sealedReplicas.contains(partitionName);
       }
-      replica.setSealedState(isSealed);
+      replica.setSealedStatus(isSealed ? ReplicaSealStatus.SEALED : ReplicaSealStatus.NOT_SEALED);
```
JingQianCloud, line 1363: Shall we check PARTIAL_SEALED as well?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1426,17 +1427,19 @@ private List<ReplicaId> initializeDisksAndReplicasOnNode(AmbryDataNode datanode,
...
-          boolean isSealed = sealedReplicas.contains(partitionName);
+          ReplicaSealStatus replicaSealStatus = resolveReplicaSealStatus(partitionName, sealedReplicas);
           if (clusterMapConfig.clusterMapEnablePartitionOverride && partitionOverrideInfoMap.containsKey(
               partitionName)) {
             // override sealed state if PartitionOverride is enabled.
-            isSealed = partitionOverrideInfoMap.get(partitionName)
+            boolean isSealed = partitionOverrideInfoMap.get(partitionName)
                 .get(ClusterMapUtils.PARTITION_STATE)
                 .equals(ClusterMapUtils.READ_ONLY_STR);
+            // Currently only sealed state is supported by PartitionOverride.
```
JingQianCloud, line 1437: Thanks for the comment.
Is that better that we add the isPartialSealed code support here?
Even it's not supported yet and it always isPartialSealed already return false, we won't forget to change it when we starts to support partially sealed.

JingQianCloud (commented):
Overall, looks good to me.
Two small coments.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
ankagrawal: Fixed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
ankagrawal: fixed.

JingQianCloud (approved):

Arun-LinkedIn (approved):

PR #2354, opened 2023-02-07
[Efficient Metadata Operations] Add logic in BlobStore to mark a replica as partially sealed when it crosses threshold.

ankagrawal:
This PR implements the logic in BlobStore to mark a replica as partially sealed when it crosses the threshold. Do note that ReplicaStatusDelegate still doesn't do the partial update in HelixPartiticipant, so this doesn't change the end to end logic yet.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -16,6 +16,7 @@ package com.github.ambry.store;
 import com.codahale.metrics.Timer;
 import com.github.ambry.account.AccountService;
 import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.clustermap.ReplicaSealStatus;
 import com.github.ambry.clustermap.ReplicaState;
 import com.github.ambry.clustermap.ReplicaStatusDelegate;
 import com.github.ambry.config.StoreConfig;
@@ -49,6 +50,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -84,8 +86,10 @@ public class BlobStore implements Store {
   private final UUID sessionId = UUID.randomUUID();
   private final ReplicaId replicaId;
   private final List<ReplicaStatusDelegate> replicaStatusDelegates;
-  private final long thresholdBytesHigh;
-  private final long thresholdBytesLow;
+  private final long sealThresholdBytesHigh;
+  private final long sealThresholdBytesLow;
+  private final long partialSealThresholdBytesHigh;
+  private final long partialSealThresholdBytesLow;
   private final long ttlUpdateBufferTimeMs;
   private final RemoteTokenTracker remoteTokenTracker;
   private final AtomicInteger errorCount;
@@ -98,7 +102,8 @@ public class BlobStore implements Store {
   private volatile ReplicaState currentState;
   private volatile boolean recoverFromDecommission;
   // TODO remove this once ZK migration is complete
-  private AtomicBoolean isSealed = new AtomicBoolean(false);
+  private AtomicReference<ReplicaSealStatus> replicaSealStatus =
+      new AtomicReference<>(ReplicaSealStatus.NOT_SEALED);
   private AtomicBoolean isDisabled = new AtomicBoolean(false);
   protected PersistentIndex index;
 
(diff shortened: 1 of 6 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1142,6 +1109,137 @@ public void shutdown() throws StoreException {
...
+   * Update the sealed status of the replica.
+   */
+  private void updateSealedStatus() {
+    // In two zk clusters case, "replicaId.isSealed()" might be true in first cluster but is still unsealed in second
+    // cluster, as server only spectates the first cluster. To guarantee both clusters have consistent seal/unseal
+    // state, we bypass "isSealed()" check if there are more than one replicaStatusDelegates.
+    ReplicaSealStatus replicaSealStatus1 = replicaSealStatus.get();
+    if (resolveReplicaSealStatusFromLogSize() == ReplicaSealStatus.SEALED && (replicaSealStatus.get() != ReplicaSealStatus.SEALED || (
+        replicaStatusDelegates.size() > 1 &&  replicaSealStatus.getAndSet(ReplicaSealStatus.SEALED) != ReplicaSealStatus.SEALED))) {
+      for (ReplicaStatusDelegate replicaStatusDelegate : replicaStatusDelegates) {
```
JingQianCloud, line 1122: Can replicaStatusDelegates have 2 and more entries?
If yes, what if one replicaStatusDelegate.seal is true but the other replicaStatusDelegate.seal is false?

The old code probably has the same code logic. Just want to understand it.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -98,7 +102,10 @@ public class BlobStore implements Store {
   private volatile ReplicaState currentState;
   private volatile boolean recoverFromDecommission;
   // TODO remove this once ZK migration is complete
+  private AtomicReference<ReplicaSealStatus> replicaSealStatus =
+      new AtomicReference<>(ReplicaSealStatus.NOT_SEALED);
   private AtomicBoolean isSealed = new AtomicBoolean(false);
+  private AtomicBoolean isPartiallySealed = new AtomicBoolean(false);
```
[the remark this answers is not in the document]
ankagrawal: removed.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
ankagrawal: So replicaStatusDelegates can have 2 entries during zk migration etc. Out strategy during zk migration was that we will use the replicaStatusDelegates for both the zk endpoints/clusters, and let them sync up for a few days before decommissioning the older zk.

ankagrawal: So whether a replica is in sealed/partially_sealed/unsealed state is determined by only three factors
1. Whether the size of used capacity of a replica's log exceeds a certain threshold.
2. Whether there is any other replica of the same partition marked as sealed in helix, so the replica gets a notification from helix.
3. Whether the partition id of the replica is present in partitionOverrideList (store in helix)

In all the cases, if there is any divergence, then it's safe for a replica to temporarily assume the most strict seal state. With compaction and writes, each replica will keep re-evaluating its state and eventually come back to a state that is guided by its log usage.

ankagrawal: In this `BlobStore` logic, the only thing that matters for a replica to mark itself as sealed/partially-sealed/unsealed is the log size alone (factor 1 above).

JingQianCloud: Thanks for the explanation.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -197,18 +202,26 @@ public BlobStore(ReplicaId replicaId, StoreConfig config, ScheduledExecutorServi
...
+    this.sealThresholdBytesHigh =
+        (long) (capacityInBytes * (config.storeReadOnlyEnableSizeThresholdPercentage / 100.0));
+    this.sealThresholdBytesLow = (long) (capacityInBytes * (
+        (config.storeReadOnlyEnableSizeThresholdPercentage
+            - config.storeReadOnlyToPartialWriteEnableSizeThresholdPercentageDelta) / 100.0));
+    this.partialSealThresholdBytesHigh =
+        (long) (capacityInBytes * (config.storePartialWriteEnableSizeThresholdPercentage / 100.0));
+    this.partialSealThresholdBytesLow = (long) (capacityInBytes * (
+        (config.storePartialWriteEnableSizeThresholdPercentage
+            - config.storePartialWriteToReadWriteEnableSizeThresholdPercentageDelta) / 100.0));
```
JingQianCloud, line 214: Currently we haven't fully supported the partially seal yet.
After we merge this code, how do we make sure there is no impact to the existing code logic?

JingQianCloud: What's the default value of partialSealThresholdBytesLow and storePartialWriteToReadWriteEnableSizeThresholdPercentageDelta we'll set?

ankagrawal: The default value for `storePartialWriteEnableSizeThresholdPercentage` is 50%. And the default value for `storePartialWriteToReadWriteEnableSizeThresholdPercentageDelta` is 5%. They are defined in StoreConfig.java.

ankagrawal: > After we merge this code, how do we make sure there is no impact to the existing code logic?

If you check the `ReplicaStatusDelegate::partialSeal()` implementation, it does the same thing that `ReplicaStatusDelegate::unseal()` does. So effectively, anything that is not sealed will remain not sealed.

JingQianCloud: I think after we set the partialSeal(), even the implementation is to set as unseal, the metrics will start to show as partial sealed.

I'm wondering if we set storePartialWriteEnableSizeThresholdPercentage to storeReadOnlyEnableSizeThresholdPercentage, looks like we won't generate partial seal status at all.

I leave it to you.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1145,6 +1110,142 @@ public void shutdown() throws StoreException {
...
+  /**
+   * Update the sealed status of the replica.
+   */
+  private void updateSealedStatus() {
+    // In two zk clusters case, "replicaId.isSealed()" might be true in first cluster but is still unsealed in second
+    // cluster, as server only spectates the first cluster. To guarantee both clusters have consistent seal/unseal
+    // state, we bypass "isSealed()" check if there are more than one replicaStatusDelegates.
+    ReplicaSealStatus replicaSealStatus1 = replicaSealStatus.get();
+    if (resolveReplicaSealStatusFromLogSize() == ReplicaSealStatus.SEALED && (
+        replicaSealStatus.get() != ReplicaSealStatus.SEALED || (replicaStatusDelegates.size() > 1
```
Arun-LinkedIn, line 1122: Should this be `replicaId.isSealed()` since `replicaSealStatus.get()` is local value?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1145,6 +1110,142 @@ public void shutdown() throws StoreException {
...
   }
 
+  /**
+   * Update the sealed status of the replica.
+   */
+  private void updateSealedStatus() {
+    // In two zk clusters case, "replicaId.isSealed()" might be true in first cluster but is still unsealed in second
+    // cluster, as server only spectates the first cluster. To guarantee both clusters have consistent seal/unseal
+    // state, we bypass "isSealed()" check if there are more than one replicaStatusDelegates.
+    ReplicaSealStatus replicaSealStatus1 = replicaSealStatus.get();
```
Arun-LinkedIn, line 1120: nit: would name like `oldReplicaSealStatus` be better?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
Arun-LinkedIn: Yeah, I think this should be `replicaId.isSealed()` since `replicaSealStatus` is only updated during two ZK clusters case

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1145,6 +1110,142 @@ public void shutdown() throws StoreException {
...
+   */
+  private void reconcileSealedStatus() {
+    // During startup, we also need to reconcile the replica state from both ZK clusters.
+    // reconcile the state by reading sealing state from both clusters
+    ReplicaSealStatus sealStatus = resolveReplicaSealStatusFromLogSize();
+    String partitionName = replicaId.getPartitionId().toPathString();
+    for (ReplicaStatusDelegate replicaStatusDelegate : replicaStatusDelegates) {
+      Set<String> sealedReplicas = new HashSet<>(replicaStatusDelegate.getSealedReplicas());
+      Set<String> partiallySealedReplicas = new HashSet<>(replicaStatusDelegate.getPartiallySealedReplicas());
+      ReplicaSealStatus sealStatus1 = sealedReplicas.contains(partitionName) ? ReplicaSealStatus.SEALED : (partiallySealedReplicas.contains(partitionName) ? ReplicaSealStatus.PARTIALLY_SEALED : ReplicaSealStatus.NOT_SEALED);
```
Arun-LinkedIn, line 1182: nit: would it be okay to rename `sealStatus1` to `sealStatus` and `sealStatus` to `globalSealStatus` or something like that?

Arun-LinkedIn (approved):

PR #2356, opened 2023-02-08
[DR] Redirect output of BackupChecker to other files

snalli:
This will make the BackupChecker redirect all its output to different files so that we have a record of missing keys.


ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -59,6 +72,22 @@ public class BackupCheckerThread extends ReplicaThread {
 
   private final Logger logger = LoggerFactory.getLogger(BackupCheckerThread.class);
   public static final String DR_Verifier_Keyword = "dr";
+  protected final AmbryCache fileDescriptorCache;
+  private final String basePath = "/export/content/lid/logs/ambry-server/i001";
```
ankagrawal, line 76: Please don't hardcode filepath. Use a config.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -175,6 +205,101 @@ protected void advanceToken(RemoteReplicaInfo remoteReplicaInfo, ExchangeMetadat
...
+   * @param remoteReplicaInfo The remote replica that is being replicated from
+   * @return Return a file descriptor of type SeekableByteChannel
+   */
+  protected SeekableByteChannel getOrCreateStatusFileFd(RemoteReplicaInfo remoteReplicaInfo) {
+    String filePath = basePath
+        + File.separator + remoteReplicaInfo.getReplicaId().getPartitionId().getId()
+        + File.separator + remoteReplicaInfo.getReplicaId().getDataNodeId().getHostname()
+        + File.separator + "replicaCheckStatus";
+    /**
+     * NOTE: Working with these file options has been a pain. They do not work as advertised in documentation.
```
ankagrawal, line 252: Can we remove this comment? As much as I sympathize with the pain, this isn't useful for the reader.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -252,7 +252,8 @@ NetworkClient getNetworkClient() {
    * Empty for current ReplicaThread but can be re-written in other cases to track progress.
    * @param remoteReplicaInfo
    */
-  protected void logReplicationCaughtUp(RemoteReplicaInfo remoteReplicaInfo) {
```
ankagrawal, line 255: nit: update javadocs for this method.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -59,6 +72,22 @@ public class BackupCheckerThread extends ReplicaThread {
 
   private final Logger logger = LoggerFactory.getLogger(BackupCheckerThread.class);
   public static final String DR_Verifier_Keyword = "dr";
+  protected final AmbryCache fileDescriptorCache;
+  private final String basePath = "/export/content/lid/logs/ambry-server/i001";
+  public static final DateFormat dateFormat = new SimpleDateFormat("dd MMM yyyy HH:mm:ss:SSS");
+
+  protected class FileDescriptor implements AmbryCacheEntry {
```
ankagrawal, line 79: Why does a `FileDescriptor` class need to implement `AmbryCacheEntry` ?

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -59,6 +72,22 @@ public class BackupCheckerThread extends ReplicaThread {
 
   private final Logger logger = LoggerFactory.getLogger(BackupCheckerThread.class);
   public static final String DR_Verifier_Keyword = "dr";
+  protected final AmbryCache fileDescriptorCache;
+  private final String basePath = "/export/content/lid/logs/ambry-server/i001";
+  public static final DateFormat dateFormat = new SimpleDateFormat("dd MMM yyyy HH:mm:ss:SSS");
```
ankagrawal, line 77: constant variable names are all upper case with _ separator.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -175,6 +205,101 @@ protected void advanceToken(RemoteReplicaInfo remoteReplicaInfo, ExchangeMetadat
...
+    return getOrCreateReportFd(filePath, EnumSet.of(StandardOpenOption.CREATE, StandardOpenOption.WRITE));
+  }
+
+  /**
+   * Creates or returns an existing file descriptor for file missingKeys
+   * @param remoteReplicaInfo The remote replica that is being replicated from
+   * @return Return a file descriptor of type SeekableByteChannel
+   */
+  protected SeekableByteChannel getOrCreateMissingKeysFd(RemoteReplicaInfo remoteReplicaInfo) {
+    String filePath = basePath
```
ankagrawal, line 265: Use `String.join` here.

ankagrawal: I am not sure if creating so many files on a prod node is a good idea. Why can't we create just one file and make the log string prefixes unique? What will creating multiple new files give us that creating log entries with appropriate prefix not give us?

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: If this is not done, the code compiles but fails during runtime with an error. The AmbryCache expects the objects inserted to be of type AmbryCacheEntry.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: I'd like to have this for the record, at least for myself if not someone else. I spent close to half the day figuring out why the code wasn't working as expected and I don't want to repeat it. I am sure it will be useful to someone other than me. At the very least, it's not misleading.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: What is the issue with creating files? There is more than enough space on the host. Logging this stuff in the logs means I'll have to later grep and sort it out. Output from multiple threads will be interleaved. Logs will be rotated and garbage collected. And I may have to rerun the whole thing. Or I'll have to transfer logs to my Mac for post processing. I'd like to avoid all of that. I am sure I will be asked to share the results of this effort. And I want to be able to quickly collect it at a moment's notice.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
JingQianCloud: Also can we use relative path instead of absolution path? Let's refer to how other modules/existing code picks the path.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
JingQianCloud: I assume we only open one or a couple files to save the outcome. File descriptor limit shouldn't be our concern, right?

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -59,6 +72,22 @@ public class BackupCheckerThread extends ReplicaThread {
...
   public static final String DR_Verifier_Keyword = "dr";
+  protected final AmbryCache fileDescriptorCache;
+  private final String basePath = "/export/content/lid/logs/ambry-server/i001";
+  public static final DateFormat dateFormat = new SimpleDateFormat("dd MMM yyyy HH:mm:ss:SSS");
+
+  protected class FileDescriptor implements AmbryCacheEntry {
+
+    SeekableByteChannel _seekableByteChannel;
+
+    public FileDescriptor(SeekableByteChannel seekableByteChannel) {
```
JingQianCloud, line 83: Do we need random access to the file? 
I thought we only need append to the end of the file.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
JingQianCloud: We can add comment to mention the flags needed. It should be enough, especially it's OSS.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -175,6 +205,101 @@ protected void advanceToken(RemoteReplicaInfo remoteReplicaInfo, ExchangeMetadat
     remoteReplicaInfo.setLocalLagFromRemoteInBytes(exchangeMetadataResponse.localLagFromRemoteInBytes);
     // reset stored metadata response for this replica so that we send next request for metadata
     remoteReplicaInfo.setExchangeMetadataResponse(new ExchangeMetadataResponse(ServerErrorCode.No_Error));
+    logReplicationStatus(remoteReplicaInfo, exchangeMetadataResponse);
+  }
+
+  /**
+   * Write to a given file at a given offset
```
JingQianCloud, line 212: In which case, we need random access?

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: I checked the config on the host. They use absolute paths. Absolute paths eliminate any confusion where the files will land in the system.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: Yeah, certain combos of the flags work and some don't.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: not one or two. For the osua event, there are 40 partitions. For each partition, there are 3 replicas. And each replica will have two files - missingKeys and replicaCheckStatus. So 40*3*2 = 240 files at most. And this can grow if the number of partitions increase. I also plan to add one more file later.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: For replicaCheckStatus, I reset the file pointer to start of the file since i don't want a history of how the token moves but just the current token and lag.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: to report current status of checking the replica - curr token, curr lag.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
ankagrawal: This is open source repo. There should be no linkedin specific assumption in this code.

Also, directory structure is more of a deployment concern. Tomorrow the infra could change, and the directory structure might change as well. So its never good to hardcode directory structures in code.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
ankagrawal: Again, atleast because its open source code, lets not make any subjective statements about APIs that aren't part of this code base. Objective statements like "ABC flags don't work as expected in library X with version Y" are ok.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
ankagrawal: Can we somehow move this file creation logic to closed source repo?

ankagrawal (changes requested):

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: I will remove the hardcoded path. I am not talking about that above.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: yes that's better

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
JingQianCloud: Can you elaborate more?
Why do we need seek and change the file position?
If we need some current token, current lag, are they supposed to be in the memory?

JingQianCloud: And in any case you want to reset the file pointer to the start, we'd better do a "file truncate". Don't seek the position to the beginning and modify the file content. You may end up with file content garbage since the remaining content still exist.

snalli: 1/ When replicating from remote server, I want to know when the local server catches up with the remote server so that I can claim the backup checker has completed and stop the process. I don't want to guess this. I want to be sure and by printing it out to a file, I can instantly and accurately know how far its behind the remote server. And this is not just for 1 partitions but 40 or more.
2/ I already looked at truncate and it does not work as expected. It expects the file to exist. If it does not, then truncate fails. So then for the case when file is absent, i'd have check if it exists, remove the truncate option if the caller passes it and then create the file. Then the next time, i'd have to check if it exists and then honor the truncate option if the option is passed from the caller. I left a comment explaining the various combos I tried. There are several cases to handle and the code as written currently handles all cases without a lot of if-else statements. You are right about the garbage. I do see some random numbers at the end of the file but the content I require is there intact. However, the length of the text printed out each time is more or less the same such that the entire content is overwritten. We can overcome the garbage by appending plenty of spaces or newline at the end of the text.

snalli: yes the current token and lag are in memory already, i am just printing it out

snalli: Let me give this a bit more thought. I do like the idea of truncate because it gives a clean output. For now, I've moved it into closed-source.

snalli: replied in closed-source

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerFileManager.java
```diff
@@ -0,0 +1,43 @@
...
+
+import com.codahale.metrics.MetricRegistry;
+import com.github.ambry.config.ReplicationConfig;
+import java.nio.file.StandardOpenOption;
+import java.util.EnumSet;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * File manager for BackupChecker, although it can be used for other purposes
```
ankagrawal, line 24: nit: if it can be used for other purposes, then let's rename this as something like `WritableFileManager` (writable because the only thing that is does is `writeToFile`. Also should it be an interface with a `default` implementation instead of being a class?

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -71,6 +84,11 @@ public BackupCheckerThread(String threadName, FindTokenHelper findTokenHelper, C
         replicationConfig, replicationMetrics, notification, storeKeyConverter, transformer, metricRegistry,
         replicatingOverSsl, datacenterName, responseHandler, time, replicaSyncUpManager, skipPredicate,
         leaderBasedReplicationAdmin);
+    try {
+      fileManager = Utils.getObj(replicationConfig.backupCheckFileManagerType, replicationConfig, metricRegistry);
```
ankagrawal, line 88: Maybe add a log message?

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -71,6 +84,11 @@ public BackupCheckerThread(String threadName, FindTokenHelper findTokenHelper, C
...
+    } catch (ReflectiveOperationException e) {
+      throw new RuntimeException(e);
```
ankagrawal, line 90: Maybe add a log message here to make it easy to debug?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
ankagrawal: This method's signature was changed but the javadocs is still not updated. It still has only one `@param` . Also no description for `remoteReplicaInfo`.

ankagrawal (commented):
Mostly looks good to me, just some minor formatting/naming type comments.

I still feel that creating so many files might not be the best idea, but I can't pin down the exact reason for that. So I don't want to block the PR due to that.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -58,7 +60,18 @@
...
   private final Logger logger = LoggerFactory.getLogger(BackupCheckerThread.class);
+  protected final BackupCheckerFileManager fileManager;
   public static final String DR_Verifier_Keyword = "dr";
+  public static final String MISSING_KEYS_FILE = "missingKeys";
+  public static final String REPLICA_STATUS_FILE = "replicaCheckStatus";
+  /**
+   * To create a file, you need both CREATE + WRITE/APPEND and not just CREATE or just WRITE/APPEND.
+   * For TRUNCATE_EXISTING to work, the file must exist and CREATE does not seem to have any effect there.
+   */
+  public static final EnumSet<StandardOpenOption> CREATE_TRUNC_WRITE = EnumSet.of(StandardOpenOption.CREATE,
```
JingQianCloud, line 71: Personally I like to encapsulate all the file system API options to the file manager. 
The caller here only cares it's an append or fresh write. 
File manager decides the API options. File manager knows if the file exists or not. If it doesn't exist, it create it. If it exists, depends on it's an append or fresh write, it may truncate or don't truncate.

ankagrawal (dismissed):

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerFileManager.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+public class BackupCheckerFileManager {
+  private final Logger logger = LoggerFactory.getLogger(BackupCheckerFileManager.class);
+
+  public BackupCheckerFileManager(ReplicationConfig replicationConfig, MetricRegistry metricRegistry) {
```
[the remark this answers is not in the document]
snalli: used in closed src impl

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerFileManager.java
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * Write to a given file at a given offset
```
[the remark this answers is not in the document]
snalli: fixed

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: done

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: done

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: WRITE = just open the file for write, don't create, don't trunc, fail if file absent
WRITE + TRUNC =  just open the file for write, don't create, trunc, fail if file absent
Add CREATE to above means create the file if absent. 
I could make the code in this class thinner by excluding some options and adding more logic in the file mgr using if-else stmts, but i am trying to not to be too specific for use case. And using a set of options and leveraging the Files package to create a file is much easier. No if-else statements. In the end, the result is the same with lesser if-else statements.

JingQianCloud: I don't know in which case the caller wants the file write to fail if the file doesn't absent. 
In the health checker, from caller's point of view, it only wants two things,
1. WRITE at the beginning = write from the beginning, if file exists, then truncate. If not, then create.
2. APPEND = if the file exists, continue to write. If it doesn't, create the file and write.

Different Java API and library may require different parameter options. Expose the option to the user/caller may limit the choice of the file manager's implementation.

But I leave it to you.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
```diff
@@ -265,8 +265,38 @@ public class ReplicationConfig {
...
+  public static final String BACKUP_CHECKER_FILE_MANAGER_TYPE = "backup.checker.file.manager.type";
+  public static final String DEFAULT_BACKUP_CHECKER_FILE_MANAGER = "com.github.ambry.replication.BackupCheckerFileManager";
+  @Config(BACKUP_CHECKER_FILE_MANAGER_TYPE)
+  public final String backupCheckFileManagerType;
+
+  /**
+   * Config for folder to store all results from backup checker
+   */
+  public static final String BACKUP_CHECKER_REPORT_DIR = "backup.checker.report.dir";
+  public static final String DEFAULT_BACKUP_CHECKER_REPORT_DIR = "/tmp";
```
JingQianCloud, line 280: Not sure if /tmp is writable on the server. But we can change the config any way.

JingQianCloud (approved):

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: removed the options to allow for flexible filemgr impl

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
snalli: this is just for open source, on the server the config will be different something like /export/lid etc.

PR #2358, opened 2023-02-09
Support update dataset

SophieGuo410:
This pr is to support partial update of the dataset metadata.

### DIFF
--- a/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
+++ b/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
@@ -787,6 +787,27 @@ public class MySqlAccountServiceIntegrationTest {
     } catch (Exception e) {
       //no op
     }
+
+    //Add a dataset with limited ttl
+    expirationTimeInMs = Utils.addSecondsToEpochTime(System.currentTimeMillis(), 30);
+    dataset = new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME_WITH_TTL).setVersionSchema(
+            Dataset.VersionSchema.TIMESTAMP)
+        .setExpirationTimeMs(expirationTimeInMs)
+        .setUserTags(userTags)
+        .setRetentionCount(retentionCount)
+        .build();
+    mySqlAccountStore.addDataset(testAccount.getId(), testContainer.getId(), dataset);
+
+    //update expiration to permanent.
+    expirationTimeInMs = -1;
+    datasetForUpdate =
+        new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME_WITH_TTL).setExpirationTimeMs(
+            expirationTimeInMs).build();
+    mySqlAccountStore.updateDataset(testAccount.getId(), testContainer.getId(), datasetForUpdate);
+    datasetFromMysql = mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
+        testContainer.getName(), DATASET_NAME_WITH_TTL);
+    assertEquals("Mismatch in updated expirationTimeInMst", (Long) expirationTimeInMs,
+        datasetFromMysql.getExpirationTimeMs());
   }
 
   /**
--- a/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
+++ b/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
@@ -67,6 +67,7 @@ public class MySqlAccountServiceIntegrationTest {
   private MySqlAccountServiceConfig accountServiceConfig;
   private MySqlAccountService mySqlAccountService;
   private static final String DATASET_NAME = "testDataset";
+  private static final String DATASET_NAME_NOT_EXIST = "testDatasetNotExist";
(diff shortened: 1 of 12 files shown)


ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -441,6 +450,32 @@ public synchronized void addDataset(int accountId, int containerId, Dataset data
...
+      long startTimeMs = System.currentTimeMillis();
+      try {
+        getDataset(accountId, containerId, dataset.getAccountName(), dataset.getContainerName(),
+            dataset.getDatasetName());
+      } catch (Exception e) {
+        throw new SQLException("Can't update the dataset if it does not exist. Dataset: " + dataset.toString());
+      }
+      dataAccessor.getDatabaseConnection(true);
+      PreparedStatement updateDatasetStatement = dataAccessor.getPreparedStatement(updateDatasetSql, true);
+      executeUpdateDatasetStatement(updateDatasetStatement, accountId, containerId, dataset);
```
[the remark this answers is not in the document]
SophieGuo410: You are right, I didn't realize it before. Thanks for caching it! Fixed.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -138,6 +139,14 @@ public AccountDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
         String.format("insert into %s (%s, %s, %s, %s, %s, %s, %s, %s) values (?, ?, ?, ?, now(3), ?, ?, ?)",
             DATASET_TABLE, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_COUNT,
             USER_TAGS, DELETE_TS);
+    //update the dataset field, in order to support partial update, if one parameter is null, keep the original value.
+    //if the delete_ts parameter from input equals to Utils.Infinite_Time(-1), set the delete_ts to null.
+    updateDatasetSql = String.format(
```
justinlin-linkedin, line 144: I have a couple question for this statement.
1. Can we just use IFNULL instead of COALESE(NULLIF(?, ''),XX). If the retentionCount is NULL from the input, we just use the existing retentioinCount, `retentionCount` = IFNULL(?, `retentionCount`)
2. for delete_ts, in the CASE WHEN statement, shouldn't we compare with -1, not '-1', which is a string?

mikeyang923 (approved):

ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
```diff
@@ -739,6 +745,48 @@ public void testAddAndGetDatasets() throws Exception {
...
+    //update the expiration time.
+    long expirationTimeInMs = Utils.addSecondsToEpochTime(System.currentTimeMillis(), 30);
+    datasetForUpdate =
+        new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME).setExpirationTimeMs(
+            expirationTimeInMs).build();
+    mySqlAccountStore.updateDataset(testAccount.getId(), testContainer.getId(), datasetForUpdate);
+    datasetFromMysql = mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
+        testContainer.getName(), DATASET_NAME);
+    assertEquals("Mismatch in updated expirationTimeInMst", (Long) expirationTimeInMs,
+        datasetFromMysql.getExpirationTimeMs());
```
justinlin-linkedin, line 768: could you please add another test where the delete_ts is not -1 when created, and later we update to -1.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -441,6 +450,32 @@ public synchronized void addDataset(int accountId, int containerId, Dataset data
...
+   */
+  public synchronized void updateDataset(int accountId, int containerId, Dataset dataset) throws SQLException {
+    try {
+      long startTimeMs = System.currentTimeMillis();
+      try {
+        getDataset(accountId, containerId, dataset.getAccountName(), dataset.getContainerName(),
+            dataset.getDatasetName());
+      } catch (Exception e) {
+        throw new SQLException("Can't update the dataset if it does not exist. Dataset: " + dataset.toString());
+      }
```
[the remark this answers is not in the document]
SophieGuo410: Yeah, one thing is update won't throw any exception, it just not update it. But I took Mike's advice to get the update count to throw exception if updatedRowCount is null.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: We could, I just simply it.
And for delete_ts, since it will compare with timestamp, so it can't be an integer. 
But I do find a bug and fix it after adding more tests. I updated the statement executeUpdateDatasetStatement portion.

ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
SophieGuo410: Updated.

justinlin-linkedin (approved):

PR #1796, opened 2021-03-05
Introduce RemoteTokenTracker in BlobStore to clean up permanent delete tombstone

jsjtzyy:
Introduce RemoteTokenTracker to help keep track of replication tokens from all peer replicas. The tokens maintained within this tracker will be consumed by compaction.

PR #2361, opened 2023-02-10
Persist RemoteTokenTracker

JingQianCloud:
We have delete record tombstone issue. As long as there is one replica down, we cannot move forward to remove any tombstone because we don't know the replication token of that replica.
After we persist the remote token, if there is a server newly down, at least we can still continue to remove the delete record as long as the delete record is before the token.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -509,6 +509,14 @@ public class StoreConfig {
   public static final String storeRebuildTokenBasedOnCompactionHistoryName =
       "store.rebuild.token.based.on.compaction.history";
 
+  /**
+   * If storePersistRemoteTokenIntervalInSeconds > 0, persist the remote token every storePersistRemoteTokenIntervalInSeconds seconds.
+   */
+  @Config(storePersistRemoteTokenIntervalInSecondsName)
+  public final int storePersistRemoteTokenIntervalInSeconds;
+  public static final String storePersistRemoteTokenIntervalInSecondsName =
+      "store.persist.remote.token.interval.in.seconds";
+
   public StoreConfig(VerifiableProperties verifiableProperties) {
 
     storeKeyFactory = verifiableProperties.getString("store.key.factory", "com.github.ambry.commons.BlobIdFactory");
@@ -634,5 +642,8 @@ public class StoreConfig {
       storePartitionsToRebuildTokenBasedOnCompactionHistory =
           Stream.of(partitions.split(",")).map(Long::parseLong).collect(Collectors.toList());
     }
+
+    storePersistRemoteTokenIntervalInSeconds =
+        verifiableProperties.getIntInRange(storePersistRemoteTokenIntervalInSecondsName, 0, 0, 60 * 60 * 24);
   }
 }
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -205,7 +205,7 @@ public class BlobStore implements Store {
     ttlUpdateBufferTimeMs = TimeUnit.SECONDS.toMillis(config.storeTtlUpdateBufferTimeSeconds);
     errorCount = new AtomicInteger(0);
     currentState = ReplicaState.OFFLINE;
-    remoteTokenTracker = replicaId == null ? null : new RemoteTokenTracker(replicaId);
+    remoteTokenTracker = replicaId == null ? null : new RemoteTokenTracker(replicaId, taskScheduler, factory);
     logger.debug(
         "The enable state of replicaStatusDelegate is {} on store {}. The high threshold is {} bytes and the low threshold is {} bytes",
         config.storeReplicaStatusDelegateEnable, storeId, this.thresholdBytesHigh, this.thresholdBytesLow);
@@ -267,6 +267,9 @@ public class BlobStore implements Store {
         metrics.initializeIndexGauges(storeId, index, capacityInBytes, blobStoreStats,
             config.storeEnableCurrentInvalidSizeMetric, config.storeEnableIndexDirectMemoryUsageMetric);
         checkCapacityAndUpdateReplicaStatusDelegate();
+        if (remoteTokenTracker != null) {
+          remoteTokenTracker.start(config.storePersistRemoteTokenIntervalInSeconds);
+        }
         logger.trace("The store {} is successfully started", storeId);
         onSuccess("START");
         isDisabled.set(false);
@@ -1157,6 +1160,9 @@ public class BlobStore implements Store {
         compactor.close(30);
         index.close(skipDiskFlush);
         log.close(skipDiskFlush);
+        if (remoteTokenTracker != null) {
+          remoteTokenTracker.close();
+        }
         metrics.deregisterMetrics(storeId);
         currentState = ReplicaState.OFFLINE;
         started = false;
--- a/ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
@@ -15,29 +15,62 @@ package com.github.ambry.store;
 
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.replication.FindToken;
+import com.github.ambry.utils.CrcInputStream;
+import com.github.ambry.utils.CrcOutputStream;
+import com.github.ambry.utils.Pair;
+import com.github.ambry.utils.Utils;
+import java.io.Closeable;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 
 /**
  * {@link RemoteTokenTracker} tracks tokens from all peer replicas and updates them when handling metadata request from
  * peer node.
+ * We persist the peer replicas' token and the timestamp when the token is updated to a file.
  */
-public class RemoteTokenTracker {
+public class RemoteTokenTracker implements Closeable {
+  private static String REMOTE_TOKEN_FILE_NAME = "remoteTokenTracker";
+
   private static final String DELIMITER = ":";
   private final ReplicaId localReplica;
   // The key of peerReplicaAndToken is a string containing hostname and path of peer replica.
-  // For example: localhost:/mnt/u001/p1
-  private ConcurrentMap<String, FindToken> peerReplicaAndToken = new ConcurrentHashMap<>();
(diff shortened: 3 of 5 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -206,7 +206,7 @@ public BlobStore(ReplicaId replicaId, StoreConfig config, ScheduledExecutorServi
     ttlUpdateBufferTimeMs = TimeUnit.SECONDS.toMillis(config.storeTtlUpdateBufferTimeSeconds);
     errorCount = new AtomicInteger(0);
     currentState = ReplicaState.OFFLINE;
-    remoteTokenTracker = replicaId == null ? null : new RemoteTokenTracker(replicaId);
+    remoteTokenTracker = replicaId == null ? null : new RemoteTokenTracker(replicaId, taskScheduler, factory);
```
snalli, line 209: just curious, this remote token tracker is in the memory of the sender - the server that sends blobs right ?
and with this change you are persisting that token. So if B pulls from A, then A persists its remoteToken corresponding to B?

JingQianCloud: Correct.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -15,29 +15,62 @@
...
+import org.slf4j.LoggerFactory;
 
 
 /**
  * {@link RemoteTokenTracker} tracks tokens from all peer replicas and updates them when handling metadata request from
  * peer node.
+ * We persist the peer replicas' token and the timestamp when the token is updated to a file.
  */
-public class RemoteTokenTracker {
+public class RemoteTokenTracker implements Closeable {
```
ankagrawal, line 46: Curious: What was the issue with the previous remote token tracker?

JingQianCloud: Nothing is wrong with the previous tracker.

We have delete record tombstone issue. As long as there is one replica down, we cannot move forward to remove any tombstone because we don't know the replication token of that replica.
After we persist the token, if there is a server down, at least we can still continue to remove the delete record as long as the delete record is before the token.

Later, we can have more improvement. With the timestamp, we know if the replica is down for 1 month, we'll ignore that down replica. But this is not in the PR.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -69,7 +103,201 @@ void refreshPeerReplicaTokens() {
...
+   * @return the file path which persists the remote token.
+   */
+  public String getPersistFilePath() {
+    return localReplica.getReplicaPath() + File.separator + REMOTE_TOKEN_FILE_NAME;
+  }
+
+  /**
+   * Provide a way to explicitly persist the token besides the background persistor
+   */
+  public boolean persistToken() {
```
snalli, line 120: Just clarifying questions
1. Were we persisting the token on the sender before this ?
2. Why are we persisting now ?

JingQianCloud: We don't persist the remote token currently. Remote token was introduced in the below PR and tried to solve the tombstone problem. It helps with the tombstone problem.
https://github.com/linkedin/ambry/pull/1796

If we don't persist the remote token, as long as there is server deployment and there is one replica down, we lose the remote token for that down replica. Because of that, we cannot remove any tombstone any more.
But if we persist the token, we can continue to remove tombstone as long as the token of the down replica has passed that tombstone.

We have more room to improve it after we persist the token. For example, we know how long the remote replicas has been down, we can decide if we remove the remote replica from the replica pool.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -69,7 +103,204 @@ void refreshPeerReplicaTokens() {
...
+    persistor.cancel();
+  }
+
+  /**
+   * Start the background persistor.
+   */
+  public void start(boolean enablePersist) {
+    if (enablePersist) {
+      if (this.scheduler != null) {
+        this.scheduler.scheduleAtFixedRate(persistor, 0, REMOTE_TOKEN_FILE_PERSIST_INTERVAL_IN_SECONDS,
```
justinlin-linkedin, line 143: we can have a configuration for the interval. As a matter of fact, we can use this value to enable the persistence. If this value is 0 or negative, we don't enable the persistence, but if it's not, we enable it.

justinlin-linkedin: anyway, just think we should probably have a configuration for this.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -268,6 +268,9 @@ public void start() throws StoreException {
         metrics.initializeIndexGauges(storeId, index, capacityInBytes, blobStoreStats,
             config.storeEnableCurrentInvalidSizeMetric, config.storeEnableIndexDirectMemoryUsageMetric);
         checkCapacityAndUpdateReplicaStatusDelegate();
+        if (remoteTokenTracker != null) {
+          remoteTokenTracker.start(config.storePersistRemoteToken);
+        }
```
justinlin-linkedin, line 273: nit: this is better
```
if (remoteTokenTracker != null && config.storePersisteRemoteToken) {
    remoteTokenTracker.start();
}
```

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -69,7 +103,204 @@ void refreshPeerReplicaTokens() {
...
    */
-  Map<String, FindToken> getPeerReplicaAndToken() {
+  Map<String, Pair<Long, FindToken>> getPeerReplicaAndToken() {
     return new HashMap<>(peerReplicaAndToken);
   }
+
+  /**
+   * @return the file path which persists the remote token.
+   */
+  public String getPersistFilePath() {
```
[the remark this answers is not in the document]
JingQianCloud: Done.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
JingQianCloud: Done

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -69,7 +103,204 @@ void refreshPeerReplicaTokens() {
...
+    return localReplica.getReplicaPath() + File.separator + REMOTE_TOKEN_FILE_NAME;
+  }
+
+  /**
+   * Provide a way to explicitly persist the token besides the background persistor
+   */
+  public boolean persistToken() {
```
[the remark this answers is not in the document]
JingQianCloud: Done.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
JingQianCloud: Changed.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -69,7 +103,204 @@ void refreshPeerReplicaTokens() {
...
+      } else {
+        return new ConcurrentHashMap<>();
+      }
+    }
+  }
+
+  /**
+   * Class to serialize and deserialize replica tokens
+   */
+  private static class RemoteReplicaTokenSerde {
```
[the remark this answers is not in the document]
JingQianCloud: The major content of the remote token is the StoreFindToken, it supports different versions. Following are the code of StoreFindToken::fromBytes and StoreFindToken::toBytes.
https://jarvis.corp.linkedin.com/codesearch/result/?name=StoreFindToken.java&path=ambry%2Fambry-store%2Fsrc%2Fmain%2Fjava%2Fcom%2Fgithub%2Fambry%2Fstore&reponame=linkedin%2Fambry#212
https://jarvis.corp.linkedin.com/codesearch/result/?name=StoreFindToken.java&path=ambry%2Fambry-store%2Fsrc%2Fmain%2Fjava%2Fcom%2Fgithub%2Fambry%2Fstore&reponame=linkedin%2Fambry#428

After we persist the data, we also need retrieve them back. It needs quite some effort to verify if the json version matches with the design of all the versions. Considering the RISK and effort, I think we'd better use the existing binary serializer and deserializer of the StoreFindToken.

I understand for debugging purpose, binary version couldn't help. Currently I print out the token in String format during bootup.
Also I think when I do serialization, I can generate two versions, 
1. binary version "remoteTokenTracker".  It'll be used by retrieval. 
2. Jason version "remoteTokenTracker.txt".  It's for debugging purpose only. We only dump it but we never use it.

Let me know your thought.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -678,5 +686,7 @@ public StoreConfig(VerifiableProperties verifiableProperties) {
 
     storeDeletedPutAsDeleteInFindEntries =
         verifiableProperties.getBoolean(storeDeletedPutAsDeleteInFindEntriesName, false);
+    storePersistRemoteTokenIntervalInSeconds =
+        verifiableProperties.getIntInRange(storePersistRemoteTokenIntervalInSecondsName, 0, 60 * 1, 60 * 60 * 24);
```
justinlin-linkedin, line 690: this would probably not going to work, the default value is no in the range.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -69,7 +101,204 @@ void refreshPeerReplicaTokens() {
...
+
+  /**
+   * Runner that persist the remote peers and their tokens.
+   */
+  private class RemoteTokenPersistor implements Runnable {
+    private volatile boolean cancelled = false;
+
+    @Override
+    public void run() {
+      if (!cancelled) {
```
justinlin-linkedin, line 157: we don't need a cancelled boolean here, we can let the schedule return a ScheduledFuture and cancel the future when closing the RemoteTokeneTracker.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -69,7 +101,204 @@ void refreshPeerReplicaTokens() {
...
+    }
+  }
+
+  /**
+   * Disable this {@link RemoteTokenTracker} and cancel the scheduled persistor.
+   */
+  @Override
+  public void close() {
+    persistor.cancel();
+  }
```
justinlin-linkedin, line 133: we can make schedule to return a future and use future to stop the execution. And then we should call persist() again to make sure we dump the latest set of tokens.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
JingQianCloud: Nice catch. Thank you.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
JingQianCloud: Thanks for the suggestion. Changed.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
JingQianCloud: Done.

justinlin-linkedin (approved):

PR #2366, opened 2023-02-14
Replicate blob, Separate the meter for TtlUpdate and Delete

JingQianCloud:
Use different metrics for ReplicateBlobOnDelete and ReplicateBlobOnTtlUpdate

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
@@ -98,7 +98,7 @@ public class OperationQuotaCharger implements Chargeable {
   @Override
   public QuotaResource getQuotaResource() {
     if (Objects.isNull(quotaChargeCallback)) {
-      if (!operationName.equals("ReplicateBlobOperation")) {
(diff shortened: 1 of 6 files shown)


ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
```diff
@@ -98,7 +98,9 @@ public QuotaAction checkAndCharge(boolean shouldCheckExceedAllowed) {
   @Override
   public QuotaResource getQuotaResource() {
     if (Objects.isNull(quotaChargeCallback)) {
-      LOGGER.warn("quota charge callback is null for operation: {}", operationName);
+      if (!operationName.equals("ReplicateBlobOperation")) {
```
justinlin-linkedin, line 101: nit: can we do`ReplicateBlobOperation.class.getSimpleName()`?

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/OperationQuotaCharger.java
JingQianCloud: Done

PR #2370, opened 2023-02-16
[FULL_AUTO] ClusterMap: Add a method to return if a data node is on FULL_AUTO

justinlin-linkedin:
This PR adds a method in clustermap interface and HelixClusterManager implementation to indicate a data node is on FULL_AUTO or not by returning true or false.

When turning on FULL_AUTO, we would first reconstruct resource so that a data node can have partitions from only one resource. Then we update the resource's ideal state to change the RebalanceMode to FULL_AUTO. In this way, a data node would know if all the partitions assigned to it are on FULL_AUTO mode or not.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -1610,6 +1610,9 @@ public class HelixClusterManager implements ClusterMap {
     }
   }
 
+  /**
+   * Properties for Resource.
+   */
   static class ResourceProperty {
     final String name;
     final IdealState.RebalanceMode rebalanceMode;
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -517,7 +517,7 @@ public class HelixClusterManager implements ClusterMap {
     }
     // Make sure all the resources turned on FULL_AUTO mode
     String resourceName = resourceNames.iterator().next();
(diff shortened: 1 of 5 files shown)


ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixAdmin.java
```diff
@@ -292,9 +292,10 @@ void setReplicaStoppedState(AmbryPartition partition, String instance, boolean i
     triggerInstanceConfigChangeNotification(tagAsInit);
   }
 
-  void removeResourceIdealState(String clusterName, String resourceName) {
+  void removeResourceIdealState(String clusterName, String resourceName) throws Exception {
```
SophieGuo410, line 295: nit: java docs, and cluster Name is never used.

SophieGuo410 (approved):
lgtm!

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -489,6 +493,38 @@ public void registerClusterMapListener(ClusterMapChangeListener clusterMapChange
...
+    if (resourceNames == null || resourceNames.isEmpty()) {
+      logger.trace("DataNode {} doesn't have any resources", instanceName);
+      return false;
+    }
+    // Before turn on FULL_AUTO, we have to make sure each host only stores partitions from one resource
+    if (resourceNames.size() != 1) {
+      return false;
+    }
+    // Make sure all the resources turned on FULL_AUTO mode
+    for (String resourceName : resourceNames) {
```
Arun-LinkedIn, line 519: nit: We may not need for loop since resourceNames size is 1 at this point.

Arun-LinkedIn (approved):

PR #2372, opened 2023-02-16
[Efficient Metadata Operations] Add logic to handle partially sealed partition in ClusterParticipant.

ankagrawal:
Before this PR, the ClusterParticipant implementations were ignoring the PARTIALLY_SEALED state. This PR adds the logic so ClusterParticipant can handle the PARTIALLY_SEALED state and update the appropriate backend.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -155,14 +155,34 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
     synchronized (helixAdministrationLock) {
       DataNodeConfig config = getDataNodeConfig();
       String partitionId = replicaId.getPartitionId().toPathString();
-      boolean success = true;
-      // TODO Handle partial seal case.
-      if (replicaSealStatus != ReplicaSealStatus.SEALED && config.getSealedReplicas().remove(partitionId)) {
-        logger.trace("Removing the partition {} from sealedReplicas list", partitionId);
-        success = dataNodeConfigSource.set(config);
-      } else if (replicaSealStatus == ReplicaSealStatus.SEALED && config.getSealedReplicas().add(partitionId)) {
-        logger.trace("Adding the partition {} to sealedReplicas list", partitionId);
-        success = dataNodeConfigSource.set(config);
+      boolean success = false;
+      boolean configChanged;
+      switch (replicaSealStatus) {
+        case NOT_SEALED:
+          configChanged = config.getSealedReplicas().remove(partitionId);
+          configChanged = config.getPartiallySealedReplicas().remove(partitionId) || configChanged;
+          if (configChanged) {
+            logger.trace("Removing partition {} from sealed lists to mark it as NOT_SEALED", partitionId);
+            success = dataNodeConfigSource.set(config);
+          }
+          break;
+        case PARTIALLY_SEALED:
+          configChanged = config.getPartiallySealedReplicas().add(partitionId);
+          configChanged = config.getSealedReplicas().remove(partitionId) || configChanged;
+          if (configChanged) {
+            logger.trace("Adding the partition {} to partially sealed list to mark it as PARTIALLY_SEALED",
+                partitionId);
+            success = dataNodeConfigSource.set(config);
+          }
+          break;
+        case SEALED:
+          configChanged = config.getSealedReplicas().add(partitionId);
+          configChanged = config.getPartiallySealedReplicas().remove(partitionId) || configChanged;
+          if (configChanged) {
+            logger.trace("Adding the partition {} to sealedReplicas list", partitionId);
+            success = dataNodeConfigSource.set(config);
+          }
+          break;
       }
       logger.trace("Set sealed state of partition {} is completed", partitionId);
       return success;
@@ -214,6 +234,11 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
   public List<String> getSealedReplicas() {
     // TODO refactor these getter methods to return set instead of list.
     return new ArrayList<>(getDataNodeConfig().getSealedReplicas());
+  }
+
+  @Override
+  public List<String> getPartiallySealedReplicas() {
+    return new ArrayList<>(getDataNodeConfig().getPartiallySealedReplicas());
   }
 
   /**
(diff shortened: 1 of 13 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -147,21 +147,34 @@ public void registerPartitionStateChangeListener(StateModelListenerType listener
...
-        success = dataNodeConfigSource.set(config);
-      } else if (isSealed && config.getSealedReplicas().add(partitionId)) {
-        logger.trace("Adding the partition {} to sealedReplicas list", partitionId);
-        success = dataNodeConfigSource.set(config);
+      boolean success = false;
+      switch (replicaSealStatus) {
+        case NOT_SEALED:
+          if (removePartitionFromSealedReplicasConfig(partitionId, config) && removePartitionFromPartiallySealedReplicasConfig(partitionId, config)) {
+            logger.trace("Removing partition {} from sealed lists to mark it as NOT_SEALED", partitionId);
+            success = dataNodeConfigSource.set(config);
```
JingQianCloud, line 163: The code logic/design is changed.
If we want to set the replica to NOT_SEALED state, in the old code, only when the partition is at sealed state (config.getSealedReplicas().remove(partitionId) returns true), we will call dataNodeConfigSource.set(config).

In the current code, even the partition currently is already at NOT_SEALED state, we still call dataNodeConfigSource.set(config)
Will the un-necessary/repeated set call any problem?

ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixParticipant.java
```diff
@@ -100,11 +101,18 @@ public void participate(List<AmbryStatsReport> ambryStatsReports, AccountStatsSt
...
-    } else {
-      sealedReplicas.remove(replicaId);
+  public boolean setReplicaSealedState(ReplicaId replicaId, ReplicaSealStatus replicaSealStatus) {
+    switch (replicaSealStatus) {
+      case SEALED:
+        sealedReplicas.add(replicaId);
+        partiallySealedReplicas.remove(replicaId);
+        break;
+      case PARTIALLY_SEALED:
+        partiallySealedReplicas.add(replicaId);
```
JingQianCloud, line 111: Are we supposed to add sealedReplicas.remove(replicaId)?

ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixParticipant.java
```diff
@@ -100,11 +101,18 @@ public void participate(List<AmbryStatsReport> ambryStatsReports, AccountStatsSt
...
+        break;
+      case NOT_SEALED:
+        partiallySealedReplicas.remove(replicaId);
```
JingQianCloud, line 114: Do we also need sealedReplica.remove(replicaId)?
It's not necessary we will go from not-sealed -> partial sealed-> sealed?
It's possible that we jump from not-sealed to sealed, right? Can we have test cases for it if we don't have them yet.

ambry-server/src/test/java/com/github/ambry/server/ParticipantsConsistencyTest.java
```diff
@@ -184,9 +197,18 @@ public void participate(List<AmbryStatsReport> ambryStatsReports, AccountStatsSt
...
-      if (isSealed) {
-        sealedReplicas.add(replicaId.getReplicaPath());
+    public boolean setReplicaSealedState(ReplicaId replicaId, ReplicaSealStatus replicaSealStatus) {
+      String replicaPath = replicaId.getReplicaPath();
+      switch (replicaSealStatus) {
+        case SEALED:
+          sealedReplicas.add(replicaPath);
+          partiallySealedReplicas.remove(replicaPath);
+          break;
+        case PARTIALLY_SEALED:
```
JingQianCloud, line 207: Same questions here.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
ankagrawal: Fixed it.

ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixParticipant.java
ankagrawal: fixed.

ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixParticipant.java
ankagrawal: fixed.

ambry-server/src/test/java/com/github/ambry/server/ParticipantsConsistencyTest.java
ankagrawal: fixed.

ambry-clustermap/src/test/java/com/github/ambry/clustermap/MockHelixParticipant.java
ankagrawal: Added tests too.

Arun-LinkedIn (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -155,14 +155,34 @@ public boolean setReplicaSealedState(ReplicaId replicaId, ReplicaSealStatus repl
...
+      switch (replicaSealStatus) {
+        case NOT_SEALED:
+          configChanged = config.getSealedReplicas().remove(partitionId);
+          configChanged = config.getPartiallySealedReplicas().remove(partitionId) || configChanged;
+          if (configChanged) {
+            logger.trace("Removing partition {} from sealed lists to mark it as NOT_SEALED", partitionId);
+            success = dataNodeConfigSource.set(config);
+          }
+          break;
+        case PARTIALLY_SEALED:
```
JingQianCloud, line 169: Just double check. At this moment, no code will call HelixParticipant.setReplicaSealedState, right?

ankagrawal: There is code that calls `HelixParticipant.setReplicaSealedState` (see ReplicaStatusDelegate.java). But note that it's never called with replicaSealStatus argument set to `ReplicaSealStatus.Partially_Sealed`.

PR #2375, opened 2023-02-17
[Efficient Metadata Operations] Complete the code to handle partial seal notifications from helix.

ankagrawal:
Completes the partial seal handling logic by handling the notifications in HelixClusterManager.java.
Also adds partial logic in the cluster upgrade/bootstrap utilities.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
@@ -193,18 +193,28 @@ public class AmbryPartition implements PartitionId {
     // if the lock could not be taken, that means the state is being updated in another thread. Avoid updating the
     // state in that case.
     if (currentCounterValue > lastUpdatedSealedStateChangeCounter && stateChangeLock.tryLock()) {
+      ReplicaSealStatus mergedReplicaSealStatus = ReplicaSealStatus.NOT_SEALED;
       try {
         lastUpdatedSealedStateChangeCounter = currentCounterValue;
-        boolean isSealed = false;
+        AmbryReplica partiallySealedReplica = null;
         for (AmbryReplica replica : clusterManagerQueryHelper.getReplicaIdsForPartition(this)) {
-          if (replica.isSealed()) {
-            LOGGER.trace("Partition {} is sealed because of replica {} is sealed at sealedStateChangeCounter {}", id,
+          mergedReplicaSealStatus =
+              ReplicaSealStatus.mergeReplicaSealStatus(mergedReplicaSealStatus, replica.getSealedStatus());
+          if (mergedReplicaSealStatus == ReplicaSealStatus.SEALED) {
+            LOGGER.trace("Partition {} is sealed because replica {} is sealed at sealedStateChangeCounter {}", id,
                 replica.getDataNodeId().toString(), lastUpdatedSealedStateChangeCounter);
-            isSealed = true;
             break;
           }
+          if (mergedReplicaSealStatus == ReplicaSealStatus.PARTIALLY_SEALED) {
+            partiallySealedReplica = replica;
+          }
         }
-        state = isSealed ? PartitionState.READ_ONLY : PartitionState.READ_WRITE;
+        if (mergedReplicaSealStatus == ReplicaSealStatus.PARTIALLY_SEALED) {
+          LOGGER.trace(
+              "Partition {} is partially sealed because replica {} is partially sealed at sealedStateChangeCounter {}",
+              id, partiallySealedReplica.getDataNodeId().toString(), lastUpdatedSealedStateChangeCounter);
+        }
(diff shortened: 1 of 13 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartition.java
```diff
@@ -193,18 +193,32 @@ void resolvePartitionState() {
...
+              ReplicaSealStatus.mergeReplicaSealStatus(mergedReplicaSealStatus, replica.getSealedStatus());
+          if (mergedReplicaSealStatus == ReplicaSealStatus.SEALED) {
+            firstSealedReplica = replica;
             break;
           }
+          if (mergedReplicaSealStatus == ReplicaSealStatus.PARTIALLY_SEALED) {
+            partiallySealedReplica = replica;
+          }
+        }
+        if (mergedReplicaSealStatus == ReplicaSealStatus.SEALED) {
```
JingQianCloud, line 212: I think we don't need firstSealedReplica variable and this if condition. 
This if condition is used to print the trace only.
We can just do it in the 

```
          if (mergedReplicaSealStatus == ReplicaSealStatus.SEALED) {
            firstSealedReplica = replica; <-- remove
            <-- add print here.
            break;
          }
```

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -889,6 +889,19 @@ long getPartitionSealedCount() {
       return count;
     }
 
+    /**
+     * @return the count of partitions that are in sealed (read-only) state.
```
JingQianCloud, line 893: nit: comment doesn't match.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1508,23 +1521,25 @@ private void ensurePartitionAbsenceOnNodeAndValidateCapacity(AmbryPartition part
...
+                .get(ClusterMapUtils.PARTITION_STATE));
+        if (replicaSealStatus == ReplicaSealStatus.SEALED) {
+          return  replicaSealStatus;
         }
       } else {
         if (sealedReplicas.contains(partitionName)) {
           return ReplicaSealStatus.SEALED;
         }
       }
       if (partiallySealedReplicas.contains(partitionName)) {
```
JingQianCloud, line 1539: When do we use clusterMapEnablePartitionOverride to overwrite the partition to sealed?
Is that the manual operation used by SRE team?

Just wondering if we need clusterMapEnablePartitionOverride for partial sealed partition as well. If it's just used for SRE team to manual sealing, I guess we don't need the override for partial sealed as your current code.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
ankagrawal: fixed.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
ankagrawal: The cluster map partition is done manually. We add a list of partitions with their seal status in a znode of helix property store. This can be used to force set the status of a partition.

We don't need to override partial-sealed using cluster map override as of now. But in case we need it in the future, it's already supported. The map entry contains not just partition id, but also the state it's being overridden to. So for forcing partial-seal, we just need to mark the partition as "PRW" in the property store.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -869,7 +869,9 @@ long getPartitionCount() {
     long getPartitionReadWriteCount() {
       long count = 0;
       for (AmbryPartition partition : partitionMap.values()) {
-        if (partition.getPartitionState() == PartitionState.READ_WRITE) {
+        // TODO note that currently even partially read write partitions are treated as read write.
```
JingQianCloud, line 872: nit: I have a habit that when I implement a feature, I use FEATURE_NAME_TODO as the comment like EFFICIENT_META_TODO
I guess we may have many TODO in the existing code. It's easy to lose track of the generic TODO.

JingQianCloud (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
ankagrawal: Thats a great idea. It will also make my job easier finding all the left out todos to close the project. I will update all the relevant TODOs in the next PR.

Arun-LinkedIn (approved):
Looks good to me

PR #2377, opened 2023-02-17
Record replication lag in seconds for the originating dc

justinlin-linkedin:
This PR changes the way we record replication lag in seconds metrics. Before this PR, whenever we get a blob from a remote peer, we just record the replication lag in seconds against this peer's datacenter. However, this peer might get this blob from other datacenter. This would skew the replication lag in seconds metric.

This PR fixes this issue.


JingQianCloud (approved):
LGTM

Arun-LinkedIn (approved):

PR #2378, opened 2023-02-17
Reduce Index Segment Size

JingQianCloud:
It's based on https://github.com/linkedin/ambry/pull/2376.
Instead of pre-config the entry size to 115, we monitor the IndexSegment Entry size with the maxEntrySize variable. When we write the IndexSegment, we pad each entry with the maxEntrySize.
There are three cases related to entry size,
1. Sealed Index Segment: The entry size is fixed. Renamed it to PersistedEntrySize.
2. Brand new index segment. maxEntrySize starts with 0. And we monitor the new entry size and update it.
3. Saved index segment but not sealed, after we read from the index segment size, we set the initial maxEntrySize to the entry size used in the file. Then we continue to change maxEntrySize if we have a new entry which is larger than that.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
@@ -76,7 +76,6 @@ class IndexSegment implements Iterable<IndexEntry> {
   static final String INDEX_SEGMENT_FILE_NAME_SUFFIX = "index";
   static final String BLOOM_FILE_NAME_SUFFIX = "bloom";
 
-  private final static int ENTRY_SIZE_INVALID_VALUE = -1;
   private final static int VALUE_SIZE_INVALID_VALUE = -1;
   private static final Logger logger = LoggerFactory.getLogger(IndexSegment.class);
   private final int VERSION_FIELD_LENGTH = 2;
@@ -105,8 +104,8 @@ class IndexSegment implements Iterable<IndexEntry> {
   private int indexSizeExcludingEntries;
   private int firstKeyRelativeOffset;
   private IFilter bloomFilter = null;
-  private int valueSize;
-  private int persistedEntrySize;
+  private int valueSize = VALUE_SIZE_INVALID_VALUE;
+  private int maxEntrySize = 0;
   private short version;
   private Offset prevSafeEndPoint = null;
   // reset key refers to the first StoreKey that is added to the index segment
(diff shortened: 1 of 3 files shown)


snalli:
trying to understand this, have a few questsion will discuss on slack

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -105,8 +104,8 @@ class IndexSegment implements Iterable<IndexEntry> {
   private int indexSizeExcludingEntries;
   private int firstKeyRelativeOffset;
   private IFilter bloomFilter = null;
-  private int valueSize;
-  private int persistedEntrySize;
+  private int valueSize = VALUE_SIZE_INVALID_VALUE;
```
snalli, line 107: what is valueSize ? Size of the value in the key-value pair in the index ?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -105,8 +104,8 @@ class IndexSegment implements Iterable<IndexEntry> {
...
+  private int maxEntrySize = 0;
```
snalli, line 108: why max value 0 ?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -133,11 +132,18 @@ class IndexSegment implements Iterable<IndexEntry> {
     this.factory = factory;
     this.metrics = metrics;
     this.time = time;
+    // When PersistentIndex tries to add one IndexEntry and it decides to create a new IndexSegment,
+    // it provides that new IndexEntry's valueSize and entrySize for the IndexSegment.
```
snalli, line 136: is entrySize = keySize + valueSize ?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -133,11 +132,18 @@ class IndexSegment implements Iterable<IndexEntry> {
...
+    // Inside one IndexSegment, all the index entries have the same valueSize.
```
snalli, line 137: // Inside one IndexSegment, all the index entries have the same valueSize.

i don't get this, i thought all entries in an index-segment are sorted by blob id. 
Is this an additional constraint that all entries must have same valueSize ?
this might disrupt the sorted order...

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
```diff
@@ -258,17 +269,24 @@ int size() {
...
 
   /**
-   * @return The persisted entry size of this segment
+   * @return The entry size of this segment
+   * It's for testing purpose only.
    */
   int getPersistedEntrySize() {
-    return persistedEntrySize;
+    rwLock.readLock().lock();
+    int sz = sealed.get() ? sealedIndex.getPersistedEntrySize() : maxEntrySize;
```
snalli, line 277: so the entry size can change if the index segment is not sealed ? and once its sealed all entry sizes are same ? how is that possible ?

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
JingQianCloud: It's the size of the IndexValue, yes, it's the size of the value in the key-value pair.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
JingQianCloud: It's the initial value. When we have the first entry, will set the entry size of the first entry. Then increase if there is larger entry.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
JingQianCloud: Yes, the JIRA ticket I mentioned have some explanation of it.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
JingQianCloud: Again, valueSize is the size of the IndexValue.
Why will valueSize impact the sorting?
Basically all the Index Value in the same IndexSegment are with the same PersistentIndex.VERSION_X.

ambry-store/src/main/java/com/github/ambry/store/IndexSegment.java
JingQianCloud: Each entry can have different size. That's why we use maxEntrySize to monitor the max entry size.
When we persist, we use the maxEntrySize to do the padding. If the entry is smaller than that size, we pad the zero to it.

snalli: oh, ok that explains why all entries are of the same size.

justinlin-linkedin (approved):
lgtm

snalli (approved):

PR #2379, opened 2023-02-18
Named Blob cleanup exclude valid blobId cases

mikeyang923:
Named Blob cleanup exclude valid blobId cases


ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobsCleanupRunner.java
```diff
@@ -39,13 +42,18 @@ public void run() {
...
       List<StaleNamedBlob> staleResultList = namedBlobDb.pullStaleBlobs().get();
+      Set<String> validBlobIds = namedBlobDb.pullValidBlobIds().get();
       List<StaleNamedBlob> failedResults = new ArrayList<>();
       for (StaleNamedBlob staleResult : staleResultList) {
-        try {
-          router.deleteBlob(staleResult.getBlobId(), "ambry-named-blobs-cleanup-runner").get();
-        } catch (Exception e) {
-          logger.error("Failed to cleanup named stale blob {}", staleResult.toString(), e);
-          failedResults.add(staleResult);
+        if (!validBlobIds.contains(staleResult.getBlobId())) {
```
SophieGuo410, line 48: I'm wondering if we can directly exclude those valid ids when pullStaleBlobs?

SophieGuo410 (approved):

PR #2381, opened 2023-02-22
Reducing the Number of StoreExceptions for checking mount path

maskndafi-linkedin:
The disk healthchecker will check if the mount is accessible every 60s, so if a mount is inaccessible, the logs will be crowded with repeated exceptions of disks being inaccessible. With this change, we reduce the number of exceptions. So, we only throw an exception, if the previous state wasn't MOUNT_NOT_ACCESSIBLE.The initial state for diskHealthStatus will be null to trigger that initial Store Exception, if the mount isn't accessible. Build was successful

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
@@ -531,14 +531,17 @@ public class DiskManager {
 
   /**
    * @throws StoreException if the disk's mount path is inaccessible.
+   *
+   * @return bool - True if mount is accessible otherwise False
    */
-  private void checkMountPathAccessible() throws StoreException {
+  private boolean checkMountPathAccessible() throws StoreException {
     File mountPath = new File(disk.getMountPath());
     if (!mountPath.exists() && diskHealthStatus != DiskHealthStatus.MOUNT_NOT_ACCESSIBLE) {
       metrics.diskMountPathFailures.inc();
       throw new StoreException("Mount path does not exist: " + mountPath + " ; cannot start stores on this disk",
           StoreErrorCodes.Initialization_Error);
     }
+    return mountPath.exists();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


litingulfs (commented):
I'm not super familiar with this code - where do the 180 seconds come from?

maskndafi-linkedin:
> I'm not super familiar with this code - where do the 180 seconds come from?
It's actually 60 seconds. I had mistaken it for 180 seconds
https://github.com/linkedin/ambry/blob/master/ambry-api/src/main/java/com/github/ambry/config/DiskManagerConfig.java#L55

ankagrawal (approved):
lgtm

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -534,7 +534,7 @@ private Map<String, Throttler> getThrottlers(StoreConfig config, Time time) {
    */
   private void checkMountPathAccessible() throws StoreException {
     File mountPath = new File(disk.getMountPath());
-    if (!mountPath.exists()) {
+    if (!mountPath.exists() && diskHealthStatus != DiskHealthStatus.MOUNT_NOT_ACCESSIBLE) {
```
justinlin-linkedin, line 537: I am afraid this is not going to fix the issue. 

So first time the disk checker is running, `diskHealthStatus` is null so it doesn't equal to MOUNT_NOT_ACCESSIBLE, it will throw an exception and change the `diskHealthStatus` to MOUNT_NOT_ACCESSIBLE.

Second time the disk checker is running, `diskHealthStatus` is MOUNT_NOT_ACCESSIABLE, so this method won't throw any exception. However, the createFileChannel method would proceed to create a file and fail. Now the `diskHealthStatus` will be set to `CREATE_EXCEPTION`.

Third time the disk checker is running, `diskHealthStatus` does not equal to MOUNT_NOT_ACCESSIBLE again. And the cycle goes on and on.

I think we need to stop or slow down the frequency of disk checker when we see any exceptions.

maskndafi-linkedin: You're absolutely right. I had that logic in my local branch but hadn't pushed it. In this new commit, checkMountPathAccessible returns whether or not it's accessible. If the function returns False we would then return null in createFileChannel which wouldn't change the state of `diskHealthStatus` and ultimately not return a repeated exception since `diskHealthStatus` is still MOUNT_NOT_ACCESSIBLE. If it's true, we would continue to attempt to create a file channel.

justinlin-linkedin (approved):
LGTM

PR #2382, opened 2023-02-23
Fix timedout request in replica thread

justinlin-linkedin:
In some cases, a droped request will not get a response back from network client. And this is wrong for ReplicaThread since ReplicaThread is expecting a reponse for each request so the RemoteReplicaGroup can be terminated successfully. 

This PR fixes it.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -2187,7 +2187,8 @@ public class ReplicaThread implements Runnable {
         final int pollTimeoutMs = (int) replicationConfig.replicationRequestNetworkPollTimeoutMs;
         List<ResponseInfo> responseInfos = networkClient.sendAndPoll(requestInfos, requestsToDrop, pollTimeoutMs);
         // Add response for dropped request because there is no guarantee that sendAndPoll would return a response for
-        // dropped requests.
+        // dropped requests. Even if the network client returns response infos for dropped requests, onResponse should
+        // still be able to handle duplicate response infos for the same request.
         responseInfos.addAll(responseInfosForTimedOutRequests);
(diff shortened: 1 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -2178,10 +2178,17 @@ public void replicateWithNetworkClient() {
...
+            filterTimedOutRequests(correlationIdToRequestInfo, correlationIdToReplicaGroup);
+        Set<Integer> requestsToDrop = responseInfosForTimedOutRequests.stream()
+            .map(r -> r.getRequestInfo().getRequest().getCorrelationId())
+            .collect(Collectors.toSet());
 
         final int pollTimeoutMs = (int) replicationConfig.replicationRequestNetworkPollTimeoutMs;
         List<ResponseInfo> responseInfos = networkClient.sendAndPoll(requestInfos, requestsToDrop, pollTimeoutMs);
+        // Add response for dropped request because there is no guarantee that sendAndPoll would return a response for
+        // dropped requests.
+        responseInfos.addAll(responseInfosForTimedOutRequests);
```
JingQianCloud, line 2191: What if the sendAndPoll returns response for the dropped requests? We may have two responses for the single requests. 
I guess we reply on the factor that inside the onResponses function, if the correlationId doesn't exist, we do nothing. 
Can we add comment for it?

JingQianCloud (approved):
LGTM. One small comment below.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
justinlin-linkedin: Done.

ankagrawal (approved):

PR #2383, opened 2023-02-24
Apply Ttl Update for named blob put

mikeyang923:
Apply Ttl Update for named blob put


ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
```diff
@@ -70,10 +70,11 @@ CompletableFuture<Page<NamedBlobRecord>> list(String accountName, String contain
    * @param record the {@link NamedBlobRecord}
    * @param state the {@link NamedBlobState}
    * @param isUpsert the {@link Boolean}
+   * @param isTtlUpdate the {@link Boolean}
    * @return a {@link CompletableFuture} that will eventually contain a {@link PutResult} or an exception if an error
    *         occurred.
    */
-  CompletableFuture<PutResult> put(NamedBlobRecord record, NamedBlobState state, Boolean isUpsert);
+  CompletableFuture<PutResult> put(NamedBlobRecord record, NamedBlobState state, Boolean isUpsert, Boolean isTtlUpdate);
```
justinlin-linkedin, line 77: can we not reuse put method here, can we just create a unique method for updating blob state? something like
`CompletableFuture<PutResult> updateBlobState(NamedBlobRecord record, NamedBlobState state)`?

justinlin-linkedin (changes requested):

ambry-api/src/main/java/com/github/ambry/named/NamedBlobDb.java
mikeyang923: Good point, i just updated to use ```updateBlobStateToReady```  instead of ```put```

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -181,7 +183,8 @@ private CompletionStage<String> convertId(String input, RestRequest restRequest,
           // Set named blob state as 'IN_PROGRESS', will set the state to be 'READY' in the ttlUpdate success callback: routerTtlUpdateCallback
           state = NamedBlobState.IN_PROGRESS;
         }
-        conversionFuture = getNamedBlobDb().put(record, state, RestUtils.isUpsertForNamedBlob(restRequest.getArgs())).thenApply(result -> result.getInsertedRecord().getBlobId());
+        conversionFuture = getNamedBlobDb().put(record, state, RestUtils.isUpsertForNamedBlob(restRequest.getArgs())).thenApply(
+            result -> result.getInsertedRecord().getBlobId() + NAMED_BLOBID_VERSION_SEPARATOR + result.getInsertedRecord().getVersion());
```
justinlin-linkedin, line 187: Let's not change the return value of id converter, let's put the version in the RestRequest. We can add an InternalKey in the RestUtils and save the version as the value of this internal key, so that the NamedBlobPutHandler can get the version out.

mikeyang923: Good idea, I just updated with that.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -622,13 +636,27 @@ private PutResult run_put_v2(NamedBlobRecord record, NamedBlobState state, short
...
   }
 
+  private PutResult apply_ttl_update(NamedBlobRecord record, short accountId, short containerId, Connection connection)
+      throws Exception{
+    try (PreparedStatement statement = connection.prepareStatement(TTL_UPDATE_QUERY)) {
+      statement.setInt(1, accountId);
+      statement.setInt(2, containerId);
+      statement.setString(3, record.getBlobName());
+      statement.setLong(4, record.getVersion());
+      statement.executeUpdate();
```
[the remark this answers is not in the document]
mikeyang923: You mean first check whether this record (..., name, version) exist in DB:
```
If yes: then apply the update.
Else: insert in a new row?
```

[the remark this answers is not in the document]
SophieGuo410: I was thinking about if this record is gone, should we consider ttl update failed and throw exception like "this record does not exist"?

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -297,11 +296,12 @@ private boolean isRetriable(Throwable throwable) {
     private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo, String blobId) {
       return buildCallback(frontendMetrics.updateBlobTtlRouterMetrics, convertedBlobId -> {
         // Set the named blob state to be 'READY' after the Ttl update succeed
+        long namedBlobVersion = Long.parseLong(restRequest.getArgs().getOrDefault(RestUtils.InternalKeys.NAMED_BLOB_VERSION, 0L).toString());
```
justinlin-linkedin, line 299: let's not return a default value for this one, since a default value is not correct. We should throw an exception when the version is missing in the restRequest.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -337,6 +343,15 @@ public CompletableFuture<PutResult> put(NamedBlobRecord record, NamedBlobState s
         }, null);
   }
 
+  @Override
+  public CompletableFuture<PutResult> updateBlobStateToReady(NamedBlobRecord record) {
+    return executeTransactionAsync(record.getAccountName(), record.getContainerName(), true,
+        (accountId, containerId, connection) -> {
+          logger.info("Updating to READY for Named Blob: {}", record);
```
justinlin-linkedin, line 350: use trace instead of info.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -337,6 +343,15 @@ public CompletableFuture<PutResult> put(NamedBlobRecord record, NamedBlobState s
         }, null);
   }
 
+  @Override
+  public CompletableFuture<PutResult> updateBlobStateToReady(NamedBlobRecord record) {
+    return executeTransactionAsync(record.getAccountName(), record.getContainerName(), true,
```
justinlin-linkedin, line 348: we should have a metric for this operation just like any other operations.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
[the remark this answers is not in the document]
mikeyang923: Yeah, I can do that, @justinlin-linkedin how do you think?
I think maybe that's why initially I am just inserting a new row instead of updating the existing one. So that it doesn't need to check and then update.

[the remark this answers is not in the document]
justinlin-linkedin: we should check out the return value of `executeUpdate`. It should not be 0. If the return value of `executeUpdate` is 0, then the record doesn't exist.

justinlin-linkedin (commented):
I don't see the unit test for apply ttl, could you please add some unit test and integration test for this new method.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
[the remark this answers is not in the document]
mikeyang923: Done, @SophieGuo410 @justinlin-linkedin can you please take another look?

mikeyang923:
> I don't see the unit test for apply ttl, could you please add some unit test and integration test for this new method.

Added in latest commit

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
mikeyang923: Done

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Done.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
mikeyang923: Added.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -297,11 +296,17 @@ private boolean isRetriable(Throwable throwable) {
     private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo, String blobId) {
       return buildCallback(frontendMetrics.updateBlobTtlRouterMetrics, convertedBlobId -> {
         // Set the named blob state to be 'READY' after the Ttl update succeed
+        if (!restRequest.getArgs().containsKey(RestUtils.InternalKeys.NAMED_BLOB_VERSION)) {
+          throw new RestServiceException(
+              "Internal key " + RestUtils.InternalKeys.NAMED_BLOB_VERSION + " is required in Named Blob TTL update callback!",
+              RestServiceErrorCode.InvalidArgs);
```
justinlin-linkedin, line 302: let's not return invalid argument here, since it would return in 400 response, we should return 500 which is internal server in this case.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -297,11 +296,17 @@ private boolean isRetriable(Throwable throwable) {
...
+        }
+        long namedBlobVersion = Long.parseLong(restRequest.getArgs().get(RestUtils.InternalKeys.NAMED_BLOB_VERSION).toString());
```
justinlin-linkedin, line 304: you can just do 
```
long namedBlobVersion = (long) restRequest.getArgs().get(NAMED_BLOB_VERSION);
```

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
```diff
@@ -141,6 +145,25 @@ public void testPullAndCleanStaleNamedBlobs() throws Exception {
...
+
+    long expirationTimeMs = calendar.getTimeInMillis() + 5000;
+    final NamedBlobRecord record = new NamedBlobRecord(account.getName(), container.getName(), "blobName", id, expirationTimeMs);
+    PutResult putResult = namedBlobDb.put(record, NamedBlobState.IN_PROGRESS, true).get();
+
+    record.setVersion(putResult.getInsertedRecord().getVersion());
+    namedBlobDb.updateBlobStateToReady(record).get();
+
+    NamedBlobRecord namedBlobRecord = namedBlobDb.get(account.getName(), container.getName(), "blobName").get();
+    assertEquals("Blob Id is not matched with the record", id, namedBlobRecord.getBlobId());
```
justinlin-linkedin, line 164: we should check that before the update we can't get anything out, and then with the update, we can get the record out.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
mikeyang923: Updated. Can you take another look?

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
mikeyang923: Done.

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
mikeyang923: I just added the check in the [integration test](https://github.com/linkedin/ambry/pull/2383/files#diff-c568b3a93a06b59db49d461110e676415b73b4dfb1b7c9035afd1929e61bacc3R531).
Here it's using mock all the places in unit test

justinlin-linkedin (approved):

PR #2384, opened 2023-02-27
Listen to Helix Ideal states in Aggregated view as well

Arun-LinkedIn:
When we move Ambry to Full auto, we need to Helix Ideal states for getting the assigned Partition to Replica mapping. In addition, we also need it for getting the Resource configs to know if a resource is in full-auto or semi-auto mode. 

This PR has following changes:

1. Currently we are only listening to Ideal states when using Non-aggregated view. Add changes to listen to helix ideal state in aggregated view as well.
2. Use Ideal states to get the Resource to Partition mapping in both aggregated and non-aggregated views.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -87,9 +87,6 @@ public class HelixClusterManager implements ClusterMap {
   private final Map<String, ConcurrentHashMap<String, String>> partitionToResourceNameByDc = new ConcurrentHashMap<>();
   private final Map<String, AtomicReference<RoutingTableSnapshot>> dcToRoutingTableSnapshotRef =
       new ConcurrentHashMap<>();
-  // A map of partition to resource names used in aggregated cluster view.
-  private final AtomicReference<ConcurrentHashMap<String, Set<String>>> globalPartitionToResourceNamesRef =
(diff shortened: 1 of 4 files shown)


justinlin-linkedin (approved):
LGTM

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -758,36 +737,20 @@ public List<AmbryReplica> getReplicaIdsByState(AmbryPartition partition, Replica
...
-          RoutingTableSnapshot routingTableSnapshot = dcToRoutingTableSnapshotRef.get(dc).get();
-          String partitionPath = partition.toPathString();
-          routingTableSnapshot.getInstancesForResource(resourceName, partitionPath, state.name())
-              .stream()
-              .map(instanceConfig -> instanceNameToAmbryDataNode.get(instanceConfig.getInstanceName()))
-              .map(dataNode -> ambryDataNodeToAmbryReplicas.get(dataNode).get(partition.toPathString()))
-              .filter(Objects::nonNull)
-              .forEach(replicas::add);
-        }
+      List<String> dcs = dcName != null ? Collections.singletonList(dcName)
```
ankagrawal, line 740: nit: what this method really does is return all replica ids, not replica ids by state. We should think about renaming it to make it clear.

ankagrawal (approved):

PR #2385, opened 2023-02-28
Release Netty byte Buf for quota rejected requests

Arun-LinkedIn:
For outgoing requests, the Netty ByteBuf would be released (reference counter decremented by 1) in the Netty layer. But, for rejects that are rejected due to Quota, they wouldn't reach Network layer. Hence, we would have release it ourselves.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
@@ -197,8 +197,8 @@ public class QuotaAwareOperationController extends OperationController {
             break;
           case REJECT:
             quotaAvailable = false;
-            logger.warn(
-                "Rejecting request for quota resource {} because of reject recommendation.", quotaResource.toString());
+            logger.warn("Rejecting request for quota resource {} because of reject recommendation.",
+                quotaResource.toString());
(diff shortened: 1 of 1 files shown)


justinlin-linkedin (approved):
sweat

ambry-router/src/main/java/com/github/ambry/router/QuotaAwareOperationController.java
```diff
@@ -252,8 +252,11 @@ private void pollQuotaExceedAllowedRequestsIfAny(List<RequestInfo> requestsToSen
 
   @Override
   protected List<ResponseInfo> getNonQuotaCompliantResponses() {
-    List<ResponseInfo> nonCompliantResponses = nonCompliantRequests.
-        stream().map(requestInfo -> new ResponseInfo(requestInfo, true)).collect(Collectors.toList());
+    List<ResponseInfo> nonCompliantResponses = nonCompliantRequests.stream().map(requestInfo -> {
+      // Release the Netty ByteBuf here since they are not sent out to Network layer.
+      requestInfo.getRequest().release();
```
ankagrawal, line 257: Good catch.
Btw, this means that a memory leak would happen only when requests are rejected. In the current implementation, rejection happens only due to storage quota.

ankagrawal (approved):

PR #2386, opened 2023-02-28
[DR] Use static map to discover VCR nodes

snalli:
It's becoming difficult to debug VCR issues because the current setup uses HelixVcrCluster to dynamically discover vcr nodes and then randomly picks a node to talk to. It is also difficult to play around with VCR configs and understand what works because one has to change configs in all VCR nodes and we don't know which VCR node will picked up next time we start a DR node.

This patch adds a support for reading files to discover vcr nodes and allows us to fix a vcr node for an experiment or debugging.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/StaticVcrClustermap.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/StaticVcrClustermap.java
@@ -1,0 +1,85 @@
+/**
+ * Copyright 2023 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.clustermap;
+
+import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.network.Port;
+import com.github.ambry.network.PortType;
+import java.util.List;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Class to parse and store static vcr clustermap, useful for debugging vcr node issues.
+ * {
+ *   datacenter: [
+ *    {
+ *      datacenter_name: "ei4",
+ *      vcr_nodes: [
+ *        {
+ *          hostname: "..",
+ *          plaintext_port: "..",
+ *          ssl_port: "..",
+ *          http2_port: ".."
+ *        },
+ *        {
+ *          ...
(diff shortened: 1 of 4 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/StaticVcrClustermap.java
```diff
@@ -0,0 +1,85 @@
...
+ *        },
+ *        {
+ *          ...
+ *        }
+ *      ]
+ *    }
+ *   ]
+ * }
+ */
+public class StaticVcrClustermap {
```
ankagrawal, line 46: We already have a static cluster map defined in the code base. Why can't we extend or refactor that instead of creating custom maps and parsers?

snalli: The code changes required are just too fragmented and intrusive. Take a look starting from protected AmbryLIServer createInstance L159. The current approach is the easiest and does not disrupt anything and is primarily for quick debugging and experiments with VCR configs.

ankagrawal: Having noted my concern, I will leave it to Jing and you to take the final call.

JingQianCloud: If we want to implement the static cluster map for VCR, I think we should implements ClusterMap interface as other types of cluster maps do.
Looks like to me, this StaticVcrClustermap is not a complete ClusterMap implementation.

snalli: Extended clusterMap. Most methods are empty since I don't need them atm.

snalli: I could extend StaticClusterManager but it doesn't exactly fit my use case. There is no mapping from partitions to VCR nodes. Any VCR node is capable of recovering any partition.

snalli: Tested in ei4. I am able to contact a single vcr and recover partitions

JingQianCloud: Sanketh,
As you also mentioned, this is not a really static cluster manager and the caller usage model is different too.
Since iI=t's a testing code. do we really need check in that? 
As you test it in the EI, can we use it as WIP code and patch the prod server for testing?

snalli: The static side of DR-VCR path has been incomplete and this is an attempt to fill that gap. Until the DR effort is complete, there will be more cases where this will be needed. I was working around helixVcr in EI for a couple of months by shutting down all but one vcr nodes. But that's a bad idea in prod. The priority is to get to the root of 429s in prod-vcr. The trade-off here is spending more days extending current code, fixing tests and ctors Vs. having a decent enough code but spending more time tuning VCR and investigating 429. I do not like the idea of prod-patching because it limits how often I can deploy to DR node and how easily I can configure it. In comparison to the task at hand, I do not think the concerns presented here are pressing enough but I will go the patching route since it has been insisted.

PR #2391, opened 2023-03-03
Fix bugs for datset version support

SophieGuo410:
This pr fixed two issues:
1. check if userTag is null.
2. set the ttl instead of expiration time to blobProperties.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
@@ -384,8 +384,10 @@ public class NamedBlobPutHandler {
       if (RestUtils.isDatasetVersionUpload(restRequest.getArgs())) {
         Dataset dataset = (Dataset) restRequest.getArgs().get(RestUtils.InternalKeys.TARGET_DATASET);
         Map<String, String> userTags = dataset.getUserTags();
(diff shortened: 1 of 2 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -524,7 +526,10 @@ private void addDatasetVersion(BlobProperties blobProperties, RestRequest restRe
         restResponseChannel.setHeader(RestUtils.Headers.TARGET_DATASET_NAME, datasetName);
         restResponseChannel.setHeader(RestUtils.Headers.TARGET_DATASET_VERSION, datasetVersionRecord.getVersion());
         long newExpirationTimeMs = datasetVersionRecord.getExpirationTimeMs();
-        blobProperties.setTimeToLiveInSeconds(newExpirationTimeMs);
+        if (expirationTimeMs != Utils.Infinite_Time || expirationTimeMs != newExpirationTimeMs) {
```
JingQianCloud, line 529: Can you add comment to explain the difference between the dataset creation time, expiration and ttl?

JingQianCloud (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
SophieGuo410: Updated.

PR #2392, opened 2023-03-06
Support delete dataset

SophieGuo410:
Support delete dataset in account service and request handler.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
@@ -31,6 +31,7 @@ import java.io.IOException;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.SQLIntegrityConstraintViolationException;
 import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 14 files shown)


ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -754,6 +795,60 @@ private void executeAddDatasetStatement(PreparedStatement statement, int account
...
+   * Execute updateDatasetSqlIfExpired statement.
+   * @param statement the updateDatasetSqlIfExpired statement.
+   * @param accountId the id for the parent account.
+   * @param containerId the id of the container.
+   * @param dataset the {@link Dataset}.
+   * @return the number of rows been updated.
+   * @throws SQLException
+   * @throws AccountServiceException
+   */
+  private int executeUpdateDatasetSqlIfExpiredStatement(PreparedStatement statement, int accountId, int containerId, Dataset dataset)
```
justinlin-linkedin, line 808: what is the purpose of this statement? when an dataset version already exist, we update it? or when it already exist but expired, we can reuse the version but with different metadata?

SophieGuo410: It will only affect the dataset, not the dataset version. So when a dataset has been marked as expired, we will support user to create dataset with same primary key before it gets auto purged. But if the dataset is still exist, they can only update the dataset metadata by calling the update dataset API. 
And for dataset version, I'm planning to do the similar thing.

justinlin-linkedin (approved):

PR #2398, opened 2023-03-09
Add metrics for request timeout to server since it can change dynamically

Arun-LinkedIn:


justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -296,6 +296,13 @@ private void cleanupExpiredInflightRequests(
         onErrorResponse(requestInfo.getReplicaId(),
             RouterUtils.buildTimeoutException(correlationId, requestInfo.getReplicaId().getDataNodeId(), blobId));
         requestRegistrationCallback.registerRequestToDrop(correlationId);
+        if (routerRequestExpiryReason == RouterUtils.RouterRequestExpiryReason.ROUTER_SERVER_NETWORK_CLIENT_TIMEOUT) {
+          // Since the network timeout value can change based on load, record it in metrics.
```
ankagrawal, line 300: Its the same code in three classes. Can we abstract it out into a method and put it in RouterUtils.

```
  public static void logTimeoutMetrics(RouterRequestExpiryReason routerRequestExpiryReason, NonBlockingRouterMetrics routerMetrics, RequestInfo requestInfo) {
    if (routerRequestExpiryReason == RouterUtils.RouterRequestExpiryReason.ROUTER_SERVER_NETWORK_CLIENT_TIMEOUT) {
      // Since the network timeout value can change based on load, record it in metrics.
      routerMetrics.dynamicRequestNetworkTimeoutMs.update(requestInfo.getNetworkTimeOutMs());
      routerMetrics.requestNetworkTimeoutCount.inc();
    } else if (routerRequestExpiryReason == RouterUtils.RouterRequestExpiryReason.ROUTER_REQUEST_TIMEOUT) {
      routerMetrics.requestFinalTimeoutCount.inc();
    }
  }
```

Arun-LinkedIn: Done..

ankagrawal (approved):

PR #2403, opened 2023-03-13
[config] Add codestyle.xml

snalli:
Borrowed from Helix team.
Instructions to import codestyle.xml into IntelliJ IDE:

1. Goto: LI IntelliJ IDEA 2022.2 > Settings > Editor > Code Style > Java
2. Click Gear icon next to "Scheme"
3. Select "Import scheme" from drop-down and then select "IntelliJ IDEA code style XML"
4. Select file: ambry/config/codestyle.xml and click OK
5. Click Apply and restart IDE

### DIFF
--- a/config/codestyle.xml
+++ b/config/codestyle.xml
@@ -1,0 +1,408 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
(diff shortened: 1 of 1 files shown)


snalli:
Found that there is a built-in "Linkedin style" option in intellij. So we don't need a new style.

snalli:
Found that there is a built-in "Linkedin style" option in intellij. So we don't need a new style.

PR #2407, opened 2023-03-16
[Efficient Metadata Operations] Enable replicas to be marked as partially sealed in DataNodeConfig.

ankagrawal:
Before this change, the BlobStore was marking replicas that met partial seal conditions, as Unsealed. This PR changes ReplicaStatusDelegate to mark partially sealed partitions in Helix DatanodeConfig.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/clustermap/ReplicaStatusDelegate.java
+++ b/ambry-api/src/main/java/com/github/ambry/clustermap/ReplicaStatusDelegate.java
@@ -44,9 +44,7 @@ public class ReplicaStatusDelegate {
    * @return {@code true} if replica is successfully partial-sealed. {@code false} if not.
    */
   public boolean partialSeal(ReplicaId replicaId) {
-    // TODO Efficient_Metadata_Operations_TODO Once the partial seal logic is complete end to end, this should set the
-    //  partial seal state.
-    return clusterParticipant.setReplicaSealedState(replicaId, ReplicaSealStatus.NOT_SEALED);
+    return clusterParticipant.setReplicaSealedState(replicaId, ReplicaSealStatus.PARTIALLY_SEALED);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ambry-api/src/main/java/com/github/ambry/clustermap/ReplicaStatusDelegate.java
```diff
@@ -46,7 +46,7 @@ public boolean seal(ReplicaId replicaId) {
   public boolean partialSeal(ReplicaId replicaId) {
     // TODO Efficient_Metadata_Operations_TODO Once the partial seal logic is complete end to end, this should set the
```
Arun-LinkedIn, line 47: nit: We can remove TODO

Arun-LinkedIn (approved):

ambry-api/src/main/java/com/github/ambry/clustermap/ReplicaStatusDelegate.java
```diff
@@ -46,7 +46,7 @@ public boolean seal(ReplicaId replicaId) {
...
     //  partial seal state.
-    return clusterParticipant.setReplicaSealedState(replicaId, ReplicaSealStatus.NOT_SEALED);
+    return clusterParticipant.setReplicaSealedState(replicaId, ReplicaSealStatus.PARTIALLY_SEALED);
```
JingQianCloud, line 49: Before the partial sealed code is fully ready, how do we protect the side effect it may rise after we change the state to PARTIALLY_SEALED?

ankagrawal: The only side effect will be that some replicas are marked as PARTIALLY_SEALED in helix. But this should not affect anything else. The PutOperation will still pick all the writable partitions only which are the partitions with state `!=PartitionState.READ_ONLY`.

ambry-api/src/main/java/com/github/ambry/clustermap/ReplicaStatusDelegate.java
ankagrawal: fixed.

PR #2409, opened 2023-03-16
Fix typo in GetBlobOperation so we dont force set unavailability error when blob is deleted

justinlin-linkedin:
There is a typo in GetBlobOperation that would mistakenly set the the chunkException to AmbryUnavailable when the exception should be either BlobDeleted or BlobExpired. This PR fix that typo and also adds some test cases.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -808,6 +808,7 @@ class GetBlobOperation extends GetOperation {
       logger.trace("BlobId {}: Retry for chunk Id: {}, failed attempts: {}, initializedTime: {}", blobId, chunkBlobId,
           failedAttempts, initializedTimeMs);
       chunkException = null;
+      chunkCompleted = false;
       chunkOperationTracker = getOperationTracker(chunkBlobId.getPartition(), chunkBlobId.getDatacenterId(),
           RouterOperation.GetBlobOperation, chunkBlobId);
       progressTracker = new ProgressTracker(chunkOperationTracker);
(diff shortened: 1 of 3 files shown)


ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -1992,7 +2033,7 @@ public void testDecompressContent() throws Exception {
 
     // Generate the test compressed buffer.
     ByteBuffer sourceBuffer =
-        ByteBuffer.wrap("Compression unit test to test compression.".getBytes(StandardCharsets.UTF_8));
+        ByteBuffer.wrap("Compression unit test to test compression." .getBytes(StandardCharsets.UTF_8));
```
JingQianCloud, line 2036: Is the space a typo?

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -808,6 +808,7 @@ void resetForRetry() {
       logger.trace("BlobId {}: Retry for chunk Id: {}, failed attempts: {}, initializedTime: {}", blobId, chunkBlobId,
           failedAttempts, initializedTimeMs);
       chunkException = null;
+      chunkCompleted = false;
```
JingQianCloud, line 811: Is that possible to add the test for the retry?

Arun-LinkedIn (approved):
No other comments from me apart from Jing's comment.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
justinlin-linkedin: updated, the formatter is messing up my PR here.

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
justinlin-linkedin: I added in the same test to make sure that there is no retries in this case.

JingQianCloud (approved):

PR #2411, opened 2023-03-16
[Efficient Metadata Operations] Add method in ClusterMap to check if partition has enough eligible replicas for put operation.

ankagrawal:
This method will be used in PutOperation to check if the reserved partition for MetadataChunk is available when metadata chunk is ready to be written.


JingQianCloud (approved):
LGTM

Arun-LinkedIn (approved):

PR #2412, opened 2023-03-16
Support retentionTime at dataset level

SophieGuo410:
Previously I treat expiration time of dataset as the dataset life span as well as the default dataset version life span. But after a second thought, I feel like treat expiration time(retention time) of dataset as the default ttl which all dataset version can be inherited from is better. And due to for all blobs we will provide ttl value or -1, it's hard for use to determine if user provided dataset version ttl or not, so I've added this header x-ambry-dataset-version-ttl-enabled to help identify if user want the dataset version ttl override the default dataset level ttl. By default it will be disabled and all the dataset version ttl are inherited from the default dataset level ttl.

### DIFF
--- a/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
+++ b/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
@@ -39,7 +39,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import org.json.JSONArray;
@@ -733,7 +732,7 @@ public class MySqlAccountServiceIntegrationTest {
     //add dataset with must provide info
     Dataset dataset =
         new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME_BASIC).setVersionSchema(
-            Dataset.VersionSchema.MONOTONIC).setExpirationTimeMs(-1).build();
+            Dataset.VersionSchema.MONOTONIC).setRetentionTime(-1).build();
     mySqlAccountStore.addDataset(testAccount.getId(), testContainer.getId(), dataset);
     Dataset datasetFromMysql =
         mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
@@ -745,7 +744,7 @@ public class MySqlAccountServiceIntegrationTest {
     int retentionCount = 5;
     dataset = new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME).setVersionSchema(
             Dataset.VersionSchema.TIMESTAMP)
-        .setExpirationTimeMs(-1)
+        .setRetentionTime(-1)
         .setUserTags(userTags)
         .setRetentionCount(retentionCount)
         .build();
@@ -777,15 +776,14 @@ public class MySqlAccountServiceIntegrationTest {
         datasetFromMysql.getRetentionCount());
 
     //update the expiration time.
-    long expirationTimeInMs = Utils.addSecondsToEpochTime(System.currentTimeMillis(), 30);
+    long datasetTtl = 30;
     datasetForUpdate =
-        new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME).setExpirationTimeMs(
-            expirationTimeInMs).build();
+        new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME).setRetentionTime(datasetTtl)
+            .build();
     mySqlAccountStore.updateDataset(testAccount.getId(), testContainer.getId(), datasetForUpdate);
     datasetFromMysql = mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
         testContainer.getName(), DATASET_NAME);
-    assertEquals("Mismatch in updated expirationTimeInMst", (Long) expirationTimeInMs,
-        datasetFromMysql.getExpirationTimeMs());
+    assertEquals("Mismatch in updated dataset ttl", (Long) datasetTtl, datasetFromMysql.getRetentionTime());
 
     //update the user tags
     userTags.clear();
@@ -809,41 +807,26 @@ public class MySqlAccountServiceIntegrationTest {
     }
 
     //Add a dataset with limited ttl
-    expirationTimeInMs = Utils.addSecondsToEpochTime(System.currentTimeMillis(), 30);
     dataset =
         new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME_WITH_TTL).setVersionSchema(
                 Dataset.VersionSchema.TIMESTAMP)
-            .setExpirationTimeMs(expirationTimeInMs)
+            .setRetentionTime(datasetTtl)
             .setUserTags(userTags)
             .setRetentionCount(retentionCount)
             .build();
     mySqlAccountStore.addDataset(testAccount.getId(), testContainer.getId(), dataset);
 
     //update expiration to permanent.
-    expirationTimeInMs = -1;
+    datasetTtl = -1;
     datasetForUpdate =
-        new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME_WITH_TTL).setExpirationTimeMs(
-            expirationTimeInMs).build();
+        new DatasetBuilder(testAccount.getName(), testContainer.getName(), DATASET_NAME_WITH_TTL).setRetentionTime(
+            datasetTtl).build();
     mySqlAccountStore.updateDataset(testAccount.getId(), testContainer.getId(), datasetForUpdate);
     datasetFromMysql = mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
         testContainer.getName(), DATASET_NAME_WITH_TTL);
-    assertEquals("Mismatch in updated expirationTimeInMst", (Long) expirationTimeInMs,
-        datasetFromMysql.getExpirationTimeMs());
+    assertEquals("Mismatch in updated dataset ttl", (Long) datasetTtl, datasetFromMysql.getRetentionTime());
 
-    //delete a permanent dataset.
-    mySqlAccountStore.deleteDataset(testAccount.getId(), testContainer.getId(), DATASET_NAME_WITH_TTL);
-    try {
-      mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
-          testContainer.getName(), DATASET_NAME_WITH_TTL);
-      fail("Should fail due to the dataset already expired");
-    } catch (AccountServiceException e) {
-      assertEquals("Mistmatch on error code", AccountServiceErrorCode.Deleted, e.getErrorCode());
-    }
-
-    //delete a dataset with future expirationTimeMs
-    long oldExpirationTime =
-        mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
-            testContainer.getName(), DATASET_NAME).getExpirationTimeMs();
+    //delete a dataset
     mySqlAccountStore.deleteDataset(testAccount.getId(), testContainer.getId(), DATASET_NAME);
     try {
       mySqlAccountStore.getDataset(testAccount.getId(), testContainer.getId(), testAccount.getName(),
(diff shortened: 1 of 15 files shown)


ambry-api/src/main/java/com/github/ambry/account/Dataset.java
```diff
@@ -61,8 +61,8 @@ public class Dataset {
   static final String CONTAINER_NAME_KEY = "containerName";
   static final String DATASET_NAME_KEY = "datasetName";
   static final String JSON_VERSION_SCHEMA_KEY = "versionSchema";
-  static final String JSON_EXPIRATION_TIME_KEY = "expirationTimeMs";
   static final String JSON_RETENTION_COUNT_KEY = "retentionCount";
+  static final String JSON_RETENTION_TIME_KEY = "retentionTime";
```
JingQianCloud, line 65: Looks like "retentionTime" is in seconds. Can we rename it to include the time unit?

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -144,23 +145,22 @@ public AccountDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
     insertDatasetSql =
         String.format("insert into %s (%s, %s, %s, %s, %s, %s, %s, %s) values (?, ?, ?, ?, now(3), ?, ?, ?)",
             DATASET_TABLE, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_COUNT,
-            USER_TAGS, DELETE_TS);
+            RETENTION_TIME, USER_TAGS);
```
JingQianCloud, line 148: From the ddl, I saw we added RETENTION_TIME but we still keep the DELETE_TS.
Why we don't deal with DELETE_TS here?

JingQianCloud: How do we save the expiration time in the old scheme? I don't see it in the schema. Are we using the "DELETE_TS" for the expiration time?

JingQianCloud (commented):
I have some question about the design.

ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
```diff
@@ -777,15 +776,14 @@ public void testAddGetAndUpdateDatasets() throws Exception {
         datasetFromMysql.getRetentionCount());
 
     //update the expiration time.
-    long expirationTimeInMs = Utils.addSecondsToEpochTime(System.currentTimeMillis(), 30);
+    long datasetTtl = 30;
```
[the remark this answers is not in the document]
SophieGuo410: It's seconds.

ambry-api/src/main/java/com/github/ambry/account/Dataset.java
SophieGuo410: Sure, updated.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: I plan to not support future ttl for dataset. so the retentionTime is the default retention value which all dataset versions under the dataset can inherited from. And the delete_ts is only used for dataset's lifespan, and the dataset can only be deleted by calling the deleteDataset API. Previously I tried to use the delete_ts to represent both default ttl and dataset's lifespan, but it's not make sense because if we create a dataset with 30days ttl, which means any dataset version upload request will be failed after the dataset creation time + 30 days.

ambry-api/src/main/java/com/github/ambry/account/Dataset.java
```diff
@@ -43,6 +42,7 @@
  *    "versionSchema": "TIMESTAMP",
  *    "expirationTimeMs": -1,
  *    "retentionCount": 10,
+ *    "retentionTimeInSeconds": 3600,
```
justinlin-linkedin, line 45: is -1 and null the same thing for retentionTime?

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
JingQianCloud: So DELETE_TS will be updated when deleteDataset is called, right?
When we insert the dataset entry, we don't provide the DELETE_TS. What's the default value for it, -1?

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -144,23 +145,22 @@ public AccountDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
...
             DATASET_TABLE, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_COUNT,
-            USER_TAGS, DELETE_TS);
+            RETENTION_TIME_IN_SECONDS, USER_TAGS);
     updateDatasetIfExpiredSql = String.format(
         "update %s set %s = ?, %s = now(3), %s = ?, %s = ?, %s = ? where %s = ? and %s = ? and %s = ? and delete_ts < now(3)",
-        DATASET_TABLE, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_COUNT, USER_TAGS, DELETE_TS, ACCOUNT_ID,
+        DATASET_TABLE, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_COUNT, RETENTION_TIME_IN_SECONDS, USER_TAGS, ACCOUNT_ID,
         CONTAINER_ID, DATASET_NAME);
     //update the dataset field, in order to support partial update, if one parameter is null, keep the original value.
     //if the delete_ts parameter from input equals to Utils.Infinite_Time(-1), set the delete_ts to null.
```
JingQianCloud, line 154: Comment doesn't match with the code.
In this updateDatasetSql, we didn't touch DELETE_TS.

ambry-account/src/main/resources/AccountSchema.ddl
```diff
@@ -51,6 +51,7 @@ CREATE TABLE IF NOT EXISTS Datasets (
     datasetName VARCHAR(235) NOT NULL,
     versionSchema INT NOT NULL,
     retentionCount INT DEFAULT NULL,
+    retentionTimeInSeconds BIGINT NOT NULL,
```
JingQianCloud, line 54: Can we add comment, for the usage of each fields?

JingQianCloud: What's retentionCount, userTags?
For lastModifiedTime, do we update it whenever there is a change of the Dataset?

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -505,24 +515,27 @@ private void executeDeleteDatasetVersionStatement(PreparedStatement statement, i
...
     statement.setLong(1, version);
-    long updatedExpirationTimeMs = expirationTimeMs;
-    if (expirationTimeMs == Infinite_Time && dataset.getExpirationTimeMs() == Infinite_Time) {
+    long updatedExpirationTimeMs;
+    if (datasetVersionTtlEnabled || dataset.getRetentionTimeInSeconds() == null) {
+      updatedExpirationTimeMs = Utils.addSecondsToEpochTime(creationTimeInMs, timeToLiveInSeconds);
+    } else {
+      updatedExpirationTimeMs = Utils.addSecondsToEpochTime(creationTimeInMs, dataset.getRetentionTimeInSeconds());
+    }
+    if (updatedExpirationTimeMs == Infinite_Time) {
```
JingQianCloud, line 536: Is that possible updatedExpirationTimesMs == Infitinite_time?
Looks like in either the if condition or the else condition, it'll be creationTimeInMs + xxxx.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -505,24 +515,27 @@ private void executeDeleteDatasetVersionStatement(PreparedStatement statement, i
...
-   * @return
+   * @param datasetVersionTtlEnabled set to true if dataset version ttl want to override the dataset level default ttl.
+   * @return the updated expiration time.
    * @throws SQLException
    * @throws AccountServiceException
    */
-  private long executeUpdateDatasetVersionIfExpiredSqlStatement(PreparedStatement statement,
-      int accountId, int containerId, String datasetName, long version, long expirationTimeMs, Dataset dataset)
-      throws SQLException, AccountServiceException {
+  private long executeUpdateDatasetVersionIfExpiredSqlStatement(PreparedStatement statement, int accountId,
```
JingQianCloud, line 526: In my opinion, the statement can be created inside this function, not passed in as parameter.
Because the PreparedStatement has to match with the code below. The caller must be aware of the statement it have to pass in.

Also creating the statement inside the function is easy for reading. Reader can match the order of ? in the statement with the code set the parameters.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -946,15 +951,15 @@ private int executeUpdateDatasetSqlIfExpiredStatement(PreparedStatement statemen
     } else {
       statement.setObject(2, null);
     }
-    if (userTags != null) {
-      statement.setString(3, userTagsInJson);
+    if (retentionTimeInSeconds != null) {
```
JingQianCloud, line 954: Same question here. 
Unless we check which function is calling the function and what the statement passes in, we cannot verify if the statement.setXXX(2, xxx) sets the right parameter.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: That's right.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: Removed.

ambry-account/src/main/resources/AccountSchema.ddl
SophieGuo410: I have the comments in dataset.java. for lastModifiedTime, it's same as Account and Container table that any change of the dataset will update the lastModifiedTime.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: Inside addSecondsToEpochTime, if either creationTimeInMs or timeToLiveInSeconds == Infitinite_time, the updatedExpirationTimesMs will be Infitinite_time.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: I just follow the previous format based on other request. I feel like it's more clear to read the mysql statement in a common place. And most of the requests are simple and by following the order it's easy to read. My personal opinion is to keep the format consistent since both account and named blob sql statement follow the same format.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: I feel like the integration test is good enough to verify it, besides, we have the local mysql database which can clearly check what we set.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
JingQianCloud: I don't mean that we need move the insertDatasetSql definition. 
I don't see the benefit that create the PreparedStatement outside and pass in to the function as below.

        PreparedStatement updateDatasetSqlIfExpiredStatement =
            dataAccessor.getPreparedStatement(updateDatasetVersionIfExpiredSql, true);
        long newExpirationTimeMs =
              executeUpdateDatasetVersionIfExpiredSqlStatement(updateDatasetSqlIfExpiredStatement, accountId,
                  containerId, datasetName, versionNumber, expirationTimeMs, dataset);

Actually many of the other functions as below generates the PreparedStatement inside the function when the set is altogether. 
- getNewAccounts
- getContainers
- getNewContainers
- getContainerById


For some batch functions, the reason it wants to pass in the statement is it tries to share the code forinsertAccountStatement/updateAccountStatement/insertContainerStatement/updateContainerStatement

justinlin-linkedin (approved):

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: Oh, I think the reason that those function has the statement and execution together is due to it's pretty simple. But my case is more complicated so I created the sub method to do the execution in order to make the function easy to read. And by checking the execute* method name and java doc, you will know which statement I'm execute.

PR #2413, opened 2023-03-20
Support getting bootstrap replica in Helix Full auto mode

Arun-LinkedIn:
This PR adds changes to create a new AmbryReplica object dynamically during OFFLINE -> BOOTSTRAP instance in Full Auto mode. 

In Semi-auto mode, the new replica information such as mount path of the disk, size of the replica is decided before hand and stored to Helix (/PROPERTYSTORE/AdminConfigs/ReplicaAddition Znode).  But, in Full-auto mode, the decision of the disk to host the replica has to be made locally. This PR adds those changes.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
@@ -97,6 +97,9 @@ public class ClusterMapUtils {
   static final long MIN_REPLICA_CAPACITY_IN_BYTES = 1024 * 1024 * 1024L;
   static final long MAX_REPLICA_CAPACITY_IN_BYTES = 10L * 1024 * 1024 * 1024 * 1024;
   static final long MIN_DISK_CAPACITY_IN_BYTES = 10L * 1024 * 1024 * 1024;
+  // TODO: Temporary defaults to be used when adding replicas in helix FULL_AUTO mode. These replica configs will need
+  //  to stored in configs or helix for each cluster.
+  static final long DEFAULT_REPLICA_CAPACITY_IN_BYTES = 100L * 1024 * 1024 * 1024;
   static final int CURRENT_SCHEMA_VERSION = 0;
   static final String WRITABLE_LOG_STR = "writable";
   static final String FULLY_WRITABLE_LOG_STR = "fully writable";
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -462,7 +462,6 @@ public class HelixClusterManager implements ClusterMap {
         checkLocalDcOnly);
   }
 
-
   /**
    * Disconnect from the HelixManagers associated with each and every datacenter.
    */
@@ -658,13 +657,13 @@ public class HelixClusterManager implements ClusterMap {
    *     "1": {
    *         "replicaCapacityInBytes": 107374182400,
    *         "partitionClass": "max-replicas-all-datacenters",
-   *         "localhost1_17088": "/tmp/c/1",
-   *         "localhost2_17088": "/tmp/d/1"
+   *         "localhost1_17088": "/tmp/c/1,536870912000",
+   *         "localhost2_17088": "/tmp/d/1,536870912000"
    *     },
    *     "2": {
    *         "replicaCapacityInBytes": 107374182400,
    *         "partitionClass": "max-replicas-all-datacenters",
-   *         "localhost3_17088": "/tmp/e/1"
+   *         "localhost3_17088": "/tmp/e/1,536870912000"
    *     }
    * }
    * </pre>
@@ -711,10 +710,14 @@ public class HelixClusterManager implements ClusterMap {
         if (diskCapacityStr != null) {
           // update disk capacity if bootstrap replica info contains disk capacity in bytes.
           targetDisk.setDiskCapacityInBytes(Long.parseLong(diskCapacityStr));
+        } else {
+          logger.info("Replica addition infos map doesn't contain disk capacity. Disk {} capacity {} is not changed",
+              targetDisk, targetDisk.getRawCapacityInBytes());
         }
         // A bootstrap replica is always ReplicaSealedStatus#NOT_SEALED.
         bootstrapReplica = new AmbryServerReplica(clusterMapConfig, currentPartition, targetDisk, true, replicaCapacity,
             ReplicaSealStatus.NOT_SEALED);
+        logger.info("Created bootstrap replica {} for Partition {}", bootstrapReplica, partitionIdStr);
       } catch (Exception e) {
         logger.error("Failed to create bootstrap replica for partition {} on {} due to exception: ", partitionIdStr,
             instanceName, e);
@@ -749,8 +752,11 @@ public class HelixClusterManager implements ClusterMap {
         logger.info("Partition {} is currently not present in cluster map, a new partition is created", partitionIdStr);
         currentPartition = mappedPartition;
       }
-      return new AmbryServerReplica(clusterMapConfig, currentPartition, disk, true, DEFAULT_REPLICA_CAPACITY_IN_BYTES,
-          ReplicaSealStatus.NOT_SEALED);
+      AmbryServerReplica replica =
+          new AmbryServerReplica(clusterMapConfig, currentPartition, disk, true, DEFAULT_REPLICA_CAPACITY_IN_BYTES,
+              ReplicaSealStatus.NOT_SEALED);
+      logger.info("Created bootstrap replica {} for Partition {}", replica, partitionIdStr);
+      return replica;
     } catch (Exception e) {
       logger.error("Failed to create bootstrap replica for partition {} on {} due to exception: ", partitionIdStr,
           dataNodeId, e);
@@ -772,7 +778,7 @@ public class HelixClusterManager implements ClusterMap {
     long maxAvailableDiskSpace = 0;
     for (AmbryDisk disk : disks) {
       if (disk.getAvailableSpaceInBytes() < DEFAULT_REPLICA_CAPACITY_IN_BYTES) {
-        logger.info("Disk {} doesn't have space to host new replica. Disk space left {}, replica capacity {}", disk,
+        logger.debug("Disk {} doesn't have space to host new replica. Disk space left {}, replica capacity {}", disk,
             disk.getAvailableSpaceInBytes(), DEFAULT_REPLICA_CAPACITY_IN_BYTES);
         continue;
       }
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -30,6 +30,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Random;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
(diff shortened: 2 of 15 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -388,8 +387,86 @@ public List<PartitionId> getAllPartitionIds(String partitionClass) {
...
+      currentPartition = mappedPartition;
+    }
+
+    // Select a disk for the replica
+    Set<AmbryDisk> disks = ambryDataNodeToAmbryDisks.get(dataNode);
+    // Get disk which has maximum remaining space
+    AmbryDisk diskToUse = null;
+    for (AmbryDisk disk : disks) {
+      if (diskToUse == null || (disk.getRawCapacityInBytes() - disk.getUsedCapacityInBytes()
+          > diskToUse.getRawCapacityInBytes() - diskToUse.getUsedCapacityInBytes())) {
```
justinlin-linkedin, line 450: Can we create a method call getRemainingCapacityInBytes()? or getFreeCapacityInBytes()?
And should we check the disk health here as well?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -388,8 +387,86 @@ public List<PartitionId> getAllPartitionIds(String partitionClass) {
...
+        diskToUse = disk;
+      }
+    }
```
justinlin-linkedin, line 453: how do we deal with concurrent operation where a bunch of replicas are added to the disk but the disk's used capacity is not yet increased?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
```diff
@@ -96,6 +96,9 @@ public class ClusterMapUtils {
   static final long MIN_REPLICA_CAPACITY_IN_BYTES = 1024 * 1024 * 1024L;
   static final long MAX_REPLICA_CAPACITY_IN_BYTES = 10L * 1024 * 1024 * 1024 * 1024;
   static final long MIN_DISK_CAPACITY_IN_BYTES = 10L * 1024 * 1024 * 1024;
+  // TODO: Temporary defaults to be used when adding replicas in helix FULL_AUTO mode. These replica configs will need
+  //  to stored in configs or helix for each cluster.
+  static final long DEFAULT_REPLICA_CAPACITY_IN_BYTES = 100L * 1024 * 1024 * 1024;
```
justinlin-linkedin, line 101: it should be in the helix cluster config. Each partition would have a weight, this weight is determined by the replica capacity. We should include the partition weight and the replica capacity in the same cluster config.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -170,7 +171,10 @@ void start() throws InterruptedException {
         }
         Thread thread = Utils.newThread("store-startup-" + partitionAndStore.getKey(), () -> {
           try {
-            partitionAndStore.getValue().start();
+            BlobStore store = partitionAndStore.getValue();
+            store.start();
+            // Increase the disk used capacity once the store is successfully started
+            disk.addUsedCapacityInBytes(store.getReplicaId().getCapacityInBytes());
```
justinlin-linkedin, line 177: A stopped replica would also take some space on disk, we should add the used capacity for stopped replicas as well.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -371,6 +375,8 @@ boolean addBlobStore(ReplicaId replica) {
         // create a bootstrap-in-progress file to distinguish it from regular stores (the file will be checked during
         // BOOTSTRAP -> STANDBY transition)
         createBootstrapFileIfAbsent(replica);
+        // Update the disk used capacity
+        disk.addUsedCapacityInBytes(replica.getCapacityInBytes());
```
justinlin-linkedin, line 379: I only saw the disk's used capacity increases when adding blob store, but we should have logic when the disk used capacity is decreased due to removing replica.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
Arun-LinkedIn: Sure. I will add these configs in subsequent PR and remove these default values.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
Arun-LinkedIn: Thank you for catching it. Fixed it.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
Arun-LinkedIn: Sure. Changed to modify the disk usage in constructor so that it will apply for all replicas that are present on the node.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: Sure. Created method getAvailableSpaceInBytes(). 

Yeah, true. Like you said, we should check disk health too. Will add it in subsequent PR.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: Yeah true. Added synchronization when selecting a disk so that it helps in concurrent operations.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -778,7 +778,6 @@ private ServerErrorCode handleRemoveStoreRequest(PartitionId partitionId) throws
     Store store = ((StorageManager) storeManager).getStore(partitionId, true);
     // Attempt to remove store from storage manager.
     if (storeManager.removeBlobStore(partitionId) && store != null) {
-      ((BlobStore) store).deleteStoreFiles();
```
SophieGuo410, line 781: QQ, I saw that you move this into the removeBlobStore in StorageManager. Would it affect the CloudStorageManager?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -713,6 +646,152 @@ private AmbryReplica fetchBootstrapReplica(String partitionName) {
...
+   * returns null.
+   */
+  synchronized private AmbryDisk getDiskForBootstrapReplica(AmbryDataNode dataNode) {
+    Set<AmbryDisk> disks = ambryDataNodeToAmbryDisks.get(dataNode);
+    List<AmbryDisk> potentialDisks = new ArrayList<>();
+    long maxAvailableDiskSpace = 0;
+    for (AmbryDisk disk : disks) {
+      if (disk.getAvailableSpaceInBytes() < DEFAULT_REPLICA_CAPACITY_IN_BYTES) {
+        logger.info("Disk {} doesn't have space to host new replica. Disk space left {}, replica capacity {}", disk,
+            disk.getAvailableSpaceInBytes(), DEFAULT_REPLICA_CAPACITY_IN_BYTES);
```
justinlin-linkedin, line 776: nit: could be a trace log

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -713,6 +646,152 @@ private AmbryReplica fetchBootstrapReplica(String partitionName) {
...
+   * @return {@link ReplicaId} if there is a new replica satisfying given partition and data node. {@code null} otherwise.
+   */
+  private ReplicaId getBootstrapReplicaInFullAuto(String partitionIdStr, DataNodeId dataNodeId) {
+    try {
+      AmbryDisk disk = getDiskForBootstrapReplica((AmbryDataNode) dataNodeId);
+      if (disk == null) {
+        logger.error("No Disk is available to host bootstrap replica. Cannot create the replica.");
+        return null;
+      }
+      AmbryPartition mappedPartition =
```
justinlin-linkedin, line 743: nit: let's log a trace log here so we know which disk is picked for this bootstrap replica.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
```diff
@@ -778,7 +778,6 @@ private ServerErrorCode handleRemoveStoreRequest(PartitionId partitionId) throws
     Store store = ((StorageManager) storeManager).getStore(partitionId, true);
     // Attempt to remove store from storage manager.
     if (storeManager.removeBlobStore(partitionId) && store != null) {
```
justinlin-linkedin, line 780: at some point we have to re-evaulate if these admin command is still valid or not. This goes against dynamic helix replica management. I always thought this is still there because we still support static cluster map. But I might be wrong. If this is not used anywhere, then we should retire these logics.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
Arun-LinkedIn: I think it might be okay since we don't have any local files deletion logic in the CloudStorageManager.

ambry-server/src/main/java/com/github/ambry/server/AmbryServerRequests.java
Arun-LinkedIn: Sure. I will check with SRE if we still use admin commands for store addition and removal. If not, I will put up a PR to remove them.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: Sure. Changed it to debug level.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
Arun-LinkedIn: Done

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -713,6 +646,152 @@ private AmbryReplica fetchBootstrapReplica(String partitionName) {
...
+    String[] segments = mountPathAndDiskCapacityFromHelix.split(DISK_CAPACITY_DELIM_STR);
+    String mountPath = segments[0];
+    String diskCapacityStr = segments.length >= 2 ? segments[1] : null;
+    Set<AmbryDisk> disks = dataNode != null ? ambryDataNodeToAmbryDisks.get(dataNode) : null;
+    Optional<AmbryDisk> potentialDisk =
+        disks != null ? disks.stream().filter(d -> d.getMountPath().equals(mountPath)).findAny() : Optional.empty();
+    if (potentialDisk.isPresent()) {
+      try {
+        AmbryDisk targetDisk = potentialDisk.get();
+        if (diskCapacityStr != null) {
```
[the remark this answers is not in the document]
Arun-LinkedIn: This information is added by us in the ReplicaAddition Znode during replica addition step. By looking at code in HelixBootstrapUpgradeUtil in  `generateReplicaAdditionMap` method, I think disk capacity will always be there in the information. I think the null check was there for backward compatibility.

Just wanted to note that, this is existing logic. I just moved it into a new method.

justinlin-linkedin (approved):

SophieGuo410 (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -713,6 +645,159 @@ private AmbryReplica fetchBootstrapReplica(String partitionName) {
...
+   * TODO Check for disk health as well (Will be added in next PR)
+   * @param dataNode the {@link DataNodeId} on which disk is needed
+   * @return {@link AmbryDisk} which has maximum available or free capacity. If none of the disks have free space,
+   * returns null.
+   */
+  synchronized private AmbryDisk getDiskForBootstrapReplica(AmbryDataNode dataNode) {
+    Set<AmbryDisk> disks = ambryDataNodeToAmbryDisks.get(dataNode);
+    List<AmbryDisk> potentialDisks = new ArrayList<>();
+    long maxAvailableDiskSpace = 0;
+    for (AmbryDisk disk : disks) {
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure. I will put up a separate PR for checking if the disk is healthy.

PR #2414, opened 2023-03-20
Add SimpleOperationTracker option for undelete

justinlin-linkedin:
This PR adds SimpleOperationTracker to Undelete in case where we need it.

Undelete is considered as an Administrative operations. We can relax the global quorum requirement for undelete some times.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
@@ -67,6 +67,9 @@ public class UndeleteOperation {
   private boolean operationCompleted = false;
   private static final Logger LOGGER = LoggerFactory.getLogger(UndeleteOperation.class);
 
+  // True when the operationTrackerType is UndeleteOperationTracker.
+  // If true, operation will verify that life versions received from replicas are the same
+  private final boolean verifyLifeVersion;
   private Short lifeVersion = null;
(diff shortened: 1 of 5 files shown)


SophieGuo410 (approved):

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
```diff
@@ -95,8 +98,20 @@ public class UndeleteOperation {
     this.callback = callback;
     this.time = time;
     this.operationTimeMs = operationTimeMs;
-    this.operationTracker = new UndeleteOperationTracker(routerConfig, blobId.getPartition(),
-        clusterMap.getDatacenterName(blobId.getDatacenterId()), routerMetrics);
+    if (routerConfig.routerUndeleteOperationTrackerType.equals(UndeleteOperationTracker.class.getSimpleName())) {
```
ankagrawal, line 101: nit: consider using enum.

ankagrawal (approved):

ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
justinlin-linkedin: the reason not to use enum here is that we have same logic for GetBlobOperation and PutBlobOperation, and they are not using enum either. Want to keep it consistent.

PR #2423, opened 2023-04-06
Support retentionCount for dataset version database

SophieGuo410:
This pr supported the retentionCount logic for our dataset version db.
When user issue a put request to put a new version of dataset, we will check if the number of existing dataset version of this dataset is out of the retention count number which has been set at dataset level.
And we will issue a delete to try to delete the dataset version out of the retention count. 
And since it happened to every request, we will handle it async and only log the error messaged if delete failed. And retry with next put. So it won't affect the availability of put.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
@@ -177,9 +177,10 @@ public class AccountDao {
     listValidVersionSql =
         String.format("select %s, %s from %s " + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ?",
             VERSION, DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME);
-    listVersionByModifiedTimeSql = String.format(
-        "select %s, %s from %s " + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ? ORDER BY %s DESC",
-        VERSION, DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME,
+    // list all valid versions sorted by last modified time, and skip the first N records which is not out of retentionCount.
+    listVersionByModifiedTimeSql = String.format("select %s, %s from %s "
+            + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ? ORDER BY %s DESC LIMIT ?, 100", VERSION,
+        DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME,
         LAST_MODIFIED_TIME);
     getDatasetVersionByNameSql =
         String.format("select %s, %s from %s where %s = ? and %s = ? and %s = ? and %s = ?", LAST_MODIFIED_TIME,
@@ -772,17 +773,14 @@ public class AccountDao {
       statement.setInt(1, accountId);
       statement.setInt(2, containerId);
       statement.setString(3, datasetName);
+      statement.setInt(4, retentionCount);
       resultSet = statement.executeQuery();
-      int idx = 0;
       while (resultSet.next()) {
-        idx++;
-        if (idx > retentionCount) {
-          long versionValue = resultSet.getLong(VERSION);
-          Timestamp deletionTime = resultSet.getTimestamp(DELETE_TS);
-          String version = convertVersionValueToVersion(versionValue, versionSchema);
-          datasetVersionRecordList.add(
-              new DatasetVersionRecord(accountId, containerId, datasetName, version, timestampToMs(deletionTime)));
-        }
+        long versionValue = resultSet.getLong(VERSION);
+        Timestamp deletionTime = resultSet.getTimestamp(DELETE_TS);
+        String version = convertVersionValueToVersion(versionValue, versionSchema);
+        datasetVersionRecordList.add(
+            new DatasetVersionRecord(accountId, containerId, datasetName, version, timestampToMs(deletionTime)));
       }
       return datasetVersionRecordList;
     } finally {
--- a/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
@@ -98,6 +98,7 @@ public class AccountDao {
   private final String updateDatasetVersionIfExpiredSql;
   private final String listVersionSql;
   private final String listValidVersionSql;
+  private final String listVersionByModifiedTimeSql;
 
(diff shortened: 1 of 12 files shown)


ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -176,6 +177,10 @@ public AccountDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
     listValidVersionSql =
         String.format("select %s, %s from %s " + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ?",
             VERSION, DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME);
+    listVersionByModifiedTimeSql = String.format(
```
justinlin-linkedin, line 180: there is a way in sql statement to skip first N rows before returning any data, could we just use that? I think LIMIT clause has two parameters to skip first N rows.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -548,5 +552,41 @@ private void addDatasetVersion(BlobProperties blobProperties, RestRequest restRe
             RestServiceErrorCode.getRestServiceErrorCode(ex.getErrorCode()));
       }
     }
+
+    /**
+     * Support delete the dataset version out of retentionCount
+     */
+    private void deleteDatasetVersionOutOfRetentionCount() {
+      Dataset dataset = (Dataset) restRequest.getArgs().get(RestUtils.InternalKeys.TARGET_DATASET);
```
justinlin-linkedin, line 560: this statement should be placed in the if statement below.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -548,5 +552,41 @@ private void addDatasetVersion(BlobProperties blobProperties, RestRequest restRe
...
+            restRequest.setArg(InternalKeys.REQUEST_PATH, newRequestPath);
+            restRequest.setArg(InternalKeys.TARGET_ACCOUNT_KEY, null);
+            restRequest.setArg(InternalKeys.TARGET_CONTAINER_KEY, null);
+            try {
+              deleteBlobHandler.handle(restRequest, restResponseChannel, null);
+            } catch (Exception e) {
+              //do not throw the exception or handle it since we don't care if it has been deleted successfully.
+              LOGGER.error("Failed to delete the datasetVersionRecord :" + record);
+            }
+          }
```
justinlin-linkedin, line 585: oh, this is probably not going to work.

You see, the delete handler is an async method, so it will not finished the execution before handle method returns. And if we have multiple versions to delete, then we would override the RequestPath in this for loop since the they share the same restRequest.

For example:
Say we have version1 and version2 to delete.
for version1, we replace RequestPath in restRequest to version1 and call deleteBlobHandler.handle(restRequest).
Before it finishes, we move on to version2, and now we replace the RequestPath in restRequest and call deleteBlobHandler.handle(restRequest) again.
But since deleting version 1 isn't done yet, any logic in deleting version1 would now see version2.

We might have to create a new restRequest for each version.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: Sure, updated it.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
SophieGuo410: Fixed.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
SophieGuo410: Yeah, I feel it's hard to clone a restRequest, so I decided to use the recursive callback to delete dataset version synchronously, and ignore the delete failures. Normally there would be only one version need to be deleted.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -548,5 +552,41 @@ private void addDatasetVersion(BlobProperties blobProperties, RestRequest restRe
...
+     */
+    private void deleteDatasetVersionOutOfRetentionCount() {
+      Dataset dataset = (Dataset) restRequest.getArgs().get(RestUtils.InternalKeys.TARGET_DATASET);
+      try {
+        if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
+          String accountName = dataset.getAccountName();
+          String containerName = dataset.getContainerName();
+          String datasetName = dataset.getDatasetName();
+          List<DatasetVersionRecord> datasetVersionRecordList =
+              accountService.getAllValidVersionsOutOfRetentionCount(accountName, containerName, datasetName);
```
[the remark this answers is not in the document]
SophieGuo410: That's a good concern. I think normally we should be good. Since we always delete once we successfully put a new version into the database. And we always leverage the same primary DB to do the get and delete. There's one case that when user failed to push the version and the version has not been deleted from DB after user successfully issue another new put, it will consider the failed put as a version and still delete the old one. But we never promise user that the version will be a valid version. Considering we already tell user that we are not guarantee to rollback the failed put version and it might need manual delete, so I think it should be good(this is how wormhole supported as well). But I think we could have some update in the future to add a new filed to the database when the put is done successfully, this might be better. Let me know if you have any concern about it.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -548,5 +559,67 @@ private void addDatasetVersion(BlobProperties blobProperties, RestRequest restRe
...
+     * Recursively delete the dataset version out of retention count at dataset level.
+     * @param datasetVersionRecordList the list of the {@link DatasetVersionRecord}
+     * @param idx the index of the datasetVersionRecordList
+     * @param accountName the name of the account.
+     * @param containerName the name of the container.
+     * @param datasetName the name of the dataset.
+     */
+    private Callback<Void> recursiveCallback(List<DatasetVersionRecord> datasetVersionRecordList, int idx, String accountName,
+        String containerName, String datasetName) {
+      if (idx == datasetVersionRecordList.size()) {
```
JingQianCloud, line 604: Can we add some print message(or metric) to record the status of the background deletion?

JingQianCloud (approved):
Overall, looks good to me.

justinlin-linkedin (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
SophieGuo410: Sure, I'll add it in a follow up pr.

PR #2429, opened 2023-04-11
Fixed two minor NPE in ambry-server

justinlin-linkedin:
This PR fixes two minor NPE in ambry server.


ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -618,6 +618,17 @@ private boolean copyDataByLogSegment(LogSegment logSegmentToCopy, FileSpan dupli
     long logSegmentStartTime = time.milliseconds();
     for (Offset indexSegmentStartOffset : getIndexSegmentDetails(logSegmentToCopy.getName()).keySet()) {
       IndexSegment indexSegmentToCopy = srcIndex.getIndexSegments().get(indexSegmentStartOffset);
+      if (indexSegmentToCopy == null) {
+        // This should not happen
```
JingQianCloud, line 622: Just curious, did you see it happens on the server?
Your comment said it shouldn't happen. So it's a workaround and we need figure out the root cause, is my understanding correct?

JingQianCloud (approved):

Arun-LinkedIn (approved):

PR #2431, opened 2023-04-17
Remove swapped host from container stats database

justinlin-linkedin:
This PR would try to remove the swapped hosts from the container stats database and save some disk space for the database.

This is an time consuming operation, but host swap won't happen that often. So I put a possibility check for this operation. Only a random value is less than the possibility we provide in the configuration file should we clean up the swapped hosts.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/accountstats/AccountStatsStore.java
+++ b/ambry-api/src/main/java/com/github/ambry/accountstats/AccountStatsStore.java
@@ -17,6 +17,8 @@ import com.github.ambry.server.HostAccountStorageStatsWrapper;
 import com.github.ambry.server.HostPartitionClassStorageStatsWrapper;
 import com.github.ambry.server.storagestats.AggregatedAccountStorageStats;
 import com.github.ambry.server.storagestats.AggregatedPartitionClassStorageStats;
+import com.github.ambry.utils.Pair;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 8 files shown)


SophieGuo410 (approved):
lgtm!

ambry-api/src/main/java/com/github/ambry/config/ClusterMapConfig.java
```diff
@@ -329,6 +331,15 @@ public class ClusterMapConfig {
...
   public final boolean clustermapEnableDeleteInvalidDataInMysqlAggregationTask;
 
+  /**
+   * The possibility to delete invalid hosts in mysql aggregation tasks. It should be a value between 0 and 1.
+   * When it's 0, it means we are never going to delete invalid hosts in the aggregation task. When it's 1, it
+   * means every time we run an aggregation task, we will try to delete the invalid hosts.
+   */
+  @Config(DELETE_INVALID_HOSTS_IN_MYSQL_AGGREGATION_TASK_POSSIBILITY)
+  @Default("0.01")
+  public final float clustermapDeleteInvalidHostsInMysqlAggregationTaskPossibility;
```
Arun-LinkedIn, line 341: I was just thinking if we could also remove invalid hosts by listening to changes in cluster map. For example, a) add a method `onHostRemoved(String hostname)` in `ClusterMapChangeListener` and b) StatsManager could register with the listener with clusterMap. Would it work?

ambry-api/src/main/java/com/github/ambry/accountstats/AccountStatsStore.java
```diff
@@ -84,6 +86,14 @@ public interface AccountStatsStore {
    */
   AggregatedAccountStorageStats queryMonthlyAggregatedAccountStorageStats() throws Exception;
 
+  /**
+   * Retain the host account storage stats for the given list of host and port, and remove the account storage stats for
+   * the other hosts.
+   * @param hosts The list of host and port pair
+   * @throws Exception
+   */
+  void retainHostAccountStorageStatsForHosts(List<Pair<String, Integer>> hosts) throws Exception;
```
Arun-LinkedIn, line 95: nit: Since we want to remove old hosts from mysql, would name like `deleteAccountStorageStatesForInvalidHosts` be better?

PR #2434, opened 2023-04-25
[Efficient Metadata Operations] Add reserved metadata id in the blob properties.

ankagrawal:
The reserved metadata id in blob properties will be used to save the parent anchor of the data chunks in the server and to pass the reserved metadata id in case of chunked and stitched requests.

Note that the only significant change in this PR is that it adds a field "reservedMetadataBlobId" in the BlobProperties class and updates its constructors. All other changes are just to update the constructor arguments in other places.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
+++ b/ambry-api/src/main/java/com/github/ambry/messageformat/BlobProperties.java
@@ -37,6 +37,7 @@ public class BlobProperties {
   private long timeToLiveInSeconds;
   // Non persistent blob properties.
   private final String externalAssetTag;
+  private final String reservedMetadataBlobId;
 
   /**
    * @param blobSize The size of the blob in bytes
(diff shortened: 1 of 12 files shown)


JingQianCloud (commented):
We added the field "reservedMetadataBlobID to BlobProperties.java.
But we don't use it yet, our deserialization and serialization ignored this field And we didn't use revision 5 either.
So it's only interface change, no real code logic change. Right?

ankagrawal:
> We added the field "reservedMetadataBlobID to BlobProperties.java. But we don't use it yet, our deserialization and serialization ignored this field And we didn't use revision 5 either. So it's only interface change, no real code logic change. Right?

Yes, this is pretty much an interface change for now. The subsequent PRs will start using this. For now the serialization-deserialization logic ignores this field.
I updated the PR to remove the version because that's not needed for this PR.

JingQianCloud (approved):

Arun-LinkedIn (approved):

PR #2438, opened 2023-04-26
[Efficient Metadata Operations] Add metric to track slipped puts for metadata chunks.

ankagrawal:


ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterMetrics.java
```diff
@@ -184,7 +184,9 @@ public class NonBlockingRouterMetrics {
   public final Meter operationErrorRate;
   public final Meter encryptedOperationErrorRate;
   public final Counter slippedPutAttemptCount;
+  public final Counter metadataSlippedPutAttemptCount;
```
JingQianCloud, line 187: Can we add comment for 
    slippedPutAttempCount,
    slippedPutSuccessCount,
    metadataSlippedPutSuccessCount
    metadataPutAttemptCount.

Looks like your intention is that slippedPutXXX includes both data chunk and meta data while metadataXXX is as name indicated.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -580,14 +580,10 @@ private void onChunkOperationComplete(PutChunk chunk) {
       logger.trace("{}: Successfully put data chunk with blob id : {}", loggingContext, chunk.getChunkBlobId());
```
JingQianCloud, line 580: I believe you don't need this log any more.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouterMetrics.java
ankagrawal: You are right about the intention.
Added the comments.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: fixed.

JingQianCloud (approved):

PR #2440, opened 2023-04-26
Add a new callback method to ClusterMapChangeListener

justinlin-linkedin:
Adding a new callback method to ClusterMapChangeListener for DataNode removal so other components that are interested on this events can be notified. 

Adding a empty callback to StatsManager. The real implementation will be added later.

### DIFF
--- a/ambry-test-utils/src/main/java/com/github/ambry/clustermap/MockClusterMap.java
+++ b/ambry-test-utils/src/main/java/com/github/ambry/clustermap/MockClusterMap.java
@@ -672,6 +672,16 @@ public class MockClusterMap implements ClusterMap {
   }
 
   /**
+   * removeDataNode to invoke clustermap change listener. This method doesn't really remove the datanode.
+   * @param dataNodeId The removed data node
+   */
+  public void removeDataNode(DataNodeId dataNodeId) {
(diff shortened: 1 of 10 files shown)


ambry-test-utils/src/main/java/com/github/ambry/clustermap/MockClusterMap.java
```diff
@@ -671,6 +671,16 @@ public ClusterMapChangeListener getClusterMapChangeListener() {
     return clusterMapChangeListener;
   }
 
+  /**
+   * removeDataNode to invoke clustermap change listener. This method doesn't really remove the datanode.
+   * @param dataNodeId The removed data node
+   */
+  public void removeDataNode(DataNodeId dataNodeId) {
```
SophieGuo410, line 678: Can we update the method name to be something like invokeListenerWhenDataNodeRemoved since this method does not remove the data node?

Arun-LinkedIn (approved):

ambry-test-utils/src/main/java/com/github/ambry/clustermap/MockClusterMap.java
justinlin-linkedin: updated

SophieGuo410 (approved):

PR #2441, opened 2023-04-27
[Efficient Metadata Operations] Reserve metadata chunkid for chunked uploads.

ankagrawal:

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/frontend/ReservedMetadataIdMetrics.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/ReservedMetadataIdMetrics.java
@@ -33,6 +33,10 @@ public class ReservedMetadataIdMetrics {
   public final Counter noReservedMetadataFoundForChunkedUploadResponseCount;
   public final Counter noReservedMetadataForChunkedUploadCount;
   public final Counter mismatchedReservedMetadataForChunkedUploadCount;
+  public final Counter metadataIdNotReservedCount;
+  public final Counter stitchedBlobMetadataIdDeserErrorCount;
+  public final Counter stitchedBlobMetadataIdMismatchCount;
+  public final Counter reservedMetadataPassedInForNonChunkedUploadCount;
 
   /**
    * Constructor for {@link ReservedMetadataIdMetrics}.
@@ -57,6 +61,14 @@ public class ReservedMetadataIdMetrics {
         MetricRegistry.name(ReservedMetadataIdMetrics.class, "NoReservedMetadataForChunkedUploadCount"));
     mismatchedReservedMetadataForChunkedUploadCount = metricRegistry.counter(
         MetricRegistry.name(ReservedMetadataIdMetrics.class, "MismatchedReservedMetadataForChunkedUploadCount"));
+    metadataIdNotReservedCount = metricRegistry.counter(
+        MetricRegistry.name(ReservedMetadataIdMetrics.class, "MetadataIdNotReservedCount"));
+    stitchedBlobMetadataIdDeserErrorCount = metricRegistry.counter(
+        MetricRegistry.name(ReservedMetadataIdMetrics.class, "StitchedBlobMetadataIdDeserErrorCount"));
+    stitchedBlobMetadataIdMismatchCount = metricRegistry.counter(
+        MetricRegistry.name(ReservedMetadataIdMetrics.class, "StitchedBlobMetadataIdMismatchCount"));
+    reservedMetadataPassedInForNonChunkedUploadCount = metricRegistry.counter(
+        MetricRegistry.name(ReservedMetadataIdMetrics.class, "ReservedMetadataPassedInForNonChunkedUploadCount"));
   }
 
   /**
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -319,12 +319,25 @@ class PutOperation {
     bytesFilledSoFar = 0;
     chunkCounter = -1;
     putChunks = new ConcurrentLinkedQueue<>();
-    reservedMetadataIdMetrics = ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(routerMetrics.getMetricRegistry());
-    metadataPutChunk = new MetadataPutChunk();
+    reservedMetadataIdMetrics =
+        ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(routerMetrics.getMetricRegistry());
     chunkFillerChannel = new ByteBufferAsyncWritableChannel(writableChannelEventListener);
     isEncryptionEnabled = passedInBlobProperties.isEncrypted();
     restRequest = options.getRestRequest();
     loggingContext = makeLoggingContext();
+    if (chunksToStitch != null && chunksToStitch.isEmpty()) {
+      // chunksToStitch is null for non stitch requests.
+      // For stitch requests, the current metadata format does not support empty chunk lists
+      setOperationExceptionAndComplete(new RouterException("Must provide at least one chunk for stitchBlob call",
+          RouterErrorCode.InvalidPutArgument));
+    }
+    MetadataPutChunk metadataPutChunk = null;
+    try {
+      metadataPutChunk = new MetadataPutChunk(chunksToStitch);
+    } catch (RouterException routerException) {
+      setOperationExceptionAndComplete(routerException);
+    }
+    this.metadataPutChunk = metadataPutChunk;
   }
 
   /**
@@ -384,11 +397,6 @@ class PutOperation {
    * @throws RouterException if validation of the chunks to stitch failed.
    */
   private void processChunksToStitch() throws RouterException {
-    if (chunksToStitch.isEmpty()) {
-      // The current metadata format does not support empty chunk lists
-      throw new RouterException("Must provide at least one chunk for stitchBlob call",
-          RouterErrorCode.InvalidPutArgument);
-    }
     long totalSize = 0;
     long intermediateChunkSize =
         routerConfig.routerMetadataContentVersion == MessageFormatRecord.Metadata_Content_Version_V2
@@ -1268,19 +1276,22 @@ class PutOperation {
         // To ensure previously attempted partitions are not retried for this PUT after a failure.
         attemptedPartitionIds.add(partitionId);
 
-        chunkBlobId =
-            new BlobId(routerConfig.routerBlobidCurrentVersion, BlobIdType.NATIVE, clusterMap.getLocalDatacenterId(),
-                passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(), partitionId,
-                passedInBlobProperties.isEncrypted(), blobDataType);
+        if (shouldUseReservedMetadataId()) {
+          chunkBlobId = metadataPutChunk.reservedMetadataChunkId;
+        } else {
+          chunkBlobId =
+              new BlobId(routerConfig.routerBlobidCurrentVersion, BlobIdType.NATIVE, clusterMap.getLocalDatacenterId(),
+                  passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(), partitionId,
+                  passedInBlobProperties.isEncrypted(), blobDataType);
+        }
 
-        // TODO: Efficient Metadata Operations: The reserved metadata blob id needs to be passed as chunk properties.
         chunkBlobProperties = new BlobProperties(chunkBlobSize, passedInBlobProperties.getServiceId(),
             passedInBlobProperties.getOwnerId(), passedInBlobProperties.getContentType(),
             passedInBlobProperties.isPrivate(), passedInBlobProperties.getTimeToLiveInSeconds(),
             passedInBlobProperties.getCreationTimeInMs(), passedInBlobProperties.getAccountId(),
             passedInBlobProperties.getContainerId(), passedInBlobProperties.isEncrypted(),
             passedInBlobProperties.getExternalAssetTag(), passedInBlobProperties.getContentEncoding(),
-            passedInBlobProperties.getFilename(), null);
+            passedInBlobProperties.getFilename(), resolveReservedMetadataId());
         operationTracker = getOperationTracker();
         correlationIdToChunkPutRequestInfo.clear();
         logger.trace("{}: Chunk {} is ready for sending out to server", loggingContext, chunkIndex);
@@ -1291,6 +1302,19 @@ class PutOperation {
         setOperationExceptionAndComplete(
             new RouterException("Prepare for sending failed", e, RouterErrorCode.UnexpectedInternalError));
       }
+    }
+
+    /**
+     * Resolve the reserved metadata id to be used for data chunks.
+     * For simple blobs or metadata chunks this will be {@code null}. For the data chunks of composite blobs, the reserved
+     * metadata id of the metadata chunk will be used.
+     * @return the blob id to use as the reserved metadata id. Can be {@code null}.
+     */
+    private String resolveReservedMetadataId() {
+      if (!routerConfig.routerReservedMetadataEnabled || isSimpleBlob || isMetadataChunk()) {
+        return null;
+      }
+      return metadataPutChunk.reservedMetadataChunkId.getID();
     }
 
     /**
@@ -1496,7 +1520,7 @@ class PutOperation {
         if (!operationTracker.hasSucceeded()) {
           failedAttempts++;
           appendSlippedPutBlobId(chunkBlobId);
-          if (failedAttempts <= routerConfig.routerMaxSlippedPutAttempts) {
+          if (failedAttempts <= routerConfig.routerMaxSlippedPutAttempts && !shouldUseReservedMetadataId()) {
             logger.trace("{}: Attempt to put chunk with id: {} failed, attempting slipped put ", loggingContext,
                 chunkBlobId);
             routerMetrics.slippedPutAttemptCount.inc();
@@ -1535,6 +1559,14 @@ class PutOperation {
         }
         state = ChunkState.Complete;
       }
+    }
+
+    /**
+     * @return {@code true} If this chunk is a metadata chunk and router reserved metadata is enabled. {@code false}
+     * otherwise.
+     */
+    boolean shouldUseReservedMetadataId() {
+      return isMetadataChunk() && routerConfig.routerReservedMetadataEnabled;
     }
 
     /**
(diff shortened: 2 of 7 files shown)


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1841,6 +1908,60 @@ void addChunkId(StoreKey storeKey, long chunkSize, int chunkIndex) {
...
     }
 
+    /**
+     * Validate the reserved metadata chunk id is passed in correctly where it needs to be specified.
+     * @throws RouterException in case of any errors.
+     */
+    private void validateReservedMetadataChunkId() throws RouterException {
+      if (!options.isChunkUpload() && Objects.nonNull(passedInBlobProperties.getReservedMetadataBlobId())) {
+        reservedMetadataIdMetrics.reservedMetadataPassedInForNonChunkedUploadCount.inc();
+        throw new RouterException("Reserved metadata should be passed in for request that isn't chunked upload.",
```
tbingol, line 1918: Is this error message correct? I think you meant "should not". Regardless, can you simplify your messages? Something like, "Only chunk uploads can reserve metadata blob id"

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1841,6 +1908,60 @@ void addChunkId(StoreKey storeKey, long chunkSize, int chunkIndex) {
...
+            RouterErrorCode.InvalidOrMismatchedStitchBlobReservedMetadataChunkId);
+      }
+      if (options.isChunkUpload() && passedInBlobProperties.getReservedMetadataBlobId() == null) {
+        reservedMetadataIdMetrics.noReservedMetadataForChunkedUploadCount.inc();
+        throw new RouterException("No reserved metadata passed in for chunked upload",
```
tbingol, line 1923: Don't we have a config to enable the reservation? If it is set to false, should expect reserved metadata id?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1248,6 +1258,29 @@ private void prepareForSending() {
...
+     * @return the blob id to use as the reserved metadata id. Can be {@code null}.
+     */
+    private String resolveReservedMetadataIdForDataChunk() {
+      if (!routerConfig.routerReservedMetadataEnabled) {
+        return null;
+      }
+      if (options.isChunkUpload()) {
+        return passedInBlobProperties.getReservedMetadataBlobId();
+      }
+      if (!chunkFillingCompletedSuccessfully || (chunkFillingCompletedSuccessfully
```
JingQianCloud, line 1277: Can you embed comment for each condition in the code instead of in the function header?
I think "chunkFillingCompletedSuccessfully" means it's a data chunk.

What does this condition mean? Is this the last data chunk of the composite blob?
(chunkFillingCompletedSuccessfully && && getBlobSize() > routerConfig.routerMaxPutChunkSizeBytes)

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1442,7 +1475,7 @@ void checkAndMaybeComplete() {
       // Now, check if this chunk is done.
       if (operationTracker.isDone() || (chunkException != null
           && chunkException.getErrorCode() == RouterErrorCode.TooManyRequests)) {
-        if (!operationTracker.hasSucceeded()) {
+        if (!operationTracker.hasSucceeded() && !shouldUseReservedMetadataId()) {
```
JingQianCloud, line 1478: Do we have an issue here?
The else condition of if (!operationTracker.hasSucceeded) will set the "chunkException" to null and "done to true".
That means if it's meta chunk failure, it'll be treated as successfully.

I guess your intention is to skip failure retry. If so, we should add the "!shouldUseReservedMetadataId" condition to the following 
```
if (failedAttempts <= routerConfig.routerMaxSlippedPutAttempts) {
}
```

JingQianCloud: Do you have the test case to cover metadata write failure?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1486,6 +1519,14 @@ void checkAndMaybeComplete() {
       }
     }
 
+    /**
+     * @return {@code true} If this chunk is a metadata chunk and router reserved metadata is enabled. {@code false}
+     * otherwise.
+     */
+    boolean shouldUseReservedMetadataId() {
+      return this instanceof MetadataPutChunk && routerConfig.routerReservedMetadataEnabled;
```
JingQianCloud, line 1527: nit: isMetadataChunk()

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1795,19 +1836,45 @@ private class MetadataPutChunk extends PutChunk {
...
-          ClusterMapUtils.reserveMetadataBlobId(partitionClass, Collections.emptyList(), reservedMetadataIdMetrics,
-              clusterMap, passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(),
-              passedInBlobProperties.isEncrypted(), routerConfig);
+      try {
+        validateReservedMetadataChunkId();
+        if (options.isChunkUpload()) {
+          setReservedMetadataChunkId(passedInBlobProperties.getReservedMetadataBlobId());
+        } else if (chunksToStitch == null) {
+          // if it's not chunked upload, and chunksToStitch is null, then it's a regular upload.
+          reservedMetadataChunkId =
```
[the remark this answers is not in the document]
JingQianCloud: +1

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
JingQianCloud: If we can have some wrap functions like IsDataChunk(), IsMetaChunk() etc, that'll be great.

ambry-api/src/main/java/com/github/ambry/frontend/ReservedMetadataIdMetrics.java
```diff
@@ -54,6 +58,14 @@ private ReservedMetadataIdMetrics(MetricRegistry metricRegistry) {
         MetricRegistry.name(ReservedMetadataIdMetrics.class, "NoReservedMetadataFoundForChunkedUploadResponseCount"));
     noReservedMetadataForChunkedUploadCount = metricRegistry.counter(
         MetricRegistry.name(ReservedMetadataIdMetrics.class, "NoReservedMetadataForChunkedUploadCount"));
+    metadataIdNotReservedCount = metricRegistry.counter(
+        MetricRegistry.name(ReservedMetadataIdMetrics.class, "MetadataIdNotReservedCount"));
+    stitchedBlobMetadataIdDeserErrorCount = metricRegistry.counter(
+        MetricRegistry.name(ReservedMetadataIdMetrics.class, "StitchedBlobMetadataIdDeserErrorCount"));
+    stitchedBlobMetadataIdMismatchCount = metricRegistry.counter(
+        MetricRegistry.name(ReservedMetadataIdMetrics.class, "StitchedBlobMetadataIdMismatchCount"));
+    reservedMetadataPassedInForNonChunkedUploadCount = metricRegistry.counter(
```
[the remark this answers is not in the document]
ankagrawal: fixed.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: Thats right. Thanks for catching it. Fixed it.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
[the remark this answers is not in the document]
ankagrawal: `setReservedMetadataId` sets the reserved metadataid for the operation using the `chunkId` String passed to it. But this line is executed if this is a non-chunked upload for a composite blob. In this case, we create a metadata blobid to reserve it.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: done.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: `validateReservedMetadataChunkId` will always throw an exception for invalid cases. The caller can then make a determination how it wants to handle the exception.
In this case, the `MetadataPutChunk` constructor that calls it, handles the exception based on whether config is enabled or not.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1846,19 +1877,51 @@ private class MetadataPutChunk extends PutChunk {
...
+              ClusterMapUtils.reserveMetadataBlobId(partitionClass, Collections.emptyList(), reservedMetadataIdMetrics,
+                  clusterMap, passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(),
+                  passedInBlobProperties.isEncrypted(), routerConfig);
+        } else if (!chunksToStitch.isEmpty()) {
+          // Empty chunksToStitch list will be handled by {@link processChunksToStitch}.
```
JingQianCloud, line 1910: Shall we just fail the PutOperation here when chunksToStitch is not null but isEmpty instead of waiting late for the processChunksToStitch.

The old code probably is ok since it hasn't used chunksToStitch yet but with the new code, it's a little bit weird to delay the error return.

And with the new code, we complete the request in constructor already with the following try and catch code block.

```
    try {
      metadataPutChunk = new MetadataPutChunk(chunksToStitch);
    } catch (RouterException routerException) {
      setOperationExceptionAndComplete(routerException);
    }
```

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1892,6 +1955,65 @@ void addChunkId(StoreKey storeKey, long chunkSize, int chunkIndex) {
       indexToChunkIdsAndChunkSizes.put(chunkIndex, new Pair<>(storeKey, chunkSize));
     }
 
+    /**
+     * Validate the reserved metadata chunk id is passed in correctly where it needs to be specified.
+     * @throws RouterException in case of any errors.
+     */
+    private void validateReservedMetadataChunkId() throws RouterException {
+      if (!options.isChunkUpload() && Objects.nonNull(passedInBlobProperties.getReservedMetadataBlobId())) {
```
JingQianCloud, line 1963: Just a question, right now our code set passedInBlobProperties.reservedMetadataBlobId already for the chunk upload case?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1846,19 +1877,51 @@ private class MetadataPutChunk extends PutChunk {
...
+          setReservedMetadataChunkId(passedInBlobProperties.getReservedMetadataBlobId());
+        } else if (chunksToStitch == null) {
+          // if it's not chunked upload, and chunksToStitch is null, then it's a regular upload.
+          reservedMetadataChunkId =
+              ClusterMapUtils.reserveMetadataBlobId(partitionClass, Collections.emptyList(), reservedMetadataIdMetrics,
+                  clusterMap, passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(),
+                  passedInBlobProperties.isEncrypted(), routerConfig);
+        } else if (!chunksToStitch.isEmpty()) {
+          // Empty chunksToStitch list will be handled by {@link processChunksToStitch}.
+          setReservedMetadataChunkId(chunksToStitch.get(0).getReservedMetadataId());
```
JingQianCloud, line 1911: With existing code, for each chunksToStitch[x], do we provide the reserved meta chunk already?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
JingQianCloud: Again, shall we use isMetadataChunk instead of "this instance of MetadataPutChunk"?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: Yes. For chunked upload when a user creates a signed url, the reserved metadata id is embedded in the signed url (AmbryUrlSigningService.java:180). And when the user attempts to upload using the signed url, the reserved metadata chunk is parsed from the signed header in `PostBlobHandler::152`.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1892,6 +1955,65 @@ void addChunkId(StoreKey storeKey, long chunkSize, int chunkIndex) {
...
+      }
+    }
+
+    /**
+     * Validate and set the specified chunkId as the reserved metadata chunk id.
+     * @param chunkId The chunk id to set.
+     * @throws RouterException In case of any issue during validation.
+     */
+    private void setReservedMetadataChunkId(String chunkId) throws RouterException {
+      if (chunkId == null) {
```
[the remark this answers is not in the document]
ankagrawal: Actually `chunkId == null` should never really happen. I am removing this check.
This method `setReservedMetadataChunkId ` is called from two places in the `MetadataChunk` constructor. And in both the cases there is enough safeguard against `chunkId == null`.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: fixed it.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -319,12 +319,25 @@ private PutOperation(RouterConfig routerConfig, NonBlockingRouterMetrics routerM
...
     putChunks = new ConcurrentLinkedQueue<>();
-    reservedMetadataIdMetrics = ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(routerMetrics.getMetricRegistry());
-    metadataPutChunk = new MetadataPutChunk();
+    reservedMetadataIdMetrics =
+        ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(routerMetrics.getMetricRegistry());
     chunkFillerChannel = new ByteBufferAsyncWritableChannel(writableChannelEventListener);
     isEncryptionEnabled = passedInBlobProperties.isEncrypted();
     restRequest = options.getRestRequest();
     loggingContext = makeLoggingContext();
+    if (chunksToStitch.isEmpty()) {
```
JingQianCloud, line 328: chunksToStitch can be null here.  Should be

```
if (chunksToStitch !=null && chunksToStitch.isEmpty())
```

Suppose we should have unit test fail. If not, we should add unit test.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1293,6 +1304,19 @@ private void prepareForSending() {
...
     }
 
+    /**
+     * Resolve the reserved metadata id to be used for data chunks.
+     * For simple blobs or metadata chunks this will be {@code null}. For the data chunks of composite blobs, the reserved
+     * metadata id of the metadata chunk will be used.
+     * @return the blob id to use as the reserved metadata id. Can be {@code null}.
+     */
+    private String resolveReservedMetadataId() {
+      if (!routerConfig.routerReservedMetadataEnabled || isSimpleBlob || isMetadataChunk()) {
```
JingQianCloud, line 1314: I thought we pass the reserved meta data id to the properties no matter we enable the flag or not?
```
!routerConfig.routerReservedMetadataEnabled
```

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1892,6 +1956,62 @@ void addChunkId(StoreKey storeKey, long chunkSize, int chunkIndex) {
...
+      }
+    }
+
+    /**
+     * Verify that the reserved metadata chunk id matches the specified chunk id.
+     * @param chunkId chunk id to be matched.
+     * @throws RouterException In case of a mismatch.
+     */
+    private void verifyStitchedBlobReservedMetadataChunkId(String chunkId) throws RouterException {
+      if (reservedMetadataChunkId == null && chunkId == null) {
```
JingQianCloud, line 2002: As you mentioned we provide the reserved meta data id already even the flag is not enable, right?
If so, this is an error case, we shouldn't ignore, right?

ankagrawal: Actually we don't need to handle this case separately. This will already be handled by the code before this and in this method.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1846,19 +1878,51 @@ private class MetadataPutChunk extends PutChunk {
...
+                  passedInBlobProperties.isEncrypted(), routerConfig);
+        } else if (!chunksToStitch.isEmpty()) {
+          // Empty chunksToStitch list will be handled by {@link processChunksToStitch}.
+          setReservedMetadataChunkId(chunksToStitch.get(0).getReservedMetadataId());
+          for (ChunkInfo chunkInfo : chunksToStitch) {
+            verifyStitchedBlobReservedMetadataChunkId(chunkInfo.getReservedMetadataId());
+          }
+        }
+      } catch (RouterException routerException) {
+        if (routerConfig.routerReservedMetadataEnabled) {
```
Arun-LinkedIn, line 1918: Minor: Would it be better if we avoid throwing exception in `validateReservedMetadataChunkId` if `routerReservedMetadataEnabled`? This seems to look like something went wrong but we are consuming the exception here :)

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1846,19 +1878,51 @@ private class MetadataPutChunk extends PutChunk {
...
+        if (options.isChunkUpload()) {
+          setReservedMetadataChunkId(passedInBlobProperties.getReservedMetadataBlobId());
+        } else if (chunksToStitch == null) {
+          // if it's not chunked upload, and chunksToStitch is null, then it's a regular upload.
+          reservedMetadataChunkId =
+              ClusterMapUtils.reserveMetadataBlobId(partitionClass, Collections.emptyList(), reservedMetadataIdMetrics,
+                  clusterMap, passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(),
+                  passedInBlobProperties.isEncrypted(), routerConfig);
+        } else if (!chunksToStitch.isEmpty()) {
+          // Empty chunksToStitch list will be handled by {@link processChunksToStitch}.
```
Arun-LinkedIn, line 1911: nit: May be this comment will be `// Empty chunksToStitch list will be handled by constructor` now.

Arun-LinkedIn (approved):
Mostly looks good. Just a couple of minor comments.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: We are not serializing the reserved metadata to disk yet, but we are still assigning reserved metadata for Post operations. For example, if its a chunked upload, the signed url created to do the upload will have a reserved metadata id embedded in it.  `validateReservedMetadataChunkId` will check against that embedded reserved metadata id. It doesn't need to get the reserved metadata id from the ambry storage.

Also note that unless `routerConfig.routerReservedMetadataEnabled` is true, no exception will be actually thrown out of the `MetadataChunk` constructor.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: Its not just `validateReservedMetadataChunkId()` method that can throw exception. The methods `setReservedMetadataChunkId`, `reserveMetadataBlobId` and `verifyStitchedBlobReservedMetadataChunkId` can also throw exceptions. Thats why I am handling everything at one place.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: updated the comments.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: We will not. `isReservedMetadataEnabled` will affect only whether the reserved metadata chunk id is sent to storage nodes or not.

PR #2442, opened 2023-05-04
StatsManager: add a real listener to remove host from account stats store

justinlin-linkedin:
This PR adds a real listener in StatsManager to remove hosts from account stats store when the host is removed from clustermap. However, there are some challenges to this. All the ambry server would listen to this clustermap change event and run the same piece of code to remove hosts from account stats store. We don't need all those ambry-servers to do the same thing, so we need some mechanism to coordinate those hosts to make sure there is only one host executing this task.

In order to do so, we are using helix's distributed lock. We will create a distributed lock for each host that is removed, and all the ambry-server hosts would try to acquire this lock. However, only one server host would succeed. That's how we guarantee that there is only one ambry-server doing this work.

### DIFF
--- a/ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
@@ -22,6 +22,7 @@ import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.ClusterMapChangeListener;
 import com.github.ambry.clustermap.ClusterParticipant;
 import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.clustermap.DistributedLock;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.PartitionStateChangeListener;
 import com.github.ambry.clustermap.ReplicaId;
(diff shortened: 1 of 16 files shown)


SophieGuo410 (approved):

ambry-server/src/main/java/com/github/ambry/server/StatsManager.java
```diff
@@ -468,13 +471,53 @@ List<String> examineUnreachablePartitions(List<PartitionId> unreachablePartition
...
+      }
+    }
+
+    private void tryRemoveHostFromAccountStatsStore(DataNodeId removedDataNode) {
+      String resource =
+          "%s_%s_%d".format(LOCK_RESOURCE_PREFIX + removedDataNode.getHostname(), removedDataNode.getPort());
+      try {
+        DistributedLock lock = clusterParticipant.getDistributedLock(resource, "AccountStatsStoreRemoval");
+        if (lock.tryLock()) {
+          logger.info("Acquired distributed lock for host removal: {}_{}", removedDataNode.getHostname(),
```
Arun-LinkedIn, line 500: Just a question. With distributed lock, would all hosts still execute the common code in a serialized manner?

justinlin-linkedin: All the hosts would be triggered to execute this task, but the first thing in this task is to acquire lock. I set the lock timeout to be 10 minutes, which means, whoever gets the lock, will hold it for 10 minutes. All the server hosts, as long as they receive the host removal messages in 10 minute, they would fail when acquiring the lock (except for the first server host). In this way, we can be sure that there are only one server host executing this code.

Arun-LinkedIn (approved):

PR #2443, opened 2023-05-05
Allow empty replicas to be deactivated immediately since there is no need for peers to sync up

Arun-LinkedIn:
Issue: An empty replica still has a log size of 18 bytes for its header. During it's deactivation process, we think that its peers are 18 bytes behind and wait indefinitely for them to catch up during INACTIVE to OFFLINE transition.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
@@ -360,7 +360,10 @@ public class ReplicationManager extends ReplicationEngine {
         throw new StateTransitionException(
             "Store " + partitionName + " is not started during Standby-To-Inactive transition", StoreNotStarted);
       }
-      replicaSyncUpManager.initiateDeactivation(localReplica);
+      if (!store.isEmpty()) {
+        // Wait for peers to sync up if this replica is not empty
+        replicaSyncUpManager.initiateDeactivation(localReplica);
+      }
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -381,7 +384,10 @@ public void onPartitionBecomeOfflineFromInactive(String partitionName) {
       }
       // set local store state to OFFLINE and initiate disconnection
       store.setCurrentState(ReplicaState.OFFLINE);
-      replicaSyncUpManager.initiateDisconnection(localReplica);
+      if (!store.isEmpty()) {
+        // Wait for peers to sync up if this replica is not empty
+        replicaSyncUpManager.initiateDisconnection(localReplica);
```
SophieGuo410, line 389: nit: not related to this pr, but the initiateDisconnection and initiateDeactivation looks a little bit confused to me that I thought every store need to go through this process when it has been marked as OFFLINE. However, it seems like only the non-empty store would need to go through this process in order to make sure peer replica catch up everything. Can we add the comment in the interface side to make it more clear?

SophieGuo410 (approved):
lgtm! just left one minor comment.

justinlin-linkedin (approved):
LGTM

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
Arun-LinkedIn: Sure. Added comments that we don't need this step for empty replica.

PR #2446, opened 2023-05-08
[Efficient Metadata Operations] Add reserved metadata id property in the ChunkInfo class.

ankagrawal:
The chunked uploads will have reserved metadata pre-reserved for them when a signed url request was made with the "chunked-upload" header set to true. This pre-reserved metadata id will be encoded in the signed-url.
On chunk upload, this information will need to be decoded and passed onto the PutOperation. Adding reserved metadata id property in the ChunkInfo class facilitates passing this information.


JingQianCloud (approved):
LGTM

Arun-LinkedIn (approved):

PR #2448, opened 2023-05-08
Remove generated colume for replication policy migration

SophieGuo410:
Mysql team mentioned that if we want mysql to support data migration, the schema should be exactly same.
And since multi upload(GG) database can't support generated column, we have to support it at application side.
This pr is to remove the generated column and support same function at application side.


Arun-LinkedIn (approved):
Code changes looks good! Do we have to follow any steps for updating schema of the table on MySql side before deploying application (Ambry) code changes?

SophieGuo410:
> Code changes looks good! Do we have to follow any steps for updating schema of the table on MySql side before deploying application (Ambry) code changes?

since for the new schema we support GG replication, so we have to ask mysql team to help us build the table based on the new schema. then we will enable this enableNewDatabaseForMigration which disableGenerateColumn set to true for cachedAccountServiceNew.

justinlin-linkedin (approved):

PR #2449, opened 2023-05-09
[Efficient Metadata Operations] Reserve metadata id for signed-url request for POST with chunked-upload enabled.

ankagrawal:

This PR handles case 2.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
@@ -13,7 +13,13 @@
  */
 package com.github.ambry.frontend;
 
+import com.github.ambry.account.Account;
+import com.github.ambry.account.Container;
 import com.github.ambry.clustermap.ClusterMap;
+import com.github.ambry.clustermap.ClusterMapUtils;
+import com.github.ambry.commons.BlobId;
+import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.config.RouterConfig;
 import com.github.ambry.rest.RestMethod;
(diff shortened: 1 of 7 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
```diff
@@ -149,6 +165,20 @@ public String getSignedUrl(RestRequest restRequest) throws RestServiceException
         }
         argsForUrl.put(RestUtils.Headers.CHUNK_UPLOAD, true);
         argsForUrl.put(RestUtils.Headers.SESSION, UUID.randomUUID().toString());
+        BlobId reservedMetadataBlobId = ClusterMapUtils.reserveMetadataBlobId(
```
JingQianCloud, line 168: No matter isReservedMetadataEnabled is enabled or disabled, we always generate the reservedMetadataBlobId.
Is this a bug?

ankagrawal: No this is not a bug. The idea is to do all the while the feature is disabled to collect the metrics. Any user facing changes (like reserved metadata chunk showing up in the signed-url, or exception getting thrown due to reserved metadata chunk), will happen only when the feature is enabled.
Seeing that all metrics look ok will give us enough confidence in a few weeks to enable the feature everywhere.

JingQianCloud (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryUrlSigningService.java
ankagrawal: Updated the comments in RouterConfig.

PR #2451, opened 2023-05-09
Support enable get from new db only

SophieGuo410:
This pr support:
1. when we have new db set up, we directly issue dataset request against new db since we don't need to migrate any data.
2. when ignoreNewDatabaseUploadError has been enabled, we ignore the upload failure against new db. this can be used at early stage where we don't want the new db upload affect the old db upload until we verify the data are in sync.
3. when enableGetFromNewDbOnly has been enabled, we only get data from new database. this will be enabled after the migration is done and been verified.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
@@ -502,6 +502,9 @@ public class CachedAccountService extends AbstractAccountService {
     try {
       Pair<Short, Short> accountAndContainerIdPair =
           getAccountAndContainerIdFromName(dataset.getAccountName(), dataset.getContainerName());
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       mySqlAccountStore.addDataset(accountAndContainerIdPair.getFirst(), accountAndContainerIdPair.getSecond(),
           dataset);
     } catch (SQLException e) {
@@ -514,6 +517,9 @@ public class CachedAccountService extends AbstractAccountService {
     try {
       Pair<Short, Short> accountAndContainerIdPair =
           getAccountAndContainerIdFromName(dataset.getAccountName(), dataset.getContainerName());
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       mySqlAccountStore.updateDataset(accountAndContainerIdPair.getFirst(), accountAndContainerIdPair.getSecond(),
           dataset);
     } catch (SQLException e) {
@@ -526,6 +532,9 @@ public class CachedAccountService extends AbstractAccountService {
       throws AccountServiceException {
     try {
       Pair<Short, Short> accountAndContainerIdPair = getAccountAndContainerIdFromName(accountName, containerName);
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       return mySqlAccountStore.getDataset(accountAndContainerIdPair.getFirst(), accountAndContainerIdPair.getSecond(),
           accountName, containerName, datasetName);
     } catch (SQLException e) {
@@ -538,6 +547,9 @@ public class CachedAccountService extends AbstractAccountService {
       throws AccountServiceException {
     try {
       Pair<Short, Short> accountAndContainerIdPair = getAccountAndContainerIdFromName(accountName, containerName);
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       mySqlAccountStore.deleteDataset(accountAndContainerIdPair.getFirst(), accountAndContainerIdPair.getSecond(),
           datasetName);
     } catch (SQLException e) {
@@ -550,6 +562,9 @@ public class CachedAccountService extends AbstractAccountService {
       String version, long timeToLiveInSeconds, long creationTimeInMs, boolean datasetVersionTtlEnabled) throws AccountServiceException {
     try {
       Pair<Short, Short> accountAndContainerIdPair = getAccountAndContainerIdFromName(accountName, containerName);
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       return mySqlAccountStore.addDatasetVersion(accountAndContainerIdPair.getFirst(),
           accountAndContainerIdPair.getSecond(), accountName, containerName, datasetName, version, timeToLiveInSeconds,
           creationTimeInMs, datasetVersionTtlEnabled);
@@ -563,6 +578,9 @@ public class CachedAccountService extends AbstractAccountService {
       String version) throws AccountServiceException {
     try {
       Pair<Short, Short> accountAndContainerIdPair = getAccountAndContainerIdFromName(accountName, containerName);
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       return mySqlAccountStore.getDatasetVersion(accountAndContainerIdPair.getFirst(),
           accountAndContainerIdPair.getSecond(), accountName, containerName, datasetName, version);
     } catch (SQLException e) {
@@ -575,6 +593,9 @@ public class CachedAccountService extends AbstractAccountService {
       String version) throws AccountServiceException {
     try {
       Pair<Short, Short> accountAndContainerIdPair = getAccountAndContainerIdFromName(accountName, containerName);
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       mySqlAccountStore.deleteDatasetVersion(accountAndContainerIdPair.getFirst(),
           accountAndContainerIdPair.getSecond(), datasetName, version);
     } catch (SQLException e) {
@@ -587,6 +608,9 @@ public class CachedAccountService extends AbstractAccountService {
       throws AccountServiceException {
     try {
       Pair<Short, Short> accountAndContainerIdPair = getAccountAndContainerIdFromName(accountName, containerName);
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
       return mySqlAccountStore.getAllValidVersion(accountAndContainerIdPair.getFirst(),
           accountAndContainerIdPair.getSecond(), datasetName);
     } catch (SQLException e) {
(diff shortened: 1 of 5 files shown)


ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -108,28 +111,56 @@ Set<String> getRecentNotFoundContainersCache() {
...
   public boolean removeAccountUpdateConsumer(Consumer<Collection<Account>> accountUpdateConsumer) {
-    return cachedAccountService.removeAccountUpdateConsumer(accountUpdateConsumer);
+    boolean result = cachedAccountService.removeAccountUpdateConsumer(accountUpdateConsumer);
+    if (config.enableNewDatabaseForMigration) {
+      boolean resultNew = cachedAccountServiceNew.removeAccountUpdateConsumer(accountUpdateConsumer);
+      if (config.enableGetFromNewDbOnly) {
+        return resultNew;
+      }
+    }
+    return result;
```
justinlin-linkedin, line 163: I probably missed these two methods in previous PR, dealing with account update consumers should be orthogonal to the migration. It should be a function of MysqlAccountService, instead of cached mysql store.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -146,18 +177,43 @@ protected Account getAccountByNameNew(String accountName) {
...
 
   @Override
   public void updateAccounts(Collection<Account> accounts) throws AccountServiceException {
     cachedAccountService.updateAccounts(accounts);
     if (config.enableNewDatabaseForMigration) {
-      cachedAccountServiceNew.updateAccounts(accounts);
+      try {
+        cachedAccountServiceNew.updateAccounts(accounts);
+      } catch (AccountServiceException e) {
+        if (config.ignoreNewDatabaseUploadError) {
```
justinlin-linkedin, line 203: can we add a metric for all the ignored errors?

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
SophieGuo410: it seems like the notifyAccountUpdateConsumers will be called in CacheAccountService::fetchAndUpdateCache.
So if we don't support addAccountUpdateConsumer, the accountUpdateConsumers in mysqlAccountService and CacheAccountService will be the different objects. So when we call notifyAccountUpdateConsumers in CacheAccountService::fetchAndUpdateCache, accountUpdateConsumers will be an empty list. Please let me know if I misunderstand anything.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
SophieGuo410: Updated.

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -497,6 +497,171 @@ protected void publishChangeNotice() {
     super.publishChangeNotice();
   }
 
+  @Override
+  public void addDataset(Dataset dataset) throws AccountServiceException {
+    try {
+      String accountName = dataset.getAccountName();
+      String containerName = dataset.getContainerName();
+      Container container = getContainerByName(accountName, containerName);
```
tbingol, line 505: Nit: make them final

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -497,6 +497,171 @@ protected void publishChangeNotice() {
...
+      if (container == null) {
+        throw new AccountServiceException("Can't find the container: " + containerName + " in account: " + accountName,
+            AccountServiceErrorCode.BadRequest);
+      }
+      short accountId = container.getParentAccountId();
+      short containerId = container.getId();
```
tbingol, line 511: Nit: make them final

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -497,6 +497,171 @@ protected void publishChangeNotice() {
...
+      throw translateSQLException(e);
+    }
+  }
+
+  @Override
+  public void updateDataset(Dataset dataset) throws AccountServiceException {
+    try {
+      String accountName = dataset.getAccountName();
+      String containerName = dataset.getContainerName();
+      Container container = getContainerByName(accountName, containerName);
```
tbingol, line 526: Nit: make them final

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -497,6 +497,171 @@ protected void publishChangeNotice() {
...
+      if (container == null) {
+        throw new AccountServiceException("Can't find the container: " + containerName + " in account: " + accountName,
+            AccountServiceErrorCode.BadRequest);
+      }
+      short accountId = container.getParentAccountId();
+      short containerId = container.getId();
```
tbingol, line 532: Nit: make them final here and all other functions below

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -497,6 +497,171 @@ protected void publishChangeNotice() {
...
+      if (mySqlAccountStore == null) {
+        mySqlAccountStore = this.supplier.get();
+      }
```
tbingol, line 576: This code repeats in all functions. Consider putting it into a helper private function

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
```diff
@@ -108,28 +111,57 @@ Set<String> getRecentNotFoundContainersCache() {
   synchronized void fetchAndUpdateCache() throws SQLException {
     cachedAccountService.fetchAndUpdateCache();
     if (config.enableNewDatabaseForMigration) {
-      cachedAccountServiceNew.fetchAndUpdateCache();
+      try {
+        cachedAccountServiceNew.fetchAndUpdateCache();
+      } catch (SQLException e) {
+        if (config.ignoreNewDatabaseUploadError) {
```
tbingol, line 117: I am curious why we felt the need for this. Do we observe failures on the new DB? If so, it should be a concern when it becomes SOT

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
SophieGuo410: Updated.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
SophieGuo410: This can be used at early stage where we don't want the new db upload affect the old db upload until we verify everything is good.

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -632,6 +729,28 @@ private Pair<List<Container>, List<Container>> getUpdatedContainers(Account acco
...
+      throws AccountServiceException {
+    final Container container = getContainerByName(accountName, containerName);
+    if (container == null) {
+      throw new AccountServiceException("Can't find the container: " + containerName + " in account: " + accountName,
+          AccountServiceErrorCode.BadRequest);
+    }
+    final short accountId = container.getParentAccountId();
+    final short containerId = container.getId();
+    if (mySqlAccountStore == null) {
+      mySqlAccountStore = this.supplier.get();
```
Arun-LinkedIn, line 749: Minor: Just thinking that getting mysqlAccountStore here is little out of place. May be we can have within addDataset, getDataset, updateDataset itself although it leads to duplicated code?

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
Arun-LinkedIn: Would you disable `config.ignoreNewDatabaseUploadError` before setting `config.enableGetFromNewDbOnly` to `true` since we would need to catch upload errors at that time?

Arun-LinkedIn (approved):
Mostly looks good. Just a couple of minor comments

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
SophieGuo410: sure, updated.

SophieGuo410: Not sure why I could not leave comments for below one, so I just reply here. yes, I'll disable config.ignoreNewDatabaseUploadError before setting config.enableGetFromNewDbOnly. so the step would be:
1. support dual upload
2. one time dump data
3. once it's in sync and don't see any issue, disable config.ignoreNewDatabaseUploadError
4. after a while when it's stable, enable config.enableGetFromNewDbOnly to get from new db only
5. disable upload to old db.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
SophieGuo410: yes, I'll disable config.ignoreNewDatabaseUploadError before setting config.enableGetFromNewDbOnly. so the step would be:

support dual upload
one time dump data
once it's in sync and don't see any issue, disable config.ignoreNewDatabaseUploadError
after a while when it's stable, enable config.enableGetFromNewDbOnly to get from new db only
disable upload to old db.

ambry-account/src/main/java/com/github/ambry/account/MySqlAccountService.java
justinlin-linkedin: I think we should be careful about this one. Basically, this consumer callback is passed by some other components to get notified on the account updates. It's basically a passive read operation. If we can reading from old database, then we should only register this consumer callback to the old cached account service. If we are reading from the new database, then we should only register this consumer callback to the new cached account service.

It might be fine for the consumer to be called twice, once from old cached account service and the other one from the new, but I am afraid before data migration, these two invocation would read two different account lists.

tbingol (approved):

justinlin-linkedin (approved):
LGTM

PR #2452, opened 2023-05-10
[Efficient Metadata Operations] For chunked upload requests decode the reserved metadata id from the request and encode the reserved metadata in the response.

ankagrawal:

This PR implements cases 3 & 4.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
@@ -425,6 +425,7 @@ public class NamedBlobPutHandler {
      */
     List<ChunkInfo> getChunksToStitch(BlobProperties stitchedBlobProperties, JSONObject stitchRequestJson)
         throws RestServiceException {
+      String reservedMetadataBlobId = null;
       List<String> signedChunkIds = StitchRequestSerDe.fromJson(stitchRequestJson);
       if (signedChunkIds.isEmpty()) {
         throw new RestServiceException("Must provide at least one ID in stitch request",
@@ -456,8 +457,9 @@ public class NamedBlobPutHandler {
         @SuppressWarnings("ConstantConditions")
         long expirationTimeMs = RestUtils.getLongHeader(metadata, EXPIRATION_TIME_MS_KEY, true);
         verifyChunkAccountAndContainer(blobId, stitchedBlobProperties);
+        reservedMetadataBlobId = getAndVerifyReservedMetadataBlobId(metadata, reservedMetadataBlobId, blobId);
 
-        chunksToStitch.add(new ChunkInfo(blobId, chunkSizeBytes, expirationTimeMs, null));
+        chunksToStitch.add(new ChunkInfo(blobId, chunkSizeBytes, expirationTimeMs, reservedMetadataBlobId));
       }
       //the actual blob size for stitched blob is the sum of all the chunk sizes
       restResponseChannel.setHeader(RestUtils.Headers.BLOB_SIZE, totalStitchedBlobSize);
@@ -486,6 +488,36 @@ public class NamedBlobPutHandler {
             + stitchedBlobProperties.getAccountId() + ", " + stitchedBlobProperties.getContainerId() + ")",
             RestServiceErrorCode.BadRequest);
       }
+    }
+
+    /**
+     * Verify that the reserved metadata id for the specified chunkId is same as seen for previous chunks.
+     * Also return the chunk's reserved metadata id.
+     * @param metadata {@link Map} of metadata set in the signed ids.
+     * @param reservedMetadataBlobId Reserved metadata id for the chunks. Can be {@code null}.
+     * @param chunkId The chunk id.
+     * @return The reserved metadata id.
+     * @throws RestServiceException in case of any exception.
+     */
+    private String getAndVerifyReservedMetadataBlobId(Map<String, String> metadata, String reservedMetadataBlobId,
+        String chunkId) throws RestServiceException {
+      String chunkReservedMetadataBlobId = RestUtils.getHeader(metadata, RestUtils.Headers.RESERVED_METADATA_ID, false);
+      if (chunkReservedMetadataBlobId == null) {
+        ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(
+            frontendMetrics.getMetricRegistry()).noReservedMetadataForChunkedUploadCount.inc();
+        throwRestServiceExceptionIfEnabled(
+            new RestServiceException(String.format("No reserved metadata id present in chunk %s signed url", chunkId),
+                RestServiceErrorCode.BadRequest), router.getRouterConfig().routerReservedMetadataEnabled);
+      }
+      if (reservedMetadataBlobId != null && !reservedMetadataBlobId.equals(chunkReservedMetadataBlobId)) {
+        ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(
+            frontendMetrics.getMetricRegistry()).mismatchedReservedMetadataForChunkedUploadCount.inc();
+        throwRestServiceExceptionIfEnabled(new RestServiceException(String.format(
+                "Reserved metadata id for the chunks are not same. For chunk: %s the reserved metadata id is %s. But reserved metadata id %s was found earlier.",
+                chunkId, chunkReservedMetadataBlobId, reservedMetadataBlobId), RestServiceErrorCode.BadRequest),
+            router.getRouterConfig().routerReservedMetadataEnabled);
+      }
+      return chunkReservedMetadataBlobId;
     }
 
     /**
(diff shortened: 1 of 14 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -488,6 +490,34 @@ private void verifyChunkAccountAndContainer(String chunkBlobId, BlobProperties s
...
+      String chunkReservedMetadataBlobId = RestUtils.getHeader(metadata, RestUtils.Headers.RESERVED_METADATA_ID, false);
+      if (reservedMetadataBlobId == null && chunkReservedMetadataBlobId == null) {
+        ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(
+            frontendMetrics.getMetricRegistry()).noReservedMetadataForChunkedUploadCount.inc();
+        if (router.getRouterConfig().routerReservedMetadataEnabled) {
+          throw new RestServiceException(String.format("No reserved metadata id present in chunk %s signed url", chunkId),
+              RestServiceErrorCode.BadRequest);
+        }
+      }
+      if (reservedMetadataBlobId != null && !reservedMetadataBlobId.equals(chunkReservedMetadataBlobId)) {
```
JingQianCloud, line 513: When router.getRouterConfig().routerReservedMetadataEnabled is true, the logic is right.
But if it's disabled, some case we throw exception but some case we don't when there is mismatch.
1. First chunk doesn't have the metadata blob id, second chunk has the metadata id. --> No exception.
2. first chunk has the meta blob id, second doesn't have --> throw exception.

The routerReservedMetadataEnabled flag doesn't fully control all the "metadata blob" related logic. It's not clear to me our intention when routerReservedMetadataEnabled is disabled.

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -322,6 +322,16 @@ private void setSignedIdMetadataAndBlobSize(BlobProperties blobProperties) throw
             RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.SESSION, true));
         metadata.put(EXPIRATION_TIME_MS_KEY,
             Long.toString(Utils.addSecondsToEpochTime(time.milliseconds(), blobProperties.getTimeToLiveInSeconds())));
+        if (blobProperties.getReservedMetadataBlobId() != null) {
+          metadata.put(RestUtils.Headers.RESERVED_METADATA_ID, blobProperties.getReservedMetadataBlobId());
+        } else {
+          ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(
+              frontendMetrics.getMetricRegistry()).noReservedMetadataFoundForChunkedUploadResponseCount.inc();
+          if (router.getRouterConfig().routerReservedMetadataEnabled) {
+            throw new RestServiceException("No reserved metadata id present to set in chunked upload response",
```
JingQianCloud, line 331: If I understand the code logic correctly, setSignedIdMetadataAndBlobSize is called in the PutBlobCallback.
I suppose we have checked the metadata blob exists beforehand already if routerReservedMetadataEnabled is enabled, right? Is that possible we hit getReservedMetadataBlobId == null ?

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -418,5 +430,33 @@ private void verifyChunkAccountAndContainer(String chunkBlobId, BlobProperties s
...
+    /**
+     * Verify that the reserved metadata id for the specified chunkId is same as seen for previous chunks.
+     * Also return the chunk's reserved metadata id.
+     * @param metadata {@link Map} of metadata set in the signed ids.
+     * @param reservedMetadataBlobId Reserved metadata id for the chunks. Can be {@code null}.
+     * @param chunkId The chunk id.
+     * @return The reserved metadata id.
+     * @throws RestServiceException in case of any exception.
+     */
+    private String getAndVerifyReservedMetadataBlobId(Map<String, String> metadata, String reservedMetadataBlobId,
```
JingQianCloud, line 443: Same questions as the namedBlobPutXXXhandler.getAndVerifyReservedMetadataBlobId

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -488,6 +490,34 @@ private void verifyChunkAccountAndContainer(String chunkBlobId, BlobProperties s
...
+     * @param reservedMetadataBlobId Reserved metadata id for the chunks. Can be {@code null}.
+     * @param chunkId The chunk id.
+     * @return The reserved metadata id.
+     * @throws RestServiceException in case of any exception.
+     */
+    private String getAndVerifyReservedMetadataBlobId(Map<String, String> metadata, String reservedMetadataBlobId,
+        String chunkId) throws RestServiceException {
+      String chunkReservedMetadataBlobId = RestUtils.getHeader(metadata, RestUtils.Headers.RESERVED_METADATA_ID, false);
+      if (reservedMetadataBlobId == null && chunkReservedMetadataBlobId == null) {
+        ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(
```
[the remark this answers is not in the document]
ankagrawal: This method will be called only for stitch blob requests. So even if someone is stitching only one data chunk, a metadata chunk will still be created. Also, when someone created the POST-signed-URL to upload the data chunk, then the metadata chunk should have been reserved and that metadata chunk should ultimately be used for saving the stitched blob.

When no metadata chunk is reserved in the signed-id of the chunk to stitch, then this will increment the metric. This is to give us some metric to track that there are some stitch blob uploads, that don't have metadata id reserved. Before enabling "routerReservedMetadataEnabled" we need to get evidence that all cases are covered and there is no flow that inadvertently not reserved metadata chunk. That's why we need the metric even though "routerReservedMetadataEnabled" is disabled.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
ankagrawal: If the first chunk didn't have a metadata blob id, then we would throw an exception on line 509. Once the first chunk is determined to have a metadata blob id, then line 518 will return that metadata blob id.

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
ankagrawal: Most likely this will never happen. But I want to cover all cases and make sure we have evidence that everything is working properly for a few weeks before I enabled this feature.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -456,8 +457,9 @@ List<ChunkInfo> getChunksToStitch(BlobProperties stitchedBlobProperties, JSONObj
         @SuppressWarnings("ConstantConditions")
         long expirationTimeMs = RestUtils.getLongHeader(metadata, EXPIRATION_TIME_MS_KEY, true);
         verifyChunkAccountAndContainer(blobId, stitchedBlobProperties);
+        reservedMetadataBlobId = getAndVerifyReservedMetadataBlobId(metadata, reservedMetadataBlobId, blobId);
```
Arun-LinkedIn, line 460: nit: would having an extra variable like `prevChunkReservedMetadataBlobId` make it easier to follow? For example:

```
String reservedMetadataBlobId = null;
String prevChunkReservedMetadataBlobId = null; 

for (String signedChunkId : signedChunkIds) {
    reservedMetadataBlobId = getAndVerifyReservedMetadataBlobId(metadata, 
    prevChunkReservedMetadataBlobId, blobId);
    chunksToStitch.add(new ChunkInfo(blobId, chunkSizeBytes, expirationTimeMs, reservedMetadataBlobId));
    prevChunkReservedMetadataBlobId = reservedMetadataBlobId;
}

```

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -322,6 +322,16 @@ private void setSignedIdMetadataAndBlobSize(BlobProperties blobProperties) throw
             RestUtils.getHeader(restRequest.getArgs(), RestUtils.Headers.SESSION, true));
         metadata.put(EXPIRATION_TIME_MS_KEY,
             Long.toString(Utils.addSecondsToEpochTime(time.milliseconds(), blobProperties.getTimeToLiveInSeconds())));
+        if (blobProperties.getReservedMetadataBlobId() != null) {
+          metadata.put(RestUtils.Headers.RESERVED_METADATA_ID, blobProperties.getReservedMetadataBlobId());
+        } else {
+          ReservedMetadataIdMetrics.getReservedMetadataIdMetrics(
+              frontendMetrics.getMetricRegistry()).noReservedMetadataFoundForChunkedUploadResponseCount.inc();
```
Arun-LinkedIn, line 329: QQ (Not related to this PR). For other metrics classes like RouterMetrics, FrontendMetrics, etc, we seem to create them in the factory classes. Would it involve lot of code changes to do similar for `ReservedMetadataIdMetrics` also? For example, would we have to create it at the beginning in `RestServer.java`, etc and pass it to all components?

Arun-LinkedIn (approved):
Looks good. Just couple of nit comments.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
ankagrawal: The thing is that the line " prevChunkReservedMetadataBlobId = reservedMetadataBlobId;" will become redundant above. Because either " prevChunkReservedMetadataBlobId" is always equal to "reservedMetadataBlobId", or `getAndVerifyReservedMetadataBlobId ` throws an exception.

JingQianCloud (approved):

PR #2453, opened 2023-05-10
[Full-Auto] Avoid using same disk for concurrent replica additions 

Arun-LinkedIn:
When multiple replicas are bootstrapped concurrently on helix callback threads, we used the same disk repeatedly since the used disk space is only updated in the end. To avoid it, update available disk space as soon as it is selected for a replica.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -522,6 +522,9 @@ public class StorageManager implements StoreManager {
         // sure old store of this replica is deleted (this store may be created in previous replica addition but failed
         // at some point). Then a brand new store associated with this replica should be created and started.
         if (!addBlobStore(replicaToAdd)) {
+          // We have decreased the available disk space in HelixClusterManager#getDiskForBootstrapReplica. Increase it
+          // back since addition of store failed.
+          replicaToAdd.getDiskId().increaseAvailableSpaceInBytes(replicaToAdd.getCapacityInBytes());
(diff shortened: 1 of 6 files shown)


justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -522,6 +522,9 @@ public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
         // sure old store of this replica is deleted (this store may be created in previous replica addition but failed
         // at some point). Then a brand new store associated with this replica should be created and started.
         if (!addBlobStore(replicaToAdd)) {
+          // We have decreased the available disk space in HelixClusterManager#getDiskForBootstrapReplica. Increase it
+          // back since addition of store failed.
+          replicaToAdd.getDiskId().increaseAvailableSpaceInBytes(replicaToAdd.getCapacityInBytes());
```
SophieGuo410, line 527: QQ: it seems like we only decreased the available disk space for FullAuto, but not for semi auto. is this backward compatible?

Arun-LinkedIn: Sure. Updated disk usage when creating a replica in Semi-auto mode too. However, for semi-auto, this value is not used since the 'disk -> replica' mapping is pre-decided during cluster map creation. We only use it in FullAuto since the 'disk -> replica' mapping is done in code.

SophieGuo410 (approved):

PR #2454, opened 2023-05-10
[Full-auto] Handle transition of a replica from offline to dropped state on host start up

Arun-LinkedIn:
In Helix full-auto mode, when a host is down, it is possible that its replicas are reassigned to a different host. After the host comes up, Helix may directly transition the replicas from `OFFLINE -> DROPPED` state without going through `OFFLINE -> BOOTSTRAP -> STANDBY -> INACTIVE -> OFFLINE -> DROPPED` steps. In this case, we have to make sure the 'replica decommission' steps are followed to make sure peers have caught up with it before deleting the files.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -714,21 +714,13 @@ public class StorageManager implements StoreManager {
       // without stopping it)
       BlobStore store = (BlobStore) getStore(replica.getPartitionId(), true);
 
-      // 1. check is the store is recovering from decommission or directly transitioning from OFFLINE to DROPPED in
-      // full-auto.
-      boolean resumeDecommission = false;
-      if (store.recoverFromDecommission()) {
-        resumeDecommission = true;
-      } else if (clusterMap.isDataNodeInFullAutoMode(replica.getDataNodeId())
-          && store.getPreviousState() == ReplicaState.OFFLINE) {
-        // In Full-auto mode, it is possible that this replica has been reassigned to a different host when it is down.
-        // After the host comes up, Helix may directly transition the replica from OFFLINE -> DROPPED without going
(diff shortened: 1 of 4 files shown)


ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -713,10 +713,22 @@ public void onPartitionBecomeDroppedFromOffline(String partitionName) {
...
-        // peer replicas have caught up with local replica and update InstanceConfig in Helix)
+        resumeDecommission = true;
+      } else if (clusterMap.isDataNodeInFullAutoMode(replica.getDataNodeId())
+          && store.getPreviousState() == ReplicaState.OFFLINE) {
+        // In Full-auto mode, it is possible that this replica has been reassigned to a different host when it is down.
+        // After the host comes up, Helix may directly transition the replica from OFFLINE -> DROPPED without going
+        // through OFFLINE -> BOOTSTRAP -> STANDBY -> INACTIVE -> OFFLINE -> DROPPED steps. So, we have to make sure
+        // peers have caught with this replica before deleting the files.
+        resumeDecommission = true;
+      }
```
justinlin-linkedin, line 729: nit: let's do
`
boolean resumeDecommision = store.recoverFromDecommission() || (clusterMap.isDataNodeInFullAutoMode(resplica.getDataNodeId()) && store.getPreviousState() == ReplicaState.OFFLINE);
`

justinlin-linkedin (commented):
Can you add a test case for this one?

SophieGuo410 (approved):

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
Arun-LinkedIn: Sure. Updated if-condition and added a test case.

PR #2455, opened 2023-05-12
[FS] Add database schema

justinlin-linkedin:
Start adding database schema for file system interface.

### DIFF
--- a/ambry-filesystem/src/main/resources/AmbryFS.ddl
+++ b/ambry-filesystem/src/main/resources/AmbryFS.ddl
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2023 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+/* CREATE DATABASE ambryfs; */
+
+CREATE TABLE IF NOT EXISTS DirectoryInfo (
+    account_id      int           NOT NULL,
+    container_id    int           NOT NULL,
+    parent_dir_id   binary(16)    NOT NULL,
+    resource_name   varchar(128)  NOT NULL,
+    resource_type   byte          NOT NULL,
+    resource_id     binary(16)    NOT NULL,
+    created_ts      datetime(6)   NOT NULL,
+    last_updated_ts datetime(6)   NOT NULL,
+    deleted_ts      datetime(6)   DEFAULT NULL,
+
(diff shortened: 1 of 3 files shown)


ambry-filesystem/src/main/resources/AmbryFS.ddl
```diff
@@ -0,0 +1,47 @@
...
+    container_id    int           NOT NULL,
+    parent_dir_id   binary(16)    NOT NULL,
+    resource_name   varchar(128)  NOT NULL,
+    resource_type   byte          NOT NULL,
+    resource_id     binary(16)    NOT NULL,
+    created_ts      datetime(6)   NOT NULL,
+    last_updated_ts datetime(6)   NOT NULL,
+    deleted_ts      datetime(6)   DEFAULT NULL,
+
+    PRIMARY KEY (account_id, container_id, parent_dir_id, resource_name)
```
JingQianCloud, line 28: If we have the resource_name in the key, the following two operation needs some effort.
1. user delete the folder and then recreate the folder with the same name.
2. rename the folder.

But if we exclude the "resource_name" from the key, the above two operation will be one shot operation.

I suggest,
1. PRIMAY KEY(account_id, container_id, parent_dir_id, resource_id).
2. Index on(account_id, container_id, parent_dir_id, resource_name).

I mentioned this in the doc as well. Looks like you have some concern about it. We can continue the discussion there.

justinlin-linkedin: There are hidden cost for secondary index. 
1. Updates to the table would be a bit slower
2. There will be some disk overhead.

For latency concern, I suppose we would have at least 200:1 ratio of insert to rename. In this case, as long as the latency cost for secondary index exceeds 0.5%, we are not getting any benefit.
For disk overhead, each innodb secondary index would include the primary key for each row so later it can use this primary key to fetch the row content. We have a pretty big primary key for this table, I am afraid the disk overhead isn't going to trivial. 

For these reasons, I prefer to not have secondary index for this table.

I can try to find some numbers to quantify the performance impact for having a secondary index.

JingQianCloud: That'll be great if we have some numbers for the secondary index.

Will see other people's opinions about it.

tbingol, line 28: I was hoping we could follow this model which is more k/v friendly: https://docs.google.com/document/d/1nXEXDhorJf4ylme9Eb8wqgs1ket5igTDgdYAhvoUp68/edit#bookmark=id.9ly0e83bv2p

Based on this model, a directory is keyed with parent_object_id/directory_name.

justinlin-linkedin: I don't think we have to worry about mysql database schema isn't kv friendly. If later we start using a simple key-value store, we would have a different implementation.

Also I think the account and container is useful to be part of the primary key in the table, so we can quickly find how many directories and files created by an account and a container.

JingQianCloud (approved):
Conditional approval. Dev needs it for coding purpose.
The schema will be modified late for real production.

SophieGuo410 (approved):

mikeyang923 (approved):

PR #2457, opened 2023-05-16
Don't allow user requests to delete data chunks.

ankagrawal:
Only delete requests generated by frontends should delete data chunks.


ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -432,7 +432,7 @@ public Future<Void> deleteBlob(String blobId, String serviceId, Callback<Void> c
         // Can skip attemptChunkDeletes if we can determine this is not a metadata blob
         boolean attemptChunkDeletes = isMaybeMetadataBlob(blobId);
         getOperationController().deleteBlob(blobId, serviceId, futureResult,
-            new BlobOperationCallbackWrapper<>(blobId, callback), attemptChunkDeletes, quotaChargeCallback);
+            new BlobOperationCallbackWrapper<>(blobId, callback), attemptChunkDeletes, quotaChargeCallback, true);
```
mikeyang923, line 435: So only user request will hit this place?

ankagrawal: There are only 4 ways delete requests are issued:
1. By user (that means either via curli or ambry-client).
2. By BackgroundDeleter
3. For OnDemandReplication (in future)
4. For NamedBlob cleanup

Router.deleteBlob will be called only for cases 1 & 4 and there is no reason to delete a data chunk in those cases.

ankagrawal: Maybe we can change the flag name to "isInternalRequest". We definitely know that requests generated by BackgroundDeleter and  OnDemandReplication are internal requests, while others are external/user requests.
In that case the claim would be that only internal requests should be able to delete the data chunks.

tbingol (approved):

PR #2459, opened 2023-05-17
[FULL_AUTO] Use tag to determine if a host is on FULL_AUTO or not

justinlin-linkedin:
## Summary 
This PR change the implementation of `isDataNodeInFullAutoMode`.

The issue with the current implementation is that when data node restarts, helix would remove it from all the IdealState and ExternalView so we can't figure out which resource this host's partitions belong to. In this case, whenever a data node restarts, it will be changed back out to SEMI AUTO.

The new implementation fixes this issue.

The new implementation relies on the tag rather then the IdealState. Before we turn on FULL_AUTO on any resource, we would add a tag to the resource and the same tag to all the instances that belong to this resource. The tag is not going to disappear when the instance (data node) restarts. In this way, we can rely on this tag to determine if the FULL_AUTO is turned on.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -30,7 +30,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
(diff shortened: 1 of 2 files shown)


Arun-LinkedIn (approved):

mikeyang923 (approved):
LGTM

PR #2460, opened 2023-05-17
Add NO_UPDATE NamedBlobMode and reject named blob upsert request for this mode

mikeyang923:
Description: 
Add NO_UPDATE NamedBlobMode and reject named blob upsert request when the container is NO_UPDATE mode


justinlin-linkedin (approved):
LGTM

SophieGuo410 (approved):

PR #2462, opened 2023-05-23
[vcr-2.0] Backup local-leader only for faster progress

snalli:
A flaw in replication causes some unnecessary traffic to Azure. Its not a bug or error, but a logical flaw.
1/ vcr-node fetches a list of blob-ids from all 3 replicas of a partition
2/ vcr-node filters missing blob-ids from all 3 lists by checking azure, these 3 missing-blob-lists are identical
3/ vcr-node fetches blob data from all 3 replicas for missing blobs
4/ vcr-node uploads missing blobs from all 3 replicas one replica at a time
The uploads from the first replica succeed, but the next two fail causing a blobUploadConflictCount error.
Its not due to multiple threads trying to replicate in parallel and writing to Azure. 

This happens in server-server replication also. But there it is cheap to keep checking the index or disk just before writing a blob to disk and there are locks to guard from concurrent PUT. Its not so cheap to keep doing that for Azure. VCR-1.0 attempted to solve this using a recentBlobCache but that had a few issues and corner cases.
Short term solution is to do just leader-based replication between VCR and server for a 1-2 months until the backup finishes. Backups are a bit slow right now. Long term solution is to find a cheaper way to avoid multiple uploads of the same blob.


snalli:
https://github.com/linkedin/ambry/pull/2730

PR #2463, opened 2023-05-23
Best effort to delete dataset version if named blob upload failed

SophieGuo410:
We add the dataset version before uploading the named blob in order to support LATEST version control.
If named blob upload failed(including upload blob and update named blob db), we will take the best effort to delete the specific dataset version.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
@@ -152,6 +152,7 @@ public class NamedBlobPutHandler {
     private final RestRequest restRequest;
     private final RestResponseChannel restResponseChannel;
     private final Callback<Void> finalCallback;
+    private final Callback<Void> deleteDatasetCallback;
     private final String uri;
 
     /**
@@ -164,6 +165,7 @@ public class NamedBlobPutHandler {
       this.restRequest = restRequest;
       this.restResponseChannel = restResponseChannel;
       this.finalCallback = finalCallback;
+      this.deleteDatasetCallback = deleteDatasetVersionIfUploadFailedCallBack(finalCallback);
       this.uri = restRequest.getUri();
     }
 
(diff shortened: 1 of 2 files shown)


JingQianCloud (commented):
NamedBlobPutHandler has a Callback Chain with multiple steps.
Looks like the PR is carefully picking for each step, shall we use the "finalCallback" or the "deleteDatasetCallback"?
Is there any harm to always call "deleteDatasetCallback"?

If so, we probably can only change one line
this.finalCallback = deleteDatasetVersionIfUploadFailedCallBack(finalCallback);

SophieGuo410:
> NamedBlobPutHandler has a Callback Chain with multiple steps. Looks like the PR is carefully picking for each step, shall we use the "finalCallback" or the "deleteDatasetCallback"? Is there any harm to always call "deleteDatasetCallback"?
> 
> If so, we probably can only change one line this.finalCallback = deleteDatasetVersionIfUploadFailedCallBack(finalCallback);

In NamedBlobPutHandler, we might call finalCallback for both successAction and failureCallback, you can refer details in buildCallback. So I just replace the finalCallback with deleteDatasetCallback when named blob upload or router.putBlob related function failed.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -321,7 +323,7 @@ private Callback<Void> routerTtlUpdateCallback(BlobInfo blobInfo, String blobId)
           deleteDatasetVersionOutOfRetentionCount();
         }
         securityService.processResponse(restRequest, restResponseChannel, blobInfo, securityProcessResponseCallback());
-      }, uri, LOGGER, finalCallback);
+      }, uri, LOGGER, deleteDatasetCallback);
     }
```
JingQianCloud, line 327: For the last step, securityProcessResponseCallback, shall we change the failure callback to deleteDatasetCallback as well?

I feel for any buildCallback after the PutBlob step, we should replace the failure callback to deleteDatasetCallback.

```   
private Callback<Void> securityProcessResponseCallback() {
      return buildCallback(frontendMetrics.putBlobSecurityProcessResponseMetrics,
          securityCheckResult -> finalCallback.onCompletion(null, null), restRequest.getUri(), LOGGER, finalCallback);
```

SophieGuo410: This step is to sumbit the response. the named blob and regular blob already in our service. So I feel like no need to rollback if only the sumbit response failed. Instead, client will take the best effort to manually delete this version.

JingQianCloud (approved):

justinlin-linkedin (approved):

PR #2464, opened 2023-05-24
[repl] Do not advance token when there are invalid messages

snalli:
In the existing replication logic, when a CRC mismatch occurs, the system simply bypasses those keys and moves forward with the token. Consequently, we end up overlooking those specific keys, and there is no opportunity to revisit them later.

For instance, we encountered a bug that caused us to unintentionally skip the backup of 1090 blobs as a direct consequence of this issue.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -1305,10 +1305,9 @@ public class ReplicaThread implements Runnable {
                     messageInfoList);
               }
 
-              remoteReplicaInfo.setToken(exchangeMetadataResponse.remoteToken);
-              remoteReplicaInfo.setLocalLagFromRemoteInBytes(exchangeMetadataResponse.localLagFromRemoteInBytes);
-              // reset stored metadata response for this replica
-              remoteReplicaInfo.setExchangeMetadataResponse(new ExchangeMetadataResponse(ServerErrorCode.No_Error));
+              if (!validMessageDetectionInputStream.hasInvalidMessages()) {
+                advanceToken(remoteReplicaInfo, exchangeMetadataResponse);
+              }
 
               logger.trace("Remote node: {} Thread name: {} Remote replica: {} Token after speaking to remote node: {}",
                   remoteNode, threadName, remoteReplicaInfo.getReplicaId(), exchangeMetadataResponse.remoteToken);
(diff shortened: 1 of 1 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1305,13 +1305,12 @@ private void handleGetResponse(GetResponse getResponse, List<RemoteReplicaInfo>
                     messageInfoList);
               }
 
-              remoteReplicaInfo.setToken(exchangeMetadataResponse.remoteToken);
-              remoteReplicaInfo.setLocalLagFromRemoteInBytes(exchangeMetadataResponse.localLagFromRemoteInBytes);
-              // reset stored metadata response for this replica
-              remoteReplicaInfo.setExchangeMetadataResponse(new ExchangeMetadataResponse(ServerErrorCode.No_Error));
+              if (!validMessageDetectionInputStream.hasInvalidMessages()) {
+                advanceToken(remoteReplicaInfo, exchangeMetadataResponse);
+              }
```
justinlin-linkedin, line 1310: There are two ways to do this.
1. Treat MessageFormatException as any other exception and just throw it out in MessageSievingInputStream constructor. This way, whenever replica thread encountered a crc error, it would throw an exception right away and restart the replication.
2. We still treat MessageFormatException differently like this PR. We would deal with other blobs that don't have any error but not advance token so next time we would fetching the same blobs.

Both approaches will perform retry on the MessageFormatException blobs. 

However the second approach would deal with non exception blobs in the first time and later it would fetch these blobs again, which seems to be a better solution.

But both approaches do not solve the infinite retry issue. If a blob is indeed corrupted, then we would fall into infinite retries for this blob. 

We should probably create some alerts for the message format exception if we want to do infinite retry.

justinlin-linkedin (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
JingQianCloud: Sanketh, thanks for reporting this issue. 
Can you help check "replicationMetrics.incrementInvalidMessageError"? We have metrics to record the invalid message. How good or bad it is right now?

Advancing token makes us lose track of the invalid message which is not acceptable.
But if we don't advance it and we'll start and retry from this token every time when the server boots up. If we don't take immediate action on that, we may have bigger and bigger replication lag.

JingQianCloud (approved):
Based on the discussion, looks like the problem happened on VCR only. Approved.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1305,13 +1305,12 @@ private void handleGetResponse(GetResponse getResponse, List<RemoteReplicaInfo>
                     messageInfoList);
               }
 
-              remoteReplicaInfo.setToken(exchangeMetadataResponse.remoteToken);
-              remoteReplicaInfo.setLocalLagFromRemoteInBytes(exchangeMetadataResponse.localLagFromRemoteInBytes);
-              // reset stored metadata response for this replica
-              remoteReplicaInfo.setExchangeMetadataResponse(new ExchangeMetadataResponse(ServerErrorCode.No_Error));
+              if (!validMessageDetectionInputStream.hasInvalidMessages()) {
+                advanceToken(remoteReplicaInfo, exchangeMetadataResponse);
```
tbingol, line 1309: What happens when we do not advance the token? I think we should either this failure visible and continue with the replication or halt the replication until we get to the bottom of the issue.

snalli: @tbingol If we do not advance the token, then the target node (vcr) will keep fetching blobs from previous token position. And if a blob is truly corrupted, then we will never make progress be stuck in an infinite loop. Currently, there isn't a finer granularity of insight of whether a blob is truly corrupted or just the message coming from source is corrupted. Nor do we have any alerts for infinite retries. One key point of difference between server and vcr is that on server side, we emit metrics for blobs skipped due to crc error or message format exceptions. But we do not emit any metrics on vcr side, so we have no idea how many blobs we skipped. I recently added those metrics though it has not been deployed yet.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
JingQianCloud: My questions and concern is around Ambry Server, not VCR since the code is shared for both Ambry server and VCR.
But I think I have the answer. For Ambry Server, we didn't hit the crc error issues. So the metrics should show zero. So Ambry Server won't have larger replication lag due to the change.

snalli:
some tests are failing because they expect the token to be advanced inspite of invalid messages, so need to modify those test. But this fundamentally changes the behaviour of replication.

PR #2465, opened 2023-05-25
Handle a cycle of standby to standby state transitions from Helix during semi-auto to full-auto migration

Arun-LinkedIn:
When migrating from semi-auto to full-auto, we observed that helix could bring down and bring up replicas on the same host. I.e. they send state transitions of STANDBY -> INACTIVE -> OFFLINE -> BOOTSTRAP -> STANDBY.

This change avoids stopping the `BlobStore` during INACTIVE -> OFFLINE transition. This is consistent with `BlobStore` state on boot up where the replica state is OFFLINE.


justinlin-linkedin (approved):
LGTM

SophieGuo410 (approved):

PR #2468, opened 2023-05-26
Schedule the named blob cleanup process to run once with delay

mikeyang923:
Schedule the named blob cleanup process to run once with delay

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
@@ -39,6 +39,7 @@ import com.github.ambry.utils.ThrowingConsumer;
 import com.github.ambry.utils.Utils;
 import java.io.IOException;
 import java.util.GregorianCalendar;
+import java.util.Random;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -209,7 +211,7 @@ public void start() throws InstantiationException {
     if (frontendConfig.enableNamedBlobCleanupTask) {
       namedBlobsCleanupScheduler = Utils.newScheduler(1, "named-blobs-cleanup-", false);
       namedBlobsCleanupTask =
-          namedBlobsCleanupScheduler.scheduleAtFixedRate(namedBlobsCleanupRunner, 60 * 10, 60 * 10, TimeUnit.SECONDS);
+          namedBlobsCleanupScheduler.schedule(namedBlobsCleanupRunner, random.nextInt(60*60*24*7), TimeUnit.SECONDS);
```
justinlin-linkedin, line 214: we still want to run this periodically, we want to run this with a delay and then periodically run it.

justinlin-linkedin (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
mikeyang923: Ok, got it, since I thought we had weekly release, we don't need to schedule it to run weekly. But I agree it's better to repeat weekly in case the ambry release delayed or canceled..

PR #2470, opened 2023-05-31
Add a network client implementation to communicate with cloud

justinlin-linkedin:
## Summary

Adding a NetworkClient implementation to return responses by talking to cloud. This would be useful in RecoveryManager to recover data from cloud, with a nonblocking network client. However, in this implementation, the client is a blocking client. But it's fine to be in this way. If later we can have a real nonblocking implementation, we can later replace it.


snalli:
Testing this out, will get back on Monday

snalli (approved):

PR #2476, opened 2023-06-14
Support list all datasets for a container

SophieGuo410:


ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -1133,6 +1162,41 @@ private List<DatasetVersionRecord> executeListAllValidVersionStatement(PreparedS
...
+   */
+  private Page<String> executeListAllValidDatasetsStatement(PreparedStatement statement, int accountId,
+      int containerId, String pageToken) throws SQLException {
+    ResultSet resultSet = null;
+    try {
+      statement.setInt(1, accountId);
+      statement.setInt(2, containerId);
+      statement.setString(3, pageToken);
+      statement.setString(4, pageToken);
+      statement.setInt(5, mySqlAccountServiceConfig.listDatasetsMaxResult + 1);
```
JingQianCloud, line 1182: I guess the way to implement the paging is that,
We get one entry more than needed. And then we use the last entry as the starting point of the next page. Very nice approach.

Another approach is that we get the exact entries as we needed. And then for the next query we use > instead of >=.

Comparing with these two, I believe we have a reason to choice the first one. May I know the reason?

JingQianCloud (approved):
Approved.
But I have a question embedded.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
SophieGuo410: As I checked, typically we will use the next_page_token to fetch the next page, also followed by named blob as well.

PR #2477, opened 2023-06-15
Add Log4J structured logging for public access logs

litingulfs:
This PR changes the public access logging so that instead of writing out a single large string for each log statement, structured logs are emitted. The advantage of this is that our logs will be much easier to search with Azure Kusto, since the LinkedIn Kusto integration breaks up the structured log fields into separate fields in Kusto. This way, we won't have to write regular expressions to parse out individual fields in Kusto.

The changes are guarded by a new configuration that is set to false by default.

The code is a little messy, essentially copying each log instruction to a new, structured log, instruction. I left it like this on purpose, since we don't really have any way to properly test these changes other than to deploy to a test host and check Kusto afterwards.

If we get this working, I intend to follow up with another PR that either removes the old logging or cleans up this code so the new and old ways can exist side by side.

### DIFF
--- a/ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
+++ b/ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
@@ -28,8 +28,9 @@ import java.security.cert.X509Certificate;
 import java.util.Collection;
 import javax.net.ssl.SSLEngine;
 import javax.security.auth.x500.X500Principal;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.message.StringMapMessage;
 
 
 /**
@@ -44,11 +45,13 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
   private long responseFirstChunkStartTimeInMs;
   private long responseBytesSent;
   private StringBuilder logMessage;
+  private StringMapMessage structuredLogMessage;
   private HttpRequest request;
   private StringBuilder sslLogMessage;
 
   private static final long INIT_TIME = -1;
-  private static final Logger logger = LoggerFactory.getLogger(PublicAccessLogHandler.class);
+  //private static final Logger logger = LoggerFactory.getLogger(PublicAccessLogHandler.class);
+  private static final Logger logger = LogManager.getLogger(PublicAccessLogHandler.class);
 
   public PublicAccessLogHandler(PublicAccessLogger publicAccessLogger, NettyMetrics nettyMetrics) {
     this.publicAccessLogger = publicAccessLogger;
@@ -71,8 +74,11 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
       requestArrivalTimeInMs = System.currentTimeMillis();
       request = (HttpRequest) obj;
       logMessage.append(ctx.channel().remoteAddress()).append(" ");
+      structuredLogMessage.put("remoteAddress", String.valueOf(ctx.channel().remoteAddress()));
       logMessage.append(request.method().toString()).append(" ");
+      structuredLogMessage.put("method", String.valueOf(request.method()));
       logMessage.append(request.uri()).append(", ");
+      structuredLogMessage.put("uri", String.valueOf(request.uri()));
       logSSLInfo(ctx);
       logMessage.append(", ");
       logHeaders("Request", request, publicAccessLogger.getRequestHeaders());
@@ -100,6 +106,7 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
         logMessage.append(", ");
         logMessage.append("status=").append(response.status().code());
         logMessage.append(", ");
+        structuredLogMessage.put("status", String.valueOf(response.status().code()));
         if (HttpUtil.isTransferEncodingChunked(response)) {
           responseFirstChunkStartTimeInMs = System.currentTimeMillis();
         } else {
@@ -116,7 +123,14 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
       }
       if (shouldReset) {
         logDurations();
-        publicAccessLogger.logInfo(logMessage.toString());
+
+        if (publicAccessLogger.structuredLoggingEnabled()) {
+          // Write out the old log message as a full string, as well as add each field individually.
+          structuredLogMessage.put("message", logMessage.toString());
+          publicAccessLogger.logInfo(structuredLogMessage);
+        } else {
+          publicAccessLogger.logInfo(logMessage.toString());
+        }
         reset();
       }
     }
@@ -151,11 +165,13 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
     for (String header : headers) {
       if (message.headers().get(header) != null) {
         logMessage.append("[").append(header).append("=").append(message.headers().get(header)).append("] ");
+        structuredLogMessage.put(header, message.headers().get(header));
       }
     }
     boolean isChunked = HttpUtil.isTransferEncodingChunked(message);
     logMessage.append("[isChunked=").append(isChunked).append("]");
     logMessage.append(")");
+    structuredLogMessage.put("isChunked", String.valueOf(isChunked));
   }
 
   /**
@@ -165,14 +181,20 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
   private void logDurations() {
     long nowMs = System.currentTimeMillis();
     logMessage.append("bytes sent=").append(responseBytesSent).append(" ");
+    structuredLogMessage.put("bytesSent", String.valueOf(responseBytesSent));
     logMessage.append("duration=").append(nowMs - requestArrivalTimeInMs).append("ms ");
+    structuredLogMessage.put("duration", String.valueOf(nowMs - requestArrivalTimeInMs).concat("ms"));
     if (requestLastChunkArrivalTimeInMs != INIT_TIME) {
       logMessage.append("(chunked request receive=")
           .append(requestLastChunkArrivalTimeInMs - requestArrivalTimeInMs)
           .append("ms) ");
+      structuredLogMessage.put("chunkedRequestReceive",
+          String.valueOf(requestLastChunkArrivalTimeInMs - requestArrivalTimeInMs).concat("ms"));
     }
     if (responseFirstChunkStartTimeInMs != INIT_TIME) {
       logMessage.append("(chunked response send=").append(nowMs - responseFirstChunkStartTimeInMs).append("ms) ");
+      structuredLogMessage.put("chunkedResponseSend",
+          String.valueOf(nowMs - responseFirstChunkStartTimeInMs).concat("ms"));
     }
   }
 
@@ -185,6 +207,7 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
     requestArrivalTimeInMs = INIT_TIME;
     responseBytesSent = 0;
     logMessage = new StringBuilder();
+    structuredLogMessage = new StringMapMessage();
     request = null;
   }
 
@@ -195,7 +218,13 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
   private void logError(String msg) {
     logDurations();
     logMessage.append(msg);
-    publicAccessLogger.logError(logMessage.toString());
+    structuredLogMessage.put("error", msg);
+
+    if (publicAccessLogger.structuredLoggingEnabled()) {
+      publicAccessLogger.logError(structuredLogMessage);
+    } else {
+      publicAccessLogger.logError(logMessage.toString());
+    }
   }
 
   /**
@@ -210,6 +239,8 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
         SslHandler sslHandler = ctx.pipeline().get(SslHandler.class);
         boolean sslUsed = sslHandler != null;
         sslLogMessage.append("[used=").append(sslUsed).append("]");
+        structuredLogMessage.put("used", String.valueOf(sslUsed));
+
         if (sslUsed) {
           SSLEngine sslEngine = sslHandler.engine();
           if (sslEngine.getNeedClientAuth()) {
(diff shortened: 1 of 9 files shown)


ankagrawal (commented):
Why are these new files and not part of their packages?

litingulfs:
Sorry @ankagrawal - I didn't realize we usually use forks of the repo to make changes and then PRs from there. I ended up copying over files to a new fork, not setting the paths correctly. Should be fixed now.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -28,8 +28,12 @@
 import java.util.Collection;
 import javax.net.ssl.SSLEngine;
 import javax.security.auth.x500.X500Principal;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+//import org.slf4j.Logger;
```
ankagrawal, line 31: nit: lets remove these two lines instead of commenting them out.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -28,8 +28,12 @@
...
+//import org.slf4j.LoggerFactory;
+import org.apache.logging.log4j.Level;
```
ankagrawal, line 33: Is this import unused?

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -44,11 +48,13 @@ public class PublicAccessLogHandler extends ChannelDuplexHandler {
   private long responseFirstChunkStartTimeInMs;
   private long responseBytesSent;
   private StringBuilder logMessage;
+  private StringMapMessage structuredLogMessage;
   private HttpRequest request;
   private StringBuilder sslLogMessage;
 
   private static final long INIT_TIME = -1;
-  private static final Logger logger = LoggerFactory.getLogger(PublicAccessLogHandler.class);
+  //private static final Logger logger = LoggerFactory.getLogger(PublicAccessLogHandler.class);
```
ankagrawal, line 56: instead of commenting out, please remove lines that you don't need here and at other places.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -116,7 +123,11 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
       }
       if (shouldReset) {
         logDurations();
-        publicAccessLogger.logInfo(logMessage.toString());
+
+        if(publicAccessLogger.structuredLoggingEnabled()) {
+          publicAccessLogger.logInfo(structuredLogMessage);
+        }
+        else publicAccessLogger.logInfo(logMessage.toString());
```
ankagrawal, line 130: nit: formatting issue. This should be

```
} else {
    publicAccessLogger.logInfo(logMessage.toString());
}
```

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -165,16 +178,24 @@ private void logHeaders(String tag, HttpMessage message, String[] headers) {
...
     long nowMs = System.currentTimeMillis();
     logMessage.append("bytes sent=").append(responseBytesSent).append(" ");
+    structuredLogMessage.put("bytesSent", String.valueOf(responseBytesSent));
     logMessage.append("duration=").append(nowMs - requestArrivalTimeInMs).append("ms ");
+    structuredLogMessage.put("duration", String.valueOf(nowMs - requestArrivalTimeInMs).concat("ms"));
     if (requestLastChunkArrivalTimeInMs != INIT_TIME) {
       logMessage.append("(chunked request receive=")
           .append(requestLastChunkArrivalTimeInMs - requestArrivalTimeInMs)
           .append("ms) ");
+      structuredLogMessage.put("chunkedRequestReceive",
```
ankagrawal, line 188: Here and at other places, do you think there's a value in keeping the error message same as before?

Like in this case keeping the message as "(chunked request receive="?

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -165,16 +178,24 @@ private void logHeaders(String tag, HttpMessage message, String[] headers) {
...
+          .concat("ms"));
     }
     if (responseFirstChunkStartTimeInMs != INIT_TIME) {
       logMessage.append("(chunked response send=").append(nowMs - responseFirstChunkStartTimeInMs).append("ms) ");
+      structuredLogMessage.put("chunkedResponseSend", String.valueOf(nowMs - responseFirstChunkStartTimeInMs)
+              .concat("ms"));
+
     }
-  }
+   }
```
ankagrawal, line 198: nit: formatting.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -195,7 +217,14 @@ private void reset() {
   private void logError(String msg) {
     logDurations();
     logMessage.append(msg);
-    publicAccessLogger.logError(logMessage.toString());
+    structuredLogMessage.put("error", msg);
+
+    if(publicAccessLogger.structuredLoggingEnabled()) {
+      publicAccessLogger.logError(structuredLogMessage);
+    }
+    else publicAccessLogger.logError(logMessage.toString());
```
ankagrawal, line 225: nit: formatting

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -218,7 +249,9 @@ private void logSSLInfo(ChannelHandlerContext ctx) {
                 X500Principal principal = ((X509Certificate) certificate).getSubjectX500Principal();
                 Collection subjectAlternativeNames = ((X509Certificate) certificate).getSubjectAlternativeNames();
                 sslLogMessage.append(", [principal=").append(principal).append("]");
+                structuredLogMessage.put("principal", principal.toString());
```
ankagrawal, line 252: Do we know why we had separate `sslLogMessage` and `logMessage` earlier. Should we also separate out `structuredLogMessage` for ssl and non ssl cases.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogger.java
```diff
@@ -13,27 +13,33 @@
...
+  private static final Logger publicAccessLogger = LogManager.getLogger("PublicAccessLogger");
 
   private final String[] requestHeaders;
   private final String[] responseHeaders;
+  private boolean enableStructuredLogging;
 
   /**
    * @param requestHeaders the request headers to log.
    * @param responseHeaders the response headers to log.
+   * @param enableStructuredLogging true if we want to use structured logging (for Kusto), false otherwise
```
ankagrawal, line 37: nit: javadocs formatting
Should be
```
@param enableStructuredLogging {@code true} if we want to use structured logging (for Kusto), {@code false} otherwise
```

ambry-rest/src/test/java/com/github/ambry/rest/NettyServerTest.java
```diff
@@ -38,7 +38,7 @@ public class NettyServerTest {
   private static final RestRequestService REST_REQUEST_SERVICE =
       new MockRestRequestService(new VerifiableProperties(new Properties()), new MockRouter());
   private static final RestRequestHandler REQUEST_HANDLER = new MockRestRequestResponseHandler(REST_REQUEST_SERVICE);
-  private static final PublicAccessLogger PUBLIC_ACCESS_LOGGER = new PublicAccessLogger(new String[]{}, new String[]{});
+  private static final PublicAccessLogger PUBLIC_ACCESS_LOGGER = new PublicAccessLogger(new String[]{}, new String[]{}, false);
```
ankagrawal, line 41: Maybe let's add a couple of tests where `enableStructuredLogging` is set to true. Doesn't really test the log message itself, but at least we will have the confidence that the code path will not just error out, and will also be maintained in the future.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
```diff
@@ -100,6 +106,7 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
         logMessage.append(", ");
         logMessage.append("status=").append(response.status().code());
         logMessage.append(", ");
+        structuredLogMessage.put("status", String.valueOf(response.status().code()));
```
Arun-LinkedIn, line 109: We also log information related to received request such as remote address, headers, etc in `channelRead()` method. May be we would need those information too in structured log message?

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
Arun-LinkedIn: Yeah. You could probably use '⌘⌥L' option to format file

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogger.java
```diff
@@ -13,27 +13,33 @@
...
+import org.apache.logging.log4j.message.StringMapMessage;
 
 /**
  * Logs requests and responses to public access log.
  */
 public class PublicAccessLogger {
 
-  private static final Logger publicAccessLogger = LoggerFactory.getLogger("PublicAccessLogger");
+  //private static final Logger publicAccessLogger = LoggerFactory.getLogger("PublicAccessLogger");
+  private static final Logger publicAccessLogger = LogManager.getLogger("PublicAccessLogger");
```
Arun-LinkedIn, line 28: Just wanted to note that this seems to directly use the Log4j APIs instead of SLF4J (which seems to be an [abstraction](https://howtodoinjava.com/log4j2/slf4j-vs-log4j-which-one-is-better/) over Log4j and can be useful for replacing with other logging frameworks). For example, we provide the log4j-slf4j-impl in the dependencies [here](https://github.com/linkedin/ambry/blob/ca77dc6bec8a84c3f2200295eba6688422f2adcf/build.gradle#L430).

But, I couldn't find on how to use structured logging (StringMapMessage) with SLF4J. May be it is okay for now?

ambry-rest/src/test/java/com/github/ambry/rest/NettyServerTest.java
litingulfs: Yeah, I modified FrontendNettyFactoryTest to include a couple of cases where where `enableStructuredLogging` is set to true. Let me know if you think that's fine. If not, I'll need some help to add tests from scratch.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: Thanks, good catch, I added those.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogger.java
litingulfs: According to the main Wiki page that I've used for guidance at https://iwww.corp.linkedin.com/wiki/cf/pages/viewpage.action?spaceKey=ENGS&title=Structural+Logging+with+Log4J we're not yet able to use SLF4J for structured logging at LinkedIn:

"Specific for structural logging, it is not supported until SLF4j version 2.x, which is not allowed yet in LinkedIn due to the compatibility issues. Therefore we now recommend Log4j API for structural logging. SLF4j support will be added later on."

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogger.java
litingulfs: Thanks, I did not know this.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: I'm not quite sure what the original reasoning here was. However, if I understand the original code correctly, it simply uses `sslLogMessage` to build up an SSL-specific log statement which is then appended to `logMessage` at the end of the method. So `sslLogMessage` is not really a log message by itself, it seems to exist for first-pass convenience.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: Thanks for pointing this out! I changed the code such that when we actually print the log statements, we include the full contents of the old log message as a "message" field. This will certainly bloat the logs a bit, but for the time being should make it easier to read the logs for humans.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: I ran the IntelliJ command suggested by Arun on this and the other changed files.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: That is done.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: Yes, I deleted it.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: OK that is done.

ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
litingulfs: Should be fixed now.

PR #2478, opened 2023-06-18
[Efficient Metadata Operations] Set the blob data type of each chunk as SIMPLE, DATACHUNK or METADATACHUNK.

ankagrawal:
Before this change, each blob was either set as a DATACHUNK or as a METADATACHUNK.

There are a few cases to consider while determining if a chunk is a simple chunk or a data chunk
1. If the user uploads data as part of chunked upload, then the chunk is always a DATACHUNK.
2. For data that is not uploaded as chunked upload: 2.1 If the user uploads less than RouterConfig.routerMaxPutChunkSizeBytes of data for blob, then the blob is a SIMPLE CHUNK. 2.2 If the user uploads RouterConfig.routerMaxPutChunkSizeBytes or more of data, then when the first chunk is full, we don't send the chunk to server immediately. We wait until we get atleast one one more byte for upload (in this case the blob is a composite blob and the first chunk is a DATACHUNK), or until chunk filling is complete without uploading any more data (in this case the blob is a simple blob of size RouterConfig.routerMaxPutChunkSizeBytes and is marked as SIMPLE CHUNK).


ambry-router/src/test/java/com/github/ambry/router/GetManagerTest.java
```diff
@@ -282,12 +282,16 @@ private String createStitchBlob(int blobSize) throws Exception {
     int curBlobSize = blobSize;
     for (int i = 0; i < NUM_STITCHED_CHUNKS; i++) {
       //Give each chunk a different size
-      int curChunkSize = Math.min(curBlobSize, chunkSize + i * 5);
+      int curChunkSize = Math.min(curBlobSize, chunkSize - i * 5);
+      if (i == NUM_STITCHED_CHUNKS - 1) {
+        curChunkSize = curBlobSize;
+      }
```
mikeyang923, line 288: Why we change logic here?

JingQianCloud:
> 
What does "uploads data as part of chunked upload" mean? Using what kind of commands?

I thought, in item 2 seems to me, it's our existing design or at least planned logic, but you mentioned so far we only set DATACHUNK or METADATACHUNK? Do you mean our existing code has some bugs so we didn't follow the planned way?

ankagrawal:
> What does "uploads data as part of chunked upload" mean? Using what kind of commands?

This will be as preparation of stitched blob API. Users can upload blob using signed-url by setting the header "chunked_upload=true". Hence such uploads are called "chunked uploads", as opposed to regular uploads where the users upload the entire blob.

> Item 2 seems to me, it's our existing design or at least our planned logic, but you mentioned so far we only set DATACHUNK or METADATACHUNK. Do you mean our existing code has some bugs so we didn't follow the planned.

Our existing code only marks the blobs as `BlobDataType::DATACHUNK` or `BlobDataType::METADATACHUNK`. It doesn't mark a blob as `BlobDataType::SIMPLE` blob. Blobs with `size <=RouterConfig::routerMaxPutChunkSizeBytes` are the ones that should be marked as `BlobDataType::SIMPLE` blob. Thats what this PR fixes.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1428,8 +1469,15 @@ int fillFrom(ByteBuf channelReadBuf) {
         }
       }
       if (buf.readableBytes() == routerConfig.routerMaxPutChunkSizeBytes) {
-        onFillComplete(true);
-        updateChunkFillerWaitTimeMetrics();
+        if (firstPutChunkPrepared) {
```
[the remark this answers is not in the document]
JingQianCloud: Just wondering if "delay write" can make the code logic a little bit easier.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -647,23 +654,35 @@ void fillChunks() {
...
         PutChunk lastChunk = getBuildingChunk();
-        if (lastChunk != null) {
-          if (chunkCounter != 0 && lastChunk.buf.readableBytes() == 0) {
-            logger.trace("{}: The last buffer(s) received from chunkFillerChannel have no data, discarding them.",
-                loggingContext);
-            lastChunk.releaseBlobContent();
-            lastChunk.state = ChunkState.Free;
-          } else {
+
+        if (!firstPutChunkPrepared && (chunkCounter == 0 || (chunkCounter == 1
```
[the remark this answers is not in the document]
ankagrawal: So, the crux of the logic is that once we have the first PutChunk object filled, then we wait for either `chunkFillingCompletedSuccessfully` to be true, or wait for reading at least one more byte on the network.

"firstPutChunkAwaitingResolution" is to indicate that the first put chunk is filled up, and is `waiting for either "chunkFillingCompletedSuccessfully" to be true, or wait for reading at least one more byte on the network`

"firstPutChunkPrepared" is to indicate that the first put chunk is ready to be sent to the server and the chunk filling logic doesn't need to worry about it.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1428,8 +1469,15 @@ int fillFrom(ByteBuf channelReadBuf) {
...
-        updateChunkFillerWaitTimeMetrics();
+        if (firstPutChunkPrepared) {
+          onFillComplete(true);
+        } else {
+          // If first put chunk is full, but not yet prepared then mark it pending resolution instead of completing it.
+          // This chunk will be prepared as soon as either more bytes are read from the channel, or the chunk filling
+          // is complete. At that point we will have enough information to mark this blob as simple or composite blob.
+          chunkToFill = null; // this will ensure that a new chunk is picked up for filling more data.
+          firstPutChunkAwaitingResolution = true;
+        }
```
justinlin-linkedin, line 1480: let's make it easier here, if we know the current chunk is the first chunk then we can verify if we have extra byte in the `channelReadBuf` here right away.
```
if (buf.readableBytes() == routerConfig.routerMaxPutChunkSizeBytes) {
    if (this == firstChunk) {
        if (channelReadBuf.readableBytes() > 0) {
            prepareFirstChunk(false);
        } else {
            firstPutChunkAwaitingResolution = true;
        }
    } else {
        onFillComplete(true);
        updateChunkFillerWaitTimeMetrics();
    }
}
```

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1428,8 +1469,15 @@ int fillFrom(ByteBuf channelReadBuf) {
...
       if (buf.readableBytes() == routerConfig.routerMaxPutChunkSizeBytes) {
-        onFillComplete(true);
-        updateChunkFillerWaitTimeMetrics();
+        if (firstPutChunkPrepared) {
+          onFillComplete(true);
+        } else {
+          // If first put chunk is full, but not yet prepared then mark it pending resolution instead of completing it.
+          // This chunk will be prepared as soon as either more bytes are read from the channel, or the chunk filling
+          // is complete. At that point we will have enough information to mark this blob as simple or composite blob.
+          chunkToFill = null; // this will ensure that a new chunk is picked up for filling more data.
```
justinlin-linkedin, line 1478: we probably don't ned to assign null to chunkToFill. When there are extra bytes from http channel, before we read any of bytes, we would already prepare first chunk to send it out. After that, the state of first chunk would be either encrypting or ready.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -847,11 +886,14 @@ private PutChunk getFreeChunk() {
   private PutChunk getBuildingChunk() {
     PutChunk chunkToReturn = null;
     for (PutChunk chunk : putChunks) {
-      if (chunk.isBuilding()) {
+      if (chunk.isBuilding() && (firstPutChunkPrepared || chunk != firstChunk)) {
```
justinlin-linkedin, line 889: I don't think we want to treat first chunk differently here with building. If the first chunk is currently building, let's just return it.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -180,6 +180,10 @@ class PutOperation {
   private final String loggingContext;
   private final CompressionService compressionService;
   private final ReservedMetadataIdMetrics reservedMetadataIdMetrics;
+  private PutChunk firstChunk = null;
+  private boolean firstPutChunkPrepared = false; // to track if the first chunk has been prepared for sending.
```
justinlin-linkedin, line 184: I suspect that we don't need `firstPutChunkPrepared`, but only `firstPutChunkAwaitingResolution` to determine if this is a simple blob.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -617,6 +620,10 @@ void fillChunks() {
           channelReadBuf = chunkFillerChannel.getNextByteBuf(0);
         }
         if (channelReadBuf != null) {
+          // If first chunk is awaiting resolution, and there are more bytes in the channel prepare it first as a data chunk.
+          if (channelReadBuf.readableBytes() > 0 && firstPutChunkAwaitingResolution && !firstPutChunkPrepared) {
+            prepareFirstChunk(false);
+          }
```
justinlin-linkedin, line 626: we can do 
```
if (firstChunk == chunkToFill && firstPutChunkAwaitForResolution) {
    prepareFirstChunk(false);
}
```
In this case, if the current chunk is still the first chunk, meaning first chunk is not in encryption or network, and the first chunk is waiting for resolution, we now know that the resolution is here.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
justinlin-linkedin, line 661: I strongly believe that we don't need two boolean variables

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: I want to reduce this to only one variable/state too. I tried to think about that but wasn't able to find a reasonable solution.

Here are two problems that make 2 variables necessary in my opinion
1. The `firstChunk` is a reference to `PutChunk` object. The `PutChunk` objects are taken from the array whose size is `RouterConfig::routerMaxInMemPutChunks`. So the chunk number 5/9/13 will end up sharing the same `PutChunk` object. Thats why I think two variables are needed. `firstPutChunkAwaitingResolution` to assert that first put chunk needs to be handled and hasn't been handled yet, and `firstPutChunkPrepared ` to indicate that we don't need to worry about `firstChunk` anymore.
2. Note that `RouterConfig::routerMaxInMemPutChunks` could be just 1 or 2. In that case every chunk, or every other chunk, will basically end up being equal to `firstChunk`. So we need a way to say that `firstPutChunkNeedsResolution` (i.e, that firstChunk is read from network but we are waiting for more data to ascertain fistChunk's type) and then to say that  `firstPutChunkIsPrepared ` (i.e, first chunk is prepared, and we don't need to worry about delaying it anymore).

ambry-router/src/test/java/com/github/ambry/router/GetManagerTest.java
ankagrawal: Because before this change, both simple blobs and chunked uploads were marked as `DATACHUNKS`. So it was totally possible to stitch `SIMPLE` blobs. Going forward, `SIMPLE` blobs will not be allowed for stitching. So, in order to upload blobs that need to be stitched later (like this test does), we need to upload it with option `chunkUpload=true`. 
Now once the blob is marked as chunked upload, then we also check and ensure that the max upload size doesn't exceed `CHUNK_SIZE`. Hence it's important to calculate the chunk size correctly.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
JingQianCloud: I'm trying to think about a way to solve the problem with less variables and more centralized code.

For variable, 
```
we need one in PutOperation as pointer to the firstChunk. Let's call it firstChunk.
PutChunk has a variable isSimple, by default it's always false unless we change it.
```

For functions, we need modify two functions,
I guess the prepareForSending is the function we set buffer to be ready to send out to server. 
We delay the first chunk's prepareForSending until either the second chunk is ready or the chunk filling is completed.
```
private void prepareForSending {
	if (!chunkFillingCompletedSuccessfully) {
		if (chunkIndex == 0) { <--- For first chunk, just save it and return immediately.
			firstChunk = this;
			return;
		} else if (chunkIndex == 1) { <--- When the second chunk comes, complete the first chunk.
			firstChunk.prepareForSending();
			firstChunk = null;
		}
	}

	// original code as before <-- Our original code.
	...
}
```

In fillChunks, we changed two places below with <---
```
void fillChunks() {
	...
  if (chunkFillingCompletedSuccessfully) {
    PutChunk lastChunk = getBuildingChunk();
    if (lastChunk != null) {
      if (chunkCounter != 0 && lastChunk.buf.readableBytes() == 0) {
        logger.trace("{}: The last buffer(s) received from chunkFillerChannel have no data, discarding them.",
            loggingContext);
        lastChunk.releaseBlobContent();
        lastChunk.state = ChunkState.Free;
      } else {
      	lastChunk.isSimple = true; <------ Set Simple chunk to true.
        lastChunk.onFillComplete(true);
        updateChunkFillerWaitTimeMetrics();
      }
      if (lastChunk.isReady()) {
        routerCallback.onPollReady();
      }
    } else {
    	if (firstChunk != null) {
    		firstChunk.isSimple = true;
    		firstChunk.prepareForSending(); <----- Closeup the 4MB chunk.
    	}
    }
  }
}
```

JingQianCloud (commented):
I'm trying a different way for the PR.
Please check if it's doable.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
ankagrawal: There are a few things that I want to note here
1. It's not a chunk's state that is simple or composite. It's the blob's state. For chunked uploads, each chunk is uploaded separately and gets handled like a separate put request. So if it's a chunked upload, the upload is always less than or equal to the chunk size (otherwise the upload fails). 
2. Second issue is that it's possible that `prepareForSending` was called with `chunkIndex =0`, and another chunk with `chunkIndex == 1` was created, but there was no data to fill in this chunk. In that case `prepareForSending` will not be called with `chunkIndex == 1` and `chunkFillingCompletedSuccessfully` will be set to `true`. This case needs to be handled in the `if(chunkFillingCompletedSuccessfully` section above. If we want to handle it here, then we will have to first check if `firstChunk` is still waiting (i.e., we will need at least one more variable which is `firstChunkAwaitingResolution`. 


If we want to rely on chunkCounter or chunkIndex, then I think there is another approach that can work with using `chunkCounter` instead of `chunkIndex` and checking the size of the buffer if `chunkCounter == 1`. Let me try to update the PR with that.

JingQianCloud: 1. Understood. That's why I  set isSimple to true only if it's not chunked Upload as below.
  ```
  lastChunk.isSimple = true if it's not chunkUpload; 
  ```
2. In the above code, I added the condition under chunkFillingCompletedSuccessfully to handle that case.

ankagrawal: But `isSimple` is not a property of a chunk. Every is chunk is always "simple" because its always <=4Mib. `isSimple` is the property of a user blob.

ankagrawal: Also there is one more issue why I don't prefer doing `firstChunk.onFillComplete`  inside the `onFillCompelete` method. The reason is that, for all large blobs (larger than 8MiB), we will need to wait for reading the entire 4MiB of the second chunk, before we can send the first chunk to the server. This shouldn't increase the upload time overall, but it's more efficient to send the first chunk to the server, as soon as we have at least one more chunk on the network.

ankagrawal: There are two more things to note in the above code snippet:
1. When we set the `firstChunk` variable in the `prepareForSending` method, then the `firstChunk` is still in `Building` state. When we need a new chunk for more data coming from the network, we will call `getBuildingChunk`, which will be the first chunk. So the above logic also needs change in `getBuildingChunk` to make it work.

2. Even if we make the change in `getBuildingChunk` as mentioned above, note that it's possible that `lastChunk != null`, and firstChunk is still waiting. Because it can happen that `firstChunk` was filled, and `prepareForSending` was called. But there we didn't really send it because we want to wait for more data. After some time we got an empty buffer on the network. And later `chunkFillingCompletedSuccessfully` was marked as `true`, without reading any more data. We will need to find a way to handle this case as well in the above logic.

JingQianCloud: > But `isSimple` is not a property of a chunk. Every is chunk is always "simple" because its always <=4Mib. `isSimple` is the property of a user blob.

Although isSimple is blob concept, we represent it on the CHUNK level in the prepareForSending, right?
I agree we can move the variable to the PutOperation instead inside the PutChunk.
```
        BlobDataType blobDataType =
            isMetadataChunk() ? BlobDataType.METADATA : isSimpleBlob ? BlobDataType.SIMPLE : 
```

JingQianCloud: > Also there is one more issue why I don't prefer doing `firstChunk.onFillComplete` inside the `onFillCompelete` method. The reason is that, for all large blobs (larger than 8MiB), we will need to wait for reading the entire 4MiB of the second chunk, before we can send the first chunk to the server. This shouldn't increase the upload time overall, but it's more efficient to send the first chunk to the server, as soon as we have at least one more chunk on the network.

I don't have much concern about it. For memory usage, we will use at least two chunks for large blob anyway. And the wait only happens on the second chunk, not the following chunks. For large blob, there wont be no difference.
For 8MB chunk, maybe there is a little bit difference but can be ignored.

ankagrawal: To be precise, what we say in the `blobDataType` is that whether this chunk belongs to a larger blob, or is this a blob in itself. In order to make that determination, we need to know the state of the blob.
Consider this in another way. If a blob has multiple data chunks, all the chunks will need to be `chunk.isSimple = false`. If the blob has only one chunk, then all the chunks (in this case only one chunk) will need to be `chunk.isSimple = true`. So having it at a chunk level doesn't help.

ankagrawal: > I don't have much concern about it
My point is, if there is a way to keep it performant, then why not. Also, that's not my main concern. The change in itself will need more changes anyways.

Let's discuss these changes offline, and I will update the comments with the discussion summary.

JingQianCloud: > 

1. Good point for the concern of the getBuildingChunk. We can solve it by adding another state before READY. Let's say RESOLVING. 
2. Solve the problem 2. 

Modified my proposal. Just to share my thought. 
You mentioned you have another version. Happy to review that version. 

Variables:
```
PutOperation: Add two variables
   boolean isSimpleBlob = !options.isChunkUpload();
   boolean blobTypeIsResolved = false;
Chunk State: Add one state
   Add one chunk state: RESOLVING.
```

Code change to two functions:
```
private void prepareForSending {
        if (chunkIndex == 1) { <--- When the second chunk comes, set all the flags.
                isSimpleBlob = false;
                blobTypeIsResolved = true;
                resolveWaitingChunk();
	}
	if (!blobTypeIsResolved) {
             state = ChunkState.RESOLVING;
             return;
	}

	// original code as before
	...
}
```

```
void fillChunks() {
	...
      if (chunkFillingCompletedSuccessfully) {
        PutChunk lastChunk = getBuildingChunk();
        if (lastChunk != null) {
           ....
        }
        blobTypeIsResolved = true; <----- 
         resolveWaitingChunk(); <----- Check any chunk which is waiting for resolving.
  }
}
```

```
void resolveWaitingChunk(){
         for (PutChunk chunk : putChunks) { <--- Check if we have a first chunk haven't been resolved.
             if (chunk.state == RESOLVING()) {
                prepareForSending();
                break;
             }
          }
}
```

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -647,23 +652,32 @@ void fillChunks() {
         }
       }
       if (chunkFillingCompletedSuccessfully) {
-        // If the blob size is less then 4MB or the last chunk size is less than 4MB, than this lastChunk will be
+        // If the blob size is less than 4MB or the last chunk size is less than 4MB, than this lastChunk will be
```
justinlin-linkedin, line 655: the logic in this if statement looks good to me. However, I do think the comments should be more detailed. There are several case we have to deal with within this if statement and there are two decisions we need to make in those cases. 
Decision 1: Is this a simple blob?
Decision 2:  Do we have to send the last chunk to ambry servers?

Case 1: there is no data at all.
In this case: last chunk is not null, and is building, chunkCounter == 0  so the isSimpleBlob == options.isChunkUpload. calling mayResolveAwaitingChunk is an nop since there is no chunk awaiting. And we will send this 0 byte chunk to ambry server at line 667.

Case 2:  there are exactly 4MB data, and there is no empty bytebuf after.
In this case: last chunk is null, chunkCounter == 0. isSimpleBlob = isChunkUpload. maybeResolveAwaitingChunk would send the data to ambry server. Since the lastChunk == null, it will skip the next two if statements.

Case 3: there are exactly 4MB data, but there is an empty bytebuf after 4MB.
In this case: last chunk is not null, and is building with readableBytes == 0. chunkcounter == 1 so the isSimpleBlob = isChunkUpload. maybeResolveAwaitingChunk would send the first chunk out to ambry server.

Case 4: there are more than 4MB data. 
In this case: isSimpleBlob would be false. and it depends on if we have any data in the last chunk to determine weather we want to send it to server.

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
JingQianCloud: Yes, Justin's comment will be helpful.

ankagrawal: Updated the code, so we don't need these comments here.
I have added the comments elsewhere, where `isSimpleBlob` status is being set.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -647,23 +652,32 @@ void fillChunks() {
...
+        if (chunkCounter == 0 || (chunkCounter == 1 && lastChunk != null && lastChunk.buf.readableBytes() == 0)) {
+          // If the chunkCounter is 0, or chunkCounter is 1 but there is no data in the second chunk (i.e, lastChunk), then
+          // effectively there is only one chunk of data it could be awaiting resolution. Resolve it, and mark the blob
+          // as simple blob if it's not a chunked upload.
+          isSimpleBlob = !options.isChunkUpload();
+          maybeResolveAwaitingChunk();
+        }
+
         if (lastChunk != null) {
-          if (chunkCounter != 0 && lastChunk.buf.readableBytes() == 0) {
```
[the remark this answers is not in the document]
ankagrawal: Updated the logic to remove those changes.

JingQianCloud (approved):
LGTM. Approved.
Please wait for Justin to take a look of the final revision although he proved the previous version. Thanks,

PR #2483, opened 2023-06-21
Optimize named blob cleanup process

mikeyang923:
Optimize named blob cleanup process:
1. Update the mysql query to avoid cleaning up any blob_id that's associated with Valid Latest Record of any named blob.
2. Adjust the config parameters for cleanup process.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
@@ -58,14 +58,14 @@ public class MySqlNamedBlobDbConfig {
    * The maximum number of entries to return per each stale blobs pull request.
    */
   @Config(QUERY_STALE_DATA_MAX_RESULTS)
-  @Default("1000")
+  @Default("10000")
   public final int queryStaleDataMaxResults;
 
   /**
    * The maximum number of days for a stale blob to say uncleaned.
    */
   @Config(STALE_DATA_RETENTION_DAYS)
-  @Default("5")
+  @Default("20")
   public final int staleDataRetentionDays;
 
   /**
@@ -86,8 +86,8 @@ public class MySqlNamedBlobDbConfig {
     this.localPoolSize = verifiableProperties.getIntInRange(LOCAL_POOL_SIZE, 5, 1, Integer.MAX_VALUE);
     this.remotePoolSize = verifiableProperties.getIntInRange(REMOTE_POOL_SIZE, 1, 1, Integer.MAX_VALUE);
     this.listMaxResults = verifiableProperties.getIntInRange(LIST_MAX_RESULTS, 100, 1, Integer.MAX_VALUE);
-    this.queryStaleDataMaxResults = verifiableProperties.getIntInRange(QUERY_STALE_DATA_MAX_RESULTS, 1000, 1, Integer.MAX_VALUE);
-    this.staleDataRetentionDays = verifiableProperties.getIntInRange(STALE_DATA_RETENTION_DAYS, 5, 1, Integer.MAX_VALUE);
+    this.queryStaleDataMaxResults = verifiableProperties.getIntInRange(QUERY_STALE_DATA_MAX_RESULTS, 10000, 1, Integer.MAX_VALUE);
+    this.staleDataRetentionDays = verifiableProperties.getIntInRange(STALE_DATA_RETENTION_DAYS, 20, 1, Integer.MAX_VALUE);
     this.dbTransition = verifiableProperties.getBoolean(DB_TRANSITION, false);
     this.dbRelyOnNewTable = verifiableProperties.getBoolean(DB_RELY_ON_NEW_TABLE, false);
   }
--- a/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
@@ -58,7 +58,7 @@ public class MySqlNamedBlobDbConfig {
    * The maximum number of entries to return per each stale blobs pull request.
    */
   @Config(QUERY_STALE_DATA_MAX_RESULTS)
-  @Default("10000")
+  @Default("1000")
   public final int queryStaleDataMaxResults;
 
   /**
@@ -86,7 +86,7 @@ public class MySqlNamedBlobDbConfig {
     this.localPoolSize = verifiableProperties.getIntInRange(LOCAL_POOL_SIZE, 5, 1, Integer.MAX_VALUE);
     this.remotePoolSize = verifiableProperties.getIntInRange(REMOTE_POOL_SIZE, 1, 1, Integer.MAX_VALUE);
     this.listMaxResults = verifiableProperties.getIntInRange(LIST_MAX_RESULTS, 100, 1, Integer.MAX_VALUE);
-    this.queryStaleDataMaxResults = verifiableProperties.getIntInRange(QUERY_STALE_DATA_MAX_RESULTS, 10000, 1, Integer.MAX_VALUE);
+    this.queryStaleDataMaxResults = verifiableProperties.getIntInRange(QUERY_STALE_DATA_MAX_RESULTS, 1000, 1, Integer.MAX_VALUE);
     this.staleDataRetentionDays = verifiableProperties.getIntInRange(STALE_DATA_RETENTION_DAYS, 20, 1, Integer.MAX_VALUE);
     this.dbTransition = verifiableProperties.getBoolean(DB_TRANSITION, false);
     this.dbRelyOnNewTable = verifiableProperties.getBoolean(DB_RELY_ON_NEW_TABLE, false);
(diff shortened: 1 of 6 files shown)


ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -154,24 +154,22 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
-   * To pull out the rows that meet the 2nd condition above, we are using left join and then pull out the rows where
-   * right side table's keys are all NULL
    */
   // @formatter:off
   private static final String GET_STALE_QUERY = String.format(""
-          + "SELECT t1.%s, t1.%s, t1.%s, t1.%s, t1.%s, t1.%s, %s "
-          + "FROM %s t1 "
-          + "LEFT JOIN "
-          + "(SELECT account_id, container_id, blob_name, max(version) as version "
+          + "SELECT %s, %s, %s, %s, %s, %s, %s "
```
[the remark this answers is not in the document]
mikeyang923: It's more complicated because we need to prevent deleting valid blob_id. The previous SQL doesn't prevent that. 
Since blob_id are indexed, it runs pretty fast:
```
mysql> SELECT account_id, container_id, blob_name, blob_id, version, deleted_ts, CURRENT_TIMESTAMP(6) FROM named_blobs_v2 WHERE blob_id not in (SELECT distinct(blob_id)   FROM named_blobs_v2   WHERE version in   (SELECT max(version) as version   FROM named_blobs_v2   WHERE blob_state = 1 and version != 0   GROUP BY account_id, container_id, blob_name) ) AND version<168680160000000000 ORDER BY blob_id LIMIT 10000;
```
It's completed ~6 seconds:

```
CC3FE4E7DD38CD9 |                  0 | NULL                       | 2023-06-23 18:17:19.738068 |
+------------+--------------+--------------------------------------------------------------------------------------------------------+------------------------------------------------------------------------+--------------------+----------------------------+----------------------------+
10000 rows in set (6.11 sec)
```

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -154,24 +154,22 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
-          + "ON (t1.account_id,t1.container_id,t1.blob_name,t1.version) = (t2.account_id,t2.container_id,t2.blob_name,t2.version) "
-          + "WHERE t1.%s<? and t2.account_id IS NULL and t2.container_id IS NULL and t2.blob_name IS NULL and t2.version IS NULL "
-          + "LIMIT ?",
+          + "WHERE blob_id not in "
+          + "   (SELECT distinct(blob_id) "
+          + "    FROM %s "
+          + "    WHERE version in  (SELECT max(version) as version "
+          + "      FROM %s "
+          + "      WHERE blob_state = 1 and version != 0 "
+          + "      GROUP BY account_id, container_id, blob_name))"
```
SophieGuo410, line 171: I would suggest you to add more comments here to explain your ideas.
Something like you select blobId which is not in the subquery, and the subquery is getting distinct blobId where the version == max version of a specific blob.
Also wondering why do we need to order by blobId?
And may I know what's the cost of previous query? We are not an expert for mysql cmd queries, if possible, maybe mysql team can check if this query can be optimized.

mikeyang923: Ok, just added more comments to make it clear for the ideas behind the SQL query.
I order by blobID so that we have deterministic results within the limit. Eg we have total 50K rows. Each cleanup runner run only pull out the 10K (the default limit number).

The time cost is actually reduced by half (see below). I am asking Bei from mysql team to take a look.

**Cost of previous query:**
```
mysql> SELECT t1.account_id, t1.container_id, t1.blob_name, t1.blob_id, t1.version, t1.deleted_ts, CURRENT_TIMESTAMP(6) FROM named_blobs_v2 t1 LEFT JOIN (SELECT account_id, container_id,
blob_name, max(version) as version  FROM named_blobs_v2  WHERE blob_state=1  GROUP BY account_id, container_id, blob_name) t2   ON (t1.account_id,t1.container_id,t1.blob_name,t1.version) = (t2.account_id,t2.container_id,t2.blob_name,t2.version)   WHERE t1.version<168680160000000000 and t2.account_id IS NULL and t2.container_id IS NULL and t2.blob_name IS NULL and t2.version IS NULL LIMIT 10000;

10000 rows in set (11.59 sec)
```

**Query of latest query with same parameter:**
```
mysql> SELECT account_id, container_id, blob_name, blob_id, version, CURRENT_TIMESTAMP(6) FROM named_blobs_v2 WHERE blob_id not in (SELECT distinct(blob_id)  FROM named_blobs_v2  WHERE version in  (SELECT max(version) as version  FROM named_blobs_v2  WHERE blob_state = 1 and version != 0  GROUP BY account_id, container_id, blob_name) ) AND version<168680160000000000 AND (deleted_ts IS NULL OR deleted_ts>"2023-06-10") ORDER BY blob_id desc LIMIT 10000;

10000 rows in set (6.70 sec)

```

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -58,14 +58,14 @@ public class MySqlNamedBlobDbConfig {
    * The maximum number of entries to return per each stale blobs pull request.
    */
   @Config(QUERY_STALE_DATA_MAX_RESULTS)
-  @Default("1000")
+  @Default("10000")
```
justinlin-linkedin, line 61: why increase the result, can we keep it as 1000, this would limit the time to run the query.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -213,11 +213,11 @@ public void start() throws InstantiationException {
     namedBlobsCleanupRunner = new NamedBlobsCleanupRunner(router, namedBlobDb);
     if (frontendConfig.enableNamedBlobCleanupTask) {
       namedBlobsCleanupScheduler = Utils.newScheduler(1, "named-blobs-cleanup-", false);
-      int oneWeekInSeconds = 60 * 60 * 24 * 7;
-      int initialDelayInSeconds = random.nextInt(oneWeekInSeconds);
+      int oneDayInSeconds = 60 * 60 * 24;
```
justinlin-linkedin, line 216: run it every 7 days would be good, or we can put it in a configuration.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
```diff
@@ -248,6 +248,7 @@ public void testCleanupBlobsPipeline() throws Exception {
 
     int staleCount = 10;
     int needCleanupCount = 0;
+    List<NamedBlobRecord> staleRecords = new ArrayList<>();
```
justinlin-linkedin, line 251: is there a new test case? Can we add a test case where the blob id is shared by two versions? and see if the query solve this issue?

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
mikeyang923: I thought 1000 is too few to handle all stale blobs in one day (since i always schedule the cleanup to run once per day). 
Now i am using config-driven schedule interval for the named blob cleanup process, I just changed it back to 1000.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
mikeyang923: I updated the schedule interval to be config-driven with weekly schedule.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
mikeyang923: Yes, that's tested in `testCleanupBlobsPipeline`, which actually failed after using the new query in `MySqlNamedBlobDb`. I need to make the change to make it pass, and as you can see it only delete the stale rows that do not share valid blob_id with latest record. I have to use a blobId2 to create a different blobId case which can be cleaned up:
```
      String blobId2 = getBlobId(account, container);
      NamedBlobRecord record2 =
          new NamedBlobRecord(account.getName(), container.getName(), blobName, blobId2, expirationTime);
      namedBlobDb.put(record2, NamedBlobState.READY, true).get();
```

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobsCleanupRunner.java
```diff
@@ -56,6 +58,8 @@ public void run() {
       namedBlobDb.cleanupStaleData(staleResultList);
       logger.info("Named Blobs Cleanup Runner is completed for {} stale cases (there are {} failed cases)",
           staleResultList.size(), failedResults.size());
+      Set<String> cleanedBlobIds = staleResultList.stream().map(StaleNamedBlob::getBlobId).collect(Collectors.toSet());
+      logger.info("The cleaned blobIds are: {}", cleanedBlobIds);
```
justinlin-linkedin, line 62: can we do a for to print out more information, like the blob name and the blob id.

justinlin-linkedin (approved):
Please make sure that mysql team checks out this query.

SophieGuo410 (approved):

PR #2485, opened 2023-06-22
Adding dataset version state for partial upload

SophieGuo410:
Previously, we support best effort when dataset version upload failed, we fallback to delete that version, and if dataset version deletion failed, we ask user to manually deleted it. That would lead to a problem that we will count the partial upload data into retention policy. So this logic is to resolve this issue. 
But the latest version should still count the partial upload data since before manual delete, the record are still in datasetVersion db which can't be used for next upload.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
@@ -27,6 +27,7 @@ import com.github.ambry.account.DatasetVersionRecord;
 import com.github.ambry.config.MySqlAccountServiceConfig;
 import com.github.ambry.frontend.Page;
 import com.github.ambry.mysql.MySqlDataAccessor;
+import com.github.ambry.protocol.DatasetVersionState;
 import com.github.ambry.utils.Utils;
 import java.io.IOException;
 import java.sql.PreparedStatement;
@@ -82,6 +83,7 @@ public class AccountDao {
 
   // Dataset version table fields
   public static final String DATASET_VERSION_TABLE = "DatasetVersions";
+  public static final String DATASET_VERSION_STATE = "datasetVersionState";
 
   // Common fields
   public static final String VERSION = "version";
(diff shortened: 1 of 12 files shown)


JingQianCloud (approved):
LGTM

justinlin-linkedin (commented):
at this moment, I think it's worth to create a different Dao file, say DatasetDao, to handle all the Dataset related queries.

SophieGuo410:
> at this moment, I think it's worth to create a different Dao file, say DatasetDao, to handle all the Dataset related queries.

Sure, I can do it as a code refactoring in a future pr.

ambry-account/src/main/java/com/github/ambry/account/mysql/AccountDao.java
```diff
@@ -203,32 +206,39 @@ public AccountDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
...
-    listValidVersionSql =
+    listValidVersionForDatasetDeletionSql =
         String.format("select %s, %s from %s " + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ?",
             VERSION, DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME);
     // list all valid versions sorted by last modified time, and skip the first N records which is not out of retentionCount.
     listVersionByModifiedTimeSql = String.format("select %s, %s from %s "
-            + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ? ORDER BY %s DESC LIMIT ?, 100", VERSION,
-        DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME,
-        LAST_MODIFIED_TIME);
+            + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ? and %s = ? ORDER BY %s DESC LIMIT ?, 100",
```
justinlin-linkedin, line 228: in cachedAccountService, you don't have a method for `getAllValidVersionsOutOfRetentionCount`? why is that?

SophieGuo410: It seems like the method has been accidentally deleted, so I added back in this pr. https://github.com/linkedin/ambry/pull/2489

justinlin-linkedin (approved):

PR #2486, opened 2023-06-22
Log named blob mysql db queries for debugging 

Arun-LinkedIn:


ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -555,7 +555,9 @@ private NamedBlobRecord run_get_v2(String accountName, String containerName, Str
       statement.setInt(1, accountId);
       statement.setInt(2, containerId);
       statement.setString(3, blobName);
-      try (ResultSet resultSet = statement.executeQuery()) {
+      logger.debug("Getting blob name from MySql. Query {}", statement);
+      ResultSet resultSet = statement.executeQuery();
+      try {
         if (!resultSet.next()) {
```
mikeyang923, line 561: If you just want to log for debug purpose. you can just log it like below, why need to move the resources outside of try() block?

```
try (ResultSet resultSet = statement.executeQuery()) {
logger.debug(.....);
}
```

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -571,18 +573,25 @@ private NamedBlobRecord run_get_v2(String accountName, String containerName, Str
           return new NamedBlobRecord(accountName, containerName, blobName, blobId, timestampToMs(deletionTime),
               version);
         }
+      } catch (Exception e) {
```
justinlin-linkedin, line 576: try-with-resource statement can actually have a catch clause, something like
```
try (ResultSet resultSet = statement.executeQuery() ) {
} catch (Exception e) {
    logger.error("xxxx");
    throw e;
}

File: ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java (hunk 1 above)
Comment (Arun-LinkedIn) [reply ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java:562]: Sure. I was also trying to print the mysql query in catch clause. Due to that, I moved it outside. But, added back the try-with-resources now

File: ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java (hunk 2 above)
Comment (Arun-LinkedIn) [reply ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java]: Sure. Added back the try-with-resource statement

**Review by justinlin-linkedin — APPROVED** [review]

File: ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java [hunk 3]
```diff
@@ -572,17 +575,23 @@ private NamedBlobRecord run_get_v2(String accountName, String containerName, Str
               version);
         }
       }
+    } catch (Exception e) {
+      logger.error("MySql query {} received exception", query, e);
+      throw e;
```
mikeyang923, line 580: nit: we are having the same ```MySql query {} received exception``` error log in multiple places. It's better to put the method name or some unique thing in the string to differentiate between them

mikeyang923 (approved):
lgtmbc

PR #2489, opened 2023-06-23
Support retention policy for dataset with default value

SophieGuo410:
1.Due to espresso back up use case, we plan to support customized retention policy.
By default, user will use the default counter based retention policy which will delete versions out of retention count.
This pr is to support a new field at dataset level so in the future we can support customized retention policy.
2.Also, I realized that the getAllValidVersionsOutOfRetentionCount is missing in mysqlAccountService. Adding the method and metrics.


justinlin-linkedin (approved):

JingQianCloud (approved):
LGTM

PR #2496, opened 2023-06-27
Add transaction isolation level for Named blobs MySql DB

Arun-LinkedIn:
Set transaction isolation level to connections for Named Blob DB. The value is taken via configuration.

### DIFF
--- a/ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDbFactory.java
+++ b/ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDbFactory.java
@@ -71,6 +71,9 @@ public class MySqlNamedBlobDbFactory implements NamedBlobDbFactory {
     hikariConfig.addDataSourceProperty("prepStmtCacheSize", "250");
     hikariConfig.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
     hikariConfig.addDataSourceProperty("useServerPrepStmts", "true");
+    if (!config.transactionIsolationLevel.equals(TransactionIsolationLevel.TRANSACTION_NONE)) {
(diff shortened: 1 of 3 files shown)


justinlin-linkedin (approved):

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDbFactory.java
```diff
@@ -71,6 +71,9 @@ public HikariDataSource buildDataSource(DbEndpoint dbEndpoint) {
     hikariConfig.addDataSourceProperty("prepStmtCacheSize", "250");
     hikariConfig.addDataSourceProperty("prepStmtCacheSqlLimit", "2048");
     hikariConfig.addDataSourceProperty("useServerPrepStmts", "true");
+    if (!config.transactionIsolationLevel.equals(TransactionIsolationLevel.TRANSACTION_NONE)) {
+      hikariConfig.setTransactionIsolation(config.transactionIsolationLevel.name());
+    }
```
mikeyang923, line 76: So we want set the transaction isolation when creating the DB instance and it will stay that level for the DB instance's whole life? i thought we want to set the isolation level only at specific method (eg soft delete): eg
```
softDelete function: 
      set isolation level
      do soft delete
      in finally reset the isolation level to its default

```

Arun-LinkedIn: Yeah, true. This sets transaction isolation level for all connections to DB. I was thinking that there is no harm in setting it for all queries. I think MySql team also suggested same. It also makes debugging easier.

mikeyang923 (approved):

PR #2498, opened 2023-06-28
Add metrics to track host capacity utilization

Arun-LinkedIn:
This will be useful to track replica distribution in Full Auto.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManagerMetrics.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManagerMetrics.java
@@ -114,6 +114,22 @@ public class StorageManagerMetrics {
   }
 
   /**
+   * Initializes gauge that track the percentage used capacity of the host
+   * @param storageManager the {@link StorageManager} instance to use to obtain values.
+   */
+  void initializeHostUtilizationTracker(final StorageManager storageManager) {
+    Gauge<Double> percentageUsedCapacity = storageManager::getHostPercentageUsedCapacity;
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/StorageManagerMetrics.java
```diff
@@ -113,6 +113,22 @@ public StorageManagerMetrics(MetricRegistry registry) {
...
+   */
+  void initializeHostUtilizationTracker(final StorageManager storageManager) {
+    Gauge<Double> percentageUsedCapacity = storageManager::getHostPercentageUsedCapacity;
+    registry.register(MetricRegistry.name(StorageManager.class, "HostPercentageUsedCapacity"), percentageUsedCapacity);
+  }
+
+  /**
+   * Deregister the Metrics related to the host utilization.
+   */
+  void deregisterHostUtilizationTracker() {
```
litingulfs, line 128: Why do we need to remove this metric from the registry?

litingulfs (approved):

SophieGuo410 (approved):

ambry-store/src/main/java/com/github/ambry/store/StorageManagerMetrics.java
Arun-LinkedIn: My understanding is that when we register gauge metric with registry, we pass the `StorageManager` object reference to it. I think removing the metric from registry will help clean up its reference and make it available for garbage collection. 

Since there is only one instance of StorageManager, it may be okay in this case not to deregister. But, if we are having metrics for objects which are creating and deleted frequently, I think we have to remove their references in registry to avoid memory build up.

PR #2499, opened 2023-06-28
Fix a bug in InputStreamReadableStreamChannel

justinlin-linkedin:
InputStreamReadableStreamChannel offers bytes to the AsyncWritableChannel with one byte array, and when the size of the stream is larger than the byte array's size, it will reuse this byte array. This is fine if the AsyncWritableChannel makes a copy of the byte array. But ByteBufferAsyncWritableChannel doesn't make any copy. So InputStreamReadableStreamChannel doesn't work with ByteBufferAsyncWritableChannel. 

This PR fixes this issue by forcing InputStreamReadableStreamChannel to not reuse byte array.


SophieGuo410 (approved):
lgtm!
Just curious, how did you find this bug?

justinlin-linkedin:
> lgtm! Just curious, how did you find this bug?

we are using this class in the client code.

Arun-LinkedIn (approved):
Nice catch!

PR #2500, opened 2023-06-28
Fix bug for retentionLogic and latest version

SophieGuo410:
1. fix the bug that only update state for new db.
2. for latest version, if the version gets deleted, it will be considered as invalid version and we need to update it.
3. update restMethod when delete dataset out of retention.


JingQianCloud (approved):
LGTM

PR #2503, opened 2023-07-10
Fix bug for acl check and retention

SophieGuo410:
1.  For dataset version upload, we should run permission check before query the dataset table.
2. For retention policy support, we need to make sure the put has been set back after last call back has been triggered, not when generating the callback. Adding the test cases to make sure the delete has been called.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
@@ -77,7 +77,7 @@ public class DeleteBlobHandler {
     RequestPath requestPath = getRequestPath(restRequest);
     // named blob requests have their account/container in the URI, so checks can be done prior to ID conversion.
     if (requestPath.matchesOperation(Operations.NAMED_BLOB)) {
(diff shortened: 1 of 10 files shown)


justinlin-linkedin (approved):
LGTM

ambry-frontend/src/main/java/com/github/ambry/frontend/DeleteBlobHandler.java
```diff
@@ -160,6 +160,7 @@ private Callback<Void> routerCallback() {
         if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
           try {
             metrics.deleteDatasetVersionRate.mark();
+            accountAndContainerInjector.injectDatasetForNamedBlob(restRequest);
```
JingQianCloud, line 163: Why do we have to set the dataset version here not in handle function?

SophieGuo410: Because we need to make sure the dataset access is after the permission check which is done in processRequest in close source.

JingQianCloud (approved):
LGTM

PR #2504, opened 2023-07-11
Add changes to make disk selection locally when rolling back from Full-Auto to Semi-auto

Arun-LinkedIn:
During roll back to Semi-auto, we will use the replica placement calculated by Helix (in Full-auto mode) to avoid too many replica movements. However, this means that we should continue to choose disk for the replica locally on the server instead of relying on the static map.

To distinguish the rollback scenario from regular semi-auto, we will keep track of list of resources migrating to Full-Auto. We will read this list on servers and choose disk locally if the resource is present in the list.


justinlin-linkedin (commented):
LGTM, could you please add some test cases?

justinlin-linkedin (approved):

SophieGuo410 (approved):

PR #2505, opened 2023-07-11
Some change to make http handling a bit less clumsy 

justinlin-linkedin:


SophieGuo410 (commented):
I don't have concern for the code refactoring, but I would suggest if you can split it into some small PRs for easy review. Let me know if it's doable.

PR #2508, opened 2023-07-14
[Compaction] Stagger the start of full compaction between multiple replicas on a disk.

ankagrawal:
This is needed to ensure that all the stores on a disk do not start full compaction at the same time.


justinlin-linkedin (approved):
LGTM

PR #2509, opened 2023-07-14
Log errors only for SQL exceptions and not internally generated ones

Arun-LinkedIn:
Avoid logging ERROR message type for internally generated exceptions such as `om.github.ambry.rest.RestServiceException: GET: Blob not found` in `MySqlNamedBlobDb.java` class since these are informative and are caught higher in the call stack.


justinlin-linkedin (approved):
Can we avoid using e.getMessage, but just log out the entire exception with stacktrace?

SophieGuo410 (approved):

Arun-LinkedIn:
> Can we avoid using e.getMessage, but just log out the entire exception with stacktrace?

It seems like we log the entire exception message in caller method in `NettyResponseChannel`. Due to that, logged only e.getMessage here.

PR #2511, opened 2023-07-15
Support list all dataset versions

SophieGuo410:


ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -144,10 +145,14 @@ public DatasetDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
         String.format("select %s, %s from %s " + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ?",
             VERSION, DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME);
     // list all valid versions sorted by last modified time, and skip the first N records which is not out of retentionCount.
-    listVersionByModifiedTimeSql = String.format("select %s, %s from %s "
+    listVersionByModifiedTimeAndFilterByRetentionSql = String.format("select %s, %s from %s "
             + "where (%s IS NULL or %s > now(3)) and %s = ? and %s = ? and %s = ? and %s = ? ORDER BY %s DESC LIMIT ?, 100",
         VERSION, DELETE_TS, DATASET_VERSION_TABLE, DELETE_TS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME,
         DATASET_VERSION_STATE, LAST_MODIFIED_TIME);
+    listValidDatasetVersionsSql = String.format("select %1$s from %2$s "
+            + "where (%3$s IS NULL or %3$s > now(3)) and %4$s = ? and %5$s = ? and %6$s = ? and %7$s = ? and (%1$s >= ? or ? IS NULL) "
```
justinlin-linkedin, line 153: we should check if the page token is NULL out side of this query statement, we should just do `AND %1$s >= ?`

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -918,6 +959,47 @@ private Page<String> executeListAllValidDatasetsStatement(PreparedStatement stat
...
+  private Page<String> executeListAllValidDatasetVersionsStatement(PreparedStatement statement, int accountId,
+      int containerId, String datasetName, String pageToken, Dataset.VersionSchema versionSchema) throws SQLException {
+    ResultSet resultSet = null;
+    try {
+      statement.setInt(1, accountId);
+      statement.setInt(2, containerId);
+      statement.setString(3, datasetName);
+      statement.setInt(4, DatasetVersionState.READY.ordinal());
+      statement.setString(5, pageToken);
+      statement.setString(6, pageToken);
```
justinlin-linkedin, line 982: we should convert pageToken back to a version number before passing it to the query statement, we should do it before calling this execute method.

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -918,6 +959,47 @@ private Page<String> executeListAllValidDatasetsStatement(PreparedStatement stat
...
+      statement.setInt(7, mySqlAccountServiceConfig.listDatasetVersionsMaxResult + 1);
+      resultSet = statement.executeQuery();
+      List<String> entries = new ArrayList<>();
+      String nextContinuationToken = null;
+      int resultIndex = 0;
+      while (resultSet.next()) {
+        long versionValue = resultSet.getLong(VERSION);
+        String version = convertVersionValueToVersion(versionValue, versionSchema);
+        if (resultIndex++ == mySqlAccountServiceConfig.listDatasetVersionsMaxResult) {
```
justinlin-linkedin, line 991: nit: I really don't like have `++` operator right after. can you increase the resultIndex right after this if statement.

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -918,6 +959,47 @@ private Page<String> executeListAllValidDatasetsStatement(PreparedStatement stat
...
+          nextContinuationToken = version;
```
JingQianCloud, line 992: If I understand correctly, the version we save in the db is "long" for all three types of versionSchema.
For the continue token, I think we don't need convert to customer readable string token. The native "long" type that we save in the database is good enough. Customer doesn't need to know the meaning of the token.
In this way, we don't need "convert to long before query" or "convert to string before sending token to the customer."

And with this PR, I don't see the code that we "convert the string token before query". Am I missing anything? If it's true, the list version with SEMANTIC won't work, right?

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -557,18 +562,21 @@ public synchronized Dataset getDataset(int accountId, int containerId, String ac
...
       dataAccessor.getDatabaseConnection(false);
+      PreparedStatement getVersionSchemaStatement = dataAccessor.getPreparedStatement(getVersionSchemaSql, false);
+      Dataset.VersionSchema versionSchema =
+          executeGetVersionSchema(getVersionSchemaStatement, accountId, containerId, datasetName);
       PreparedStatement listAllValidVersionForDatasetDeletionStatement =
           dataAccessor.getPreparedStatement(listValidVersionForDatasetDeletionSql, false);
       List<DatasetVersionRecord> datasetVersionRecords =
           executeListAllValidVersionForDatasetDeletionStatement(listAllValidVersionForDatasetDeletionStatement,
-              accountId, containerId, datasetName);
+              accountId, containerId, datasetName, versionSchema);
```
JingQianCloud, line 576: Looks like right now we always serialize the version based on the versionSchema. Because of that, for each API, we need two mysql API calls. One is to get the versionSchema. The other is the real one which lists the version set either for deletion or for listAll.
Not sure if the extra getSchema query will have any performance impact. For long term, probably we can think about a way to avoid the getSchema mysql call. For example, probably client has this information already and it can provide to the API.

ambry-api/src/main/java/com/github/ambry/frontend/DatasetVersionPath.java
```diff
@@ -0,0 +1,106 @@
...
+    int expectedMinimumSegments = isListRequest ? 4 : 5;
+    if (splitPath.length < expectedMinimumSegments) {
+      throw new RestServiceException(String.format(
+          "Blob name must have format '/named/<account_name>/<container_name>/<dataset_name>%s'.  Received path='%s'",
+          isListRequest ? "" : "/<version>'", path), RestServiceErrorCode.BadRequest);
+    }
+    String accountName = splitPath[1];
+    String containerName = splitPath[2];
+    String datasetName;
+    if (isListRequest) {
```
JingQianCloud, line 63: Can you add some comment for the datasetName composition for both of the cases when isListRequest true or false?
Giving the example of the expected args.

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -667,6 +667,21 @@ public Page<String> listAllValidDatasets(String accountName, String containerNam
     }
   }
 
+  @Override
+  public synchronized Page<String> listAllValidDatasetVersions(String accountName, String containerName,
```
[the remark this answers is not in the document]
JingQianCloud: +1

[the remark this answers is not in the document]
SophieGuo410: Removed.

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
SophieGuo410: Fixed.

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
SophieGuo410: I forgot that I store version as long type before. So when page token is semantic, the comparison is comparing a string x.x.x with the long value which is wrong. And I also missed the test for that use case, so I added the new test. Now it should work.

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
SophieGuo410: the version schema is provided at dataset level(one time) before user adding the dataset version, so they don't need to provide the version schema during dataset version upload.

ambry-api/src/main/java/com/github/ambry/frontend/DatasetVersionPath.java
SophieGuo410: the isListRequest is just determined by the header when people want to enable this feature.
boolean isListRequest = RestUtils.getBooleanHeader(args, ENABLE_DATASET_VERSION_LISTING, false);
May I know what kind of extra comment you want me to add?

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
JingQianCloud: Yes, I understand the dataset version schema is pre-determined by customer and it is saved in dataset table, not the version table.

Just for this API, in order to serialize the version, we have to make two mysql calls when we really care is the version table. I don't know if we have any latency data, for example how long is one mysql call takes and how long   we usually finish this API. Does the getSchema mysql call have any performance impact?

JingQianCloud (approved):
In my opinion, we don't need "synchronize" the listDataSetVersions with deleteVersion etc.
It's ok to show up deleted version. And since the requests can get to any frontend node, the "synchronization" won't help either.

Other than that, it's good to me. Approved.

SophieGuo410:
> In my opinion, we don't need "synchronize" the listDataSetVersions with deleteVersion etc. It's ok to show up deleted version. And since the requests can get to any frontend node, the "synchronization" won't help either.
> 
> Other than that, it's good to me. Approved.

Checked with @Arun-LinkedIn, the synchronize is actually to prevent request sending from same frontend might close a connection in one method while other method is still in progress. I'll move the synchronize at one layer above which is MySqlAccountStore. And will try to use hikari library to create a connection pool instead of using single connection in a future pr since the qps has been increased.

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
justinlin-linkedin: ?
```
if (resultIndex == mysqlAccountServiceConfig.listDatasetVersionsMaxResult) {
    nextContinuationToken = version;
    break;
}
resultIndex ++;
```
This is exactly the same as your PR since you are not using resultIndex within the if statement and after while loop.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -290,6 +294,9 @@ public void handleGet(final RestRequest restRequest, final RestResponseChannel r
           && NamedBlobPath.parse(requestPath, restRequest.getArgs()).getBlobName() == null) {
         namedBlobListHandler.handle(restRequest, restResponseChannel,
             (result, exception) -> submitResponse(restRequest, restResponseChannel, result, exception));
+      } else if (RestUtils.getBooleanHeader(restRequest.getArgs(), ENABLE_DATASET_VERSION_LISTING, false)) {
+        listDatasetVersionHandler.handle(restRequest, restResponseChannel,
+            (result, exception) -> submitResponse(restRequest, restResponseChannel, result, exception));
```
justinlin-linkedin, line 299: This predicate is too simple, can we add some condition on the url as well.

ambry-frontend/src/main/java/com/github/ambry/frontend/ListDatasetVersionHandler.java
```diff
@@ -0,0 +1,162 @@
...
+     * request time security checks that rely on the request being fully parsed and any additional arguments set.
+     * @return a {@link Callback} to be used with {@link SecurityService#processRequest}.
+     */
+    private Callback<Void> securityPostProcessRequestCallback() {
+      return buildCallback(frontendMetrics.getDatasetsSecurityPostProcessRequestMetrics, securityCheckResult -> {
+        LOGGER.debug("Received request for listing all datasets with arguments: {}", restRequest.getArgs());
+        accountAndContainerInjector.injectDatasetForNamedBlob(restRequest);
+        Page<String> datasetList = listAllValidDatasetVersions();
+        ReadableStreamChannel channel = FrontendUtils.serializeJsonToChannel(
+            datasetList.toJson(datasetVersion -> new JSONObject().put(DATASET_VERSION_KEY, datasetVersion)));
```
justinlin-linkedin, line 128: do we need to return a list of object that only contains one version key back to client? Can we just return a list of versions?

ambry-api/src/main/java/com/github/ambry/frontend/DatasetVersionPath.java
```diff
@@ -0,0 +1,108 @@
+/*
```
justinlin-linkedin, line 1: please format this file.

ambry-frontend/src/main/java/com/github/ambry/frontend/ListDatasetVersionHandler.java
SophieGuo410: Updated to use new serialize and deserialize logic in Page.

ambry-api/src/main/java/com/github/ambry/frontend/DatasetVersionPath.java
SophieGuo410: Done.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
SophieGuo410: Added one more condition to check if the version info is empty. But the major condition is ENABLE_DATASET_VERSION_LISTING because it's hard for us to detect if they provide the version or not, since the datasetName could include "/".

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
SophieGuo410: You are right, updated.

justinlin-linkedin (approved):

PR #2514, opened 2023-07-17
Make the other StaticClusterAgentsFactory constructor public.

panchow:
This is required by Caspian to instantiate StaticClusterAgentsFactory using an a PartitionLayout instance.  Caspian does not use the other public constructor that requires passing in full path of the Hardware and Partition Layout file.


justinlin-linkedin (approved):
LGTM

Arun-LinkedIn (approved):

PR #2515, opened 2023-07-18
Fix Full-auto migration tool 

Arun-LinkedIn:
Add minor fixes and prints to Full-auto migration tool. Verified it by remigrating Perf cluster from Semi-auto to Full-auto with the tool.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
@@ -410,24 +410,29 @@ public class HelixBootstrapUpgradeUtil {
 
   /**
    * Migrate resources from semi-auto to full-auto.
-   * @param hardwareLayoutPath the path to the hardware layout file.
-   * @param partitionLayoutPath the path to the partition layout file.
-   * @param zkLayoutPath the path to the zookeeper layout file.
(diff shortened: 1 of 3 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixBootstrapUpgradeUtil.java
```diff
@@ -897,73 +901,50 @@ public void migrateToFullAuto(String commaSeparatedResources, WagedHelixConfig w
...
-            }
-          }
-          if (!invalidInstances.isEmpty()) {
-            logger.error("Validation of waged rebalancer configuration failed for instances {}", invalidInstances);
-            throw new IllegalStateException(
-                "Instances" + invalidInstances + "are not configured correctly for waged rebalancer");
-          }
-          // 4.2 Resource validation
-          Map<String, Boolean> resourceValidationResultMap =
-              helixAdmin.validateResourcesForWagedRebalance(clusterName, new ArrayList<>(resources));
```
Arun-LinkedIn, line 924: Removed validation for now since `ZkHelixAdmin.validateResourcesForWagedRebalance` APIs don't seem to support tags and expect all the instances of the cluster to be configured for Waged rebalancer. We can probably do manual validation via curl if that works.

justinlin-linkedin (approved):

SophieGuo410 (approved):

PR #2520, opened 2023-07-26
FULL_AUTO: use the old way to deal with disk failures, also remove replicas from replication manager

justinlin-linkedin:
This PR reverted some of the logic in this PR https://github.com/linkedin/ambry/pull/2516. 

We will be doing all the operations in disk failure handler, and also remove replicas from replication and stats manager. 
The reason to not rely on the partition state change listener is because when a disk already failed before server starts, all the replicas will be errored out when going through upward state transition. In this case, when we reset partitions, those replicas would be transitioned to offline and then to bootstrap, not dropped, if helix decided to keep those replicas in the same host. If we don't remove those replicas from property store, then those replicas would not be treated as bootstrap replicas, and server won't find a new disk for them. That's why we have to remove them from property store. So we do all the operations in the disk failure handler.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -736,6 +736,9 @@ public class StorageManager implements StoreManager {
         // During decommission, imagine an edge case where the node crashed immediately after removing replica from
         // InstanceConfig. So next time when this node is restarted, old replica is no longer present in Helix cluster
         // map. We still need to clean up file/directory on disk that is associated with removed replica.
+
+        // There is another case where the replica would be null, which is when the disk failed, we removed all the
+        // replicas on the disk. We can try to remove the diretory, but it won't find any.
         try {
           maybeDeleteResidualDirectory(partitionName);
         } catch (IOException e) {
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -756,31 +759,32 @@ public void onPartitionBecomeDroppedFromOffline(String partitionName) {
...
-              "Exception occurred when resuming decommission on replica " + partitionName, ReplicaOperationFailure);
+      if (store == null && isReplicaOnFailedDisk(replica)) {
+        logger.info("Replica is in a failed disk, blob store not started. skip");
+      } else {
+        // 1. Check if the store is recovering from decommission or directly transitioning from OFFLINE to DROPPED in
+        // full-auto (i.e. if this replica has been reassigned to a different host when it is down, Helix may directly
+        // transition the replica from OFFLINE -> DROPPED without going through OFFLINE -> BOOTSTRAP -> STANDBY ->
+        // INACTIVE -> OFFLINE -> DROPPED steps). If so, go through decommission steps to make sure peer replicas are
+        // caught up with local replica and we update DataNodeConfig in Helix.
+        if (store.recoverFromDecommission() || (clusterMap.isDataNodeInFullAutoMode(replica.getDataNodeId())
```
Arun-LinkedIn, line 770: Minor: Would we need null check here? It seems like we could come here if store == null and isReplicaOnFailedDisk(replica) = false.

Arun-LinkedIn (approved):

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
justinlin-linkedin: I don't think we need to. When deal with disk failures, the replicas has already removed from partitionNameToReplicaId, so the replicaId will be null and the execution would enter the if statement above and return in the if statement.

JingQianCloud (approved):
LGTM.

PR #2530, opened 2023-08-04
Fix the order of blob store to passed to compaction manager

justinlin-linkedin:
We are passing a collection of blob store to compaction manager from disk manager, but this collection is generated from a concurrent hashmap's values() method. It doens't guarantee the order of the blob stores. So we might run into a case when ambry-server restarts, we get different order on the same collection. Some blob store might be compacted multiple times before others. This PR fixes it.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
@@ -135,6 +135,9 @@ public class DiskManager {
     this.replicaStatusDelegates = replicaStatusDelegates;
     this.stoppedReplicas = stoppedReplicas;
     expectedDirs.add(reserveFileDir.getAbsolutePath());
+    // Create an array list to add all the stores and pass it to compaction manager so that we have a fixed order
+    // of blob stores to compact, even between restarts.
+    List<BlobStore> storesToCompact = new ArrayList<>();
     for (ReplicaId replica : replicas) {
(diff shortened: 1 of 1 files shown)


ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -135,6 +135,9 @@ public class DiskManager {
     this.replicaStatusDelegates = replicaStatusDelegates;
     this.stoppedReplicas = stoppedReplicas;
     expectedDirs.add(reserveFileDir.getAbsolutePath());
+    // Create an array list to add all the stores and pass it to compaction manager so that we have a fixed order
+    // of blob stores to compact, even between restarts.
+    List<BlobStore> storesToCompact = new ArrayList<>();
```
ankagrawal, line 140: The list of replicas is originally obtained from `ClusterMap::getReplicaIds()` which makes no ordering guarantees for list of replicas returned each time its called (e.g, for `HelixClusterManager` it comes from a map). Since this original list of replicas has no ordering guarantee, how will the order of blob stores be fixed across restarts? Also, `CompactionManager` takes this list and converts it into a Set backed by a `ConcurrentMap`, so even if the list of replicas always had same order, they won't preserve the ordering in CompactionManager.

Also, the ordering will change only for the first cycle after the restart. After that the ordering will remain same. Is this still a big issue generally?

PR #2533, opened 2023-08-08
Allow SimpleOperationTracker continue to run even there is not enough quorum

JingQianCloud:
If we only have one single eligible replica, SimpleOperationTracker fails the Delete Request right away. We don't have chance to run Offline Repair.
When offline repair is enabled, allow SimpleOperationTracker continue to run even there is not enough replicas to fulfill the success target count.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -350,7 +350,12 @@ class SimpleOperationTracker implements OperationTracker {
         generateErrorMessage(partitionId, examinedReplicas, replicaPool, backupReplicasToCheck, downReplicasToCheck,
             routerOperation));
     if (totalReplicaCount < getSuccessTarget()) {
-      throw notEnoughReplicasException.get();
+      if (possibleRunOfflineRepair(totalReplicaCount)) {
+        logger.info("RepairRequest: Not enough quorum for delete but give it a try since offline repair is enabled {}",
+            blobId);
+      } else {
+        throw notEnoughReplicasException.get();
+      }
     }
 
     originatingDcTotalReplicaCount = (int) allReplicas.stream()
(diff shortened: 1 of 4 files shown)


ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -567,7 +581,13 @@ public boolean hasFailed() {
       }
       // if there is no possible way to use the remaining replicas to meet the success target,
       // deem the operation a failure.
-      return replicaInPoolOrFlightCount + replicaSuccessCount < replicaSuccessTarget;
+      // For Delete, even there is not enough quorum, if offline repair is enabled, continue to run it.
+      if (possibleRunOfflineRepair(replicaInPoolOrFlightCount + replicaSuccessCount)) {
```
SophieGuo410, line 585: QQ: it seems like once the offline repair enabled, we will not fail the request. But will it be any chance that the record has not been successfully stored in the repair db? Or probably due to the availability is high, so we can ignore that?
Also, we probably want to have metrics to record how many times this issue happened?

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -369,6 +374,15 @@ class SimpleOperationTracker implements OperationTracker {
         originatingDcName);
   }
 
+  /**
+   * @param eligibleReplicaCount eligible replicas to send Delete Requests.
+   * @return true if it's possible to run offline repair for the Delete
+   */
+  public boolean possibleRunOfflineRepair(int eligibleReplicaCount) {
+    return (eligibleReplicaCount >= 1 && routerOperation == RouterOperation.DeleteOperation
+        && routerConfig.routerDeleteOfflineRepairEnabled && routerConfig.routerRepairRequestsDbFactory != null);
```
SophieGuo410, line 383: Would it be safer if we check getRepairRequestsDb is not null? Since it might failed to create the db.

justinlin-linkedin, line 383: as long as the offline repair is on, we should allow it, we don't have to check the DbFactory.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
justinlin-linkedin, line 585: I suppose we don't have to run this check again, if we know in the constructor that we are going to allow simple operation tracker to continue even if there is no sufficient replicas, we can use a field various to save this flag and use it here rightaway.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
JingQianCloud: Remove the routerRepairRequestsDBFactory check.
At the same time, make routerDeleteOfflineRepairEnabled check the factory already.

SophieGuo410 (approved):
lgtm!

justinlin-linkedin (approved):

PR #2534, opened 2023-08-08
[Named Blob] Filter out expired/deleted entries in listing named blob

mikeyang923:
Filter out expired/deleted entries in listing named blob


justinlin-linkedin (commented):
Please add some test cases

mikeyang923:
> Please add some test cases

Ok, just added. thanks!

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
```diff
@@ -164,6 +167,50 @@ public void testUpdateBlobStateToReady() throws Exception {
     assertEquals("Blob Id is not matched with the record", id, namedBlobRecord.getBlobId());
   }
 
+  @Test
+  public void testListNamedBlobs() throws Exception {
```
justinlin-linkedin, line 171: please add test cases to integration test as well, since integration test will run the real mysql database query.

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
```diff
@@ -164,6 +167,50 @@ public void testUpdateBlobStateToReady() throws Exception {
...
+            namedBlobDb.put(record, NamedBlobState.READY, true).get();
+          } catch (Exception e) {
+            throw new RuntimeException(e);
+          }
+        }
+    );
+
+    // List named blob should only put out valid ones without empty entries.
+    Page<NamedBlobRecord> page =  namedBlobDb.list(account.getName(), container.getName(), blobNamePrefix, null).get();
+    assertEquals("List named blob entries size should match", validNames.size(), page.getEntries().size());
```
justinlin-linkedin, line 210: we can't get compare the size, we also have to compare the content of these two lists.

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
mikeyang923: Added the actual test in the MySqlNamedBlobDbIntegrationTest

ambry-named-mysql/src/test/java/com/github/ambry/named/MySqlNamedBlobDbTest.java
mikeyang923: Right, I find out the Unit test db query is just a mock, and it will return all rows, each row will have null for all fields (including NextPageToken). So it's not meaningful test there, I just moved to do the real test in integration test

justinlin-linkedin (approved):

PR #2536, opened 2023-08-11
Fix the bug to not set response channel for delete out of retention request

SophieGuo410:
Since we don't care about if delete out of retention request need to be done after put request, this pr is to make sure when we submit response to client, we don't set extra info for the response channel.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
@@ -32,6 +32,7 @@ import com.github.ambry.named.NamedBlobRecord;
 import com.github.ambry.protocol.DatasetVersionState;
 import com.github.ambry.quota.QuotaManager;
 import com.github.ambry.quota.QuotaUtils;
+import com.github.ambry.rest.NoOpResponseChannel;
 import com.github.ambry.rest.RequestPath;
 import com.github.ambry.rest.RestMethod;
 import com.github.ambry.rest.RestRequest;
@@ -155,6 +156,7 @@ public class NamedBlobPutHandler {
     private final RestResponseChannel restResponseChannel;
     private final Callback<Void> finalCallback;
     private final Callback<Void> deleteDatasetCallback;
+    private final Callback<Void> setRestMethodBackToPutCallback;
     private final String uri;
 
     /**
@@ -168,6 +170,7 @@ public class NamedBlobPutHandler {
       this.restResponseChannel = restResponseChannel;
       this.finalCallback = finalCallback;
       this.deleteDatasetCallback = deleteDatasetVersionIfUploadFailedCallBack(finalCallback);
+      this.setRestMethodBackToPutCallback = setRestMethodBackToPutCallback(finalCallback);
       this.uri = restRequest.getUri();
     }
 
@@ -287,10 +290,13 @@ public class NamedBlobPutHandler {
               routerTtlUpdateCallback(blobInfo, blobId));
         } else {
           if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
-            updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount();
+            //We have to make sure the restMethod been set back to PUT before we process response.
+            updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount(
+                deleteDatasetVersionOutOfRetentionCallback(blobInfo));
+          } else {
+            securityService.processResponse(restRequest, restResponseChannel, blobInfo,
+                securityProcessResponseCallback());
           }
-          securityService.processResponse(restRequest, restResponseChannel, blobInfo,
-              securityProcessResponseCallback());
         }
       }, uri, LOGGER, deleteDatasetCallback);
     }
@@ -325,10 +331,25 @@ public class NamedBlobPutHandler {
             namedBlobPath.getBlobName(), blobIdClean, Utils.Infinite_Time, namedBlobVersion);
         namedBlobDb.updateBlobStateToReady(record).get();
         if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
-          updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount();
+          //We have to make sure the restMethod been set back to PUT before we process response.
+          updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount(
+              deleteDatasetVersionOutOfRetentionCallback(blobInfo));
+        } else {
+          securityService.processResponse(restRequest, restResponseChannel, blobInfo,
+              securityProcessResponseCallback());
         }
-        securityService.processResponse(restRequest, restResponseChannel, blobInfo, securityProcessResponseCallback());
       }, uri, LOGGER, deleteDatasetCallback);
+    }
+
+    /**
+     * After updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount, call {@link SecurityService#processResponse}.
+     * @param blobInfo the {@link BlobInfo} to use for security checks.
+     * @return a {@link Callback} to be used with updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount.
+     */
+    private Callback<Void> deleteDatasetVersionOutOfRetentionCallback(BlobInfo blobInfo) {
+      return buildCallback(frontendMetrics.deleteDatasetOutOfRetentionRequestMetrics,
+          deleteResult -> securityService.processResponse(restRequest, restResponseChannel, blobInfo,
+              securityProcessResponseCallback()), restRequest.getUri(), LOGGER, setRestMethodBackToPutCallback);
     }
 
     /**
@@ -602,7 +623,7 @@ public class NamedBlobPutHandler {
     /**
      * Support delete the dataset version out of retentionCount
      */
-    private void updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount() throws RestServiceException {
+    private void updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount(Callback<Void> callback) throws RestServiceException {
       Dataset dataset = (Dataset) restRequest.getArgs().get(RestUtils.InternalKeys.TARGET_DATASET);
       String accountName = dataset.getAccountName();
       String containerName = dataset.getContainerName();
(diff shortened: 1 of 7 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -705,5 +733,19 @@ private <T> Callback<T> deleteDatasetVersionIfUploadFailedCallBack(Callback<T> c
...
     }
+
+    /**
+     * If updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount failed, this method is to help set the rest method
+     * back to PUT.
+     * @param callback callback the final callback which submit the response.
+     */
+    private <T> Callback<T> setRestMethodBackToPutCallback(Callback<T> callback) {
+      return (r, e) -> {
+        restRequest.setRestMethod(RestMethod.PUT);
```
justinlin-linkedin, line 744: can we create new implementation of RestRequest that takes in the `restRequest` when it needs to fetch headers, but return DELETE when it returns RestMethod. This way we would have a new object for delete handler so we don't have to reset the RestMethod.

SophieGuo410: Thanks! Updated to use the WrappedRestRequest which always return DELETE restMethod.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -632,18 +650,23 @@ private void updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount() thro
...
           LOGGER.trace("New request path : " + newRequestPath);
-          // Replace RequestPath in the RestRequest and call DeleteBlobHandler.handle.
-          restRequest.setArg(InternalKeys.REQUEST_PATH, newRequestPath);
-          restRequest.setArg(InternalKeys.TARGET_ACCOUNT_KEY, null);
-          restRequest.setArg(InternalKeys.TARGET_CONTAINER_KEY, null);
-          //set the rest method to delete in order to delete data in named blob.
-          restRequest.setRestMethod(RestMethod.DELETE);
-          deleteBlobHandler.handle(restRequest, restResponseChannel,
-              recursiveCallback(datasetVersionRecordList, 1, accountName, containerName, datasetName));
+          // Replace RequestPath in the WrappedRestRequest for delete and call DeleteBlobHandler.handle.
```
justinlin-linkedin, line 653: BTW, in line 643, could you also print out the account name, container name and the dataset name?

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -654,34 +677,37 @@ private void updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount() thro
...
+                    NAMED_BLOB_PREFIX + SLASH + accountName + SLASH + containerName + SLASH + datasetName + SLASH + version,
+                    requestPath.getSubResource(), requestPath.getBlobSegmentIdx());
+            LOGGER.debug("New request path in recursive call : " + newRequestPath);
+            // Replace RequestPath in the RestRequest and call DeleteBlobHandler.handle.
+            wrappedRestRequest.setArg(InternalKeys.REQUEST_PATH, newRequestPath);
+            wrappedRestRequest.setArg(InternalKeys.TARGET_ACCOUNT_KEY, null);
+            wrappedRestRequest.setArg(InternalKeys.TARGET_CONTAINER_KEY, null);
+            deleteBlobHandler.handle(wrappedRestRequest, restResponseChannel, nextCallBack);
+          }, uri, LOGGER,
+          (r, e) -> LOGGER.error("Delete dataset" + wrappedRestRequest.getUri() + " failed with exception" + e));
```
justinlin-linkedin, line 710: this should not just be a logger, this callback will be invoked when the delete failed, we should still call processResponseCallback.onComplete to return 201 back to client.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -654,34 +677,37 @@ private void updateVersionStateAndDeleteDatasetVersionOutOfRetentionCount() thro
...
-          LOGGER.debug("Change rest method back to PUT");
-          restRequest.setRestMethod(RestMethod.PUT);
+          //make sure to process response delete is done.
+          processResponseCallback.onCompletion(null, e);
         };
       }
       Callback<Void> nextCallBack =
-          recursiveCallback(datasetVersionRecordList, idx + 1, accountName, containerName, datasetName);
+          recursiveCallback(datasetVersionRecordList, idx + 1, accountName, containerName, datasetName,
+              processResponseCallback, null);
```
justinlin-linkedin, line 694: wrapped request shouldn't be null in this function call.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
justinlin-linkedin: can you please add a test when the delete handler fails, client would still get a success response.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
SophieGuo410: Sure, Updated.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
SophieGuo410: Thanks for catching that, fixed it and the updated test should cover this case.

justinlin-linkedin (approved):
LGTM, just one more question, can we reuse the code in recursiveCallback, to something like this.
```
        List<DatasetVersionRecord> datasetVersionRecordList =
            accountService.getAllValidVersionsOutOfRetentionCount(accountName, containerName, datasetName);
        if (datasetVersionRecordList.size() > 0) {
          WrappedRestRequest wrappedRestRequest = new WrappedRestRequest(restRequest);
          LOGGER.info(
              "Number of records need to be deleted due to out of retention count: " + datasetVersionRecordList.size()
                  + " for accountName: " + accountName + " containerName: " + containerName + " datasetName: "
                  + datasetName);
          recursiveDeleteDatasetVersions(wrappedRestRequest, datasetVersionRecordList, 0, callback);
        } else {
          callback.onCompletion(null, null);
        }

// This is the recursive call
    private void recursiveDeleteDatasetVersions(RestRequest restRequest,
        List<DatasetVersionRecord> datasetVersionRecordList, int idx, Callback<Void> callback) throws Exception {
      if (idx == datasetVersionRecordList.size()) {
        callback.onCompletion(null, null);
        return;
      }
      Dataset dataset = (Dataset) restRequest.getArgs().get(RestUtils.InternalKeys.TARGET_DATASET);
      String accountName = dataset.getAccountName();
      String containerName = dataset.getContainerName();
      String datasetName = dataset.getDatasetName();
      DatasetVersionRecord record = datasetVersionRecordList.get(idx);
      String version = record.getVersion();
      RequestPath requestPath = getRequestPath(restRequest);
      RequestPath newRequestPath =
          new RequestPath(requestPath.getPrefix(), requestPath.getClusterName(), requestPath.getPathAfterPrefixes(),
              NAMED_BLOB_PREFIX + SLASH + accountName + SLASH + containerName + SLASH + datasetName + SLASH + version,
              requestPath.getSubResource(), requestPath.getBlobSegmentIdx());
      LOGGER.debug("New request path : " + newRequestPath);
      // Replace RequestPath in the WrappedRestRequest for delete and call DeleteBlobHandler.handle.
      restRequest.setArg(InternalKeys.REQUEST_PATH, newRequestPath);
      restRequest.setArg(InternalKeys.TARGET_ACCOUNT_KEY, null);
      restRequest.setArg(InternalKeys.TARGET_CONTAINER_KEY, null);
      //for delete out of retention request, we don't want to set anything to response channel.
      deleteBlobHandler.handle(restRequest, new NoOpResponseChannel(), (r, e) -> {
        try {
          recursiveDeleteDatasetVersions(restRequest, datasetVersionRecordList, idx + 1, callback);
        } catch (Exception exc) {
          callback.onCompletion(null, null);
        }
      });
    }
```

SophieGuo410:
> LGTM, just one more question, can we reuse the code in recursiveCallback, to something like this.
> 
> ```
>         List<DatasetVersionRecord> datasetVersionRecordList =
>             accountService.getAllValidVersionsOutOfRetentionCount(accountName, containerName, datasetName);
>         if (datasetVersionRecordList.size() > 0) {
>           WrappedRestRequest wrappedRestRequest = new WrappedRestRequest(restRequest);
>           LOGGER.info(
>               "Number of records need to be deleted due to out of retention count: " + datasetVersionRecordList.size()
>                   + " for accountName: " + accountName + " containerName: " + containerName + " datasetName: "
>                   + datasetName);
>           recursiveDeleteDatasetVersions(wrappedRestRequest, datasetVersionRecordList, 0, callback);
>         } else {
>           callback.onCompletion(null, null);
>         }
> 
> // This is the recursive call
>     private void recursiveDeleteDatasetVersions(RestRequest restRequest,
>         List<DatasetVersionRecord> datasetVersionRecordList, int idx, Callback<Void> callback) throws Exception {
>       if (idx == datasetVersionRecordList.size()) {
>         callback.onCompletion(null, null);
>         return;
>       }
>       Dataset dataset = (Dataset) restRequest.getArgs().get(RestUtils.InternalKeys.TARGET_DATASET);
>       String accountName = dataset.getAccountName();
>       String containerName = dataset.getContainerName();
>       String datasetName = dataset.getDatasetName();
>       DatasetVersionRecord record = datasetVersionRecordList.get(idx);
>       String version = record.getVersion();
>       RequestPath requestPath = getRequestPath(restRequest);
>       RequestPath newRequestPath =
>           new RequestPath(requestPath.getPrefix(), requestPath.getClusterName(), requestPath.getPathAfterPrefixes(),
>               NAMED_BLOB_PREFIX + SLASH + accountName + SLASH + containerName + SLASH + datasetName + SLASH + version,
>               requestPath.getSubResource(), requestPath.getBlobSegmentIdx());
>       LOGGER.debug("New request path : " + newRequestPath);
>       // Replace RequestPath in the WrappedRestRequest for delete and call DeleteBlobHandler.handle.
>       restRequest.setArg(InternalKeys.REQUEST_PATH, newRequestPath);
>       restRequest.setArg(InternalKeys.TARGET_ACCOUNT_KEY, null);
>       restRequest.setArg(InternalKeys.TARGET_CONTAINER_KEY, null);
>       //for delete out of retention request, we don't want to set anything to response channel.
>       deleteBlobHandler.handle(restRequest, new NoOpResponseChannel(), (r, e) -> {
>         try {
>           recursiveDeleteDatasetVersions(restRequest, datasetVersionRecordList, idx + 1, callback);
>         } catch (Exception exc) {
>           callback.onCompletion(null, null);
>         }
>       });
>     }
> ```

Yes, I think that one is more simplify. I just use the way you recommended and adding some logs and metrics when delete failed.

PR #1679, opened 2020-10-29
[AZURE_AD_AUTH] Implementation of Ad based authentication for azure storage.

ankagrawal:
This PR adds the functionality of AD based authentication for Blob Storage in ambry-cloud. The current authentication method using connection string is unsafe because connection string is visible to anyone who is able to login the account. With AD based authentication, permissions can be assigned to make sure that the client secret is not visible to everyone.

To add this functionality, a class `StorageClientFactory` provides the base implementation for `Azure BlobServiceClient`. The client is built in one of `ConnectionStringBasedStorageClientFactory` or `ADAuthBasedStorageClientFactory` classes based on config.

PR #2537, opened 2023-08-14
Azure blob storage classes

snalli:
This patch introduces foundational classes to implement Backup 2.0 that are identical counterparts of objects in Backup 1.0 and preserve the same relationships. We create new classes to avoid meddling with old classes and obfuscating code which makes it harder to review and maintain.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -117,7 +117,7 @@ public class CloudConfig {
   /**
    * Azure backup version
    */
-  public static final String AMBRY_BACKUP_VERSION = "azure.backup.version";
+  public static final String AMBRY_BACKUP_VERSION = "ambry.backup.version";
   public static final String AMBRY_BACKUP_VERSION_1 = "1.0";
   public static final String AMBRY_BACKUP_VERSION_2 = "2.0";
   public static final String DEFAULT_AMBRY_BACKUP_VERSION = AMBRY_BACKUP_VERSION_1;
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -115,6 +115,16 @@ public class CloudConfig {
   public static final String VCR_HELIX_CONFIG_READY = "vcrHelixConfigReady";
 
   /**
+   * Azure backup version
+   */
+  public static final String AZURE_BACKUP_VERSION = "azure.backup.version";
+  public static final String AZURE_BACKUP_VERSION_1 = "1.0";
+  public static final String AZURE_BACKUP_VERSION_2 = "2.0";
+  public static final String DEFAULT_AZURE_BACKUP_VERSION = AZURE_BACKUP_VERSION_1;
+  @Config(AZURE_BACKUP_VERSION)
+  public final String azureBackupVersion;
+
+  /**
    * True for VCR node, false for live serving node.
    */
   @Config(CLOUD_IS_VCR)
@@ -429,6 +439,7 @@ public class CloudConfig {
 
   public CloudConfig(VerifiableProperties verifiableProperties) {
 
+    azureBackupVersion = verifiableProperties.getString(AZURE_BACKUP_VERSION, DEFAULT_AZURE_BACKUP_VERSION);
     cloudIsVcr = verifiableProperties.getBoolean(CLOUD_IS_VCR, false);
     cloudDestinationFactoryClass =
         verifiableProperties.getString(CLOUD_DESTINATION_FACTORY_CLASS, DEFAULT_CLOUD_DESTINATION_FACTORY_CLASS);
--- a/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
@@ -117,12 +117,12 @@ public class CloudConfig {
   /**
    * Azure backup version
    */
-  public static final String AZURE_BACKUP_VERSION = "azure.backup.version";
-  public static final String AZURE_BACKUP_VERSION_1 = "1.0";
-  public static final String AZURE_BACKUP_VERSION_2 = "2.0";
-  public static final String DEFAULT_AZURE_BACKUP_VERSION = AZURE_BACKUP_VERSION_1;
-  @Config(AZURE_BACKUP_VERSION)
-  public final String azureBackupVersion;
+  public static final String AMBRY_BACKUP_VERSION = "azure.backup.version";
+  public static final String AMBRY_BACKUP_VERSION_1 = "1.0";
+  public static final String AMBRY_BACKUP_VERSION_2 = "2.0";
+  public static final String DEFAULT_AMBRY_BACKUP_VERSION = AMBRY_BACKUP_VERSION_1;
+  @Config(AMBRY_BACKUP_VERSION)
+  public final String ambryBackupVersion;
 
   /**
    * True for VCR node, false for live serving node.
@@ -439,7 +439,7 @@ public class CloudConfig {
 
   public CloudConfig(VerifiableProperties verifiableProperties) {
 
-    azureBackupVersion = verifiableProperties.getString(AZURE_BACKUP_VERSION, DEFAULT_AZURE_BACKUP_VERSION);
+    ambryBackupVersion = verifiableProperties.getString(AMBRY_BACKUP_VERSION, DEFAULT_AMBRY_BACKUP_VERSION);
     cloudIsVcr = verifiableProperties.getBoolean(CLOUD_IS_VCR, false);
     cloudDestinationFactoryClass =
         verifiableProperties.getString(CLOUD_DESTINATION_FACTORY_CLASS, DEFAULT_CLOUD_DESTINATION_FACTORY_CLASS);
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorage.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorage.java
@@ -1,0 +1,194 @@
+/**
+ * Copyright 2023 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.github.ambry.cloud;
+
+import com.azure.storage.blob.BlobContainerClient;
+import com.codahale.metrics.MetricRegistry;
+import com.github.ambry.cloud.azure.AzureCloudConfig;
+import com.github.ambry.cloud.azure.AzureMetrics;
+import com.github.ambry.clustermap.ClusterMap;
+import com.github.ambry.clustermap.ReplicaState;
+import com.github.ambry.config.CloudConfig;
+import com.github.ambry.config.VerifiableProperties;
+import com.github.ambry.replication.FindToken;
+import com.github.ambry.store.FindInfo;
+import com.github.ambry.store.MessageInfo;
+import com.github.ambry.store.MessageWriteSet;
+import com.github.ambry.store.Store;
+import com.github.ambry.store.StoreException;
+import com.github.ambry.store.StoreGetOptions;
+import com.github.ambry.store.StoreInfo;
+import com.github.ambry.store.StoreKey;
+import com.github.ambry.store.StoreStats;
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Implements methods to write {@link com.github.ambry.clustermap.AmbryPartition} in Azure blob storage.
+ */
+public class AzureStorage implements Store {
+
+  private final ClusterMap clusterMap;
+  private final AzureMetrics azureMetrics;
+  private final CloudConfig cloudConfig;
+  private final AzureCloudConfig azureCloudConfig;
+  private final Object vcrMetrics;
+  private final BlobContainerClient blobContainerClient;
+  private static final Logger logger = LoggerFactory.getLogger(AzureStorage.class);
+
+  public AzureStorage(VerifiableProperties properties, MetricRegistry metricRegistry, ClusterMap clusterMap,
+      BlobContainerClient blobContainerClient) {
+    this.azureMetrics = new AzureMetrics(metricRegistry);
+    this.cloudConfig = new CloudConfig(properties);
+    this.clusterMap = clusterMap;
+    this.azureCloudConfig = new AzureCloudConfig(properties);
+    this.vcrMetrics = new VcrMetrics(metricRegistry);
+    this.blobContainerClient = blobContainerClient;
+  }
+
+  @Override
+  public void put(MessageWriteSet messageSetToWrite) throws StoreException {
+    // TODO
+  }
+
+  @Override
+  public void delete(List<MessageInfo> messageInfos) throws StoreException {
+    // TODO
+  }
+
+  @Override
+  public short undelete(MessageInfo messageInfo) throws StoreException {
+    // TODO
+    return 0;
+  }
+
+  @Override
+  public void updateTtl(List<MessageInfo> messageInfos) {
+    // TODO
+  }
+
+  @Override
+  public Set<StoreKey> findMissingKeys(List<StoreKey> storeKeys) {
+    // TODO
+    return new HashSet<>();
+  }
+
+  @Override
+  public MessageInfo findKey(StoreKey storeKey) {
+    // TODO
+    return null;
+  }
+
+  @Override
+  public boolean isStarted() {
+    return true;
+  }
+
+  @Override
+  public boolean isEmpty() {
+    // TODO
+    return false;
+  }
+
+  @Override
+  public ReplicaState getCurrentState() {
+    // TODO
+    return null;
+  }
+
+  @Override
+  public void setCurrentState(ReplicaState state) {
+    // TODO
+  }
+
+  @Override
+  public long getSizeInBytes() {
+    // TODO
+    return 0;
+  }
+
+  @Override
+  public long getEndPositionOfLastPut() {
+    // TODO
+    return 0;
+  }
+
+  /////////////////////////////////////////// Unimplemented methods /////////////////////////////////////////////////
+
+  @Override
+  public void start() throws StoreException {
+
+  }
+
+  @Override
+  public StoreInfo get(List<? extends StoreKey> ids, EnumSet<StoreGetOptions> storeGetOptions) throws StoreException {
+    return null;
+  }
+
+  @Override
+  public void forceDelete(List<MessageInfo> infosToDelete) {
+
+  }
+
+  @Override
+  public FindInfo findEntriesSince(FindToken token, long maxTotalSizeOfEntries, String hostname, String remoteReplicaPath) throws StoreException {
+    return null;
+  }
+
+  @Override
+  public StoreStats getStoreStats() {
+    return null;
+  }
+
+  @Override
+  public boolean isDisabled() {
+    return false;
+  }
+
+  @Override
+  public boolean isKeyDeleted(StoreKey key) {
+    return false;
+  }
+
+  @Override
+  public boolean isBootstrapInProgress() {
+    return false;
+  }
+
+  @Override
+  public boolean isDecommissionInProgress() {
+    return false;
+  }
+
+  @Override
+  public void completeBootstrap() {
+
+  }
+
+  @Override
+  public boolean recoverFromDecommission() {
+    return false;
+  }
+
+  @Override
+  public void shutdown() throws StoreException {
+
+  }
+}
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageHelixParticipant.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageHelixParticipant.java
@@ -1,0 +1,49 @@
+/**
+ * Copyright 2023 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.github.ambry.cloud;
+
+import com.codahale.metrics.MetricRegistry;
+import com.github.ambry.account.AccountService;
+import com.github.ambry.clustermap.ClusterMap;
+import com.github.ambry.config.CloudConfig;
+import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.config.StoreConfig;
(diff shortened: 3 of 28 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorage.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.List;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Implements methods to write {@link com.github.ambry.clustermap.AmbryPartition} in Azure blob storage.
+ */
+public class AzureStorage implements Store {
```
JingQianCloud, line 46: For the class naming, isn't AzureBlobStore better than AzureStorage?
We are implementing the BlobStore with the Azure Storage. 

And AzureBlobStore matches with the v1 cloudBlobStore as well.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageHelixParticipant.java
```diff
@@ -0,0 +1,49 @@
...
+import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.config.StoreConfig;
+import com.github.ambry.config.VerifiableProperties;
+import org.apache.helix.participant.StateMachineEngine;
+
+
+/**
+ * Helix participant
+ */
+public class AzureStorageHelixParticipant extends HelixVcrClusterParticipant {
```
JingQianCloud, line 30: For the Helix Participant, do we need do any change?
What's the problem with the existing HelixVcrClusterParticipant?

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageManager.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Storage manager that tracks {@link AzureStorage} objects.
+ */
+public class AzureStorageManager implements StoreManager {
```
JingQianCloud, line 45: I think we need the backup implementation change, but for the higher level class like CloudStorageManager, the code logic should be the same. 
Do we need create this one?

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageReplicationManager.java
```diff
@@ -0,0 +1,95 @@
...
+import com.github.ambry.utils.Time;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Predicate;
+
+
+/**
+ * Manages replication from servers to VCR
+ */
+public class AzureStorageReplicationManager extends VcrReplicationManager {
```
JingQianCloud, line 52: Any new method we need for the AzureStorageReplicatonManager, other than the methods from the parent class?

Same question for AzureStorageReplicationThread. Although we have V1 and V2, if the functionality is the same for some of the high level classes, we don't need create the duplicated classes.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageTokenWriter.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Map;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Implements methods to persist {@link com.github.ambry.replication.RemoteReplicaInfo.ReplicaTokenInfo} to Azure
+ */
+public class AzureStorageTokenWriter extends ReplicaTokenPersistor {
```
JingQianCloud, line 38: What's our conclusion for the token? Do we use the existing abs for token or cosmos to save token?

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
```diff
@@ -189,9 +186,7 @@ public void participate() throws Exception {
         cloudConfig.vcrClusterZkConnectString);
     VcrStateModelFactory stateModelFactory = Utils.getObj(cloudConfig.vcrHelixStateModelFactoryClass, this);
     manager.getStateMachineEngine().registerStateModelFactory(stateModelFactory.getStateModelName(), stateModelFactory);
-    if (cloudConfig.cloudContainerCompactionEnabled) {
-      registerContainerDeletionSyncTask(manager.getStateMachineEngine());
-    }
+    registerContainerDeletionSyncTask(manager.getStateMachineEngine());
```
JingQianCloud, line 189: What's the reason that we don't need check the "cloudContainerCompactionEnable" flag?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
```diff
@@ -382,6 +410,22 @@ protected BlobServiceAsyncClient createBlobStorageAsyncClient() {
     }
   }
 
+  protected BlobServiceClient createBlobStorageSyncClient() {
```
JingQianCloud, line 413: Why do we need sync client? Is this for testing purpose?

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorage.java
snalli: I prefer that the names of all related classes begin with the same prefix, for eg AzureStorage in this case. V1 code is unreadable and readability for others is my goal in v2. It is easy to search for related classes using a common prefix with grep or IDE or any internal code browser. I can call it AzureStorageBlobStore if you want but I am not interested in following v1.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageHelixParticipant.java
snalli: Readability and Operability.

There are several problems with using HelixVcrClusterParticipant for V2. 

1. Creating new counterpart classes enables switching from V1 to V2 and back easily, using _exactly_ one config variable which is azureBackupVersion. Otherwise, there are 4-5 config variables you'll have to remember and change in order to switch. cloudContainerCompactionEnabled is one among those configs when we use V1 helix participant. Furthermore, V2 will first be deployed in a single cluster while others remain on V1. Additionally, switching a V2 cluster back to V1 in case of any issues would be a no-brainer if there is exactly one config. A single config minimizes operator errors during such deployments.

2. Secondly, registerContainerDeletionSyncTask needs to be over-written. The backup format changes from V1 to V2 and V1 clean-up tasks will simply not work. Over-writing it completely is cleaner than meddling with existing implementation.

3. Finally, creating new classes does not harm performance, reliability or availability. It doesn't even harm readability in this case. So if there are other concerns, I don't understand them.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageManager.java
snalli: Readability and better code.

We need an AzureStorageManager for following reasons.

1. AzureStorageManager creates ABS containers, which CloudStorageManager does not. We could modify it using some configs and if-else statements but as I have mentioned above, readability is an important goal for V2. Meddling with CloudStorageManager creates unreadable code, may break something or may make it harder to review slowing down the overall project. I would just like V1 code to run as it is separate from V2. 

2. CloudStorageManager tracks CloudBlobStore objects in a map<partitionId, CloudBlobStore>. V2 creates AzureStorage objects by implementing the Store interface. I could change the declaration of the map to <partitionId, Store> but as mentioned above, readability and operability are important goals for V2. 

3. CloudStorageManager is sub-optimal. It uses its own locking mechanism instead of using a ConcurrentHashMap, which is what AzureStorageManager does.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageReplicationManager.java
snalli: Yes and yes.

AzureStorageReplicationManager over writes getReplicaThread to return AzureStorageReplicationThread. Going forward, it will schedule a new type of thread for compaction.

AzureStorageReplicationThread over writes advanceToken and simplifies token persistence. It brings it to the foreground and makes debugging a lot more easier. Otherwise, the token persistence thread runs somewhere in the background and doesn't persist tokens as expected. I expect the token to be persisted after each batch of replication messages deterministically and would like to see it advance in the web-portal during a live debugging session.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageTokenWriter.java
snalli: I will discuss this when we get to that PR but my focus now is to get this one out.

ambry-cloud/src/main/java/com/github/ambry/cloud/HelixVcrClusterParticipant.java
snalli: This check here is redundant. It is being checked below registerContainerDeletionSyncTask.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
snalli: This is not for testing. V2 will use syncClient. There are several issues with the AsyncClient.

1. Async is a good principle in general but we need to evaluate it in each context before applying it. AsyncClient offers absolutely ZERO benefits for V2. V1 is a readability and debugging nightmare because of AsyncClient. Please read the V1 code and/or talk to Arun. 
2. AsyncClient was implemented in V1 because VCR was expected to server customer traffic. That is simply not the case anymore. 
3. Replication does not proceed to fetch a blob from server without copying the previous one out to Azure. So AsyncClient simply doesn't make sense.
4. AsyncClient was implemented to override some of the handlers post-completion and do some operations on cosmosDB including storing metadata. Not the case anymore. 
5. SyncClient is a wrapper over AsyncClient in the library. They complete the future returned by AsyncClient to implement the syncClient. 
6. I am willing to sacrifice any performance improvements that AsyncClient may offer for the sake of readability, speed of development and ease of debugging that syncClient offers. 
7. We can revisit this if the performance of syncClient is unacceptable.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorage.java
snalli: Done. Renamed all classes with prefix AzureStore

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
```diff
@@ -0,0 +1,231 @@
...
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Storage manager that tracks {@link AzureStore} objects.
+ */
+public class AzureStoreManager implements StoreManager {
```
Arun-LinkedIn, line 45: Same as earlier comment in AzureStore class. May be we could rename the classes under `/ambry/cloud` to `{oldName}V2` such `CloudStorageManagerV2` and have Azure names for classes only under `/ambry/cloud/azure`?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -82,23 +82,16 @@ public class VcrReplicationManager extends ReplicationEngine {
...
+      NotificationSystem requestNotification, StoreKeyConverterFactory storeKeyConverterFactory,
+      String transformerClassName) throws ReplicationException, IllegalStateException {
     super(replicationConfig, clusterMapConfig, storeConfig, storeKeyFactory, clusterMap, scheduler,
         vcrClusterParticipant.getCurrentDataNodeId(), Collections.emptyList(), networkClientFactory,
         vcrMetrics.getMetricRegistry(), requestNotification, storeKeyConverterFactory, transformerClassName, null,
         storeManager, null, true);
     this.cloudConfig = cloudConfig;
     this.vcrClusterParticipant = vcrClusterParticipant;
     this.vcrMetrics = vcrMetrics;
-    this.persistor =
```
snalli, line 95: Refactoring to create two ctors, one of which is used for this project

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
```diff
@@ -165,23 +167,11 @@ public void startup() throws InstantiationException {
       // verify the configs
       properties.verify();
 
-      // initialize cloud destination
```
snalli, line 168: Refactoring to bring all V1 code in the if-case below.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
```diff
@@ -0,0 +1,231 @@
...
+  public AzureStoreManager(VerifiableProperties properties, MetricRegistry metricRegistry, ClusterMap clusterMap) {
+    this.partitionToAzureStore = new ConcurrentHashMap<>();
+    this.metricRegistry = metricRegistry;
+    this.azureMetrics = new AzureMetrics(metricRegistry);
+    this.cloudConfig = new CloudConfig(properties);
+    this.azureCloudConfig = new AzureCloudConfig(properties);
+    this.properties = properties;
+    this.clusterMap = clusterMap;
+    this.azureStorageClient =
+        new ConnectionStringBasedStorageClient(cloudConfig, azureCloudConfig, azureMetrics).getStorageSyncClient();
```
Arun-LinkedIn, line 73: Are you planning to make it configurable in later PRs since we are not using connection string based authentication in production because it is unsafe ([Reference PR](https://github.com/linkedin/ambry/pull/1679)). We are using either ADAuth based or ClientSecretCredential based.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageTokenWriter.java
snalli: I found that ABS has a table offering similar to cosmos 
https://learn.microsoft.com/en-us/azure/cosmos-db/table/support?toc=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fazure%2Fstorage%2Ftables%2Ftoc.json&bc=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fazure%2Fbread%2Ftoc.json

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorage.java
JingQianCloud: You probably misunderstood my meaning. I don't have concern with AzureXXXX.
For this class itself, "AzureStorage" means interface to provide Azure Storage APIs. But this class is specifically to implement BlobStore on Azure. So the AzureBlobStore naming is better. And the previous version uses CloudBlobStore which matches the interface as well.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
```diff
@@ -0,0 +1,237 @@
...
+
+  /**
+   * Returns blob container in Azure for an Ambry partition
+   * @param partitionId Ambry partition
+   * @return blobContainerClient
+   */
+  protected BlobContainerClient getBlobStore(PartitionId partitionId) {
+    BlobContainerClient blobContainerClient;
+    try {
+      blobContainerClient = azureStorageClient.getBlobContainerClient(String.valueOf(partitionId.getId()));
```
Arun-LinkedIn, line 97: Would it be possible to put this Azure container creation logic inside AzureBlobStore (or if possible inside a new class called AzureCloudDestinationV2) similar to how we do today? 

I think today we have:
CloudStorageManager -> CloudBlobStore (contains Ambry logic like retries, encryption, etc) -> CloudDestination / AzureCloudDestination (contains Azure exception handling, retries, etc) -> AzureStorageClient (which uses azure sdk client). 

May be we could have similar to that as below so that Azure interactions are only in  :
CloudStorageManagerV2 -> CloudBlobStoreV2 -> CloudDestination / AzureCloudDestinationV2 -> AzureStorageClient (which uses azure sdk client. (or)

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
```diff
@@ -0,0 +1,237 @@
...
+    }
+    return blobContainerClient;
+  }
+
+  /**
+   * Creates blob container in Azure for an Ambry partition
+   * @param partitionId Ambry partition
+   * @return blobContainerClient
+   */
+  protected BlobContainerClient createBlobStore(PartitionId partitionId) {
```
Arun-LinkedIn, line 117: Why protected?

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -114,6 +114,16 @@ public class CloudConfig {
...
 
+  /**
+   * Azure backup version
+   */
+  public static final String AZURE_BACKUP_VERSION = "azure.backup.version";
+  public static final String AZURE_BACKUP_VERSION_1 = "1.0";
+  public static final String AZURE_BACKUP_VERSION_2 = "2.0";
+  public static final String DEFAULT_AZURE_BACKUP_VERSION = AZURE_BACKUP_VERSION_1;
+  @Config(AZURE_BACKUP_VERSION)
+  public final String azureBackupVersion;
```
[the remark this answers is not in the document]
snalli: done

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageHelixParticipant.java
JingQianCloud: Item2 is a good reason for me. Yes, we can use the new class.

But,
item 1: If we use template and provide the corresponding <T>, it can solve problem 1.
item 3: Although creating new class won't harm performance, we don't help with readability since have more code to deal with.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ConnectionStringBasedStorageClient.java
```diff
@@ -62,14 +73,23 @@ public ConnectionStringBasedStorageClient(BlobServiceAsyncClient blobServiceAsyn
...
-            : azureCloudConfig.azureStorageConnectionString)
+            storageAccountInfo() != null ? storageAccountInfo().getStorageConnectionString()
+                : azureCloudConfig.azureStorageConnectionString)
         .httpClient(httpClient)
         .retryOptions(retryOptions)
         .configuration(configuration)
         .buildAsyncClient();
   }
 
+  protected BlobServiceClient buildBlobServiceSyncClient(HttpClient httpClient, Configuration configuration,
```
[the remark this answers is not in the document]
snalli: I will override the method for whateverr class is being used.  Are you sure we are using `ClientSecretCredentialStorageClient` ? vcr-video is using `ConnectionStringBasedStorageClient`.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageManager.java
JingQianCloud: Why doesn't CloudStorageManager create containers? Or do you mean we are creating different containers between v1 and v2?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/StorageClient.java
JingQianCloud: @Arun-LinkedIn Arun, do you have any concern here? I know we have put quite some effort to change the cloud implement from sync mode to async way.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStore.java
```diff
@@ -0,0 +1,194 @@
...
+import java.util.List;
+import java.util.Set;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+/**
+ * Implements methods to write {@link com.github.ambry.clustermap.AmbryPartition} in Azure blob storage.
+ */
+public class AzureStore implements Store {
```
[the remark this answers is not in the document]
snalli: This class will directly talk to Azure and use Azure APIs. In V2, there is no more CloudDestination and I highly doubt we will ever use S3 or GCP. If that happens, then we can refactor the code but I don't want to create those destination classes and objects at the expense of current OKR in the anticipation of an extremely rare possibility.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
snalli: Yeah, if that is truly the case, I will make it configurable or just directly use the client we plan on using. But currently we are using connection-based client in vcr-video.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
snalli: I do not plan on exposing this method. The public API is addBlobStore (called by addReplica) and getBlobStore.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
snalli: I highly doubt we will ever depart from Azure, so just using the name `Azure` directly is descriptive. Giving new classes a distinct prefix makes searching easier with grep or any code browser. For eg 
```
find . -name 'AzureStore'
```
will show up all the V2 code and not V1 code.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
snalli: Those destination objects in V2 don't really add any value. The code to write to Azure is a 20 lines at most and is best positioned in `AzureStore`

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -114,6 +114,16 @@ public class CloudConfig {
 
   public static final String VCR_HELIX_CONFIG_READY = "vcrHelixConfigReady";
 
+  /**
+   * Azure backup version
+   */
+  public static final String AMBRY_BACKUP_VERSION = "azure.backup.version";
```
JingQianCloud, line 120: azure.backup.version -> ambry.backup.version?

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStore.java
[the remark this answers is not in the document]
JingQianCloud: Thanks Arun for the explanation. I second Arun's opinion. 
1. The current architecture to separate the generic cloud vs Azure implement makes sense in the design point of view.
2. And since V1 and V2 will co-exist for quite some time, let's follow the previous naming conversion. Otherwise, the person who familiar with the V1 will get very confused.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStoreManager.java
JingQianCloud: If there is no specific reason, if possible, let's be compatible with V1.  
V1 and V2 will coexist for quite some time, it'll improve readability.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrServer.java
snalli: good catch !

JingQianCloud (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorage.java
snalli: got it, understood.

ambry-cloud/src/main/java/com/github/ambry/cloud/AzureStorageHelixParticipant.java
snalli: got it, understood.

snalli:
Answered comments. I have more PRs to follow, so if I left some out, we can address in the next one.

PR #2538, opened 2023-08-15
[Compaction] Add metric to capture time taken for each store's compaction.

ankagrawal:


JingQianCloud (approved):
LGTM

justinlin-linkedin (commented):
Do you want to add a metrics for cost as well? And maybe print out the cost for CompactAll? And BTW, CompactAll policy would mostly to skew this timer since we don't have many compactions and CompactAll is so much slower than the StatsBasedCompaction. Maybe we want to create yet another metrics for StatsBaseCompaction.

PR #2541, opened 2023-08-16
Deprioritize bootstrap replicas in get operations

justinlin-linkedin:
When a host or a replica is bootstrapping, it likely that this replica doesn't have the blob for get operation, in this way, we should deprioritize this replica for get operation.


ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -749,6 +751,9 @@ public class RouterConfig {
   @Default("false")
   public final boolean routerReservedMetadataEnabled;
 
+  @Config(ROUTER_GET_OPERATION_DEPRIORITIZE_BOOTSTRAP_REPLICAS)
```
ankagrawal, line 754: Do we really need the config. No harm in deprioritizing the bootstrap replica by default.

justinlin-linkedin: just want to have a layer of protection if this logic is somehow wrong.

ankagrawal (approved):

Arun-LinkedIn (approved):

PR #2542, opened 2023-08-17
Split the pr to support named blob ttl update only

SophieGuo410:


mikeyang923 (approved):
🚀

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -764,6 +764,30 @@ private void applySoftDelete(short accountId, short containerId, String blobName
     }
   }
 
+  @Override
+  public CompletableFuture<PutResult> ttlUpdate(NamedBlobRecord record, NamedBlobState state) {
```
justinlin-linkedin, line 768: can we have a dedicated sql statement for updating ttl. And there should be some limitation to when a named blob's ttl can be updated, like if there is an expiration date, it needs to be more than one day. We should share the same configuration with what is in the server nodes.

justinlin-linkedin (changes requested):

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
SophieGuo410: As we discussed offline, I think I can re-use this updateBlobStateToReady method, just change the name to updateBlobTtlAndStateToReady to be more clear. 
And in covertId logic, when operation is TTL_UPDATE, we just get the named blob for id and version.

`          conversionFuture = getNamedBlobDb().get(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
              namedBlobPath.getBlobName(), getOption).thenApply(result -> {
            restRequest.setArg(RestUtils.InternalKeys.NAMED_BLOB_VERSION, result.getInsertedRecord().getVersion());
            restRequest.setArg(RestUtils.InternalKeys.NAMED_BLOB_MAPPED_ID, result.getInsertedRecord().getBlobId());
            return result.getInsertedRecord().getBlobId();
          });`

And in ttlUpdateHanlder -> routerCallback we can directly call this method. updateBlobTtlAndStateToReady with the blobId and Version updated in the newNamedBlobRecord.
`          NamedBlobRecord record = new NamedBlobRecord(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
              namedBlobPath.getBlobName(), (String) restRequest.getArgs().get(NAMED_BLOB_MAPPED_ID), Utils.Infinite_Time,
              namedBlobVersion);
          namedBlobDb.updateBlobStateToReady(record).get();`

justinlin-linkedin (approved):

PR #2544, opened 2023-08-17
Implemented benefits based middle range compaction

JingQianCloud:
Two major tunings,
1. Periodically pack the middle range log segments. It's similar with existing full compaction but smarter.
2. Stats based compaction overly care about the IO cost. When IO cost is lower than some threshold, the difference can be ignored. Rather we should value "compaction benefit" more when the IO cost is very low.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -233,6 +233,20 @@ public class StoreConfig {
   public final int storeMinLogSegmentCountToReclaimToTriggerCompaction;
 
   /**
+   * Only the log segment whose valid data percentage is less or equal than the specified number
+   * will be qualified for compaction.number.
+   */
+  @Config("store.max.log.segment.valid.data.percentage.to.qualify.compaction")
+  public final double storeMaxLogSegmentValidDataPercentageToQualifyCompaction;
+
+  /**
+   * the time interval to run the "middle range compaction" in the stalts based compaction.
+   * the interval is in milliseconds.
+   */
+  @Config("store.stats.based.middle.range.compaction.interval.in.ms")
+  public final long storeStatsBasedMiddleRangeCompactionIntervalInMs;
+
+  /**
    * The number of buckets for stats bucketing, a value of 0 will disable bucketing.
    */
   @Config("store.stats.bucket.count")
@@ -615,6 +629,12 @@ public class StoreConfig {
         "com.github.ambry.store.CompactAllPolicyFactory");
     storeMinLogSegmentCountToReclaimToTriggerCompaction =
         verifiableProperties.getIntInRange("store.min.log.segment.count.to.reclaim.to.trigger.compaction", 1, 1, 1000);
+    storeMaxLogSegmentValidDataPercentageToQualifyCompaction =
+        verifiableProperties.getDoubleInRange("store.max.log.segment.valid.data.percentage.to.qualify.compaction", 0.30,
+            0.0, 1.0);
+    storeStatsBasedMiddleRangeCompactionIntervalInMs =
+        verifiableProperties.getLongInRange("store.stats.based.middle.range.compaction.interval.in.ms", 0, 0,
+            Long.MAX_VALUE);
     storeStatsBucketCount = verifiableProperties.getIntInRange("store.stats.bucket.count", 0, 0, 10000);
     storeStatsBucketSpanInMinutes =
         verifiableProperties.getLongInRange("store.stats.bucket.span.in.minutes", 60, 1, 10000);
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -234,7 +234,7 @@ public class StoreConfig {
 
   /**
    * Only the log segment whose valid data percentage is less or equal than the specified number
-   * will be qualified for compaction.number.
+   * will be qualified for compaction.
    */
   @Config("store.max.log.segment.valid.data.percentage.to.qualify.compaction")
   public final double storeMaxLogSegmentValidDataPercentageToQualifyCompaction;
@@ -247,7 +247,7 @@ public class StoreConfig {
   public final double storeStatsBasedCompactionMinCostInPercentage;
 
   /**
-   * the time interval to run the "middle range compaction" in the stalts based compaction.
+   * the time interval to run the "middle range compaction" in the stats based compaction.
    * the interval is in milliseconds.
    */
   @Config("store.stats.based.middle.range.compaction.interval.in.ms")
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -240,6 +240,13 @@ public class StoreConfig {
   public final double storeMaxLogSegmentValidDataPercentageToQualifyCompaction;
 
   /**
+   * In the stats based compaction, set the min cost for each log segment.
+   * If the IO effort is less than it, still count the cost as this min cost.
+   */
+  @Config("store.stats.based.compaction.min.cost.in.percentage")
+  public final double storeStatsBasedCompactionMinCostInPercentage;
+
+  /**
    * the time interval to run the "middle range compaction" in the stalts based compaction.
    * the interval is in milliseconds.
    */
@@ -629,12 +636,14 @@ public class StoreConfig {
         "com.github.ambry.store.CompactAllPolicyFactory");
     storeMinLogSegmentCountToReclaimToTriggerCompaction =
         verifiableProperties.getIntInRange("store.min.log.segment.count.to.reclaim.to.trigger.compaction", 1, 1, 1000);
-    storeMaxLogSegmentValidDataPercentageToQualifyCompaction =
-        verifiableProperties.getDoubleInRange("store.max.log.segment.valid.data.percentage.to.qualify.compaction", 0.30,
-            0.0, 1.0);
     storeStatsBasedMiddleRangeCompactionIntervalInMs =
         verifiableProperties.getLongInRange("store.stats.based.middle.range.compaction.interval.in.ms", 0, 0,
             Long.MAX_VALUE);
+    storeMaxLogSegmentValidDataPercentageToQualifyCompaction =
+        verifiableProperties.getDoubleInRange("store.max.log.segment.valid.data.percentage.to.qualify.compaction", 0.30,
+            0.0, 1.0);
+    storeStatsBasedCompactionMinCostInPercentage =
+        verifiableProperties.getDoubleInRange("store.stats.based.compaction.min.cost.in.percentage", 0.06, 0.0, 1.0);
     storeStatsBucketCount = verifiableProperties.getIntInRange("store.stats.bucket.count", 0, 0, 10000);
     storeStatsBucketSpanInMinutes =
         verifiableProperties.getLongInRange("store.stats.bucket.span.in.minutes", 60, 1, 10000);
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -381,6 +381,17 @@ class BlobStoreStats implements StoreStats, Closeable {
     return retValue;
   }
 
+  String dumpLogSegmentSize(NavigableMap<LogSegmentName, Long> validDataSizeByLogSegment, long segmentCapacity,
+      String dataDir) {
+    final StringBuilder sizeLog = new StringBuilder("Valid data size for " + dataDir + " from log segments: ");
+    validDataSizeByLogSegment.forEach((logSegmentName, validDataSize) -> {
+      sizeLog.append(
+          logSegmentName + " " + validDataSize / 1000 / 1000 / 1000.0 + "GB " + validDataSize * 1.0 / segmentCapacity
+              + "%;");
+    });
+    return sizeLog.toString();
+  }
+
   /**
    * Gets the storage stats for all serviceIds and their containerIds as of now (the time when the API is called).
    * Storage stats is comprised of 3 values: 1. valid data size (logicalStorageUsage) 2. physical data size 3. number of blobs.
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
@@ -381,6 +381,13 @@ class BlobStoreStats implements StoreStats, Closeable {
     return retValue;
   }
 
+  /**
+   * Dump the log segment size in a human readable way.
+   * @param validDataSizeByLogSegment a map from log segment name to valid size
+   * @param segmentCapacity Segment capacity of one {@link LogSegment}
+   * @param dataDir data directory on the disk
+   * @return the dump string of the log segment valid size in a human readable way.
+   */
   String dumpLogSegmentSize(NavigableMap<LogSegmentName, Long> validDataSizeByLogSegment, long segmentCapacity,
       String dataDir) {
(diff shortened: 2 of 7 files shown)


ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -121,23 +191,42 @@ private CostBenefitInfo getBestCandidateToCompact(NavigableMap<LogSegmentName, L
...
+    long adjustedTotalCost = 0;
     int totalSegmentsToBeCompacted = 0;
     LogSegmentName curSegmentName = firstLogSegmentName;
+    // weight more on benefits, don't let one single log segment with extreme small valid data leads to small compactions.
+    // Be default, the storeStatsBasedCompactionMinCostInPercentage is 0.06. 16G*0.06=1G
+    long minSegmentCost = storeConfig.storeStatsBasedCompactionMinCostInPercentage == 0 ? 0
+        : (long) (segmentCapacity * storeConfig.storeStatsBasedCompactionMinCostInPercentage);
     while (!curSegmentName.equals(lastLogSegmentName)) {
       totalCost += validDataSizePerLogSegment.get(curSegmentName);
+      adjustedTotalCost += validDataSizePerLogSegment.get(curSegmentName) < minSegmentCost ? minSegmentCost
```
ankagrawal, line 208: I don't follow this logic. So here we are saying that if valid data to copy is less than 1GiB, then take the cost as 1GiB. Doesn't this penalize segments that are smaller?

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -121,23 +191,42 @@ private CostBenefitInfo getBestCandidateToCompact(NavigableMap<LogSegmentName, L
...
     // add the last log segment
     totalCost += validDataSizePerLogSegment.get(curSegmentName);
+    adjustedTotalCost += validDataSizePerLogSegment.get(curSegmentName) < minSegmentCost ? minSegmentCost
+        : validDataSizePerLogSegment.get(curSegmentName);
     totalSegmentsToBeCompacted++;
     long maxCapacityPerSegment = segmentCapacity - segmentHeaderSize - maxBlobSize;
     int benefit = totalSegmentsToBeCompacted - (int) Math.ceil(totalCost / (maxCapacityPerSegment * 1.0));
+
+    if (weightOnBenefit) {
+      if (totalCost != adjustedTotalCost) {
```
ankagrawal, line 223: `totalCost` is always going to be great than `adjustedTotalCost` right? Should we change this condition to reflect that?

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -85,16 +101,70 @@ public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapac
...
-      long segmentCapacity, long segmentHeaderSize, long maxBlobSize) {
+      long segmentCapacity, long segmentHeaderSize, long maxBlobSize, BlobStoreStats blobStoreStats) {
     Map.Entry<LogSegmentName, Long> firstEntry = validDataPerLogSegments.firstEntry();
     Map.Entry<LogSegmentName, Long> lastEntry = validDataPerLogSegments.lastEntry();
     CostBenefitInfo bestCandidateToCompact = null;
+    boolean weightOnBenefit = false;
+
+    // weigh more on compaction benefit
+    // try to reclaim as many log segments as possible with reasonable cost.
+    if (storeConfig.storeStatsBasedMiddleRangeCompactionIntervalInMs != 0) {
```
ankagrawal, line 112: Can we separate out the two implementations (once where `weightOnBenefit` is `true`, and other where `weightOnBenefit` is `false`) into separate helper functions to make is easy to understand and make this function smaller.

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -232,6 +232,27 @@ public class StoreConfig {
   @Default("1")
   public final int storeMinLogSegmentCountToReclaimToTriggerCompaction;
 
+  /**
+   * Only the log segment whose valid data percentage is less or equal than the specified number
```
ankagrawal, line 236: nit: remove ".number." at the end of this comment.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
JingQianCloud: Yes, the logic is just as you mentioned.
If we set the threshold to 1GB, it means it worths to compact the log segments have less than 1GB valid size. It won't penalize the smallest log segments. It gives chance for the smaller but not smallest log segments to get compacted together with the smallest log segments.

In our current design, if we have log segments as below, (below is the valid size percentage). With current IO effort formula, we'll only compact log segment 3rd and 4th since their valid size is extreme small which is 1% and 1%.
```
double[] validPercentage = {0.9, 0.8, 0.05, 0.01, 0.01, 0.05, 0.1, 0.2, 0.4, 0.7}; 
                                                 1st, 2nd, 3rd, 4th,  ...............  
```

**But actually when the valid size percentage is less than some threshold, they are all ready to get compacted.** 

In the above example, if we set the threshold to 6%, with the adjustedTotalCost tuning, we'll compact log segment 2,3,4 and 5. The compaction benefit is 3, much better than the existing one which is 1. The IO cost added is very small.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
JingQianCloud: totalCost is always equal or smaller than adjustedTotalCost. Yes, I will change to "<".

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -54,14 +60,24 @@ public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapac
         Pair<Long, NavigableMap<LogSegmentName, Long>> validDataSizeByLogSegment =
             blobStoreStats.getValidDataSizeByLogSegment(
                 new TimeRange(time.milliseconds() - messageRetentionTimeInMs - ERROR_MARGIN_MS, ERROR_MARGIN_MS));
-        logger.info("Valid data size for {} from BlobStoreStats {} ", dataDir, validDataSizeByLogSegment);
+        if (validDataSizeByLogSegment != null) {
+          final StringBuilder sizeLog = new StringBuilder(
+              "Valid data size for " + dataDir + " from BlobStoreStats " + validDataSizeByLogSegment.getFirst()
```
justinlin-linkedin, line 65: we can create a utility method for this one.

justinlin-linkedin (commented):
I started to get what you want to do here, but i have to put some thought to it.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
JingQianCloud: Added one function for middleRangeCompaction.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -35,12 +37,16 @@ class StatsBasedCompactionPolicy implements CompactionPolicy {
   private final Time time;
   private final StoreConfig storeConfig;
   private final long messageRetentionTimeInMs;
+  // map from store id to the time when last time we run the "middle range compaction"
+  private final Map<String, Long> blobToLastMiddleRangeCompaction;
```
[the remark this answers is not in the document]
JingQianCloud: I see. Very good point. Let me think about a way to fix it.

ambry-store/src/main/java/com/github/ambry/store/CompactAllPolicyFactory.java
```diff
@@ -58,13 +61,28 @@ class CompactAllPolicy implements CompactionPolicy {
...
+        Pair<Long, NavigableMap<LogSegmentName, Long>> validDataSizeByLogSegment =
+            blobStoreStats.getValidDataSizeByLogSegment(
+                new TimeRange(time.milliseconds() - messageRetentionTimeInMs - ERROR_MARGIN_MS, ERROR_MARGIN_MS));
+        if (validDataSizeByLogSegment != null) {
+          final StringBuilder sizeLog = new StringBuilder(
+              "Valid data size for " + dataDir + " from BlobStoreStats " + validDataSizeByLogSegment.getFirst()
+                  + " segments: ");
+          validDataSizeByLogSegment.getSecond().forEach((logSegmentName, validDataSize) -> {
+            sizeLog.append(logSegmentName + " " + validDataSize / 1000 / 1000 / 1000.0 + "GB "
+                + validDataSize * 1.0 / segmentCapacity + "%;");
```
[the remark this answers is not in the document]
JingQianCloud: Yes, it's good to have file size or in-memory total size(not valid size) data. 
I try to avoid system call to get the file size for debug purpose. I don't find a in-memory one handy. Let me know if I miss anything.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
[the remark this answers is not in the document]
JingQianCloud: In my test, I'll set full compaction far away like 10 days.
So for testing purpose, it think it's ok. Don't want to change the hybridCompactionPolicy too much right now.
Actually with good tuning, we should get rid of the "time based" tuning totally. We may run "middle range compaction" when it's needed.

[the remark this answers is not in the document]
JingQianCloud: Fixed accordingly. Change the HybridPolicy to reuse statsBasedCompactionPolicy and compactAllPolicy.
Thanks, Justin.

justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
```diff
@@ -615,6 +636,14 @@ public StoreConfig(VerifiableProperties verifiableProperties) {
         "com.github.ambry.store.CompactAllPolicyFactory");
     storeMinLogSegmentCountToReclaimToTriggerCompaction =
         verifiableProperties.getIntInRange("store.min.log.segment.count.to.reclaim.to.trigger.compaction", 1, 1, 1000);
+    storeStatsBasedMiddleRangeCompactionIntervalInMs =
```
ankagrawal, line 639: Can we have a reasonable default value here? How often are we planning to run it currently?

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -121,23 +207,42 @@ private CostBenefitInfo getBestCandidateToCompact(NavigableMap<LogSegmentName, L
...
     // add the last log segment
     totalCost += validDataSizePerLogSegment.get(curSegmentName);
+    adjustedTotalCost += validDataSizePerLogSegment.get(curSegmentName) < minSegmentCost ? minSegmentCost
+        : validDataSizePerLogSegment.get(curSegmentName);
     totalSegmentsToBeCompacted++;
     long maxCapacityPerSegment = segmentCapacity - segmentHeaderSize - maxBlobSize;
     int benefit = totalSegmentsToBeCompacted - (int) Math.ceil(totalCost / (maxCapacityPerSegment * 1.0));
+
+    if (weightOnBenefit) {
+      if (totalCost < adjustedTotalCost) {
```
ankagrawal, line 239: `adjustedCost` will always be higher than or equal to `totalCost`. Then why do we need this `totalCost < adjustedTotalCost` check? Can we just assign the value without the check?

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -121,23 +207,42 @@ private CostBenefitInfo getBestCandidateToCompact(NavigableMap<LogSegmentName, L
...
+        logger.info("getCostBenefitInfo {} cost {} adjusted cost {}",
+            validDataSizePerLogSegment.subMap(firstLogSegmentName, true, lastLogSegmentName, true).keySet(), totalCost,
+            adjustedTotalCost);
+        totalCost = adjustedTotalCost;
```
ankagrawal, line 243: Also, I still don't understand why calculating the `adjustedTotalCost` matters. I understand that we want to take everything that has lower than `StoreConfig.storeMaxLogSegmentValidDataPercentageToQualifyCompaction` valid data size. But why change their cost? What will happen if we use the `totalCost` instead of `adjustedTotalCost`?

ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
JingQianCloud: I use this as a "enable" flag so I don't need another "enable" flog.
I probably will start with 2 day as our full compaction frequency. Ideally we will get rid of this "time based" frequency after we tune it well.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
JingQianCloud: Because I also need a if condition to print the log that we hit the tuning condition.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreStats.java
```diff
@@ -381,6 +381,17 @@ && getLogSegmentDeleteRefTimeMs(indexScanner.newScanResults, timeRange) != REF_T
     return retValue;
   }
 
+  String dumpLogSegmentSize(NavigableMap<LogSegmentName, Long> validDataSizeByLogSegment, long segmentCapacity,
```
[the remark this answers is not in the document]
JingQianCloud: Updated javadoc.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
JingQianCloud: The reason is that our cost is too sensitive on the IO cost.
One extreme small valid size log segment will populate the data to make us pick just two log segments while all the neighbor log segments are already ready for compaction.
Taking the previous example,


If we set the threshold to 1GB, it means it worths to compact the log segments have less than 1GB valid size. 
In our current design, if we have log segments as below, (below is the valid size percentage). With current IO effort formula, we'll only compact log segment 3rd and 4th since their valid size is extreme small which is 1% and 1%.
It's cost is (0.01 + 0.01) * logSegmentSize / 1 = 0.02*logSegmentSize

double[] validPercentage = {0.9, 0.8, 0.05, 0.01, 0.01, 0.05, 0.1, 0.2, 0.4, 0.7}; 
But actually when the valid size percentage is less than some threshold, they are all ready to get compacted.

In the above example, if we set the threshold to 6%, with the adjustedTotalCost tuning, we'll compact log segment 2,3,4 and 5. The compaction benefit is 3, much better than the existing one which is 1. The IO cost added is very small.
The cost for the four log segment is (0.05+0.01+0.01+0.05) * logSegmentSize / 3 = 0.04*logSegmentSize.

You can see without the tuning, we'll pick log segment 3th and 4th only. But actually all 2, 3, 4 and 5 are ready for compaction. After the tuning, the cost of the group of 2 and 3 will increase. And then group of 2,3,4 and 5 will win.

New calculation for two log segments: (0.06+0.06) * logSegmentSize / 1 = 0.12 * logSegmentSize
New calculation for 4 log segments: (0.06 + 0.06 + 0.06+ 0.06) * logSegmentSize / 3 = 0.08 * logSegmentSize. So the second one wins.

ankagrawal (approved):

PR #2545, opened 2023-08-18
[vcr-2.0] Impl put-blob to Azure

snalli:
Uploads Ambry blob to Azure blob storage. If blob already exists, then it catches BLOB_ALREADY_EXISTS exception and proceeds to next blob. It fails for any other exception.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
@@ -174,18 +174,20 @@ public class CloudBlobStoreV2 implements Store {
         azureMetrics.blobUploadSuccessRate.mark();
         // Measure ingestion rate, helps decide fleet size
         azureMetrics.backupSuccessByteRate.mark(messageInfo.getSize());
-        logger.trace("Successful upload of blob {} to Azure blob storage with statusCode = {}, etag = {}",
+        logger.debug("Successful upload of blob {} to Azure blob storage with statusCode = {}, etag = {}",
             messageInfo.getStoreKey().getID(), blockBlobItemResponse.getStatusCode(),
             blockBlobItemResponse.getValue().getETag());
       } catch (Exception e) {
-        logger.error("Failed to upload blob {} to Azure blob storage because {}", messageInfo.getStoreKey().getID(),
-            e.getMessage());
         if (e instanceof BlobStorageException
             && ((BlobStorageException) e).getErrorCode() == BlobErrorCode.BLOB_ALREADY_EXISTS) {
-          // This should never happen. If we invoke put(), then blob must be absent in the Store
+          // Since VCR replicates from all replicas, a blob can be uploaded by at least two threads concurrently.
+          logger.trace("Failed to upload blob {} to Azure blob storage because it already exists",
+              messageInfo.getStoreKey().getID());
           azureMetrics.blobUploadConflictCount.inc();
           continue;
         }
+        logger.error("Failed to upload blob {} to Azure blob storage because {}", messageInfo.getStoreKey().getID(),
+            e.getMessage());
         azureMetrics.blobUploadErrorCount.inc();
         throw new RuntimeException(e);
       } finally {
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
@@ -156,6 +156,7 @@ public class CloudBlobStoreV2 implements Store {
         // To avoid overwriting, pass "*" to setIfNoneMatch(String ifNoneMatch)
         // https://learn.microsoft.com/en-us/java/api/com.azure.storage.blob.blobclient?view=azure-java-stable
         blobParallelUploadOptions.setRequestConditions(new BlobRequestConditions().setIfNoneMatch("*"));
+        // This is ambry metadata, uninterpreted by azure
         blobParallelUploadOptions.setMetadata(messageInfoToMap(messageInfo));
         // Without content-type, get-blob floods log with warnings
         blobParallelUploadOptions.setHeaders(new BlobHttpHeaders().setContentType("application/octet-stream"));
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
@@ -181,7 +181,7 @@ public class CloudBlobStoreV2 implements Store {
         if (e instanceof BlobStorageException
             && ((BlobStorageException) e).getErrorCode() == BlobErrorCode.BLOB_ALREADY_EXISTS) {
           // Since VCR replicates from all replicas, a blob can be uploaded by at least two threads concurrently.
-          logger.trace("Failed to upload blob {} to Azure blob storage because it already exists",
+          logger.debug("Failed to upload blob {} to Azure blob storage because it already exists",
               messageInfo.getStoreKey().getID());
           azureMetrics.blobUploadConflictCount.inc();
           continue;
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
@@ -102,6 +102,27 @@ public class CloudBlobStoreV2 implements Store {
   }
 
   /**
+   * Detects duplicates in list of {@link MessageInfo}
+   * @param messageInfos list of {@link MessageInfo} to check for duplicates
+   */
+  protected void checkDuplicates(List<MessageInfo> messageInfos) {
+    HashMap<String, MessageInfo> seenKeys = new HashMap<>();
+    for (MessageInfo messageInfo : messageInfos) {
+      // Check for duplicates. We cannot continue as stream could be corrupted.
+      // Server always sends one message per key that captures full state of the key including ttl-update & delete.
+      // Emit a log and metric for investigation as it indicates a flaw in (de)serialization or replication.
+      // TODO: This should be in replication layer to avoid repeated code at Store level.
+      String blobId = messageInfo.getStoreKey().getID();
+      if (seenKeys.containsKey(blobId)) {
+        throw new IllegalArgumentException(
+            String.format("Duplicate message in replication-batch, original message = %s, duplicate message = %s",
+                seenKeys.get(blobId).toString(), messageInfo));
+      }
+      seenKeys.put(blobId, messageInfo);
+    }
+  }
+
+  /**
    * Uploads Ambry blob to Azure blob storage.
    * If blob already exists, then it catches BLOB_ALREADY_EXISTS exception and proceeds to next blob.
    * It fails for any other exception.
@@ -113,6 +134,9 @@ public class CloudBlobStoreV2 implements Store {
   public void put(MessageWriteSet messageSetToWrite) throws StoreException {
     MessageFormatWriteSet messageFormatWriteSet = (MessageFormatWriteSet) messageSetToWrite;
     Timer.Context storageTimer = null;
+
+    checkDuplicates(messageFormatWriteSet.getMessageSetInfo());
+
     // For-each loop must be outside try-catch. Loop must continue on BLOB_ALREADY_EXISTS exception
     for (MessageInfo messageInfo : messageFormatWriteSet.getMessageSetInfo()) {
       try {
@@ -161,13 +185,14 @@ public class CloudBlobStoreV2 implements Store {
             && ((BlobStorageException) e).getErrorCode() == BlobErrorCode.BLOB_ALREADY_EXISTS) {
           // This should never happen. If we invoke put(), then blob must be absent in the Store
           azureMetrics.blobUploadConflictCount.inc();
-          return;
+          continue;
         }
         azureMetrics.blobUploadErrorCount.inc();
         throw new RuntimeException(e);
       } finally {
         if (storageTimer != null) {
           storageTimer.stop();
+          storageTimer = null;
         }
       } // try-catch
     } // for-each
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
@@ -135,15 +135,15 @@ public class CloudBlobStoreV2 implements Store {
    */
   @Override
   public void put(MessageWriteSet messageSetToWrite) throws StoreException {
-    MessageFormatWriteSet messageFormatWriteSet = (MessageFormatWriteSet) messageSetToWrite;
+    MessageSievingInputStream messageSievingInputStream =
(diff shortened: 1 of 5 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -14,14 +14,24 @@
 
 package com.github.ambry.cloud;
 
+import com.azure.core.http.rest.Response;
```
JingQianCloud, line 17: I see why you against using the same naming convention as before.
Now Cloud is Azure. 
Before Cloud is the abstract layer and Azure is the implementation layer which can support different types of cloud implementation.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +78,100 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+        // To avoid overwriting, pass "*" to setIfNoneMatch(String ifNoneMatch)
+        // https://learn.microsoft.com/en-us/java/api/com.azure.storage.blob.blobclient?view=azure-java-stable
+        blobParallelUploadOptions.setRequestConditions(new BlobRequestConditions().setIfNoneMatch("*"));
+        blobParallelUploadOptions.setMetadata(messageInfoToMap(messageInfo));
+        // Without content-type, get-blob floods log with warnings
+        blobParallelUploadOptions.setHeaders(new BlobHttpHeaders().setContentType("application/octet-stream"));
+
+        ////////////////////////////////// Upload blob to Azure blob storage ////////////////////////////////////////
+        storageTimer = azureMetrics.blobUploadTime.time();
+        Response<BlockBlobItem> blockBlobItemResponse =
```
JingQianCloud, line 143: Now the operation is in sync mode.
Existing code is using async mode. We may have performance concern.
@Arun-LinkedIn 
Arun knows the details.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +78,100 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+            blobContainerClient.getBlobClient(messageInfo.getStoreKey().getID())
+                .uploadWithResponse(blobParallelUploadOptions, Duration.ofMillis(cloudConfig.cloudRequestTimeout),
+                    Context.NONE);
+        ////////////////////////////////// Upload blob to Azure blob storage ////////////////////////////////////////
```
JingQianCloud, line 147: Do we still support encryption?

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +78,100 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+  /**
+   * Uploads Ambry blob to Azure blob storage.
+   * If blob already exists, then it catches BLOB_ALREADY_EXISTS exception and proceeds to next blob.
+   * It fails for any other exception.
+   * @param messageSetToWrite The message set to write to the store
+   *                          Only the StoreKey, OperationTime, ExpirationTime, LifeVersion should be used in this method.
+   * @throws StoreException
+   */
   @Override
   public void put(MessageWriteSet messageSetToWrite) throws StoreException {
```
JingQianCloud, line 113: Old code has sanity check.
Is that safe to remove them?

```
    if (messageSetToWrite.getMessageSetInfo().isEmpty()) {
      throw new IllegalArgumentException("Message write set cannot be empty");
    }
    checkDuplicates(messageSetToWrite.getMessageSetInfo());
```

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
snalli: Yes, thank you. My experimental branch is far ahead of master branch and I have explored different combinations of the code, classes and naming. What I am putting out is the most straightforward solution.

snalli: Yes, thank you.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
Arun-LinkedIn: Thank you @JingQianCloud for sharing the concern. I am thinking sync client may be okay for VCR. I added async clients mainly for blueshift so that we could scale to QPS from users. Since replication happens in background, sync client might also be okay as code is more readable and easier to debug. (Today also, for VCR we use sync API `CloudBlobStore.putBlob()`)

May be in future, if we want to improve performance, we could check async apis?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureMetrics.java
```diff
@@ -142,6 +147,8 @@ public AzureMetrics(MetricRegistry registry) {
         registry.counter(MetricRegistry.name(AzureCloudDestination.class, BLOB_UPLOAD_REQUEST_COUNT));
     blobUploadSuccessCount =
         registry.counter(MetricRegistry.name(AzureCloudDestination.class, BLOB_UPLOAD_SUCCESS_COUNT));
+    blobUploadSuccessRate = registry.meter(MetricRegistry.name(AzureCloudDestination.class, BLOB_UPLOAD_SUCCESS_RATE));
+    blobUploadErrorCount = registry.counter(MetricRegistry.name(AzureCloudDestination.class, BLOB_UPLOAD_ERROR_COUNT));
```
Arun-LinkedIn, line 151: Are you planning to change the names to correct class (`CloudBlobStoreV2.class, etc`) later?

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
snalli: 1. Empty-list check

```
    if (messageSetToWrite.getMessageSetInfo().isEmpty()) {
      throw new IllegalArgumentException("Message write set cannot be empty");
    }
```
The list will never be empty in `put()` because of L1364 in `applyPut`.  The check for empty-list in `put()` is when VCR receives `PUT` from front-end, which is not a goal anymore. As of today, VCR receives `PUT` only from backend-server through replication, which goes through L1364 in `applyPut`.

2. Added check-for-duplicates, although this should be at a higher level in the replication-layer because I see two implementations of `checkDuplicates` at the store level, and now a third one.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
snalli: Spoke to Arun, and there is no performance concern. He can elaborate for the record.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
snalli: V1 does not have encryption as `vcrRequireEncryption = false` on all VCR clusters.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
JingQianCloud: Got you. Thanks for the explanation. @Arun-LinkedIn

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
JingQianCloud: Ack

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
JingQianCloud: Ack.

JingQianCloud (commented):
Overall looks good to me.
Waiting for the fix of the stream read as Arun pointed.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +78,100 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+    Timer.Context storageTimer = null;
+    // For-each loop must be outside try-catch. Loop must continue on BLOB_ALREADY_EXISTS exception
+    for (MessageInfo messageInfo : messageFormatWriteSet.getMessageSetInfo()) {
+      try {
+        // Read blob from input stream. It is not possible to just pass the input stream to azure-SDK.
+        // The SDK expects the stream to contain exactly the number of bytes to be read. If there are
+        // more bytes, it fails assuming the stream is corrupted. The inputStream we have here is a
+        // concatenation of many blobs and not just one blob.
+        ByteBuffer messageBuf = ByteBuffer.allocate((int) messageInfo.getSize());
+        if (messageFormatWriteSet.getStreamToWrite().read(messageBuf.array()) == -1) {
```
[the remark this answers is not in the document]
snalli: Good observation ! 

azure-sdk allows us to pass an `inputStream` to upload-blob. The stream must have number of bytes equal to size of the blob being uploaded, no more and no less. We have a list of `messageInfo` and we also have a list of `inputStream` in `MessageSievingInputStream`. I exposed this list of input streams and pass one stream at a time to the sdk. This simplifies the code.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +78,100 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+                .uploadWithResponse(blobParallelUploadOptions, Duration.ofMillis(cloudConfig.cloudRequestTimeout),
+                    Context.NONE);
+        ////////////////////////////////// Upload blob to Azure blob storage ////////////////////////////////////////
+
+        // Metrics and log
+        // Success rate is effective, Counter is ineffective because it just monotonically increases
+        azureMetrics.blobUploadSuccessRate.mark();
+        // Measure ingestion rate, helps decide fleet size
+        azureMetrics.backupSuccessByteRate.mark(messageInfo.getSize());
+        logger.trace("Successful upload of blob {} to Azure blob storage with statusCode = {}, etag = {}",
```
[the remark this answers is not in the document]
snalli: good idea !

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +78,100 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+            messageInfo.getStoreKey().getID(), blockBlobItemResponse.getStatusCode(),
+            blockBlobItemResponse.getValue().getETag());
+      } catch (Exception e) {
+        logger.error("Failed to upload blob {} to Azure blob storage because {}", messageInfo.getStoreKey().getID(),
```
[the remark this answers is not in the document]
snalli: Good catch !

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +78,100 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+            e.getMessage());
+        if (e instanceof BlobStorageException
+            && ((BlobStorageException) e).getErrorCode() == BlobErrorCode.BLOB_ALREADY_EXISTS) {
+          // This should never happen. If we invoke put(), then blob must be absent in the Store
```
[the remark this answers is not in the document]
snalli: Good catch !

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureMetrics.java
snalli: Good catch !!

JingQianCloud (approved):
LGTM

snalli:
Tested in prod and able to upload a few blobs with no errors.
@Arun-LinkedIn can u confirm that the code where we are reading for the stream is ok ?

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +79,145 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+
+  /**
+   * Detects duplicates in list of {@link MessageInfo}
+   * @param messageInfos list of {@link MessageInfo} to check for duplicates
+   */
+  protected void checkDuplicates(List<MessageInfo> messageInfos) {
+    if (messageInfos.size() < 2) {
+      return;
+    }
+    HashMap<String, MessageInfo> seenKeys = new HashMap<>();
```
Arun-LinkedIn, line 113: nit: `HashSet` may be sufficient

Arun-LinkedIn (approved):
Looks good to me too. Just had one question.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
snalli: Actually, I wish to print out the full message info along with key and that's why its a map

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -63,10 +79,145 @@ public CloudBlobStoreV2(VerifiableProperties properties, MetricRegistry metricRe
...
+        (MessageSievingInputStream) ((MessageFormatWriteSet) messageSetToWrite).getStreamToWrite();
+
+    // This is a list of blob-ids from remote server
+    List<MessageInfo> messageInfos = messageSievingInputStream.getValidMessageInfoList();
+    checkDuplicates(messageInfos);
+
+    // This is a list of data-streams associated with blob-ids. i-th stream is for i-th blob-id.
+    // This allows us to pass stream to azure-sdk and let it handle read() logic.
+    List<InputStream> messageStreamList = messageSievingInputStream.getValidMessageStreamList();
+    if (messageInfos.size() != messageStreamList.size()) {
```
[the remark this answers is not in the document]
snalli: The separation of msg infos and msg stream into two lists happens only in `handleGetResponse` L1232 and immediately after that is `applyPut`. There are no checks in between that assert the length of the two lists are the same. So whether this check is `handleGetResponse` or `applyPut`, it happens just once. However, unequal lengths would be really really rare because the addition to each list happens together like this
```
      sievedMessageInfoList.add(inMsg.getMessageInfo());
      msgStreamList.add(inMsg.getStream());
```

PR #2546, opened 2023-08-18
Use one single connection for dataset version request

SophieGuo410:
When we query the dataset version, we need to get the Dataset and make sure it's not expired or deleted. 
But we need to make sure when getDataset is for add datasetVersion or update datasetVersion related call, we need to enable writable connection in order to make sure it's exist.
And anything related to auto increment or latest version should also be writable since we need to support strong consistency.


JingQianCloud (approved):
LGTM

justinlin-linkedin (approved):

PR #2547, opened 2023-08-18
Support ttl update handler for named blob and dataset version

SophieGuo410:


ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
```diff
@@ -154,6 +169,26 @@ private Callback<Void> securityPostProcessRequestCallback(BlobId blobId) {
      */
     private Callback<Void> routerCallback() {
       return buildCallback(metrics.updateBlobTtlRouterMetrics, result -> {
+        if (RequestPath.matchesOperation(blobIdStr, Operations.NAMED_BLOB)) {
```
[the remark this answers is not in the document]
SophieGuo410: You are right! Updated it.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -340,6 +341,9 @@ private String getDatasetVersion(RestRequest restRequest) throws RestServiceExce
         restResponseChannel.setHeader(RestUtils.Headers.TARGET_CONTAINER_NAME, containerName);
         restResponseChannel.setHeader(RestUtils.Headers.TARGET_DATASET_NAME, datasetName);
         restResponseChannel.setHeader(RestUtils.Headers.TARGET_DATASET_VERSION, datasetVersionRecord.getVersion());
+        restResponseChannel.setHeader(RestUtils.Headers.DATASET_EXPIRATION_TIME,
+            datasetVersionRecord.getExpirationTimeMs() == -1 ? "Infinite_Time"
+                : new Date(datasetVersionRecord.getExpirationTimeMs()));
```
justinlin-linkedin, line 346: standard http response header "Expires" can show the expiration date/time for an entity. I think it would be better to use standard http header here. If this dataset version has no ttl, we don't have to set this header. but it does, we can set "Expires" header.

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
```diff
@@ -143,17 +158,34 @@ private Callback<String> idConverterCallback() {
...
       }, restRequest.getUri(), LOGGER, finalCallback);
     }
 
     /**
      * After {@link Router#updateBlobTtl} finishes, call {@link SecurityService#processResponse}.
+     * @param blobId the {@link BlobId} to update
      * @return a {@link Callback} to be used with {@link Router#updateBlobTtl}.
      */
-    private Callback<Void> routerCallback() {
+    private Callback<Void> routerCallback(BlobId blobId) {
```
justinlin-linkedin, line 171: can you rename this parameter to convertedBlobId, since there is already a blobIdStr in the class?

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
```diff
@@ -143,17 +158,34 @@ private Callback<String> idConverterCallback() {
...
+        if (RequestPath.matchesOperation(blobIdStr, Operations.NAMED_BLOB)) {
+          // Set the named blob state to be 'READY' after the Ttl update succeed
+          if (!restRequest.getArgs().containsKey(NAMED_BLOB_VERSION)) {
+            throw new RestServiceException(
+                "Internal key " + NAMED_BLOB_VERSION + " is required in Named Blob TTL update callback!",
+                RestServiceErrorCode.InternalServerError);
+          }
+          long namedBlobVersion = (long) restRequest.getArgs().get(NAMED_BLOB_VERSION);
+          NamedBlobPath namedBlobPath = NamedBlobPath.parse(blobIdStr, restRequest.getArgs());
+          NamedBlobRecord record = new NamedBlobRecord(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
```
justinlin-linkedin, line 182: BTW, in NamedBlobRecord, can we make sure that all the members are final. It's weird that for a data structure type of class, some members are not final.

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
```diff
@@ -143,17 +158,34 @@ private Callback<String> idConverterCallback() {
...
+              namedBlobPath.getBlobName(), blobId.getID(), Utils.Infinite_Time, namedBlobVersion);
+          namedBlobDb.updateBlobTtlAndStateToReady(record).get();
```
justinlin-linkedin, line 184: nit: do you think we should put the rest of the logic as a callback to the CompletableFuture returned by `updateBlobTtlAndStateToReady`?

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
SophieGuo410: I can do that, but I checked we have this setCacheHeaders in AmbrySecurityService which will always set the EXPIRES to DATE, I'm kind of confused about why we always set that for a get call.
restResponseChannel.setHeader(RestUtils.Headers.EXPIRES, restResponseChannel.getHeader(RestUtils.Headers.DATE));

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
SophieGuo410: Sure, updated.

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
SophieGuo410: Updated.

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
SophieGuo410: I've updated it, but just wondering what's your concern about that. 
Is it because you think it's more clear if we chain the flow with callback?

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
```diff
@@ -126,38 +141,59 @@ private Callback<Void> securityProcessRequestCallback() {
...
+          if (!restRequest.getArgs().containsKey(NAMED_BLOB_VERSION)) {
+            throw new RestServiceException(
+                "Internal key " + NAMED_BLOB_VERSION + " is required in Named Blob TTL update callback!",
+                RestServiceErrorCode.InternalServerError);
+          }
+          long namedBlobVersion = (long) restRequest.getArgs().get(NAMED_BLOB_VERSION);
+          NamedBlobPath namedBlobPath = NamedBlobPath.parse(blobIdStr, restRequest.getArgs());
+          NamedBlobRecord record = new NamedBlobRecord(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
+              namedBlobPath.getBlobName(), convertedBlobId.getID(), Utils.Infinite_Time, namedBlobVersion);
+          namedBlobDb.updateBlobTtlAndStateToReady(record).thenRun(() -> {
```
justinlin-linkedin, line 184: nit: there is actually a utility method that deals with this type of conversion. It's in the CallbackUtils. You can do 
```
CallbackUtils.callCallbackAfter(namedBlobDb.updateBlobTtlAndStateToReady(record), (r, e) -> {
    if (e != null) {
        finalCallback.onComplete(null, e);
    } else {
        try {
            if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
                updateTtlForDatasetVersion();
            }
            processResponseHelper();
        } catch (RestServiceException ex) {
            finalCallback.onCompletion(null, ex);
        }
    }
})
```

justinlin-linkedin (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/TtlUpdateHandler.java
justinlin-linkedin: database call can take some time, and we are blocking it here, by changing it to callback chain (completable future chain), we are not blocking it.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
justinlin-linkedin: oh, my bad, EXPIRES is for browser cache, not for the blob itself. I don't think EXPIRES is going to work. :(

justinlin-linkedin (approved):

PR #2550, opened 2023-08-25
[vcr-2.0] Impl findKey and findMissingKey

snalli:
This patch implements findKey and findMissingKey for Azure blob storage.
findKey is used to fetch a blob metadata to update it including ttl-update, delete and undelete timestamps.
findMissingKey is used to determine keys absent in Azure blob storage.

This patch also prints errlog in replica thread that was not being printed before.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
@@ -19,6 +19,7 @@ import com.azure.core.util.Context;
 import com.azure.storage.blob.BlobContainerClient;
 import com.azure.storage.blob.models.BlobErrorCode;
 import com.azure.storage.blob.models.BlobHttpHeaders;
+import com.azure.storage.blob.models.BlobProperties;
 import com.azure.storage.blob.models.BlobRequestConditions;
 import com.azure.storage.blob.models.BlobStorageException;
 import com.azure.storage.blob.models.BlockBlobItem;
@@ -38,6 +39,7 @@ import com.github.ambry.store.FindInfo;
 import com.github.ambry.store.MessageInfo;
 import com.github.ambry.store.MessageWriteSet;
 import com.github.ambry.store.Store;
+import com.github.ambry.store.StoreErrorCodes;
 import com.github.ambry.store.StoreException;
 import com.github.ambry.store.StoreGetOptions;
 import com.github.ambry.store.StoreInfo;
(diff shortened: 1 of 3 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -235,17 +237,67 @@ public void updateTtl(List<MessageInfo> messageInfos) {
...
   @Override
   public Set<StoreKey> findMissingKeys(List<StoreKey> storeKeys) {
-    // TODO
-    return new HashSet<>();
+    Set<StoreKey> missingKeys = new HashSet<>();
+    Timer.Context timer = azureMetrics.missingKeysQueryTime.time();
+    try {
+      for (StoreKey storeKey : storeKeys) {
+        try {
+          if (!blobContainerClient.getBlobClient(storeKey.getID()).exists()) {
```
JingQianCloud, line 252: Is there any batched API to check the existence for multiple blobs in Azure?

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
```diff
@@ -235,17 +237,67 @@ public void updateTtl(List<MessageInfo> messageInfos) {
...
+
+  public static final String DELETE_TIME_MS = "deleteTimeMs";
+  public static final String TTL_UPDATE_TIME_MS = "ttlUpdateTimeMs";
+  public static final String UNDELETE_TIME_MS = "undeleteTimeMs";
+
+  /**
+   * @return a {@link HashMap} of metadata key-value pairs.
+   */
+  public MessageInfo messageInfoFromMap(StoreKey storeKey, Map<String, String> map) {
+    return new MessageInfo(storeKey, Long.parseLong(map.get(SIZE)), map.containsKey(DELETE_TIME_MS),
```
JingQianCloud, line 275: If the blob is deleted and then undeleted, isDeleted should be true or false?
map.containsKey(DELETE_TIME_MS) will be true even it's undeleted, right?

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
snalli: Good question ! I looked for this first and there is no such API
https://stackoverflow.com/questions/54879762/how-to-check-if-multiple-files-exists-in-azure-container

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStoreV2.java
snalli: This depends on how I implement delete() and undelete(), which is next PR. find() will just report what it finds. 
I think that if a blob deleted and then undeleted, the delete_timestamp should be cleared or removed from metadata and the lifeVersion must be incremented. isDeleted will be false.

JingQianCloud: Can we find out what do we do in the Ambry server? Do we set isDelete to true or false when the blob is deleted and undeleted? If you want to delay this, please add TODO.

Anything we are uncertain at this moment, I'll suggest to add one specific TODO. Otherwise, when these questions pile up, we'll lose track of them. 
For example, TODO_VCR_BACKUP_V2, something like that and keep using the same TODO for all your PRs.

snalli: Ok, I have a few thoughts and questions about delete/undelete. I will setup a meeting or add to my doc and discuss.

PR #2552, opened 2023-08-28
[Efficient Metadata Operations] Implement Request and Response for Purge Message from frontend to server.

ankagrawal:
In the efficient metadata operations design, the frontend will need to send message to storage nodes to indicate that a data chunk is ready to be compacted away, when the metadata chunk is expired or deleted and past retention time. This PR implements the Purge Message will be used to pass this message from frontend to server.

### DIFF
--- a/ambry-protocol/src/main/java/com/github/ambry/protocol/PurgeRequest.java
+++ b/ambry-protocol/src/main/java/com/github/ambry/protocol/PurgeRequest.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2023 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


ambry-protocol/src/main/java/com/github/ambry/protocol/PurgeRequest.java
```diff
@@ -0,0 +1,129 @@
...
+  public String toString() {
+    return "PurgeRequest[" + "BlobID=" + blobId + ", " + "PartitionId=" + blobId.getPartition() + ", " + "ClientId="
+        + clientId + ", " + "CorrelationId=" + correlationId + ", " + "AccountId=" + blobId.getAccountId() + ", "
+        + "ContainerId=" + blobId.getContainerId() + "]";
+  }
+
+  /**
+   * Class to read protocol version 1 PurgeRequest from the stream.
+   */
+  public static class CompactMessageRequest_V1 {
```
JingQianCloud, line 121: Shall we use consistent name, both Purge or both Compact?
I prefer Purge more, since we have used compact for server already.

JingQianCloud (commented):
Looks good to me. One minor comment.

ambry-protocol/src/main/java/com/github/ambry/protocol/PurgeRequest.java
ankagrawal: fixed it.

JingQianCloud (approved):
Approved.
For TODO, I suggest to add specific TODO like EFFICIENT_META_TODO etc.
Otherwise, the TODO will get lost in the massive amount of generic TODOs.

PR #2563, opened 2023-09-05
Change server request queue to be bounded

Arun-LinkedIn:
This PR:
1. Changes Ambry server request queue to bounded
2. Rejects the response from NettyRequestResponseChannel immediately when queue is full
3. Removes separate background thread for dropping requests but instead uses existing request handler threads
4. Removes putting dummy request into queue to signal shutting down of request handler threads. Instead it uses executor service to shutdown request handler threads


ambry-network/src/main/java/com/github/ambry/network/http2/AmbryNetworkRequestHandler.java
```diff
@@ -32,18 +35,42 @@ public class AmbryNetworkRequestHandler extends SimpleChannelInboundHandler<Full
...
+      boolean isRequestQueued = requestResponseChannel.sendRequest(networkRequest);
+      // Reject the request if input queue is full.
+      if (!isRequestQueued) {
+        try {
+          // 1. Consume the first 8 bytes of the input stream which represents the size of the request. This is added to the
+          // request since Ambry's SocketServer stack needs to know the number of the bytes to read for a request
+          // (see BoundedNettyByteBufReceive#readFrom(ReadableByteChannel) method). This is not needed when using
+          // Netty HTTP2 server stack. We can remove this once SocketServer stack is retired. For now, we are
+          // consuming the bytes here so that rest of request handling in server is same.
+          DataInputStream stream = new DataInputStream(networkRequest.getInputStream());
```
JingQianCloud, line 61: Any reason that we don't call consumeSizeHeader as other places?

JingQianCloud (commented):
Overall looks good to me, just have two questions embedded,

snalli:
Why is this PR so huge ? I thought changing to bounded queue would be a small change ? Are there two PRs clubbed into one ?

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
```diff
@@ -37,6 +37,7 @@ public class NetworkConfig {
   public static final String SELECTOR_USE_DIRECT_BUFFERS = "selector.use.direct.buffers";
   public static final String REQUEST_QUEUE_TYPE = "request.queue.type";
   public static final String REQUEST_QUEUE_TIMEOUT_MS = "request.queue.timeout.ms";
+  public static final String REQUEST_QUEUE_CAPACITY = "request.queue.capacity";
```
snalli, line 40: I think this a good config. If we want an unbounded queue, we can set it to a high val or -1, if we interpret -1 to be infinity ?

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
```diff
@@ -170,6 +177,8 @@ public NetworkConfig(VerifiableProperties verifiableProperties) {
     requestQueueType = verifiableProperties.getEnum(REQUEST_QUEUE_TYPE, RequestQueueType.class,
         RequestQueueType.BASIC_QUEUE_WITH_FIFO);
     requestQueueTimeoutMs = verifiableProperties.getIntInRange(REQUEST_QUEUE_TIMEOUT_MS, 2000, 0, Integer.MAX_VALUE);
+    // Used 5000 since Ambry incoming max qps seems to be 2.5K in last few weeks and default request.queue.timeout is 2 seconds
```
snalli, line 180: How did you decide 5000 ? Just double the max qps ?

ambry-api/src/main/java/com/github/ambry/config/NetworkConfig.java
```diff
@@ -170,6 +177,8 @@ public NetworkConfig(VerifiableProperties verifiableProperties) {
...
+    requestQueueCapacity = verifiableProperties.getIntInRange(REQUEST_QUEUE_CAPACITY, 5000, 0, Integer.MAX_VALUE);
     adaptiveLifoQueueCodelTargetDelayMs =
```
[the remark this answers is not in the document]
snalli: 1. Does queue have requests or chunks ? If requests, then can the requests be of arbitrary size ?
2. 20GB is ok, that's where most servers are but I'd say make it even more like 25,000 if each item in queue is no more than 4MB ? That way the outliers will stick out. Otherwise we cannot distinguish a bad host from a good one because all are 25-30g directmem. 100g of direct-mem will not crash server because the upper limit is 150gb.

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -56,11 +57,18 @@ public NettyServerRequestResponseChannel(NetworkConfig config, Http2ServerMetric
...
-  /** Send a request to be handled */
+  /**
+   * Send a request to be handled
+   * @return {@code True} if we are able to send the request over the channel. Else {@code False}
+   */
   @Override
-  public void sendRequest(NetworkRequest request) throws InterruptedException {
-    networkRequestQueue.offer(request);
-    http2ServerMetrics.requestEnqueueTime.update(System.currentTimeMillis() - request.getStartTimeInMs());
+  public boolean sendRequest(NetworkRequest networkRequest) throws InterruptedException {
```
snalli, line 65: Does this fn take requests coming in from the frontend and add it to a queue in the server ?
What are the threads in the server consuming this queue ?

ambry-network/src/main/java/com/github/ambry/network/FifoNetworkRequestQueue.java
```diff
@@ -29,30 +28,31 @@
...
   @Override
-  public void offer(NetworkRequest request) throws InterruptedException {
-    queue.offer(request);
-    if (request.equals(EmptyRequest.getInstance())) {
-      // This is a internal generated request used to signal shutdown. Add it to dropped requests queue as well so that
-      // request-dropper thread shuts down gracefully.
-      droppedRequestsQueue.offer(request);
-    }
+  public boolean offer(NetworkRequest request) throws InterruptedException {
+    return queue.offer(request);
```
snalli, line 55: What happens if the queue is full ?

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestHandler.java
```diff
@@ -40,10 +39,6 @@ public void run() {
     while (true) {
       try {
         req = requestChannel.receiveRequest();
-        if (req.equals(EmptyRequest.getInstance())) {
```
snalli, line 43: Why is the EmptyRequest removed everywhere ?

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestHandlerPool.java
```diff
@@ -43,19 +45,27 @@ public class RequestHandlerPool implements Closeable {
...
   public RequestHandlerPool(int numThreads, RequestResponseChannel requestResponseChannel, RequestAPI requests,
       String handlerPrefix, boolean enableDropper) {
-    threads = new Thread[numThreads];
-    handlers = new RequestHandler[numThreads];
     this.requestResponseChannel = requestResponseChannel;
+    // 1. Start worker threads for handling requests
+    ThreadFactory requestHandlerThreadFactory =
+        new ThreadFactoryBuilder().setNameFormat(handlerPrefix + "request-handler-%d")
+            .setDaemon(true)
+            .setUncaughtExceptionHandler((t, e) -> logger.error("Encountered throwable in {}", t, e))
```
snalli, line 53: Should it be two curly braces ? one for t and other for e ?

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestHandlerPool.java
```diff
@@ -43,19 +45,27 @@ public class RequestHandlerPool implements Closeable {
...
+
+    // 2. Start a worker thread for rejecting stale requests.
     if (enableDropper) {
-      requestDropper = new RequestDropper(requestResponseChannel, requests);
-      requestDropperThread = Utils.daemonThread(handlerPrefix + "request-dropper", requestDropper);
-      requestDropperThread.start();
+      ThreadFactory requestDropperThreadFactory =
+          new ThreadFactoryBuilder().setNameFormat(handlerPrefix + "request-dropper-%d")
+              .setDaemon(true)
+              .setUncaughtExceptionHandler((t, e) -> logger.error("Encountered throwable in {}", t, e))
```
snalli, line 65: same as above

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestHandlerPool.java
```diff
@@ -43,19 +45,27 @@ public class RequestHandlerPool implements Closeable {
    */
   public RequestHandlerPool(int numThreads, RequestResponseChannel requestResponseChannel, RequestAPI requests,
       String handlerPrefix, boolean enableDropper) {
-    threads = new Thread[numThreads];
-    handlers = new RequestHandler[numThreads];
     this.requestResponseChannel = requestResponseChannel;
+    // 1. Start worker threads for handling requests
+    ThreadFactory requestHandlerThreadFactory =
+        new ThreadFactoryBuilder().setNameFormat(handlerPrefix + "request-handler-%d")
```
snalli, line 51: these are the consumers of the bounded queue right ?

ambry-network/src/main/java/com/github/ambry/network/http2/AmbryNetworkRequestHandler.java
```diff
@@ -32,18 +35,42 @@ public class AmbryNetworkRequestHandler extends SimpleChannelInboundHandler<Full
...
+          stream.readLong();
```
snalli, line 62: Didn't understand the logic of consuming first 8 bytes. We are not using it anywhere here.

PR #2566, opened 2023-09-07
Add shared variable to communicate server shutting down to its worker threads

Arun-LinkedIn:
When we change incoming request queue to bounded, it is possible that EmptyRequest (that is used to signal shutdown to worker threads) is not queued due to queue being full. To handle this case, adding a shared variable to communicate server shutdown to worker threads.

### DIFF
--- a/ambry-protocol/src/main/java/com/github/ambry/protocol/RequestDropper.java
+++ b/ambry-protocol/src/main/java/com/github/ambry/protocol/RequestDropper.java
@@ -17,6 +17,7 @@ import com.github.ambry.network.NetworkRequest;
 import com.github.ambry.network.RequestResponseChannel;
 import com.github.ambry.server.EmptyRequest;
 import java.util.Collection;
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.slf4j.Logger;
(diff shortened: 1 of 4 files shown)


JingQianCloud (approved):
LGTM

ambry-protocol/src/main/java/com/github/ambry/protocol/RequestDropper.java
```diff
@@ -64,7 +66,17 @@ public void run() {
   }
 
   public void shutdown() throws InterruptedException {
+    running.set(false);
     requestChannel.sendRequest(EmptyRequest.getInstance());
```
ankagrawal, line 70: Do we need to append empty request anymore?

Arun-LinkedIn: Yeah, we may need it. `requestChannel.getDroppedRequests()` is a blocking call. Due to that, if there are no requests (for ex: a server is up but not taking any requests yet), the execution won't come out of the loop and doesn't check the `running` status. Server would only forcefully shutdown after 5 mins (timeout) in this case.

ankagrawal (approved):

PR #2569, opened 2023-09-07
Add helper class to deserialize server requests

Arun-LinkedIn:
Adding helper class to deserialize server requests. This will be helpful to deserialize incoming request and send response from Netty IO threads itself in future PR.

### DIFF
--- a/ambry-protocol/src/main/java/com/github/ambry/protocol/ServerRequestResponseHelper.java
+++ b/ambry-protocol/src/main/java/com/github/ambry/protocol/ServerRequestResponseHelper.java
@@ -1,0 +1,130 @@
+/**
+ * Copyright 2023 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.github.ambry.protocol;
+
+import com.github.ambry.clustermap.ClusterMap;
+import com.github.ambry.network.LocalRequestResponseChannel;
+import com.github.ambry.network.NetworkRequest;
+import com.github.ambry.replication.FindTokenHelper;
+import com.github.ambry.server.ServerErrorCode;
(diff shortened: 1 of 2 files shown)


ambry-protocol/src/main/java/com/github/ambry/protocol/ServerRequestResponseHelper.java
```diff
@@ -0,0 +1,130 @@
...
+        default:
+          throw new UnsupportedOperationException("Request type not supported");
+      }
+    } else {
+      request = (RequestOrResponse) ((LocalRequestResponseChannel.LocalChannelRequest) networkRequest).getRequestInfo()
+          .getRequest();
+      if (request.getRequestType() == RequestOrResponseType.PutRequest) {
+        // For PutRequest sent via local channel, we have to create a new PutRequest object to wrap the data buffer as a
+        // stream. Also, we need to set the crc value to null since the request is not coming over network.
+        PutRequest receivedPutRequest = ((PutRequest) request);
```
JingQianCloud, line 83: For the PutBlob local request, getDecodedRequest will return the newly created PutRequest. 
When we send out the response for that request, we'll use this internal created PutRequest as the reference.
But it's not a LocalChannelRequest, right? Do we have problem for the sendResponse as below?
In the old implementation, although we also create the new PutBlob, but we send out the response based on the original request.

```
    @Override
  public void sendResponse(Send payloadToSend, NetworkRequest originalRequest, ServerNetworkResponseMetrics metrics) {
    LocalChannelRequest localRequest = (LocalChannelRequest) originalRequest;
    ResponseInfo responseInfo =
        new ResponseInfo(localRequest.requestInfo, null, localRequest.requestInfo.getReplicaId().getDataNodeId(),
            payloadToSend);
    BatchBlockingQueue<ResponseInfo> responseQueue = getResponseQueue(localRequest.processorId);
    responseQueue.put(responseInfo);
    logger.debug("Added response for {}, size now {}", localRequest.processorId, responseQueue.size());
  }

```

Arun-LinkedIn: Hmm, actually PutRequest is present inside `NetworkRequest` as `networkRequest.requestInfo.request`. In both old and new code, we are using original `NetworkRequest` in the `sendResponse()`. I think creating a new `PutRequest` should be okay.

JingQianCloud: I see. With the new code, we still send out the response with the NetworkRequest. I missed that.

JingQianCloud (approved):

Arun-LinkedIn:
Thank you @JingQianCloud . Merging the PR so that I can put up dependent PR.

PR #2576, opened 2023-09-14
log error in doPrefetch

snalli:


Arun-LinkedIn (approved):

justinlin-linkedin (approved):

snalli:
cancel because ambry can request chunk size greater than 4mb (data + some md like crc) and netty's chunk size of pooled mem is 4mb, anything more goes to unpooled.

PR #2577, opened 2023-09-15
Reject expired requests in request handler threads

Arun-LinkedIn:
This pr uses `RequestHandler` threads for rejecting expired requests and removes the `RequestDropper` thread. 

This helps to:
1. Reject expired requests quicker (and avoid memory build up) since there are 160 request-handler threads as compared to 1 request-dropper thread. 
2. Avoids logic to reject requests in 2 places.


JingQianCloud (approved):
I love this code cleanup.
Thanks for keeping improve the code quality.

justinlin-linkedin (approved):

PR #2582, opened 2023-09-20
Fix unsupported request type error handling and catch exception

Arun-LinkedIn:
Catch exception in `NettyRequestResponseChannel.rejectRequests` so that it is not propagated to request handler threads and cause server shutdown

### DIFF
--- a/ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
+++ b/ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
@@ -23,7 +23,6 @@ import com.github.ambry.utils.SystemTime;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
-import java.io.IOException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -37,6 +36,7 @@ public class NettyServerRequestResponseChannel implements RequestResponseChannel
   private final NetworkRequestQueue networkRequestQueue;
   private final ServerMetrics serverMetrics;
   private final ServerRequestResponseHelper requestResponseHelper;
+  protected static final Logger publicAccessLogger = LoggerFactory.getLogger("PublicAccessLogger");
 
   public NettyServerRequestResponseChannel(NetworkConfig config, Http2ServerMetrics http2ServerMetrics,
       ServerMetrics serverMetrics, ServerRequestResponseHelper requestResponseHelper) {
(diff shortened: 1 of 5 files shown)


ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -37,6 +36,7 @@ public class NettyServerRequestResponseChannel implements RequestResponseChannel
   private final NetworkRequestQueue networkRequestQueue;
   private final ServerMetrics serverMetrics;
   private final ServerRequestResponseHelper requestResponseHelper;
+  protected static final Logger publicAccessLogger = LoggerFactory.getLogger("PublicAccessLogger");
```
JingQianCloud, line 39: I'm not sure how the LoggerFactory works. With that, will publicAccessLogger write to the public log instead of the ambry log?

Also, right now we are using LogManager instead of LoggerFactory for the PublicAccessLogger. I don't know the difference between them.
  private static final Logger publicAccessLogger = LogManager.getLogger("PublicAccessLogger");

JingQianCloud (approved):
LGTM

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
```diff
@@ -137,18 +148,21 @@ public void shutdown() {
...
       } else {
         http2ServerMetrics.requestResponseChannelDroppedOnOverflowCount.inc();
       }
       serverMetrics.totalRequestDroppedRate.mark();
+      // Log the request and response in public access logs
+      long requestProcessingTime = SystemTime.getInstance().milliseconds() - networkRequest.getStartTimeInMs();
+      publicAccessLogger.info("{} {} processingTime {}", request, response, requestProcessingTime);
       sendResponse(response, networkRequest, null);
-    } catch (IOException | InterruptedException e) {
+    } catch (Exception e) {
```
justinlin-linkedin, line 164: we should capture Throwable, instead of Exception.

justinlin-linkedin (approved):

ambry-network/src/main/java/com/github/ambry/network/NettyServerRequestResponseChannel.java
Arun-LinkedIn: Yeah, I think the Logger obtained via `PublicAccessLogger` name will write logs to public access logs. I think log4j2 has concept of "appenders" which allows us to choose the destination log file for a logger via configuration. If we see in internal AmbryLI configuration, the appender for `PublicAccessLogger` redirects logs to public access logs.

LogManager is used for structured logging. But I think we only use it for frontends. For servers, we are still using LoggerFactory. All the server public access logging happens in AmbryRequests.java. 

This is my understanding :)

PR #2585, opened 2023-09-21
Adding more logs and metrics and make transactionIsolationLevel tunable

SophieGuo410:
This pr support:
1. Adding more logs and metrics.
2. Make the transactionIsolationLevel tunable for metadata db since we will support auto purge. This is similar like what we've been done for named blob db.

### DIFF
--- a/ambry-mysql/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
@@ -13,6 +13,8 @@
  */
 package com.github.ambry.mysql;
 
+import com.github.ambry.config.MySqlAccountServiceConfig;
+import com.github.ambry.named.TransactionIsolationLevel;
 import com.github.ambry.utils.Pair;
 import com.mysql.cj.exceptions.MysqlErrorNumbers;
 import java.sql.BatchUpdateException;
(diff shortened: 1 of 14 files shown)


ambry-mysql/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
```diff
@@ -177,6 +181,10 @@ public synchronized Connection getDatabaseConnection(boolean needWritable) throw
 
     // If the active connection is good and it's the best endpoint, keep it.
     if (activeConnection != null && !isBetterEndpoint(bestEndpoint, connectedEndpoint)) {
+      if (config != null && !config.transactionIsolationLevel.equals(TransactionIsolationLevel.TRANSACTION_NONE)) {
+        activeConnection.setTransactionIsolation(
```
Arun-LinkedIn, line 185: May be we can set the transaction isolation level immediately after creation connection (after below code). In that way, we can have it just in one place? 

```
 Connection connection = mysqlDriver.connect(candidateEndpoint.getUrl(), credentials);
```

Arun-LinkedIn (approved):
LGTM. Just had one comment.

ambry-mysql/src/main/java/com/github/ambry/mysql/MySqlDataAccessor.java
SophieGuo410: yeah, we can do that, the reason I set to two place is due to if the connection is still active, we are still able to set the transaction isolation level, but since the connection will be closed eventually and go through the connectToBestAvailableEndpoint logic, so I guess it should be fine.

Arun-LinkedIn (approved):

JingQianCloud (approved):

PR #2588, opened 2023-09-22
Server: check size of prefetched data

justinlin-linkedin:
## Summary
After prefetch data from file, check the size of the actual returned data with the desired size, and make sure that are equal, otherwise, throw an error.

### DIFF
--- a/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatSend.java
+++ b/ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatSend.java
@@ -22,6 +22,7 @@ import com.github.ambry.store.StoreKeyFactory;
 import com.github.ambry.utils.AbstractByteBufHolder;
 import com.github.ambry.utils.ByteBufferInputStream;
 import com.github.ambry.utils.ByteBufferOutputStream;
+import com.github.ambry.utils.NettyByteBufDataInputStream;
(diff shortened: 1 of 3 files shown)


ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatSend.java
```diff
@@ -129,7 +130,7 @@ private void fetchDataFromReadSet(MessageFormatMetrics metrics) throws MessageFo
           // If we can successfully deserialize BlobAll, then the message content is correct
           try {
             ByteBuf messageData = readSet.getPrefetchedData(i).duplicate();
-            MessageFormatRecord.deserializeBlobAll(new ByteBufInputStream(messageData), storeKeyFactory);
+            MessageFormatRecord.deserializeBlobAll(new NettyByteBufDataInputStream(messageData), storeKeyFactory);
```
JingQianCloud, line 133: What's the reason that we want to change the InputStream type?

justinlin-linkedin: to avoid one extra copy. If we use NettyByteBufDataInputStream, we don't have to copy 4MB of data chunk out to a java ByteBuffer to to calculate crc.

JingQianCloud (approved):

Arun-LinkedIn (approved):

PR #2591, opened 2023-09-22
[vcr-2.0] Impl UNDELETE

snalli:


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -393,7 +393,56 @@ public CompletableFuture<Boolean> deleteBlobAsync(BlobId blobId, long deletionTi
...
+    String blobIdStr = blobLayout.blobFilePath;
+    BlobProperties blobProperties = getBlobProperties(blobLayout);
+    Map<String, String> cloudMetadata = blobProperties.getMetadata();
+    // lifeVersion must always be present
+    short cloudlifeVersion = Short.parseShort(cloudMetadata.get(CloudBlobMetadata.FIELD_LIFE_VERSION));
+    // Below is the correct behavior. For ref, look at BlobStore::undelete and ReplicaThread::applyUndelete
+    if (cloudlifeVersion == lifeVersion && cloudMetadata.containsKey(CloudBlobMetadata.FIELD_DELETION_TIME)
+        && Long.parseLong(cloudMetadata.get(CloudBlobMetadata.FIELD_DELETION_TIME)) == Utils.Infinite_Time) {
+      String error = String.format("Cannot undelete id %s as it is undeleted in cloud", blobIdStr);
+      throw new CloudStorageException(error, new StoreException(error, StoreErrorCodes.ID_Undeleted));
```
Arun-LinkedIn, line 407: Similar comments as delete PR. Would it be okay to return success response here instead of throwing an exception since this is a valid scenario which could happen due to concurrent updates.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -393,7 +393,56 @@ public CompletableFuture<Boolean> deleteBlobAsync(BlobId blobId, long deletionTi
...
+    }
+    // Don't rely on the CloudBlobStore.recentCache to do the "right" thing.
+    // Below is the correct behavior. For ref, look at BlobStore::undelete and ReplicaThread::applyUndelete
+    if (cloudlifeVersion > lifeVersion) {
+      String error = String.format("Cannot undelete id %s as it has a higher lifeVersion than the message info: %s > %s",
+          blobIdStr, cloudlifeVersion, lifeVersion);
+      throw new CloudStorageException(error, new StoreException(error, StoreErrorCodes.Life_Version_Conflict));
```
Arun-LinkedIn, line 414: Same here. Would returning success be correct here since this can happen when a VCR tries to replicate from a lagging replica?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
Arun-LinkedIn: In case we are going with throwing exception, hopefully it is caught and replication is able to proceed.

Arun-LinkedIn (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: Yes, it definitely is. 

```
  protected void applyUndelete(MessageInfo messageInfo, RemoteReplicaInfo remoteReplicaInfo) throws StoreException {
    DataNodeId remoteNode = remoteReplicaInfo.getReplicaId().getDataNodeId();
    try {
      messageInfo = new MessageInfo.Builder(messageInfo).isUndeleted(true).isDeleted(false).build();
      remoteReplicaInfo.getLocalStore().undelete(messageInfo);
      logger.trace("Remote node: {} Thread name: {} Remote replica: {} Key undelete id: {}", remoteNode, threadName,
          remoteReplicaInfo.getReplicaId(), messageInfo.getStoreKey());
    } catch (StoreException e) {
      // The blob may be undeleted, which is alright
      if (e.getErrorCode() == StoreErrorCodes.Life_Version_Conflict
          || e.getErrorCode() == StoreErrorCodes.ID_Undeleted) {
        logger.trace("Remote node: {} Thread name: {} Remote replica: {} Key {}: {}", remoteNode, threadName,
            remoteReplicaInfo.getReplicaId(), messageInfo.getStoreKey(), e.getErrorCode().name());
      } else {
        throw e;
      }
    }
    // A Repair event for an undelete signifies that an undelete message was received from the remote and it is fired
    // as long as the undelete is guaranteed to have taken effect locally.
    if (notification != null) {
      notification.onBlobReplicaUndeleted(dataNodeId.getHostname(), dataNodeId.getPort(),
          messageInfo.getStoreKey().getID(), BlobReplicaSourceType.REPAIRED);
    }
  }
```

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: ReplicaThread does all these checks and we should almost never hit this . Throwing an error then makes more sense and it is caught in applyUndelete(). Please see ReplicaThread::applyUndelete

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -408,7 +408,75 @@ public CompletableFuture<Boolean> deleteBlobAsync(BlobId blobId, long deletionTi
...
+          Cloud is either ahead of server or caught up.
+         */
+        // lifeVersion must always be present
+        short cloudlifeVersion = Short.parseShort(cloudMetadata.get(CloudBlobMetadata.FIELD_LIFE_VERSION));
+        /*
+          if cloudLifeVersion == replicaLifeVersion && deleteTime absent in cloudMetatadata, then something is wrong. Throw Life_Version_Conflict.
+          if cloudLifeVersion == replicaLifeVersion && deleteTime != -1, then something is wrong. Throw Life_Version_Conflict.
+         */
+        if (cloudlifeVersion == lifeVersion && cloudMetadata.containsKey(CloudBlobMetadata.FIELD_DELETION_TIME)
+            && Long.parseLong(cloudMetadata.get(CloudBlobMetadata.FIELD_DELETION_TIME)) == Utils.Infinite_Time) {
```
justinlin-linkedin, line 435: could it be because other peers have already synced the undelete to ABS. If so, we don't have to throw an exception here, but just treat it as a no-op.

snalli: Yes, if we are here it means other threads (or may be peers) got to the cloud first. However, replicaThread swallows these  exc and previous backup code also does the same thing.

justinlin-linkedin (approved):

PR #2593, opened 2023-09-25
Dont increase refCnt in Utils methods

justinlin-linkedin:
## Summary
When using NettyByteBufDataInputStream, we would increase the refCnt when calling readNettyByteBufFromCrcInputStream in Utils class. This is a bad design as an utility function should never increase the reference counter. It's very hard to know that refCnt is changed here. 

This PR would fix that.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
@@ -880,9 +880,14 @@ class GetBlobOperation extends GetOperation {
             routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
             ByteBuf decryptedContent = result.getDecryptedBlobContent();
             ByteBuf decompressedContent = decompressContent(decryptedContent);
+            // If the decompressed content is valid, don't use decryptedContent anymore, so we release it.
+            // If the decompressed content is null, which means the operation failed, just release the decryptedContent.
+            decryptedContent.release();
             if (decompressedContent != null) {
               chunkIndexToBuf.put(chunkIndex, filterChunkToRange(decompressedContent));
               numChunksRetrieved.incrementAndGet();
+            } else {
+              progressTracker.setCryptoJobFailed();
             }
           } else {
             // only in maybeReleaseDecryptionResultBuffer() will result be set to null, this means the bytebuf has been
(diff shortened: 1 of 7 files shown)


ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -1112,25 +1115,25 @@ void handleBody(InputStream payload, MessageMetadata messageMetadata, MessageInf
...
+        boolean launchedJob = maybeLaunchCryptoJob(chunkBuf, null, encryptionKey, chunkBlobId);
+        if (!launchedJob) {
+          if (chunkIndex == 0) {
+            routerMetrics.getFirstDataChunkLatencyMs.update(System.currentTimeMillis() - submissionTimeMs);
+          }
+          routerMetrics.getDataChunkLatencyMs.update(System.currentTimeMillis() - initializedTimeMs);
+          ByteBuf decompressedContent = decompressContent(chunkBuf);
+          // If the decompressed content is valid, don't use chunkBuf anymore, so we release it.
+          // If the decompressed content is null, which means the operation failed, just release the chunkBuf.
+          chunkBuf.release();
```
Arun-LinkedIn, line 1130: Just a thought. Since `decompressContent()` method is the one who is creating a new ByteBuf for storing decompressed content, would it be better if it releases the passed input ByteBuf?

Arun-LinkedIn (approved):
Looks good to me. Nice catch :)

ambry-router/src/main/java/com/github/ambry/router/GetBlobOperation.java
```diff
@@ -914,12 +916,12 @@ protected void maybeProcessCallbacks() {
      */
     protected ByteBuf decompressContent(ByteBuf sourceBuffer) {
       if (!isChunkCompressed) {
-        // Blob is not compressed.  Return source buffer.
+        // Blob is not compressed.  Return source buffer and increase the counter.
```
JingQianCloud, line 919: Shall we call sourceBuffer.release here since we add refer count before calling this function?

justinlin-linkedin: we don't want to release source buffer since we are going to use source buffer later now that the chunk is not compressed.

JingQianCloud (approved):

PR #2594, opened 2023-09-26
Support deleted_ts for auto_purge

SophieGuo410:
Due to previously we have delete_ts, but auto_purge only recognize deleted_ts, so we have to do this migration.
1. Add one more column for deleted_ts.
2. update the logic in our frontend to write to both deleted_ts and delete_ts, and read the max(deleted_ts, delete_ts)
3. Do one time migration to migrate all delete_ts to deleted_ts if it's null.
4. Delete the delete_ts column after migration is done.
**This pr is to support step2, and need to be merged once step1 is done.**

The pr has three major changes:
1.When update the delete_ts, we will also update same value for deleted_ts.
2.Since we do step1, so when we get the delete_ts and deleted_ts, there will be two cases
    2.1 delete_ts has value and deleted_ts is null.
    2.2 delete_ts has same value(or null) as deleted_ts.
So calling the COALESCE(%2$s, %8$s) will get the real delete timestamp we want as it will return the first non-null value.
3.when we use select, we get both delete_ts and deleted_ts and use below cmd to get the latest timestamp, as it will return the first non-null value as well.
`deletionTime = getLatestTimeStamp(resultSet.getTimestamp(DELETE_TS), resultSet.getTimestamp(DELETED_TS))`

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
@@ -106,13 +106,16 @@ public class DatasetDao {
     this.dataAccessor = dataAccessor;
     this.mySqlAccountServiceConfig = mySqlAccountServiceConfig;
     this.metrics = metrics;
+    //adding a dataset to db, we only support permanent dataset, and the only way to delete dataset is by calling deleteDatasetByIdSql
     insertDatasetSql =
         String.format("insert into %s (%s, %s, %s, %s, %s, %s, %s, %s, %s) values (?, ?, ?, ?, now(3), ?, ?, ?, ?)",
             DATASET_TABLE, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_POLICY,
             RETENTION_COUNT, RETENTION_TIME_IN_SECONDS, USER_TAGS);
+    //if dataset is expired, we can create a dataset with same primary key, but we have to use update instead of insert.
+    //need to make sure the delete_ts is set back to null as the only way to delete a dataset is by calling deleteDatasetByIdSql.
     updateDatasetIfExpiredSql = String.format(
         "update %1$s set %2$s = ?, %3$s = now(3), %4$s = ?, %5$s = ?, %6$s = ?, %7$s = ?, %8$s = NULL, %12$s = NULL "
-            + "where %9$s = ? and %10$s = ? and %11$s = ? and GREATEST(%8$s, %12$s) < now(3)", DATASET_TABLE,
+            + "where %9$s = ? and %10$s = ? and %11$s = ? and COALESCE(%8$s, %12$s) < now(3)", DATASET_TABLE,
         VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_POLICY, RETENTION_COUNT, RETENTION_TIME_IN_SECONDS, USER_TAGS,
         DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME, DELETED_TS);
     //update the dataset field, in order to support partial update, if one parameter is null, keep the original value.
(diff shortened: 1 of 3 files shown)


ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -110,74 +111,80 @@ public DatasetDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
             DATASET_TABLE, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_POLICY,
             RETENTION_COUNT, RETENTION_TIME_IN_SECONDS, USER_TAGS);
     updateDatasetIfExpiredSql = String.format(
```
JingQianCloud, line 113: Do you mind add one line comment for each sql prepared statement?

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -110,74 +111,80 @@ public DatasetDao(MySqlDataAccessor dataAccessor, MySqlAccountServiceConfig mySq
...
-        "update %s set %s = ?, %s = now(3), %s = ?, %s = ?, %s = ?, %s = ?, %s = NULL where %s = ? and %s = ? and %s = ? and delete_ts < now(3)",
-        DATASET_TABLE, VERSION_SCHEMA, LAST_MODIFIED_TIME, RETENTION_POLICY, RETENTION_COUNT,
-        RETENTION_TIME_IN_SECONDS, USER_TAGS, DELETE_TS, ACCOUNT_ID, CONTAINER_ID, DATASET_NAME);
+        "update %1$s set %2$s = ?, %3$s = now(3), %4$s = ?, %5$s = ?, %6$s = ?, %7$s = ?, %8$s = NULL, %12$s = NULL "
```
JingQianCloud, line 114: It's not about this PR. Just for this statement,
1. why do we set deleted_ts to NULL?
2. here we assume deleted_ts is not NULL. What if deleted_ts and delete_ts are both null?

JingQianCloud (commented):
For the data migration purpose,
I think for "get" type of query, right now we need read both "deleted_ts" and "delete_ts".
But for update/insert type of statement, I think we don't need write to the "delete_ts" any more. We can just write to the "deleted_ts", right?

SophieGuo410:
> For the data migration purpose, I think for "get" type of query, right now we need read both "deleted_ts" and "delete_ts". But for update/insert type of statement, I think we don't need write to the "delete_ts" any more. We can just write to the "deleted_ts", right? For read, we allow either deleted_ts or delete_bs is NULL.

I feel like setting both delete_ts and deleted_ts is easier for the read. Basically we don't need to check the max value between delete_ts and deleted_ts, and having the delete_ts and deleted_ts different we will need to think about more condition like delete_ts=null, deleted_ts=value, or delete_ts = small_value and deleted_ts = large_value.... By setting both delete_ts and deleted_ts, it's easy for us to detect if delete_ts and deleted_ts is same, we don't need to worry about migration because we know the deleted_ts is latest states, and if delete_ts and deleted_ts is different, we know the actual value we need to pick is delete_ts. This also makes the one time migration easier(step 3). 
update table set deleted_ts = COALESCE(deleted_ts, delete_ts)

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
SophieGuo410: Sure, I'll add the comment for each sql.

ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
SophieGuo410: This statement is used when dataset is expired but has not been autopurged yet. So when user create a dataset with same datatset name, we should create a new dataset without setting any delete_ts or deleted_ts, since we don't support ttl for dataset(just for dataset version). The only way to delete the dataset is to call the deleteDataset API. You can see the addDataset will not update the delete_ts either. In summary, addDataset and updateDatasetIfExpired will make sure the delete_ts and deleted_ts in non-null.

JingQianCloud (commented):
Overall it looks good to me.
I'm not sure if you have time before the leave.
Can we add unit test specifically for the migration?   Insert/Update records and compare the "deleted_ts" and "delete_ts".
Don't worry if you don't have time for it.

SophieGuo410:
> Overall it looks good to me. I'm not sure if you have time before the leave. Can we add unit test specifically for the migration? Insert/Update records and compare the "deleted_ts" and "delete_ts". Don't worry if you don't have time for it.

I'll try to see if I can add some test. And we shouldn't use GREATEST, we should use COALESCE. Somehow I forget to update that one. Will fix it.

JingQianCloud:
Close this PR. 
Submitted another one

PR #2607, opened 2023-10-04
Warning when the number of replicas are decreased

justinlin-linkedin:
Print out some warning if the number of replicas from the static clustermap is less than the number of replicas in helix.


MikeDafi (approved):
LGTM

Arun-LinkedIn (approved):

PR #2611, opened 2023-10-09
[vcr-2.0] Fix updateTTL and logging issues

snalli:
This patch fixes updateTTL issues from prev backup code and a few logging issues detailed below.

1. Replication: When repl-layer receives PUT+TTL together, it applies PUT to store a blob with expiry = -1 followed by updating its expiry to -1. This is redundant and was caught by the new backup code. It floods the logs with unnecessary messages. We can't touch repl-layer at this point so we just suppress the logs.
2. Backup code V1: The previous backup code reported a blob was not ttl-updated even though it was. This sent duplicate writes to Azure and was caught by the new backup code. This is not only wrong but also floods the logs with unnecessary messages.
3. Azure SDK: The SDK prints out unnecessary warnings when searching for an absent blob in Azure Storage. The issue is confirmed by MSFT and they'll fix it in their next release. Additionally, we also print some unnecessary messages when we encounter blob_not_found. This patch eliminates those as well.

MSFT issue: https://github.com/Azure/azure-sdk-for-java/issues/37117

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
@@ -14,7 +14,14 @@
 package com.github.ambry.cloud;
 
 import com.github.ambry.config.CloudConfig;
+import com.github.ambry.store.Store;
+import com.github.ambry.store.StoreErrorCodes;
+import com.github.ambry.store.StoreException;
 import com.github.ambry.utils.Utils;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -86,9 +93,16 @@ public class CloudRequestAgent {
         vcrMetrics.retryWaitTimeMsec.inc(delay);
       } else {
         // Either not retryable or exhausted attempts.
-        // Not very useful to log error here as they are printed in several other places
-        logger.trace("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
-            statusCode, cause, attempts);
+        Throwable t = cse.getCause();
+        Set<StoreErrorCodes> errorCodes = new HashSet<>(Arrays.asList(StoreErrorCodes.Already_Updated));
+        if (t instanceof StoreException && errorCodes.contains(((StoreException)t).getErrorCode())) {
+          // Not very useful to log error here as they are printed in several other places
+          logger.trace("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
+              statusCode, cause, attempts);
+        } else {
+          logger.error("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
+              statusCode, cause, attempts);
+        }
         throw cse;
       }
     } else {
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
@@ -87,7 +87,7 @@ public class CloudRequestAgent {
       } else {
         // Either not retryable or exhausted attempts.
         // Not very useful to log error here as they are printed in several other places
-        logger.debug("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
+        logger.trace("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
             statusCode, cause, attempts);
         throw cse;
       }
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
@@ -86,7 +86,8 @@ public class CloudRequestAgent {
         vcrMetrics.retryWaitTimeMsec.inc(delay);
       } else {
         // Either not retryable or exhausted attempts.
-        logger.error("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
+        // Not very useful to log error here as they are printed in several other places
+        logger.debug("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
             statusCode, cause, attempts);
         throw cse;
       }
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
@@ -85,8 +85,11 @@ public class AzureCloudDestinationSync implements CloudDestination {
     /*
      * These are the configs to be changed for vcr-2.0
      *
-     *    azureCloudConfig.azureNameSchemeVersion = 1
      *    azureCloudConfig.azureBlobContainerStrategy = PARTITION
+     *    azureCloudConfig.azureNameSchemeVersion = 1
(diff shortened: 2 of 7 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
```diff
@@ -86,7 +86,8 @@ private void throwOrDelay(Throwable e, String actionName, String partitionPath,
         vcrMetrics.retryWaitTimeMsec.inc(delay);
       } else {
         // Either not retryable or exhausted attempts.
-        logger.error("{} failed partition {} statusCode {} cause {} made {} attempts.", actionName, partitionPath,
+        // Not very useful to log error here as they are printed in several other places
```
JingQianCloud, line 89: It'll hide all exceptions.
If you want to avoid ttl updated exception log, please do it for the ttl update exception. Not all exceptions.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -330,6 +328,10 @@ protected BlobProperties getBlobProperties(AzureBlobLayoutStrategy.BlobLayout bl
     try {
       return createOrGetBlobStore(blobLayout.containerName).getBlobClient(blobLayout.blobFilePath).getProperties();
     } catch (BlobStorageException bse) {
+      if (bse.getErrorCode() == BlobErrorCode.BLOB_NOT_FOUND) {
```
JingQianCloud, line 331: I suggest we still throw BLOB_NOT_FOUND, just let getBlobMetadata handle it in the right way.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -597,14 +614,16 @@ public Map<String, CloudBlobMetadata> getBlobMetadata(List<BlobId> blobIds) thro
...
-      } catch (Throwable t) {
-        if (t instanceof CloudStorageException
-            && ((CloudStorageException) t).getStatusCode() == CloudBlobStore.STATUS_NOT_FOUND) {
-          // We should never be here because replication logic checks if a blob exists or not before undeleting it.
-          String dbg = String.format("Blob %s absent in Azure blob storage", blobLayout);
-          logger.error(dbg);
+        // We are trying to find blobs present on servers but absent in cloud. Many PUTs fall into this bucket.
+        // Do not print any errors as this is expected when the servers are ahead of cloud and cloud is a bit stale.
+        // blobProperties == null when a blob is absent in cloud.
+        if (blobProperties != null) {
```
JingQianCloud, line 620: Just let getBlobProperties continues to throw NOT_FOUND as most of our code does.
Here if the exception is not_FOUND, swallow it or print with trace level.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudRequestAgent.java
snalli: done

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: done

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: done

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -598,13 +624,19 @@ public Map<String, CloudBlobMetadata> getBlobMetadata(List<BlobId> blobIds) thro
...
-        if (t instanceof CloudStorageException
-            && ((CloudStorageException) t).getStatusCode() == CloudBlobStore.STATUS_NOT_FOUND) {
-          // We should never be here because replication logic checks if a blob exists or not before undeleting it.
-          String dbg = String.format("Blob %s absent in Azure blob storage", blobLayout);
-          logger.error(dbg);
+      } catch (CloudStorageException cse) {
+        if (cse.getCause() instanceof BlobStorageException &&
+            ((BlobStorageException) cse.getCause()).getErrorCode() == BlobErrorCode.BLOB_NOT_FOUND) {
+          /*
+            We mostly get here from findMissingKeys, and we will encounter many blobs missing from cloud before we upload them.
```
Arun-LinkedIn, line 631: I agree with you that this is happening because findMissingKeys() is using `getBlobMetadata` instead of `exists()` to check if blob is present or that. But, I just feel that ignoring `BLOB_NOT_FOUND` exception here assuming that it is always called from `findMissingKeys()` may be not correct. If we look at this method independently, `BLOB_NOT_FOUND` seems to be a valid exception (For example, if this method is used to get blob metadata in recovery path, this exception could be valid).

If it is flooding the logging, would it be possible (in future, if not now) to change the log type to `trace` in this method and log the `error` in caller methods?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -524,21 +535,36 @@ public short updateBlobExpiration(BlobId blobId, long expirationTime, CloudUpdat
...
+          For ref, look at BlobStore::updateTTL and ReplicaThread::applyTtlUpdate.
+          ReplicaThread::applyUpdatesToBlobInLocalStore does all necessary checks before calling updateTTL however,
+          ReplicaThread::handleGetResponse does not.
+
+          We can come here from ReplicaThread::handleGetResponse L1258 that applies PUT+TTL without checking if the blob
+          uploaded is permanent. The validator reports the blob has already been ttl-updated, and we end up flooding the logs
+          and incrementing metrics. To avoid this, just don't inc any error metric or print any logs. However, this would
+          mask a scenario where we are erroneously trying to update the ttl of permanent blob. This is ok because we prevent
+          an unnecessary request to cloud without sacrificing any correctness. The ReplicaThread should actually check before
+          applying a ttl-update, but it was written for disk-based log, and not a cloud-based backup system.
```
Arun-LinkedIn, line 557: Since we are replicating from multiple replicas in separate threads, I am just thinking that this situation is unavoidable even if a replica thread checks the blob before applying a ttl-update.

Arun-LinkedIn (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: `getBlobProperties` logs the trace message for blob-not-found. I can log it here in `getBlobMetadata` as well but it'd just be a duplicate.  However, I cannot throw the blob-not-found err as this fn accepts a list of blobs and must loop through the entire list. 

Overall, the design and structure of current code is restricted by the choices made by previous authors. There is quite a bit inconsistency. Some fns accept a list while others accept a single item. Some return boolean, others int and others void.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: Is there a change you are asking me to make ? 

You are referring to a concurrent update which is possible but less probable (and harmless) than the PUT+TTL case mentioned here. PUT+TTL  _always_ falls in this branch. I have tested this for two weeks in ei. As mentioned, the fix should be in the repl-layer to not apply ttl-update if the expiry is already -1.

JingQianCloud (approved):

snalli:
Merging as i got two approvals

PR #2616, opened 2023-10-13
Filter out DNS fields from certificate in public access log

justinlin-linkedin:
## Summary

We are filtering out all the DNS fields from certificate in the public access log. We don't need those information.

### DIFF
--- a/ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
+++ b/ambry-rest/src/main/java/com/github/ambry/rest/PublicAccessLogHandler.java
@@ -26,10 +26,12 @@ import io.netty.handler.ssl.SslHandler;
 import java.security.cert.Certificate;
 import java.security.cert.X509Certificate;
 import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
 import javax.net.ssl.SSLEngine;
 import javax.security.auth.x500.X500Principal;
-import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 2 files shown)


JingQianCloud:
Do you mind adding one example string, before the PR and after the PR, in the PR summary?

litingulfs (changes requested):
It looks to me like pretty much all of the `san` entries in the SSL log message are of the form:

```
san=[[2, <linkedin server hostname>]
```
and you're removing these, in order to reduce the public access log size so we can get inLogs to keep ingesting that log? I think that's a good idea, so if you can update the unit test at https://github.com/linkedin/ambry/blob/master/ambry-rest/src/test/java/com/github/ambry/rest/PublicAccessLogHandlerTest.java#L283 then I'll be happy to approve this PR.

We might even consider taking this a step further and either drop the entire SSL log statement, or put the `logSSLInfo()` invocation behind a config that we can turn on if we think we need it. It's never been quite clear to me how useful these SSL logs really are.

I am not 100% sure, but I think the SSL log may include the LI service principal, which is usually printed later on in the access log under the x-ambry-service-id header anyway:

```
 SSL ([used=true], [principal=C=US, ST=California, L=Mountain View, O=LinkedIn, OU=Security, CN=lva1-app104007.prod.linkedin.com], [san=[[2, lva1-app104007.prod.linkedin.com], [6, urn:li:servicePrincipal(nearline-recipient-features-fedex-workflow;None;i001)], [6, urn:li:servicePrincipalMetadata(prod-lva1;1.0.0)]]])
```

litingulfs (approved):
LGTM

JingQianCloud (approved):

PR #2620, opened 2023-10-17
ReplicaSyncupManager: add separate target for downward state transition

justinlin-linkedin:
Adding a separate catch up target for downward state transition.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
@@ -316,12 +316,14 @@ public class AmbryReplicaSyncUpManager implements ReplicaSyncUpManager {
           remoteDcPeerReplicaAndLag.put(peerReplica, Long.MAX_VALUE);
         }
       }
-      catchupTarget = clusterMapConfig.clustermapReplicaCatchupTarget == 0 ? localDcPeerReplicaAndLag.size()
-          : clusterMapConfig.clustermapReplicaCatchupTarget;
+      int defaultCatchUpTarget =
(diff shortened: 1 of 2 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -316,12 +316,14 @@ private class LocalReplicaLagInfos {
           remoteDcPeerReplicaAndLag.put(peerReplica, Long.MAX_VALUE);
         }
       }
-      catchupTarget = clusterMapConfig.clustermapReplicaCatchupTarget == 0 ? localDcPeerReplicaAndLag.size()
-          : clusterMapConfig.clustermapReplicaCatchupTarget;
+      int defaultCatchUpTarget =
+          currentState == ReplicaState.BOOTSTRAP ? clusterMapConfig.clustermapReplicaCatchupTarget
+              : clusterMapConfig.clustermapReplicaCatchupTargetForDownwardTransition;
+      catchupTarget = Math.max(defaultCatchUpTarget, localDcPeerReplicaAndLag.size());
```
ankagrawal, line 322: This changes the previous logic a little bit for "upward transition". Previously for upward transition, `catchupTarget` value could have been 2, while local dc peer replica size could have been 3. With this change, it will always be atleast 3. Is that intended?

justinlin-linkedin: it will not always be 3. For upward state transition, aka bootstrap, it still 2. You can set this catchupTarget to 2 and then there are 2 other leader and standby peer replicas in the local datacenter, then it will be 2.

ankagrawal (approved):

Arun-LinkedIn (approved):

PR #2621, opened 2023-10-17
Add a dedicated scheduler for persisting index segments

Arun-LinkedIn:


ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -127,6 +129,7 @@ public class DiskManager {
     this.time = time;
     diskIOScheduler = new DiskIOScheduler(getThrottlers(storeConfig, time));
     longLivedTaskScheduler = Utils.newScheduler(1, true);
+    indexPersistScheduler = Utils.newScheduler(1, "index-persistor-for-disk-" + disk.getMountPath(), false);
```
justinlin-linkedin, line 132: you have to shutdown this scheduler in the shutdown method.

justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
Arun-LinkedIn: Thank you for catching it. Updated the PR.

JingQianCloud (approved):
LGTM

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -86,6 +87,8 @@ public class DiskManager {
   private boolean running = false;
   private DiskHealthStatus diskHealthStatus;
   private final DiskHealthCheck diskHealthCheck;
+  // Have a dedicated scheduler for persisting index segments to ensure index segments are always persisted
```
ankagrawal, line 90: Just curious if we saw degradation with index persisting. Was the previous scheduler overused due to `diskFailureHandler`?
`RemoteTokenTracker` seems to be using the previous shared scheduler as well. Do we think that it can be impacted too?

Arun-LinkedIn: Yeah, based on thread dumps, we are suspecting the all the threads in previous scheduler are busy collecting stats (Running `ValidDataSizeCollector` task). We are checking whether we really need this task. We mainly noticed that index segments are not persisted at times. This is mainly like a temporary workaround for persisting index segments. Need to see if `RemoteTokenTracker` is affected too

PR #2626, opened 2023-10-20
Search duplicate resources in clustermap when partition cant be found in given resource

justinlin-linkedin:
## Summary

In helix, when we add new resources to ideal state, it will take some time to populate that information to aggregated external view. Within this time, helix cluster manager would know the resources for partitions, but getReplicaIdsByState would fail because the routing table hasn't updated yet. 

This PR would fix this issue by searching in old resources to find the replica ids.

## Test
Unit test

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -1181,6 +1181,12 @@ public class HelixClusterManager implements ClusterMap {
         if (replicas.size() != replicasBefore) {
           continue;
         }
+        logger.trace("Can't find any replicas in dc {} for partition {} at resources {} at state {}", dc, partition,
+            resourceNames, state);
+        if (!partitionToDuplicateResourceNameByDc.get(dc).containsKey(partitionName)) {
+          logger.trace("No duplicate resource found for partition {} in dc {}, continue", partitionName, dc);
+          continue;
+        }
         Map<String, ExternalView> externalViewMap =
             clusterMapConfig.clusterMapUseAggregatedView ? globalResourceToExternalView
                 : dcToResourceToExternalView.get(dc);
@@ -1188,9 +1194,12 @@ public class HelixClusterManager implements ClusterMap {
         if (!resourceNames.stream()
             .anyMatch(
                 resource -> externalViewHasInstanceForPartition(externalViewMap.get(resource), dc, partitionName))) {
+          String duplicateResource = partitionToDuplicateResourceNameByDc.get(dc).get(partitionName);
+          logger.trace("Partition {} doesn't exist in resources {} in dc {}, try duplicate resource {}", partitionName,
+              resourceNames, dc, duplicateResource);
           helixClusterManagerMetrics.resourceNameMismatchCount.inc();
-          getReplicaIdsByStateInRoutingTableSnapshot(routingTableSnapshot, dc,
-              partitionToDuplicateResourceNameByDc.get(dc).get(partitionName), partitionName, state, replicas);
+          getReplicaIdsByStateInRoutingTableSnapshot(routingTableSnapshot, dc, duplicateResource, partitionName, state,
+              replicas);
         }
       }
       logger.debug("Replicas for partition {} with state {} in dc {} are {}. Query time in Ms {}", partitionName,
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
@@ -1171,34 +1171,76 @@ public class HelixClusterManager implements ClusterMap {
           // We found the replicas from the resource name, just continue to next dc.
           continue;
         }
-        // There is some change made to resources in IDEALSTATES and it's not yet reflected in EXTERNAALVIEW.
-        // Try to search the partition in all resources
-        helixClusterManagerMetrics.resourceNameMismatchCount.inc();
-        String resourceNameInCache =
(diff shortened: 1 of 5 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1153,13 +1154,31 @@ public List<AmbryReplica> getReplicaIdsByState(AmbryPartition partition, Replica
...
                 : dcToRoutingTableSnapshotRef.get(dc).get();
+        Collection<String> resourcesFromRoutingTable = routingTableSnapshot.getResources();
         routingTableSnapshot.getInstancesForResource(resourceName, partition.toPathString(), state.name())
             .stream()
             .map(instanceConfig -> instanceNameToAmbryDataNode.get(instanceConfig.getInstanceName()))
             .filter(dataNode -> dataNode.getDatacenterName().equals(dc))
             .map(dataNode -> ambryDataNodeToAmbryReplicas.get(dataNode).get(partition.toPathString()))
             .filter(Objects::nonNull)
             .forEach(replicas::add);
+        if (replicas.isEmpty()) {
```
Arun-LinkedIn, line 1165: Nice Idea! One doubt. Since `replicas` is a set of replicas from all dcs, we may not know if replicas in a specific dc returned empty list. Should we have a new set like `currDCReplicas` and check if it is not empty to enter below logic?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1153,13 +1154,31 @@ public List<AmbryReplica> getReplicaIdsByState(AmbryPartition partition, Replica
...
+            routingTableSnapshot.getInstancesForResource(resource, partition.toPathString(), state.name())
+                .stream()
+                .map(instanceConfig -> instanceNameToAmbryDataNode.get(instanceConfig.getInstanceName()))
+                .filter(dataNode -> dataNode.getDatacenterName().equals(dc))
+                .map(dataNode -> ambryDataNodeToAmbryReplicas.get(dataNode).get(partition.toPathString()))
+                .filter(Objects::nonNull)
+                .forEach(replicas::add);
+            if (!replicas.isEmpty()) {
+              // We found the partition
+              break;
```
Arun-LinkedIn, line 1178: If we break out of for loop, we may not collect replicas in other dcs?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
justinlin-linkedin: Comparing with the size of the list now.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
justinlin-linkedin: it will break out the for loop to search resources, not the dcs.

Arun-LinkedIn (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1148,26 +1152,60 @@ public List<AmbryReplica> getReplicaIdsByState(AmbryPartition partition, Replica
...
-        routingTableSnapshot.getInstancesForResource(resourceName, partition.toPathString(), state.name())
-            .stream()
-            .map(instanceConfig -> instanceNameToAmbryDataNode.get(instanceConfig.getInstanceName()))
-            .filter(dataNode -> dataNode.getDatacenterName().equals(dc))
-            .map(dataNode -> ambryDataNodeToAmbryReplicas.get(dataNode).get(partition.toPathString()))
-            .filter(Objects::nonNull)
-            .forEach(replicas::add);
+        getReplicaIdsByStateInRoutingTableSnapshot(routingTableSnapshot, dc, resourceName, partitionName, state,
+            replicas);
+        if (replicas.size() != replicaSizeBefore) {
```
ankagrawal, line 1164: Should this be `replicas.size() > replicaSizeBefore` to be more precise?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -1148,26 +1152,60 @@ public List<AmbryReplica> getReplicaIdsByState(AmbryPartition partition, Replica
...
+                replicas);
+            if (replicas.size() != replicaSizeBefore) {
+              // We found the partition in this resource, put this resource in the case.
+              dcToPartitionNameToResourceCache.get(dc).putIfAbsent(partitionName, resource);
+              break;
+            }
+          }
+        } else {
+          helixClusterManagerMetrics.resourceNameCacheHitCount.inc();
+          getReplicaIdsByStateInRoutingTableSnapshot(routingTableSnapshot, dc, resourceNameInCache, partitionName,
```
ankagrawal, line 1188: Just curious, after the move to full-auto is complete, would this case happen that a partition id is moved from one resource to another?

Arun-LinkedIn (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
justinlin-linkedin: we are only adding to the list, it should be fine.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
justinlin-linkedin: I updated the PR to reflect the case where a partition is moved to a different resource. Basically we want to allow multiple resources for any given partition as long as the instance list is different.

ankagrawal (approved):

PR #2629, opened 2023-10-24
Swallow all exceptions for background deleter.

JingQianCloud:
For background deleter, we swallow any exception including server unavailable.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
@@ -460,13 +460,22 @@ class DeleteOperation {
           }
         }
 
-        // ignore the NOT_FOUND error for the background deleter.
+        // Handle the exceptions of the background deleter
         if (serviceId != null && serviceId.startsWith(BackgroundDeleteRequest.SERVICE_ID_PREFIX)
-            && operationException.get() != null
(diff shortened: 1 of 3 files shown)


ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -460,13 +460,22 @@ record =
...
+          RouterErrorCode code = ((RouterException) operationException.get()).getErrorCode();
+          if (code == RouterErrorCode.BlobDoesNotExist) {
+            // NOT_FOUND is not an error for the background deleter.
+            routerMetrics.backgroundDeleterNotFoundCount.inc();
+            logger.info("Ignore BlobNotFound error from the background deleter. {}", blobId);
+          } else {
+            // We may hit AmbryUnavailable or other exceptions.
+            // Since we don't do any further thing for it anyway, swallow the exception.
+            // Just emit with metrics and log error message
+            routerMetrics.backgroundDeleterNotAvailableCount.inc();
```
ankagrawal, line 475: This is not just unavailable count. We should either make it a generic metric, or handle the case for not available error.

ankagrawal (approved):

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
JingQianCloud: Fixed and thanks for the approval.

PR #2631, opened 2023-10-25
Use clustermap to avoid state transitions from old resource

justinlin-linkedin:
When reconstructing resources, if some hosts are down, and when they wake up, they would receive state transition messages for the old resources, which is to drop partitions. At the time, if the message of old resources comes before the message of new resources, ambry would drop those partitions. This is not good.

This PR fixes that by relying clustermap to provide the resource for a partition. If the resource provided by the clustermap is higher than the resource from the message, we would ignore this transition.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
@@ -78,7 +78,7 @@ public class AmbryPartitionStateModel extends StateModel {
     StateModelParser parser = new StateModelParser();
     _currentState = parser.getInitialState(AmbryPartitionStateModel.class);
     this.partitionNameToResourceName = partitionNameToResourceName;
-    this.clusterManager = clusterManager;
+    this.clusterManager = Objects.requireNonNull(clusterManager, "Clustermap is missing");
   }
 
   @Transition(to = "BOOTSTRAP", from = "OFFLINE")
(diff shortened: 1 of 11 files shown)


JingQianCloud:
The state transition message for the old resources is to drop these partitions.
What about the new resource message, is that trying to add back the same partitions? Or the new resource is irrelevant to the old partition?

```
When reconstructing resources, if some hosts are down, and when they wake up, they would receive state transition messages for the old resources, which is to drop partitions. 
At the time, if the message of old resources comes before the message of new resources, ambry would drop those partitions. This is not good.
```

This PR fixes that by relying clustermap to provide the resource for a partition. If the resource provided by the clustermap is higher than the resource from the message, we would ignore this transition.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryPartitionStateModel.java
```diff
@@ -194,14 +196,23 @@ public void reset() {
   boolean shouldTransition(Message message) {
     String resourceName = message.getResourceName();
     String partitionName = message.getPartitionName();
+    String resourceNameToCompare = clusterManager.getResourceForPartitionInLocalDc(partitionName);
+    if (resourceNameToCompare != null && !resourceNameToCompare.equals(resourceName)) {
```
JingQianCloud, line 200: Looks like the resource name is an incremental integer. 
I guess the following code is trying to get the latest resource for this partition, right?
Do you mind adding some comment for it.

Arun-LinkedIn (approved):

JingQianCloud (approved):

PR #2632, opened 2023-10-25
[vcr-2.0] Use azure-sdk-bom to manage azure libs

snalli:
Issue: https://github.com/Azure/azure-sdk-for-java/issues/37223
Ref: https://devblogs.microsoft.com/azure-sdk/dependency-management-for-java/

MSFT recommends using azure-sdk-bom to manage azure dependencies instead of us pulling in specific versions. Azure SDK BOM automatically pulls in right versions, so we don't have to specify them.

### DIFF
--- a/gradle/dependency-versions.gradle
+++ b/gradle/dependency-versions.gradle
@@ -8,6 +8,7 @@
 // under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
 // CONDITIONS OF ANY KIND, either express or implied.
 ext {
+    azureSdkBom = "1.2.17"
     junitVersion = "4.8.1"
     slf4jVersion = "1.7.35"
     joptSimpleVersion = "4.9"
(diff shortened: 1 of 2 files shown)


gradle/dependency-versions.gradle
```diff
@@ -23,13 +24,8 @@ ext {
     helixVersion = "1.1.0"
     jacksonVersion = "2.10.2"
     jaydioVersion = "0.1"
-    azureStorageBlobVersion = "12.15.0"
-    azureStorageBlobBatchVersion = "12.12.0"
-    azureCosmosDbAsyncVersion = "4.39.0"
     azureMsal4jVersion="1.7.1"
```
JingQianCloud, line 27: Looks like we moved some azure libraries but not all, is that any reason for it?

snalli: if u r referring to `com.microsoft.azure:msal4j`, its not part of azure-sdk.

JingQianCloud (approved):

PR #2633, opened 2023-10-26
Add RepairRequests DB query pagination

JingQianCloud:
If Repairing job fails to fix some errors, with pagination and token, the job continues to fix other requests.

### DIFF
--- a/ambry-mysql/src/main/java/com/github/ambry/repair/MysqlRepairRequestsDb.java
+++ b/ambry-mysql/src/main/java/com/github/ambry/repair/MysqlRepairRequestsDb.java
@@ -18,6 +18,7 @@ import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.github.ambry.config.MysqlRepairRequestsDbConfig;
+import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.sql.Connection;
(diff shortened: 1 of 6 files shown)


ambry-mysql/src/main/java/com/github/ambry/repair/MysqlRepairRequestsDb.java
```diff
@@ -263,10 +268,22 @@ public List<RepairRequestRecord> getRepairRequestsExcludingHost(long partitionId
...
                     lifeVersion, expirationTime != null ? expirationTime.getTime() : Utils.Infinite_Time);
-            result.add(record);
+            if (result.size() < config.listMaxResults) {
+              result.add(record);
+            }
+            // update the pagination token
+            // LOCAL_CONSISTENCY_TODO
+            // one risk is that all the records has the same operation time and RepairRequest handler fails to fix any of them.
+            // then, the next page won't advance
+            nextPageToken = record.getOperationTimeMs();
```
ankagrawal, line 278: Its not required for all records to have the same operation time. If greater than "pageSize" records have the same operation time, then token will not advance.
Can we make something more unique as a token? Maybe a combination of `blobId` and `operationType`?

JingQianCloud: This is the risk I listed as the LOCAL_CONSISTENCY_TODO with the comment. The risk to hit it is
1. massive records fails at the exact the operationTime on the same partition, number is more than the page size.
2. None of them can be fixed by the repair thread.

RepairRequests is not a generic database. It's used to fix the requests and the record will be removed as long as it's fixed. If we hit this condition, that means there are massive records somehow we couldn't fix for this partition. It's already a red flag. We have a metrics for it and will set SLA0 to SLA2 alert for that.

The table is indexed by operationTime because we handle the error in order. If we want to totally solve the problem, we need a combined index of operationTime with BlobId.

justinlin-linkedin (approved):
LGTM

ankagrawal (approved):

PR #2641, opened 2023-11-03
Remove the index object from StoreMetrics map, if the store is removed or shutdown

ankagrawal:
This can avoid potential memory leak when stores are removed from a host.


justinlin-linkedin (approved):
LGTM

Arun-LinkedIn (approved):

snalli (approved):
LGTM

PR #2644, opened 2023-11-05
[vcr-2.0] Check delete-status during compaction

snalli:
Adds a fix to eliminate all snapshots of a blob during compaction and a minor improvement to avoid any unnecessary 404 errors.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
@@ -756,9 +756,6 @@ public class AzureCloudDestinationSync implements CloudDestination {
               vcrMetrics.blobCompactionRate.mark();
               break;
             case HttpStatus.SC_NOT_FOUND:
-              logger.trace("[COMPACT] Tried to erase blob {} not found in Azure blob storage, reason = {}, status {}", blobItem.getName(),
-                  eraseReason, response.getStatusCode());
-              break;
             default:
               // Just increment a counter and set an alert on it. No need to throw an error and fail the thread.
               vcrMetrics.compactionFailureCount.inc();
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
@@ -30,6 +30,7 @@ import com.azure.storage.blob.models.BlobProperties;
 import com.azure.storage.blob.models.BlobRequestConditions;
 import com.azure.storage.blob.models.BlobStorageException;
 import com.azure.storage.blob.models.BlockBlobItem;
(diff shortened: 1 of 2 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -739,14 +742,30 @@ protected int eraseBlobs(List<BlobItem> blobItemList, BlobContainerClient blobCo
...
           storageTimer.stop();
-          vcrMetrics.blobCompactionRate.mark();
+          switch (response.getStatusCode()) {
+            case HttpStatus.SC_ACCEPTED:
+              logger.trace("[COMPACT] Erased blob {} from Azure blob storage, reason = {}, status = {}", blobItem.getName(),
+                  eraseReason, response.getStatusCode());
+              numBlobsPurged += 1;
+              vcrMetrics.blobCompactionRate.mark();
+              break;
+            case HttpStatus.SC_NOT_FOUND:
```
JingQianCloud, line 758: What's the case that compaction tries to delete the blob but it's gone?
I assume there is only one compaction thread for this partition. If so, it's an error and need log as error and add metrics for it, right?

Arun-LinkedIn (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: I imagine two extremely rare cases:
1. Helix reassigns a partition to second VCR node while compaction is running on the first node. The second node may start compaction concurrently. This is not harmful, just wastes resources. Eventually the first node will complete its loop and realize the partition has been reassigned.
2. Azure falsely returns 404 briefly. Not harmful again. We skip compacting the blob for now but will pick it up in the next cycle, if it reappears.

These are highly unlikely, though not serious issues. 

Do you want a metric emitted here for this case ?

JingQianCloud: To me, it's an error not normal case. We should log as error. Also add metrics.
For the first case, ideally Helix should un-assign the first VCR and reassign the second VCR. It can happen if somehow the first vcr didn't get the request.
For the second case, I don't think it would happen. It's data integrity issue which is critical.

I'm wondering if you hit this case in your test.

snalli: No I didn't hit it during tests but I recall many blob-not-found errors in prod. 
I am not sure if that is from compaction though. 
I have changed the switch case to fall through and emit a metric.

snalli: I have modified compaction v2 to stop immediately if a partition is removed by helix in https://github.com/linkedin/ambry/pull/2643

JingQianCloud (approved):

PR #2646, opened 2023-11-10
Added partialLogSegment and CompactionInterval metrics

JingQianCloud:
1. Added two metrics, 
    a. the partial written log segment count per host
    b. the interval between two compactions on the same partition, it's per host metrics as well.
2. Refine debugging message of compaction. 
3. Separate the configuration for middle range compaction and benefit weighted compaction.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -17,6 +17,7 @@ import com.codahale.metrics.Timer;
 import com.github.ambry.config.StoreConfig;
 import com.github.ambry.replication.FindToken;
 import com.github.ambry.replication.FindTokenType;
+import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Time;
 import com.github.ambry.utils.Utils;
 import java.io.File;
@@ -512,18 +513,29 @@ class PersistentIndex {
   }
 
   /**
-   * @return count of partially written log segments
+   * @return a pair: first entry is the count of partially written log segments.
+   * the second entry is the total wasted log segment space
    */
-  int getPartialLogSegmentCount() {
+  Pair<Integer, Long> getPartialLogSegmentInfo() {
     int partialLogSegmentCount = 0;
+    long wastedLogSegmentSpace = 0;
     LogSegment segment = log.getFirstSegment();
+    LogSegmentName journalName = null;
+    if (journal.getCurrentNumberOfEntries() != 0) {
+      journalName = journal.getFirstOffset().getName();
+    }
     while (segment != null) {
+      // exclude journal
+      if (journalName != null && segment.getName().equals(journalName)) {
+        break;
+      }
       if (segment.getEndOffset() < segment.getCapacityInBytes() * 0.8) {
         partialLogSegmentCount++;
       }
+      wastedLogSegmentSpace += segment.getCapacityInBytes() - segment.getEndOffset();
       segment = log.getNextSegment(segment);
     }
-    return partialLogSegmentCount;
+    return new Pair(partialLogSegmentCount, wastedLogSegmentSpace);
   }
 
   /**
--- a/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
@@ -512,6 +512,21 @@ class PersistentIndex {
   }
 
   /**
+   * @return count of partially written log segments
+   */
+  int getPartialLogSegmentCount() {
+    int partialLogSegmentCount = 0;
+    LogSegment segment = log.getFirstSegment();
+    while (segment != null) {
+      if (segment.getEndOffset() < segment.getCapacityInBytes() * 0.8) {
+        partialLogSegmentCount++;
+      }
+      segment = log.getNextSegment(segment);
+    }
+    return partialLogSegmentCount;
+  }
+
+  /**
    * Atomically adds {@code segmentFilesToAdd} to and removes {@code segmentsToRemove} from the map of {@link Offset} to
    * {@link IndexSegment} instances.
    * @param segmentFilesToAdd the backing files of the {@link IndexSegment} instances to add.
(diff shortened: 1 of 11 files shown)


ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -511,6 +511,21 @@ List<LogSegment> getLogSegments() {
...
+  /**
+   * @return count of partially written log segments
+   */
+  int getPartialLogSegmentCount() {
+    int partialLogSegmentCount = 0;
+    LogSegment segment = log.getFirstSegment();
+    while (segment != null) {
+      if (segment.getEndOffset() < segment.getCapacityInBytes() * 0.8) {
+        partialLogSegmentCount++;
+      }
```
justinlin-linkedin, line 523: Can we filter out the active log segment? since this metric is part of compaction. Compaction doesn't care about the last active log segment?
And can we also have a metric to show how many bytes we wasted because of partial log segments?

JingQianCloud: Filtered the journal.
Added the wastedLogSegmentSpace metric.

ambry-store/src/main/java/com/github/ambry/store/StoreMetrics.java
```diff
@@ -101,8 +101,11 @@ public class StoreMetrics {
 
   // Compaction related metrics
   public final Counter compactionFixStateCount;
-  public final Meter compactionCopyRateInBytes;
-  public final Counter compactionBytesReclaimedCount;
+  public final Meter compactionCopyRateInBytes;       // per host, physical data are copied during compaction
+  public final Counter compactionBytesReclaimedCount; // per host, the real reclaimed log segment size
+  public final Meter compactionIntervalInMin;           // the interval between two compactions on the same partition.
+  public final Meter compactionPartialLogSegmentCount;  // per host, partial written log segment count
+  public final Meter compactionWastedLogSegmentSpace;   // per host, wasted log segment space
```
justinlin-linkedin, line 108: can we use Gauge for these two metrics, Gauge will show us the value, meter is basically a counter + histogram, which doesn't show us the direct value.

JingQianCloud: Changed back to Gauge. Thanks for the suggestion.

justinlin-linkedin (approved):
LGTM

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -511,6 +516,32 @@ List<LogSegment> getLogSegments() {
...
+    }
+    while (segment != null) {
+      // exclude journal
+      if (journalName != null && segment.getName().equals(journalName)) {
+        break;
+      }
+      if (segment.getEndOffset() < segment.getCapacityInBytes() * 0.8) {
+        partialCount++;
+      }
+      partialCount += segment.getCapacityInBytes() - segment.getEndOffset();
```
ankagrawal, line 538: I think this should be 
`wastedSpace += segment.getCapacityInBytes() - segment.getEndOffset();`

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -131,6 +133,9 @@ public int compare(IndexEntry e1, IndexEntry e2) {
   // When undelete is enabled, DELETE is not the final state of the blob, since a DELETEd blob can be UNDELTEd.
   private final boolean isDeleteFinalStateOfBlob;
 
+  private final AtomicInteger partialLogSegmentCount = new AtomicInteger(0);
```
ankagrawal, line 136: nit: Does this need to be AtomicInteger? I guess `BlobStoreCompactor` call `updatePartialLogSegmentInfo` only one partition at a time.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
```diff
@@ -239,7 +238,7 @@ private CostBenefitInfo getCostBenefitInfo(LogSegmentName firstLogSegmentName, L
 
     if (weightOnBenefit) {
       if (totalCost < adjustedTotalCost) {
-        logger.info("getCostBenefitInfo {} cost {} adjusted cost {}",
+        logger.trace("getCostBenefitInfo {} cost {} adjusted cost {}",
```
ankagrawal, line 241: nit: why change it to trace? Does it log too much?

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
JingQianCloud: It's about the thread which reads the variable. The ingraph metric collector reads it while the compactor is updating it.
We can argue that JVM make at least 32bit primary type atomic. But it's good practice to use atomic since different threads are reading and updating it.

ambry-store/src/main/java/com/github/ambry/store/StatsBasedCompactionPolicy.java
JingQianCloud: yes

ambry-store/src/main/java/com/github/ambry/store/PersistentIndex.java
```diff
@@ -511,6 +516,32 @@ List<LogSegment> getLogSegments() {
...
+    LogSegmentName journalName = null;
+    if (journal.getCurrentNumberOfEntries() != 0) {
+      journalName = journal.getFirstOffset().getName();
+    }
+    while (segment != null) {
+      // exclude journal
+      if (journalName != null && segment.getName().equals(journalName)) {
+        break;
+      }
+      if (segment.getEndOffset() < segment.getCapacityInBytes() * 0.8) {
```
ankagrawal, line 535: My understanding so far is that  `partialCount` is supposed to give us some idea around which log segments were good candidates for compaction but maybe were not compacted by the current scheme? If yes, then maybe `0.8` might not be a good threshold. Its totally fine if something that is 0.79 full and is not picked up for compaction. We should worry about something that is 0.3 full and is not picked up. So probably we should lower this threshold to something like `0.5` or `0.6`

JingQianCloud: No, it's not for picking compaction candidates. We use it to monitor the problem caused by stats based compaction which leave a lot of unused space. 
This is not the valid data vs invalid data threshold. It's the end of write to the log segment ratio.
For that segment, it's possible that all data are valid but we didn't write the full log segment and wasted space.

ankagrawal (approved):

PR #2647, opened 2023-11-15
Support deleted_ts and delete_ts for auto_purge. Based on PR 2594

JingQianCloud:
Based on https://github.com/linkedin/ambry/pull/2594/files
Fixed two issues:
1. fixed getLatestTimeStamp bug which takes the null timestamp over tthe non-null timestamp 
2. Fixed the updateDatasetVersionTtl only updated delete_ts issue. It's due to the mysql syntax issue.

### DIFF
--- a/ambry-utils/src/main/java/com/github/ambry/utils/Utils.java
+++ b/ambry-utils/src/main/java/com/github/ambry/utils/Utils.java
@@ -1503,7 +1503,10 @@ public class Utils {
    * Get the latest timestamp, and if one of the timestamp is null, return the non-null timestamp.
    */
   public static Timestamp getLatestTimeStamp(Timestamp timestamp1, Timestamp timestamp2) {
-    if (compareTimestamp(timestamp1, timestampToMs(timestamp2)) > 0) {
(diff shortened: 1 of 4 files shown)


justinlin-linkedin (approved):

ambry-utils/src/main/java/com/github/ambry/utils/Utils.java
```diff
@@ -1499,6 +1499,17 @@ public static int compareTimestamp(Timestamp timestamp, long otherTimeMs) {
     return Utils.compareTimes(timestampToMs(timestamp), otherTimeMs);
   }
 
+  /**
+   * Get the latest timestamp, and if one of the timestamp is null, return the non-null timestamp.
+   */
+  public static Timestamp getLatestTimeStamp(Timestamp timestamp1, Timestamp timestamp2) {
```
JingQianCloud, line 1505: when timestamp1 is not null but timestamp2 is null, this function will return the null one.
It's because compareTimestamp thinks -1 is infinity and it's bigger. 

Need a fix for it.

justinlin-linkedin (approved):

PR #2649, opened 2023-11-15
Replicate from all replicas if the partition doesn't have a leader

Arun-LinkedIn:
When there is no leader for a partition, replication is very slow since we wait for 2 mins in each replication cycle before issuing GET requests. To avoid this, replicate from all remote replicas when we detect that a partition doesn't have a leader.

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -57,6 +57,7 @@ import com.github.ambry.store.StoreKey;
 import com.github.ambry.store.StoreKeyConverter;
 import com.github.ambry.store.Transformer;
 import com.github.ambry.utils.NettyByteBufDataInputStream;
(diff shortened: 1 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -366,45 +367,21 @@ public void replicate() {
...
-          if (standbyReplicasTimedOutOnNoProgress.size() > 0) {
-            RemoteReplicaGroup group =
-                new RemoteReplicaGroup(standbyReplicasTimedOutOnNoProgress, remoteNode, true, remoteReplicaGroupId++);
-            remoteReplicaGroups.add(group);
-          }
-        }
-      }
+
+      // Get remote replicas grouped by node.
+      remoteReplicaGroups = getRemoteReplicaGroups();
```
Arun-LinkedIn, line 377: Just abstracting the code into a method here so that it is easy to read.

justinlin-linkedin (approved):

PR #2650, opened 2023-11-17
Resolving Empty Concurrency Map when Disk Token Persisting

MikeDafi:
## Summary

This PR fixes an issue where the disk token persister would throw an exception if there were no partitions on a mount path, instead of gracefully handling the empty partition list. 

## Motivation

Currently, if there are no partitions on a mount path that experienced an IO exception, the disk token persister tries to iterate through the partition list for that mount path to check the disk status. This results in a NullPointerException.

We should handle the case where the partition list is empty and skip the persistor for that path, rather than throwing an exception.

## Changes

- Check if the partition list for a mount path is empty 
- If empty, skip the mount path instead of throwing exception
- If not empty, proceed with checking disk status as before

This allows the persistor to gracefully handle cases where IO exceptions occurred but there are no partitions on that path.

Fixes [EXC-#374169](https://jira01.corp.linkedin.com:8443/browse/EXC-374169)

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/DiskTokenPersistor.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/DiskTokenPersistor.java
@@ -21,6 +21,7 @@ import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


justinlin-linkedin (approved):
LGTM

PR #2653, opened 2023-11-21
Allow prioritization of remote replicas for GET if there are very few local replicas.

ankagrawal:
For get operations, if there are very few local replicas, then shuffle the pool with remote replicas so that the few remaining local replicas don't get overloaded with all the traffic.
Define a config with the threshold for minimum number of local replicas to do this.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -299,6 +299,7 @@ class SimpleOperationTracker implements OperationTracker {
     List<ReplicaId> examinedReplicas = new ArrayList<>();
     originatingDcName = originatingDcName == null ? reassignedOriginDc : originatingDcName;
     int numLocalAndLiveReplicas = 0;
+    int numRemoteOriginatingDcAndLiveReplicas = 0;
     for (ReplicaId replicaId : replicas) {
       examinedReplicas.add(replicaId);
       String replicaDcName = replicaId.getDataNodeId().getDatacenterName();
@@ -310,6 +311,7 @@ class SimpleOperationTracker implements OperationTracker {
           numLocalAndLiveReplicas++;
           addToBeginningOfPool(replicaId);
         } else if (crossColoEnabled && isOriginatingDcReplica) {
+          numRemoteOriginatingDcAndLiveReplicas++;
           addToEndOfPool(replicaId);
         } else if (crossColoEnabled) {
           backupReplicas.addFirst(replicaId);
@@ -349,6 +351,7 @@ class SimpleOperationTracker implements OperationTracker {
     }
 
     maybeDeprioritizeLocalBootstrapReplicas(numLocalAndLiveReplicas);
+    maybeShuffleWithRemoteReplicas(numLocalAndLiveReplicas, numRemoteOriginatingDcAndLiveReplicas);
     totalReplicaCount = replicaPool.size();
 
     // MockPartitionId.getReplicaIds() is returning a shared reference which may cause race condition.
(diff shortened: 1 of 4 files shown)


justinlin-linkedin (commented):
please add tests to the new code.

ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
```diff
@@ -614,6 +617,32 @@ void maybeDeprioritizeLocalBootstrapReplicas(int numLocalAndLiveReplicas) {
...
 
+  /**
+   * For get operations, if there are very few local replicas, then shuffle the pool with remote replicas so that the
+   * few (or one) remaining local replicas don't get all the traffic. The threshold for minimum number of local replicas
+   * is defined in {@link RouterConfig#routerGetOperationMinLocalReplicaCountToPrioritizeLocal}.
+   * @param numLocalAndLiveReplicas the number of local and live replicas.
+   * @param numRemoteOriginatingDcAndLiveReplicas the number of remote originating DC and live replicas.
+   */
+  void maybeShuffleWithRemoteReplicas(int numLocalAndLiveReplicas, int numRemoteOriginatingDcAndLiveReplicas) {
+    if (isGetOperation() && numLocalAndLiveReplicas < routerConfig.routerGetOperationMinLocalReplicaCountToPrioritizeLocal) {
```
justinlin-linkedin, line 628: let's do `numLocalAndLiveReplicas <= routerConfig.xxxx` here.

ankagrawal: done.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -897,6 +901,39 @@ public void testOrigDcUnavailability() throws Exception {
...
+   * replicas available, then we will prioritize remote DC.
+   * @throws Exception
+   */
+  @Test
+  public void testGetGoesToRemoteIfLocalDcHasFewAvailableReplicas() throws Exception {
+    doPut();
+    Properties props = getDefaultNonBlockingRouterProperties(true);
+    props.setProperty(RouterConfig.ROUTER_GET_ELIGIBLE_REPLICAS_BY_STATE_ENABLED, "true");
+    props.setProperty(RouterConfig.ROUTER_GET_BLOB_RETRY_LIMIT_COUNT, "10");
+    props.setProperty(RouterConfig.ROUTER_GET_BLOB_RETRY_LIMIT_IN_SEC, "10");
```
justinlin-linkedin, line 915: nit: this confusing, let's not change any retry related configuration.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
```diff
@@ -897,6 +901,39 @@ public void testOrigDcUnavailability() throws Exception {
...
+      MockPartitionId partitionId = (MockPartitionId) blobId.getPartition();
+      List<ReplicaId> localReplicas = partitionId.replicaIds.stream()
+          .filter(replicaId -> replicaId.getDataNodeId().getDatacenterName().equals(localDcName))
+          .collect(Collectors.toList());
+      partitionId.setReplicaState(localReplicas.get(0), ReplicaState.LEADER);
+      partitionId.setReplicaState(localReplicas.get(1), ReplicaState.OFFLINE);
+      partitionId.setReplicaState(localReplicas.get(2), ReplicaState.OFFLINE);
+
+      getAndAssertSuccess();
+      assertTrue(routerMetrics.shuffledWithRemoteReplicasForGetDueToFewLocalReplicas.getCount() > 0);
```
justinlin-linkedin, line 931: A better assertion would be get the value before calling `getAndAssertSuccess` and the value after `getAndAssertSuccess` and make sure the value after = value before + 1.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
ankagrawal: fixed.

ambry-router/src/test/java/com/github/ambry/router/GetBlobOperationTest.java
ankagrawal: Its actually a little more complicated. So the number of times the counter `shuffledWithRemoteReplicasForGetDueToFewLocalReplicas` is incremented depends on number of chunks in a blob and if that chunk was placed in the partition that has doesn't have enough replicas in the local colo.

In order to make the test more deterministic, and also to not introduce unnecessary complication, I have changed the test to use simple blobs only.

justinlin-linkedin (approved):

Arun-LinkedIn (approved):
Just a QQ. We are assuming that if local replicas are less in number, they would be overloaded. As an later optimization, is there any way to know that they are actually overloaded?

PR #2654, opened 2023-11-21
[vcr-2.0] Create ambry-vcr package

snalli:
This patch isolates VCR code into its own package ambry-vcr. Prior to this, the VCR code was spread across multiple packages but mostly present in ambry-cloud. This creates circular dependencies for future projects. For instance, the VCR code depends on AmbryRequests from the server-package. This creates a circular dependency if the server code needs to use cloud code.

Secondly, this patch moves out all VCR integration from server integration tests. This reduces the time for server-int tests to finish.

Lastly. this patch Ignores all the old and outdated VCR and Azure tests. Azure has changed significantly since these tests were written and don't really add any value.


justinlin-linkedin (approved):

JingQianCloud (approved):
Thanks for the re-organizing the code structure

PR #2655, opened 2023-11-27
[Efficient Metadata Operations] For reserved metadata chunks, the Operation Tracker should use reserved metadata chunk's partition id instead of attempting to get a random one.

ankagrawal:
This PR fixes a bug in PutOperation where, while sending Put requests to ambry server, the code was trying to pick a random partition id even for the reserved metadata chunks. For reserved metadata chunk, it should pick the partition id already present in the reserved metadata chunk id.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
@@ -1054,6 +1054,13 @@ class PutOperation {
   }
 
   /**
+   * @return MetadataPutChunk associated with this operation.
+   */
+  MetadataPutChunk getMetadataPutChunk() {
+    return metadataPutChunk;
+  }
+
+  /**
    * Gets the precedence level for a {@link RouterErrorCode}. A precedence level is a relative priority assigned
    * to a {@link RouterErrorCode}. If a {@link RouterErrorCode} has not been assigned a precedence level, a
    * {@code Integer.MIN_VALUE} will be returned.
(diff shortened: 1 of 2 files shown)


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1277,19 +1291,21 @@ private void prepareForSending() {
       try {
         BlobDataType blobDataType =
             isMetadataChunk() ? BlobDataType.METADATA : isSimpleBlob ? BlobDataType.SIMPLE : BlobDataType.DATACHUNK;
-        partitionId = getPartitionForPut(partitionClass, attemptedPartitionIds);
-        // To ensure previously attempted partitions are not retried for this PUT after a failure.
-        attemptedPartitionIds.add(partitionId);
 
         if (shouldUseReservedMetadataId()) {
+          partitionId = metadataPutChunk.reservedMetadataChunkId.getPartition();
```
JingQianCloud, line 1296: Ankur, the PR looks good to me.

Just wondering,
Before the fix, the chunkBlobId and partitionId are from two different partitions? What's the consequence or behavior after it?

ankagrawal: The consequence was that Puts were failing.
OperationTracker took the blobid of the reserved metadata, and in some cases a different partition id. And so the put request landed on hosts that didnt have the partition in the blobid. These put requests just failed.

JingQianCloud (approved):

PR #2659, opened 2023-11-28
Don't use unavailable disks for bootstrapping new replicas

Arun-LinkedIn:
Set disk as unavailable if DiskManager fails to start so that we don't use the disk for bootstrapping new replicas

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
@@ -16,6 +16,7 @@ package com.github.ambry.store;
 
 import com.github.ambry.account.AccountService;
 import com.github.ambry.clustermap.DiskId;
+import com.github.ambry.clustermap.HardwareState;
 import com.github.ambry.clustermap.PartitionId;
 import com.github.ambry.clustermap.ReplicaId;
 import com.github.ambry.clustermap.ReplicaStatusDelegate;
@@ -215,6 +216,12 @@ public class DiskManager {
     } catch (StoreException e) {
       logger.error("Error while starting the DiskManager for {} ; no stores will be accessible on this disk.",
           disk.getMountPath(), e);
+      // Set the state of the disk to UNAVAILABLE. This will prevent new replicas to be bootstrapped on this disk.
+      // TODO: We might need to use DiskFailureHandler to immediately reset existing replicas on this disk and reduce
+      //  instance capacity so that existing replicas are reassigned to new hosts and Helix doesn't assign more replicas
+      //  than the host can handle.
+      logger.error("Setting disk {} as UNAVAILABLE locally", disk.getMountPath());
+      disk.setState(HardwareState.UNAVAILABLE);
     } finally {
       if (!running) {
         metrics.totalStoreStartFailures.inc(stores.size());


ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -215,6 +216,12 @@ void start() throws InterruptedException {
     } catch (StoreException e) {
       logger.error("Error while starting the DiskManager for {} ; no stores will be accessible on this disk.",
           disk.getMountPath(), e);
+      // Set the state of the disk to UNAVAILABLE. This will prevent new replicas to be bootstrapped on this disk.
+      // TODO: We might need to use DiskFailureHandler to immediately reset existing replicas on this disk and reduce
+      //  instance capacity so that existing replicas are reassigned to new hosts and Helix doesn't assign more replicas
+      //  than the host can handle.
+      logger.info("Setting disk {} as UNAVAILABLE locally", disk.getMountPath());
+      disk.setState(HardwareState.UNAVAILABLE);
```
justinlin-linkedin, line 224: this is probably not going to work since it will be overridden by the HelixClusterManager once there is a DataNodeConfig update.

justinlin-linkedin (changes requested):

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -578,29 +578,43 @@ private class PartitionStateChangeListenerImpl implements PartitionStateChangeLi
...
+          if (!addBlobStore(replicaToAdd)) {
+            // We have decreased the available disk space in HelixClusterManager#getDiskForBootstrapReplica. Increase it
+            // back since addition of store failed.
+            replicaToAdd.getDiskId().increaseAvailableSpaceInBytes(replicaToAdd.getCapacityInBytes());
+            if (!clusterMap.isDataNodeInFullAutoMode(currentNode)) {
+              logger.error("Failed to add store {} into storage manager", partitionName);
+              throw new StateTransitionException("Failed to add store " + partitionName + " into storage manager",
+                  ReplicaOperationFailure);
+            } else {
+              // TODO: Delete any files added in store and reserve directory
```
justinlin-linkedin, line 609: we should do this, otherwise, it would just take some disk space and won't let it go before we restart. I suppose we can just add a new method in disk manager to clean up the "unexpected dirs". I have a PR to do that, let's merge these two PRs and then use the method in my PR to remove the partition directories later.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -16,6 +16,7 @@
 
 import com.github.ambry.account.AccountService;
 import com.github.ambry.clustermap.DiskId;
+import com.github.ambry.clustermap.HardwareState;
```
justinlin-linkedin, line 19: this import is not used anywhere.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManager.java
```diff
@@ -113,6 +113,7 @@ public class HelixClusterManager implements ClusterMap {
   private final PartitionSelectionHelper partitionSelectionHelper;
   private final Map<String, Map<String, String>> partitionOverrideInfoMap = new HashMap<>();
   private final Map<String, ReplicaId> bootstrapReplicas = new ConcurrentHashMap<>();
+  private final Map<String, Set<DiskId>> disksAttemptedForBootstrap = new ConcurrentHashMap<>();
```
justinlin-linkedin, line 116: one slightly different approach here is avoid using a disk whenver it fails a bootstrap replica. But this also would do the work

justinlin-linkedin (approved):

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
Arun-LinkedIn: Sure. I will merge this PR and put up a new PR

PR #2661, opened 2023-11-29
Include leaderless replicas for GET replication requests

Arun-LinkedIn:
During leader-based replication, we only include leaders for fetching actual data. For non-leaders, we wait for 2 minutes hoping that the data would be fetched via intra-colo replication. This causes issue of very slow replication if leader is missing. To avoid that, we are fetching data for replicas without any leaders in this PR.


justinlin-linkedin (approved):
Looks much better now

PR #2664, opened 2023-12-01
POC for S3 API experiments (Not to be checked in)

Arun-LinkedIn:


JingQianCloud:
I'm thinking of adding a enum of AmazonS3_Action
https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_Amazon_Simple_Storage_Service.html

For example, followings are related to multiple part upload.

- CreateMultipartUpload
- UploadPart
- CompleteMultipartUpload
- AbortMultipartUpload
- ListParts
- ListMultipartUploads

At the early stage like somewhere at 
We generate the action type. In our handling code, we don't parse or rely on keywords like "upload", "uploadId" etc.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -541,7 +556,7 @@ public static BlobProperties buildBlobProperties(Map<String, Object> args) throw
    * @throws RestServiceException if required arguments aren't present or if they aren't in the format expected.
    */
   public static long getTtlFromRequestHeader(Map<String, Object> args) throws RestServiceException {
-    long ttl = Utils.Infinite_Time;
+    long ttl = 86400;
```
JingQianCloud, line 559: Did we ever discuss the life cycle management with flink and TiKV team?

JingQianCloud:
We ever discussed for S3, do we need dedicated handler for it.
I'm thinking inherit namedBlob handler and create s3 handler. 
In S3 handler, we only overwrite the code which we need change.

PR #2665, opened 2023-12-02
[vcr-2.0] Add ctor to secret-storage client

snalli:


snalli:
Merging because Arun approved it on slack

PR #2667, opened 2023-12-06
[Efficient Metadata Operations] Return reserved metadata id of a blob in the response header.

ankagrawal:


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbrySecurityService.java
```diff
@@ -256,6 +256,10 @@ public void processResponse(RestRequest restRequest, RestResponseChannel respons
                   setAccountAndContainerHeaders(restRequest, responseChannel);
                   setUserMetadataHeaders(blobInfo.getUserMetadata(), responseChannel, container.getUserMetadataKeysToNotPrefixInResponse());
                   responseChannel.setHeader(RestUtils.Headers.LIFE_VERSION, blobInfo.getLifeVersion());
+                  if (blobInfo.getBlobProperties().getReservedMetadataBlobId() != null) {
```
JingQianCloud, line 259: For the close end source, we release reserved chunk id for internal request only, can we have the similar flag in OSS?

JingQianCloud (approved):

PR #2668, opened 2023-12-08
Fix some logic in SimpleOperationTracker

justinlin-linkedin:
## Summary

This PR changes some logic in SimpleOperationTracker as well as removing some unused RouterConfig and unit test

1. Return NOT_FOUND if all replicas return NOT_FOUND even if we have bootstrap replicas.
2. Add a new configuration to only require two NOT_FOUND in originating DC to return NOT_FOUND. 
3. Remove some unused configurations from RouterConfig.

## Detail

We used to treat bootstrap replicas as unavailable replicas when it returns NOT_FOUND. But when all the other replicas return NOT_FOUND, bootstrap replicas will eventually return NOT_FOUND for sure, since it gets blob content from other replicas. This PR would count the number of NOT_FOUND response and when it's the same as the total number of replicas, we just return NOT_FOUND even if there are bootstrap replicas in originating DC.

This PR also adds a new configuration to only require two NOT_FOUND response in originating DC to return NOT_FOUND. In FULL AUTO, a host can become unavailable and some time later, helix would start a new bootstrap replica somewhere else. In this case, we would have 4 replicas in originating DC, 1 leader, 1 standby, 1 bootstrap and 1 unavailable. When blob id doesn't exist, leader and standby and bootstrap would all return NOT_FOUND, but we would treat bootstrap as unavailable. In this case, we don't return NOT_FOUND since we only tolerate one unavailable replicas. This PR would change that. The reason why this is reasonable is that PutOperation would always make sure it succeeds on N-1 replicas. So if we have 2 leader/standby replicas returning NOT_FOUND, we can return NOT_FOUND.

## Test
Unit test

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -120,7 +120,6 @@ public class RouterConfig {
       "router.cross.colo.request.to.dc.with.most.replicas";
   public static final String ROUTER_BACKGROUND_DELETER_MAX_CONCURRENT_OPERATIONS =
       "router.background.deleter.max.concurrent.operations";
-  public static final String ROUTER_PUT_REQUEST_USE_JAVA_NATIVE_CRC32 = "router.put.request.use.java.native.crc32";
   public static final String OPERATION_CONTROLLER = "router.operation.controller";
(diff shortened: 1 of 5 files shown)


JingQianCloud (approved):
LGTM.
Thanks for the fix as well as the code refine/cleanup.

PR #2670, opened 2023-12-11
[vcr-2.0] Remove unused recentBlobCache

snalli:
Final planned fix for backup system 2.0.

We do not use the `recentBlobCache` for VCR 2.0 and setting the `maxCacheLimit` to 0 doesn't achieve that. The code still creates the `recentBlobCache` with a hardcoded initial size of 1000 entries (`cacheInitialCapacity`). This patch just completely removes the `recentBlobCache`. It is an incomplete and untested legacy remnant that gives erroneous results.

There are couple of other improvements. For a PUT with no encryption, it is better to just pass the blob's un-encrypted inputStream to azure-sdk instead of creating an in-memory copy and passing that to azure-sdk. The SDK will handle retries.

Lastly, this patch replaces the deprecated `download` with recommended `downloadStream`

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -23,6 +23,7 @@ import com.github.ambry.config.CloudConfig;
 import com.github.ambry.config.ClusterMapConfig;
 import com.github.ambry.config.StoreConfig;
 import com.github.ambry.config.VerifiableProperties;
+import com.github.ambry.messageformat.MessageFormatWriteSet;
 import com.github.ambry.replication.FindToken;
 import com.github.ambry.store.FindInfo;
 import com.github.ambry.store.MessageInfo;
@@ -93,6 +94,7 @@ public class CloudBlobStore implements Store {
   private final boolean isVcr;
   private boolean started;
   private volatile ReplicaState currentState = ReplicaState.OFFLINE;
+  private CloudConfig cloudConfig;
 
   /**
    * Constructor for CloudBlobStore
@@ -105,7 +107,7 @@ public class CloudBlobStore implements Store {
    */
   CloudBlobStore(VerifiableProperties properties, PartitionId partitionId, CloudDestination cloudDestination,
       ClusterMap clusterMap, VcrMetrics vcrMetrics) throws IllegalStateException {
-    CloudConfig cloudConfig = new CloudConfig(properties);
+    this.cloudConfig = new CloudConfig(properties);
     ClusterMapConfig clusterMapConfig = new ClusterMapConfig(properties);
     this.clusterMap = clusterMap;
     this.storeConfig = new StoreConfig(properties);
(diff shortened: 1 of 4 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -116,13 +118,13 @@ public class CloudBlobStore implements Store {
     minTtlMillis = TimeUnit.DAYS.toMillis(cloudConfig.vcrMinTtlDays);
     requireEncryption = cloudConfig.vcrRequireEncryption;
     isVcr = cloudConfig.cloudIsVcr;
-    if (isVcr) {
+    if (isVcr && cloudConfig.ambryBackupVersion.equals(CloudConfig.AMBRY_BACKUP_VERSION_1)) {
```
JingQianCloud, line 121: There is CloudBlobStoreV2. Are you still using it?
If yes, should the v2 code be there?
Looks like there are many implementation difference between V1 and V2, removed retry, removed cache, removed WriteChannel.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -371,13 +373,23 @@ static boolean isBlobExpired(CloudBlobMetadata metadata, long currentTimeStamp)
...
   @Override
   public void put(MessageWriteSet messageSetToWrite) throws StoreException {
-    // TODO: Remove the duplicate code by calling putAsync() method and wait on it once we verify that upload works correctly.
     checkStarted();
     if (messageSetToWrite.getMessageSetInfo().isEmpty()) {
       throw new IllegalArgumentException("Message write set cannot be empty");
     }
     checkDuplicates(messageSetToWrite.getMessageSetInfo());
 
+    if (cloudConfig.ambryBackupVersion.equals(CloudConfig.AMBRY_BACKUP_VERSION_2) && !cloudConfig.vcrRequireEncryption) {
```
JingQianCloud, line 382: Do we see performance or memory difference by bypassing the CloudWriteChannel?

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
snalli: I will delete all V2 classes next month. I designed them to avoid these cases and implement separately from V1 but that idea was not accepted early this semester. Right now, I don't have the time to refactor and go back to it. 

I tried to set the max cache size to 0 in configs, but the V1 code is so bad that it silently creates a cache of size 1000.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
snalli: I will have some numbers by the evening.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
snalli:
```
  private static final int cacheInitialCapacity = 1000;

```

Arun-LinkedIn (approved):

JingQianCloud (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
snalli: No change in perf, but slight decrease in overall memory usage. Around 4GB.

JingQianCloud: In this case, we don't need this bypass code.

PR #2671, opened 2023-12-12
Add file path in the crc error log for given store key

justinlin-linkedin:
## Summary
Adding the file path in the crc error log so we can easily identify the disk. CRC error can be caused by failed disk. This log message make identity the disk easier.


JingQianCloud (approved):
LGTM

PR #2675, opened 2023-12-14
[vcr-2.0] Log corrupt BlobIDs to Azure Table 

snalli:
This patch adds support to record corrupted blobIDs in Azure Table. It extends ReplicaThread to create VcrReplicaThread which overrides logToExternalTable, such that replication between servers is unaffected.


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ConnectionStringBasedStorageClient.java
```diff
@@ -90,6 +95,16 @@ protected BlobServiceClient buildBlobServiceSyncClient(HttpClient httpClient, Co
         .buildClient();
   }
 
+  @Override
+  protected TableServiceClient buildTableServiceClient(HttpClient httpClient, Configuration configuration,
+      RetryOptions retryOptions, AzureCloudConfig azureCloudConfig) {
+    return new TableServiceClientBuilder().connectionString(azureCloudConfig.azureTableConnectionString)
```
JingQianCloud, line 101: By default, azureTableConnectionString is null. 
Looks like getTableServiceClient will throw exception to the constructor.
Do you test the case when  azureTableConnectionString is null?

JingQianCloud (commented):
overall looks good to me. one comment left.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/ConnectionStringBasedStorageClient.java
snalli: Added a method `validateTableServiceConfigs` which will fail VCR startup if the table-conn-str is null and we are using a conn-str client.  The conn-str varies depending on the azure account, so we cannot hardcode it here.

JingQianCloud (approved):

PR #2687, opened 2024-01-09
[Recovery] Impl get-metadata from azure-storage

snalli:
This patch implements the basic functionality required to initiate recovery of data from Azure. The recovery code will run directly on the server, which can contact azure and fetch blobs. The first step is to fetch metadata. Azure Storage provides a continuation token which the server can use to fetch paginated metadata. When the server runs in recovery mode, it does not replicate from its peer servers. It only replicates data from Azure. Recovery for a partition is complete when the token issued by Azure is null. This patch uses a json format to store the token along with some additional info for convenience.

The major files are RecoveryNetworkClient and RecoveryNetworkClientTest


ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryNetworkClient.java
```diff
@@ -124,134 +149,125 @@ public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Inte
...
+      }
 
-          break;
+      // Extract ambry metadata
+      logger.trace("For container {}, number of blobItems from Azure = {}", containerName, response.getValue().size());
+      List<MessageInfo> messageInfoList = new ArrayList<>();
+      long bytesRead = 0, blobsRead = 0;
+      for (BlobItem blobItem: response.getValue()) {
+        MessageInfo messageInfo = getMessageInfo(blobItem);
+        if (messageInfo != null) {
```
JingQianCloud, line 206: 1. What's the else case? If messageInfo is null, what does that mean? 
2. Do you need error log for it?

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryNetworkClient.java
```diff
@@ -124,134 +149,125 @@ public List<ResponseInfo> sendAndPoll(List<RequestInfo> requestsToSend, Set<Inte
...
+       * There is no way to know which case happened, so just put ttlUpdated as false.
+       * expiryTimeMs is included however and replication-code can decide necessary course of action.
+       * Same goes for undeleted field and lifeVersion is included for replication-code to decide.
+       *
+       * We don't upload CRC during backups. This is probably an issue.
+       */
+      return new MessageInfo(new BlobId(blobItem.getName(), clustermap),
+          Long.parseLong(metadata.get(CloudBlobMetadata.FIELD_SIZE)),
+          metadata.containsKey(CloudBlobMetadata.FIELD_DELETION_TIME),
+          false,
```
JingQianCloud, line 253: Do you mind adding the comment for the field we fill in false/null? It'll be easier for reader to know which field it is.

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryNetworkClient.java
```diff
@@ -67,17 +70,38 @@
...
+  public RecoveryNetworkClient(VerifiableProperties properties, MetricRegistry registry, ClusterMap clusterMap,
+      StoreManager storeManager, AccountService accountService, RecoveryNetworkClientCallback clientCallback) {
+    this.clientCallback = clientCallback == null ? new EmptyCallback() : clientCallback;
+    this.clustermap = clusterMap;
     this.storeManager = storeManager;
-    this.cosmosContainer = cosmosContainer;
+    this.clusterMapConfig = new ClusterMapConfig(properties);
+    this.azureCloudConfig = new AzureCloudConfig(properties);
+    this.recoveryMetrics = new RecoveryMetrics(registry);
+    this.messageInfoCache = new ConcurrentHashMap<>();
```
JingQianCloud, line 94: Looks like we don't use the cache for this version. 
Do we use cache before for the restore? How did we use it?

snalli:
MSFT issue https://github.com/Azure/Azurite/issues/2349

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryNetworkClient.java
snalli: null messageInfo means there was an error extracting ambry-metadata from azure-metadata for a blob. The error is logged in getMessageInfo.

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryNetworkClient.java
snalli: yup done

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryNetworkClient.java
snalli: Brought it back, the cache is used to store metadata fetched during the first part of replication - get-metadata. This  is later used to fetch blob data. This reduces the number of metadata calls to azure.

justinlin-linkedin (approved):
LGTM

JingQianCloud (approved):

PR #2689, opened 2024-01-12
Fix NPE in getReplicaIdsByState when partition only exist in one DC

justinlin-linkedin:
## Summary

When creating new partitions, we create them in on DC first and make sure they are working, then move on to other DCs. This would put cluster in a state where some partitions only exist in one DC and not in other DCs. 

This PR fixes an NPE issue in getReplicaIdsByState when partitions are newly created in only one DC.


MikeDafi (approved):
LGTM, makes sense when a resource doesn't exist in another colo yet, especially in terms of a cluster uplift, we would need to continue

PR #2690, opened 2024-01-18
Fail storage manager when too many disk failed

justinlin-linkedin:
## Summary
Fail the initialization of StorageManager when there are too many disks fail to start or marked as unavailable. If enough disks fail to start up, there is no way ambry-server would be able to function. Then just fail the ambry-server so the helix would move all the replicas out sometime.

## Test
Unit test


Arun-LinkedIn (approved):

JingQianCloud (approved):

justinlin-linkedin:
Fail the initialization of StorageManager when there are too many disks fail to start or marked as unavailable. If enough disks fail to start up, there is no way ambry-server would be able to function. Then just fail the ambry-server so the helix would move all the replicas out sometime.

PR #2693, opened 2024-01-19
Remove blob store directory when it can't start

justinlin-linkedin:
## Summary

When a blob store can't start (while some other blob stores can start), we currently just mark this blob store as down and don't do anything about it. This error would require human intervention. Someone has to go to ambry server host, stop jvm process and remove the replica directory and then restart the process and hope replication would work to restore all the blobs.

This PR tries to automate this process by removing the blob store directory and recreate a new one.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
@@ -243,6 +243,14 @@ public class DiskManager {
     }
   }
 
+  /**
+   * Maybe recover the blob stores that failed to start. The configuration has to be set to true and the number
+   * of failed blob stores are not the same as the total number on the disk, then we would try to recover each
+   * blob store. We only recover the blob store if the exception thrown in the start method is StoreException and
+   * the store error code is one of the recoverable error codes.
+   * @param numStoreFailures
+   * @param startExceptions
+   */
   void maybeRecoverBlobStores(int numStoreFailures, Map<PartitionId, Exception> startExceptions) {
     if (!storeConfig.storeRemoveDirectoryAndRestartBlobStore || numStoreFailures == stores.size()) {
       return;
--- a/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
@@ -36,6 +36,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -90,6 +91,9 @@ public class DiskManager {
   private final DiskHealthCheck diskHealthCheck;
   // Have a dedicated scheduler for persisting index segments to ensure index segments are always persisted
   private final ScheduledExecutorService indexPersistScheduler;
+  private final EnumSet<StoreErrorCodes> recoverableStoreErrorCodes =
+      EnumSet.of(StoreErrorCodes.Log_File_Format_Error, StoreErrorCodes.Index_File_Format_Error,
+          StoreErrorCodes.Log_End_Offset_Error, StoreErrorCodes.Index_Recovery_Error);
 
   private static final Logger logger = LoggerFactory.getLogger(DiskManager.class);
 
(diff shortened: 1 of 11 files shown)


ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -199,6 +202,33 @@ void start(boolean shouldRemoveUnexpectedDirs) throws InterruptedException {
       }
       if (numStoreFailures.get() > 0) {
         logger.error("Could not start {} out of {} stores on the disk {}", numStoreFailures.get(), stores.size(), disk);
+        if (storeConfig.storeRemoveDirectoryAndRestartBlobStore && numStoreFailures.get() != stores.size()) {
+          for (Map.Entry<PartitionId, BlobStore> entry : stores.entrySet()) {
```
Arun-LinkedIn, line 206: nit: May be extract the code to a separate method?

Arun-LinkedIn (approved):
lgtm!

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -199,6 +202,33 @@ void start(boolean shouldRemoveUnexpectedDirs) throws InterruptedException {
       }
       if (numStoreFailures.get() > 0) {
         logger.error("Could not start {} out of {} stores on the disk {}", numStoreFailures.get(), stores.size(), disk);
+        if (storeConfig.storeRemoveDirectoryAndRestartBlobStore && numStoreFailures.get() != stores.size()) {
```
JingQianCloud, line 205: What's the reasons so far the BlobStore cannot start?
If it's one log segment or index segment is broken, it's reasonable to remove it.

But is there any other possibilities which cause almost all BlobStore fails to boot up?
If there is any software bug which blocks booting, we'll remove all the blobs. It's not recoverable. 

I see that you have the check the numberStoreFailures != stores.size and there is shouldRemoveDirectory check.

Can we be more cautions? I don't have a good solution yet. 
1. Maybe we add one threshold, like  failedBlob threshold, only when the failed number is under that threshold, we'll remove blobs?
2. Or Just every bootup, no matter how many failed blobs, we only recover one.

justinlin-linkedin: we currently have several error cases in clusters.

1. EndOffset of log is less than end offset of index
2. Recovery results in duplicate puts
3. Some index segment file only exist in directory metadata, but doesn't really exist.
4. File corruptions

These are the ones that I saw right now.

JingQianCloud: Is that possible to pass these "error messages" into the Exception content? And then we only wipe out disk if it fails to these listed error messages.

justinlin-linkedin: I will add some new error code and use these code to decide whether to remove the directory

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -89,6 +91,9 @@ public class DiskManager {
   private final DiskHealthCheck diskHealthCheck;
   // Have a dedicated scheduler for persisting index segments to ensure index segments are always persisted
   private final ScheduledExecutorService indexPersistScheduler;
+  private final EnumSet<StoreErrorCodes> recoverableStoreErrorCodes =
+      EnumSet.of(StoreErrorCodes.Log_File_Format_Error, StoreErrorCodes.Index_File_Format_Error,
+          StoreErrorCodes.Log_End_Offset_Error, StoreErrorCodes.Index_Recovery_Error);
```
justinlin-linkedin, line 96: should I use a configuration?

JingQianCloud (approved):
LGTM.
Thanks for taking care of that. The broken replica issue has been there for a long time and we didn't have an automated way to fix it.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
JingQianCloud: That's fine for now, I guess

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -235,6 +243,52 @@ void start(boolean shouldRemoveUnexpectedDirs) throws InterruptedException {
     }
   }
 
+  void maybeRecoverBlobStores(int numStoreFailures, Map<PartitionId, Exception> startExceptions) {
+    if (!storeConfig.storeRemoveDirectoryAndRestartBlobStore || numStoreFailures != stores.size()) {
```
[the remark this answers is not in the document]
ankagrawal: Why dont we want to attempt restart when only a subset of stores fail?

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -139,10 +140,10 @@ public class DiskManager {
     this.replicaStatusDelegates = replicaStatusDelegates;
     this.stoppedReplicas = stoppedReplicas;
     expectedDirs.add(reserveFileDir.getAbsolutePath());
+    diskMetrics = new DiskMetrics(storeMainMetrics.getRegistry(), disk.getMountPath(),
```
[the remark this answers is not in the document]
justinlin-linkedin: not exactly, we don't have to create multiple disk metrics, that's why I only create one for it. But creating different disk metrics is not wrong, since each disk metric only has the references to the real histogram, counter metrics. There is only one metric being created in the MetricRegistry. And different disk metric objects are just referencing to the same metric.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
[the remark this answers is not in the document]
justinlin-linkedin: 1. It's going to be ok, since recovery means first remove the directory. And then we are left with an empty replica directory to restart. This would be super fast since the majority of the starting time is allocated at reading index files.
2. Changed the logic. The correct logic should be, if all stores fail, we don't want to recover. Since that's a strong indication that the disk has some issues.

justinlin-linkedin (commented):
I will add another test case to make sure we don't recover when the store error code is not recoverable.

ankagrawal (approved):

PR #2694, opened 2024-01-23
[S3 API] Add support for parsing S3 requests and listing blobs in a container

Arun-LinkedIn:
This PR adds changes to:
1. Support S3 endpoint by converting S3 request to named blob request internally
2. Support S3 list API as per https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/account/Container.java
+++ b/ambry-api/src/main/java/com/github/ambry/account/Container.java
@@ -126,6 +126,12 @@ public class Container {
   public static final String DEFAULT_PRIVATE_CONTAINER_NAME = "default-private-container";
 
   /**
+   * Default name for the containers associated with S3 APIs. Since, S3 requests on client side only take Account
+   * (i.e. Bucket) name, we use a default name for containers.
+   */
+  public static final String DEFAULT_S3_CONTAINER_NAME = "container-a";
+
+  /**
    * The status of {@link #UNKNOWN_CONTAINER}.
    */
   public static final ContainerStatus UNKNOWN_CONTAINER_STATUS = ContainerStatus.ACTIVE;
--- a/ambry-api/src/main/java/com/github/ambry/account/Container.java
+++ b/ambry-api/src/main/java/com/github/ambry/account/Container.java
@@ -129,7 +129,7 @@ public class Container {
    * Default name for the containers associated with S3 APIs. Since, S3 requests on client side only take Account
    * (i.e. Bucket) name, we use a default name for containers.
    */
-  public static final String DEFAULT_S3_CONTAINER_NAME = "container-a";
+  public static final String DEFAULT_S3_CONTAINER_NAME = "container-s3";
 
   /**
    * The status of {@link #UNKNOWN_CONTAINER}.
--- a/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
@@ -13,14 +13,19 @@
  */
 package com.github.ambry.rest;
 
+import com.github.ambry.account.Container;
+import com.github.ambry.frontend.Operations;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static com.github.ambry.rest.RestUtils.*;
+import static com.github.ambry.rest.RestUtils.InternalKeys.*;
 import static com.github.ambry.router.GetBlobOptions.*;
 
 
@@ -35,6 +40,8 @@ public class RequestPath {
   private static final char PATH_SEPARATOR_CHAR = '/';
   private static final String PATH_SEPARATOR_STRING = String.valueOf(PATH_SEPARATOR_CHAR);
   private static final String SEGMENT = SubResource.Segment.toString();
+  private static final Logger logger = LoggerFactory.getLogger(RequestPath.class);
+  private static final String S3_PATH = PATH_SEPARATOR_CHAR + Operations.S3;
 
   /**
    * Parse the request path (and additional headers in some cases). The path will match the following regex-like
@@ -74,6 +81,13 @@ public class RequestPath {
     } catch (IllegalArgumentException e) {
       throw new RestServiceException("Invalid URI path", e, RestServiceErrorCode.BadRequest);
     }
+
+    if (path.startsWith(S3_PATH)) {
+      // Convert to named blob request internally
+      path = getNamedBlobPath(path);
+      restRequest.setArg(S3_REQUEST, "true");
+    }
+
     return parse(path, restRequest.getArgs(), prefixesToRemove, clusterName);
   }
 
@@ -297,5 +311,30 @@ public class RequestPath {
       }
     }
     return nextPathSegmentOffset;
+  }
+
+  /**
+   * Get named blob path from S3 request path
+   * @param path s3 request path
+   * @return named blob request path
+   */
+  private static String getNamedBlobPath(String path) {
+    // S3 requests to Ambry are in the form "/s3/account-name/key-name". We convert it to named blob path
+    // "/named/account-name/container-name/key-name" internally.
+    // For ex: for S3 path, "/s3/named-blob-sandbox/checkpoints/87833badf879a3fc7bf151adfe928eac/chk-1/_metadata",
+    // the corresponding named blob path is "/named/named-blob-sandbox/container-a/checkpoints/87833badf879a3fc7bf151adfe928eac/chk-1/_metadata".
+    // Please note that we hardcode container-name to 'container-a'.
+
+    int accountStart = S3_PATH.length() + PATH_SEPARATOR_STRING.length();
+    int accountEnd = path.indexOf(PATH_SEPARATOR_CHAR, accountStart);
+    if (accountEnd == -1) {
+      accountEnd = path.length();
+    }
+    String accountName = path.substring(accountStart, accountEnd);
+    String containerName = Container.DEFAULT_S3_CONTAINER_NAME;
+    String keyName = path.substring(accountEnd);
+    String namedBlobPath = "/named/" + accountName + "/" + containerName + (keyName.length() > 0 ? keyName : "");
+    logger.info("S3 API | Converted S3 request path {} to NamedBlob path {}", path, namedBlobPath);
+    return namedBlobPath;
   }
 }
--- a/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
@@ -82,7 +82,7 @@ public class RequestPath {
       throw new RestServiceException("Invalid URI path", e, RestServiceErrorCode.BadRequest);
     }
 
-    if (path.startsWith(S3_PATH)) {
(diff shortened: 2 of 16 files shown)


ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -125,6 +125,12 @@ public class Container {
    */
   public static final String DEFAULT_PRIVATE_CONTAINER_NAME = "default-private-container";
 
+  /**
+   * Default name for the containers associated with S3 APIs. Since, S3 requests on client side only take Account
+   * (i.e. Bucket) name, we use a default name for containers.
+   */
+  public static final String DEFAULT_S3_CONTAINER_NAME = "container-a";
```
JingQianCloud, line 132: Is this our final solution or temporary solution?
I thought we would go with
1. The endpoint is
https://ambry-frontend-main.ambry-frontend.xxxx.linkedin.com:3183/s3
2. The S3 path is
```
"s3://<ambry-account-name>/<ambry-container-name>/<folder-name>/<file-name>"
```

ambry-frontend/src/main/java/com/github/ambry/frontend/Contents.java
```diff
@@ -0,0 +1,72 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ */
+
+package com.github.ambry.frontend;
+
+import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;
+
+
+public class Contents {
```
JingQianCloud, line 21: For all the new classes,
do we want to create one folder as XXX/ambry/s3,
then put all these classes, especially the Content.java, ListBucketResult there?

ambry-frontend/src/main/java/com/github/ambry/frontend/S3ListHandler.java
```diff
@@ -0,0 +1,101 @@
...
+import java.util.GregorianCalendar;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import org.json.JSONObject;
+
+
+/**
+ * Handles S3 requests for listing blobs that start with a provided prefix.
+ */
+public class S3ListHandler {
```
JingQianCloud, line 27: I like this approach which separates from the named blob code. It's much clean.

ambry-api/src/main/java/com/github/ambry/account/Container.java
Arun-LinkedIn: This was actually @justinlin-linkedin's idea in the doc.  Since S3 has the concept of only buckets, I am also wondering that some S3 APIs may assume that path after `/bucket-name/` is the `key-name`.  For example, they have `HEAD /bucket-name` API. 

We may have to also tell S3 customers that they can have only 1 container (with name `container-a`) under Ambry accounts. 

What do you think?

ambry-frontend/src/main/java/com/github/ambry/frontend/Contents.java
Arun-LinkedIn: Sure. I will create a new folder /s3/ and put these new classes there.

ambry-api/src/main/java/com/github/ambry/account/Container.java
JingQianCloud: Sounds good to me so far.
Let's discuss it with our customers next time including the ttl and path.

ambry-frontend/src/main/java/com/github/ambry/frontend/S3ListHandler.java
```diff
@@ -0,0 +1,101 @@
...
+      contentsList.add(contents);
+    }
+
+    listBucketResult.setMaxKeys(namedBlobRecords.size());
+    listBucketResult.setDelimiter("/");
+    listBucketResult.setEncodingType("url");
+
+    listBucketResult.setContents(contentsList);
+
+    ObjectMapper objectMapper = new XmlMapper();
```
weisong44, line 96: Nit: perhaps keep a static instance of this, I think it's thread-safe.

ambry-api/src/main/java/com/github/ambry/account/Container.java
Arun-LinkedIn: Hmm, I am thinking that we shouldn't have any scalability concerns. We use container as a logically group blobs with similar properties (encrypted, cacheable, media-scan). 

Containers also provide ability to compact all blobs in it in background. It may happen that compaction of containers may take longer time. But I am thinking that it should be okay as it is asynchronous process and no one is actually waiting on it.

JingQianCloud (approved):
LGTM.
One nic: Since we put every class to s3 folder now, do we still need the "S3" prefix for "S3ListHandler"? 
Is "ListHandler" good enough?

ambry-api/src/main/java/com/github/ambry/account/Container.java
weisong44: Got it, this makes sense. I think then we could use different account for different quality of service.

justinlin-linkedin: Do we want to use a different container name for default s3 container? container-a is too common amount accounts. If customers want to enable s3 API for its application, it should put files under s3 specific container. Something like "s3" might be better.

ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
```diff
@@ -298,4 +312,30 @@ private static int matchPathSegments(String path, int pathOffset, String pathSeg
...
+    if (accountEnd == -1) {
+      accountEnd = path.length();
+    }
+    String accountName = path.substring(accountStart, accountEnd);
+    String containerName = Container.DEFAULT_S3_CONTAINER_NAME;
+    String keyName = path.substring(accountEnd);
+    String namedBlobPath =
+        PATH_SEPARATOR_STRING + Operations.NAMED_BLOB + PATH_SEPARATOR_STRING + accountName + PATH_SEPARATOR_STRING
+            + containerName + (keyName.length() > 0 ? keyName : "");
+    logger.info("S3 API | Converted S3 request path {} to NamedBlob path {}", path, namedBlobPath);
```
justinlin-linkedin, line 335: make this as a debug log.

ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
```diff
@@ -298,4 +312,30 @@ private static int matchPathSegments(String path, int pathOffset, String pathSeg
...
+    // S3 requests to Ambry are in the form "/s3/account-name/key-name". We convert it to named blob path
+    // "/named/account-name/container-name/key-name" internally.
+    // For ex: for S3 path, "/s3/named-blob-sandbox/checkpoints/87833badf879a3fc7bf151adfe928eac/chk-1/_metadata",
+    // the corresponding named blob path is "/named/named-blob-sandbox/container-a/checkpoints/87833badf879a3fc7bf151adfe928eac/chk-1/_metadata".
+    // Please note that we hardcode container-name to 'container-a'.
+    int accountStart = S3_PATH.length() + PATH_SEPARATOR_STRING.length();
+    int accountEnd = path.indexOf(PATH_SEPARATOR_CHAR, accountStart);
+    if (accountEnd == -1) {
+      accountEnd = path.length();
+    }
```
justinlin-linkedin: this seems to be a 404 issue. If request goes like "/s3/", then we should return 404.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -293,8 +296,13 @@ public void handleGet(final RestRequest restRequest, final RestResponseChannel r
             (result, exception) -> submitResponse(restRequest, restResponseChannel, result, exception));
       } else if (requestPath.matchesOperation(Operations.NAMED_BLOB)
           && NamedBlobPath.parse(requestPath, restRequest.getArgs()).getBlobName() == null) {
-        namedBlobListHandler.handle(restRequest, restResponseChannel,
-            (result, exception) -> submitResponse(restRequest, restResponseChannel, result, exception));
+        if (isS3Request(restRequest)) {
```
justinlin-linkedin, line 299: We are converting the path in the RequestPath so here we have to rely on the isS3Request method. What if this is an upload or download request? We would have to use directly NamedBlob handlers directly?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
```diff
@@ -0,0 +1,151 @@
...
+        result.readInto(writableChannel, null).get();
+        JSONObject jsonObject = FrontendUtils.readJsonFromChannel(writableChannel);
+        Page<NamedBlobListEntry> page = Page.fromJson(jsonObject, NamedBlobListEntry::new);
+        ReadableStreamChannel readableStreamChannel = serializeAsXml(page);
+        restResponseChannel.setHeader(RestUtils.Headers.DATE, new GregorianCalendar().getTime());
+        restResponseChannel.setHeader(RestUtils.Headers.CONTENT_TYPE, "application/xml");
+        restResponseChannel.setHeader(RestUtils.Headers.CONTENT_LENGTH, readableStreamChannel.getSize());
+        callback.onCompletion(readableStreamChannel, null);
+      } catch (Exception e) {
+        throw new RuntimeException(e);
```
justinlin-linkedin, line 90: This should be a 503 status. Let's just return a runtime exception here.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
```diff
@@ -0,0 +1,151 @@
...
+      try {
+        // Convert from json response to S3 xml response as defined in
+        // https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_ResponseSyntax.
+        RetainingAsyncWritableChannel writableChannel = new RetainingAsyncWritableChannel();
+        result.readInto(writableChannel, null).get();
+        JSONObject jsonObject = FrontendUtils.readJsonFromChannel(writableChannel);
+        Page<NamedBlobListEntry> page = Page.fromJson(jsonObject, NamedBlobListEntry::new);
+        ReadableStreamChannel readableStreamChannel = serializeAsXml(page);
+        restResponseChannel.setHeader(RestUtils.Headers.DATE, new GregorianCalendar().getTime());
+        restResponseChannel.setHeader(RestUtils.Headers.CONTENT_TYPE, "application/xml");
```
justinlin-linkedin, line 86: Does S3 API only return xml object? Or does it decide the serialization format based on "Accept" header from request?

Arun-LinkedIn: Hmm, I think S3 ListObjects API only returns xml response - https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
Arun-LinkedIn: Done..

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
justinlin-linkedin: sure

justinlin-linkedin (approved):

PR #2700, opened 2024-01-30
PutBlob, add one option: skipCompositeChunk.

JingQianCloud:
PutBlob, we add one option: skipCompositeChunk.
When it's true, we don't generate the composite blob.
And then we return serialized PutBlobMetaInfo string instead of the composite blob id.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
@@ -20,6 +20,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.ambry.utils.Pair;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import org.json.JSONArray;
 import org.json.JSONObject;
@@ -73,7 +74,7 @@ public class PutBlobMetaInfo {
    * @return the list of chunk with blob id and blob size
    */
   public List<Pair<String, Long>> getOrderedChunkIdSizeList() {
-    return orderedChunkIdSizeList;
+    return Collections.unmodifiableList(orderedChunkIdSizeList);
   }
 
   /**
--- a/ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright 2024 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+
+package com.github.ambry.router;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.ambry.utils.Pair;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+
+/**
+ * This class holds information about the data chunk list or other metadata of the PutOperation
+ * It contains a list of key size pair for the blob's data chunks, reserved metadata etc.
+ */
+public class PutBlobMetaInfo {
+  private static final String BLOB = "blob";
+  private static final String SIZE = "size";
+  private static final String CHUNKS = "chunks";
+  private static final String RESERVED_METADATA_CHUNK_ID = "reservedMetadataChunkId";
+
+  // reserved metadata chunk id
+  private final String reservedMetadataChunkId;
+  // the list of data chunk pairs <blob id, blob size> in order
+  private final List<Pair<String, Long>> orderedChunkIdSizeList;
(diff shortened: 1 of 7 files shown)


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -2077,6 +2084,10 @@ void poll(RequestRegistrationCallback<PutOperation> requestRegistrationCallback)
       }
     }
 
+    private boolean isS3MultipartPartUpload() {
```
JingQianCloud, line 2087: When we have the MultiPart code, we may set a flag called "Skip-Composite-Blob" flag.
So here we rely on this flag to skip the compositeBlob step.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -2089,7 +2100,34 @@ private void finalizeMetadataChunk() {
...
+          jsonObject.accumulate("chunks", oneObject);
+        }
+        jsonObject.put("ttl", ""); // placeholder
+        jsonObject.put("reservedMetadataId", ""); // placeholder
+        jsonObject.put("session", ""); // placeholder
+        String result = jsonObject.toString();
+        logger.info("S3 5MB : finalizeMetadataChunk generated json {}", result);
+
+        // use the S3_BLOB_LIST to pass the value
+        restRequest.setArg(S3_BLOB_LIST, result);
```
JingQianCloud, line 2129: Ideally we let S3 code path to generate the ETag, not here. 
Anyway to pass the dataChunk raw data to the caller?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -2089,7 +2100,34 @@ private void finalizeMetadataChunk() {
               passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(),
               passedInBlobProperties.isEncrypted(), passedInBlobProperties.getExternalAssetTag(),
               passedInBlobProperties.getContentEncoding(), passedInBlobProperties.getFilename(), null);
-      if (isStitchOperation() || getNumDataChunks() > 1) {
+
+      if (isS3MultipartPartUpload()) {
```
justinlin-linkedin, line 2104: I don't think this is a good idea to have s3 knowledge inside of PutOperation. I would propose another way of convey these information in and out. Basically, we have extra input (a flag to indicate if we want to have composite blob id) and extra output (a list of blob ids in a string). I think we can use PutBlobOptions to pass this information, something PutBlobOption.ignoreCompositeBlobId, and just return a string (maybe in base64) that contains all blob ids and sizes. Something like that.

justinlin-linkedin (approved):
LGTM

ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
```diff
@@ -0,0 +1,137 @@
...
+   */
+  public int getChunkNumber() {
+    return orderedChunkIdSizeList.size();
+  }
+
+  /**
+   * Get the chunk list
+   * @return the list of chunk with blob id and blob size
+   */
+  public List<Pair<String, Long>> getOrderedChunkIdSizeList() {
```
weisong44, line 75: Nit: would it make sense to make the returned list immutable?

weisong44 (commented):
I don't think I know enough to do a proper review, just left a very minor comment :slightly_smiling_face:

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -2089,7 +2101,25 @@ private void finalizeMetadataChunk() {
               passedInBlobProperties.getAccountId(), passedInBlobProperties.getContainerId(),
               passedInBlobProperties.isEncrypted(), passedInBlobProperties.getExternalAssetTag(),
               passedInBlobProperties.getContentEncoding(), passedInBlobProperties.getFilename(), null);
-      if (isStitchOperation() || getNumDataChunks() > 1) {
+
+      if (options.skipCompositeChunk()) {
+        // close the request as the single blob. we don't generate the composite blob.
+        blobId = (BlobId) indexToChunkIdsAndChunkSizes.get(0).getFirst();
```
Arun-LinkedIn, line 2107: Setting blobId to the first chunk ID looks bit hacky :) But, I also don't know any better way.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -2089,7 +2101,25 @@ private void finalizeMetadataChunk() {
...
+        // now generate the PutBlobMetaInfo which includes the data chunk list
+        // values returned are in the right order as TreeMap returns them in key-order.
+        // .stream.map.collect is supposed to return in order
+        List<Pair<String, Long>> orderedChunkList = new ArrayList<>(indexToChunkIdsAndChunkSizes.values()
+            .stream()
+            .map(p -> new Pair<>(p.getFirst().toString(), p.getSecond()))
+            .collect(Collectors.toList()));
+        PutBlobMetaInfo putBlobMetaInfoObj = new PutBlobMetaInfo(orderedChunkList, reservedMetadataChunkId.getID());
+        putBlobMetaInfo = PutBlobMetaInfo.serialize(putBlobMetaInfoObj);
+        logger.info("S3 5MB : finalizeMetadataChunk orderedChunkIdList is {} {}", indexToChunkIdsAndChunkSizes,
```
Arun-LinkedIn, line 2120: nit: Would `debug` level be sufficient?

Arun-LinkedIn (approved):
Looks good to me. Just had couple of nit comments.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
JingQianCloud: That's true. Added comment to explain it. Hopefully it's less confusing. :-)

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
JingQianCloud: Changed.

PR #2702, opened 2024-02-01
Add metrics to get the time taken for deserializing and validating signed ids for named blob stitch puts

ankagrawal:
For very large blobs, the number of chunks to deserialize and validate can be very very high (~250000 for 1TiB blob). This metrics tracks the time taken for deserialization and validation of stitch blob uploads.


JingQianCloud (approved):

JingQianCloud (commented):
Approved.
Looks good to me. Not sure if you want to add one logger message when the time spent is longer than 10s or so.

PR #2704, opened 2024-02-01
[WIP] Detect conflict during named blob put operation before upload or stitch operation.

ankagrawal:


ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -208,24 +208,34 @@ private Callback<Void> securityProcessRequestCallback() {
...
-          }
-          RetainingAsyncWritableChannel channel =
-              new RetainingAsyncWritableChannel(frontendConfig.maxJsonRequestSizeBytes);
-          restRequest.readInto(channel, fetchStitchRequestBodyCallback(channel, blobInfo));
+        if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
+          accountAndContainerInjector.injectDatasetForNamedBlob(restRequest);
+          frontendMetrics.addDatasetVersionRate.mark();
+          addDatasetVersion(blobInfo.getBlobProperties(), restRequest);
+        }
+        idConverter.detectConflict(restRequest, detectConflictCallback(blobInfo));
```
Arun-LinkedIn, line 216: Hmm, this will result in checking for conflicts for regular named blob uploads as well, i.e. not just `stitch` requests? Wondering if it may increase in latency of regular uploads due to additional call to mysql db. Would it be better to check for conflicts only for `stitch` requests?

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -112,37 +113,97 @@ public Future<String> convert(RestRequest restRequest, String input, BlobInfo bl
...
+      frontendMetrics.idConverterDetectConflictRate.mark();
+      if (restRequest.getRestMethod() == RestMethod.PUT && RestUtils.getRequestPath(restRequest)
+          .matchesOperation(Operations.NAMED_BLOB) && !RestUtils.isUpsertForNamedBlob(restRequest.getArgs())) {
+        // A named blob put without upsert is allowed only if the named blob does not exist.
+        // If this is a named blob put request without upsert, then we check if a mapping for the name already exists.
+        NamedBlobPath namedBlobPath = NamedBlobPath.parse(RestUtils.getRequestPath(restRequest), restRequest.getArgs());
+
+        detectionFuture = getNamedBlobDb().get(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
+            namedBlobPath.getBlobName(), GetOption.Include_All).thenApply(namedBlobRecord -> {
+          LOGGER.info("A mapping for the namedBlob {} already exists. Original BlobId: {}",
```
JingQianCloud, line 171: What about if the named blob is deleted?
If the named blob is deleted, we should allow upload even without the upsert option.
If it's the case, we need check the db record status.

ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -112,37 +113,97 @@ public Future<String> convert(RestRequest restRequest, String input, BlobInfo bl
...
+    /**
+     * Detects if the ID in the request can cause a conflict.
+     * @param restRequest @link RestRequest} representing the request.
+     * @return the {@link CompletionStage} that will be completed with the conflict status.
+     * @throws RestServiceException if the operation cannot be performed.
+     */
+    private CompletionStage<Boolean> detectConflict(RestRequest restRequest) throws RestServiceException {
+      final CompletionStage<Boolean> detectionFuture;
+      frontendMetrics.idConverterDetectConflictRate.mark();
+      if (restRequest.getRestMethod() == RestMethod.PUT && RestUtils.getRequestPath(restRequest)
```
JingQianCloud, line 163: Can we verify the Dataset behavior?
Does it allow update with the same version number if upsert is not set?

PR #2707, opened 2024-02-03
[S3 API] Remove Ambry header from response

weisong44:
Changes introduced in this PR are
* Introduced base class S3BaseHandler, which allows pre- and post-processing
* Added post-processing to remove Ambry header from response right before invoking the final callback
* Some basic erroring checking and logging

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2024 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *
+ */
+package com.github.ambry.frontend.s3;
+
+import com.github.ambry.commons.Callback;
+import com.github.ambry.frontend.Operations;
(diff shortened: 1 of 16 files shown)


JingQianCloud (approved):
I like the idea to have the base S3 class.
Code looks good to me.
Let's see if Arun has any opinion if the S3 base class fits all his S3 handlers.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
```diff
@@ -0,0 +1,90 @@
...
+   * @param restRequest the {@link RestRequest} that contains the request headers and body.
+   * @param restResponseChannel the {@link RestResponseChannel} that contains the response headers and body.
+   * @param callback the {@link Callback} to invoke when the response is ready (or if there is an exception).
+   * @throws RestServiceException exception when the processing fails
+   */
+  public void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
+      Callback<R> callback) throws RestServiceException {
+    String path = ((RequestPath) restRequest.getArgs().get(REQUEST_PATH)).getOperationOrBlobId(true);
+    if (!path.startsWith(Operations.NAMED_BLOB)) {
+      throw new RuntimeException("S3 request handler can only handle named blob requests");
```
Arun-LinkedIn, line 65: Just a question. Would it be better to send `400 bad request` to user by completing the callback with `new RestServiceException("S3 request handler can only handle named blob requests", RestServiceErrorCode.BadRequest` ? I think `RuntimeException` translates to `500 internal server error` to user.

Arun-LinkedIn (approved):
Looks good to me. Thank you for introducing the base class :) Just had one minor question.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
weisong44: My understanding is that this should never happen, i.e. we internally translate /s3 to /named and do other checks before we let the request proceed, so I chose 500. However, if there is a chance for client to construct to land here, then I agree that we should return 400.

PR #2713, opened 2024-02-13
[DRAFT] Recovery impl getblob

snalli:


ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryManager.java
```diff
@@ -107,30 +135,27 @@ public CloudToStoreReplicationManager(ReplicationConfig replicationConfig, Clust
...
-
-    // start background persistent thread
-    // start scheduler thread to persist index in the background
-    scheduler.scheduleAtFixedRate(persistor, replicationConfig.replicationTokenFlushDelaySeconds,
-        replicationConfig.replicationTokenFlushIntervalSeconds, TimeUnit.SECONDS);
-
+    // Use a static clustermap file on the host to know partitions assigned to the host.
+    // Assume that there is no helix/zookeeper available post disaster.
+    // In this design, server does not talk to vcr but directly to Azure.
+    this.clusterMap.getAllPartitionIds(clusterMapConfig.MAX_REPLICAS_ALL_DATACENTERS).forEach(partition -> {
```
Arun-LinkedIn, line 149: May be this needs to be changed to `clusterMapConfig.clusterMapDefaultPartitionClass` instead of hardcoded `clusterMapConfig.MAX_REPLICAS_ALL_DATACENTERS` ? Video cluster has default partition class configured as `video-3111`.

Also, QQ.. Would this recovery code be running on ambry server hosts? If so, would we need to filter partitions that belong to this particular host?

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryThread.java
```diff
@@ -0,0 +1,182 @@
...
+    super(threadName, findTokenHelper, clusterMap, correlationIdGenerator, dataNodeId, networkClient, replicationConfig,
+        replicationMetrics, notification, storeKeyConverter, transformer, metricRegistry, replicatingOverSsl,
+        datacenterName, responseHandler, time, replicaSyncUpManager, skipPredicate, leaderBasedReplicationAdmin);
+    this.fileDescriptorCache = new AmbryCache(String.format("%s-FDCache", threadName), true,
+        replicationConfig.maxBackupCheckerReportFd, metricRegistry);
+    this.recoveryManager = recoveryManager;
+  }
+
+  /**
+   * Persists token to cloud in each replication cycle
```
Arun-LinkedIn, line 87: nit: Persists token `locally`?

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryThread.java
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  /**
+   * Truncates a file and then writes to it.
+   * Creates the file if absent.
+   * @param filePath Path of the file in the system
+   * @param text Text to append
+   * @return True if append was successful, false otherwise
+   */
+  protected boolean truncateAndWriteToFile(String filePath, String text) {
```
Arun-LinkedIn, line 153: nit: Return type could be `void` since it is not used?

Arun-LinkedIn (approved):
Approved with couple of comments..

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryManager.java
snalli: Thanks, I will add the filtering. The disaster scenario assumed is that we have no helix and each server is configured using a static clustermap. 

As for why this code runs in server, there are only two places recovery code can run - server or VCR. VCR is just a big bottleneck and a proxy to Azure. It doesn't server any other purpose. A server is fully capable of directly fetching data from Azure and would be a lot faster. We discussed this in a meeting last sem.

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryThread.java
snalli: yes

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryThread.java
snalli: This is fine. A function must always return something, the caller can choose to use it or not.

ambry-cloud/src/main/java/com/github/ambry/cloud/RecoveryThread.java
```diff
@@ -0,0 +1,182 @@
...
+    EnumSet<StandardOpenOption> options = EnumSet.of(StandardOpenOption.WRITE);
```
JingQianCloud, line 154: I think you want TRUNCATE_EXISTING, CREATE and WRITE, three options.

JingQianCloud (approved):
Have one comment embedded. Please comment or address it before merge. 

Another general comment, when you squash and commit the PR, please check the commit comment, only the summary is needed. Don't include all the revision comment for all the 18 commits.

PR #2714, opened 2024-02-14
[vcr-2.0] Deprecate V1 tests

snalli:
Deprecate V1 tests for VCR, since V1 is gone and the tests just take too much time


ambry-cloud/src/test/java/com/github/ambry/cloud/CloudBlobMetadataTest.java
```diff
@@ -23,6 +23,7 @@
 import java.util.concurrent.TimeUnit;
 import org.apache.commons.lang3.ArrayUtils;
 import org.junit.Before;
+import org.junit.Ignore;
```
ankagrawal, line 26: nit: we should remember to get rid of all these files eventually.

ankagrawal (approved):
lgtm

PR #2717, opened 2024-02-22
Remove hardcoding of container name for S3 requests

Arun-LinkedIn:
This PR removes hardcoding of container name for S3 Requests so that users don't have restriction of creating separate Ambry accounts for different use cases.

It also consists of a minor fix in `S3PutHandler` to send blob ID correctly in ETag header.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
+++ b/ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
@@ -13,7 +13,6 @@
  */
 package com.github.ambry.rest;
 
-import com.github.ambry.account.Container;
 import com.github.ambry.frontend.Operations;
 import java.io.UnsupportedEncodingException;
 import java.net.URLDecoder;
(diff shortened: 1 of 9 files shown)


JingQianCloud (approved):
LGTM.
I'll rebase my PR when your RP is merged.

ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
```diff
@@ -319,25 +318,23 @@ private static int matchPathSegments(String path, int pathOffset, String pathSeg
...
-          String.format("S3 Path must have format '/s3/<account_name>/<key-name>.  Received path='%s'", path),
-          RestServiceErrorCode.BadRequest);
+          String.format("S3 Path must have format '/s3/<account_name>/container-name/<key-name>.  Received path='%s'",
+              path), RestServiceErrorCode.BadRequest);
     }
     String accountName = splitPath[1];
-    String containerName = Container.DEFAULT_S3_CONTAINER_NAME;
-    String keyName = splitPath.length > 2 ? splitPath[2] : "";
+    String containerName = splitPath[2];
+    String keyName = splitPath.length > 3 ? splitPath[3] : "";
```
weisong44, line 333: Should we even allow empty keys?

weisong44 (approved):
LGTM, left one minor comment

ambry-api/src/main/java/com/github/ambry/rest/RequestPath.java
Arun-LinkedIn: Hmm, I think for `HEAD bucket` requests, the key name can be empty. Due to that, kept the logic to accept empty keys for now.

PR #2720, opened 2024-02-24
Multipart part upload uses its own CallbackChain, don't use namedBlobPutUploader.

JingQianCloud:
S3 Multipart part upload don't use namedBlobPutHandler.
1. Part upload only have the security check and PutBlob. It doesn't need TtlUpdate, Stitch function, db operation or dataset logic. So create its own simple CallbackChain.
2. Removed the S3 part upload code from the namedBlobPutHandler.

After this, all the Multipart handlers have their own CallbackChain

- S3CreateMultipartUploadHandler
- S3CompleteMultipartUploadHandler
- S3MultipartUploadPartHandler

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
@@ -152,7 +152,6 @@ public class S3MultipartUploadPartHandler {
      */
     private Callback<Void> securityProcessRequestCallback() {
       return buildCallback(frontendMetrics.putSecurityProcessRequestMetrics, securityCheckResult -> {
-        //make sure this has been called after processRequest(permission check) since it needs to query dataset db.
         BlobInfo blobInfo = getBlobInfoFromRequest();
         securityService.postProcessRequest(restRequest, securityPostProcessRequestCallback(blobInfo));
       }, uri, logger, finalCallback);
@@ -227,35 +226,7 @@ public class S3MultipartUploadPartHandler {
       // TODO [S3] may consolidate and move it to a central place.
       long propsBuildStartTime = System.currentTimeMillis();
       accountAndContainerInjector.injectAccountContainerForNamedBlob(restRequest, frontendMetrics.putBlobMetricsGroup);
-      if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
-        accountAndContainerInjector.injectDatasetForNamedBlob(restRequest);
-      }
       BlobProperties blobProperties = RestUtils.buildBlobProperties(restRequest.getArgs());
-      Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
-      if (blobProperties.getTimeToLiveInSeconds() + TimeUnit.MILLISECONDS.toSeconds(
-          blobProperties.getCreationTimeInMs()) > Integer.MAX_VALUE) {
-        logger.debug("TTL set to very large value in PUT request with BlobProperties {}", blobProperties);
-        frontendMetrics.ttlTooLargeError.inc();
-      } else if (container.isTtlRequired() && (blobProperties.getTimeToLiveInSeconds() == Utils.Infinite_Time
-          || blobProperties.getTimeToLiveInSeconds() > frontendConfig.maxAcceptableTtlSecsIfTtlRequired)) {
-        String descriptor = RestUtils.getAccountFromArgs(restRequest.getArgs()).getName() + ":" + container.getName();
-        if (frontendConfig.failIfTtlRequiredButNotProvided) {
-          throw new RestServiceException(
-              "TTL < " + frontendConfig.maxAcceptableTtlSecsIfTtlRequired + " is required for upload to " + descriptor,
-              RestServiceErrorCode.InvalidArgs);
-        } else {
-          logger.debug("{} attempted an upload with ttl {} to {}", blobProperties.getServiceId(),
-              blobProperties.getTimeToLiveInSeconds(), descriptor);
-          frontendMetrics.ttlNotCompliantError.inc();
-          restResponseChannel.setHeader(RestUtils.Headers.NON_COMPLIANCE_WARNING,
-              "TTL < " + frontendConfig.maxAcceptableTtlSecsIfTtlRequired + " will be required for future uploads");
-        }
-      }
-      // inject encryption frontendMetrics if applicable
-      if (blobProperties.isEncrypted()) {
-        restRequest.getMetricsTracker()
-            .injectMetrics(frontendMetrics.putBlobMetricsGroup.getRestRequestMetrics(restRequest.isSslUsed(), true));
-      }
       Map<String, Object> userMetadataFromRequest = new HashMap<>(restRequest.getArgs());
       byte[] userMetadata = RestUtils.buildUserMetadata(userMetadataFromRequest);
       frontendMetrics.blobPropsBuildForNameBlobPutTimeInMs.update(System.currentTimeMillis() - propsBuildStartTime);
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
@@ -141,12 +141,8 @@ public class S3MultipartUploadPartHandler {
     private void start() {
       restRequest.getMetricsTracker()
           .injectMetrics(frontendMetrics.putBlobMetricsGroup.getRestRequestMetrics(restRequest.isSslUsed(), false));
-      try {
-        // Start the callback chain by parsing blob info headers and performing request security processing.
-        securityService.processRequest(restRequest, securityProcessRequestCallback());
-      } catch (Exception e) {
-        finalCallback.onCompletion(null, e);
-      }
+      // Start the callback chain by parsing blob info headers and performing request security processing.
+      securityService.processRequest(restRequest, securityProcessRequestCallback());
     }
 
     /**
@@ -209,20 +205,15 @@ public class S3MultipartUploadPartHandler {
      */
     private Callback<Void> securityProcessResponseCallback() {
       return buildCallback(frontendMetrics.putBlobSecurityProcessResponseMetrics, securityCheckResult -> {
-        try {
-          // Set the response status to 200 since Ambry named blob PUT has response as 201.
-          if (restResponseChannel.getStatus() == ResponseStatus.Created) {
-            restResponseChannel.setStatus(ResponseStatus.Ok);
-          }
-
-          // Set S3 ETag header
-          String blobId = (String) restResponseChannel.getHeader(LOCATION);
-          restResponseChannel.setHeader("ETag", blobId);
-          //TODO [S3]: remove x-ambry- headers
-          finalCallback.onCompletion(null, null);
-        } catch (RestServiceException e) {
-          finalCallback.onCompletion(null, e);
+        // Set the response status to 200 since Ambry named blob PUT has response as 201.
+        if (restResponseChannel.getStatus() == ResponseStatus.Created) {
+          restResponseChannel.setStatus(ResponseStatus.Ok);
         }
+
+        // Set S3 ETag header
+        String blobId = (String) restResponseChannel.getHeader(LOCATION);
+        restResponseChannel.setHeader("ETag", blobId);
+        finalCallback.onCompletion(null, null);
       }, restRequest.getUri(), logger, finalCallback);
     }
 
@@ -233,6 +224,7 @@ public class S3MultipartUploadPartHandler {
      * @throws RestServiceException if there is an error while parsing the {@link BlobInfo} arguments.
      */
     private BlobInfo getBlobInfoFromRequest() throws RestServiceException {
+      // TODO [S3] may consolidate and move it to a central place.
       long propsBuildStartTime = System.currentTimeMillis();
       accountAndContainerInjector.injectAccountContainerForNamedBlob(restRequest, frontendMetrics.putBlobMetricsGroup);
       if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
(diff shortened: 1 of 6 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -65,28 +107,202 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+      this.uri = restRequest.getUri();
+    }
+
+    /**
+     * Start the chain by calling {@link SecurityService#processRequest}.
+     */
+    private void start() {
+      restRequest.getMetricsTracker()
+          .injectMetrics(frontendMetrics.putBlobMetricsGroup.getRestRequestMetrics(restRequest.isSslUsed(), false));
       try {
```
weisong44, line 144: Any exception should be caught in S3BaseHandler, shouldn't need to catch here.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -65,28 +107,202 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
-        callback.onCompletion(null, null);
-      } catch (RestServiceException e) {
-        callback.onCompletion(null, e);
+    /**
+     * Parse {@link BlobInfo} from the request arguments. This method will also ensure that the correct account and
+     * container objects are attached to the request.
+     * @return the {@link BlobInfo} parsed from the request arguments.
+     * @throws RestServiceException if there is an error while parsing the {@link BlobInfo} arguments.
+     */
+    private BlobInfo getBlobInfoFromRequest() throws RestServiceException {
```
weisong44, line 235: Can this be moved to a util class shared with NamedBlobPutHandler and PostBlobHandler?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -65,28 +107,202 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+
+    /**
+     * After {@link SecurityService#processResponse}, call {@code finalCallback}.
+     * @return a {@link Callback} to be used with {@link SecurityService#processResponse}.
+     */
+    private Callback<Void> securityProcessResponseCallback() {
+      return buildCallback(frontendMetrics.putBlobSecurityProcessResponseMetrics, securityCheckResult -> {
+        try {
+          // Set the response status to 200 since Ambry named blob PUT has response as 201.
+          if (restResponseChannel.getStatus() == ResponseStatus.Created) {
```
weisong44, line 214: Otherwise, should we abort?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -65,28 +107,202 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+            restResponseChannel.setStatus(ResponseStatus.Ok);
+          }
+
+          // Set S3 ETag header
+          String blobId = (String) restResponseChannel.getHeader(LOCATION);
+          restResponseChannel.setHeader("ETag", blobId);
+          //TODO [S3]: remove x-ambry- headers
```
weisong44, line 221: This should already be done in S3BaseHandler

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -65,28 +107,202 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+      }, uri, logger, finalCallback);
+    }
+
+    /**
+     * After {@link SecurityService#processResponse}, call {@code finalCallback}.
+     * @return a {@link Callback} to be used with {@link SecurityService#processResponse}.
+     */
+    private Callback<Void> securityProcessResponseCallback() {
+      return buildCallback(frontendMetrics.putBlobSecurityProcessResponseMetrics, securityCheckResult -> {
+        try {
```
weisong44, line 212: Don't think we need try/catch block here, exception should be handled in CallbackUtils.chainCallback()

weisong44 (commented):
Overall LGTM, left a few comments. Didn't give my approval, since I don't know enough about some of the business logic.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: You are right. Removed.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: Got you. Removed.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: The status can be "ResponseStatus.ok" as well.
If it's some error, exception should have been thrown.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: Got you.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: Yes, it has the potential to consolidate. 
But not sure if we should consolidate with generic NamedBlobPutHandler in case S3 has some special implementation.
I added the TODO. Let's revisit it after we have done with the S3 blob life cycle management.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -65,13 +107,104 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+    }
+
+    /**
+     * After {@link SecurityService#processRequest} finishes, call {@link SecurityService#postProcessRequest} to perform
+     * request time security checks that rely on the request being fully parsed and any additional arguments set.
+     * @return a {@link Callback} to be used with {@link SecurityService#processRequest}.
+     */
+    private Callback<Void> securityProcessRequestCallback() {
+      return buildCallback(frontendMetrics.putSecurityProcessRequestMetrics, securityCheckResult -> {
+        //make sure this has been called after processRequest(permission check) since it needs to query dataset db.
```
Arun-LinkedIn, line 155: nit: This comment can be removed? Seems to be associated with dataset db

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -80,13 +213,88 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+     * @return the {@link BlobInfo} parsed from the request arguments.
+     * @throws RestServiceException if there is an error while parsing the {@link BlobInfo} arguments.
+     */
+    private BlobInfo getBlobInfoFromRequest() throws RestServiceException {
+      // TODO [S3] may consolidate and move it to a central place.
+      long propsBuildStartTime = System.currentTimeMillis();
+      accountAndContainerInjector.injectAccountContainerForNamedBlob(restRequest, frontendMetrics.putBlobMetricsGroup);
+      if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
+        accountAndContainerInjector.injectDatasetForNamedBlob(restRequest);
+      }
```
Arun-LinkedIn, line 232: Since it is specific to dataset code, it can be removed?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -80,13 +213,88 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+      BlobProperties blobProperties = RestUtils.buildBlobProperties(restRequest.getArgs());
+      Container container = RestUtils.getContainerFromArgs(restRequest.getArgs());
+      if (blobProperties.getTimeToLiveInSeconds() + TimeUnit.MILLISECONDS.toSeconds(
+          blobProperties.getCreationTimeInMs()) > Integer.MAX_VALUE) {
+        logger.debug("TTL set to very large value in PUT request with BlobProperties {}", blobProperties);
+        frontendMetrics.ttlTooLargeError.inc();
+      } else if (container.isTtlRequired() && (blobProperties.getTimeToLiveInSeconds() == Utils.Infinite_Time
```
Arun-LinkedIn, line 239: Since TTL for S3 multipart requests would be set by us, may be we don't need these `if/else if` checks?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -80,13 +213,88 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+        } else {
+          logger.debug("{} attempted an upload with ttl {} to {}", blobProperties.getServiceId(),
+              blobProperties.getTimeToLiveInSeconds(), descriptor);
+          frontendMetrics.ttlNotCompliantError.inc();
+          restResponseChannel.setHeader(RestUtils.Headers.NON_COMPLIANCE_WARNING,
+              "TTL < " + frontendConfig.maxAcceptableTtlSecsIfTtlRequired + " will be required for future uploads");
+        }
+      }
+      // inject encryption frontendMetrics if applicable
+      if (blobProperties.isEncrypted()) {
```
Arun-LinkedIn, line 255: May be this check is also not required?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -80,13 +213,88 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
+     * @param blobInfoFromRequest the {@link BlobInfo} parsed from the request.
+     * @return a {@link BlobProperties} for a TTL-ed initial router call.
+     */
+    BlobProperties getPropertiesForRouterUpload(BlobInfo blobInfoFromRequest) {
+      BlobProperties properties;
+      if (blobInfoFromRequest.getBlobProperties().getTimeToLiveInSeconds() == Utils.Infinite_Time) {
+        properties = new BlobProperties(blobInfoFromRequest.getBlobProperties());
+        // For blob with infinite time, the procedure is putBlob with a TTL, record insert to database with
+        // infinite TTL, and ttlUpdate.
+        properties.setTimeToLiveInSeconds(frontendConfig.permanentNamedBlobInitialPutTtl);
```
Arun-LinkedIn, line 293: Just wondering if this would also not be needed since part requests may always be temporary? But not sure though

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: Removed. Thanks,

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: You are right. Removed.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: Removed it as well.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: Removed all the Ttl related code. Thanks

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: Leave it for now.
When we do the S3 blob life cycle management, we can revisit it.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
weisong44: In this case, perhaps just set it to OK without checking?

weisong44 (approved):
LGTM

Arun-LinkedIn (approved):
lgtm..

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
JingQianCloud: I hope it's ok I leave it as is just in case there will other status than the OK or Created.

PR #2721, opened 2024-02-27
[vcr-2.0] Fix multiple blob-upload errors to Azure

snalli:
This patch addresses the error blobUploadConflictCount, occurring when attempting to upload a blob to Azure that already exists there.

Initially, we suspected a race condition due to concurrent threads accessing the same blob from different replicas and attempting simultaneous uploads. However, investigations disproved this idea. Even when assigning all replicas of a partition to a single thread, the error continued. The root cause was identified as the replication thread iterating over all replicas of a partition within the same loop iteration, leading to redundant upload attempts.

The solution involves segregating replicas R1, R2, and R3 into distinct iterations of the replication loop. As the loop is continuous, each replica gets its turn. Consequently, in the first iteration, R1 is processed, followed by R2, and then R3. This approach ensures that missing blob B is uploaded from R1 to Azure during the first iteration. Subsequent iterations involving R2 and R3 skip the fetch and upload step for blob B, as it is already present.

This strategy not only eliminates the error at its source but also reduces server-to-server and server-to-Azure traffic. Previous attempts to address the issue through caching were only partially effective, allowing some redundant requests to pass through due to cache eviction and presenting correctness issues. With this patch, however, the error is completely and permanently resolved.

Before:

i-th iteration: 

VCR -> {R1, R2, S1, S2}

After:

i-th iteration:       VCR -> {R1, S1}
i+1-st iteration:   VCR -> {R2, S2}
i+2-nd iteration:  VCR -> {R1, S1}
i+3-rd iteration:  VCR -> {R2, S2}
.. so on


ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -183,6 +186,28 @@ protected ReplicaThread getReplicaThread(String threadName, FindTokenHelper find
...
+  protected void addRemoteReplicaInfoToReplicaThread(List<RemoteReplicaInfo> remoteReplicaInfos, boolean startThread) {
+    for (RemoteReplicaInfo rinfo : remoteReplicaInfos) {
+      String dc = rinfo.getReplicaId().getDataNodeId().getDatacenterName();
+      List<ReplicaThread> replicaThreads = getOrCreateThreadPoolIfNecessary(dc, startThread);
+      if (replicaThreads != null) {
+        ReplicaThread rthread = partitionToReplicaThread.computeIfAbsent(
+            rinfo.getLocalReplicaId().getPartitionId().getId(),
+            key -> replicaThreads.get(getReplicaThreadIndexToUse(dc)));
+        rthread.addRemoteReplicaInfo(rinfo);
+        rinfo.setReplicaThread(rthread);
```
Arun-LinkedIn, line 204: This may cause each thread to talk to more Ambry servers? For example, if ReplicaThread1 is responsible for P1, P2, P3, it could talk to 9 hosts (in worst case) in each replication cycle. Just wondering if it could cause increase in replication lag. I think tests may help to know if there are any side effects.

Arun-LinkedIn (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
snalli: hmm good point, lemme think abt it

snalli:
Closing this as the issue is not due to concurrent updates. It's because Azure is busy and we fail to check if the blob exists. We assume the blob does not exist and try to upload it at which point Azure says blob-exists. This is not harmful

snalli:
Will reopen when ready

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
```diff
@@ -58,6 +65,17 @@ public class VcrReplicaThread extends ReplicaThread {
   protected AzureCloudConfig azureCloudConfig;
   protected VerifiableProperties properties;
   protected CloudDestination cloudDestination;
+  protected HashMap<String, List<Pair<RemoteReplicaInfo, MessageInfo>>> blobToReplicas;
```
justinlin-linkedin, line 68: It seems there is no code cleaning up these two maps so these two maps would probably blow up the entire memory.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
```diff
@@ -76,6 +94,88 @@ public VcrReplicaThread(String threadName, FindTokenHelper findTokenHelper, Clus
...
+      if (oldMetadata.serverErrorCode != ServerErrorCode.No_Error) {
+        newMetadata = oldMetadata;
+      } else if (replicaToMessages.containsKey(replica)) {
+        newMetadata = new ExchangeMetadataResponse(oldMetadata, replicaToMessages.get(replica));
+      } else {
+        newMetadata = new ExchangeMetadataResponse(oldMetadata, new HashSet<>());
+      }
+      newMetadataResponseList.add(newMetadata);
+    }
+
```
justinlin-linkedin, line 168: I think this method needs more explanation and comments. Using a blob map and replica map to exchange information between multiple RemoteReplicaGroup is fine, but it needs some synchronization to clean up the maps so we don't keep all blob ids in memory.

We can add a preReplicationCycle hook to ReplicaThread and override it here to clear the blob and replica map.

justinlin-linkedin (changes requested):

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
snalli: A single thread already talks to multiple nodes today. There are more number of nodes than there are threads.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
```diff
@@ -76,6 +81,44 @@ public VcrReplicaThread(String threadName, FindTokenHelper findTokenHelper, Clus
...
+      String d1 = r1.getReplicaId().getDataNodeId().getHostname();
+      String d2 = r2.getReplicaId().getDataNodeId().getHostname();
+      return d1.compareTo(d2);
+    }
+  }
+
+  /**
+   * A custom filter that inheritors can override depending upon use case to pick which replicas to replicate from
+   * @param replicas A map of replicas {host -> {replicas}}
+   */
```
Arun-LinkedIn, line 99: Minor: Could you please add comments on what this method is doing?

Arun-LinkedIn (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
```diff
@@ -76,6 +81,44 @@ public VcrReplicaThread(String threadName, FindTokenHelper findTokenHelper, Clus
...
+    Map<DataNodeId, List<RemoteReplicaInfo>> nodes = new HashMap<>();
+    partitions.values().forEach(rlist -> {
+      rlist.sort(new ReplicaComparator());
+      RemoteReplicaInfo replica = rlist.get(numReplIter % rlist.size());
+      DataNodeId dnode = replica.getReplicaId().getDataNodeId();
+      List alist = nodes.getOrDefault(dnode, new ArrayList<>());
+      alist.add(replica);
+      nodes.putIfAbsent(dnode, alist);
+    });
+    numReplIter = (numReplIter % Integer.MAX_VALUE) + 1; // Prevent integer overflow
```
justinlin-linkedin, line 120: we don't have to go as high as Integer.MAX_VALUE, just reset it to 0 when it gets to a huge value, like 100, or 1000.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
```diff
@@ -76,6 +81,44 @@ public VcrReplicaThread(String threadName, FindTokenHelper findTokenHelper, Clus
...
+   * A custom filter that inheritors can override depending upon use case to pick which replicas to replicate from
+   * @param replicas A map of replicas {host -> {replicas}}
+   */
+  @Override
+  public Map<DataNodeId, List<RemoteReplicaInfo>> customFilter(Map<DataNodeId, List<RemoteReplicaInfo>> replicas) {
+    HashMap<Long, List<RemoteReplicaInfo>> partitions = new HashMap<>();
+    // Group replicas by partition
+    replicas.values().forEach(rlist -> rlist.forEach(replica -> {
+      long pid = replica.getReplicaId().getPartitionId().getId();
+      List alist = partitions.getOrDefault(pid, new ArrayList<>());
```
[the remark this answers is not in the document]
snalli: `List` cannot be instantiated.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
```diff
@@ -76,6 +81,44 @@ public VcrReplicaThread(String threadName, FindTokenHelper findTokenHelper, Clus
...
+      partitions.putIfAbsent(pid, alist);
+    }));
+    // Pick one replica per partition in this iteration
+    Map<DataNodeId, List<RemoteReplicaInfo>> nodes = new HashMap<>();
+    partitions.values().forEach(rlist -> {
+      rlist.sort(new ReplicaComparator());
+      RemoteReplicaInfo replica = rlist.get(numReplIter % rlist.size());
+      DataNodeId dnode = replica.getReplicaId().getDataNodeId();
+      List alist = nodes.getOrDefault(dnode, new ArrayList<>());
+      alist.add(replica);
```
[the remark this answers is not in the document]
snalli: `List` cannot be instantiated.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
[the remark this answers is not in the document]
snalli: The code here is trying to group replicas by partition, so partition -> {r1, r2, r3}.
`alist` is not created each time we enter the loop, but only once when partition is not in the map.

When partition p is not in the map:
1. create new alist
2. add r1 to alist
3. insert p -> {r1}

When partition p is already in the map:
1. get existing alist
2. append r2 to alist
3. put nothing since p already exists

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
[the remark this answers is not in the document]
justinlin-linkedin: I mean we can do something like this 
```
nodes.computeIfAbsent(dnode, k -> new ArrayList<>()).add(replica);
```
This can replace 
```
List alist = nodes.getOrDefault(dnode, new ArrayList<>());
alist.add(replica);
nodes.putIfAbsent(dnode, alist);
```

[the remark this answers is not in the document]
snalli: Got it, i didn't know computeIfAbsent returns the entry value.

justinlin-linkedin (approved):
LGTM

PR #2722, opened 2024-03-01
[S3 API] Consolidated all S3 payload POJO into a single class

weisong44:
As per subject, consolidated all S3 payload POJO classes into class S3MessagePayload.


JingQianCloud (approved):
LGTM. Thanks,

PR #2723, opened 2024-03-02
Don't reserve metadata chunk for S3 regular or part upload.

JingQianCloud:
For S3 commands, we doesn't use reserved metadata chunk.
1. For multipart upload, S3 doesn't use signed url. And for the data chunk, we don't reserve metadata chunk.
    Customer may upload part by part for days.
2. For regular update, it's not chunked upload. It's uploaded with single PutObject.
    Don't reserve metadata chunk either. We manage the S3 life cycle right away.
3. For multipart Complete request, it uses stitch command. Avoid the metadata chunk exceptions.


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -2116,7 +2124,8 @@ private void finalizeMetadataChunk() {
             .stream()
             .map(p -> new Pair<>(p.getFirst().toString(), p.getSecond()))
             .collect(Collectors.toList()));
-        PutBlobMetaInfo putBlobMetaInfoObj = new PutBlobMetaInfo(orderedChunkList, reservedMetadataChunkId.getID());
+        // S3 Part Upload, we don't reserve metadata chunk id for each data chunk. Refer to resolveReservedMetadataId.
+        PutBlobMetaInfo putBlobMetaInfoObj = new PutBlobMetaInfo(orderedChunkList, null);
```
Arun-LinkedIn, line 2128: Would we have to ignore the reserved metadata Id at the time of stitching too? https://github.com/linkedin/ambry/blob/0b197de3b1b2db94e71afd261534c327702f8b26/ambry-router/src/main/java/com/github/ambry/router/PutOperation.java#L1944

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1347,6 +1347,14 @@ private String resolveReservedMetadataId() {
...
         return null;
       }
+      // S3 doesn't use reserved metadata chunk.
+      // 1. For multipart upload, S3 doesn't use signed url. And for the data chunk, we don't reserve metadata chunk.
+      //    Customer may upload part by part for days.
+      // 2. For regular update, it's not chunked upload. It's uploaded with single PutObject.
+      //    Don't reserve metadata chunk either. We manage the S3 life cycle right away.
+      if (RestUtils.isS3Request(restRequest)) {
+        return null;
+      }
```
justinlin-linkedin, line 1357: let's use PutBlobOptions.skipCompositeChunk() here. Checking if the request is s3 is tying the feature too much with the s3.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
JingQianCloud: Added handling for the stitch command.

justinlin-linkedin (approved):
LGTM

Arun-LinkedIn (approved):

PR #2725, opened 2024-03-05
[vcr-2.0] Fix DateFormat concurrency error

snalli:
SimpleDateFormat object is not thread-safe. Multiple replication threads were using the same instance causing exceptions. These exceptions were caught by replication logic which then tried to update a non-existing metric causing null pointer exc. I am not sure why the metric is absent but that's not important right now.


justinlin-linkedin (approved):
LGTM

PR #2726, opened 2024-03-06
Add partial support for listing parts of Multipart Upload

Arun-LinkedIn:
This PR adds support for listing parts for a completed multipart upload. For now, the API returns empty list. Flink customer only uses it before starting a new multipart upload. So, sending an empty list unblocks them. In future, we may need to fetch the actual part IDs.

Reference API: https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListParts.html

Also renamed `S3CreateMultipartUploadHandler` to `S3MultipartCreateUploadHandler` and `S3CompleteMultipartUploadHandler` to `S3MultipartCompleteUploadHandler`

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
@@ -229,13 +229,13 @@ class FrontendRestRequestService implements RestRequestService {
         accountAndContainerInjector);
     s3DeleteHandler = new S3DeleteHandler(deleteBlobHandler, frontendMetrics);
     s3HeadHandler = new S3HeadHandler(headBlobHandler, securityService, frontendMetrics, accountService);
-    s3ListHandler = new S3ListHandler(namedBlobListHandler, frontendMetrics);
     s3HeadHandler = new S3HeadHandler(headBlobHandler, securityService, frontendMetrics, accountService);
     s3MultipartUploadHandler =
         new S3MultipartUploadHandler(securityService, frontendMetrics, accountAndContainerInjector, frontendConfig,
             namedBlobDb, idConverter, router, quotaManager);
     s3PostHandler = new S3PostHandler(s3MultipartUploadHandler);
     s3PutHandler = new S3PutHandler(namedBlobPutHandler, s3MultipartUploadHandler, frontendMetrics);
+    s3ListHandler = new S3ListHandler(namedBlobListHandler, frontendMetrics, s3MultipartUploadHandler);
     namedBlobsCleanupRunner = new NamedBlobsCleanupRunner(router, namedBlobDb);
     if (frontendConfig.enableNamedBlobCleanupTask) {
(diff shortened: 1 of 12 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -315,15 +315,14 @@ public void handleGet(final RestRequest restRequest, final RestResponseChannel r
       } else if (requestPath.matchesOperation(Operations.STATS_REPORT)) {
         getStatsReportHandler.handle(restRequest, restResponseChannel,
             (result, exception) -> submitResponse(restRequest, restResponseChannel, result, exception));
+      } else if (S3ListHandler.isListBucketRequest(restRequest) || S3MultipartUploadHandler.isListPartRequest(
+          restRequest)) {
+        s3ListHandler.handle(restRequest, restResponseChannel,
+            (result, exception) -> submitResponse(restRequest, restResponseChannel, result, exception));
```
justinlin-linkedin, line 321: I suggest to have a method like `isS3Request` to just check if the request is an s3 request regardless of sub types. In this case, we can have logic like this
```
else if (S3BaseHandler.isS3Request(restRequest)) {
    if (isListBucketRequest(restRequest)) {
    } else if (isMultipartListRequest(restRequest)) {
    } else {
        // throw error
    }
}
```

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -229,13 +229,13 @@ public void start() throws InstantiationException {
...
     s3DeleteHandler = new S3DeleteHandler(deleteBlobHandler, frontendMetrics);
     s3HeadHandler = new S3HeadHandler(headBlobHandler, securityService, frontendMetrics, accountService);
-    s3ListHandler = new S3ListHandler(namedBlobListHandler, frontendMetrics);
     s3HeadHandler = new S3HeadHandler(headBlobHandler, securityService, frontendMetrics, accountService);
     s3MultipartUploadHandler =
         new S3MultipartUploadHandler(securityService, frontendMetrics, accountAndContainerInjector, frontendConfig,
             namedBlobDb, idConverter, router, quotaManager);
     s3PostHandler = new S3PostHandler(s3MultipartUploadHandler);
     s3PutHandler = new S3PutHandler(namedBlobPutHandler, s3MultipartUploadHandler, frontendMetrics);
+    s3ListHandler = new S3ListHandler(namedBlobListHandler, frontendMetrics, s3MultipartUploadHandler);
```
justinlin-linkedin, line 238: why do we have to merge list bucket and list multipart handlers in the same one. we can just rename S3ListHandler to S3ListBucketHandler and call  S3MultipartHandler.handle to handle the list multipart requests.

justinlin-linkedin (approved):

JingQianCloud (approved):

PR #2728, opened 2024-03-06
[S3 API] S3 multipart blob lifecycle

weisong44:
* Setting parts of multipart upload to be temporary with an expiration time
* Setting final blob to be permanent with infinite expiration time

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
@@ -14,7 +14,6 @@
  */
 package com.github.ambry.frontend.s3;
 
-import com.github.ambry.account.Container;
 import com.github.ambry.commons.Callback;
 import com.github.ambry.config.FrontendConfig;
 import com.github.ambry.frontend.AccountAndContainerInjector;
@@ -29,7 +28,6 @@ import com.github.ambry.quota.QuotaUtils;
 import com.github.ambry.rest.ResponseStatus;
 import com.github.ambry.rest.RestRequest;
 import com.github.ambry.rest.RestResponseChannel;
-import com.github.ambry.rest.RestServiceErrorCode;
 import com.github.ambry.rest.RestServiceException;
 import com.github.ambry.rest.RestUtils;
 import com.github.ambry.router.PutBlobOptions;
@@ -39,7 +37,6 @@ import com.github.ambry.router.Router;
 import com.github.ambry.utils.Utils;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.TimeUnit;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -104,10 +101,10 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
     String uploadId = RestUtils.getHeader(restRequest.getArgs(), UPLOAD_ID_QUERY_PARAM, true);
     restRequest.setArg(S3_CHUNK_UPLOAD, true);
     restRequest.setArg(SESSION, uploadId);
-    // TODO [S3] : set part ttl value
```
JingQianCloud, line 107: @weisong44 @asinghai 

I'm thinking, 
1. For part upload, we always use 29 days or 30 days ttl. We don't let user to control it. So if the part upload fails or user doesn't complete the multipart upload, these blobs will expire in one month.
2. In the CompleteMultiPart callchain, we set ttl to permanent.

We may have some discussion about it.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
```diff
@@ -104,10 +101,10 @@ void handle(RestRequest restRequest, RestResponseChannel restResponseChannel,
...
-    // restRequest.setArg(TTL, xxx);
-
     restRequest.setArg(SEND_FAILURE_REASON, Boolean.TRUE);
+    if (Utils.Infinite_Time == RestUtils.getTtlFromRequestHeader(restRequest.getArgs())) {
+      restRequest.setArg(TTL, frontendConfig.permanentNamedBlobInitialPutTtl);
```
Arun-LinkedIn, line 106: I am wondering if `frontendConfig.chunkUploadInitialChunkTtlSecs` or `frontendConfig.chunkUploadMaxChunkTtlSecs` may be more appropriate configs? These are the default and max TTL we allow for chunks in Ambry multipart uploads.

Also, may be we don't need to check `RestUtils.getTtlFromRequestHeader(restRequest.getArgs())` since S3 requests won't have the `x-ambry-ttl` http header?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
weisong44: Thanks! Based on our offline discussion, I will make the following changes
1. Use `chunkUploadInitialChunkTtlSecs` for upload
2. Use `permanentNamedBlobInitialPutTtl` during stitching

Arun-LinkedIn (approved):

JingQianCloud (approved):
LGTM

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartUploadPartHandler.java
weisong44: Agreed and fixed

PR #2731, opened 2024-03-09
Add buffered directio to compaction

justinlin-linkedin:
## Summary

Add a buffered implemetation of direct io for writing to log segment in compaction.

## Details
We are add a configuration to switch between a buffered implementation and a non-buffered implementation. The code looks a bit weird right now as we have to compatible with the existing logic and use configuration to switch between the new logic and the existing one. After we verify the new logic (buffered direct io) would improve the disk performance, we have to refactor the code for io operation.

## Test
Unit test

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
@@ -543,6 +543,7 @@ class LogSegment implements Read, Write {
    */
   void flush() throws IOException {
     fileChannel.force(true);
+    directIOExecutor.flush();
   }
 
   /**
@@ -554,10 +555,12 @@ class LogSegment implements Read, Write {
     if (open.compareAndSet(true, false)) {
       if (!skipDiskFlush) {
         flush();
+        directIOExecutor.flush();
       }
       try {
         // attempt to close file descriptors even when there is a disk I/O error.
         fileChannel.close();
+        directIOExecutor.close();
       } catch (IOException e) {
         if (!skipDiskFlush) {
           throw e;
--- a/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
@@ -91,6 +91,7 @@ class LogSegment implements Read, Write {
     this.name = name;
     this.capacityInBytes = capacityInBytes;
     this.metrics = metrics;
+    this.directIOExecutor = createDirectIOExecutor(config.storeCompactionDirectIOBufferSize);
     try {
       fileChannel = Utils.openChannel(file, true);
       segmentView = new Pair<>(file, fileChannel);
@@ -104,7 +105,6 @@ class LogSegment implements Read, Write {
       if (config.storeSetFilePermissionEnabled) {
         Files.setPosixFilePermissions(this.file.toPath(), config.storeDataFilePermission);
       }
-      directIOExecutor = createDirectIOExecutor(config.storeCompactionDirectIOBufferSize);
     } catch (IOException e) {
       throw new StoreException("File not found while creating the log segment", e, StoreErrorCodes.File_Not_Found);
     }
--- a/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
@@ -151,13 +151,13 @@ class LogSegment implements Read, Write {
       this.name = name;
       this.metrics = metrics;
       fileChannel = Utils.openChannel(file, true);
+      directIOExecutor = createDirectIOExecutor(config.storeCompactionDirectIOBufferSize);
       segmentView = new Pair<>(file, fileChannel);
       // externals will set the correct value of end offset.
       endOffset = new AtomicLong(startOffset);
       if (config.storeSetFilePermissionEnabled) {
         Files.setPosixFilePermissions(this.file.toPath(), config.storeDataFilePermission);
       }
-      directIOExecutor = createDirectIOExecutor(config.storeCompactionDirectIOBufferSize);
     } catch (FileNotFoundException e) {
       throw new StoreException("File not found while creating log segment [" + file.getAbsolutePath() + "]", e,
           StoreErrorCodes.File_Not_Found);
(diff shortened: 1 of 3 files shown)


ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -32,6 +33,7 @@
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.zip.CRC32;
 import net.smacke.jaydio.DirectRandomAccessFile;
+import net.smacke.jaydio.align.DirectIoByteChannelAligner;
```
litingulfs, line 36: I think it's great that we're investigating using direct I/O 👍 

The https://github.com/smacke/jaydio appears to be abandoned, so do you have any thoughts on how we might fork the repo or otherwise maintain it ourselves in the future?

How will we measure whether or not this implementation is an improvement?

justinlin-linkedin: That's right, this library hasn't been updated in the last three years, but it should still be working.

It's hard to messure the disk io performance we have some many disk iops from all sources. One thing I can think so is the average write size and the number of write requests. We are using any buffer when performing direct io in compaction, and we have a lot of small records to write to the target log segment. After this PR, we should be able to see number of write requests goes down and the average of write size increases. 

This is a good thing for disk io performance as we don't want that many small writes after all.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -572,4 +593,100 @@ public String toString() {
...
+  }
+
+  /**
+   * A non-buffered implementation of {@link DirectIOExecutor}. It will open a direct io file every time an operation
+   * is executed.
+   */
+  private class DirectIOOneTimeExecutor implements DirectIOExecutor {
+    @Override
+    public <T> T execute(DirectIOOp<T> op) throws IOException {
+      try (DirectIoByteChannelAligner directIo = DirectIoByteChannelAligner.open(file, 2 * 1024 * 1024, false)) {
```
Arun-LinkedIn, line 677: Just a q.. does the jaydio library need 2 MB buffer for direct io as well?

Arun-LinkedIn (approved):

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
justinlin-linkedin: not really, I don't want to change any of the old code and this is what's in the old code.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -532,10 +551,12 @@ void close(boolean skipDiskFlush) throws IOException {
     if (open.compareAndSet(true, false)) {
       if (!skipDiskFlush) {
         flush();
+        directIOExecutor.flush();
```
JingQianCloud, line 554: Isn't this redundant? 
We have added the directIOExecutor.flush inside the flush() already.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -521,6 +539,7 @@ long getCapacityInBytes() {
    */
   void flush() throws IOException {
     fileChannel.force(true);
+    directIOExecutor.flush();
```
JingQianCloud, line 542: I'm not sure if it makes any difference.
Shall we can directIOExecutor.flush() before fileChannel.force()?

Same question for the close.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -572,4 +593,100 @@ public String toString() {
...
+    }
+
+    @Override
+    public <T> T execute(DirectIOOp<T> op) throws IOException {
+      // the direct io executes all the operation in the same thread so we don't need
+      // to care able the thread safety.
+      if (directIo == null) {
+        directIo = DirectIoByteChannelAligner.open(file, bufferSize, false);
+      }
+      return op.run(directIo);
```
JingQianCloud, line 652: Not necessary in this PR, if we have some metrics to show the how much times execute is called and how many time flush is called, we may see if it helps to accumulate the IO.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
justinlin-linkedin: You are right.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
justinlin-linkedin: I don't actually if it's necessary to even call both of them. I suppose it's fine to do so.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
justinlin-linkedin: It's probably not going to be helpful, since when the buffer is full, data will be flushed. But the flush operation is done in the directio library and we don't really know when that happens.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
JingQianCloud: DirectIOExecutor is based on fileChannel and utilize file channel. I think it's better to flush it before fileChannel to be safe.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -540,6 +559,7 @@ void close(boolean skipDiskFlush) throws IOException {
       try {
         // attempt to close file descriptors even when there is a disk I/O error.
         fileChannel.close();
+        directIOExecutor.close();
```
JingQianCloud, line 562: Can we close directIO before we close fileChannel?

JingQianCloud (commented):
Overall it looks good to me.
But I think we'd better be careful about the flush and close of the directIO and fileChannel.
DirectIO is using file channel. Like we start file channel first, we need stop directIO first.

JingQianCloud (approved):
After discussion, the directIO is not using this file channel. It has its own file channel.

PR #2732, opened 2024-03-10
Add two metrics to get more insight on compaction read

justinlin-linkedin:
## Summary
Adding two new metric to show
1. in compaction, how many bytes are we reading from src log segment every time
2. in those bytes, how many of them will be written to target log.

## Test
 no tests.


JingQianCloud (approved):
LGTM.

PR #2733, opened 2024-03-12
EOFException in log segment should be consider as format error

justinlin-linkedin:
## Summary
If we don't have enough bytes in log segment, it would throw an EOF exception. We should consider this exception as log format error.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
@@ -18,6 +18,7 @@ import com.github.ambry.utils.CrcInputStream;
 import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Utils;
 import java.io.DataInputStream;
+import java.io.EOFException;
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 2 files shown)


ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -154,6 +155,9 @@ class LogSegment implements Read, Write {
     } catch (FileNotFoundException e) {
       throw new StoreException("File not found while creating log segment [" + file.getAbsolutePath() + "]", e,
           StoreErrorCodes.File_Not_Found);
+    } catch (EOFException e) {
```
JingQianCloud, line 158: Good to me.
Currently even we don't catch it, it goes with the generic Exception. With generic exception, we also generate the same StoreException with Log_File_Format_Error as well, right?

JingQianCloud (approved):

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
justinlin-linkedin: Not really, EOFException is a subclass of IOException, it will be caught by the catch statement and then return Uknown_Error code.

JingQianCloud (approved):

Arun-LinkedIn (approved):

PR #2734, opened 2024-03-12
Fix file reads in the prefetched data

justinlin-linkedin:
## Summary
There is an error when reading bytes from log segment file to a ByteBuf in prefetched data. This PR fixes that error.

## Test
Unit test


ambry-utils/src/main/java/com/github/ambry/utils/Utils.java
```diff
@@ -612,6 +612,35 @@ public static int readFileToByteBuffer(FileChannel fileChannel, long offset, Byt
...
+    // If the writable bytes in ByteBuf is less then the given size, ByteBuf would expand its capacity.
+    // Put a check here to make sure that we don't expand the capacity.
+    if (buffer.writableBytes() < size) {
+      throw new IllegalArgumentException(
+          "ByteBuf doesn't have enough writable bytes [" + buffer.writableBytes() + "<" + size + "]");
+    }
+    int bytesRead = 0;
+    while (bytesRead < size) {
+      int readSize = buffer.writeBytes(fileChannel, offset + bytesRead, size - bytesRead);
+      if (readSize == -1 || readSize == 0) {
```
JingQianCloud, line 636: In which case will we hit the EOF?
I thought when we read, we checked the getMessageInfo().getSize() already.

justinlin-linkedin: This is a generic method that might deal with other usage. That's why we check EOF here.

JingQianCloud: Then I assume we don't hit issue. The code is just to be safe. Right?

JingQianCloud (approved):

Arun-LinkedIn (approved):
Lgtm! Just a q.. what is the bug with current code?

justinlin-linkedin:
> Lgtm! Just a q.. what is the bug with current code?

In some partition, reading a whole blob out would fail because file channel won't return all the required bytes at one read. And if we fail three times, this would mistakenly mark this partition unavailable.

PR #2739, opened 2024-03-16
Add S3MultipartETag class with version control

JingQianCloud:
Add S3MultipartETag class with version control

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/frontend/PutBlobMetaInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/PutBlobMetaInfo.java
@@ -101,7 +101,7 @@ public class PutBlobMetaInfo {
    * @param metaInfo the {@link PutBlobMetaInfo}
    * @return the JSON object
    */
-  public static ObjectNode serializeToJasonObject(PutBlobMetaInfo metaInfo) {
+  public static ObjectNode serializeToJsonObject(PutBlobMetaInfo metaInfo) {
     ObjectNode rootObject = objectMapper.createObjectNode();
     // The order of elements in JSON arrays is preserved.
     // https://www.rfc-editor.org/rfc/rfc7159.html
@@ -123,7 +123,7 @@ public class PutBlobMetaInfo {
    * @return the JSON string representation of the object
    */
   public static String serialize(PutBlobMetaInfo metaInfo) {
-    ObjectNode rootObject = serializeToJasonObject(metaInfo);
+    ObjectNode rootObject = serializeToJsonObject(metaInfo);
 
     return rootObject.toString();
   }
--- a/ambry-api/src/main/java/com/github/ambry/frontend/PutBlobMetaInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/PutBlobMetaInfo.java
@@ -12,18 +12,19 @@
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
  */
 
-package com.github.ambry.router;
+package com.github.ambry.frontend;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.github.ambry.utils.Pair;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import org.json.JSONArray;
-import org.json.JSONObject;
+import java.util.Objects;
 
 
 /**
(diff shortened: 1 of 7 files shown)


ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
```diff
@@ -132,8 +164,15 @@ public static PutBlobMetaInfo deserialize(String metaInfo) throws IOException {
       throw new IOException("Not expected JSON content " + metaInfo + e.getMessage());
     }
 
+    // version id
+    JsonNode jsonNode = rootNode.get(VERSION);
+    short version = jsonNode.shortValue();
+    if (version != VERSION_1) {
+      throw new IOException("Wrong version number " + metaInfo);
+    }
+
```
justinlin-linkedin, line 173: for json serialization and deserialization, why are we using both jackson and org.json at the same time? We only need one library to deal with JSON. I suppose we just use jackson and remove org.json.

JingQianCloud: Changed to Jackson. Thanks for the suggestion.

ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
```diff
@@ -101,17 +133,17 @@ public List<Pair<String, Long>> getOrderedChunkIdSizeList() {
    * @return the JSON string representation of the object
    */
   public static String serialize(PutBlobMetaInfo metaInfo) {
-    JSONObject rootObject = new JSONObject();
-
+    ObjectNode rootObject = objectMapper.createObjectNode();
```
justinlin-linkedin, line 136: can we just do 
```
return objectMapper.writeValueAsString(metaInfo);
```

ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
```diff
@@ -45,32 +46,63 @@
  *             "size": 104
  *         }
  *     ],
- *     "reservedMetadataChunkId": "AAYIAkJbXNUAAQAAAAAAAAACEEbJqS49Sz2YIfqoDrpHmw"
+ *     "version": 1
```
justinlin-linkedin, line 49: BTW, if we are adding versions to etag, then we should add version outside of the serialized PutBlobMetaInfo, so later if we decide not to use PutBlobMetaInfo as etag, we can use the version number to tell.

For example, the etag should be v1:{"blob": {}}, notice that v1 is outside of the serialized json string. Later if we decide to not use PutBlobMetaInof, we can do v2:xxxxxx

What do you think?

ambry-router/src/main/java/com/github/ambry/router/PutBlobMetaInfo.java
JingQianCloud: As we discussed, the pair key and value name needs to use meaningful naming instead of "first" and "second".

justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/frontend/PutBlobMetaInfo.java
```diff
@@ -96,24 +97,34 @@ public List<Pair<String, Long>> getOrderedChunkIdSizeList() {
...
-   * Serialize a {@link PutBlobMetaInfo} object to a JSON string
+   * Serialize a {@link PutBlobMetaInfo} object to a JSON Object
    * @param metaInfo the {@link PutBlobMetaInfo}
-   * @return the JSON string representation of the object
+   * @return the JSON object
    */
-  public static String serialize(PutBlobMetaInfo metaInfo) {
-    JSONObject rootObject = new JSONObject();
-
+  public static ObjectNode serializeToJasonObject(PutBlobMetaInfo metaInfo) {
```
Arun-LinkedIn, line 104: nit: Probably typo in method name for JasonObject?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -364,30 +363,28 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
 
-            ChunkInfo chunk = new ChunkInfo(blobId, chunkSize, expirationTimeInMs, reservedMetadataId);
-            chunkInfos.add(chunk);
+              ChunkInfo chunk = new ChunkInfo(blobId, chunkSize, expirationTimeInMs, reservedMetadataId);
+              chunkInfos.add(chunk);
+            }
+          } else {
+            String error = "Wrong ETag version " + completeMultipartUpload + " rest request " + restRequest;
+            LOGGER.error(error);
+            throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
```
Arun-LinkedIn, line 382: Would we need `if and else` condition here? When ETag version is wrong, `S3MultipartETag.deserialize()` throws exception and we might directly go to `catch` block.

Arun-LinkedIn (approved):
lgtm..

ambry-api/src/main/java/com/github/ambry/frontend/PutBlobMetaInfo.java
JingQianCloud: You are right.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
JingQianCloud: That's true. 
But with the else, I think it increases the readability. And if we have version 2 and S3MultipartETag can deserialize it,  we won't miss the case in the handler code.

PR #2741, opened 2024-03-17
[vcr-2.0] Introduce CPU scaling factor

snalli:
Not all nodes are identical; some possess more cores while others fewer.
Therefore, adopting a predetermined count of replication threads is not advisable.
Instead, the number of replication threads should be determined based on the number of cores available.
The scaling factor provided below can assume three potential states: negative, zero, or positive.
A negative value signifies the creation of a number of threads equal to the absolute value of the negative integer.
Zero denotes the absence of thread creation.
A positive value will serve as a multiplier for the number of CPUs to ascertain the number of threads to generate.


ambry-cloud/src/test/java/com/github/ambry/cloud/VcrReplicaThreadTest.java
```diff
@@ -33,23 +37,26 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static org.mockito.Mockito.*;
+
 
 public class VcrReplicaThreadTest {
-  protected final VerifiableProperties properties;
+  protected VerifiableProperties properties;
```
justinlin-linkedin, line 44: why removing `final`?

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
```diff
@@ -448,8 +448,32 @@ public class CloudConfig {
...
+   *  A cpu_scale of -2.5 will create 2 threads since int(|-2.5|) = 2
+   *  A cpu_scale of +2.5 will create 10 threads since int(2.5 * 4) = 10
+   */
+  public static final String BACKUP_NODE_CPU_SCALE = "backup.node.cpu.scale";
+  public static final float DEFAULT_BACKUP_NODE_CPU_SCALE = -1.0f; // 1 thread
+  // num_cpu-1 threads, prevent a fork bomb
+  public static final float MIN_BACKUP_NODE_CPU_SCALE = 1f - Float.valueOf(Runtime.getRuntime().availableProcessors());
+  public static final float MAX_BACKUP_NODE_CPU_SCALE = 4.0f; // 4 * num_cpus, prevent a fork bomb
+  @Config(BACKUP_NODE_CPU_SCALE)
+  public final float backupNodeCpuScale;
```
justinlin-linkedin, line 470: I recommend keeping this simple. If `a predetermined count of replication threads is not advisable`, then let's just rely on the scale factor and only permit the positive (and 0) values.

Arun-LinkedIn: Also, for replication, I think CPU is not the bottleneck since most of the time threads are waiting on response from network than doing busy work. 

We could still have config to scale up and down threads. Probably remove `CPU` references in config names since people may mistake it to be a CPU bottleneck?

snalli: @justinlin-linkedin Thanks. The idea behind a negative scale is to allow creating 1, 2 or a known number of threads for debugging and experiments. It does not have to be used and will not be used in production. You can still create 1 or 2 threads using the positive scale, but you'd have compute the fraction 1/num_cpus or 2/num_cpus and so on..

@Arun-LinkedIn cpu is not the bottleneck and this pr is not alleviate any cpu bottlenecks. network still remains the limiting factor, but only if ensure we over-subscribe the cpus. This pr removes the guess work of guessing the num threads. This pr future-proofs the code as well. If we get hosts with more cores, then the num threads automatically goes up without any config change. Conversely, it will also go down if we get smaller hosts. I don't think anyone would confuse cpu with bottleneck.

ambry-cloud/src/test/java/com/github/ambry/cloud/VcrReplicaThreadTest.java
snalli: good catch !

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -186,6 +186,30 @@ protected ReplicaThread getReplicaThread(String threadName, FindTokenHelper find
         replicaSyncUpManager, skipPredicate, leaderBasedReplicationAdmin, this.persistor, cloudDestination, properties);
   }
 
+  /**
+   * Returns the number of replication threads to create
+   * @return num replication threads
+   */
+  protected int getNumReplThreads(float cpu_scale) {
```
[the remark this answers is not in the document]
snalli: good catch !

ambry-api/src/main/java/com/github/ambry/config/CloudConfig.java
snalli: Also I am not really interested in the absolute number of threads. I want to know what is the throughput of if I subscribe the cluster as a whole to 2x, 3x or some X times the num avlb cpus.

justinlin-linkedin (approved):
LGTM

snalli:
Arun approved on slack

PR #2743, opened 2024-03-21
[DRAFT] Backup verification

snalli:


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -281,6 +281,20 @@ private List<RemoteReplicaInfo> createRemoteReplicaInfos(List<? extends ReplicaI
     return remoteReplicaInfos;
   }
 
+  public ReplicaThread getReplicationThread(String threadName) {
```
justinlin-linkedin, line 284: I don't see who is using this method.

ambry-api/src/main/java/com/github/ambry/config/ServerConfig.java
```diff
@@ -127,7 +127,18 @@ public class ServerConfig {
...
 
+  /**
+   * Server execution mode
+   * - Data recovery mode
+   * - Data verification mode
+   * - Data serving mode
+   */
+  public static final String SERVER_EXECUTION_MODE = "server.execution.mode";
+  @Config("server.execution.mode")
+  public final String serverExecutionMode;
```
justinlin-linkedin, line 138: There is a method in VerifiableProperty to getEnum from string. We can use Enum type here directly

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
```diff
@@ -263,16 +255,44 @@ public void startup() throws InstantiationException {
...
+        networkClientFactory =
+            new RecoveryNetworkClientFactory(properties, registry, clusterMap, storageManager, accountService);
+        recoveryManager =
+            new RecoveryManager(replicationConfig, clusterMapConfig, storeConfig, storageManager, storeKeyFactory,
+                clusterMap, scheduler, nodeId, networkClientFactory, registry, notificationSystem,
+                storeKeyConverterFactory, serverConfig.serverMessageTransformer, null, null);
+        recoveryManager.start();
+      } else if (serverConfig.serverExecutionMode.equals(ServerExecutionMode.DATA_VERIFICATION_MODE.toString())) {
+        // Backup integrity monitor here because vcr does not have code to store to disk and the code to create that is here
+        RecoveryManager recoveryManager =
```
justinlin-linkedin, line 267: Looks like we should try to shutdown this recovery manager when shutting down the server. There is already a class variable recoveryManager, why not just use it?

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
```diff
@@ -263,16 +255,44 @@ public void startup() throws InstantiationException {
...
+            new RecoveryManager(replicationConfig, clusterMapConfig, storeConfig, storageManager, storeKeyFactory,
+                clusterMap, null, nodeId,
+                new RecoveryNetworkClientFactory(properties, registry, clusterMap, storageManager, accountService),
+                registry, null, storeKeyConverterFactory, serverConfig.serverMessageTransformer,
+                null, null);
+        ReplicationManager serverReplicationManager =
```
justinlin-linkedin, line 273: same above.

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
```diff
@@ -263,16 +255,44 @@ public void startup() throws InstantiationException {
...
-        _recoveryManager =
-            new RecoveryManager(replicationConfig, clusterMapConfig, storeConfig, storageManager,
-                storeKeyFactory, clusterMap, scheduler, nodeId, networkClientFactory, registry, notificationSystem,
-                storeKeyConverterFactory, serverConfig.serverMessageTransformer, null,
-                clusterParticipants.get(0));
-        _recoveryManager.start();
-      } else {
+        networkClientFactory =
+            new RecoveryNetworkClientFactory(properties, registry, clusterMap, storageManager, accountService);
+        recoveryManager =
```
justinlin-linkedin, line 260: Recovery manager is not copying from VCR anymore, it copies data directly from cloud. It's time to clean up some RecoveryManager code, we still have some code handles VCR nodes etc..

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
```diff
@@ -263,16 +255,44 @@ public void startup() throws InstantiationException {
...
-            accountService);
-        vcrClusterSpectator = null; // Server does not talk to vcr during recovery
-        _recoveryManager =
-            new RecoveryManager(replicationConfig, clusterMapConfig, storeConfig, storageManager,
-                storeKeyFactory, clusterMap, scheduler, nodeId, networkClientFactory, registry, notificationSystem,
-                storeKeyConverterFactory, serverConfig.serverMessageTransformer, null,
-                clusterParticipants.get(0));
-        _recoveryManager.start();
-      } else {
+        networkClientFactory =
```
justinlin-linkedin, line 258: The comment above doesn't seem to be up to date. We don't have CloudToStorageManager no more.

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
```diff
@@ -263,16 +255,44 @@ public void startup() throws InstantiationException {
...
+        ReplicationManager serverReplicationManager =
+            new ReplicationManager(replicationConfig, clusterMapConfig, storeConfig, storageManager, storeKeyFactory,
+                clusterMap, null, nodeId,
+                new Http2NetworkClientFactory(new Http2ClientMetrics(registry), new Http2ClientConfig(properties), sslHttp2Factory, time),
+                registry, null, storeKeyConverterFactory, serverConfig.serverMessageTransformer,
+                null, null);
+        backupIntegrityMonitor = new BackupIntegrityMonitor(recoveryManager, serverReplicationManager,
+            (CompositeClusterManager) clusterMap, storageManager, nodeId, findTokenHelper, properties);
+        backupIntegrityMonitor.start();
+      } else if (serverConfig.serverExecutionMode.equals(ServerExecutionMode.DATA_SERVING_MODE.toString())) {
```
justinlin-linkedin, line 282: I also don't quite get some the verification works.

The first step should pick a random partition and use recovery manager to recover data from cloud. Now recovery mananger rely on the clustermap to know which partitions/replicas to recover. In order to make it work, we need to chang clustermap's method so it will return a random partition/replica. And since verification process runs on a different server host, we might not have any replicas in the clustermap for this server host. That indicates we really have to make some change in clustermap.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
snalli: spillover from next pr, will remove

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
snalli: good pt !

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
snalli: good pt ! removed the vcr-cluster-spectator.  i assume that's what u r referring to.

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
snalli: good pt ! used the existing one.

ambry-server/src/main/java/com/github/ambry/server/AmbryServer.java
snalli: I deal with some of these things in the next pr, since this one is big enough. 
But I agree with your points.

>> The first step should pick a random partition and use recovery manager to recover data from cloud.

Correct. I do this.

>> Now recovery mananger rely on the clustermap to know which partitions/replicas to recover. In order to make it work, we need to chang clustermap's method so it will return a random partition/replica.

Hmm I can see how you got here. RecoveryManager.start() will use a static clustermap to recover selected partitions from cloud. (Or it could be helix clustermap if that's avlb post-disaster, though I have not tested it with helix). 

>> And since verification process runs on a different server host, we might not have any replicas in the clustermap for this server host.

Correct, we do not have any replicas for this host. I work around this in two steps:
1/ Use a static clustermap to discover disks on the verification-host and assign the replica to one of the disks.
2/ Call RecoveryThread::replicate(), instead of RecoveryManager.start()

justinlin-linkedin: I am fine with the steps, my concern is where those steps are going to happen. 
1. Use a statick clustermap to pick a disk D on the verification-host? Say verification host is HV, will this host placed in the static clustermap? 
2. Replica R should replicate data via RecoveryThread, which means it copies data from cloud. Now do we wait for the copy to finish? or do we verify as the copy goes on?
3. We also created a replication manager to replicate from other peers. I suppose this is trying to get data from on-prem server so we can compare them with data copied from cloud. Where does this data live? Disk D stores data from cloud, do we not store those data on disk and just compare them in memory?

snalli: >> Use a statick clustermap to pick a disk D on the verification-host? Say verification host is HV, will this host placed in the static clustermap?

Yes, the static clustermap on the host records the hostname, port, disk mountpaths.

>> Replica R should replicate data via RecoveryThread, which means it copies data from cloud. Now do we wait for the copy to finish? or do we verify as the copy goes on?

We wait for recovery to finish. Azure signals us when there is no more data to recover and we can stop the loop.

>> We also created a replication manager to replicate from other peers. I suppose this is trying to get data from on-prem server so we can compare them with data copied from cloud. Where does this data live? Disk D stores data from cloud, do we not store those data on disk and just compare them in memory?

yes, `ReplicationManager` will fetch on-prem data from server hosting replica R. For this, I plan to only fetch metadata + CRC from server and compare it with metadata + CRC from cloud stored in disk D. I do not plan to store metadata + CRC from server on any disk in HV, but just compare and discard or print out mismatches in log/email. Fetching entire blob from server, I think, would be a waste of time and resources and a stress on the server esp when it is busy with customer traffic.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
snalli: actually, its used in the ctor for BackupIntegMonitor

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -0,0 +1,132 @@
...
+    clusterMapConfig = new ClusterMapConfig(properties);
+    compositeClusterManager = cluster;
+    executor = Utils.newScheduler(1, "ambry_backup_integrity_monitor_", true);
+    nodeId = node;
+    serverReplicationManager = server;
+    serverReplicator = server.getBackupCheckerThread("ambry_backup_integrity_monitor");
+    stop = new AtomicBoolean(false);
+    storageManager = storage;
+    tokenFactory = token;
+    // log disk state
```
snalli, line 72: @justinlin-linkedin added this, logging avlb disks on the node at startup, lmk if this is to be removed

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -0,0 +1,192 @@
...
+
+      /** Create local Store S */
+      BlobStore store = startLocalStore(partition);
+
+      /** Restore cloud backup C */
+      logger.info("[BackupIntegrityMonitor] Restoring backup partition-{} to disk [{}]", partition.getId(), store);
+      cloudReplica = azureReplicationManager.getCloudReplica(store.getReplicaId());
+      // TODO: Reload tokens
+      azureReplicator.addRemoteReplicaInfo(cloudReplica);
+      while (!((RecoveryToken) cloudReplica.getToken()).isEndOfPartition()) { azureReplicator.replicate(); }
```
justinlin-linkedin, line 177: please reformat this file.

justinlin-linkedin, line 177: is there anyway we can print out some message to show progress?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/CompositeClusterManager.java
```diff
@@ -34,48 +36,51 @@
...
   final StaticClusterManager staticClusterManager;
   final HelixClusterManager helixClusterManager;
   final HelixClusterManagerMetrics helixClusterManagerMetrics;
+  final boolean check;
+  public HelixClusterManager getHelixClusterManager() {
+    return helixClusterManager;
+  }
+  public StaticClusterManager getStaticClusterManager() {
+    return staticClusterManager;
+  }
```
justinlin-linkedin, line 50: reformat this file please.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/CompositeClusterManager.java
```diff
@@ -34,48 +36,51 @@
  * instance and uses the {@link StaticClusterManager} as the source-of-truth. It relays events to both and checks for
  * and reports inconsistencies in the views from the two underlying cluster managers.
  */
-class CompositeClusterManager implements ClusterMap {
+public class CompositeClusterManager implements ClusterMap {
```
justinlin-linkedin, line 39: it seems like you only need CompositeClusterManager to carry HelixClusterManager and StaticClusterMap to AmbryServer, you don't really care about if the CompositeClusterManager would compare the results from both ClusterMap. If this is the case, we might as well leave CompositeClusterManager intact and only expose the methods to return two clustermaps.

snalli: I started with that, exposing just the two clustermaps. But that was not enough.

Deep in the code, BlobId::L302 calls clusterMap.getPartitionIdFromStream. The current composite mgr defaults to returning the result from static-cluster-map and this does not fly on the verification node because it relies on helix-cluster-map to discover replicas and extract partitionId from blobIDs.  So I had to change CompositeClusterManager::getPartitionIdFromStream.

Next, HelixClusterManagerTest::basicInterfaceTest fails because it calls clusterManager.getWritablePartitionIds and clusterManager.getPartitionIdFromStream. The types of return values don't match. So then I had to change other methods.

Finally, I made the checking of val from static and helix maps optional because the result will always be false on the verification node. The check is not needed there.

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
snalli: yes done

justinlin-linkedin (approved):

PR #2744, opened 2024-03-22
MessageFormatException should be treated as Log format error

justinlin-linkedin:
## Summary
When recovering records from log to index, we would just skip malformed records. But this should be treated as Log format error so we can remove this replica's directory and start over.

## Test
Unit test


JingQianCloud (approved):
LGTM.

Arun-LinkedIn (approved):

PR #2749, opened 2024-03-29
Check crc for last index segment file before parsing its content

justinlin-linkedin:
## Summary

Checking CRC of the unsealed index segment file before parsing its content to get all the index entries. Checking the CRC before parsing would expose file corruption error. If CRC matches with the content of the file, then all the errors in the parsing, will be treated as logical error.

## Test
Unit test


Arun-LinkedIn (approved):
lgtm!

JingQianCloud (approved):

PR #2750, opened 2024-03-30
Restore disk's availability when starting up storage manager

justinlin-linkedin:
## Summary
When disk fails, we have a disk failure handler to set the availability of this disk to false and decrease instance capacity. After  disk is fixed, we need to manually restore the disk's availability and update instance capacity in helix. This can be done in code automatically. This PR is trying to do that.

There is a simple method to check if the disk is recovered, we just have to check the mount path of the disk and make sure it exists and is a directory. This is by no mean a prefect condition, but most of the time, it works.

After that, we would update the property store to mark those unavailable disks back to available and increase the instance capacity. This is a two-step operation and they are not atomic, so it's possible we only successfully finish first step and fail on second step. In order to recover it, we will alway try to update the instance capacity when starting up.

## Test
Unit test

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -268,7 +268,7 @@ public class StorageManager implements StoreManager {
     if (!storeConfig.storeRestoreUnavailableDiskInFullAuto || !clusterMap.isDataNodeInFullAutoMode(currentNode)) {
       return failedDiskIds;
     }
-    // We should try to retore the availability of failed disks
+    // We should try to restore the availability of failed disks
     if (!failedDiskIds.isEmpty()) {
       List<DiskId> disksToRecover = new ArrayList<>();
       for (DiskId diskId : failedDiskIds) {
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -1059,11 +1059,7 @@ public class StorageManager implements StoreManager {
       // messages, then when we reset the partitions, replica list in the StorageManager would be obsolete. Fortunately
       // all the replicas in the failed disks are already stopped, we just have to remove the disk from disk maps.
       failedDisks.addAll(newFailedDisks);
-      long healthyDiskCapacity = currentNode.getDiskIds()
-          .stream()
-          .filter(((Predicate<DiskId>) failedDisks::contains).negate())
-          .mapToLong(DiskId::getRawCapacityInBytes)
-          .sum();
+      int actualCapacityInGB = getCapacityOfHealthyDisks(currentNode.getDiskIds(), failedDisks);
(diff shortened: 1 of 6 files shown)


ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -259,6 +256,49 @@ void verifyDiskHealth() throws StoreException {
...
+   */
+  Set<DiskId> maybeRestoreDiskAvailability() {
+    List<DiskId> allDiskIds = currentNode.getDiskIds();
+    Set<DiskId> failedDiskIds = allDiskIds.stream()
+        .filter(diskId -> diskId.getState() == HardwareState.UNAVAILABLE)
+        .collect(Collectors.toSet());
+    if (!storeConfig.storeRestoreUnavailableDiskInFullAuto || !clusterMap.isDataNodeInFullAutoMode(currentNode)) {
+      return failedDiskIds;
+    }
+    // We should try to retore the availability of failed disks
```
Arun-LinkedIn, line 271: nit: typo for restore

Arun-LinkedIn (approved):
lgtm!

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -259,6 +256,49 @@ void verifyDiskHealth() throws StoreException {
...
+        .collect(Collectors.toSet());
+    if (!storeConfig.storeRestoreUnavailableDiskInFullAuto || !clusterMap.isDataNodeInFullAutoMode(currentNode)) {
+      return failedDiskIds;
+    }
+    // We should try to restore the availability of failed disks
+    if (!failedDiskIds.isEmpty()) {
+      List<DiskId> disksToRecover = new ArrayList<>();
+      for (DiskId diskId : failedDiskIds) {
+        // If the mount path exist and is a directory, it's likely disk is already recovered
+        File mount = new File(diskId.getMountPath());
```
JingQianCloud, line 276: You have some explanation in the PR title but I don't fully understand.
Let's say one partition is corrupted and we marked it as failure in the Helix. Do we remove the folder right away?
Can the mount point existing is just the old mount point we haven't removed yet? How do we know it's the new one?

justinlin-linkedin: This is not for one partition, this is for the disk. If a disk failed, all replicas on this disk would be moved to a different host/disk in DiskFailureHandler. And in DiskFailureHandler we would also mark this disk as "UNAVALABLE" in property store and decrease the instance disk capacity. 

Now if this disk is fixed, (or replaced), then we want to mark this disk as "AVAILABLE" in property store and increase the instance disk capacity when starting up storage manager so we don't have to manually do this every time.

But how do we know if the disk is fixed? One simple way is to check if the disk mount path exists or not. Because when a disk is broken, its mount path doesn't exist in the root filesystem. And now if it exists, it's a very strong evidence that this disk is good now.

This is not a perfect solution, but based on what I observed, it solves most of the time.

PR #2751, opened 2024-04-01
Added Base64 encoding for S3MultipartETag serialization

weisong44:
As per subject, added Base64 encoding for S3MultipartETag serialization, so that the internals aren't exposed externally.


JingQianCloud (approved):
LGTM.
Do you have any other response string we should encode?

PR #2756, opened 2024-04-05
[S3 API] Support S3 GetObjectLockConfiguration command

JingQianCloud:
1. Added DefaultCallbackChain the S3BaseHandler.
2. We return NOT_FOUND for S3 GetObjectLockConfiguration command.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
@@ -21,6 +21,7 @@ import com.github.ambry.rest.RestUtils;
 import java.util.Map;
 import java.util.Objects;
 
+import static com.github.ambry.rest.RestUtils.*;
 
 /**
  * Represents the blob url parsing results for named blob.
@@ -47,6 +48,9 @@ public class NamedBlobPath {
   public static NamedBlobPath parse(String path, Map<String, Object> args) throws RestServiceException {
     Objects.requireNonNull(path, "path should not be null");
     Objects.requireNonNull(args, "args should not be null");
+    if (RestUtils.isS3Request(args)) {
+      return parseS3(path, args);
+    }
     path = path.startsWith("/") ? path.substring(1) : path;
     String[] splitPath = path.split("/", 4);
     String blobNamePrefix = RestUtils.getHeader(args, PREFIX_PARAM, false);
@@ -73,6 +77,44 @@ public class NamedBlobPath {
     }
   }
 
+  /**
+   * Parse the input path for S3 request.
+   * @param path the URI path that needs to be parsed. This path should already be URL decoded and have query paramaters
+   *             removed. In the request handling path, this would be done by the {@link RestRequest} implementation.
+   * @param args a map containing any query parameters from the request.
+   * @return the {@link NamedBlobPath} that indicates the parsing result from blobUrl.
+   * @throws RestServiceException on parsing errors.
+   */
+  public static NamedBlobPath parseS3(String path, Map<String, Object> args) throws RestServiceException {
+    path = path.startsWith("/") ? path.substring(1) : path;
+    String[] splitPath = path.split("/", 4);
+    String blobNamePrefix = RestUtils.getHeader(args, PREFIX_PARAM, false);
+    boolean isGetObjectLockRequest = args.containsKey(OBJECT_LOCK_PARAM);
+    boolean isListRequest = blobNamePrefix != null;
+    int expectedSegments = (isListRequest || isGetObjectLockRequest) ? 3 : 4;
+    if (splitPath.length != expectedSegments || !Operations.NAMED_BLOB.equalsIgnoreCase(splitPath[0])) {
+      throw new RestServiceException(String.format(
+          "Path must have format '/named/<account_name>/<container_name>%s.  Received path='%s', arg='%s'",
+          isListRequest || isGetObjectLockRequest ? "" : "/<blob_name>'", path, args), RestServiceErrorCode.BadRequest);
+    }
+    String accountName = splitPath[1];
+    String containerName = splitPath[2];
+    if (isListRequest) {
+      String pageToken = RestUtils.getHeader(args, PAGE_PARAM, false);
+      return new NamedBlobPath(accountName, containerName, null, blobNamePrefix, pageToken);
+    }
+    if (isGetObjectLockRequest) {
+      return new NamedBlobPath(accountName, containerName, null, null, null);
+    } else {
+      String blobName = splitPath[3];
+      if (blobName.length() > MAX_BLOB_NAME_LENGTH) {
+        throw new RestServiceException(
+            String.format("Blob name maximum length should be less than %s", MAX_BLOB_NAME_LENGTH),
+            RestServiceErrorCode.BadRequest);
+      }
+      return new NamedBlobPath(accountName, containerName, blobName, null, null);
+    }
+  }
   /**
    * Parse the input path if it's a named blob request.
    * @param requestPath the {@link RequestPath} to be parsed.
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
@@ -18,8 +18,8 @@ import com.github.ambry.accountstats.AccountStatsStore;
 import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.commons.Callback;
 import com.github.ambry.config.FrontendConfig;
-import com.github.ambry.frontend.s3.S3BaseHandler;
 import com.github.ambry.frontend.s3.S3DeleteHandler;
+import com.github.ambry.frontend.s3.S3GetHandler;
 import com.github.ambry.frontend.s3.S3HeadHandler;
 import com.github.ambry.frontend.s3.S3ListHandler;
 import com.github.ambry.frontend.s3.S3MultipartUploadHandler;
@@ -115,6 +115,7 @@ class FrontendRestRequestService implements RestRequestService {
   private S3HeadHandler s3HeadHandler;
   private S3PostHandler s3PostHandler;
   private S3MultipartUploadHandler s3MultipartUploadHandler;
+  private S3GetHandler s3GetHandler;
   private QuotaManager quotaManager;
   private boolean isUp = false;
   private final Random random = new Random();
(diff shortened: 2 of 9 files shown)


ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
```diff
@@ -33,6 +33,8 @@ public class NamedBlobPath {
   private final String pageToken;
 
   static final String PREFIX_PARAM = "prefix";
+  static final String OBJECT_LOCK_PARAM = "object-lock";
```
weisong44, line 36: Can we just reuse the one defined in RestUtils?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
```diff
@@ -119,4 +140,58 @@ private ReadableStreamChannel serializeAsXml(RestRequest restRequest, Page<Named
     xmlMapper.writeValue(outputStream, result);
     return new ByteBufferReadableStreamChannel(ByteBuffer.wrap(outputStream.toByteArray()));
   }
+
+  private class CallbackChain {
```
weisong44, line 144: I suspect that this isn't an isolated case, I wonder whether it makes sense to make this generic to handle all features we don't yet support and share among all handlers?

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
```diff
@@ -119,4 +140,58 @@ private ReadableStreamChannel serializeAsXml(RestRequest restRequest, Page<Named
...
+    }
+
+    /**
+     * After {@link SecurityService#postProcessRequest} finishes, return response for the request.
+     * @return a {@link Callback} to be used with {@link SecurityService#postProcessRequest}.
+     */
+    private Callback<Void> securityPostProcessRequestCallback() {
+      return buildCallback(metrics.putSecurityPostProcessRequestMetrics, securityCheckResult -> {
+        // We don't support Bucket Object Configuration. So always return NOT_FOUND
+        throw new RestServiceException("Object Lock configuration is not set.", RestServiceErrorCode.NotFound);
```
weisong44, line 193: Question: will RestServiceErrorCode.NotFound translates to 404?

weisong44 (commented):
LGTM, left a few comments/questions.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
JingQianCloud: I also have the same thought. 
But the problem is that for HTTP request, we don't know the request type until the time the code is passing the parameters.
In the handleGet, we have the code as below, besides the multipart code, we have two condition,
1. the path has blob name, it goes to the "else". 
2. the path doesn't have the blob name, it goes to the s3ListHandler.

If we want to create a standalone s3OtherHandlers, we need a way to know what are the "other handles". 
Let me know if you have any suggestions.

```
      } else if (isS3Request(restRequest)) {
        if (S3BaseHandler.isListBucketRequest(restRequest)) {
          s3ListHandler.handle(restRequest, restResponseChannel,
        } else if (S3BaseHandler.isMultipartListPartRequest(restRequest)) {
          s3MultipartUploadHandler.handle(restRequest, restResponseChannel,
        } else {
          getBlobHandler.handle(requestPath, restRequest, restResponseChannel,
        }
      } else if (requestPath.matchesOperation(Operations.NAMED_BLOB)
```

weisong44: I didn't mean to create an other S3 handler, but just the CallbackChain class as a utility that can be shared, so that we don't have to repeat the same CallbackChain class repeatedly.

ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java
```diff
@@ -236,7 +236,8 @@ public void start() throws InstantiationException {
             namedBlobDb, idConverter, router, quotaManager);
     s3PostHandler = new S3PostHandler(s3MultipartUploadHandler);
     s3PutHandler = new S3PutHandler(namedBlobPutHandler, s3MultipartUploadHandler, frontendMetrics);
-    s3ListHandler = new S3ListHandler(namedBlobListHandler, frontendMetrics);
+    s3ListHandler =
+        new S3ListHandler(namedBlobListHandler, securityService, frontendMetrics, accountAndContainerInjector);
```
justinlin-linkedin, line 240: Honestly, I think we should create a new Handler, S3GetObjectLockConfigurationHandler. It would be so much simpler if we have another handler. We will call s3BaseHandler.isGetObjectLockConfigurationRequest in the handleGet method and call this new handler right away. This way, it doesn't mix with S3ListHandler.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
JingQianCloud: Just wrap the CallbackChain is a little bit weird to me.
I'm thinking creating one S3GetHandler and wrap all the s3 handlers together. Let me work on it first and discuss more after the change.

weisong44: Thanks!

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3GetHandler.java
```diff
@@ -0,0 +1,86 @@
...
+   * @param restRequest the {@link RestRequest} that contains the request parameters and body.
+   * @param restResponseChannel the {@link RestResponseChannel} where headers should be set.
+   * @param callback the {@link Callback} to invoke when the response is ready (or if there is an exception).
+   * @throws RestServiceException exception when the processing fails
+   */
+  @Override
+  protected void doHandle(RestRequest restRequest, RestResponseChannel restResponseChannel,
+      Callback<ReadableStreamChannel> callback) throws RestServiceException {
+    RequestPath requestPath = (RequestPath) restRequest.getArgs().get(REQUEST_PATH);
+    if (S3BaseHandler.isGetObjectLockConfigurationRequest(restRequest)) {
```
weisong44, line 74: Nit, you can remove "S3BaseHander."

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
```diff
@@ -135,4 +148,70 @@ public static boolean isMultipartListPartRequest(RestRequest restRequest) {
...
+    }
+
+    /**
+     * Start the chain by calling {@link SecurityService#processRequest}.
+     */
+    protected void start() {
+      try {
+        accountAndContainerInjector.injectAccountContainerForNamedBlob(restRequest, metrics.postBlobMetricsGroup);
+        securityService.processRequest(restRequest, securityProcessRequestCallback());
+      } catch (Exception e) {
```
weisong44, line 191: Catch clause shouldn't be needed as it's already handled in BaseS3Handler.handle()

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
```diff
@@ -135,4 +148,70 @@ public static boolean isMultipartListPartRequest(RestRequest restRequest) {
     return restRequest.getRestMethod() == RestMethod.GET && restRequest.getArgs().containsKey(S3_REQUEST)
         && restRequest.getArgs().containsKey(UPLOAD_ID_QUERY_PARAM);
   }
+
+  protected class DefaultCallbackChain {
```
weisong44, line 152: Can we use a different name indicating that it's erroring out.

weisong44 (approved):
LGTM, left a few comments ..

ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
```diff
@@ -73,6 +77,44 @@ public static NamedBlobPath parse(String path, Map<String, Object> args) throws
...
 
+  /**
+   * Parse the input path for S3 request.
+   * @param path the URI path that needs to be parsed. This path should already be URL decoded and have query paramaters
+   *             removed. In the request handling path, this would be done by the {@link RestRequest} implementation.
+   * @param args a map containing any query parameters from the request.
+   * @return the {@link NamedBlobPath} that indicates the parsing result from blobUrl.
+   * @throws RestServiceException on parsing errors.
+   */
+  public static NamedBlobPath parseS3(String path, Map<String, Object> args) throws RestServiceException {
```
justinlin-linkedin, line 88: Now that we created a new Handler to handle getObjectLockRequest, do we still need this method here? We won't deal with object-lock parameter in NamedBlobPath anymore.

justinlin-linkedin (approved):
Most code looks good to me, just one comment.

ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
JingQianCloud: We still need this parseS3.
We call parse in many places including accountAndContainerSanityCheck etc. 
We throw exception if we don't have 3 part for list and 4 part for getBlob.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3GetHandler.java
JingQianCloud: We still need S3BaseHandler, right? We have all the helper functions.
The S3Gethandler is for Get only.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
JingQianCloud: Some caller will call DefaultCallbackChain directly like below in S3GetHandler.java. It didn't go through BaseS3Handler.handle()

```
  protected void doHandle(RestRequest restRequest, RestResponseChannel restResponseChannel,
      Callback<ReadableStreamChannel> callback) throws RestServiceException {
    RequestPath requestPath = (RequestPath) restRequest.getArgs().get(REQUEST_PATH);
    if (S3BaseHandler.isGetObjectLockConfigurationRequest(restRequest)) {
      // We don't support Bucket Object Configuration. So always return NOT_FOUND
      new DefaultCallbackChain(restRequest, securityService, frontendMetrics, accountAndContainerInjector, callback,
          "Object Lock configuration is not set.", RestServiceErrorCode.NotFound).start();
    } else if (S3BaseHandler.isListObjectRequest(restRequest)) {
      ...
    }
```
  }

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3BaseHandler.java
JingQianCloud: Sure. Let me think about the better name for it. May change at the next PR.
For now, I'll merge it in since Sophie will start the list based on the PR.

PR #2759, opened 2024-04-09
Add a new metric to show the number of full range compaction

justinlin-linkedin:
## Summary
Adding a new metric to show how many full range compaction is going on each host. Full range compaction takes in all the inactive log segment files and run compaction logic on top of them. It's a very slow process and consumes a lot of disk iops. Having a metric to show the number of full range compaction will provide us with more insight to the cluster.

## Test
Unit test


ambry-store/src/main/java/com/github/ambry/store/CompactionDetails.java
```diff
@@ -112,7 +136,9 @@ byte[] toBytes() {
      */
 
     List<byte[]> segmentNameBytesList = new ArrayList<>();
-    int size = VERSION_SIZE + REFERENCE_TIME_SIZE + SEGMENT_COUNT_SIZE;
+    int size =
```
JingQianCloud, line 139: Can you also update the comment in "toBytes"

```
    /*
      Description of format
      version
      referenceTimeMs
      size of the logSegmentsUnderCompaction list
      segment_1_name_length segment_1_name
      segment_2_name_length segment_2_name
      ....
     */

```

JingQianCloud: Probably also:
```
  @Override
  public String toString() {
    return "(Delete Ref Time [" + referenceTimeMs + "] Segments to compact [" + logSegmentsUnderCompaction + "])";
  }
```

JingQianCloud (approved):
Overall looks good to me.
Added two comments in CompactionDetails.

PR #2760, opened 2024-04-10
Proactively examine blob store availability when there is disk io error

justinlin-linkedin:
## Summary
Proactively test storage's availability for all the blob stores on a disk when any blob store on the same disk encountered an io error. 

When disk fails, all blob stores would become unavailable due to io error. But this process might take a long time since it relies on the frontend (or replication) to trigger several io operations, like put or get. Without any io operation, a blob store won't know that the underlying disk has already failed. 

This is particularly an issue when we have lots of old partitions that barely getting any frontend requests and barely getting any replication requests. When one blob store fails due to disk io error, this blob store would be shut down and the replica would enter ERROR state, but the rest of replicas on the same disk would still remain as STANDBY or LEADER. In this case, disk failure handler won't be able to handle this disk failure. 

In this PR, we proactively issue read requests to all blob stores on this disk to trigger read io operations and then force replicas to enter ERROR state. 

## Test
Unit test

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/StoreConfig.java
@@ -633,6 +633,14 @@ public class StoreConfig {
   public static final String storeProactivelyTestStorageAvailabilityName =
       "store.proactively.test.storage.availability";
 
+  /**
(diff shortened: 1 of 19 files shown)


JingQianCloud (commented):
LGTM. Thanks, Justin.

JingQianCloud (approved):

PR #2761, opened 2024-04-12
[DR] Add replication between server-backup node in Backup Integrity Monitor 

snalli:
This patch adds replication between server and backup-node. The integ. monitor gets and compares metadata from server and compares it with on-disk metadata (which is a copy of cloud metadata). It then prints any mismatches to an output file.


ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -182,72 +228,63 @@ protected void advanceToken(RemoteReplicaInfo remoteReplicaInfo, ExchangeMetadat
...
+      changeMap.setAccessible(false);
+    } catch (IllegalAccessException | NoSuchFieldException e) {
+      logger.error(e.getMessage());
+      json = new JSONObject();
+    }
+    // Maintain alphabetical order of fields for string match
+    json.put("replica_token", rinfo.getToken().toString());
+    json.put("replicated_until", rinfo.getReplicatedUntilUTC());
+    // Pretty print with indent for easy viewing
+    fileManager.truncateAndWriteToFile(getFilePath(rinfo, REPLICA_STATUS_FILE), json.toString(4));
```
justinlin-linkedin, line 287: jackson's `ObjectMapper` implementation can respect any order you set for the object. Do you consider using jackson library here. 

1. We have two json libraries, in the code, jackson and org.json, this is annoying.
2. Using a reflection is always something to avoid if we have other ways.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -171,7 +171,12 @@ public class ReplicationMetrics {
   // This metric tracks cross colo get requests bytes rate for Standby replicas.
   public final Map<String, Meter> interColoReplicationFetchBytesRateForStandbyReplicas = new ConcurrentHashMap<>();
 
+  public static final String BACKUP_INTEGRITY_ERROR = "BackupIntegrityError";
+  public final Counter backupIntegrityError;
+
   public ReplicationMetrics(MetricRegistry registry, List<? extends ReplicaId> replicaIds) {
+    backupIntegrityError =
+        registry.counter(MetricRegistry.name(ReplicationMetrics.class, BACKUP_INTEGRITY_ERROR));
```
justinlin-linkedin, line 179: we usually just put the metric name in the register.meter.

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -172,34 +235,88 @@ public void run() {
...
         }
       }
-      logger.info("[BackupIntegrityMonitor] Restored backup partition-{} to disk [{}]", partition.getId(), store);
+      logger.info("[BackupIntegrityMonitor] Restored backup partition-{} to disk [{}], num_blobs = {}, num_bytes = {}",
+          partition.getId(), store, token.getNumBlobs(), token.getBytesRead());
 
-      /** TODO: Replicate from server and compare */
+      /** Create a temporary map of all keys recovered from cloud */
+      // Get all entries in one call, so keep the ceil buffer size high 1GB, all of it will not be used
+      FindInfo finfo = store.findEntriesSince(new StoreFindToken(), 1000 * (2 << 20),
```
justinlin-linkedin, line 256: this will not give you all the entries in one call. The size you passed to the `findEntriesSince` is the chunk data size. A partition can have up to 384 GB of worth of data. I suggest you do a while loop to get all the entries.

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -172,34 +235,88 @@ public void run() {
...
+        metrics.backupCheckerRuntimeError.inc();
+        logger.error("[BackupIntegrityMonitor] Mismatch, num_blobs from cloud = {}, num_blobs from disk = {}",
+            token.getNumBlobs(), azureBlobs.size());
+      }
+      serverScanner.setAzureBlobMap(azureBlobs);
 
-      logger.info("[BackupIntegrityMonitor] Verified backup partition-{}", partition.getId());
+      /** Replicate from server and compare */
+      List<AmbryReplica> replicas = partition.getReplicaIds();
+      AmbryReplica replica = replicas.get(random.nextInt(replicas.size()));
```
justinlin-linkedin, line 268: one filter to apply here to make sure the replica is not currently offline.

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -172,34 +235,88 @@ public void run() {
...
+      logger.info("[BackupIntegrityMonitor] Selected peer server replica [{}]", replica);
+      serverReplica = serverReplicationManager.createRemoteReplicaInfos(
+          Collections.singletonList(replica), store.getReplicaId()).get(0);
+      logger.info("[BackupIntegrityMonitor] Created peer server replica info [{}]", serverReplica);
+      serverScanner.addRemoteReplicaInfo(serverReplica);
+      logger.info("[BackupIntegrityMonitor] Queued peer server replica info [{}]", serverReplica);
+      long scanStartTime, scanEndTime = serverReplica.getReplicatedUntilUTC();
```
justinlin-linkedin, line 275: The epoch time is always UTC, there is no need to attach `UTC` in the name.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -143,29 +146,72 @@ protected void applyUndelete(MessageInfo remoteBlob, RemoteReplicaInfo remoteRep
...
+   * present: absent  => print to file, blob may not have been backed up or somehow missing in cloud backup
+   * present: present => check blob-state matches, if not then print to file
+   * @param response The {@link ReplicaMetadataResponse}.
+   * @param replicas
+   * @param server The remote {@link DataNodeId}.
+   * @return
+   */
+  List<ExchangeMetadataResponse> handleReplicaMetadataResponse(ReplicaMetadataResponse response,
+      List<RemoteReplicaInfo> replicas, DataNodeId server) {
+    IntStream.range(0, response.getReplicaMetadataResponseInfoList().size())
```
justinlin-linkedin, line 180: A for loop would be better than a IntStream.range.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerFileManager.java
```diff
@@ -70,8 +153,17 @@ protected boolean appendToFile(String filePath, RemoteReplicaInfo remoteReplicaI
    * @return True if append was successful, false otherwise
    */
   protected boolean truncateAndWriteToFile(String filePath, String text) {
-    // open-source impl
-    return true;
+    EnumSet<StandardOpenOption> options = EnumSet.of(StandardOpenOption.CREATE, StandardOpenOption.WRITE,
+        StandardOpenOption.TRUNCATE_EXISTING);
+    SeekableByteChannel seekableByteChannel = getFd(filePath, options);
+    try {
+      seekableByteChannel.truncate(0);
```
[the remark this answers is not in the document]
snalli: TRUNCATE_EXISTING truncates when opening the file for the first time. After that, the fd is cached and the flag has no effect. trucate(0) is to truncate before each write. 
I'll remove TRUNCATE_EXISTING.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerFileManager.java
```diff
@@ -70,8 +153,17 @@ protected boolean appendToFile(String filePath, RemoteReplicaInfo remoteReplicaI
...
+      seekableByteChannel.write(ByteBuffer.wrap(text.getBytes()));
```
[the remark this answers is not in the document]
snalli: merged `writeToFile` with `appendToFile`, had just one caller

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
snalli: sorry i don't get it, do you mean to change the class name of the metric ?

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
snalli: I see, how would I know if I've acquired a list of all keys in local store ?

snalli: or what if pass an insanely large size like 1000 TiB ? will it return all keys in one call ?

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -172,34 +236,98 @@ public void run() {
...
 
-      /** TODO: Replicate from server and compare */
+      /** Create a temporary map of all keys recovered from cloud */
+      StoreFindToken newDiskToken = new StoreFindToken(), oldDiskToken = null;
+      while (!newDiskToken.equals(oldDiskToken)) {
+        FindInfo finfo = store.findEntriesSince(newDiskToken, 1000 * (2 << 20),
+            null, null);
+        finfo.getMessageEntries().forEach(msg -> azureBlobs.put(msg.getStoreKey().getID(), msg));
+        oldDiskToken = newDiskToken;
+        newDiskToken = (StoreFindToken) finfo.getFindToken();
```
justinlin-linkedin, line 262: let's log out the token so if this doesn't work, we would know what happened.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
justinlin-linkedin: I mean you don't need to create a static final string variable for the metric name.

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: So we need the index i because we are iterating two lists - response-list and replica-list. So the for loop would iterate using indices like for (i=0; i < size; ++i) and IntStream is the same and a little cleaner, don't you think ?

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
snalli: done

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: done

ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
snalli: i looked at jackson, that is useful if i wanted to print the entire object as json to a file. But here i am only interested in two fields from two different objects, and i don't want to create a third object combining those two just for this purpose.

justinlin-linkedin (approved):

PR #2762, opened 2024-04-12
Support 100-continue PUT request

SophieGuo410:
This pr supported the Expect:100-continue feature. 
This header indicates that the client is asking the server to confirm if it is acceptable to send the request body before actually sending it. The server, upon receiving such a request with the "Expect: 100-continue" header, can respond with a "100 Continue" status code if it is ready to receive the request body. If not ready or if there's an issue, it can respond with an appropriate error code.


ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -96,6 +97,10 @@ public static final class Headers {
      * {@code "Date"}
      */
     public static final String DATE = "Date";
+    /**
+     * {@code "Expect"}
+     */
+    public static final String EXPECT = "Expect";
```
justinlin-linkedin, line 103: Please add some comments here that this is only supported for named blob Put

SophieGuo410: Sure, updated.

JingQianCloud (approved):
LGTM. Thanks for adding the support for "Continue".

PR #2763, opened 2024-04-17
Simplify exception generation when there are not enough replicas

litingulfs:
Make the code easier to understand by moving the generation of the exception to the place where it's used, and eliminate an extraneous variable.


ankagrawal (approved):
lgtm

PR #2767, opened 2024-04-23
DiskFailureHandler: check MM reason before exist MM

justinlin-linkedin:
## Summary
In disk failure handler, before exiting MaintenanceMode, we should check if the MM reason has been updated or not. If it has already been updated (overridden) by other application, we should not exit MM in disk failure handler.

## Reason for this change
When doing deployment, we don't want to back off if the cluster is in MM due to any other reason, so we would override the MM reason before the deployment. If we run into a race condition where
1. DiskFailureHandler Enter MM
2. Deployment Enter MM by overriding the reason
3. DiskFailureHandler Exit MM

In this case, DiskFailureHandler would exit MM and leave deployment without any protection from MM.

## Test
Unit test

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -14,6 +14,8 @@
 package com.github.ambry.clustermap;
 
 import com.codahale.metrics.MetricRegistry;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.ambry.accountstats.AccountStatsStore;
 import com.github.ambry.commons.Callback;
 import com.github.ambry.commons.CommonUtils;
@@ -39,9 +41,11 @@ import org.apache.helix.AccessOption;
 import org.apache.helix.HelixAdmin;
 import org.apache.helix.HelixManager;
 import org.apache.helix.InstanceType;
+import org.apache.helix.PropertyKey;
 import org.apache.helix.lock.LockScope;
 import org.apache.helix.lock.helix.HelixLockScope;
 import org.apache.helix.lock.helix.ZKDistributedNonblockingLock;
+import org.apache.helix.model.ControllerHistory;
 import org.apache.helix.model.InstanceConfig;
 import org.apache.helix.participant.StateMachineEngine;
 import org.apache.helix.store.HelixPropertyStore;
@@ -78,6 +82,7 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
   private volatile boolean inMaintenanceMode = false;
   private volatile String maintenanceModeReason = null;
   final Map<StateModelListenerType, PartitionStateChangeListener> partitionStateChangeListeners;
+  private static final ObjectMapper objectMapper = new ObjectMapper();
 
   private static final Logger logger = LoggerFactory.getLogger(HelixParticipant.class);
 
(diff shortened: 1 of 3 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -481,9 +486,16 @@ public boolean exitMaintenanceMode() {
...
+        MaintenanceRecord record = getLastMaintenanceRecord();
+        if (record.operationType.equals(MaintenanceRecord.OPERATION_TYPE_ENTER) && record.reason.equals(
+            maintenanceModeReason)) {
+          helixAdmin.manuallyEnableMaintenanceMode(clusterName, false, maintenanceModeReason, null);
+          logger.info("Cluster {} exits maintenance mode: {}", clusterName, maintenanceModeReason);
+        } else {
+          logger.error("Maintenance record is not expected, operation type: {}, reason: {}", record.operationType,
+              record.reason);
+        }
         inMaintenanceMode = false;
```
MikeDafi, line 498: I'm not sure how useful this `inMaintenanceMode variable is if we will be checking in realtime if we're in MM or not anyways

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -481,9 +486,16 @@ public boolean exitMaintenanceMode() {
         throw new IllegalStateException("Cluster is not in maintenance mode");
       }
       try {
-        helixAdmin.manuallyEnableMaintenanceMode(clusterName, false, maintenanceModeReason, null);
+        MaintenanceRecord record = getLastMaintenanceRecord();
+        if (record.operationType.equals(MaintenanceRecord.OPERATION_TYPE_ENTER) && record.reason.equals(
+            maintenanceModeReason)) {
+          helixAdmin.manuallyEnableMaintenanceMode(clusterName, false, maintenanceModeReason, null);
```
MikeDafi, line 492: nit: I think it's a little weird to use `.manuallyEnableMaintenanceMode` to disable MM. And I believe the maintenance mode reason isn't utilized when disabling the MM, right? Python docs: https://docs.google.com/document/d/13UVVnWHqY1qBd8KXSZnIQs4dIFLeviDKOa7wpZp7Mf8/edit#bookmark=id.xps9dnsnrjw

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
justinlin-linkedin: save one roundtrip to zookeeper to get the maintenance info.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -493,6 +505,46 @@ public boolean exitMaintenanceMode() {
...
+    public String operationType;
+    @JsonProperty("REASON")
+    public String reason;
+    @JsonProperty("TRIGGERED_BY")
+    public String triggeredBy;
+    @JsonProperty("USER")
+    public String user;
+
+    public static final String OPERATION_TYPE_EXIT = "EXIT";
+    public static final String OPERATION_TYPE_ENTER = "ENTER";
```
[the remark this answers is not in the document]
justinlin-linkedin: Just realized that Maintenance operation type can't be used to determine if a cluster is in MM or not. Enable/Disable MM takes two steps, first step is to send a maintenance signal, which would create a znode at /ClusterName/CONTROLLER/MAINTENANCE. The second step is to create an entry in HISTORY. 
Upon receive the maintenance signal, the cluster will be put in to MM. 

In this way, if the process, who sent the signal, dies after the signal is sent, then we won't be able to see the exit history.

Anyway, we don't even have to check the operation type in the last maintenance history entry. If the reason is different, we should not exit MM.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
justinlin-linkedin: The reason is put in the maintenance history even if this is a disable operation.  I agree it's a bit weird, but this is the only API for manipulating MM in java client unfortunately.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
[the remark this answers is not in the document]
JingQianCloud: In the above procedure, when do we write the ENTER history?

I saw two steps: 
write: /ClusterName/CONTROLLER/MAINTENANCE
write: EXIT HISTORY

MikeDafi:
Also, where is the `maintenancemodeReason` defined on the AmbryLI side, or will that be in a future change?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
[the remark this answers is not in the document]
justinlin-linkedin: Enter and Exit history are the same thing, it's just a record to insert to the maintenance history list. When we enter MM, helix would create  /ClusterName/CONTROLLER/MAINTENANCE and after that it will write an entry to /ClusterName/CONTROLLER/HISTORY.

justinlin-linkedin:
> Also, where is the `maintenancemodeReason` defined on the AmbryLI side, or will that be in a future change?

It's defined in the DiskFailureHandler, its format looks like this HOSTNAME_PORT_DISKFAILURE

JingQianCloud (approved):

MikeDafi (approved):

PR #2770, opened 2024-04-25
Always enable persistRemoteToken

JingQianCloud:
Always enable persistRemoteToken.
So it's served as a indicator that the BlobStore is alive.

Now caller has to provide the following two parameters when create BlobStore.
1. replicaID
2. scheduler.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
@@ -66,7 +66,7 @@ public class RemoteTokenTracker implements Closeable {
   public RemoteTokenTracker(ReplicaId localReplica, ScheduledExecutorService scheduler,
       StoreKeyFactory storeKeyFactory) {
     if (scheduler == null || localReplica == null) {
-      // TODO: throw exception
+      // TODO [TOMBSTONE]: throw exception
       logger.error("localReplica and schedule CANNOT BE NULL.");
     }
     this.localReplica = localReplica;
--- a/ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
@@ -63,11 +63,11 @@ public class RemoteTokenTracker implements Closeable {
 
   private static final Logger logger = LoggerFactory.getLogger(RemoteTokenTracker.class);
 
-  public RemoteTokenTracker(ReplicaId localReplica, ScheduledExecutorService scheduler,
-      StoreKeyFactory storeKeyFactory) {
+  public RemoteTokenTracker(ReplicaId localReplica, ScheduledExecutorService scheduler, StoreKeyFactory storeKeyFactory)
+      throws StoreException {
     if (scheduler == null || localReplica == null) {
-      // TODO [TOMBSTONE]: throw exception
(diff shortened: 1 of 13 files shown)


ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -65,6 +65,10 @@ public class RemoteTokenTracker implements Closeable {
 
   public RemoteTokenTracker(ReplicaId localReplica, ScheduledExecutorService scheduler,
       StoreKeyFactory storeKeyFactory) {
+    if (scheduler == null || localReplica == null) {
+      // TODO [TOMBSTONE]: throw exception
+      logger.error("localReplica and schedule CANNOT BE NULL.");
```
justinlin-linkedin, line 70: can we just throw an exception here.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
```diff
@@ -145,14 +149,15 @@ public void close() {
...
-      }
+    if (this.scheduler != null) {
+      persistorFuture =
+          this.scheduler.scheduleAtFixedRate(persistor, new Random().nextInt(Time.SecsPerMin), persistIntervalInSeconds,
+              TimeUnit.SECONDS);
+      logger.info("flush peer's remote token every {} seconds", persistIntervalInSeconds);
+    } else {
+      logger.error("scheduler is null, couldn't persist peer's remote token.");
+      // at least write once
+      persistor.run();
```
justinlin-linkedin, line 160: this is not good enough, persist only once at the beginning won't be a real solution .

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
JingQianCloud: I was thinking,
1. we start with error message and deploy it first.
2. After we monitor the cluster and there is no surprise, the following PR will always throw exception.

justinlin-linkedin: I don't think that's necessary, we have staging and canary.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
JingQianCloud: It should happen either.
Because we check it in RemoteTokenTracker that schedule won't be null. 
But yes, it's confusing, I will remove the else condition and throw the exception in the constructor at the same  PR.

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
JingQianCloud: Sure. Will change it.

JingQianCloud: Changed. Thanks,

ambry-store/src/main/java/com/github/ambry/store/RemoteTokenTracker.java
JingQianCloud: Removed the else condition.

justinlin-linkedin (approved):
LGTM, please fix the build failure.

PR #2773, opened 2024-04-29
Filter out all valid log semgents in full range compaction

justinlin-linkedin:
## Summary

In full range compaction (CompactAllPolicy), we would return all log segments what are not in journal to compactor, even if some of the log segments are having 100% of valid data. This is totally unnecessary, especially for leading and trailing log segments.  Since we would have to copy all of their data out to a new log segment file. 

### Add a configuration to `StoreConfig` to control this feature
Added a new configuration to `StoreConfig` to control this feature. This will be turn off by default and we need to change configuration to turn this on. This feature will be first turned on on only a few hosts and make sure it works, then propagate to other hosts
### Add a new interface `LogSegmentSizeProvider` to expose the data size of each log segment.
`BlobStoreStats` object has a reference to `PersistentIndex`, which knows the data size of each `LogSegment`. To best test the new logic, we are creating a new interface `LogSegmentSizeProvider` and let `PersistentIndex` implement this interface. 
In test, we have a map-based mock implementation.
## Test
Unit tests

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactAllPolicyFactory.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactAllPolicyFactory.java
@@ -128,7 +128,7 @@ class CompactAllPolicy implements CompactionPolicy {
       return null;
     }
 
-    for (int i = logSegmentsNotInJournal.size() - 1; i >= 0; i--) {
+    for (int i = logSegmentsNotInJournal.size() - 1; i >= allValidLogSegments.size(); i--) {
       if (!isLogSegmentDataAllValid.test(logSegmentsNotInJournal.get(i))) {
         break;
       }
--- a/ambry-store/src/main/java/com/github/ambry/store/CompactAllPolicyFactory.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/CompactAllPolicyFactory.java
@@ -16,9 +16,11 @@ package com.github.ambry.store;
 import com.github.ambry.config.StoreConfig;
 import com.github.ambry.utils.Pair;
 import com.github.ambry.utils.Time;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.NavigableMap;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Predicate;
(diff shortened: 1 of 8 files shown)


SophieGuo410 (approved):
lgtm!

ambry-store/src/main/java/com/github/ambry/store/CompactAllPolicyFactory.java
```diff
@@ -66,20 +68,74 @@ public CompactionDetails getCompactionDetails(long totalCapacity, long usedCapac
...
+      if (!isLogSegmentDataAllValid.test(segmentName)) {
+        break;
+      }
+    }
+    if (allValidLogSegments.size() == logSegmentsNotInJournal.size()) {
+      logger.info("All LogSegments have 100% of valid data, skip compaction");
+      return null;
+    }
+
+    for (int i = logSegmentsNotInJournal.size() - 1; i >= 0; i--) {
```
JingQianCloud, line 131: I think we are filtering the top consecutive 100% valid segments and bottom consecutive valid segments.
So here, shall we do 
for (int i = logSegmentsNotInJournal.size() - 1; i >= allValidLogSegments.size(); i--) {

justinlin-linkedin: if we are here, then it means there is a log segment that is not 100% valid in the middle of this list, which means we would break this for loop before reaching the log segment that is last element of `allValidLogSegments`.

JingQianCloud: I think both
 for (int i = logSegmentsNotInJournal.size() - 1; i >= 0; i--) {
and
 for (int i = logSegmentsNotInJournal.size() - 1; i >= allValidLogSegments.size(); i--) {
Will work.

The second one probably is easier to read. The while loop will break before allValidLogSegments.size.

justinlin-linkedin: updated

JingQianCloud (approved):

PR #2776, opened 2024-05-01
Filter out offline peers when checking delete tombstones

justinlin-linkedin:
## Summary

When a peer went offline for a long time (X days), we should not consider its replication token when checking if a delete tombstone is valid or not. 

## Reason 
The way we remove delete tombstone is fast, it keeps all the remote replication tokens from its peers, and compare offsets from tokens with the offset of the delete tombstone. This approach requires only local information so it's very fast, but it has some issues. When a peer went offline for a long time, it's remote replication token would not move at all. After a couple compactions, this peer's token offset would be less than most of delete tombstones' offsets. Thus all the delete tombstones can't compacted any more.

## Change
Filter those offline replicas out. Making sure that we enable the remote replication token persistor so we would persist remote tokens every few hours(or minutes). This would give us a good picture on when was the last time remote peer was active. 

If this active timestamp is longer than X days(configurable), we just ignore this peer.

## Test
Added unit test

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
@@ -1393,6 +1393,17 @@ class BlobStoreCompactor {
     }
     for (Map.Entry<String, Pair<Long, FindToken>> entry : remoteTokenTracker.getPeerReplicaAndToken().entrySet()) {
       Pair<Long, FindToken> pair = entry.getValue();
+      // Ignore this peer if the last active replication's timestamp of this peer is longer than the pre-configured days.
+      if (config.storeCompactionIgnorePeersUnavailableForDays != 0
+          && pair.getFirst() + TimeUnit.DAYS.toMillis(config.storeCompactionIgnorePeersUnavailableForDays)
+          < time.milliseconds()) {
+        if (logger.isTraceEnabled()) {
+          logger.trace(
+              "Store {}: dealing with delete index entry: key {}, offset {}, peer {}'s last active time is {}, ignore it",
+              dataDir, deleteIndexEntry.getKey(), deleteIndexEntry.getValue().getOffset(), pair.getFirst());
+        }
+        continue;
+      }
       FindToken token = srcIndex.resetTokenIfRequired((StoreFindToken) pair.getSecond());
       if (!token.equals(pair.getSecond())) {
         // incarnation id has changed or there is unclean shutdown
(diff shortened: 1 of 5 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1393,6 +1393,17 @@ private boolean isDeleteTombstoneRemovable(IndexEntry deleteIndexEntry, IndexSeg
     }
     for (Map.Entry<String, Pair<Long, FindToken>> entry : remoteTokenTracker.getPeerReplicaAndToken().entrySet()) {
       Pair<Long, FindToken> pair = entry.getValue();
+      // Ignore this peer if the last active replication's timestamp of this peer is longer than the pre-configured days.
```
JingQianCloud, line 1396: By default, logger trace is not enabled. So we won't know if the "skip stale replica" condition is hit.
We probably can add logger.error or logger.info for the first time when we start to ignore this particular peer.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -1393,6 +1393,17 @@ private boolean isDeleteTombstoneRemovable(IndexEntry deleteIndexEntry, IndexSeg
...
+      if (config.storeCompactionIgnorePeersUnavailableForDays != 0
+          && pair.getFirst() + TimeUnit.DAYS.toMillis(config.storeCompactionIgnorePeersUnavailableForDays)
+          < time.milliseconds()) {
+        if (logger.isTraceEnabled()) {
```
SophieGuo410, line 1400: May I know why do we need this condition for trace log?

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
justinlin-linkedin: This would not be necessary. I prefer to adding a new metric for this.

JingQianCloud (approved):
When the ignoring peer token happens, we need some visibility, either log or metrics are good for me. 
I'm approving this PR. Please add the metrics either in this PR or next PR.

justinlin-linkedin:
> When the ignoring peer token happens, we need some visibility, either log or metrics are good for me. I'm approving this PR. Please add the metrics either in this PR or next PR.

Added logs in the new commit

JingQianCloud (approved):
Approved. Thanks,

PR #2780, opened 2024-05-07
[Frontend] Fix memory leak in 100-continue feature

snalli:
This patch fixes a memory bloat issue in 100-continue feature.

First, the if-clause to handle 100-continue should be `success && (a || b || c)` but was this `success && (a || b) || c`.

Second, the Response wasn't linked to Request. As a result, the Request is dangling and not released into the heap causing memory leak. In the future, the Request could be accepted as constructor arg to Response to avoid null pointers.


SophieGuo410 (approved):
Thanks for fixing this.

SophieGuo410 (approved):

PR #2786, opened 2024-05-16
[DR] Fix false alarms in backup integrity monitor

snalli:
This patch fixes two issues in backup integ. monitor.
* First, it fixes a bug that causes a lot of false alarms making us think blobs were missing either on-prem or in Azure. Specifically, we were recording blobs restored from Azure in map as `azureBlobs.put(storeKey.ID)` but we were looking for it as `azureBlobs.put(storeKey)`. This was a mistake.
* Second, this patch adds more clarity as to where a blob is missing - in server or in Azure.
* This patch also tries to reduce the number of false alarms by ensuring the server data is scanned and compared with Azure data until the point it is backed-up in Azure.


justinlin-linkedin (approved):
LGTM

PR #2787, opened 2024-05-20
[DR] Compare metadata from all replicas of a partition with its backup

snalli:
This patch adds support in the verification-app to compare blob metadata between all 3 server replicas of a partition and its cloud backup. It also adds some filters to reduce false alarms and unhelpful comparisons.
Lastly, we leave the partition on the local disk of the verification app to allow for debugging, instead of wiping it out. It will be wiped out later anyways.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
@@ -288,7 +288,7 @@ public class BackupIntegrityMonitor implements Runnable {
    * @throws IOException
    */
   HashMap<String, MessageInfo> getAzureBlobsFromLocalStore(BlobStore store) throws StoreException, IOException {
-    MessageReadSet rdset = null;
+    MessageReadSet rdset;
     HashMap<String, MessageInfo> azureBlobs = new HashMap<>();
     StoreFindToken newDiskToken = new StoreFindToken(), oldDiskToken = null;
     EnumSet<StoreGetOptions> storeGetOptions = EnumSet.of(StoreGetOptions.Store_Include_Deleted,
@@ -314,7 +314,7 @@ public class BackupIntegrityMonitor implements Runnable {
             }
           }
         }
-        azureBlobs.put(msg.getStoreKey().getID(), msg);
+        azureBlobs.put(msg.getStoreKey().getID(), new MessageInfo(msg, crc));
       }
       oldDiskToken = newDiskToken;
       newDiskToken = (StoreFindToken) finfo.getFindToken();
@@ -369,7 +369,7 @@ public class BackupIntegrityMonitor implements Runnable {
       if (partitionBackedUpUntil == Utils.Infinite_Time) {
         partitionBackedUpUntil = System.currentTimeMillis() - SCAN_STOP_RELTIME;
       }
-      
+
       /** Create local Store S */
       BlobStore store = startLocalStore(partition);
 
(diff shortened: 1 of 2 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -272,71 +386,33 @@ public void run() {
...
-      logger.info("[BackupIntegrityMonitor] Verified cloud backup partition-{} against peer server replica {}"
-              + ", num_azure_blobs = {}, num_server_blobs = {}, num_mismatches = {}",
-          partition.getId(), replica, azureToken.getNumBlobs(), serverScanner.getNumBlobScanned(),
-          rmetrics.backupIntegrityError.getCount());
+      /** Compare metadata from server replicas with metadata from Azure */
+      List<RemoteReplicaInfo> serverReplicas =
+          serverReplicationManager.createRemoteReplicaInfos(replicas, store.getReplicaId());
+      for (RemoteReplicaInfo serverReplica : serverReplicas) {
+        /** Create a temporary map of all keys recovered from cloud */
+        HashMap<String, MessageInfo> azureBlobs = getAzureBlobsFromLocalStore(store);
```
justinlin-linkedin, line 406: we only need to do this once, instead of for every server replica.

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -210,6 +215,109 @@ private boolean stopLocalStore(AmbryPartition partition) {
...
+      e.printStackTrace();
+    }
+
+    try {
+      if (serverReplica != null) {
+        serverScanner.removeRemoteReplicaInfo(serverReplica);
+      }
+    } catch (Throwable e) {
+      metrics.backupCheckerRuntimeError.inc();
+      logger.error("[BackupIntegrityMonitor] Failed to dequeue server replica {} due to {}",
```
justinlin-linkedin, line 270: you can add exception to the end of the log error method, and in your message string, you don't have to reserve a {} for the exception. In this way, logger would be able to print out the message of the exception as well as the stack trace.

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
```diff
@@ -210,6 +215,109 @@ private boolean stopLocalStore(AmbryPartition partition) {
...
+      serverScanner.printKeysAbsentInServer(serverReplica);
+      ReplicationMetrics rmetrics = new ReplicationMetrics(cluster.getMetricRegistry(), Collections.emptyList());
+      logger.info("[BackupIntegrityMonitor] Verified cloud backup partition-{} against peer server replica {}"
+              + ", num_azure_blobs = {}, num_server_blobs = {}, num_mismatches = {}",
+          partitionId, serverReplica, ((RecoveryToken) cloudReplica.getToken()).getNumBlobs(),
+          serverScanner.getNumBlobScanned(),
+          rmetrics.backupIntegrityError.getCount());
+    } catch (Throwable e) {
+      metrics.backupCheckerRuntimeError.inc();
+      logger.error("[BackupIntegrityMonitor] Failed to verify server replica {} due to {}",
```
justinlin-linkedin, line 258: same here

ambry-cloud/src/main/java/com/github/ambry/cloud/BackupIntegrityMonitor.java
snalli: Actually, I use this map to get a diff of keys present in server and cloud. This map contains keys present in azure. When replicating from server, i remove keys from this map that belong to server. At the end, i am left with keys that are in azure but not in server. So before each loop, i need to recreate this map.

justinlin-linkedin: I understand that, but it's probably better to just do a shadow copy of this map for each replica then to recreate this map.
However, this is fine, it doesn't change its correctness.

justinlin-linkedin (approved):

PR #2788, opened 2024-05-20
[WIP][DR] Compute CRC for backup recovery and verification

snalli:


ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -693,6 +695,34 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+              Store storeToGet = storeManager.getStore(partitionId);
+              EnumSet<StoreGetOptions> storeGetOptions = EnumSet.of(StoreGetOptions.Store_Include_Deleted,
+                  StoreGetOptions.Store_Include_Expired);
+              List<MessageInfo> newMessageInfos = new ArrayList<>();
+              MessageFormatSend msgFmt = new MessageFormatSend();
+              // for-each blob
+              findInfo.getMessageEntries().forEach(minfo -> {
+                MessageInfo newMsgInfo = minfo;
+                try {
+                  List<StoreKey> keys = getConvertedStoreKeys(Collections.singletonList(minfo.getStoreKey()))
```
justinlin-linkedin, line 709: we only have one storekey, we don't really need stream().distinct().collect. The returned list from getConvertedStoreKeys is totally fine.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -693,6 +695,34 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
             logger.trace("{} Time used to find entry since: {}", partitionId,
                 (SystemTime.getInstance().milliseconds() - partitionStartTimeInMs));
 
+            // Compute CRC for backup recovery and verification
+            if (replicaMetadataRequestInfo.getReplicaPath().startsWith(BackupCheckerThread.DR_Verifier_Keyword)) {
+              Store storeToGet = storeManager.getStore(partitionId);
```
justinlin-linkedin, line 700: you don't need a new variable `storeToGet` here, the `store` before this if statement can be used here.

ambry-messageformat/src/main/java/com/github/ambry/messageformat/MessageFormatSend.java
```diff
@@ -241,6 +245,13 @@ private void fetchDataFromReadSet(MessageFormatMetrics metrics) throws MessageFo
     }
   }
 
+  public long getBlobCRCOffset(MessageReadSet messages, int idx) throws MessageFormatException, IOException {
```
justinlin-linkedin, line 248: there is a better way to know the offset of CRC. When you call `store.get(storeKeys, storeGetOptions)`, it returns a `MessageReadSet` and a list of `MessageInfo`. In this case, we only have one `MessageInfo` and one `MessageReadSet`. The `MessageInfo`'s `getSize()` would return the size of this entire message, what we are trying to get is the crc of this message. As long as we know the message size, we can just do `MessageInfo.getSize() - MessageFormatRecord.Crc_Size`. Something like that. 

So you don't really need to have a new method to get the crc offset.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -693,6 +695,34 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+                  StoreGetOptions.Store_Include_Expired);
+              List<MessageInfo> newMessageInfos = new ArrayList<>();
+              MessageFormatSend msgFmt = new MessageFormatSend();
+              // for-each blob
+              findInfo.getMessageEntries().forEach(minfo -> {
+                MessageInfo newMsgInfo = minfo;
+                try {
+                  List<StoreKey> keys = getConvertedStoreKeys(Collections.singletonList(minfo.getStoreKey()))
+                      .stream().distinct().collect(Collectors.toList());
+                  MessageReadSet rdset = storeToGet.get(keys, storeGetOptions).getMessageReadSet();
```
justinlin-linkedin, line 711: we should only do this when the MessageInfo shows that it's not DELETE, TTL_UPDATE and UNDELETE, that this is a PUT message info.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -693,6 +695,34 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+                  long crcOffset = msgFmt.getBlobCRCOffset(rdset, 0); // index = 0, as we have 1 msg
+                  rdset.doPrefetch(0, crcOffset, 8); // crc is 8 bytes
+                  long crc = rdset.getPrefetchedData(0).getLong(0);
+                  newMsgInfo = new MessageInfo(crc, minfo);
```
justinlin-linkedin, line 715: We have to release the prefetched data buffer. 
`rdset.getPrefetchedData(0).release()`

ambry-api/src/main/java/com/github/ambry/store/MessageInfo.java
```diff
@@ -191,6 +191,11 @@ public MessageInfo(StoreKey key, MessageInfo other) {
         other.getOperationTimeMs(), other.getLifeVersion());
   }
 
+  public MessageInfo(long crc, MessageInfo other) {
```
justinlin-linkedin, line 194: we should put MessageInfo as first parameter and crc as second one.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
snalli: Hmm.. the `Store` returns a single `MessageInfo` with all state changes for a blob including Put, Del, TtlUpdate etc. And the verification app is essentially always bootstrapping as it starts from an empty disk. So the verification app sees just 1 message for a blob with all state changes. And this is its only chance at comparing CRCs. If we exclude messages that have Del, TtlUpdate etc. flags set, then we will be unable to compare CRCs.

justinlin-linkedin: we would be fine. We probably should remove the delete at least. I think ttl update and undelete should be included. The deleted MessageInfo could indicate that there is no blob content anymore.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -693,6 +696,37 @@ public void handleReplicaMetadataRequest(NetworkRequest request) throws IOExcept
...
+              List<MessageInfo> newMessageInfos = new ArrayList<>();
+              // for-each blob
+              findInfo.getMessageEntries().forEach(minfo -> {
+                MessageInfo newMsgInfo = minfo;
+                MessageReadSet rdset = null;
+                // get crc of entire blob including blob-properties, user-md, content etc.
+                try {
+                  List<StoreKey> keys = getConvertedStoreKeys(Collections.singletonList(minfo.getStoreKey()));
+                  rdset = store.get(keys, storeGetOptions).getMessageReadSet();
+                  rdset.doPrefetch(0, minfo.getSize() - MessageFormatRecord.Crc_Size,
```
justinlin-linkedin, line 712: you can't use `minfo.getSize()` here, if this is a Delete message info, then the size is 100. You should get the new MessageInfo from store.get method. This method returns StoreInfo, which has MessageInfo, which would be the size of the content.

justinlin-linkedin (approved):

justinlin-linkedin (approved):

PR #2790, opened 2024-05-22
Preventing retentionTime and retentionCount to be set to 0 for dataset

SophieGuo410:


ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -595,6 +595,10 @@ public static long getTtlFromRequestHeader(Map<String, Object> args) throws Rest
         throw new RestServiceException(Headers.TTL + "[" + ttlFromHeader + "] is not valid (has to be >= -1)",
             RestServiceErrorCode.InvalidArgs);
       }
+      if (ttlFromHeader == 0) {
+        throw new RestServiceException(Headers.TTL + "[" + ttlFromHeader + "] should not be 0",
+            RestServiceErrorCode.InvalidArgs);
+      }
```
justinlin-linkedin, line 601: we probably don't need this one here.

justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
SophieGuo410: Remove this part and updated to add test and filter out null(default) after serialization as we discussed.

ambry-api/src/main/java/com/github/ambry/account/Dataset.java
```diff
@@ -228,3 +232,30 @@ public String toString() {
...
+
+    // Use reflection to get all fields of the Dataset class
+    Field[] fields = Dataset.class.getDeclaredFields();
+    for (Field field : fields) {
+      if (field.isAnnotationPresent(JsonProperty.class)) {
+        try {
+          field.setAccessible(true);
+          Object value = field.get(dataset);
+          if (value != null) {
+            gen.writeObjectField(field.getName(), value);
```
justinlin-linkedin, line 252: we should probably use annotation value here. something like
```
gen.writeObjectField(field.getAnnotation(JsonProperty.class).value(), value);
```

justinlin-linkedin (approved):

ambry-api/src/main/java/com/github/ambry/account/Dataset.java
SophieGuo410: Sure, updated it.

PR #2792, opened 2024-05-23
[WIP] Paranoid durability changes - not to be merged yet

litingulfs:


ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,405 @@
...
+    LinkedList<ReplicaId> downReplicas = new LinkedList<>();
+    if (shuffleReplicas) {
+      Collections.shuffle(replicas);
+    }
+
+    // The priority here is local dc replicas, originating dc replicas, other dc replicas, down replicas.
+    // To improve read-after-write performance across DC, we prefer to take local and originating replicas only,
+    // which can be done by setting includeNonOriginatingDcReplicas False.
+    List<ReplicaId> examinedReplicas = new ArrayList<>();
+    int numLocalAndLiveReplicas = 0;
```
justinlin-linkedin, line 95: we are not using these two variables anywhere, we can remove them.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,405 @@
...
+
+  int getCurrentRemoteParallelism() {
+    return remoteReplicaParallelism;
+  }
+
+  @Override
+  public Iterator<ReplicaId> getReplicaIterator() {
+    replicaIterator = replicaPool.iterator();                  //Iterator<ReplicaId>, since replicaPool is a List<ReplicaId>
+    // replicaByDcIterator will be a new iterator for the map
+    return otIterator;
```
justinlin-linkedin, line 319: This `otIterator` is initialized in `SimpleOperationTracker` to a private class defined inside `SimpleOperationTracker`. We need to change this. The `otIterator` now should be initialized as an instance of `ParanoidDurabilityTrackerIterator`.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,405 @@
...
+    return replicaParallelism;
+  }
+
+  int getCurrentRemoteParallelism() {
+    return remoteReplicaParallelism;
+  }
+
+  @Override
+  public Iterator<ReplicaId> getReplicaIterator() {
+    replicaIterator = replicaPool.iterator();                  //Iterator<ReplicaId>, since replicaPool is a List<ReplicaId>
```
justinlin-linkedin, line 317: we don't need this iterator anymore, this is not used any where in this class.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,405 @@
...
+
+    localReplicaSuccessTarget =
+        routerConfig.routerGetEligibleReplicasByStateEnabled ? Math.max(localEligibleReplicas.size() - 1, routerConfig.routerPutSuccessTarget) : routerConfig.routerPutSuccessTarget;
+    remoteReplicaSuccessTarget = routerConfig.routerPutRemoteSuccessTarget;
+
+    // Tommy: End of switch statement operations.
+
+    addReplicasToPool(allReplicas, offlineReplicas, true, false);
+
+
```
justinlin-linkedin, line 77: We should initialize a ParanoidDurabilityTrackerIterator here and return it in `getReplicaIterator`.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,405 @@
...
+    return true;
+  }
+
+  public boolean hasSucceeded() {
+    boolean hasSucceeded;
+    // Tommy: The original method in SimpleOperationTracker checks whether or not the current routerOperation is
+    //        a RouterOperation.PutOperation, but we already know that must be true if this class is being used at all.
+    if (routerConfig.routerPutUseDynamicSuccessTarget) {
+      // this logic only applies to replicas where the quorum can change during replica movement
+      int dynamicSuccessTarget = Math.max(totalReplicaCount - disabledCount - 1, routerConfig.routerPutSuccessTarget);
```
justinlin-linkedin, line 184: this is not right, the dynamicSuccessTarget is based on the replica pool size, which is not valid in the `ParanodDuabilityOperationTracker`. What should be compared here is whether we succeed in local and remote colo.

Can we have two methods, `hasSucceededInLocal` and `hasSucceedInRemote`. 

`hasSucceededInRemote` is simple, we just have to compare the remote success target with the remote success count.
`hasSucceededInLocal`, we just have to compare with the local success target that you initialized in the constructor.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,405 @@
...
+  public void onResponse(ReplicaId replicaId, TrackedRequestFinalState trackedRequestFinalState) {
+    super.onResponse(replicaId, trackedRequestFinalState);
+    modifyReplicasInPoolOrInFlightCount(replicaId, -1); // Tommy: This counter only used for offline repair so can ignore?
+    modifyReplicasInPoolOrInFlightCount(replicaId, -1);
+
+    String dcName = replicaId.getDataNodeId().getDatacenterName();
+
+    switch (trackedRequestFinalState) { // Paranoid durability - do we need to consider this switch block at all?
+      case SUCCESS:
+        replicaSuccessCountByDc.put(dcName, replicaSuccessCountByDc.get(dcName) + 1);
```
justinlin-linkedin, line 233: we should increase the local success count or remote success count here.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,405 @@
...
+          totalNotFoundCount++;
+          if (replicaId.getDataNodeId().getDatacenterName().equals(originatingDcName)) {
+            if (originatingDcLeaderOrStandbyReplicas.contains(replicaId)) {
+              // Since bootstrap replicas could still be catching up, treat NotFound responses only from leader or
+              // standby replicas as valid.
+              originatingDcNotFoundCount++;
+            }
+          }
+        }
+    }
```
justinlin-linkedin, line 258: for uploads, this will never be sure.

justinlin-linkedin:
I think we don't want to differentiate different remote colos. We only care about the success in local replicas and one remote replicas. For that, we can create two different set of replicas, one for local replicas, one for remote. 

So the map `replicaByDc` can have two keys, "LOCAL" and "REMOTE", or just "true" or "false" to indicate if it's local or remote. 

For the remote replicas, we also want to make sure that we get leaders and standbys, so we still need to call getEligibleReplicas to get the leader and standby replicas from the remote.

PR #2794, opened 2024-05-31
Paranoid durability - basic functionality

litingulfs:
This PR adds the basic required functionality for paranoid durability, where we require PUTs to succeed in at least one remote colo, reducing the risk of data loss in exceptional situations.

The writes in the local colo follow the same logic as before, but the remote (i.e. cross colo) writes are executed as follows:

- We allow simultaneous local and remote writes (respecting the parallelism configurations).
- We prefer to write to remote replicas that are in Helix LEADER state.
- We alternate writing to each remote data center. So, for example, if the current (local) colo is prod-ltx1, then we might attempt to write first to the LEADER in prod-lor1. If that write fails, we'll try to write to the LEADER in prod-lva1 instead of attempting further replicas in prod-lor1. The assumption behind this is that if a remote colo write fails, it is likely that we have hit a networking or colo-related issue, and so we should try the other colo now.

I've tested this locally, and as agreed with the rest of the team I'll follow up with pull requests with tests and metrics and the required wiring.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -273,7 +273,7 @@ public class RouterConfig {
    * Paranoid durability: The min number of successful responses in remote data centers required for a put operation.
    */
   @Config(ROUTER_PUT_REMOTE_SUCCESS_TARGET)
-  @Default("0")
+  @Default("1")
   public final int routerPutRemoteSuccessTarget;
 
   /**
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -259,7 +259,7 @@ public class RouterConfig {
    * Paranoid durability: Max parallel requests issued at a time in remote data centers for a chunk.
    */
   @Config(ROUTER_PUT_REMOTE_REQUEST_PARALLELISM)
-  @Default("0")
+  @Default("1")
   public final int routerPutRemoteRequestParallelism;
 
   /**
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -60,10 +60,8 @@ public class RouterConfig {
   public static final String ROUTER_DROP_REQUEST_ON_TIMEOUT = "router.drop.request.on.timeout";
   public static final String ROUTER_MAX_PUT_CHUNK_SIZE_BYTES = "router.max.put.chunk.size.bytes";
   public static final String ROUTER_PUT_REQUEST_PARALLELISM = "router.put.request.parallelism";
-  public static final String ROUTER_PUT_LOCAL_REQUEST_PARALLELISM = "router.put.local.request.parallelism";
   public static final String ROUTER_PUT_REMOTE_REQUEST_PARALLELISM = "router.put.remote.request.parallelism";
   public static final String ROUTER_PUT_SUCCESS_TARGET = "router.put.success.target";
-  public static final String ROUTER_PUT_LOCAL_SUCCESS_TARGET = "router.put.local.success.target";
   public static final String ROUTER_PUT_REMOTE_SUCCESS_TARGET = "router.put.remote.success.target";
   public static final String ROUTER_REPLICATE_BLOB_REQUEST_PARALLELISM = "router.replicate.blob.request.parallelism";
   public static final String ROUTER_REPLICATE_BLOB_SUCCESS_TARGET = "router.replicate.blob.success.target";
@@ -258,13 +256,6 @@ public class RouterConfig {
   public final int routerPutRequestParallelism;
 
   /**
-   * Paranoid durability: Max parallel requests issued at a time in the local data center for a chunk.
-   */
-  @Config(ROUTER_PUT_LOCAL_REQUEST_PARALLELISM)
-  @Default("3")
-  public final int routerPutLocalRequestParallelism;
-
-  /**
    * Paranoid durability: Max parallel requests issued at a time in remote data centers for a chunk.
    */
   @Config(ROUTER_PUT_REMOTE_REQUEST_PARALLELISM)
@@ -277,13 +268,6 @@ public class RouterConfig {
   @Config(ROUTER_PUT_SUCCESS_TARGET)
   @Default("2")
   public final int routerPutSuccessTarget;
-
-  /**
-   * Paranoid durability: The min number of successful responses in the local data center required for a put operation.
-   */
-  @Config(ROUTER_PUT_LOCAL_SUCCESS_TARGET)
-  @Default("2")
-  public final int routerPutLocalSuccessTarget;
 
   /**
    * Paranoid durability: The min number of successful responses in remote data centers required for a put operation.
@@ -784,13 +768,10 @@ public class RouterConfig {
         verifiableProperties.getIntInRange(ROUTER_MAX_PUT_CHUNK_SIZE_BYTES, 4 * 1024 * 1024, 1, Integer.MAX_VALUE);
     routerPutRequestParallelism =
         verifiableProperties.getIntInRange(ROUTER_PUT_REQUEST_PARALLELISM, 3, 1, Integer.MAX_VALUE);
-    routerPutLocalRequestParallelism =
-        verifiableProperties.getIntInRange(ROUTER_PUT_LOCAL_REQUEST_PARALLELISM, 3, 1, Integer.MAX_VALUE);
     routerPutRemoteRequestParallelism =
-        verifiableProperties.getIntInRange(ROUTER_PUT_REMOTE_REQUEST_PARALLELISM, 0, 0, Integer.MAX_VALUE);
+        verifiableProperties.getIntInRange(ROUTER_PUT_REMOTE_REQUEST_PARALLELISM, 1, 1, Integer.MAX_VALUE);
     routerPutSuccessTarget = verifiableProperties.getIntInRange(ROUTER_PUT_SUCCESS_TARGET, 2, 1, Integer.MAX_VALUE);
-    routerPutLocalSuccessTarget = verifiableProperties.getIntInRange(ROUTER_PUT_LOCAL_SUCCESS_TARGET, 2, 1, Integer.MAX_VALUE);
-    routerPutRemoteSuccessTarget = verifiableProperties.getIntInRange(ROUTER_PUT_REMOTE_SUCCESS_TARGET, 0, 0, Integer.MAX_VALUE);
+    routerPutRemoteSuccessTarget = verifiableProperties.getIntInRange(ROUTER_PUT_REMOTE_SUCCESS_TARGET, 1, 1, Integer.MAX_VALUE);
     routerReplicateBlobRequestParallelism =
         verifiableProperties.getIntInRange(ROUTER_REPLICATE_BLOB_REQUEST_PARALLELISM, 3, 1, Integer.MAX_VALUE);
     routerReplicateBlobSuccessTarget =
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -60,7 +60,11 @@ public class RouterConfig {
   public static final String ROUTER_DROP_REQUEST_ON_TIMEOUT = "router.drop.request.on.timeout";
   public static final String ROUTER_MAX_PUT_CHUNK_SIZE_BYTES = "router.max.put.chunk.size.bytes";
   public static final String ROUTER_PUT_REQUEST_PARALLELISM = "router.put.request.parallelism";
+  public static final String ROUTER_PUT_LOCAL_REQUEST_PARALLELISM = "router.put.local.request.parallelism";
+  public static final String ROUTER_PUT_REMOTE_REQUEST_PARALLELISM = "router.put.remote.request.parallelism";
   public static final String ROUTER_PUT_SUCCESS_TARGET = "router.put.success.target";
+  public static final String ROUTER_PUT_LOCAL_SUCCESS_TARGET = "router.put.local.success.target";
+  public static final String ROUTER_PUT_REMOTE_SUCCESS_TARGET = "router.put.remote.success.target";
   public static final String ROUTER_REPLICATE_BLOB_REQUEST_PARALLELISM = "router.replicate.blob.request.parallelism";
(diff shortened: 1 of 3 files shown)


ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -253,13 +257,41 @@ public class RouterConfig {
...
+   */
+  @Config(ROUTER_PUT_LOCAL_REQUEST_PARALLELISM)
+  @Default("3")
+  public final int routerPutLocalRequestParallelism;
+
+  /**
+   * Paranoid durability: Max parallel requests issued at a time in remote data centers for a chunk.
+   */
+  @Config(ROUTER_PUT_REMOTE_REQUEST_PARALLELISM)
+  @Default("0")
```
justinlin-linkedin, line 271: default value should be at least 1. I think we should use 1 as default value.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -253,13 +257,41 @@ public class RouterConfig {
...
+   */
+  @Config(ROUTER_PUT_LOCAL_SUCCESS_TARGET)
+  @Default("2")
+  public final int routerPutLocalSuccessTarget;
+
+  /**
+   * Paranoid durability: The min number of successful responses in remote data centers required for a put operation.
+   */
+  @Config(ROUTER_PUT_REMOTE_SUCCESS_TARGET)
+  @Default("0")
```
justinlin-linkedin, line 292: same thing here.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+  ParanoidDurabilityOperationTracker(RouterConfig routerConfig, PartitionId partitionId, String originatingDcName,
+      NonBlockingRouterMetrics routerMetrics) {
+    super(routerConfig, RouterOperation.PutOperation, partitionId, originatingDcName, true, routerMetrics);
+
+    localReplicaParallelism = routerConfig.routerPutLocalRequestParallelism;
+    remoteReplicaParallelism = routerConfig.routerPutRemoteRequestParallelism;
+    localReplicaSuccessTarget = routerConfig.routerPutLocalSuccessTarget;
+    remoteReplicaSuccessTarget = routerConfig.routerPutRemoteSuccessTarget;
+
+    List<ReplicaId> allEligibleReplicas = getEligibleReplicas(datacenterName, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER));
```
justinlin-linkedin, line 58: To get eligible replicas for all datacenters, we have to pass a "null" parameter to first argument.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+    addReplicasToPool(allEligibleReplicas);
+    this.otIterator = new ParanoidDurabilityTrackerIterator();
+  }
+
+  private void addReplicasToPool(List<? extends ReplicaId> replicas) {
+    Collections.shuffle(replicas);
+
+    for (ReplicaId replicaId : replicas) {
+      if (!replicaId.isDown()) {
+          addToBeginningOfPool(replicaId);
```
justinlin-linkedin, line 68: the indentation is a bit wacky here.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+    for (String currentDc : replicaPoolByDc.keySet()) {
+      if(currentDc.equals(datacenterName)) {
+        if (replicaPoolByDc.get(currentDc).size() < localReplicaSuccessTarget) {
+          logger.error("Not enough replicas in local data center for partition " + partitionId +
+              " to satisfy paranoid durability requirements " +
+              "(wanted " + localReplicaSuccessTarget + ", but found " + replicaPoolByDc.get(currentDc).size() + ")");
+          return false;
+        }
+      } else {
+        if (replicaPoolByDc.get(currentDc).size() < remoteReplicaSuccessTarget) {
```
justinlin-linkedin, line 116: The success target should not be per dc for remote dcs.
If we have two remote dcs, and the success target is 1, as long as we have 1 replica available from either DC, it would be good. But I guess this is fine. The remote success target should be a low number for both dcs.

justinlin-linkedin (commented):
first round of review

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+  public Iterator<ReplicaId> getReplicaIterator() {
+    return otIterator;
+  }
+
+  private class ParanoidDurabilityTrackerIterator implements Iterator<ReplicaId> {
+    private int currentDatacenterIndex = 0;
+
+    @Override
+    public boolean hasNext() {
+      return hasNextLocal() || hasNextRemote();
```
justinlin-linkedin, line 246: I think we should formalize the order of each replica somewhere so there is no confusion.

The idea is that we should try local datacenter first, make sure local datacenter succeeds, then we move on to remote. If this is true, then the logic of `hasNext` should look something like
```
return hasNextLocal() || (localFinished() && hasNextRemote());
```

This is important. For example, if the local parallelism is 2 and we have 3 local replicas, then 
```
Iterator<ReplicaId> iter = operationTracker.getReplicaIterator();
iter.hasNext(); // should return true;
ReplicaId r1 = iter.next();
iter.remove();
iter.hasNext(); // should return true;
ReplicaId r2 = iter.next();
iter.remove();
it.hasNext() // should return false, we only need two parallel local replicas.
```
We should not return true on third `hasNext` call, since the local replicas are not yet finished.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+            routerConfig.routerPutRemoteSuccessTarget + disabledCountByDc.get(dcName))) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
+  private boolean hasFailedLocally() {
+    return replicaInPoolOrFlightCountByDc.get(datacenterName) + replicaSuccessCountByDc.get(datacenterName) < replicaSuccessTarget;
```
justinlin-linkedin, line 215: you mean `localReplicaSuccessTarget`?

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+    return true;
+  }
+
+  public boolean hasSucceeded() {
+    return hasSucceededLocally() && hasSucceededRemotely();
+  }
+
+  private boolean hasSucceededLocally() {
+    return localReplicaSuccessCount >= localReplicaSuccessTarget;
+  }
```
justinlin-linkedin, line 133: we should cconsider dynamic target for local success here, just like the `hasFailed` method.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+    } else {
+      return hasFailedLocally() || hasFailedRemotely();
+    }
+  }
+
+  private boolean hasFailedDynamically(String dc, int successTarget)  {
+    return (totalReplicaCountByDc.get(dc) - failedCountByDc.get(dc)) < Math.max(totalReplicaCountByDc.get(dc) - 1, successTarget + disabledCountByDc.get(dc));
+  }
+
+  private boolean hasFailedRemotely() {
```
justinlin-linkedin, line 202: for `hasFailedRemotely`, we have replicas from remote dcs, and we have a success target, as long as the success target + inflight and inpool replicas is less than remote success target, we can return true.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
...
+    }
+  }
+
+  public boolean hasFailed() {
+    if (routerConfig.routerPutUseDynamicSuccessTarget) {
+      for(String dcName : allDatacenters) {
+        if (dcName.equals(datacenterName)) {  // Local colo
+          if (hasFailedDynamically(dcName, routerConfig.routerPutSuccessTarget)) return true;
+        } else {
+          if (hasFailedDynamically(dcName, routerConfig.routerPutRemoteSuccessTarget)) return true;
```
justinlin-linkedin, line 189: the dynamic target should only apply to local dc.

justinlin-linkedin (commented):
The more I think about this, the more i feel that we shouldn't differentiate remote dcs, We should just put all remote replicas in a list. After local dc succeeds, we should just iterate through the remote list. It would probably be a bit cleaner this way.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -60,7 +60,11 @@ public class RouterConfig {
   public static final String ROUTER_DROP_REQUEST_ON_TIMEOUT = "router.drop.request.on.timeout";
   public static final String ROUTER_MAX_PUT_CHUNK_SIZE_BYTES = "router.max.put.chunk.size.bytes";
   public static final String ROUTER_PUT_REQUEST_PARALLELISM = "router.put.request.parallelism";
+  public static final String ROUTER_PUT_LOCAL_REQUEST_PARALLELISM = "router.put.local.request.parallelism";
```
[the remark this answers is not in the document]
litingulfs: OK, I simplified this.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -60,7 +60,11 @@ public class RouterConfig {
...
+  public static final String ROUTER_PUT_REMOTE_REQUEST_PARALLELISM = "router.put.remote.request.parallelism";
   public static final String ROUTER_PUT_SUCCESS_TARGET = "router.put.success.target";
+  public static final String ROUTER_PUT_LOCAL_SUCCESS_TARGET = "router.put.local.success.target";
```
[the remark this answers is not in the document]
litingulfs: 👍

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
litingulfs: Agreed, and changed this as requested.

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
litingulfs: 👍

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: Good catch, I meant to pass the null but somehow missed it.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: Yep, I changed it.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: Yeah, as discussed below I changed the tracking of replicas from per-DC tracking to just tracking local and remote, so this should be fixed now.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -0,0 +1,314 @@
+package com.github.ambry.router;
```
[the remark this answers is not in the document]
litingulfs: Good catch, I added the license as required.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: We talked about this at length in a meeting and agreed to keep this part of the PR as it exists. The idea is that we fire off requests to the local colo as well as the remote colos in parallel (while staying within the configured bounds), in order to reduce the full latency as seen by our clients.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: Yeah, changed this.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: Yeah, I had misunderstood the dynamic target and its relevance, so updated the local success and failure computations.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: You're right, I simplified this as suggested.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: 👍 agreed and changed.

litingulfs:
> The more I think about this, the more i feel that we shouldn't differentiate remote dcs, We should just put all remote replicas in a list. After local dc succeeds, we should just iterate through the remote list. It would probably be a bit cleaner this way.

Yeah, we spoke about this offline, and I've changed the code so that instead of tracking replicas on a DC-by-DC basis, we just track local vs remote replicas. There is a slight cost to this, since we now end up doing more work to sort the remote replicas so that we can alternate between colos and also prefer LEADER replicas, but I think overall the code is now easier to understand.

justinlin-linkedin (approved):

PR #2797, opened 2024-06-05
[DR] Re-check integrity of Azure blobs

snalli:
If the verification-app encounters a crc or blob-size mismatch, it is most likely due to differing versions of blob-properties on the server and cloud. So, re-verify the integrity of Azure blob by deserializing it.


ambry-replication/src/main/java/com/github/ambry/replication/BackupCheckerThread.java
```diff
@@ -199,8 +241,18 @@ List<ExchangeMetadataResponse> handleReplicaMetadataResponse(ReplicaMetadataResp
                 MessageInfo azureBlob = azureBlobMap.remove(serverBlob.getStoreKey().getID());
                 Set<BlobMatchStatus> status = azureBlob == null ? Collections.singleton(BLOB_ABSENT_IN_AZURE)
                     : serverBlob.isEqual(azureBlob);
+                // if size mismatch, then re-check
+                // this is the only workaround due to a field reservedMetadataBlobId
```
justinlin-linkedin, line 245: one minor comment: due to different versions on serialization.

justinlin-linkedin (approved):

PR #2801, opened 2024-06-12
[vcr-2.0] Exploit metadata caching to speed up backups

snalli:
The ReplicationEngine operates by first identifying keys that are missing in Azure, which incurs one metadata request per blob. For existing blobs, the engine then retrieves their metadata, adding another request per blob. Additionally, the engine performs checks for expiry and deletion, each costing two requests per blob. These requests can be minimized to one by using caching.

This patch introduces a thread-local cache that selectively stores metadata. The `findMissingKeys` function caches metadata for existing blobs, which is subsequently used by the `findKeys` and `applyXXX` methods.


justinlin-linkedin (approved):
LGTM

PR #2802, opened 2024-06-13
Metrics for the count of partitions where there are 1, 2, or 3 local replicas down

Arun-LinkedIn:
We are adding two types of metrics. One includes 'bootstrap/inactive' replicas as 'down' replicas while other doesn't.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
@@ -18,6 +18,7 @@ import com.codahale.metrics.Gauge;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 import com.github.ambry.clustermap.HelixClusterManager.HelixClusterManagerQueryHelper;
+import com.github.ambry.utils.SystemTime;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
@@ -63,8 +64,8 @@ class HelixClusterManagerMetrics {
   // These maps are updated when 'partitionCountWithOneReplicaDown' metric is populated. Rest of the metrics such as
   // 'partitionCountWithTwoReplicaDown', 'partitionCountWithThreeReplicaDown' use these values instead of re-calculating
   // the partitionCounts inorder to save cpu cycles.
-  public Map<Integer, Integer> localPartitionCountsByDownReplicas = new ConcurrentHashMap<>();
-  public Map<Integer, Integer> localPartitionCountsByNonLeaderStandbyReplicas = new ConcurrentHashMap<>();
+  private final Map<Integer, Integer> localPartitionCountsByDownReplicas = new ConcurrentHashMap<>();
+  private final Map<Integer, Integer> localPartitionCountsByNonLeaderStandbyReplicas = new ConcurrentHashMap<>();
 
   /**
    * Metrics for the {@link HelixClusterManager}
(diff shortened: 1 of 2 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
```diff
@@ -59,6 +60,11 @@ class HelixClusterManagerMetrics {
   public final Timer routingTableQueryTime;
   public final Counter resourceNameMismatchCount;
   public final Counter paranoidDurabilityIneligibleReplicaCount;
+  // These maps are updated when 'partitionCountWithOneReplicaDown' metric is populated. Rest of the metrics such as
+  // 'partitionCountWithTwoReplicaDown', 'partitionCountWithThreeReplicaDown' use these values instead of re-calculating
+  // the partitionCounts inorder to save cpu cycles.
+  public Map<Integer, Integer> localPartitionCountsByDownReplicas = new ConcurrentHashMap<>();
+  public Map<Integer, Integer> localPartitionCountsByNonLeaderStandbyReplicas = new ConcurrentHashMap<>();
```
justinlin-linkedin, line 67: If these maps are not referenced outside of this metric cluster, we can make then privite.
Nit: an int[] would just be fine. The index would represent the number of down replicas and the value would represents how many partitions.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
```diff
@@ -190,15 +196,104 @@ void initializePartitionMetrics() {
...
           return 1L;
         }
       }
       return 0L;
     };
     registry.gauge(MetricRegistry.name(HelixClusterManager.class, "isMajorityReplicasDownForAnyPartition"),
         () -> isMajorityReplicasDownForAnyPartition);
+
+    Gauge<Integer> getLocalPartitionCountWithOneReplicaDown = () -> {
+      String dcName = clusterMapCallback.getDatacenterName();
```
justinlin-linkedin, line 210: you wanna clear the map every time this method is called. so the previous value won't affect each invocation.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
```diff
@@ -190,15 +196,104 @@ void initializePartitionMetrics() {
...
+        int localLeaderStandByReplicaCount = 0;
+        for (List<? extends ReplicaId> replicaIds : partition.getReplicaIdsByStates(
+            EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY), dcName).values()) {
+          localLeaderStandByReplicaCount += replicaIds.size();
+        }
+        int localBootstrapReplicaCount = partition.getReplicaIdsByState(ReplicaState.BOOTSTRAP, dcName).size();
+        int localInactiveReplicaCount = partition.getReplicaIdsByState(ReplicaState.INACTIVE, dcName).size();
+
+        // Calculate the number of replicas that are down. Bootstrap (upcoming) and Inactive (deactivating) replicas are considered as good.
+        int downReplicaCount = totalLocalReplicaCount - localLeaderStandByReplicaCount - localBootstrapReplicaCount
```
MikeDafi, line 229: Shouldn't we count Inactive replicas as part of our  down states count so we have visibility on processes that gracefully shutdown the apps

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
```diff
@@ -190,15 +196,104 @@ void initializePartitionMetrics() {
...
+            .count();
+
+        // Get local leader/standby/bootstrap replica counts of this partition
+        int localLeaderStandByReplicaCount = 0;
+        for (List<? extends ReplicaId> replicaIds : partition.getReplicaIdsByStates(
+            EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY), dcName).values()) {
+          localLeaderStandByReplicaCount += replicaIds.size();
+        }
+        int localBootstrapReplicaCount = partition.getReplicaIdsByState(ReplicaState.BOOTSTRAP, dcName).size();
+        int localInactiveReplicaCount = partition.getReplicaIdsByState(ReplicaState.INACTIVE, dcName).size();
```
justinlin-linkedin, line 226: the query of bootstrap and inactive can be combined. 
```
int localLeaderStandbyCount = partition.getReplicaIdsByState(EnumSet.of(ReplicaState.LEADER, ReplicaState.STANDBY), dcName).values().stream().mapToInt(List::size).sum();
int localBootstrapInactiveCount = partition.getReplicaIdsByState(EnumSet.of(ReplicaState.BOOTSTRAP, ReplicaState.INACTIVE), dcName).values().stream().mapToInt(List::size).sum();
```

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
```diff
@@ -190,15 +196,104 @@ void initializePartitionMetrics() {
...
+          localPartitionCountsByNonLeaderStandbyReplicas.put(nonLeaderStandbyReplicaCount,
+              localPartitionCountsByNonLeaderStandbyReplicas.getOrDefault(nonLeaderStandbyReplicaCount, 0) + 1);
+        }
+      }
+
+      return localPartitionCountsByDownReplicas.getOrDefault(1, 0);
+    };
+
+    // Metric to see number of local partitions with 1 replica down
+    registry.gauge(MetricRegistry.name(HelixClusterManager.class, "localPartitionCountWithOneReplicaDown"),
```
justinlin-linkedin, line 250: we can probably remove "local" from the metric names (and variable names), just putting some comments here and in the ingraph, it should be fine. And do we have any metric showing number of partitions?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
```diff
@@ -190,15 +196,104 @@ void initializePartitionMetrics() {
...
+    Gauge<Integer> getLocalPartitionCountWithTwoReplicaDown =
+        () -> localPartitionCountsByDownReplicas.getOrDefault(2, 0);
+    registry.gauge(MetricRegistry.name(HelixClusterManager.class, "localPartitionCountWithTwoReplicaDown"),
+        () -> getLocalPartitionCountWithTwoReplicaDown);
+
+    // Metric to see number of local partitions with 3 or more replica down
+    Gauge<Integer> getLocalPartitionCountWithThreeOrMoreReplicaDown = () -> {
+      int numOfPartitions = 0;
+      for (int downReplicaCount : localPartitionCountsByDownReplicas.keySet()) {
+        if (downReplicaCount >= 3) {
```
MikeDafi, line 263: Instead of hardcoding the values 3,2,1. Can we have it be a function of the number of replicas in case of local cluster's 4-replica RT and video cluster's cluster 1-replica RT or should we just skip 1-replica RTs since that may skew the metric whenever a host goes down in video

justinlin-linkedin:
I am not sure about the speed of iterating through 35K partitions every minutes to check out its state, can we put some trace log in the Gauge method to print out how long it takes to do that. And if this take a long time, they are some ways we can improve the metric collection.

1. Do this every 10 minutes, as we don't need it real time. We can have a `lastUpdatedTimestamp` in the `HelixClusterManagerMetric` so we only have to recalculate every 10 minutes.
2. Do it in the `HelixClusterManager` in the `ExternalViewCallback`. We will only calculate whenever there is an update in `ExternalView`.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
Arun-LinkedIn: Sure. Made them private. Since these might need be to be thread safe, for now, left them as ConcurrentHashMap instead of using AtomicInteger[].

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
Arun-LinkedIn: Modified the metric and variable names. Yeah, we have below metric showing number of partitions..

```
    Gauge<Long> partitionCount = clusterMapCallback::getPartitionCount;
    registry.gauge(MetricRegistry.name(HelixClusterManager.class, "partitionCount"), () -> partitionCount);
```

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
Arun-LinkedIn: Sorry, didn't get you. In `partitionCountWithOneReplicaDown`, we ignore inactive replicas but include them in  `partitionCountWithTwoNonLeaderStandByReplica`. I am thinking that using both the metrics, we could get idea on the partitions which are unhealthy.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixClusterManagerMetrics.java
Arun-LinkedIn: Hmm, in 4-replica cluster too, we have count of partitions which lost 1 replica, 2 replica and 3+ replica. May be we could get on the partitions which lost 2 or more replicas? 

Yeah, you are right. For video cluster since we have a mix of 3 and 1 replica partitions, we may ignore 1-replica partitions which lost 1 replica. I am thinking we can fix it later.

justinlin-linkedin (approved):

PR #2803, opened 2024-06-14
[Paranoid Durability] Unit Tests and Bugfix

litingulfs:
This PR fixes a bug in ParanoidDurabilityOperationTracker where we didn't complete initialization before checking whether enough good replicas are present. It also adds unit tests.

### DIFF
--- a/ambry-router/src/test/java/com/github/ambry/router/ParanoidDurabilityOperationTrackerTest.java
+++ b/ambry-router/src/test/java/com/github/ambry/router/ParanoidDurabilityOperationTrackerTest.java
@@ -1,0 +1,267 @@
+/**
+ * Copyright 2024 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


ambry-router/src/test/java/com/github/ambry/router/ParanoidDurabilityOperationTrackerTest.java
```diff
@@ -0,0 +1,267 @@
...
+    mockPartition = new MockPartitionId();
+    mockClusterMap = new MockClusterMap(false, buildDataNodeList(testReplicas), 1, Collections.singletonList(mockPartition), localDcName);
+    populateReplicaList(testReplicas);
+
+    ParanoidDurabilityOperationTracker ot = getParanoidDurabilityOperationTracker(2, 1, 2, 1, true);
+    while (!ot.hasSucceeded()) {
+      sendRequests(ot, 3);
+      for (int i = 0; i < 3; i++) {
+        ot.onResponse(inflightReplicas.poll(), TrackedRequestFinalState.SUCCESS);
+      }
```
justinlin-linkedin, line 87: the parallelism for local is 2 and remote is 1, we are expecting to send 3 requests out, which is right. Can we verify that the first two inflight replicas are local and the last one is remote. And after all replicas return success, the operation tracker should return success.

litingulfs: OK, I added the verification to count local and remote successes. I think it was already checking that the operation tracker was returning success, but also added a check to confirm that no failure was detected in successful cases.

justinlin-linkedin (approved):

PR #2805, opened 2024-06-21
[Paranoid Durability] Add new container field for paranoid durability.

litingulfs:
This PR adds a new field for paranoid durability to Ambry's `Container` class. The paranoid durability feature applies to PUTs/writes, and requires a minimum of one replica in a remote colo to acknowledge a write before returning success. We want this feature to be used sparingly, since it will add additional latency and cross-colo I/O to all writes. Therefore, the default setting for all containers is false (i.e. paranoid durability is OFF by default). So this new setting will need to be explicitly turned on for each container that wants to use it.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/account/Container.java
+++ b/ambry-api/src/main/java/com/github/ambry/account/Container.java
@@ -67,6 +67,7 @@ public class Container {
   static final long CONTAINER_DELETE_TRIGGER_TIME_DEFAULT_VALUE = 0;
   static final boolean SECURE_PATH_REQUIRED_DEFAULT_VALUE = false;
   static final boolean OVERRIDE_ACCOUNT_ACL_DEFAULT_VALUE = false;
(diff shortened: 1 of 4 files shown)


ambry-api/src/main/java/com/github/ambry/account/Container.java
```diff
@@ -549,6 +569,11 @@ public boolean isMediaScanDisabled() {
     return mediaScanDisabled;
   }
 
+  /**
+   * @return {@code true} if paranoid durability should be enabled for PUTs in this container.
+   */
+  public boolean isParanoidDurabilityEnabled() { return paranoidDurabilityEnabled; }
```
justinlin-linkedin, line 575: please reformat this file.

justinlin-linkedin (approved):

Arun-LinkedIn (approved):

ambry-api/src/main/java/com/github/ambry/account/Container.java
litingulfs: Thanks, that is now done.

PR #2808, opened 2024-06-24
[vcr-2.0] Thread-local metadata cache test

snalli:
Adding a test to enable thread-local-metadata cache in VCR cluster.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobMetadata.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobMetadata.java
@@ -505,6 +505,9 @@ public class CloudBlobMetadata {
     if (!(o instanceof CloudBlobMetadata)) {
       return false;
     }
+    if (o == this) {
+      return true;
+    }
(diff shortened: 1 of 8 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobMetadata.java
```diff
@@ -505,6 +505,9 @@ public boolean equals(Object o) {
     if (!(o instanceof CloudBlobMetadata)) {
       return false;
     }
+    if (o == this) {
+      return true;
+    }
```
justinlin-linkedin, line 510: this should return false.

snalli: hmm if we are comparing an object to itself, the equals() should be true right ?
o.equals(o) == true ?

justinlin-linkedin: oh, missed that

justinlin-linkedin (approved):

PR #2810, opened 2024-06-25
Simple operation tracker bug fix for delete and ttlupdate for unwante…

vaibhav-mittal-linkedin:
### About
In SimpleOperationTracker we have an optimization for not hitting last replica if we are sure that we won't be able to meet the required quorum even if we get the response from that last replica. Now consider a scenario:-  if a blob is ***already*** deleted and client sends request ***again*** for deleting the blob, now if two of the replicas in originating dc are in Bootstrap then when we send final response to client we would send 503 because we considers NOT_FOUND response from Bootstrap replicas as UNAVAILABLE. Now we already know once those replicas are in STANDBY they will not have blob.

This PR will make sure that we hit all replicas before sending response to client for cases like above, and send NOT_FOUND(404) response if all replicas including Bootstrap doesn't has the blob.

### Testing
I have added unit test to test this change for Delete and Tllupdate for case when blob is already deleted and we are querying again.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/SimpleOperationTracker.java
@@ -681,11 +681,8 @@ class SimpleOperationTracker implements OperationTracker {
       if (routerConfig.routerOperationTrackerTerminateOnNotFoundEnabled && hasFailedOnNotFound()) {
         return true;
       }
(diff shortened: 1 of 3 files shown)


justinlin-linkedin (commented):
looking good, but you have to fix other failed unit tests.

justinlin-linkedin (approved):
LGTM

PR #2811, opened 2024-06-25
[Paranoid Durability] Add metrics and logs for paranoid durability writes.

litingulfs:
This PR adds more log statements and metrics to help in evaluating and debugging the performance of the paranoid durability write feature:

* A single metric for when not enough successful writes can be made to satisfy the requirements (i.e. not enough local or remote successful writes).
  * Log messages with additional information to assist in debugging.
* A metric for when there are not enough replicas in total to attempt a paranoid durability write (log messages already exist).


justinlin-linkedin (approved):
LGTM

PR #2814, opened 2024-06-26
Add new version of DeleteRequest format

Arun-LinkedIn:
This version contains flag to indicate if we need to insert a delete tombstone even if blob is missing on servers

### DIFF
--- a/ambry-protocol/src/main/java/com/github/ambry/protocol/DeleteRequest.java
+++ b/ambry-protocol/src/main/java/com/github/ambry/protocol/DeleteRequest.java
@@ -18,8 +18,6 @@ import com.github.ambry.commons.BlobId;
 import com.github.ambry.utils.Utils;
 import java.io.DataInputStream;
 import java.io.IOException;
-import java.nio.ByteBuffer;
-import java.nio.channels.WritableByteChannel;
 
 
 /**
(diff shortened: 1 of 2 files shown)


ambry-protocol/src/main/java/com/github/ambry/protocol/DeleteRequest.java
```diff
@@ -147,7 +178,18 @@ static DeleteRequest readFrom(DataInputStream stream, ClusterMap map) throws IOE
...
+  }
+
+  private static class DeleteRequest_V3 {
+    static DeleteRequest readFrom(DataInputStream stream, ClusterMap map) throws IOException {
+      int correlationId = stream.readInt();
+      String clientId = Utils.readIntString(stream);
+      BlobId id = new BlobId(stream, map);
+      long deletionTimeInMs = stream.readLong();
+      boolean isForceDelete = stream.readByte() == 1;
+      return new DeleteRequest(correlationId, clientId, id, deletionTimeInMs, isForceDelete);
```
justinlin-linkedin, line 192: Should add VERSION_3 here. This constructor would use version 2

Arun-LinkedIn: Done

justinlin-linkedin (approved):

PR #2815, opened 2024-06-27
Add force delete operation on server

Arun-LinkedIn:
Changes to place a delete tombstone on server even if blob is not present

### DIFF
--- a/ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
+++ b/ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
@@ -456,21 +456,8 @@ public class AmbryRequests implements RequestAPI {
       }
     } catch (StoreException e) {
       if (e.getErrorCode() == StoreErrorCodes.ID_Not_Found && deleteRequest.shouldForceDelete()) {
-        try {
-          // If frontend forces a delete operation, place a tombstone even though blob is not present
-          Objects.requireNonNull(storeToDelete).forceDelete(Collections.singletonList(info));
-        } catch (StoreException ex) {
-          if (ex.getErrorCode() == StoreErrorCodes.Already_Exist) {
-            // Blob was not present originally but was replicated while we were force deleting it. Try normal delete now
-            try {
-              Objects.requireNonNull(storeToDelete).delete(Collections.singletonList(info));
-            } catch (StoreException exc) {
-              serverErrorCode = ErrorMapping.getStoreErrorMapping(e.getErrorCode());
-            }
-          } else {
-            serverErrorCode = ErrorMapping.getStoreErrorMapping(e.getErrorCode());
-          }
-        }
+        // If frontend forces a delete operation, place a tombstone even though blob is not present
+        serverErrorCode = maybeForceDelete(info, storeToDelete);
       } else {
         serverErrorCode = ErrorMapping.getStoreErrorMapping(e.getErrorCode());
       }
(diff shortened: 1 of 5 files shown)


ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -435,64 +434,86 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
-      } else if (e.getErrorCode() == StoreErrorCodes.TTL_Expired) {
-        metrics.ttlExpiredError.inc();
-      } else if (e.getErrorCode() == StoreErrorCodes.ID_Deleted) {
-        metrics.idDeletedError.inc();
-      } else if (e.getErrorCode() == StoreErrorCodes.Authorization_Failure) {
-        metrics.deleteAuthorizationFailure.inc();
+      if (e.getErrorCode() == StoreErrorCodes.ID_Not_Found && deleteRequest.shouldForceDelete()) {
+        try {
+          // If frontend forces a delete operation, place a tombstone even though blob is not present
+          Objects.requireNonNull(storeToDelete).forceDelete(Collections.singletonList(info));
```
justinlin-linkedin, line 461: if we are here, the `storeToDelete` will not be null.

justinlin-linkedin, line 461: one thing about the `forceDelete` is the the `MessageInfo` passed to this method can't have -1 lifeVersion. It has to be something meaningful. We have to change the MessageInfo to have 0 as lifeVersion.

Can you also add a comment to forceDelete method?

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -435,64 +434,86 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
-      boolean logInErrorLevel = false;
-      if (e.getErrorCode() == StoreErrorCodes.ID_Not_Found) {
-        metrics.idNotFoundError.inc();
-      } else if (e.getErrorCode() == StoreErrorCodes.TTL_Expired) {
-        metrics.ttlExpiredError.inc();
-      } else if (e.getErrorCode() == StoreErrorCodes.ID_Deleted) {
-        metrics.idDeletedError.inc();
-      } else if (e.getErrorCode() == StoreErrorCodes.Authorization_Failure) {
-        metrics.deleteAuthorizationFailure.inc();
+      if (e.getErrorCode() == StoreErrorCodes.ID_Not_Found && deleteRequest.shouldForceDelete()) {
```
justinlin-linkedin, line 458: can we create a new method for this logic, something like `maybeForceDelete`.

justinlin-linkedin (commented):
most parts LGTM. could you also add some comments.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
Arun-LinkedIn: Sure. Created a new method for this logic.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
Arun-LinkedIn: Yeah.. true.

Arun-LinkedIn: Sure. Added comments to the forceDelete method that the lifeVersion of input should be a valid value.

justinlin-linkedin (commented):
LGTM, please add some test cases for the new code.

justinlin-linkedin (approved):

PR #2817, opened 2024-07-01
changing storeKeyconverter logic to count based logic

manbearpig1996:
Added tryDropCache for keys in StoreKeyConverter to try and drop some keys.
Adding an AbstractStoreKeyConverter and implementing in current test classes of StoreKeyConverter
to have same expected logic everywhere.
AbstractStoreKeyConverter is using a counting cache logic for keys.
Whenever a key is added its count will keep increasing, when tryDropCache is called for keys, count will decrease.
Keys will be dropped, when count becomes 0.
Rest of the behviour is same as expected

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/store/AbstractStoreKeyConverter.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/AbstractStoreKeyConverter.java
@@ -1,0 +1,72 @@
+package com.github.ambry.store;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+
+public abstract class AbstractStoreKeyConverter implements StoreKeyConverter {
+  private final Map<StoreKey, StoreKey> conversionCache = new HashMap<>();
+  private final Map<StoreKey, Integer> cachedKeyCount = new HashMap<>();
+
+  @Override
+  public Map<StoreKey, StoreKey> convert(Collection<? extends StoreKey> input) throws Exception {
+    List<StoreKey> alreadyPresentStoreKeys = new ArrayList<>();
+    List<StoreKey> storeKeysTobeConverted = new ArrayList<>();
+
+    input.forEach((storeKey) -> {
+      if (conversionCache.containsKey(storeKey)) {
+        alreadyPresentStoreKeys.add(storeKey);
+      } else {
+        storeKeysTobeConverted.add(storeKey);
+      }
+    });
+
+    Map<StoreKey, StoreKey> map = convertKeys(storeKeysTobeConverted);
+
+    alreadyPresentStoreKeys.forEach((storeKey -> {
+      map.put(storeKey, conversionCache.get(storeKey));
+    }));
+
+    map.forEach((storeKey, convertedStoreKey) -> {
+      cachedKeyCount.put(storeKey, cachedKeyCount.getOrDefault(storeKey, 0) + 1);
+    });
+
+    conversionCache.putAll(map);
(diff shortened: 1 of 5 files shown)


justinlin-linkedin (approved):
LGTM

ambry-commons/src/test/java/com/github/ambry/store/AbstractStoreKeyConverterTest.java
```diff
@@ -0,0 +1,128 @@
...
+    StoreKey storeKey1 = mock(StoreKey.class);
+    assertNotSame("storeKeys should not be equal", storeKey0, storeKey1);
+    List<StoreKey> list = Arrays.asList(storeKey0, storeKey1);
+
+    try {
+      storeKeyConverter.getConverted(storeKey0);
+      fail("Get succeeded for not converted keys");
+    } catch (Exception e) {
+      //expected
+    }
```
mudit-saxena, line 71: Can you out this as a separate test method if this is expected to throw exception? Also, preferably use `assertThrows` or `@Test(expected = ExceptionClass.class)`

ambry-commons/src/test/java/com/github/ambry/store/AbstractStoreKeyConverterTest.java
```diff
@@ -0,0 +1,128 @@
...
+
+    Map<StoreKey, StoreKey> convertedMap = storeKeyConverter.convert(list);
+    assertEquals("Size of keys in input and output is different", list.size(), convertedMap.size());
+
+    try {
+      StoreKey storeKey0Res = storeKeyConverter.getConverted(storeKey0);
+      assertNotNull("Conversion of storeKey0 should not be null", storeKey0Res);
+    } catch (Exception e) {
+      fail("Exception thrown while storeKey0 is present");
```
mudit-saxena, line 80: If you expect test failure when an exception is thrown you can avoid catching it

ambry-api/src/main/java/com/github/ambry/store/AbstractStoreKeyConverter.java
```diff
@@ -0,0 +1,124 @@
...
+  protected abstract Map<StoreKey, StoreKey> convertKeys(Collection<? extends StoreKey> input) throws Exception;
+
+  /**
+   * This methods response will be returned from {@link #getConverted(StoreKey)} method.
+   * @param storeKey  StoreKey that was queried
+   * @param isKeyPresent whether key is present in cache
+   * @param cachedMapping convertedMapping that is stored in cache. Will be null if if key is not present in cache.
+   * @return
+   */
+  protected abstract StoreKey getConvertedKey(StoreKey storeKey, Boolean isKeyPresent, StoreKey cachedMapping);
```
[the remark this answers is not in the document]
manbearpig1996: Changed it.

ambry-commons/src/test/java/com/github/ambry/store/AbstractStoreKeyConverterTest.java
manbearpig1996: Need to have a test where first we are throwing exception then we are not and then we are again throwing and so on. So need to write code in a single place this way.
assertThrows is not present in junit4 that we are using in ambry-oss.

ambry-commons/src/test/java/com/github/ambry/store/AbstractStoreKeyConverterTest.java
manbearpig1996: Need to have a test where first we are throwing exception then we are not and then we are again throwing and so on. So need to write code in a single place this way.

mudit-saxena (approved):
GTG

PR #2818, opened 2024-07-02
Adding metric for fastest and slowest replica in a cycle

manbearpig1996:
Adding metric for fastest and slowest replica in a cycle


justinlin-linkedin (approved):
LGTM

PR #2819, opened 2024-07-04
Changing replication logic to not wait for replicas to finish

manbearpig1996:
Changing the logic of replication cycle. Currently in a replication cycle, we will create groups and then wait for all groups to finish. With this change, we all replica groups will run multiple iterations until any group reaches a predefined limit.
Also added logic to terminate cycle, when  we are adding/removing replicas or shutting the thread down. 

Added tests for group generation, active replication & leader based replication.

If we set limit to 1, then the new logic will work exactly as same as old logic.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -104,11 +104,11 @@ public class ReplicationConfig {
   /**
    *
    */
-  @Config("replicationIntraMaxIterationsPerReplicationGroupPerCycle")
+  @Config("replication.intra.max.iterations.per.replication.group.per.cycle")
   @Default("1")
   public final int replicationIntraMaxIterationsPerReplicationGroupPerCycle;
 
-  @Config("replicationInterMaxIterationsPerReplicationGroupPerCycle")
+  @Config("replication.inter.max.iterations.per.replication.group.per.cycle")
   @Default("1")
   public final int replicationInterMaxIterationsPerReplicationGroupPerCycle;
 
@@ -355,9 +355,9 @@ public class ReplicationConfig {
         verifiableProperties.getLongInRange("replication.inter.replica.thread.throttle.sleep.duration.ms", 0, 0,
             Long.MAX_VALUE);
     replicationIntraMaxIterationsPerReplicationGroupPerCycle =
-        verifiableProperties.getIntInRange("replication.intra.max.iterations.per.replication.group.per.cycle", 1, 1, Integer.MAX_VALUE);
+        verifiableProperties.getIntInRange("replication.intra.max.iterations.per.replication.group.per.cycle", 1, 1, 100000);
     replicationInterMaxIterationsPerReplicationGroupPerCycle =
-        verifiableProperties.getIntInRange("replication.inter.max.iterations.per.replication.group.per.cycle", 1, 1, Integer.MAX_VALUE);
+        verifiableProperties.getIntInRange("replication.inter.max.iterations.per.replication.group.per.cycle", 1, 1, 100000);
     replicationReplicaThreadIdleSleepDurationMs =
         verifiableProperties.getLongInRange("replication.replica.thread.idle.sleep.duration.ms", 0, 0, Long.MAX_VALUE);
     replicationSyncedReplicaBackoffDurationMs =
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -102,12 +102,17 @@ public class ReplicationConfig {
   public final long replicationInterReplicaThreadThrottleSleepDurationMs;
 
   /**
-   *
+   * The max number of times a replication group will be considered for iteration in a cycle
+   * in an intra colo replica thread
    */
   @Config("replication.intra.max.iterations.per.replication.group.per.cycle")
   @Default("1")
   public final int replicationIntraMaxIterationsPerReplicationGroupPerCycle;
 
+  /**
+   * The max number of times a replication group will be considered for iteration in a cycle
+   * in a cross colo replica thread
+   */
   @Config("replication.inter.max.iterations.per.replication.group.per.cycle")
   @Default("1")
   public final int replicationInterMaxIterationsPerReplicationGroupPerCycle;
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -101,9 +101,9 @@ public class ReplicationConfig {
   @Default("0")
   public final long replicationInterReplicaThreadThrottleSleepDurationMs;
 
-  @Config("replication.enable.acyclic.replication")
+  @Config("replication.enable.continuous.replication")
   @Default("true")
-  public final boolean replicationEnableAcyclicReplication;
+  public final boolean replicationEnableContinuousReplication;
   /**
    * The max number of times a replication group will be considered for iteration in a cycle
    * in an intra colo replica thread
@@ -362,7 +362,7 @@ public class ReplicationConfig {
     replicationInterReplicaThreadThrottleSleepDurationMs =
         verifiableProperties.getLongInRange("replication.inter.replica.thread.throttle.sleep.duration.ms", 0, 0,
             Long.MAX_VALUE);
-    replicationEnableAcyclicReplication = verifiableProperties.getBoolean("replication.enable.acyclic.replication", true);
+    replicationEnableContinuousReplication = verifiableProperties.getBoolean("replication.enable.continuous.replication", false);
     replicationIntraMaxIterationsPerReplicationGroupPerCycle =
         verifiableProperties.getIntInRange("replication.intra.max.iterations.per.replication.group.per.cycle", 1, 1, 100000);
     replicationInterMaxIterationsPerReplicationGroupPerCycle =
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -102,7 +102,7 @@ public class ReplicationConfig {
   public final long replicationInterReplicaThreadThrottleSleepDurationMs;
 
   @Config("replication.enable.acyclic.replication")
-  @Default("false")
+  @Default("true")
   public final boolean replicationEnableAcyclicReplication;
   /**
    * The max number of times a replication group will be considered for iteration in a cycle
@@ -362,7 +362,7 @@ public class ReplicationConfig {
     replicationInterReplicaThreadThrottleSleepDurationMs =
         verifiableProperties.getLongInRange("replication.inter.replica.thread.throttle.sleep.duration.ms", 0, 0,
             Long.MAX_VALUE);
-    replicationEnableAcyclicReplication = verifiableProperties.getBoolean("replication.enable.acyclic.replication", false);
+    replicationEnableAcyclicReplication = verifiableProperties.getBoolean("replication.enable.acyclic.replication", true);
     replicationIntraMaxIterationsPerReplicationGroupPerCycle =
         verifiableProperties.getIntInRange("replication.intra.max.iterations.per.replication.group.per.cycle", 1, 1, 100000);
     replicationInterMaxIterationsPerReplicationGroupPerCycle =
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -101,6 +101,9 @@ public class ReplicationConfig {
   @Default("0")
   public final long replicationInterReplicaThreadThrottleSleepDurationMs;
 
+  @Config("replication.enable.acyclic.replication")
+  @Default("false")
+  public final boolean replicationEnableAcyclicReplication;
   /**
    * The max number of times a replication group will be considered for iteration in a cycle
    * in an intra colo replica thread
@@ -359,6 +362,7 @@ public class ReplicationConfig {
     replicationInterReplicaThreadThrottleSleepDurationMs =
         verifiableProperties.getLongInRange("replication.inter.replica.thread.throttle.sleep.duration.ms", 0, 0,
             Long.MAX_VALUE);
+    replicationEnableAcyclicReplication = verifiableProperties.getBoolean("replication.enable.acyclic.replication", false);
     replicationIntraMaxIterationsPerReplicationGroupPerCycle =
         verifiableProperties.getIntInRange("replication.intra.max.iterations.per.replication.group.per.cycle", 1, 1, 100000);
     replicationInterMaxIterationsPerReplicationGroupPerCycle =
--- a/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
@@ -102,6 +102,17 @@ public class ReplicationConfig {
   public final long replicationInterReplicaThreadThrottleSleepDurationMs;
 
   /**
+   *
+   */
+  @Config("replicationIntraMaxIterationsPerReplicationGroupPerCycle")
+  @Default("1")
+  public final int replicationIntraMaxIterationsPerReplicationGroupPerCycle;
+
+  @Config("replicationInterMaxIterationsPerReplicationGroupPerCycle")
+  @Default("1")
+  public final int replicationInterMaxIterationsPerReplicationGroupPerCycle;
+
+  /**
    * The time (in ms) to sleep between replication cycles when the replica thread is not doing any useful work
    */
   @Config("replication.replica.thread.idle.sleep.duration.ms")
@@ -343,6 +354,10 @@ public class ReplicationConfig {
     replicationInterReplicaThreadThrottleSleepDurationMs =
         verifiableProperties.getLongInRange("replication.inter.replica.thread.throttle.sleep.duration.ms", 0, 0,
             Long.MAX_VALUE);
+    replicationIntraMaxIterationsPerReplicationGroupPerCycle =
+        verifiableProperties.getIntInRange("replication.intra.max.iterations.per.replication.group.per.cycle", 1, 1, Integer.MAX_VALUE);
+    replicationInterMaxIterationsPerReplicationGroupPerCycle =
+        verifiableProperties.getIntInRange("replication.inter.max.iterations.per.replication.group.per.cycle", 1, 1, Integer.MAX_VALUE);
     replicationReplicaThreadIdleSleepDurationMs =
         verifiableProperties.getLongInRange("replication.replica.thread.idle.sleep.duration.ms", 0, 0, Long.MAX_VALUE);
     replicationSyncedReplicaBackoffDurationMs =
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -402,11 +402,13 @@ public class ReplicaThread implements Runnable {
       //for each of smaller array of remote replicas update with current group id and add to groupId map
       for (List<RemoteReplicaInfo> remoteReplicaInfos : remoteReplicaSegregatedList) {
         groupIdToRemoteReplicaMap.put(groupId, remoteReplicaInfos);
+        logger.trace("Thread name: {} Assigned groupId {} to replicas {}", threadName, groupId, remoteReplicaInfos);
         groupId++;
       }
 
       //now, for this data node add one stand by group id for standby remote replicas
       standByNoProgressGroupId.put(remoteHost, groupId);
+      logger.trace("Thread name: {} Assigned groupId for standBy groups {} to remoteHost {}",threadName, groupId, remoteHost);
       groupId++;
     }
   }
@@ -432,9 +434,12 @@ public class ReplicaThread implements Runnable {
         long throttledTill = currentTime + threadThrottleDurationMs;
         remoteReplicaGroup.remoteReplicaInfos.forEach(remoteReplicaInfo -> {
           remoteReplicaToThrottledTill.put(remoteReplicaInfo, throttledTill);
+          logger.trace("Thread name: {} Remote Replica {} is throttled till {} in replica", threadName,
+              remoteReplicaInfo, throttledTill);
         });
         storeKeyConverter.tryDropCache(remoteReplicaGroup.getStoreKeysConversionLog());
         inflightRemoteReplicaGroupIterator.remove();
+        logger.trace("Thread name: {} Group Id {} removed from inflight groups", threadName, entry.getKey());
       }
     }
   }
@@ -457,6 +462,9 @@ public class ReplicaThread implements Runnable {
 
     //check if thread is in shutdown or cycle termination is received
     if (!running || shouldTerminateCurrentCycle) {
+      logger.trace(
+          "Thread name: {} not creating new groups as thread is in shutdown mode {} or terminate cycle received {}",
+          threadName, !running, shouldTerminateCurrentCycle);
       return false;
     }
 
@@ -472,8 +480,11 @@ public class ReplicaThread implements Runnable {
         maxIterationsReached.setTrue();
         if (inflightRemoteReplicaGroup.isEmpty()) {
           allReplicasCaughtUpEarly.setTrue();
+          logger.trace("Thread name: {} All replicas marked to caught up early", threadName);
         }
       }
+      logger.trace("Thread name: {} not creating new groups as iteration limit {} reached", threadName,
+          maxIterationsPerGroupPerCycle);
       return false;
     }
     return true;
@@ -507,11 +518,13 @@ public class ReplicaThread implements Runnable {
 
     //remove all replicas of corresponding group id if group is in iteration
     if(inflightRemoteReplicaGroup.containsKey(groupId)){
+      logger.trace("Thread name: {} group {} is in iteration", threadName, groupId);
       nonIteratingNonPendingRemoteReplicas.removeAll(inflightRemoteReplicaGroup.get(groupId).remoteReplicaInfos);
     }
 
     //remove all replicas of corresponding standby group if group is in iteration
     if(inflightRemoteReplicaGroup.containsKey(standbyReplicasNoProgressReplicaGroupId)) {
+      logger.trace("Thread name: {} stand by group {} is in iteration", threadName, standbyReplicasNoProgressReplicaGroupId);
       nonIteratingNonPendingRemoteReplicas.removeAll(inflightRemoteReplicaGroup.get(standbyReplicasNoProgressReplicaGroupId).remoteReplicaInfos);
     }
 
@@ -546,16 +559,19 @@ public class ReplicaThread implements Runnable {
 
     //check and process finished groups
     processFinishedGroups(inflightRemoteReplicaGroup, remoteReplicaToThrottledTill);
-
+    logger.trace("Thread name: {} Finished processing finished groups", threadName);
     //check if we should create new groups, return other wise
     if (!shouldCreateNewRemoteReplicaGroups(inflightRemoteReplicaGroup, groupIdIterationCountMap,
         allReplicasCaughtUpEarly, maxIterationsReached)) {
+      logger.trace("Thread name: {} Not creating any new groups", threadName);
       return;
     }
 
+    logger.trace("Thread name: {} Trying to create new active groups", threadName);
     //now we create all active groups and add stand by no progress replicas to queue to be processed later
     for (Map.Entry<Integer, List<RemoteReplicaInfo>> entry : groupIdToRemoteReplicaMap.entrySet()) {
       Integer groupId = entry.getKey();
+      logger.trace("Thread name: {} Processing for group id {}", threadName, groupId);
       List<RemoteReplicaInfo> remoteReplicas = entry.getValue();
       DataNodeId remoteNode = remoteReplicas.get(0).getReplicaId().getDataNodeId();
 
@@ -568,6 +584,8 @@ public class ReplicaThread implements Runnable {
       //generate active and standby replicas with no progress
       getActiveAndStandByNoProgressReplicas(activeReplicas, standbyReplicasWithNoProgress, remoteReplicas,
           remoteReplicaToThrottledTill, standbyReplicasNoProgressReplicaGroupId, groupId, inflightRemoteReplicaGroup);
+      logger.trace("Thread name: {} Active replicas are {}, standby Replicas with no progress are {}", threadName,
+          activeReplicas, standbyReplicasWithNoProgress);
 
       //if the current group is not in iteration, and there are active replicas, we will start a new group
       //and increase iteration count if we are able to create
@@ -576,6 +594,8 @@ public class ReplicaThread implements Runnable {
           RemoteReplicaGroup groupForActiveReplicas = new RemoteReplicaGroup(activeReplicas, remoteNode, false, groupId);
           inflightRemoteReplicaGroup.put(groupId, groupForActiveReplicas);
           groupIdIterationCountMap.put(groupId, groupIdIterationCountMap.getOrDefault(groupId, 0) + 1);
+          logger.trace("Thread name: {} Created new group for groupId {}, iteration count is {}", threadName, groupId,
+              groupIdIterationCountMap.getOrDefault(groupId, 0));
         }
       }
 
@@ -590,12 +610,16 @@ public class ReplicaThread implements Runnable {
       //add timed out stand by replicas with no progress to queue
       standByNoProgressReplicaQueue.putIfAbsent(remoteNode, new HashSet<>());
       standByNoProgressReplicaQueue.get(remoteNode).addAll(standByReplicasTimedOutOnNoProgress);
+      logger.trace("Thread name: {} Added standby no progress replicas {} to queue", threadName,
+          standByReplicasTimedOutOnNoProgress);
     }
 
+    logger.trace("Thread name: {} Trying to create new standby groups", threadName);
     //iterate over all data nodes and stand by no progress timed out replica queue and create groups
     for(Map.Entry<DataNodeId, Set<RemoteReplicaInfo>> entry : standByNoProgressReplicaQueue.entrySet()){
       DataNodeId remoteNode = entry.getKey();
       Integer groupId = standByNoProgressGroupIds.get(remoteNode);
+      logger.trace("Thread name: {} Trying for remoteNode {}, standBy groupId {}}", threadName, remoteNode, groupId);
 
       //if group is already in iteration, do nothing
       if(inflightRemoteReplicaGroup.containsKey(groupId)) {
@@ -611,6 +635,8 @@ public class ReplicaThread implements Runnable {
       inflightRemoteReplicaGroup.put(groupId, remoteReplicaGroup);
       //increase iteration count
       groupIdIterationCountMap.put(groupId, groupIdIterationCountMap.getOrDefault(groupId, 0) + 1);
+      logger.trace("Thread name: {} Create standBy group with id {} ,iteration count {}", threadName, groupId,
+          groupIdIterationCountMap.getOrDefault(groupId, 0));
 
       //clear the queue as all replicas are added in the group
       standByNoProgressReplicaQueue.get(remoteNode).clear();
@@ -619,6 +645,7 @@ public class ReplicaThread implements Runnable {
     //if we are not able to create any new group and no groups are remaining, all replicas caught up early
     if (inflightRemoteReplicaGroup.isEmpty()) {
       allReplicasCaughtUpEarly.setTrue();
+      logger.trace("Thread name: {} All replicas marked to caught up early", threadName);
     }
   }
 
@@ -626,10 +653,13 @@ public class ReplicaThread implements Runnable {
    * Checks if acyclic replication is enabled and calls appropriate logic for replication
    */
   public void replicate(){
-    if(enableAcyclicReplication)
+    if (enableAcyclicReplication) {
+      logger.trace("Thread name: {}, starting acyclic replication", threadName);
       replicateAcyclic();
-    else
+    } else {
+      logger.trace("Thread name: {}, starting cyclic replication", threadName);
       replicateCyclic();
+    }
   }
 
   /**
@@ -722,6 +752,8 @@ public class ReplicaThread implements Runnable {
         }
         final int pollTimeoutMs = (int) replicationConfig.replicationRequestNetworkPollTimeoutMs;
         List<ResponseInfo> responseInfos = networkClient.sendAndPoll(requestInfos, requestsToDrop, pollTimeoutMs);
+        logger.trace("Got {} responses for replica groups from network client in replicaThread {}",
+            responseInfos.size(), threadName);
         // Add response for dropped request because there is no guarantee that sendAndPoll would return a response for
         // dropped requests. Even if the network client returns response infos for dropped requests, onResponse should
         // still be able to handle duplicate response infos for the same request.
@@ -797,6 +829,7 @@ public class ReplicaThread implements Runnable {
 
       // Generate group ids and stand by group ids, only these group ids will be used
       generateGroupIdsForReplicas(groupIdToRemoteReplicaMap, standByNoProgressGroupId);
+      logger.trace("Thread name: {} Generated group ids for replication", threadName);
 
       Map<Integer, RemoteReplicaGroup> inflightRemoteReplicaGroups = new HashMap<>();
       Map<Integer, Integer> groupIdIterationCount = new HashMap<>();
@@ -814,15 +847,19 @@ public class ReplicaThread implements Runnable {
         generateRemoteReplicaGroups(groupIdToRemoteReplicaMap, standByNoProgressGroupId,
             inflightRemoteReplicaGroups, standByNoProgressReplicaQueue,
             remoteReplicaToThrottledTill, groupIdIterationCount, allReplicasCaughtUpEarly, maxIterationsReached);
+        logger.trace("Thread name: {} Remote replica groups {} generated", threadName,
+            inflightRemoteReplicaGroups.keySet());
 
         //get request infos from all inflight remote replica groups
         List<RequestInfo> requestInfos =
             pollRemoteReplicaGroups(new ArrayList<>(inflightRemoteReplicaGroups.values()), correlationIdToRequestInfo,
                 correlationIdToReplicaGroup);
+        logger.trace("Thread name: {} Requests generated", threadName);
 
         //generate responses for requests that are timed out
         List<ResponseInfo> responseInfosForTimedOutRequests =
             filterTimedOutRequests(correlationIdToRequestInfo, correlationIdToReplicaGroup);
+        logger.trace("Thread name: {} Filtered timed out Requests", threadName);
 
         //mark timed out requests to be dropped by network client
         Set<Integer> requestsToDrop = responseInfosForTimedOutRequests.stream()
@@ -838,6 +875,7 @@ public class ReplicaThread implements Runnable {
 
         //submit requests and get response for earlier submitted requests
         List<ResponseInfo> responseInfos = networkClient.sendAndPoll(requestInfos, requestsToDrop, pollTimeoutMs);
+        logger.trace("Thread name: {} received {} responses from network client", threadName, responseInfos.size());
         responseInfos.addAll(responseInfosForTimedOutRequests);
         //process responses
         onResponses(responseInfos, correlationIdToRequestInfo, correlationIdToReplicaGroup);
@@ -861,9 +899,9 @@ public class ReplicaThread implements Runnable {
             .filter(g -> g.isDone() && g.getException() != null)
             .forEach(g -> logger.error("Remote node: {} Thread name: {} RemoteReplicaGroup {} has exception {}",
                 g.getRemoteDataNode(), threadName, g.getRemoteReplicaInfos(), g.getException()));
-      } while (!inflightRemoteReplicaGroups.isEmpty());
 
-      logger.trace("Thread name: {} Finish all RemoteReplicaGroup replication", threadName);
+        logger.trace("Thread name: {} Finish one iteration for RemoteReplicaGroup replication", threadName);
+      } while (!inflightRemoteReplicaGroups.isEmpty());
     }catch (Exception e){
       logger.error("Thread name: {} found some error while replicating from remote hosts", threadName, e);
     }finally {
@@ -875,6 +913,7 @@ public class ReplicaThread implements Runnable {
     int allProcessedReplicasCount = groupIdToRemoteReplicaMap.values().stream().mapToInt(List::size).sum();
     //check and make thread sleep and publish metrics for throttling
     maybeSleepAfterReplication(allReplicasCaughtUpEarly.isTrue(), allProcessedReplicasCount);
+    logger.trace("Thread name: {} Exiting replication, all iterations Done!", threadName);
   }
 
   /**
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -403,15 +403,18 @@ public class ReplicaThread implements Runnable {
   }
 
   /**
-   * Generates groups
-   * @param groupIdToRemoteReplicaMap {@link Map}
-   * @param dataNodeIdStandbyReplicasNoProgressReplicaGroupId {@link Map}
-   * @param inflightRemoteReplicaGroup {@link Map}
-   * @param dataNodeIdToPendingReplicasMarkedForStandByNoProgress {@link Map}
-   * @param remoteReplicaToThrottledTill {@link Map}
-   * @param groupIdIterationCountMap {@link Map}
-   * @param allReplicasCaughtUpEarly {@link MutableBoolean}
-   * @param maxIterationsReached {@link MutableBoolean}
+   * Generates groups and standby groups and adds these to inflightRemoteReplicaGroup. Removes groups that are done from the
+   * inflightRemoteReplicaGroup. It will generate new groups until shouldTerminateCurrentCycle is false running is true and
+   * maxIterationsPerGroupPerCycle limit is not reached by any iteration count of any group. Replicas will only be added
+   * in the group in groupIdToRemoteReplicaMap and for stand by group in dataNodeIdStandbyReplicasNoProgressReplicaGroupId.
+   * @param groupIdToRemoteReplicaMap {@link Map} mapping of group id to remote replicas
+   * @param dataNodeIdStandbyReplicasNoProgressReplicaGroupId {@link Map} mapping of data node id to standby group id
+   * @param inflightRemoteReplicaGroup {@link Map} map of inflight group id to remote replica groups
+   * @param dataNodeIdToPendingReplicasMarkedForStandByNoProgress {@link Map} map of data node id to standby replicas to pick up
+   * @param remoteReplicaToThrottledTill {@link Map} map of remote replica to time until which it is throttled
+   * @param groupIdIterationCountMap {@link Map} map of group id to number of iterations
+   * @param allReplicasCaughtUpEarly {@link MutableBoolean} stores whether all replicas have caught up early
+   * @param maxIterationsReached {@link MutableBoolean} stores whether max iterations reached
    */
   void generateRemoteReplicaGroups(
       Map<Integer, List<RemoteReplicaInfo>> groupIdToRemoteReplicaMap,
@@ -537,6 +540,9 @@ public class ReplicaThread implements Runnable {
 
   /**
    * Do replication for replicas grouped by {@link DataNodeId}
+   * We will keep trying to generate groups until we do not get any groups by calling generateRemoteReplicaGroups.
+   * We will generate requests and then process responses when received for these groups.
+   *
    * A replication cycle between two replicas involves the following steps:
    *    1. Exchange metadata : fetch the metadata of blobs added to remote replica since the last synchronization point
    *    and filter the ones missing in local store.
@@ -641,7 +647,14 @@ public class ReplicaThread implements Runnable {
     maybeSleepAfterReplication(allReplicasCaughtUpEarly.isTrue(), allProcessedReplicasCount);
   }
 
-
+  /**
+   * We should call this method to convert store keys, This will convert the keys and add the keys which were
+   * converted to the list.
+   * @param storeKeysToConvert store keys to convert
+   * @param storeKeysConversionLog list to store which keys were converted
+   * @return Map
+   * @throws Exception
+   */
   Map<StoreKey, StoreKey> convertStoreKeys(Collection<StoreKey> storeKeysToConvert, List<StoreKey> storeKeysConversionLog)
       throws Exception {
     Map<StoreKey, StoreKey> conversionMap =  this.storeKeyConverter.convert(storeKeysToConvert);
@@ -799,6 +812,7 @@ public class ReplicaThread implements Runnable {
    * Maybe sleep for a while after one round of replication. If all the replicas are caught up and the configuration
    * shows we should sleep, then sleep for a while so we can save some CPU.
    * @param allCaughtUp True when all replicas are caught up.
+   * @param throttledReplicaCount total replicas which are getting throttled
    */
   private void maybeSleepAfterReplication(boolean allCaughtUp, int throttledReplicaCount) {
     long sleepDurationMs = 0;
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -963,8 +963,8 @@ public class ReplicaThread implements Runnable {
         }
       } else {
         replicationMetrics.updateMetadataRequestError(remoteReplicaInfo.getReplicaId());
-        logger.error("Remote node: {} Thread name: {} Remote replica: {} Server error: {}", remoteNode, threadName,
-            remoteReplicaInfo.getReplicaId(), replicaMetadataResponseInfo.getError());
+        logger.error("Remote node: {} Thread name: {} Remote replica: {} Server metadata-response error: {}",
+            remoteNode, threadName, remoteReplicaInfo.getReplicaId(), replicaMetadataResponseInfo.getError());
         exchangeMetadataResponseList.add(new ExchangeMetadataResponse(replicaMetadataResponseInfo.getError()));
       }
 
@@ -1459,7 +1459,7 @@ public class ReplicaThread implements Runnable {
                 }
                 // messageInfo is the Blob final state and the operation time is the blob creation time.
                 // So the applyTtlUpdate will use the creation time as the Ttl operation time.
-                if (messageInfo.isTtlUpdated()) {
+                if (isTtlUpdateNeededAfterPut(messageInfo)) {
                   applyTtlUpdate(messageInfo, remoteReplicaInfo);
                 }
                 StoreKey key = messageInfo.getStoreKey();
@@ -1530,8 +1530,8 @@ public class ReplicaThread implements Runnable {
                 exchangeMetadataResponse.getMissingStoreKeys());
           } else {
             replicationMetrics.updateGetRequestError(remoteReplicaInfo.getReplicaId());
-            logger.error("Remote node: {} Thread name: {} Remote replica: {} Server error: {}", remoteNode, threadName,
-                remoteReplicaInfo.getReplicaId(), partitionResponseInfo.getErrorCode());
+            logger.error("Remote node: {} Thread name: {} Remote replica: {} Server get-response error: {}",
+                remoteNode, threadName, remoteReplicaInfo.getReplicaId(), partitionResponseInfo.getErrorCode());
           }
         }
       }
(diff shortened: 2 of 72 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+        long currentTime = time.milliseconds();
+        long throttledTill = currentTime + threadThrottleDurationMs;
+        remoteReplicaGroup.remoteReplicaInfos.forEach(remoteReplicaInfo -> {
+          remoteReplicaToThrottledTill.put(remoteReplicaInfo, throttledTill);
+        });
+        storeKeyConverter.tryDropCache(remoteReplicaGroup.getStoreKeysConversionLog());
+        inflightRemoteReplicaGroupIterator.remove();
+      }
+    }
+
```
snalli, line 453: too much whitespace here, and can u add some comments with /** .. */ explaining the code. It's hard to follow if one has no idea of replication

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +672,131 @@ public void replicate() {
...
+        inflightRemoteReplicaGroup.values()
+            .stream()
+            .filter(g -> g.isDone() && g.getException() != null)
+            .forEach(g -> logger.error("Remote node: {} Thread name: {} RemoteReplicaGroup {} has exception {}",
+                g.getRemoteDataNode(), threadName, g.getRemoteReplicaInfos(), g.getException()));
+      } while (!inflightRemoteReplicaGroup.isEmpty());
+
+      logger.trace("Thread name: {} Finish all RemoteReplicaGroup replication", threadName);
+    }catch (Exception e){
```
snalli, line 776: formatting

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +672,131 @@ public void replicate() {
...
+      logger.error("Thread name: {} found some error while replicating from remote hosts", threadName, e);
+    }finally {
```
snalli, line 778: formatting

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +672,131 @@ public void replicate() {
...
+      Map<DataNodeId, Set<RemoteReplicaInfo>> dataNodeIdToPendingReplicasMarkedForStandByNoProgress = new HashMap<>();
+      Map<RemoteReplicaInfo, Long> remoteReplicaToThrottledTill = new HashMap<>();
+
+      Map<Integer, RemoteReplicaGroup> correlationIdToReplicaGroup = new HashMap<>();
+      Map<Integer, RequestInfo> correlationIdToRequestInfo = new LinkedHashMap<>();
+
+      MutableBoolean maxIterationsReached = new MutableBoolean(false);
+
+      do {
+        generateRemoteReplicaGroups(groupIdToRemoteReplicaMap, remoteHostToStandbyNoProgressReplicaGroupId,
```
snalli, line 733: are there no more active/standby replicas ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+      if (maxIterationsReached.isFalse()) {
+        maxIterationsReached.setTrue();
+        if (inflightRemoteReplicaGroup.isEmpty()) {
+          allReplicasCaughtUpEarly.setTrue();
+        }
+      }
+
+      return;
+    }
+
```
snalli, line 474: also consider breaking this fn into smaller ones, hard to follow and understand diff from prev logic

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+    for(Map.Entry<DataNodeId, Set<RemoteReplicaInfo>> entry : dataNodeIdToPendingReplicasMarkedForStandByNoProgress.entrySet()){
+      DataNodeId remoteNode = entry.getKey();
+      Integer groupId = dataNodeIdStandbyReplicasNoProgressReplicaGroupId.get(remoteNode);
+
+      if(inflightRemoteReplicaGroup.containsKey(groupId))
+        continue;
+
+
+      List<RemoteReplicaInfo> remoteReplicaInfos = new ArrayList<>(entry.getValue());
+      if(remoteReplicaInfos.isEmpty())
```
snalli, line 532: formatting {}

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+
+      dataNodeIdToPendingReplicasMarkedForStandByNoProgress.putIfAbsent(remoteNode, new HashSet<>());
+      dataNodeIdToPendingReplicasMarkedForStandByNoProgress.get(remoteNode).addAll(standByReplicasTimedOutOnNoProgress);
+    }
+
+    for(Map.Entry<DataNodeId, Set<RemoteReplicaInfo>> entry : dataNodeIdToPendingReplicasMarkedForStandByNoProgress.entrySet()){
+      DataNodeId remoteNode = entry.getKey();
+      Integer groupId = dataNodeIdStandbyReplicasNoProgressReplicaGroupId.get(remoteNode);
+
+      if(inflightRemoteReplicaGroup.containsKey(groupId))
```
snalli, line 527: formatting {}

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+      allReplicasCaughtUpEarly.setTrue();
+    }
+  }
+
+  /**
+   * Checks if acyclic replication is enabled and calls appropriate logic for replication
+   */
+  public void replicate(){
+    if(enableAcyclicReplication)
+      replicateAcyclic();
```
snalli, line 552: its not truly acyclic right ? the name is confusing.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+      }
+
+      List<RemoteReplicaInfo> standByReplicasTimedOutOnNoProgress =
+          getRemoteStandbyReplicasTimedOutOnNoProgress(standbyReplicasWithNoProgress);
+
+      if (standByReplicasTimedOutOnNoProgress.isEmpty()) {
+        continue;
+      }
+
+      dataNodeIdToPendingReplicasMarkedForStandByNoProgress.putIfAbsent(remoteNode, new HashSet<>());
```
snalli, line 519: consider smaller fn names, this is just hard to read

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+    if (inflightRemoteReplicaGroup.isEmpty()) {
+      allReplicasCaughtUpEarly.setTrue();
+    }
+  }
+
+  /**
+   * Checks if acyclic replication is enabled and calls appropriate logic for replication
+   */
+  public void replicate(){
+    if(enableAcyclicReplication)
```
snalli, line 551: Lets try to reduce the number of configs. Eliminate this boolean and switch between the two modes based on the value of maxIter.
if maxIter <= 1 -> old code
if maxIter > 1 -> your code.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+      List<RemoteReplicaInfo> remoteReplicas = entry.getValue();
+      DataNodeId remoteNode = remoteReplicas.get(0).getReplicaId().getDataNodeId();
+
+      Integer standbyReplicasNoProgressReplicaGroupId = dataNodeIdStandbyReplicasNoProgressReplicaGroupId.get(remoteNode);
+
+      Set<RemoteReplicaInfo> nonIteratingNonPendingRemoteReplicas = new LinkedHashSet<>(remoteReplicas);
+
+      remoteReplicas.forEach(remoteReplicaInfo -> {
+        if(remoteReplicaToThrottledTill.containsKey(remoteReplicaInfo) && remoteReplicaToThrottledTill.get(remoteReplicaInfo) > time.milliseconds()){
+          nonIteratingNonPendingRemoteReplicas.remove(remoteReplicaInfo);
```
snalli, line 486: Too big a loop, hard to follow and debug, pls break it down.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +374,186 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+        maxIterationsReached.setTrue();
+        if (inflightRemoteReplicaGroup.isEmpty()) {
+          allReplicasCaughtUpEarly.setTrue();
+        }
+      }
+
+      return;
+    }
+
+    for (Map.Entry<Integer, List<RemoteReplicaInfo>> entry : groupIdToRemoteReplicaMap.entrySet()) {
```
snalli, line 475: Too big for-loop, hard to follow and debug, pls break it down.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
```diff
@@ -101,6 +101,25 @@ public class ReplicationConfig {
...
+  @Default("true")
+  public final boolean replicationEnableAcyclicReplication;
+  /**
+   * The max number of times a replication group will be considered for iteration in a cycle
+   * in an intra colo replica thread
+   */
+  @Config("replication.intra.max.iterations.per.replication.group.per.cycle")
+  @Default("1")
+  public final int replicationIntraMaxIterationsPerReplicationGroupPerCycle;
+
```
aga9900, line 114: Do we plan to handle inter and intra colo replication iterations separately? In what scenarios , these values will be different?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -123,6 +124,9 @@ public class ReplicaThread implements Runnable {
   private final Predicate<MessageInfo> skipPredicate;
   private volatile boolean allDisabled = false;
   private final ReplicationManager.LeaderBasedReplicationAdmin leaderBasedReplicationAdmin;
+  private boolean enableAcyclicReplication;
+  private int maxIterationsPerGroupPerCycle;
+  private volatile boolean shouldTerminateCurrentCycle = false;
```
aga9900, line 129: Can you explain the use of this flag and in what scenarios, the value will be false.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
snalli: +1, let merge the two configs and have one like replication.max.iter.per.cycle

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: there are active and standby replica groups, for both separate id is getting generated. For standby groups there is 1 id per data node to reduce number of cross colo calls, it is same as right now. Both will be added to inflight groups.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: changed the name.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: fixed

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: fixed

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: fixed

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: fixed

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: changed the name to replicateLimitedAcyclic

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: We will deprecate old logic with this config after some more validations. If we set limit to 1 , it will beahve exactly as old code.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
manbearpig1996: I am planning to test both threads separately, so we need both configs currently. We will deprecate these two configs after doing some testing internally.

manbearpig1996: Will merge it to single config as suggested by @snalli after deprecating these 2.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: If this is set true and if current cycle is running, we will not create any new iterations of any group and will wait for existing groups to finish.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
snalli: One way you can test both intra and inter-colo separately is by setting the the num threads of intra and inter-colo separately. Set one to 0 and other to non-zero. Does this work for your testing ?

ambry-vcr/src/test/java/com/github/ambry/vcr/CloudTokenPersistorTest.java
```diff
@@ -169,7 +169,7 @@ public void testPersistToken() throws IOException {
     token =  new StoreFindToken(FindTokenType.Uninitialized, null, null, null, null,
         true, (short) 3, null, null, (short) -1);
     response = new ReplicaThread.ExchangeMetadataResponse(Collections.emptySet(), token,
-            -1, Collections.emptyMap(), new SystemTime());
+        -1, Collections.emptyMap(), new SystemTime());
```
snalli, line 172: can u restore this file from master ? there are no code changes here, just some formatting changes due to different IDE settings..

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+  /**
+   * Generates group ids for active iteration for replicas, all replicas are mapped to one group id only
+   * Generates group ids for all datanodes for standby no progress groups
+   * All group ids are unique
+   * All replicas are mapped to particular group id and each datanode has one stand by remote replica group id
+   * so we make less number of parallel calls to remote hosts
+   * @param groupIdToRemoteReplicaMap {@link Map} this map will be populated with group ids to replicas map
+   * @param standByNoProgressGroupId {@link Map}  this map will be populated with datanode to standby group id
+   */
+  void generateGroupIdsForReplicas(Map<Integer, List<RemoteReplicaInfo>> groupIdToRemoteReplicaMap,
```
snalli, line 389: There are too many maps mapping an int to some object here and takes too much time and effort to mentally keep track of these structures.

What is the goal of mapping an ID to a list of replicas ? 
Is it to organize the replicas of a node into active and standby lists ?
Can you define an abstraction like and possibly pass this around ?
```
class PeerNode {
  List activeReplicas;
  List standByReplicas;
  DataNode node;
  <some more variables>
  getters()
  setters()
}
```

Try to define a function with a return type that tells us instantly what the fn is returning. This fn here doesn't return anything and the return values are embedded in the arglist. Can you separate out the return value from this arglist ?

Try to avoid using mutable-booleans here and define some abstraction to pass around. That abstraction should have all the info easily accessible.

Rename standByNoProgress to just `standby` because that is enough to explain what the replica does - it stands by without any progress on data replication.

Lets not rush to merge this.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+          "Thread name: {} not creating new groups as thread is in shutdown mode {} or terminate cycle received {}",
+          threadName, !running, shouldTerminateCurrentCycle);
+      return false;
+    }
+
+    //get max iterations by any remote replica group
+    int maxIterationCount = 0;
+    for (Map.Entry<Integer, Integer> entry : groupIdIterationCountMap.entrySet()) {
+      maxIterationCount = Integer.max(maxIterationCount, entry.getValue());
+    }
```
justinlin-linkedin, line 479:
```
int maxIterationCount = groupIdIterationMap.values().stream().max(Integer:compare).orElse(0).
```
something like this?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+   * @param remoteReplicas {@link List}
+   * @param remoteReplicaToThrottledTill {@link Map}
+   * @param standbyReplicasNoProgressReplicaGroupId {@link Integer}
+   * @param groupId {@link Integer}
+   * @param inflightRemoteReplicaGroup {@link Map}
+   */
+  private void getActiveAndStandByNoProgressReplicas(List<RemoteReplicaInfo> activeReplicas,
+      List<RemoteReplicaInfo> standbyReplicasWithNoProgress, List<RemoteReplicaInfo> remoteReplicas,
+      Map<RemoteReplicaInfo, Long> remoteReplicaToThrottledTill, Integer standbyReplicasNoProgressReplicaGroupId,
+      Integer groupId, Map<Integer, RemoteReplicaGroup> inflightRemoteReplicaGroup) {
```
justinlin-linkedin, line 511: let's use primitive type for these two integers.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+
+      //if group is already in iteration, do nothing
+      if (inflightRemoteReplicaGroup.containsKey(groupId)) {
+        continue;
+      }
+
+      List<RemoteReplicaInfo> remoteReplicaInfos = new ArrayList<>(entry.getValue());
+      if (remoteReplicaInfos.isEmpty()) {
+        continue;
+      }
```
justinlin-linkedin, line 636: we can combine these two if statement.

justinlin-linkedin (commented):
first round of review, will do another round.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
snalli, line 389: Use shorter names like create or make instead of generate.
If you're using get, then the fn must return something,

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+   * Generates group ids for all datanodes for standby no progress groups
+   * All group ids are unique
+   * All replicas are mapped to particular group id and each datanode has one stand by remote replica group id
+   * so we make less number of parallel calls to remote hosts
+   * @param groupIdToRemoteReplicaMap {@link Map} this map will be populated with group ids to replicas map
+   * @param standByNoProgressGroupId {@link Map}  this map will be populated with datanode to standby group id
+   */
+  void generateGroupIdsForReplicas(Map<Integer, List<RemoteReplicaInfo>> groupIdToRemoteReplicaMap,
+      Map<DataNodeId, Integer> standByNoProgressGroupId) {
+
```
snalli, line 391: rm whitespace

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+    Map<DataNodeId, List<RemoteReplicaInfo>> dataNodeToRemoteReplicaInfo = selectReplicas(getRemoteReplicaInfos());
+
+    int groupId = 0;
+
+    for (Map.Entry<DataNodeId, List<RemoteReplicaInfo>> entry : dataNodeToRemoteReplicaInfo.entrySet()) {
+      DataNodeId remoteHost = entry.getKey();
+      List<RemoteReplicaInfo> remoteReplicasPerNode = entry.getValue();
+
```
snalli, line 399: rm whitespace

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+  /**
+   * Removes finished groups from inflight remote replica groups, drops cached keys from storeKeyConverter
+   * For finished groups adds time until which replicas are throttled
+   * @param inflightRemoteReplicaGroup {@link Map}
+   * @param remoteReplicaToThrottledTill {@link Map}
+   */
+  private void processFinishedGroups(Map<Integer, RemoteReplicaGroup> inflightRemoteReplicaGroup,
+      Map<RemoteReplicaInfo, Long> remoteReplicaToThrottledTill) {
+
```
snalli, line 428: rm whitespace

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+    Iterator<Map.Entry<Integer, RemoteReplicaGroup>> inflightRemoteReplicaGroupIterator =
+        inflightRemoteReplicaGroup.entrySet().iterator();
+
```
snalli, line 431: rm whitespace

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +787,153 @@ public void replicate() {
...
+   *              |   Local Leader    |     Local Standby   |   Remote Leader   |  Remote Standby
+   *            -------------------------------------------------------------------------------------
+   *     Leader:  |        ---        |  metadata and data  | metadata and data |   metadata only
+   *     Standby: | metadata and data |  metadata and data  | metadata only     |   metadata only
+   *
+   */
+  public void replicateLimitedAcyclic() {
+    long oneRoundStartTimeMs = time.milliseconds();
+    shouldTerminateCurrentCycle = false;
+
```
snalli, line 822: rm whitespace

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +787,153 @@ public void replicate() {
...
+    logger.trace("Thread name: {} Start RemoteReplicaGroup replication", threadName);
+
+    exchangeMetadataResponsesInEachCycle = new HashMap<>();
+    Set<Integer> groupsAddedToExchangeMetadatResponse = new HashSet<>();
+
```
snalli, line 827: rm whitespace

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +787,153 @@ public void replicate() {
...
+   *            -------------------------------------------------------------------------------------
+   *     Leader:  |        ---        |  metadata and data  | metadata and data |   metadata only
+   *     Standby: | metadata and data |  metadata and data  | metadata only     |   metadata only
+   *
+   */
+  public void replicateLimitedAcyclic() {
+    long oneRoundStartTimeMs = time.milliseconds();
+    shouldTerminateCurrentCycle = false;
+
+    logger.trace("Thread name: {} Start RemoteReplicaGroup replication", threadName);
```
snalli, line 823: rm "Thread name", the logger prints that anyways. Just put your msg in the logger

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+          groupIdIterationCountMap.put(groupId, groupIdIterationCountMap.getOrDefault(groupId, 0) + 1);
+          logger.trace("Thread name: {} Created new group for groupId {}, iteration count is {}", threadName, groupId,
+              groupIdIterationCountMap.getOrDefault(groupId, 0));
+        }
+      }
+
+      //get timed out stand by replicas with no progress
+      List<RemoteReplicaInfo> standByReplicasTimedOutOnNoProgress =
+          getRemoteStandbyReplicasTimedOutOnNoProgress(standbyReplicasWithNoProgress);
+
```
snalli, line 609: rm whitespace

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +787,153 @@ public void replicate() {
...
+      Map<Integer, RemoteReplicaGroup> inflightRemoteReplicaGroups = new HashMap<>();
+      Map<Integer, Integer> groupIdIterationCount = new HashMap<>();
+      Map<DataNodeId, Set<RemoteReplicaInfo>> standByNoProgressReplicaQueue = new HashMap<>();
+      Map<RemoteReplicaInfo, Long> remoteReplicaToThrottledTill = new HashMap<>();
+
+      Map<Integer, RemoteReplicaGroup> correlationIdToReplicaGroup = new HashMap<>();
+      Map<Integer, RequestInfo> correlationIdToRequestInfo = new LinkedHashMap<>();
+
+      MutableBoolean maxIterationsReached = new MutableBoolean(false);
```
snalli, line 847: please avoid this mutable boolean, use the global map that track iters per group to determine at the call site itself if any group has reached the max-iters

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+    remoteReplicas.forEach(remoteReplicaInfo -> {
+      if (remoteReplicaToThrottledTill.containsKey(remoteReplicaInfo)
+          && remoteReplicaToThrottledTill.get(remoteReplicaInfo) > time.milliseconds()) {
+        nonIteratingNonPendingRemoteReplicas.remove(remoteReplicaInfo);
+        throttleCount.inc();
+      }
+    });
+
+    //remove all replicas of corresponding group id if group is in iteration
+    if (inflightRemoteReplicaGroup.containsKey(groupId)) {
```
snalli, line 525: this can be written better, if you use an abstraction like GroupId. I am guessing the intent here is if this group has replicas that are in the middle of replication, then we don't include them in the next set of RemoteReplicaGroups right ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +787,153 @@ public void replicate() {
...
+
+    try {
+      storeKeyConverter.dropCache();
+      Map<DataNodeId, Integer> standByNoProgressGroupId = new HashMap<>();
+
+      // Generate group ids and stand by group ids, only these group ids will be used
+      generateGroupIdsForReplicas(groupIdToRemoteReplicaMap, standByNoProgressGroupId);
+      logger.trace("Thread name: {} Generated group ids for replication", threadName);
+
+      Map<Integer, RemoteReplicaGroup> inflightRemoteReplicaGroups = new HashMap<>();
```
snalli, line 839: just say in-Flight-Replicas

snalli (commented):
did second round, will do more next week after comments are addressed.

snalli:
For large values of iterations in a cycle, we would not pick up any new replicas added to a server because that set is refreshed only at teh end of a cycle. Similarly we won't pick up any removals. Is this correct? @justinlin-linkedin

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+
+      //generate active and standby replicas with no progress
+      getActiveAndStandByNoProgressReplicas(activeReplicas, standbyReplicasWithNoProgress, remoteReplicas,
+          remoteReplicaToThrottledTill, standbyReplicasNoProgressReplicaGroupId, groupId, inflightRemoteReplicaGroup);
+      logger.trace("Thread name: {} Active replicas are {}, standby Replicas with no progress are {}", threadName,
+          activeReplicas, standbyReplicasWithNoProgress);
+
+      //if the current group is not in iteration, and there are active replicas, we will start a new group
+      //and increase iteration count if we are able to create
+      if (!inflightRemoteReplicaGroup.containsKey(groupId)) {
```
snalli, line 595: instead of these maps and ids, please use an abstraction that can group together relevant info. For eg, this could be the state variable of an abstraction like IN_FLIGHT. Please cut back on the maps.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+    int groupId = 0;
+
+    for (Map.Entry<DataNodeId, List<RemoteReplicaInfo>> entry : dataNodeToRemoteReplicaInfo.entrySet()) {
+      DataNodeId remoteHost = entry.getKey();
+      List<RemoteReplicaInfo> remoteReplicasPerNode = entry.getValue();
+
+      //for this data node break a larger array of remote replicas to smaller multiple arrays of predefined size given in config
+      List<List<RemoteReplicaInfo>> remoteReplicaSegregatedList =
+          maxReplicaCountPerRequest > 0 ? Utils.partitionList(remoteReplicasPerNode, maxReplicaCountPerRequest)
+              : Collections.singletonList(remoteReplicasPerNode);
```
snalli, line 403: can we get rid of this config altogether ? and just put 1 replica in a group ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: @snalli check this flag for replica addition and deletion.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
manbearpig1996: We will also need to test it in production, we cannot set thread count to 0 there.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
```diff
@@ -343,6 +362,11 @@ public ReplicationConfig(VerifiableProperties verifiableProperties) {
     replicationInterReplicaThreadThrottleSleepDurationMs =
         verifiableProperties.getLongInRange("replication.inter.replica.thread.throttle.sleep.duration.ms", 0, 0,
             Long.MAX_VALUE);
+    replicationEnableAcyclicReplication = verifiableProperties.getBoolean("replication.enable.acyclic.replication", true);
```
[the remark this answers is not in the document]
manbearpig1996: Setting it to true will make more sense, as all tests and any thing i am missing will be using new code only.
and old code is kept as failsafe only.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: changed it.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: changed and refactored.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: combined and refactored.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: changed to name.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: In this code, everywhere it is written like this.
also in test logs, we need to print manually.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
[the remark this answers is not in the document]
manbearpig1996: refactored and moved.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +787,153 @@ public void replicate() {
...
+
+    MutableBoolean allReplicasCaughtUpEarly = new MutableBoolean(false);
+    Map<Integer, List<RemoteReplicaInfo>> groupIdToRemoteReplicaMap = new HashMap<>();
+
+    try {
+      storeKeyConverter.dropCache();
+      Map<DataNodeId, Integer> standByNoProgressGroupId = new HashMap<>();
+
+      // Generate group ids and stand by group ids, only these group ids will be used
+      generateGroupIdsForReplicas(groupIdToRemoteReplicaMap, standByNoProgressGroupId);
```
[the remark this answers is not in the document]
manbearpig1996: refactored and moved.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+          logger.trace("Thread name: {} Created new group for groupId {}, iteration count is {}", threadName, groupId,
+              groupIdIterationCountMap.getOrDefault(groupId, 0));
+        }
+      }
+
+      //get timed out stand by replicas with no progress
+      List<RemoteReplicaInfo> standByReplicasTimedOutOnNoProgress =
+          getRemoteStandbyReplicasTimedOutOnNoProgress(standbyReplicasWithNoProgress);
+
+      if (standByReplicasTimedOutOnNoProgress.isEmpty()) {
```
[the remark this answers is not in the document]
manbearpig1996: writing continue seems to make code more clear.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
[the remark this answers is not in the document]
manbearpig1996: inflight means groups that are currently in progress.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: removed and fixed

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+    if (!shouldCreateNewRemoteReplicaGroups(inflightRemoteReplicaGroup, groupIdIterationCountMap,
+        allReplicasCaughtUpEarly, maxIterationsReached)) {
+      logger.trace("Thread name: {} Not creating any new groups", threadName);
+      return;
+    }
+
+    logger.trace("Thread name: {} Trying to create new active groups", threadName);
+    //now we create all active groups and add stand by no progress replicas to queue to be processed later
+    for (Map.Entry<Integer, List<RemoteReplicaInfo>> entry : groupIdToRemoteReplicaMap.entrySet()) {
+      Integer groupId = entry.getKey();
```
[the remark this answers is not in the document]
manbearpig1996: refactored it.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: changed to inflightGroups

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: refactored it.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: We need this so , at any point of time, we make less calls to other node.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+        int groupId = entry.getKey();
+        logger.trace("Thread name: {} Processing for group id {}", threadName, groupId);
+        List<RemoteReplicaInfo> remoteReplicas = entry.getValue();
+        DataNodeId remoteNode = remoteReplicas.get(0).getReplicaId().getDataNodeId();
+
+        // get stand by group id for the data node
+        int standByGroupId = standByGroupIds.get(remoteNode);
+
+        List<RemoteReplicaInfo> activeReplicas = new ArrayList<>();
+        List<RemoteReplicaInfo> standbyReplicasWithNoProgress = new ArrayList<>();
```
snalli, line 2273: rename to standByReplicas

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+
+        // generate active and standby replicas with no progress
+        getActiveAndStandByNoProgressReplicas(activeReplicas, standbyReplicasWithNoProgress, remoteReplicas,
```
snalli, line 2276: rename to fillActiveAndStandByReplicas

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+        List<RemoteReplicaInfo> standByReplicasTimedOutOnNoProgress =
+            getRemoteStandbyReplicasTimedOutOnNoProgress(standbyReplicasWithNoProgress);
+
+        if (standByReplicasTimedOutOnNoProgress.isEmpty()) {
+          continue;
+        }
+
+        // add timed out stand by replicas with no progress to queue
+        standByReplicaQueue.putIfAbsent(remoteNode, new HashSet<>());
+        standByReplicaQueue.get(remoteNode).addAll(standByReplicasTimedOutOnNoProgress);
```
snalli, line 2304: these can be combined

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+            inflightGroups.put(groupId, groupForActiveReplicas);
+            groupIdIterationCountMap.put(groupId, groupIdIterationCountMap.getOrDefault(groupId, 0) + 1);
+            logger.trace("Thread name: {} Created new group for groupId {}, iteration count is {}", threadName, groupId,
+                groupIdIterationCountMap.getOrDefault(groupId, 0));
+          }
+        }
+
+        // get timed out stand by replicas with no progress
+        List<RemoteReplicaInfo> standByReplicasTimedOutOnNoProgress =
+            getRemoteStandbyReplicasTimedOutOnNoProgress(standbyReplicasWithNoProgress);
```
snalli, line 2296: rename to getTimedOutStandByReplicas

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+
+        if (standByReplicasTimedOutOnNoProgress.isEmpty()) {
```
snalli, line 2298: if (!X.isEmtpy()) { then do something }
rm the continue

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+     * Generates group ids for all datanodes for standby no progress groups
+     * All group ids are unique
+     * All replicas are mapped to particular group id and each datanode has one stand by remote replica group id
+     * so we make less number of parallel calls to remote hosts
+     * {@link #groupIdToRemoteReplicaMap} this map will be populated with group ids to replicas map
+     * {@link #standByGroupIds}  this map will be populated with datanode to standby group id
+     */
+    private void createGroupIds() {
+      Map<DataNodeId, List<RemoteReplicaInfo>> dataNodeToRemoteReplicaInfo = selectReplicas(getRemoteReplicaInfos());
+
```
snalli, line 2128: groupIdToRemoteReplicaMap : id -> replicas
standByGroupIds : node -> id
states: inflight
var: num_iter, throttled_till

How about this ? Lmk your thoughts

List<PeerReplicas> listPeerReplicas; // iterate on this list in the do-while loop

class RemoteReplicaGroupInfo {
 RemoteReplicaGroup group;
 int numIterations;
 long throttleuntil;
 // use the group.state to check if inflight, remove the inflight state
}

class PeerReplicas {
  DataNodeId datanode;
  List<RemoteReplicaInfo> remoteReplicas;
  List<RemoteReplicaGroupInfo> activeGroups; // iterate on this for each peer
  List<RemoteReplicaGroupInfo> standByGroups; // iterate on this for each peer
}

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+        // generate active and standby replicas with no progress
+        getActiveAndStandByNoProgressReplicas(activeReplicas, standbyReplicasWithNoProgress, remoteReplicas,
+            standByGroupId, groupId);
+        logger.trace("Thread name: {} Active replicas are {}, standby Replicas with no progress are {}", threadName,
+            activeReplicas, standbyReplicasWithNoProgress);
+
+        // if the current group is not in iteration, and there are active replicas, we will start a new group
+        // and increase iteration count if we are able to create
+        if (!inflightGroups.containsKey(groupId)) {
```
justinlin-linkedin, line 2283: We should check if the group id is already in the inflight map before getting the active replica and standby with no progress. This method would be called quite often, every time in the do-while loop, we would try to poll and create new groups, which means most of the time, the group is still inflight. If we check the inflight group first, we would avoid the logic above most of the time.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+    }
+
+    /**
+     * checks if any signal to terminate cycle is received,
+     * checks whether any group has reached max iteration count
+     * @return {@link boolean} returns true if new groups can be created, false otherwise
+     */
+    private boolean shouldCreateNewGroups() {
+      // check if thread is in shutdown or cycle termination is received or max iterations reached
+      if (!running || shouldTerminateCurrentCycle) {
```
justinlin-linkedin, line 2192: Let's add more comments here, to list out the conditions when we don't want to create new groups.
1. the process is shutting down, so no need to create new group even if we have more iteration for them to run.
2. shouldTerminateCurrentCycle is true, which means there are replicas added to this thread or removed from this thread, don't create new group and just let existing groups finish so we can later create new groups with the new set of replicas in next cycle.
3. any group reaches max iteration, which means existing groups should just finish its current work and stop, thus no new groups should be created.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
       lock.unlock();
     }
+    shouldTerminateCurrentCycle = true;
     if (networkClient != null) {
       networkClient.close();
     }
     shutdownLatch.await();
   }
 
+  class RemoteReplicaGroupPoller {
```
justinlin-linkedin, line 2067: I would love to add more comment to explain the life cycle of a group.

I suppose it works like this
1. We preassign group id to each group. 
2. Once a group finishes one iteration of work, include metadata exchange and data download, we remove group id from the inflight group map and potentially backoff. If this group need to go through another iteration, the same group would be added back to inflight group map.
3. PollGroups would try to create (and add back) those groups.
4. When there is a new replica added to the thread, or a replica removed from this thread, we don't create new groups, just return the existing ones, and after they all finish current iteration, inflightGroupMap would become empty and do-while loop would break.
5. Or there is one group reaches max iteration.

justinlin-linkedin (commented):
second round of review, will do another round

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
snalli: Instead of creating new groups, is it possible to reuse a group by just resetting its state to the start state ?

manbearpig1996: Replicas can change in the group in each iteration, as some replicas may go to backoff mode 
or vice-versa.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: For this group id, all replicas may not be in the inflight group, some may have been stand by and timeout can happen any time also.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: added comments.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: Added comments

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: We are specifically wanting standby no progress replicas, so kept name here.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: This method is existing method and getting used in other parts too, so kept is as such.

ambry-vcr/src/test/java/com/github/ambry/vcr/CloudTokenPersistorTest.java
manbearpig1996: fixed.

ambry-api/src/main/java/com/github/ambry/config/ReplicationConfig.java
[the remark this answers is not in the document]
manbearpig1996: Changed it to false.
My intention was to roll out config as false internally first and later merge this to oss. So all new code is compatible with new code.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -348,6 +377,296 @@ protected Map<DataNodeId, List<RemoteReplicaInfo>> selectReplicas(Map<DataNodeId
...
+   * All replicas are mapped to particular group id and each datanode has one stand by remote replica group id
+   * so we make less number of parallel calls to remote hosts
+   * @param groupIdToRemoteReplicaMap {@link Map} this map will be populated with group ids to replicas map
+   * @param standByNoProgressGroupId {@link Map}  this map will be populated with datanode to standby group id
+   */
+  void generateGroupIdsForReplicas(Map<Integer, List<RemoteReplicaInfo>> groupIdToRemoteReplicaMap,
+      Map<DataNodeId, Integer> standByNoProgressGroupId) {
+
+    Map<DataNodeId, List<RemoteReplicaInfo>> dataNodeToRemoteReplicaInfo = selectReplicas(getRemoteReplicaInfos());
+
```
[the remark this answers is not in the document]
manbearpig1996: fixed.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -466,7 +787,153 @@ public void replicate() {
...
+   *  Here is a table listing on what is exchanged between local and remote replicas based on their roles
+   *  (leader/standby) when {@link ReplicationModelType is LEADER_BASED}.
+   *
+   *              |   Local Leader    |     Local Standby   |   Remote Leader   |  Remote Standby
+   *            -------------------------------------------------------------------------------------
+   *     Leader:  |        ---        |  metadata and data  | metadata and data |   metadata only
+   *     Standby: | metadata and data |  metadata and data  | metadata only     |   metadata only
+   *
+   */
+  public void replicateLimitedAcyclic() {
```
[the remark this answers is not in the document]
manbearpig1996: changed it.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: fixed.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: couldn't find a way to combine right now.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: fixed.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1881,18 +2057,318 @@ void shutdown() throws InterruptedException {
...
+    }
+  }
+
   /**
    * ReplicaGroupReplicationState indicates different state of RemoteReplicaGroup in one replication cycle.
    * Each replication cycle, we will break all the replicas in order different groups and each group
    * has to go through these states to finish replication with nonblocking network client.
    */
-  enum ReplicaGroupReplicationState {
+   enum ReplicaGroupReplicationState {
```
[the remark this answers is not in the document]
manbearpig1996: fixed.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1967,6 +2444,7 @@ public RemoteReplicaGroup(List<RemoteReplicaInfo> remoteReplicaInfos, DataNodeId
       this.remoteDataNode = dataNodeId;
       this.isNonProgressStandbyReplicaGroup = isNonProgressStandbyReplicaGroup;
       this.id = id;
+      this.storeKeysConversionLog =  new ArrayList<>();
```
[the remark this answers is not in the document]
manbearpig1996: fixed.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
justinlin-linkedin: For the active replicas that are not in the group, they will not be added to the group anyway, since the group is still inflight. Even we find more active replicas this time, they are still going on next iteration. 
For timedout standby replicas, it's the same case. Even we find more timed out standby replicas, if the standby group is still inflight, they will not be added to the current group anyway. 

That's why i suggest separate the function to add active replicas and timed out standby replicas. They belong to different groups and will not be reevaluated at the same time.

manbearpig1996: If we do find timedout standby replicas, we are adding to queue in  code below. and whenever standBygroup is free they will be picked up.
Sure, we can  add to queue when active group is done, but here i am checking if any replica not added to active group has timedout, then i can add to queue as soon as possible.
We will add standby replica to queue sooner in current approach

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1892,12 +2069,332 @@ void shutdown() throws InterruptedException {
...
+    /**
+     * Removes finished groups from inflight remote replica groups  {@link #inflightGroups}
+     * drops cached keys from storeKeyConverter.
+     * For finished groups adds time until which replicas are throttled in to {@link #replicaToThrottledTill}
+     */
+    private void processFinishedGroups() {
+      Iterator<Map.Entry<Integer, RemoteReplicaGroup>> inflightRemoteReplicaGroupIterator =
+          inflightGroups.entrySet().iterator();
+
+      while (inflightRemoteReplicaGroupIterator.hasNext()) {
```
snalli, line 2194: just iterate the map as key-value pairs, get rid of the iterator

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1892,12 +2069,332 @@ void shutdown() throws InterruptedException {
...
+    private void createStandByGroups() {
+      logger.trace("Thread name: {} Trying to create new standby groups", threadName);
+      for (Map.Entry<DataNodeId, Set<RemoteReplicaInfo>> entry : standByReplicaQueue.entrySet()) {
+        DataNodeId remoteNode = entry.getKey();
+        List<RemoteReplicaInfo> remoteReplicaInfos = new ArrayList<>(entry.getValue());
+        int groupId = standByGroupIds.get(remoteNode);
+        logger.trace("Thread name: {} Trying for remoteNode {}, standBy groupId {}}", threadName, remoteNode, groupId);
+
+        // if group is already in iteration, or there are no replicas in queue do nothing
+        if (inflightGroups.containsKey(groupId) || remoteReplicaInfos.isEmpty()) {
```
snalli, line 2355: here again, rm the continue; if (X) {then do somethin}

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1892,12 +2069,332 @@ void shutdown() throws InterruptedException {
...
+            groupIdIterationCountMap.put(groupId, groupIdIterationCountMap.getOrDefault(groupId, 0) + 1);
+            logger.trace("Thread name: {} Created new group for groupId {}, iteration count is {}", threadName, groupId,
+                groupIdIterationCountMap.getOrDefault(groupId, 0));
+          }
+        }
+
+        // get timed out stand by replicas with no progress
+        List<RemoteReplicaInfo> standByReplicasTimedOutOnNoProgress =
+            getRemoteStandbyReplicasTimedOutOnNoProgress(standbyReplicasWithNoProgress);
+
```
snalli, line 2331: There seems to be 3 types of replicas: active, standBy, standByTimedOut ? correct ?

First, can you please consistently get-rid of the term of NoProgress everywhere in your PR. I see its removed in some places where I pointed out but left in others. Please be consistent in the naming in all places.

Second, make a clear distinction between standBy and standByTimedOut replicas. 
For eg. this queue standByReplicaQueue is really a queue for timed-out replicas. correct ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1892,12 +2069,332 @@ void shutdown() throws InterruptedException {
...
+              standByReplicasTimedOutOnNoProgress);
+        }
+      }
+    }
+
+    /**
+     * iterate over all data nodes and stand by queue of all data nodes {@link #standByReplicaQueue},
+     * and tries to create standBy Groups and add to {@link #inflightGroups}
+     */
+    private void createStandByGroups() {
```
snalli, line 2346: this fn operates on standby-timed-out replicas right ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1892,12 +2069,332 @@ void shutdown() throws InterruptedException {
...
+      logger.trace("Thread name: {} Trying to create new standby groups", threadName);
+      for (Map.Entry<DataNodeId, Set<RemoteReplicaInfo>> entry : standByReplicaQueue.entrySet()) {
+        DataNodeId remoteNode = entry.getKey();
+        List<RemoteReplicaInfo> remoteReplicaInfos = new ArrayList<>(entry.getValue());
+        int groupId = standByGroupIds.get(remoteNode);
+        logger.trace("Thread name: {} Trying for remoteNode {}, standBy groupId {}}", threadName, remoteNode, groupId);
+
```
snalli, line 2353: if the standby replicas here are always timed-out, will their replica-group be marked as DONE somehwere ? because only if the group is marked Done, the code can move fwd here right ?

snalli (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1892,12 +2069,332 @@ void shutdown() throws InterruptedException {
...
+      // remove all throttled replicas
+      remoteReplicas.forEach(remoteReplicaInfo -> {
+        if (replicaToThrottledTill.containsKey(remoteReplicaInfo)
+            && replicaToThrottledTill.get(remoteReplicaInfo) > time.milliseconds()) {
+          nonIteratingNonPendingRemoteReplicas.remove(remoteReplicaInfo);
+          throttleCount.inc();
+        }
+      });
+
+      // remove all replicas of corresponding group id if group is in iteration
```
snalli, line 2272: I don't quite understand this logic here. At the beginning, the replicas in a remote node are mapped to group-id. Then here in the fill(...), if that group-id is in-flight, then you rm it from this non-iter-non-xxx list. Wont this list be empty after this ?
is it possible that the some replicas belonging to a group-id are inflight and others not  ? Don't we track in-flight at a group level ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: each data node has only one timed out standby group. When this timed out stand by group finishes, only then, we can create a new group for timedout standby replicas in queue.

Current code also has only one standby group, I am assuming that is being done to reduce cross colo bandwidth usage.

snalli (changes requested):
I do not approve of this design and implementation. I've made it very clear in my comments and in slack to redesign with better abstractions and remove all the unnecessary maps and queues. I've even demonstrated how this can be done here in this template pr

https://github.com/linkedin/ambry/pull/2842

Please use this and fill in the blanks. Write tests and complete it.

PR #2821, opened 2024-07-09
Server returns Blob_Deleted for deleted blob out of retention

justinlin-linkedin:
## Summary
When issuing a get request with include_deleted get option, frontend would send a get request to server to fetch the content of the blob. However, if any chunk blob is deleted and out of retention window, even if we have metadata chunk still in the server, we can't finish downloading the entire blob. 

This PR fixes this issue by not returning content for metadata. Now if client issues a get request with include_deleted get option and trying to download a blob that was already deleted and out of retention window, server would just return ID_Deleted error back to frontend so frontend would just return 410 back to client right away.

To download deleted and out of retention blobs, we can issue a get request with include_all.

## Test
Unit test

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/store/StoreGetOptions.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/StoreGetOptions.java
@@ -25,8 +25,12 @@ public enum StoreGetOptions {
   Store_Include_Expired,
   /**
    * This option indicates that the store needs to return the message even if it has been
-   * marked for deletion as long as the message has not been physically deleted from the
-   * store.
+   * marked for deletion as long as the message is still in delete retention window
    */
-  Store_Include_Deleted
+  Store_Include_Deleted,
+  /**
+   * This option indicates that the store needs to return the message as long as the message
+   * has not been physically deleted from the store.
+   */
+  Store_Include_Compaction_Ready
 }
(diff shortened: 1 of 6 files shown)


ambry-api/src/main/java/com/github/ambry/store/StoreGetOptions.java
```diff
@@ -25,8 +25,12 @@ public enum StoreGetOptions {
...
-   * store.
+   * marked for deletion as long as the message is still in delete retention window
    */
-  Store_Include_Deleted
+  Store_Include_Deleted,
+  /**
+   * This option indicates that the store needs to return the message as long as the message
+   * has not been physically deleted from the store.
+   */
+  Store_Include_Compaction_Ready
```
SophieGuo410, line 35: QQ: currently we have several user who enable Store_Include_Deleted to get the blob as long as it not been physically deleted. So seems like this is not a backward compatible change. Will this affect any of their current use cases?

ambry-api/src/main/java/com/github/ambry/store/StoreGetOptions.java
```diff
@@ -25,8 +25,12 @@ public enum StoreGetOptions {
...
   /**
    * This option indicates that the store needs to return the message even if it has been
-   * marked for deletion as long as the message has not been physically deleted from the
-   * store.
+   * marked for deletion as long as the message is still in delete retention window
    */
-  Store_Include_Deleted
+  Store_Include_Deleted,
+  /**
+   * This option indicates that the store needs to return the message as long as the message
```
Arun-LinkedIn, line 32: QQ. Do we need this new option? Would it be easier if we tell customers that `Store_Include_Deleted` only works if they issue GET within the delete retention window?

Arun-LinkedIn (approved):

SophieGuo410 (approved):

ambry-api/src/main/java/com/github/ambry/store/StoreGetOptions.java
justinlin-linkedin: The reason to add this option is to make sure that Include_All get option would continue to work for compaction_ready blobs.

PR #2825, opened 2024-07-12
Issue force delete for 503 delete operations

justinlin-linkedin:
## Summary

Issue a force delete request when the delete request is getting 503. 

## Details
When originating replicas returns more than 1 unavailable to frontend for delete operations, we have to return 503 back to client since we can't be sure about the status of the blob. If we have two unavailable hosts, (due to any reason, server busy, server unavailable, etc), it's impossible to know if this blob exists on these two hosts or not. In this case, we have to return 503 so client can retry later.

However, when executing delete, in most cases, we know it would be fine to place a delete tombstone to available replicas and just return 200 back to client, since the delete is the final state.

There are some catches here. If the cluster enables undelete, we might have some unavailable replicas that have blob at certain lifeVersion. A force delete would place a delete tombstone with life version 0, which will be overritten by the higher life version later. 

## Test
Unit tests.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
@@ -81,6 +81,8 @@ class DeleteOperation {
   private final OperationQuotaCharger operationQuotaCharger;
   // Denotes whether the operation is complete.
   private boolean operationCompleted = false;
+  // When this variable is true, create DeleteRequest with force delete to be true.
+  private boolean enableForceDeleteInRequest = false;
 
   private final NonBlockingRouter nonBlockingRouter;
   private ReplicateBlobCallback replicateBlobCallback;
@@ -174,8 +176,13 @@ class DeleteOperation {
    * @return The DeleteRequest.
    */
   private DeleteRequest createDeleteRequest() {
-    return new DeleteRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
-        blobId, deletionTimeMs);
+    if (!enableForceDeleteInRequest) {
+      return new DeleteRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
+          blobId, deletionTimeMs);
+    } else {
+      return new DeleteRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
+          blobId, deletionTimeMs, DeleteRequest.DELETE_REQUEST_VERSION_3, enableForceDeleteInRequest);
+    }
   }
 
   /**
@@ -387,7 +394,7 @@ class DeleteOperation {
     return (routerConfig.routerRepairWithReplicateBlobOnDeleteEnabled && operationTracker.hasNotFound()
         && operationTracker.getSuccessCount() > 0
         && getPrecedenceLevel(errorCode) >= getPrecedenceLevel(RouterErrorCode.AmbryUnavailable)
-        && (serviceId == null || !serviceId.startsWith(BackgroundDeleteRequest.SERVICE_ID_PREFIX)));
+        && !isServiceBackgroundDeleter());
     // @formatter:on
   }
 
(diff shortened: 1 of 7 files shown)


ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -407,7 +414,38 @@ private boolean shouldRunOfflineRepair() {
...
-        && (serviceId == null || !serviceId.startsWith(BackgroundDeleteRequest.SERVICE_ID_PREFIX)));
+        && !isServiceBackgroundDeleter());
+    // @formatter:on
+  }
+
+  private boolean shouldIssueForceDelete() {
+    // The conditions to issue force delete to replicas
+    // 1. the feature is enabled
+    // 2. there is no successful replica. If we have success replicas, we should try on demand replication and retry
+    // 3. error code precedence is over AmbryUnavailable. We don't do replication if the error is BlobExpired or TooManyRequests.
```
Arun-LinkedIn, line 425: nit: Seems like comment is referencing `replication`.

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -407,7 +414,38 @@ private boolean shouldRunOfflineRepair() {
...
+    // 5. it's not a delete request from BackgroundDeleter.
+
+    // What does it mean? Here are the cases where we would need to force delete
+    // 1. 2 or 3 originating replicas return unavailable, including replica unavailable, disk unavailable, retry_after_backoff, io_error, network error...
+    // 2. we have 1 leader, 1 bootstrap, 1 offline replica in originating dc. They all return BlobNotFound
+    // 3. we have 1 leader, 1 standby, 1 bootstrap, 1 offline replica in originating dc. They all return BlobNotFound.
+    // These three cases won't return NOT_FOUND and won't trigger on demand replication and offline repair. But force delete
+    // might be able to help.
+    // Force delete shouldn't interfere with on demand replication and offline repair. But when force delete fails, it might
+    // trigger on demand replication.
```
Arun-LinkedIn, line 436: Just a doubt. Why could on-demand replication be triggered if force delete fails?

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
```diff
@@ -174,8 +176,13 @@ private void fetchRequests(RequestRegistrationCallback<DeleteOperation> requestR
...
    */
   private DeleteRequest createDeleteRequest() {
-    return new DeleteRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
-        blobId, deletionTimeMs);
+    if (!enableForceDeleteInRequest) {
+      return new DeleteRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
+          blobId, deletionTimeMs);
+    } else {
+      return new DeleteRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
+          blobId, deletionTimeMs, DeleteRequest.DELETE_REQUEST_VERSION_3, enableForceDeleteInRequest);
```
Arun-LinkedIn, line 184: nit: If we come here, `enableForceDeleteInRequest` in `DeleteRequest` constructor can be replaced  `true` :)

Also, when server change is rolled out, we could also update `DeleteRequest.CURRENT_VERSION` to 3 and avoid passing version in the constructor call? 
```
return new DeleteRequest(NonBlockingRouter.correlationIdGenerator.incrementAndGet(), routerConfig.routerHostname,
          blobId, deletionTimeMs, enableForceDeleteInRequest)
```

Arun-LinkedIn (approved):
Looks good to me!

SophieGuo410 (approved):

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
justinlin-linkedin: updated

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
justinlin-linkedin: Force delete might fail on some replicas, but succeed in one replica. If so, it will trigger on-demand replication, which is fine.

ambry-router/src/main/java/com/github/ambry/router/DeleteOperation.java
justinlin-linkedin: We can do that later.

PR #2827, opened 2024-07-16
[vcr-2.0] Add configurable azure-retry policy

snalli:
We are experiencing some throttling from Azure Storage as we keep expanding the VCR cluster. Azure support suggested trying exponential backoff. This patch changes the FIXED backoff policy into a configurable one allowing us to try different policies and retry-delays.


litingulfs (approved):
👍   looks good

PR #2828, opened 2024-07-16
[vcr-2.0] Fix nullptr in ReplicationMetrics

snalli:
Encountering missing metric in VCR. Not sure why because the code that adds it seems to be ok. We don't really rely on this metric in VCR, so just patching up the code to not run into a nullptr.


litingulfs (approved):
👍

PR #2830, opened 2024-07-17
[vcr-2.0] Remove unnecessary sleep in azure-client

snalli:
The Azure SDK interprets the tryTimeoutInSeconds as sleep duration after each request to Azure. However, we are using a sync-client which blocks for the same duration waiting on a response. So we don't need an additional sleep in the client due to tryTimeoutInSeconds. 

The above is azure-blob-client. Also, fixes azure-table-client similarly.


litingulfs (approved):
👍 - let's not sleep or wait longer than we really need to.

PR #2834, opened 2024-07-19
[vcr-2.0] Remove redundant TTL-updates

snalli:
The replication code performs a TTL-update on a permanent blob. This might be ok for disks but costs at least 1 network call to Azure. So just get rid of it. 

+ Add some cache related stats info
+ Remove logs that flood vcr.log
+ Remove unused code and variables

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
@@ -856,52 +856,7 @@ public class AzureCloudDestinationSync implements CloudDestination {
     BlobProperties blobProperties = getBlobPropertiesCached(blobLayout);
     Map<String, String> cloudMetadata = blobProperties.getMetadata();
 
-    try {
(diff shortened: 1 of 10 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -837,93 +847,52 @@ public boolean doesBlobExist(BlobId blobId) {
   }
 
   @Override
-  public short updateBlobExpiration(BlobId blobId, long expirationTime, CloudUpdateValidator cloudUpdateValidator)
+  public short
```
justinlin-linkedin, line 850: could you format this file?

justinlin-linkedin (approved):
LGTM

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: the latest version of intellij did this!

PR #2836, opened 2024-07-19
[vcr-2.0] Delete blob eagerly

snalli:
Implementing immediate blob deletion from the cloud once the VCR receives a DEL message from the server via replication. This approach offers several advantages over our current method, where instead of deleting the blob outright, we currently add a delete-timestamp to Azure blob metadata. Compaction then deletes the blob later based on this timestamp.

The benefits of the new approach include:
1. Proactively reclaiming space in Azure.
2. Reducing the number of network calls required to delete a blob; currently, it involves 2 network calls to attach a timestamp and another 2 network calls from compaction to remove the blob.
3. Decreasing the size of the response from the list-metadata API used by compaction to list all blobs in an Azure partition.

Based on customer usage patterns, few customers utilize the undelete feature. If the VCR receives an undelete request after a blob has been removed from the cloud, it will be re-uploaded to the cloud using the replication protocol.

This patch also removes some unused legacy code

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -636,24 +636,16 @@ public class CloudBlobStore implements Store {
 
   @Override
   public void delete(List<MessageInfo> infos) throws StoreException {
-    // TODO: Remove the duplicate code by calling deleteAsync() method.
-    checkStarted();
-    checkDuplicates(infos);
-
     try {
-      for (MessageInfo msgInfo : infos) {
-        BlobId blobId = (BlobId) msgInfo.getStoreKey();
-        // If the cache has been updated by another thread, retry may be avoided
-        requestAgent.doWithRetries(() -> deleteIfNeeded(blobId, msgInfo.getOperationTimeMs(), msgInfo.getLifeVersion()),
-            "Delete", partitionId.toPathString());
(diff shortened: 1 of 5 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -636,24 +636,15 @@ private boolean shouldUpload(MessageInfo messageInfo) {
...
-    checkDuplicates(infos);
-
     try {
-      for (MessageInfo msgInfo : infos) {
-        BlobId blobId = (BlobId) msgInfo.getStoreKey();
-        // If the cache has been updated by another thread, retry may be avoided
-        requestAgent.doWithRetries(() -> deleteIfNeeded(blobId, msgInfo.getOperationTimeMs(), msgInfo.getLifeVersion()),
-            "Delete", partitionId.toPathString());
+      for (MessageInfo msg : infos) {
+        cloudDestination.deleteBlob((BlobId) msg.getStoreKey(), msg.getOperationTimeMs(), msg.getLifeVersion(), null);
```
justinlin-linkedin, line 641: are we not doing any retries anymore?

snalli: the retries are handled by the azure client sdk, so i removed any retries from ambry a long time ago. Now i am just removing the code itself because its just unreadable.

justinlin-linkedin (approved):

PR #2839, opened 2024-07-24
[vcr-2.0] Debug replication error

snalli:
Encountered replication errors in VCR but not sure where they are coming from because of identical error messages. This patch distinguishes those message by adding some more info.


litingulfs (approved):
👍  this should make it easier to debug, I agree.

PR #2841, opened 2024-07-26
Batch Delete server changes

piyujai:
## Summary
Ambry batch delete support - server changes

## Description
Existing delete operation handles one blob in a single request, the batch delete change introduces handling on server side of a BatchDeleteRequest received from frontend OperationController. At the store level, this includes deleting a list of blobs in a partition and handling the aggregated response in batchDelete util. This is needed for deletion of composite blobs, along with significant performance improvement on large sized composite blobs.

## Testing
Unit tests done.
Integration tests done.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
+++ b/ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
@@ -56,6 +56,15 @@ public interface RequestAPI {
   void handleDeleteRequest(NetworkRequest request) throws IOException, InterruptedException;
 
   /**
+   * Deletes a list of blobs from the store.
+   * @param request The request that contains the list of partition and id of the blob that needs to be deleted.
+   * @throws IOException if there are I/O errors carrying our the required operation.
+   * @throws InterruptedException if request processing is interrupted.
+   */
+
+  void handleBatchDeleteRequest(NetworkRequest request) throws IOException, InterruptedException;
+
+  /**
    * Purges the blob from the store.
    * @param request The request that contains the partition and id of the blob that needs to be purged.
    * @throws IOException if there are I/O errors carrying our the required operation.
--- a/ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
+++ b/ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
@@ -56,8 +56,9 @@ public interface RequestAPI {
   void handleDeleteRequest(NetworkRequest request) throws IOException, InterruptedException;
 
   /**
-   * Deletes a list of blobs from the store.
-   * @param request The request that contains the list of partition and id of the blob that needs to be deleted.
+   * Deletes a list of blobs from the store. Also handles the scenarios of complete success, partial success and full
+   * failure by returning appropriate response {@code BatchDeleteResponse} to network channel.
+   * @param request The request that contains the list of partitions and ids of the blob that needs to be deleted.
    * @throws IOException if there are I/O errors carrying our the required operation.
    * @throws InterruptedException if request processing is interrupted.
    */
--- a/ambry-api/src/main/java/com/github/ambry/store/MessageErrorInfo.java
+++ b/ambry-api/src/main/java/com/github/ambry/store/MessageErrorInfo.java
@@ -1,0 +1,32 @@
+/**
+ * Copyright 2024 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.store;
+
+public class MessageErrorInfo {
+  private final MessageInfo messageInfo;
+  private final StoreErrorCodes error;
+
+  public MessageErrorInfo(MessageInfo messageInfo, StoreErrorCodes error) {
+    this.messageInfo = messageInfo;
+    this.error = error;
+  }
+
+  public MessageInfo getMessageInfo() {
+    return messageInfo;
+  }
+
+  public StoreErrorCodes getError() {
+    return error;
+  }
+}
--- a/ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
+++ b/ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
@@ -106,8 +106,11 @@ public class ServerMetrics {
   public final Histogram batchDeleteBlobResponseQueueTimeInMs;
   public final Histogram purgeBlobResponseQueueTimeInMs;
   public final Histogram deleteBlobSendTimeInMs;
+
+  public final Histogram batchDeleteBlobSendTimeInMs;
   public final Histogram purgeBlobSendTimeInMs;
   public final Histogram deleteBlobTotalTimeInMs;
+  public final Histogram batchDeleteBlobTotalTimeInMs;
   public final Histogram purgeBlobTotalTimeInMs;
 
   public final Histogram undeleteBlobRequestQueueTimeInMs;
@@ -262,6 +265,10 @@ public class ServerMetrics {
   public final Counter unExpectedStoreGetError;
   public final Counter unExpectedStoreTtlUpdateError;
   public final Counter unExpectedStoreDeleteError;
+
+  public final Counter unExpectedStoreDeleteErrorInBatchDelete;
+
+  public final Counter unExpectedStoreBatchDeleteError;
   public final Counter unExpectedStorePurgeError;
   public final Counter unExpectedStoreUndeleteError;
   public final Counter unExpectedAdminOperationError;
@@ -270,22 +277,31 @@ public class ServerMetrics {
   public final Counter dataCorruptError;
   public final Counter unknownFormatError;
   public final Counter idNotFoundError;
+
+  public final Counter idNotFoundErrorInBatchDelete;
   public final Counter idDeletedError;
+
+  public final Counter idDeletedErrorInBatchDelete;
   public final Counter idPurgedError;
   public final Counter idUndeletedError;
   public final Counter idNotDeletedError;
   public final Counter lifeVersionConflictError;
   public final Counter ttlExpiredError;
+
+  public final Counter ttlExpiredErrorInBatchDelete;
   public final Counter badRequestError;
   public final Counter temporarilyDisabledError;
   public final Counter getAuthorizationFailure;
   public final Counter deleteAuthorizationFailure;
+
+  public final Counter deleteAuthorizationFailureInBatchDelete;
   public final Counter purgeAuthorizationFailure;
   public final Counter undeleteAuthorizationFailure;
   public final Counter ttlUpdateAuthorizationFailure;
   public final Counter ttlAlreadyUpdatedError;
   public final Counter ttlUpdateRejectedError;
   public final Counter replicationResponseMessageSizeTooHigh;
+  public final Counter batchDeleteOperationError;
   public Counter sealedReplicasMismatchCount = null;
   public Counter partiallySealedReplicasMismatchCount = null;
   public Counter stoppedReplicasMismatchCount = null;
@@ -411,8 +427,10 @@ public class ServerMetrics {
     purgeBlobResponseQueueTimeInMs =
         registry.histogram(MetricRegistry.name(requestClass, "PurgeBlobResponseQueueTimeInMs"));
     deleteBlobSendTimeInMs = registry.histogram(MetricRegistry.name(requestClass, "DeleteBlobSendTime"));
+    batchDeleteBlobSendTimeInMs = registry.histogram(MetricRegistry.name(requestClass, "BatchDeleteBlobSendTime"));
     purgeBlobSendTimeInMs = registry.histogram(MetricRegistry.name(requestClass, "PurgeBlobSendTimeInMs"));
     deleteBlobTotalTimeInMs = registry.histogram(MetricRegistry.name(requestClass, "DeleteBlobTotalTime"));
+    batchDeleteBlobTotalTimeInMs = registry.histogram(MetricRegistry.name(requestClass, "BatchDeleteBlobTotalTime"));
     purgeBlobTotalTimeInMs = registry.histogram(MetricRegistry.name(requestClass, "PurgeBlobTotalTimeInMs"));
 
     undeleteBlobRequestQueueTimeInMs =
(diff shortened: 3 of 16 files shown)


manbearpig1996 (commented):
What does current delete api do for batch request ?
What does new api do differently ?

piyujai:
> What does current delete api do for batch request ? What does new api do differently ?

Current delete api has a list of blobs in params, but handles it in a All or Nothing way, since it raises exception on first error in  the blob list. Also, all current usages of the existing delete operation is in the form of a singleton list itself, so no partial failure handling is required as well.
 In Batch Delete store util, we are handling a list of blobs for all possible cases (all_success, partial_failure, full_failure) by returning an aggregated response.

ambry-api/src/main/java/com/github/ambry/protocol/RequestAPI.java
```diff
@@ -55,6 +55,15 @@ public interface RequestAPI {
    */
   void handleDeleteRequest(NetworkRequest request) throws IOException, InterruptedException;
 
+  /**
+   * Deletes a list of blobs from the store.
```
manbearpig1996, line 59: Can you add more info in your comment about what this api does, like handling partial complete and full delete scenario.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
             metrics.deleteBlobTotalTimeInMs, null, null, totalTimeSpent));
   }
 
+  /**
+   * Handles a {@link BatchDeleteRequest} and sends out the response via the {@link RequestResponseChannel}.
```
manbearpig1996, line 523: Can you add more info in your comment about what this api does, like handling partial complete and full delete scenario.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+   * @param request the {@link NetworkRequest} that contains the {@link BatchDeleteRequest}.
+   * @throws IOException if there is an I/O error reading from the {@link NetworkRequest}.
+   * @throws InterruptedException if the operation is interrupted.
+   */
+  @Override
+  public void handleBatchDeleteRequest(NetworkRequest request) throws IOException, InterruptedException {
```
manbearpig1996, line 529: Can you add comment on cod where you are handling partial error scenarios

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+    logger.info("Batch delete request received.");
```
manbearpig1996, line 530: Can we add trace logs in the condition where you are handling partial error scenarios ?

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+    BatchDeleteRequest batchDeleteRequest;
+    if (request instanceof LocalChannelRequest) {
+      // This is a case where handleDeleteRequest is called when frontends are talking to Azure. In this case, this method
+      // is called by request handler threads running within the frontend router itself. So, the request can be directly
```
manbearpig1996, line 534: Can you add more comments in code for understanding purpose ?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -772,6 +775,183 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
     }
   }
 
+  @Override
+  public StoreBatchDeleteInfo batchDelete(List<MessageInfo> infosToDelete) throws StoreException {
```
manbearpig1996, line 779: Can you add comments here too with trace logs ?

DevenWalia (commented):
Review phase 1 done.
Pending - AmbryRequests, BlobStore, BlobStoreTest

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+        batchDeleteRequest.accept(metricsUpdater);
+        for (BlobDeleteStatus blobDeleteStatus: partitionResponseInfo.getBlobsDeleteStatus()){
+          // Set operation level error to Bad_Request if any of the blobs fails
+          ServerErrorCode error = blobDeleteStatus.getStatus();
+          if (error != ServerErrorCode.No_Error) {
+            // Handling for control test when request type is disabled + setting default to Bad_Request for any error.
+            operationServerErrorCode = error==ServerErrorCode.Temporarily_Disabled?ServerErrorCode.Temporarily_Disabled:ServerErrorCode.Bad_Request;
+            boolean logInErrorLevel = false;
+            if (error == ServerErrorCode.Blob_Not_Found) {
+              metrics.idNotFoundError.inc();
```
skandasa23, line 609: Can we also please add one more kind of metrics that has counters at BatchDelete level? Will be useful to understand how the batchdelete is trending.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+              logger.trace("Store exception on a delete within a batch delete operation with error code {} for request {}", error, batchDeleteRequest);
+            }
+          }
+        }
+        partitionResponseInfoList.add(partitionResponseInfo);
+
+      }
+    }
+    response = new BatchDeleteResponse(batchDeleteRequest.getCorrelationId(), batchDeleteRequest.getClientId(), partitionResponseInfoList, operationServerErrorCode);
+    logger.info(response.toString());
```
skandasa23, line 632: This might not be needed as well.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+    requestResponseChannel.sendResponse(response, request,
+    new ServerNetworkResponseMetrics(metrics.deleteBlobResponseQueueTimeInMs, metrics.deleteBlobSendTimeInMs,
```
skandasa23, line 634: If you are planning to add metrics at batchdelete level, think its more appropriate to use them here?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -772,6 +775,183 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
...
+      List<IndexValue> originalPuts = new ArrayList<>();
+      List<Short> lifeVersions = new ArrayList<>();
+      Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
+      filteredInfos = new ArrayList<>();
+      for (MessageInfo info : infosToDelete) {
+        try {
+          IndexValue value =
+              index.findKey(info.getStoreKey(), new FileSpan(index.getStartOffset(), indexEndOffsetBeforeCheck));
+          if (value == null) {
+            throw new StoreException(
```
skandasa23, line 798: Seems the infoToDelete are iterated twice

    validate and filter only the qualified Ids
    Issue delete for the ids identified via#1

Iteration(1) seems to be throwing error when an id is not present in index..etc, think we need to handle this exception local to the function and ensure the batchDelete operation in attempted in fullest.

skandasa23 (changes requested):

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
DevenWalia, line 779: Can we look at breaking this function in smaller pieces?
Currently it stands at 180+ loc.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -772,6 +775,183 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
...
+    final Timer.Context context = metrics.batchDeleteResponse.time();
+    int i = 0;
+    List<MessageInfo> filteredInfos = null;
+    try {
+      List<IndexValue> indexValuesPriorToDelete = new ArrayList<>();
+      List<IndexValue> originalPuts = new ArrayList<>();
+      List<Short> lifeVersions = new ArrayList<>();
+      Offset indexEndOffsetBeforeCheck = index.getCurrentEndOffset();
+      filteredInfos = new ArrayList<>();
+      for (MessageInfo info : infosToDelete) {
```
DevenWalia, line 793: This corresponds to `infoList` which further corresponds to `convertedStoreKeys` in calling method in `AmbryRequests` class.
Do we have a upper bound on how big this list can be? 

Per design, there needs to be an upper bound to these batches.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -772,6 +775,183 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
...
+              index.findKey(info.getStoreKey(), new FileSpan(index.getStartOffset(), indexEndOffsetBeforeCheck));
+          if (value == null) {
+            throw new StoreException(
+                "Cannot delete id " + info.getStoreKey() + " because it is not present in the index",
+                StoreErrorCodes.ID_Not_Found);
+          }
+          if (!info.getStoreKey().isAccountContainerMatch(value.getAccountId(), value.getContainerId())) {
+            if (config.storeValidateAuthorization) {
+              throw new StoreException(
+                  "DELETE authorization failure. Key: " + info.getStoreKey() + "Actually accountId: "
```
DevenWalia, line 805: nit: Better to create the error string as a separate statement.

```
string error = "DELETE authorization failure. Key: " + info.getStoreKey() + "Actually accountId: " + value.getAccountId() + "Actually containerId: " + value.getContainerId();

if (config.storeValidateAuthorization) {
    throw new StoreException(error);
} else {
    logger.warn(error);
}
```

That way the error strings stay consistent & can evolve consistently across both places.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -772,6 +775,183 @@ public void delete(List<MessageInfo> infosToDelete) throws StoreException {
...
+          }
+          infosToDelete = filteredInfos;
+          if (infosToDelete.isEmpty()){
+            logger.trace("Batch Delete completely failed since all blobs failed");
+            return storeBatchDeleteInfo;
+          }
+        }
+        List<InputStream> inputStreams = new ArrayList<>(infosToDelete.size());
+        List<MessageInfo> updatedInfos = new ArrayList<>(infosToDelete.size());
+        i = 0;
```
DevenWalia, line 909: How are we using `i` ?
It was being incremented in previous blocks, then set to 0 here and its being incremented again in next blocks but we don't seem to use the value anywhere?

DevenWalia (changes requested):
Reviewed AmbryRequests, BlobStore changes

High level comments 
1. Break functions into smaller chunks. The 180+ loc BlobStore::BatchDelete function is hard to parse.

ambry-api/src/main/java/com/github/ambry/store/MessageErrorInfo.java
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.store;
+
+public class MessageErrorInfo {
```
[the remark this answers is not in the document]
piyujai: @DevenWalia We will need this class to be public since it is being accessed outside the com.github.ambry.store package.
Agreed on the lombok annotations, will look to including this new dependency and integrate it atleast for the new classes.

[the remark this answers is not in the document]
DevenWalia: Yea, wherever possible, try to use package-private access for new classes.
There might be a whole lot of places where this cannot be done due to the earlier way of packaging classes, which is fine.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+    boolean operationSuccessful = true;
+    ServerErrorCode operationServerErrorCode = ServerErrorCode.No_Error;
+    List<BatchDeletePartitionResponseInfo> partitionResponseInfoList = new ArrayList<>();
+    BatchDeleteResponse response = new BatchDeleteResponse(batchDeleteRequest.getCorrelationId(), batchDeleteRequest.getClientId(), new ArrayList<>(), ServerErrorCode.No_Error);
+    for (BatchDeletePartitionRequestInfo batchDeletePartitionRequestInfo : batchDeleteRequest.getPartitionRequestInfoList()) {
+      BatchDeletePartitionResponseInfo partitionResponseInfo = null;
+      try {
+        List<StoreKey> convertedStoreKeys = getConvertedStoreKeys(batchDeletePartitionRequestInfo.getBlobIds());
+        ServerErrorCode error =
+            validateRequest(batchDeletePartitionRequestInfo.getPartition(), RequestOrResponseType.BatchDeleteRequest,
```
mudit-saxena, line 552: This method is name confusing. It is validating partition and whether PUT request is going to READ only partition. Also, FYI @piyujai it doesn't use `skipPartitionAndDiskAvailableCheck` in case you are relying on this check for disk and partition availability

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+        } else {
+          List<MessageInfo> infoList = new ArrayList<>();
+          for (StoreKey convertedStoreKey : convertedStoreKeys) {
+            BlobId convertedBlobId = (BlobId) convertedStoreKey;
+            infoList.add(new MessageInfo.Builder(convertedBlobId, -1, convertedBlobId.getAccountId(),
+                convertedBlobId.getContainerId(), batchDeleteRequest.getDeletionTimeInMs()).isDeleted(true)
+                .lifeVersion(MessageInfo.LIFE_VERSION_FROM_FRONTEND)
+                .build());
+          }
+          Store storeToDelete = storeManager.getStore(batchDeletePartitionRequestInfo.getPartition());
```
mudit-saxena, line 570: `storeToDelete` gives a signal that we are deleting this store. Maybe rename this to something like `targetStore`?

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -508,6 +519,122 @@ public void handleDeleteRequest(NetworkRequest request) throws IOException, Inte
...
+            for (StoreKey convertedStoreKey : convertedStoreKeys) {
+              notification.onBlobReplicaDeleted(currentNode.getHostname(), currentNode.getPort(),
+                  convertedStoreKey.getID(), BlobReplicaSourceType.PRIMARY);
+            }
+          }
+        }
+      } catch (Exception e) {
+        List<BlobDeleteStatus> blobDeleteStatuses = new ArrayList<>();
+        for (StoreKey storeKey : batchDeletePartitionRequestInfo.getBlobIds()){
+          blobDeleteStatuses.add(new BlobDeleteStatus((BlobId)storeKey, ServerErrorCode.Bad_Request));
```
mudit-saxena, line 589: There might be cases of partial failure and success, isn't it?

ambry-commons/src/main/java/com/github/ambry/commons/ServerMetrics.java
```diff
@@ -399,6 +403,11 @@ public ServerMetrics(MetricRegistry registry, Class<?> requestClass, Class<?> se
     deleteBlobProcessingTimeInMs = registry.histogram(MetricRegistry.name(requestClass, "DeleteBlobProcessingTime"));
     deleteBlobResponseQueueTimeInMs =
         registry.histogram(MetricRegistry.name(requestClass, "DeleteBlobResponseQueueTime"));
+    batchDeleteBlobRequestQueueTimeInMs =
+        registry.histogram(MetricRegistry.name(requestClass, "BatchDeleteBlobRequestQueueTime"));
```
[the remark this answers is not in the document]
piyujai: Possible to elaborate? @mudit-saxena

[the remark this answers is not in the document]
mudit-saxena: Should be ok, having a prefix helps when looking at the metric

mudit-saxena (approved):
GTG

manbearpig1996 (approved):

DevenAhluwalia (approved):
🚢

PR #2846, opened 2024-07-30
[vcr-2.0] Fix error handling

snalli:
We are getting too many 503s from Azure and that is flooding the logs with giant stack traces because we throw the error up. This patch eliminates that and only throws up the error message. We don't need the full stack.


justinlin-linkedin (approved):
LGTM

PR #2850, opened 2024-08-02
[vcr-2.0] Remove repl-until-time field

snalli:
This field ReplicateUntilUTC was meant to tell us
how far a partition has been copied to cloud. But this field is unreliable. This time is derived by getting the largest timestamp from a stream of messages from the server. But the server combines all messages pertaining to a blob into a single message with the timestamp of the last message embedded into that single message.

For example, consider this stream of messages in the server log segment.

t1(put-b1) | t2(put-b2) | t3(ttl-b1)

Assume the server sends two message m1 and m2

m1: t3(put-b1, ttl-b1)
m2: t2(put-b2)

When we receive m1, we'd think we have copied the partition until t3. But then we receive m2 and we think we have copied till t2. The time is not monotonically increasing. Hence unreliable.


justinlin-linkedin (approved):
LGTM

PR #2853, opened 2024-08-02
Metric for idle time for remote replica group

manbearpig1996:
Publishing metric for idle time for remote replica group i.e. time difference b/w when group is finished and when cycle is finished.
Added metric for cumulative time groups spend in a cycle idle (doing nothing i.e time after group is done)
Added metric for percentage of time groups spend in a cycle idle (total cumlative time / (groups count * total cycle time))

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -1975,7 +1975,7 @@ public class ReplicaThread implements Runnable {
     private long getRequestStartTimeMs;
     private long exchangeMetadataStartTimeInMs;
     private long fixMissingStoreKeysStartTimeInMs;
-    long finishTime;
+    private long finishTime;
 
     /**
      * Constructor of {@link RemoteReplicaGroup}.
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -476,7 +476,7 @@ public class ReplicaThread implements Runnable {
 
     long totalIdleTime = 0;
     for (RemoteReplicaGroup group : groups) {
-      totalIdleTime = totalIdleTime + group.getFinishTime() - roundStartTime;
+      totalIdleTime = totalIdleTime + roundEndTime - group.getFinishTime();
     }
     replicationMetrics.updateReplicaBatchTotalIdleTimeMs(totalIdleTime, replicatingFromRemoteColo, datacenterName);
 
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -374,12 +374,9 @@ public class ReplicaThread implements Runnable {
   public void replicate() {
     exchangeMetadataResponsesInEachCycle = new HashMap<>();
     long oneRoundStartTimeMs = time.milliseconds();
-    Long fastestReplicaFinishTime = null;
     logger.trace("Thread name: {} Start RemoteReplicaGroup replication", threadName);
     List<RemoteReplicaGroup> remoteReplicaGroups = new ArrayList<>();
     int remoteReplicaGroupId = 0;
-    Map<Integer, Long> groupFinishTime = new HashMap<>();
-
     try {
       // Before each cycle of replication, we clean up the cache in key converter.
       storeKeyConverter.dropCache();
@@ -440,12 +437,9 @@ public class ReplicaThread implements Runnable {
         // still be able to handle duplicate response infos for the same request.
         responseInfos.addAll(responseInfosForTimedOutRequests);
         onResponses(responseInfos, correlationIdToRequestInfo, correlationIdToReplicaGroup);
-        if ((fastestReplicaFinishTime == null) && remoteReplicaGroups.stream().anyMatch(RemoteReplicaGroup::isDone)) {
-          fastestReplicaFinishTime = time.milliseconds();
-        }
-        remoteReplicaGroups.stream().filter(RemoteReplicaGroup::isDone).forEach(group -> {
-          groupFinishTime.putIfAbsent(group.getId(), time.milliseconds());
-        });
+        remoteReplicaGroups.stream()
+            .filter(RemoteReplicaGroup::shouldSetFinishTime)
+            .forEach(group -> group.setFinishTime(time.milliseconds()));
       }
       logger.trace("Thread name: {} Finish all RemoteReplicaGroup replication", threadName);
       remoteReplicaGroups.stream()
@@ -462,18 +456,28 @@ public class ReplicaThread implements Runnable {
     } catch (Throwable e) {
       logger.error("Thread name: {} found some error while replicating from remote hosts", threadName, e);
     } finally {
-      replicationMetrics.updateOneCycleReplicationTime(time.milliseconds() - oneRoundStartTimeMs,
-          replicatingFromRemoteColo, datacenterName);
-      groupFinishTime.forEach((groupId, finishTime) -> {
-        replicationMetrics.updateRemoteReplicaGroupIdleTime(time.milliseconds() - finishTime, replicatingFromRemoteColo,
-            datacenterName);
-      });
-      if (fastestReplicaFinishTime != null) {
-        replicationMetrics.updateFastestSlowestReplicaPerCycleTimeDifference(
-            time.milliseconds() - fastestReplicaFinishTime, replicatingFromRemoteColo, datacenterName);
-      }
+      emmitReplicationCycleMetrics(remoteReplicaGroups, oneRoundStartTimeMs, time.milliseconds());
     }
     maybeSleepAfterReplication(remoteReplicaGroups.isEmpty());
+  }
+
+  private void emmitReplicationCycleMetrics(List<RemoteReplicaGroup> groups, long roundStartTime,
+      long roundEndTime) {
+
+    replicationMetrics.updateOneCycleReplicationTime(roundEndTime - roundStartTime, replicatingFromRemoteColo,
+        datacenterName);
+
+    long totalIdleTime = 0;
+
+    for (RemoteReplicaGroup group : groups) {
+      totalIdleTime = totalIdleTime + group.getFinishTime() - roundStartTime;
+    }
+
+    replicationMetrics.updateReplicaBatchTotalIdleTimeMs(totalIdleTime, replicatingFromRemoteColo, datacenterName);
+
+    long idleTimePercentage = (totalIdleTime * 100) / (groups.size() * (roundEndTime - roundStartTime));
+    replicationMetrics.updateReplicaBatchTotalIdleTimePercentage(idleTimePercentage, replicatingFromRemoteColo,
+        datacenterName);
   }
 
   void setExchangeMetadataListener(ExchangeMetadataListener exchangeMetadataListener) {
@@ -1963,6 +1967,7 @@ public class ReplicaThread implements Runnable {
     private long getRequestStartTimeMs;
     private long exchangeMetadataStartTimeInMs;
     private long fixMissingStoreKeysStartTimeInMs;
+    long finishTime;
 
     /**
      * Constructor of {@link RemoteReplicaGroup}.
@@ -1979,6 +1984,7 @@ public class ReplicaThread implements Runnable {
       this.remoteDataNode = dataNodeId;
       this.isNonProgressStandbyReplicaGroup = isNonProgressStandbyReplicaGroup;
       this.id = id;
+      finishTime = 0;
       if (isNonProgressStandbyReplicaGroup) {
         exchangeMetadataResponseList = remoteReplicaInfos.stream()
             .map(remoteReplicaInfo -> new ExchangeMetadataResponse(remoteReplicaInfo.getExchangeMetadataResponse()))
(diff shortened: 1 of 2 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -259,6 +261,8 @@ public ReplicationMetrics(MetricRegistry registry, List<? extends ReplicaId> rep
         registry.histogram(MetricRegistry.name(ReplicaThread.class, "IntraColoOneCycleReplicationTime"));
     intraColoFastestSlowestReplicaPerCycleTimeDifference =
         registry.histogram(MetricRegistry.name(ReplicaThread.class, "IntraColoFastestSlowestReplicaPerCycleTimeDifference"));
+    intraColoRemoteReplicaGroupIdleTime =
+        registry.histogram(MetricRegistry.name(ReplicaThread.class, "IntraColoRemoteReplicaGroupIdleTime"));
```
mudit-saxena, line 265: For brevity you can suffix the time unit in metric name like `IntraColoRemoteReplicaGroupIdleTimeMs`

manbearpig1996: All the time units in across all metrics in ambry replication are in ms.
No other time metrics is having time unit suffix. This is an existing code smell. So can we keep it as such ?

snalli: Please add Ms suffix because when we are reading code, knowing the suffix is immensely useful rather than digging into the code to understand the units. And when we are setting the configs, we will instantly know what value to set based on the suffix. Excluding suffixes in the past was wrong.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -461,6 +464,10 @@ public void replicate() {
     } finally {
       replicationMetrics.updateOneCycleReplicationTime(time.milliseconds() - oneRoundStartTimeMs,
           replicatingFromRemoteColo, datacenterName);
+      groupFinishTime.forEach((groupId, finishTime) -> {
```
snalli, line 467: Ms granularity is too fine for this metric. We don't care at that granularity. There is not much we can do and not worth optimizing for milliseconds. We care if a group idle for a minute or more or even many seconds. Since we don't know that yet, I'd say measure this in seconds.

snalli, line 467: Move it out of the finally block and put it right outside the loop. We don't want to measure any exception cases.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
snalli: On second thought, measure it in seconds. suffix = Secs.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -704,6 +711,14 @@ public void updateFastestSlowestReplicaPerCycleTimeDifference(long timeDifferenc
     }
   }
 
+  public void updateRemoteReplicaGroupIdleTime(long idleTime, boolean remoteColo, String datacenter) {
```
snalli, line 714: private

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -443,6 +440,12 @@ public void replicate() {
         // still be able to handle duplicate response infos for the same request.
         responseInfos.addAll(responseInfosForTimedOutRequests);
         onResponses(responseInfos, correlationIdToRequestInfo, correlationIdToReplicaGroup);
+        if ((fastestReplicaFinishTime == null) && remoteReplicaGroups.stream().anyMatch(RemoteReplicaGroup::isDone)) {
```
snalli, line 443: You may remove this metric.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -443,6 +440,12 @@ public void replicate() {
...
+          fastestReplicaFinishTime = time.milliseconds();
+        }
+        remoteReplicaGroups.stream().filter(RemoteReplicaGroup::isDone).forEach(group -> {
+          groupFinishTime.putIfAbsent(group.getId(), time.milliseconds());
+        });
```
snalli, line 448: We wanna look at the median and max idle time of a each batch, because replication runs in batches currently. Plotting individual idle times, I imagine, doesn't give us an accurate picture. n-1 replicas could finish together and so their idle-times would be small. While 1 replica finished first and the p90 would be close to it, which isn't quite right. 

Emit ReplicaGroupMedianIdleTimeSecs and ReplicaGroupMaxIdleTimeSecs - just two metrics per batch. That tells the number of seconds for which 50% of groups were idle in a batch.

snalli: On second thought, to make it easier, you could sum up the idle time for a batch of ReplicaGroups.

skandasa23: Think we can keep it simple, Sum(idle_time_of_individual_group)  divided by
((cycle_finish_time-cycle_start_time) * number_of_groups). And think it should be fine to have these emitted in milliseconds.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
skandasa23: @snalli @manbearpig1996 I understand the claim from skew perspective. How often we get exceptions? If its close to 0, it should be okay to ignore. If its significant, can we have the exception time as the cycle finish time and still publish the wait time?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
snalli: I agree with keeping it simple and I think `Sum(idle_time_of_individual_group)` is simpler than `Sum(idle_time_of_individual_group) divided by ((cycle_finish_time-cycle_start_time) * number_of_groups)`.  What does the latter mean ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -378,6 +378,7 @@ public void replicate() {
     logger.trace("Thread name: {} Start RemoteReplicaGroup replication", threadName);
     List<RemoteReplicaGroup> remoteReplicaGroups = new ArrayList<>();
     int remoteReplicaGroupId = 0;
+    Map<Integer, Long> groupFinishTime = new HashMap<>();
```
snalli, line 381: minor comment: In general, consider moving away from maps and think in terms of objects. Maps are useful for competitive programming but objects are the norm in shared code bases for several reasons. I am not too particular about it in this case.

For example, this metric could have been added as follows.

```
class RemoteReplicaGroup {
   long idleTime = 0;
   public void incIdleTime(long delta) {...}
   public long getIdleTime() {...}
}
```

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
skandasa23: Among the total available time, how much of time was wasted being idle. 
Total available time = ((cycle_finish_time-cycle_start_time) * number_of_groups).

snalli: `((cycle_finish_time-cycle_start_time) * number_of_groups)` is not the total time available because its not strictly a sequential execution i.e the loop is not waiting for one group to finish before moving to the next. The loop is sending out messages for one group after another without waiting for responses. It's parallel, sorta.
 
The actual total time is just `cycle_finish_time-cycle_start_time`, assuming these are times when the loop starts and ends. I see that you are trying to go for a percentage but the problem with percentages is that it does not tell us absolute value. For eg. if the outcome is that we wasted 50% of the time idling - which on the face of it sounds like a lot but isn't very much if the loop actually just took 100ms, and that's hidden.

Let's keep it simple for now and go for a sum. I am trying to get some clear visibility. If the total idle time is small, like a few millisecs, then its not worth optimizing replication. If the total idle time is large, in the order of seconds or minutes, then we can proceed to the next phase and optimize replication without chopping up the metric further or we could refine it if needed.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: Added in Ms, while plotting we can change in Seconds. Seeing fractional values are not supported in our Histogram.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
manbearpig1996: This needs to be public as it is in ReplicationMetrics class and we will call it from ReplicaThread

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: Removed.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: Agreed on your point , I have added cumulative and percentage metrics both.
Even if cumulative metrics comes out to be 10 seconds, but if we have 1000 groups it will have less impact than if we have 1 second idle time for 1 group. But if cumulative is too less, we should also consider behaviour then. So I have added both metric so we have better data points.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: Added like this.
Will try to follow object oriented design and not rely on individual data structures from now on.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
skandasa23: @snalli Get your point, but for the very same reason thought percentage could be helpful. Thought we are already having the metrics for cycle duration. As long as we have the data points for the aspects we are talking about and could take right decisions, it should be sufficient. Thanks!

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -459,16 +456,38 @@ public void replicate() {
...
+   * @param roundStartTime start time of replication cycle
+   * @param roundEndTime end time of replication cycle
+   */
+  private void emitReplicationCycleMetrics(List<RemoteReplicaGroup> groups, long roundStartTime, long roundEndTime) {
+    replicationMetrics.updateOneCycleReplicationTime(roundEndTime - roundStartTime, replicatingFromRemoteColo,
+        datacenterName);
+
+    long totalIdleTime = 0;
+    for (RemoteReplicaGroup group : groups) {
+      totalIdleTime = totalIdleTime + group.getFinishTime() - roundStartTime;
```
snalli, line 479: This formula is wrong. Please fix it.
Idle time is the time between when the group was Done and when the loop ended.
totalIdleTime += endTime - group.finishTime

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -459,16 +456,38 @@ public void replicate() {
...
+    }
+    replicationMetrics.updateReplicaBatchTotalIdleTimeMs(totalIdleTime, replicatingFromRemoteColo, datacenterName);
+
+    long cumulativeTime = groups.size() * (roundEndTime - roundStartTime);
```
snalli, line 483: cumulative time is the total time for which the loop ran which is just endtime - starttime.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
snalli: @skandasa23 Thanks, I understand now why its multiplied by number of groups.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
manbearpig1996: Fixed this in code.

snalli (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -443,6 +437,9 @@ public void replicate() {
         // still be able to handle duplicate response infos for the same request.
         responseInfos.addAll(responseInfosForTimedOutRequests);
         onResponses(responseInfos, correlationIdToRequestInfo, correlationIdToReplicaGroup);
+        remoteReplicaGroups.stream()
+            .filter(RemoteReplicaGroup::shouldSetFinishTime)
+            .forEach(group -> group.setFinishTime(time.milliseconds()));
```
mudit-saxena, line 442: Do we set finishTime anywhere else as well?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -459,16 +456,38 @@ public void replicate() {
...
+  /**
+   * Emits metrics for
+   * 1. Replication cycle duration
+   * 2. Cumulative time groups are idle in a cycle i.e groups are in DONE state but cycle continues
+   * 3. Cumulative percentage of time per cycle groups are idle
+   * @param groups list of RemoteReplicaGroups
+   * @param roundStartTime start time of replication cycle
+   * @param roundEndTime end time of replication cycle
+   */
+  private void emitReplicationCycleMetrics(List<RemoteReplicaGroup> groups, long roundStartTime, long roundEndTime) {
```
mudit-saxena, line 473: Nit : `roundStartTime` -> `cycleStartTime` & `roundEndTime` -> `cycleEndTime`. This will keep the naming coherent with `replicationCycle` we are using everywhere else

manbearpig1996: changed in the code.

mudit-saxena (approved):
GTG

PR #2866, opened 2024-08-15
[vcr-2.0] Delay replica-thread start and token init

snalli:
There is a burst of calls to Azure to reload replica tokens when a node starts because helix adds replicas very quickly. Due to this, there are a lot of errors resulting in a token reset. To avoid this, let's add a small random delay before a replica thread starts and delay the token initialization until necessary.

Also remove a bunch of unwanted configs and code.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
@@ -117,6 +117,10 @@ public class VcrReplicaThread extends ReplicaThread {
     super.run();
   }
 
+  /**
+   * Fetch replica-token from Azure and set it in the thread
+   * @param replicaInfo
+   */
   void setReplicaToken(RemoteReplicaInfo replicaInfo) {
     String partitionKey = String.valueOf(replicaInfo.getReplicaId().getPartitionId().getId());
     String rowKey = replicaInfo.getReplicaId().getDataNodeId().getHostname();
@@ -127,6 +131,7 @@ public class VcrReplicaThread extends ReplicaThread {
         logger.warn("Resetting token for replica {}/{} because no token found", partitionKey, rowKey);
         replicaInfo.setToken(findTokenFactory.getNewFindToken());
       } else {
+        logger.info("Loading token for replica {}/{}", partitionKey, rowKey);
         DataInputStream inputStream = new DataInputStream(
             new ByteArrayInputStream((byte[]) row.getProperty(VcrReplicationManager.BINARY_TOKEN)));
         replicaInfo.setToken(findTokenFactory.getFindToken(inputStream));
@@ -223,10 +228,10 @@ public class VcrReplicaThread extends ReplicaThread {
       return;
     }
     if (token.equals(oldToken)) {
-      logger.info("Skipping token upload due to unchanged token, oldToken = {}, newToken = {}", oldToken, token);
+      logger.trace("Skipping token upload due to unchanged token, oldToken = {}, newToken = {}", oldToken, token);
       return;
     }
-    logger.info("[snkt] replica = {}, token = {}", remoteReplicaInfo, token);
+    logger.trace("Uploading token = {} for replica = {}", token, remoteReplicaInfo);
     // Table entity = Table row
     // ======================================================================================
     // | partition-key | row-key | tokenType | logSegment | offset | storeKey | binaryToken |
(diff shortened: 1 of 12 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
```diff
@@ -99,6 +101,44 @@ public int compare(RemoteReplicaInfo r1, RemoteReplicaInfo r2) {
     }
   }
 
+  @Override
+  public void run() {
+    try {
+      int delay = new Random().nextInt(900) + 1;
+      logger.info("Starting replica thread {} in {} seconds", Thread.currentThread().getName(), delay);
+      Thread.sleep(TimeUnit.SECONDS.toMillis(delay));
```
justinlin-linkedin, line 109: let's use a configuration here. just in case we want to change it in the future.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -198,16 +193,22 @@ protected int getNumReplThreads(double cpuScale) {
     return (int) (Double.valueOf(Runtime.getRuntime().availableProcessors()) * cpuScale);
   }
 
+  @Override
+  protected void runThread(ReplicaThread replicaThread) {
+    Utils.daemonThread(replicaThread.getName(), replicaThread).start();
+    logger.info("Started replica thread {}", replicaThread.getName());
+  }
```
justinlin-linkedin, line 200: Why do we run vcr replication thread as daemon thread? It doesn't make sense. We should run it as user thread.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -168,6 +159,10 @@ public VcrReplicationManager(VerifiableProperties properties, StoreManager store
     this.cloudDestination.getTableClient(this.azureCloudConfig.azureTableNameCorruptBlobs);
     azureTableNameReplicaTokens = this.azureCloudConfig.azureTableNameReplicaTokens;
     this.cloudDestination.getTableClient(azureTableNameReplicaTokens);
+    String datacenter = clusterMap.getDatacenterName(clusterMap.getLocalDatacenterId());
+    logger.info("Local datacenter id = {}, Local datacenter name = {}", clusterMap.getLocalDatacenterId(), datacenter);
+    replicaThreadPoolByDc.put(datacenter,
+        createThreadPool(datacenter, getNumReplThreads(cloudConfig.backupNodeCpuScale), true));
```
justinlin-linkedin, line 165: I understand what you are trying to do here. VCR should only replicate from local datacenter. And you only create threads for local datacenter.

But we are using a configuration to control which datacenter to replicate from for vcr, which should only contains local datacenter in configuration. I feel that we should just remove the configuration and only rely on the logic where we just need to replicate from local datacenter. 

At least let's just put a comment here saying that there can't be any other datacenter in the configuration, otherwise, vcr won't run.

snalli: got it !

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
snalli: My idea was to reduce the shutdown time for vcr. If the main thread exits and the others are daemon, then jvm will exit immediately. If this isn't right, i can change it back.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicaThread.java
snalli: done

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
snalli: done

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
justinlin-linkedin: Let's not use daemon thread here.

justinlin-linkedin (approved):

PR #2867, opened 2024-08-17
Fix missing data chunk metric

justinlin-linkedin:
## Summary

We have a missing data chunk error count metric to record the number of get blob operations that encountered a missing data chunk error. However, this metric is incremented in the wrong place, we increment it every time we see a NOT_FOUND response from any replica, this is not very useful.

This PR fixes that and only increments this counter metric at the end of get chunk operation when it fails with Blob_Not_Found error. 

##Test

Unit test


MikeDafi (approved):
I don't see in the codebase where `missingDataChunkErrorCount` was being incremented on `NOT_FOUND` as you mentioned in the PR Description, but this logic makes sense despite

PR #2869, opened 2024-08-17
[vcr-2.0] Fix duplicate uploads

snalli:
We were attempting duplicate uploads because of a typo
in the findMissingKeys fn. This patch fixes that and also
adds some guard code to prevent that and debug it if it happens
again.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
@@ -229,12 +229,8 @@ public class CloudBlobStore implements Store {
     // Keys may be missing if we are here, we are here to find out
     for (StoreKey key : keys) {
       try {
-        cloudDestination.getBlobMetadata(Collections.singletonList((BlobId) key));
-      } catch (CloudStorageException e) {
-        if (e.getStatusCode() == HttpStatus.SC_NOT_FOUND) {
+        if (!cloudDestination.getBlobMetadata(Collections.singletonList((BlobId) key)).containsKey(key.getID())) {
           missingKeys.add(key);
-        } else {
-          throw new StoreException(e.getMessage(), StoreErrorCodes.IOError);
         }
       } catch (Throwable e) {
         throw new StoreException(e.getMessage(), StoreErrorCodes.IOError);
@@ -255,15 +251,16 @@ public class CloudBlobStore implements Store {
       // Key must exist if we are here
       Map<String, CloudBlobMetadata> cloudBlobMetadataListMap =
           cloudDestination.getBlobMetadata(Collections.singletonList((BlobId) key));
+      if (!cloudBlobMetadataListMap.containsKey(key.getID())) {
+        // If we are here, the key must exist
+        throw new StoreException(String.format("%s not found", key), StoreErrorCodes.ID_Not_Found);
+      }
       CloudBlobMetadata cloudBlobMetadata = cloudBlobMetadataListMap.get(key.getID());
       return new MessageInfo(key, cloudBlobMetadata.getSize(), cloudBlobMetadata.isDeleted(),
           cloudBlobMetadata.isTtlUpdated(), cloudBlobMetadata.isUndeleted(), cloudBlobMetadata.getExpirationTime(), null,
           (short) cloudBlobMetadata.getAccountId(), (short) cloudBlobMetadata.getContainerId(),
           cloudBlobMetadata.getLastUpdateTime(), cloudBlobMetadata.getLifeVersion());
     } catch (CloudStorageException e) {
-      if (e.getStatusCode() == HttpStatus.SC_NOT_FOUND) {
-        throw new StoreException(e.getMessage(), StoreErrorCodes.ID_Not_Found);
-      }
       throw new StoreException(e, StoreErrorCodes.IOError);
     }
   }
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
@@ -389,7 +389,7 @@ public class VcrReplicationManager extends ReplicationEngine {
     } catch (Throwable e) {
       // Helix will run into error state if exception throws in Helix context.
       vcrMetrics.addPartitionErrorCount.inc();
-      logger.error("Failed to add Store for partition {} due to {}", partitionId.getId(), e.getMessage());
+      logger.error("Failed to add Store for partition {} due to {}", partitionId.getId(), e);
     } finally {
       rwLock.writeLock().unlock();
     }
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
@@ -110,12 +110,18 @@ public class AzureCloudDestinationSync implements CloudDestination {
   protected class AzureBlobProperties implements AmbryCacheEntry {
 
     private final BlobProperties properties;
-    public AzureBlobProperties(BlobProperties properties) {
+    private final int azureStatus;
+    public AzureBlobProperties(BlobProperties properties, int azureStatus) {
       this.properties = properties;
+      this.azureStatus = azureStatus;
     }
 
     public BlobProperties getProperties() {
       return properties;
+    }
+
+    public int getAzureStatus() {
+      return azureStatus;
     }
   }
   /**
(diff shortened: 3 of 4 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -872,27 +896,48 @@ public CompletableFuture<Short> updateBlobExpirationAsync(BlobId blobId, long ex
...
-        We could use one-liner Azure SDK provided lightweight exists() method, which internally fetches blob-properties
-        and does the same thing that findMissingKeys does.
+    /**
+     * Just keeping this fn around for tests and legacy reasons but imo this is a horrible function.
+     * The fn takes a list as an arg, why not just a single blob ? It is only ever used to fetch metadata of a blob,
+     * not a list. And because of this design, the higher levels have to wrap the blob-id in a singleton-list.
+     * More importantly, the return type is map. The higher levels have to explicitly look for the result in the map
+     * with blob-id. If there is typo in the code in looking up the map, for eg. instead of storeKey.getId(),
+     * if we type storeKey() alone, then we make a mistake of thinking the blob is not found. Just stop using maps!
+     * Such lookup errors are easy to miss in code-reviews. And most legacy tests are broken beyond repair.
```
justinlin-linkedin, line 930: Yeah, this is odd. If parameter is a list of BlobId, then the key of the map should be BlobId as well. Anyway, we are not caling getBlobMetadata with multiple blob ids anyway.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
```diff
@@ -165,13 +165,10 @@ public void delete(List<MessageInfo> infos) throws StoreException {
...
         cloudDestination.deleteBlob((BlobId) msg.getStoreKey(), msg.getOperationTimeMs(), msg.getLifeVersion(), null);
       }
-    } catch (CloudStorageException e) {
-      if (e.getStatusCode() == HttpStatus.SC_NOT_FOUND) {
-        throw new StoreException(e.getMessage(), StoreErrorCodes.ID_Not_Found);
-      }
-      throw new StoreException(e.getMessage(), StoreErrorCodes.IOError);
     } catch (StoreException e) {
       throw e;
+    } catch (Throwable e) {
```
justinlin-linkedin, line 170: Any reason to not have a SC_NOT_FOUND?

snalli: When calling these fn that update some metadata or deleting a blob, the blob must exist. This is checked by replication logic before calling these fn. If the blob doesn't exist for some reason, I changed the code to return a null.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
snalli: Yeah, this fn created some bad code in the upper layers.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureCloudDestinationSync.java
```diff
@@ -872,27 +896,48 @@ public CompletableFuture<Short> updateBlobExpirationAsync(BlobId blobId, long ex
...
+      BlobProperties blobPropertiesCached = getBlobPropertiesCached(blobLayout);
+      if (blobPropertiesCached != null) {
+        return CloudBlobMetadata.fromMap(blobPropertiesCached.getMetadata());
+      }
+    } catch (CloudStorageException cse) {
+      throw cse;
+    } catch (Throwable t) {
+      throw new CloudStorageException(t.getMessage());
+    }
+    return null;
```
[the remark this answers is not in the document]
snalli: null means the blob doesn't exist. The lower layers return `AzureBlobProperties(blob_properties=null, status=not_found...)` at L666. I thought the code became a bit easier to read by returning null. I can consider throwing an exc if a null is encountered in getCloudBlobMetadata.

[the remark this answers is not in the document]
justinlin-linkedin: I think it makes more sense to throw an exception when it doesn't exist.

ambry-cloud/src/main/java/com/github/ambry/cloud/CloudBlobStore.java
justinlin-linkedin: make sense

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -107,6 +107,8 @@ public class VcrReplicationManager extends ReplicationEngine {
   private DistributedLock vcrUpdateDistributedLock = null;
   private final List<ReplicaThread> threadPool;
   private final AtomicInteger threadIndex;
+  private final HashMap<Long, ReplicaThread> partitionIdThreadMap;
+  private final ReadWriteLock partitionIdThreadMapLock = new ReentrantReadWriteLock();
```
justinlin-linkedin, line 111: We probably don't need this new lock , as we have a rwLock defined in ReplicationEngine.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -202,23 +205,88 @@ protected void createThread(ReplicaThread replicaThread, boolean unused) {
...
+   */
+  @Override
+  protected void addRemoteReplicaInfoToReplicaThread(List<RemoteReplicaInfo> remoteReplicaInfos, boolean unused) {
+    /**
+     * This fn must be called with threadMapLock.readLock() held.
+     * It is called by helix when a new replica (not a new partition) is assigned to VCR node.
+     * onReplicaAddedOrRemoved -> addPartitionToReplicaThread
+     *
+     * Cannot use the same rwlock.writeLock because this is called with rwlock.readLock.
+     * Lock upgrades are not allowed.
```
justinlin-linkedin, line 254: you can change rwlock.readLock to rwLock.writeLock in ReplicationEngine. It's weird that modifying states only requires a read lock. It should be write lock anyway.

snalli: i agree, but as the comment in onReplicaAddedOrRemoved indicates, it was done for performance reasons. although i don't think it affects performance drastically since the lock is held for a very short duration.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
snalli: yeah, agree, more locks creates a potential for deadlocks

snalli: removed the lock and the thread-map, tracking replicaThread in partitionInfo.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -202,23 +199,68 @@ protected void createThread(ReplicaThread replicaThread, boolean unused) {
...
+     * This fn must be called with rwlock.writeLock() held.
+     * It is called by helix when a new replica (not a new partition) is assigned to VCR node.
+     * onReplicaAddedOrRemoved -> addPartitionToReplicaThread
+     */
+    try {
+      for (RemoteReplicaInfo rinfo : remoteReplicaInfos) {
+        try {
+          // Replicas here may belong to different partitions, but they must already be assigned to this node
+          PartitionId partitionId = rinfo.getReplicaId().getPartitionId();
+          ReplicaThread rthread = partitionToPartitionInfo.get(partitionId).getReplicaThread();
```
justinlin-linkedin, line 246: I actually prefer to create a separate map for partition id to thread mapping since it's not the same logic used in server-to-server replication. But if you want to do this, it's fine, please add some comments to the PartitionInfo so we know it's only used for VCR replication.

justinlin-linkedin (approved):

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
snalli: Thanks, I thought using another map is fragmenting the information. We have so many maps in the code.

snalli: Added a comment.

PR #2871, opened 2024-08-22
Enable 100-continue for S3 only and control by configs

SophieGuo410:


ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -196,7 +196,8 @@ private void start() {
      */
     private Callback<Void> securityProcessRequestCallback() {
       return buildCallback(frontendMetrics.putSecurityProcessRequestMetrics, securityCheckResult -> {
-        if (CONTINUE.equals(restRequest.getArgs().get(EXPECT))) {
+        if (frontendConfig.oneHundredContinueEnable && CONTINUE.equals(restRequest.getArgs().get(EXPECT))
+            && RestUtils.isS3Request(restRequest)) {
```
Arun-LinkedIn, line 200: QQ. If we are doing only for S3 requests, just wondering if it would be possible move `if` condition for `100 response` inside `S3PutHandler.doHandle()` and return from there itself? If we do it that way, we can avoid S3 checks in `NamedBlobPutHandler`

Arun-LinkedIn (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
SophieGuo410: That's true, but the reason I add to named blob is due to I want the first phase of 100-continue also check the ACLAccess to make sure frontend is accessible.

justinlin-linkedin (approved):

PR #2874, opened 2024-08-28
Allow overwrites by default for S3 uploads

Arun-LinkedIn:
Reference: https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html  (_If it receives multiple write requests for the same object simultaneously, it overwrites all but the last object written._)

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3PutHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3PutHandler.java
@@ -88,6 +88,9 @@ public class S3PutHandler extends S3BaseHandler<Void> {
     }
     restRequest.setArg(Headers.AMBRY_CONTENT_ENCODING, restRequest.getArgs().get(Headers.CONTENT_ENCODING));
 
+    // Overwrites are allowed by default for S3 APIs https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html
+    restRequest.setArg(NAMED_UPSERT, true);
+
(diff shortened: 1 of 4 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3PutHandler.java
```diff
@@ -88,6 +88,9 @@ protected void doHandle(RestRequest restRequest, RestResponseChannel restRespons
     }
     restRequest.setArg(Headers.AMBRY_CONTENT_ENCODING, restRequest.getArgs().get(Headers.CONTENT_ENCODING));
 
+    // Overwrites are allowed by default for S3 APIs https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObject.html
+    restRequest.setArg(NAMED_UPSERT, true);
```
SophieGuo410, line 92: QQ: should we also enable this for multiupload?

SophieGuo410 (approved):

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3PutHandler.java
justinlin-linkedin: right, seems like we need to do this for all uploads.

Arun-LinkedIn: Sure. Thank you for mentioning it! Updated the code

justinlin-linkedin (approved):

PR #2876, opened 2024-08-28
[Paranoid Durability] Limit the number of remote PUT attempts.

litingulfs:
This PR adds a new config and code for limiting the number of times we attempt to PUT to a remote colo. The intention is to limit the memory usage during times of extra high network latency. When the network is slow or failing, ambry-frontend will hold request attempts in memory, which can lead to exhaustion of Java direct memory.

If we reach the limit for remote writes, but local writes are successful, we will still return success to the client. The assumption here is that it is better to degrade performance than to completely fail, since we assume that network latency/error spikes will be fairly brief.

A new metric and log message is added to allow debugging. Also, an old log message is removed because it was found to make the logs noisy without adding much value.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
@@ -49,6 +49,8 @@ public class ParanoidDurabilityOperationTracker extends SimpleOperationTracker {
   protected int remoteInflightCount = 0;
   protected int localReplicaSuccessCount = 0;
   protected int remoteReplicaSuccessCount = 0;
+  protected int remoteAttempts = 0;
+  protected int remoteAttemptLimit = 2;
   private int currentDatacenterIndex = 0;
   int totalLocalReplicaCount;
   int failedLocalReplicas = 0;
@@ -67,6 +69,7 @@ public class ParanoidDurabilityOperationTracker extends SimpleOperationTracker {
     remoteReplicaParallelism = routerConfig.routerPutRemoteRequestParallelism;
     localReplicaSuccessTarget = routerConfig.routerPutSuccessTarget;
     remoteReplicaSuccessTarget = routerConfig.routerPutRemoteSuccessTarget;
+    remoteAttemptLimit = routerConfig.routerPutRemoteAttemptLimit;
 
     // Calling getEligibleReplicas with null will return replicas from all data centers.
     List<ReplicaId> allEligibleReplicas = getEligibleReplicas(null, EnumSet.of(ReplicaState.STANDBY, ReplicaState.LEADER));
@@ -225,7 +228,6 @@ public class ParanoidDurabilityOperationTracker extends SimpleOperationTracker {
    */
   public boolean hasSucceeded() {
     if (hasSucceededLocally() && hasSucceededRemotely()) {
-      logger.info("Paranoid durability PUT succeeded for partition " + partitionId);
       routerMetrics.paranoidDurabilitySuccessCount.inc();
       return true;
     }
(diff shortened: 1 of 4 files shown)


ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -250,7 +252,21 @@ private boolean hasSucceededLocally() {
    * @return true if we have received enough successful responses from remote replicas, false otherwise.
    */
   private boolean hasSucceededRemotely() {
-    return remoteReplicaSuccessCount >= remoteReplicaSuccessTarget;
+    if (remoteReplicaSuccessCount >= remoteReplicaSuccessTarget) {
+      return true;
+    }
+    else if(remoteAttempts >= remoteAttemptLimit) {
```
justinlin-linkedin, line 258: the indentation is a bit weird, it should be
```
} else if (xxx) {
}
```
could you please format this file?

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
```diff
@@ -334,7 +350,13 @@ private boolean hasFailedDynamically()  {
...
    */
   private boolean hasFailedRemotely() {
-    if ((remoteReplicaSuccessCount + remoteInflightCount + remoteReplicas.size()) < remoteReplicaSuccessTarget) {
+    // If we have exceeded the number of attempts to remote colos, we should give up and return success to the client.
+    // This situation may occur at times of excessive network latency, where we may end up issuing a large number of
+    // retries to remote colos. This is undesirable, since it will lead to excessive direct memory usage (incident-1455).
+    if (remoteAttempts >= remoteAttemptLimit) {
+      return false;
+    }
+    else if ((remoteReplicaSuccessCount + remoteInflightCount + remoteReplicas.size()) < remoteReplicaSuccessTarget) {
```
justinlin-linkedin, line 359: same above

ambry-router/src/test/java/com/github/ambry/router/ParanoidDurabilityOperationTrackerTest.java
```diff
@@ -233,6 +235,55 @@ public void justEnoughRemoteReplicasTest() {
...
+          ot.onResponse(currentReplica, TrackedRequestFinalState.TIMED_OUT);
+        } else {
+          ot.onResponse(currentReplica, TrackedRequestFinalState.SUCCESS);
+        }
+        //ot.onResponse(currentReplica, TrackedRequestFinalState.SUCCESS);
+      }
+    }
+    assertTrue("Operation should have succeeded", ot.hasSucceeded());
+    assertTrue("Operation should be done", ot.isDone());
+    assertFalse("Operation should not have failed", ot.hasFailed());
```
justinlin-linkedin, line 282: could you also verify that the counter metric is incremented by one?

justinlin-linkedin (approved):

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: Yeah, thanks for catching this - it's fixed.

ambry-router/src/main/java/com/github/ambry/router/ParanoidDurabilityOperationTracker.java
litingulfs: Yep, also fixed.

ambry-router/src/test/java/com/github/ambry/router/ParanoidDurabilityOperationTrackerTest.java
litingulfs: Yeah, added a check to confirm that the metrics is correctly incremented.

justinlin-linkedin (approved):

PR #2878, opened 2024-08-28
Add blob size and modified timestamp in named blobs metadata db

Arun-LinkedIn:
Flink is requesting to provide blob size and modified timestamp when listing blob information (via s3 list api). 

This PR:
1. Includes blob size in the INSERT query. The modified timestamp is auto populated by the mysql server
2. Gets blob size and modified timestamp as well in SELECT query and includes them in S3 list api output

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobListEntry.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobListEntry.java
@@ -116,6 +116,6 @@ public class NamedBlobListEntry {
     }
     NamedBlobListEntry that = (NamedBlobListEntry) o;
     return expirationTimeMs == that.expirationTimeMs && Objects.equals(blobName, that.blobName)
-        && modifiedTimeMs == that.modifiedTimeMs && blobSize == that.getBlobSize();
+        && modifiedTimeMs == that.modifiedTimeMs && blobSize == that.blobSize;
   }
 }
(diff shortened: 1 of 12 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/AmbryIdConverterFactory.java
```diff
@@ -192,7 +192,7 @@ private CompletionStage<String> convertId(String input, RestRequest restRequest,
         long expirationTimeMs =
             Utils.addSecondsToEpochTime(properties.getCreationTimeInMs(), properties.getTimeToLiveInSeconds());
         NamedBlobRecord record = new NamedBlobRecord(namedBlobPath.getAccountName(), namedBlobPath.getContainerName(),
-            namedBlobPath.getBlobName(), blobId, expirationTimeMs);
+            namedBlobPath.getBlobName(), blobId, expirationTimeMs, 0, properties.getBlobSize(), 0);
```
justinlin-linkedin, line 195: please add a comment here that we don't need modified time passed to database, since it will be default to current_timestamp in database.

Arun-LinkedIn: Sure. Added comments

ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobListEntry.java
```diff
@@ -90,6 +115,7 @@ public boolean equals(Object o) {
       return false;
     }
     NamedBlobListEntry that = (NamedBlobListEntry) o;
-    return expirationTimeMs == that.expirationTimeMs && Objects.equals(blobName, that.blobName);
+    return expirationTimeMs == that.expirationTimeMs && Objects.equals(blobName, that.blobName)
+        && modifiedTimeMs == that.modifiedTimeMs && blobSize == that.getBlobSize();
```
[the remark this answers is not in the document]
Arun-LinkedIn: Sure

justinlin-linkedin (approved):

PR #2879, opened 2024-08-30
Bring in Composite Router

SophieGuo410:
Prototype test passed.


ambry-rest/src/main/java/com/github/ambry/rest/RestServer.java
```diff
@@ -206,9 +206,17 @@ public RestServer(VerifiableProperties verifiableProperties, ClusterMap clusterM
...
-    RouterFactory routerFactory =
-        Utils.getObj(restServerConfig.restServerRouterFactory, verifiableProperties, clusterMap, notificationSystem,
-            routerSslFactory, accountService);
+    RouterFactory routerFactory;
+    if (restServerConfig.restServerEnableCompositeRouter) {
+      RouterFactory ambryRouterFactory =
+          Utils.getObj(restServerConfig.restServerRouterFactory, verifiableProperties, clusterMap,
+              notificationSystem, routerSslFactory, accountService);
+      Router ambryRouter = ambryRouterFactory.getRouter();
+      routerFactory = Utils.getObj(restServerConfig.restServerCompositeRouterFactory, ambryRouter);
```
Arun-LinkedIn, line 215: This is good. I was just looking at [CompositeAccountServiceFactory](https://github.com/linkedin/ambry/blob/master/ambry-account/src/main/java/com/github/ambry/account/CompositeAccountServiceFactory.java#L46) example. Similar to it, we could also have `restServerPrimaryRouterFactory` and `restServerSecondaryRouterFactory`. Then, we wouldn't need `restServerEnableCompositeRouter` config. We could directly point `restServerRouterFactory` to `CompositeRouterFactory` in the configs to enable composite router.

Arun-LinkedIn (approved):

PR #2881, opened 2024-09-03
[Continuous replication PR1] adding trackers classes for continuous replication

manbearpig1996:
Adding tracker classes for Continuous Replication.
We will use these store state for continuous replication


aga9900 (approved):
LGTM

PR #2882, opened 2024-09-04
[Continuous Replication PR 2] Config to switch b/w old and new code path

manbearpig1996:
Adding a config to switch between old and new code path for continuous replication


aga9900 (approved):
LGTM

PR #2884, opened 2024-09-10
[vcr-2.0] Measure Azure Storage Container drift in bytes

snalli:
This patch measures the drift of a Azure Storage container from a Ambry partition in bytes. It introduces a class responsible for aggregating container metrics. We deliberately avoid emitting per-container metrics to prevent an overwhelming increase in the number of metrics, which would strain the telemetry system. This approach has been tried previously and was unsuccessful.

A daemon will run periodically, emitting aggregate metrics in a controlled and predictable manner. The drift of the Azure container from the Ambry partition is set using a compare-and-set mechanism to prevent accidental multithreading errors. However, it's unlikely that multiple threads will handle the same partition in the VCR, as a single thread manages all replicas of a partition using ROUND_ROBIN policy.

We use the `min()` function because bootstrapping replicas can skew the data, falsely indicating a large drift when the partition is fully backed up. If the lag or drift is -1, we round it up to 0 and proceed.

### DIFF
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
@@ -126,7 +126,7 @@ public class VcrReplicationManager extends ReplicationEngine {
     this.azureCloudConfig = new AzureCloudConfig(properties);
     this.vcrMetrics = new VcrMetrics(metricRegistry);
     this.azureMetrics = new AzureMetrics(metricRegistry);
-    this.azureStorageContainerMetricsCollector = AzureStorageContainerMetricsCollector.getInstance(this.azureMetrics);
+    this.azureStorageContainerMetricsCollector = AzureStorageContainerMetricsCollector.getInstance(metricRegistry, properties);
     this.vcrClusterParticipant = vcrClusterParticipant;
     try {
       vcrHelixConfig =
--- a/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
+++ b/ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
@@ -17,8 +17,9 @@ import com.codahale.metrics.MetricRegistry;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.ambry.cloud.azure.AzureCloudConfig;
-import com.github.ambry.cloud.azure.AzureContainerMetrics;
+import com.github.ambry.cloud.azure.AzureStorageContainerMetrics;
 import com.github.ambry.cloud.azure.AzureMetrics;
+import com.github.ambry.cloud.azure.AzureStorageContainerMetricsCollector;
 import com.github.ambry.clustermap.CloudReplica;
 import com.github.ambry.clustermap.ClusterMap;
 import com.github.ambry.clustermap.ClusterMapChangeListener;
@@ -62,8 +63,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -86,6 +85,7 @@ public class VcrReplicationManager extends ReplicationEngine {
   private final VcrMetrics vcrMetrics;
   private final VcrClusterParticipant vcrClusterParticipant;
   private final String localDatacenterName;
+  protected final AzureStorageContainerMetricsCollector azureStorageContainerMetricsCollector;
   protected String azureTableNameReplicaTokens;
   protected AzureCloudConfig azureCloudConfig;
   protected AzureMetrics azureMetrics;
@@ -103,7 +103,6 @@ public class VcrReplicationManager extends ReplicationEngine {
   private DistributedLock vcrUpdateDistributedLock = null;
   private final List<ReplicaThread> threadPool;
   private final AtomicInteger threadIndex;
-  private final ConcurrentHashMap<Long, AzureContainerMetrics> azureContainerMetricsMap;
 
   public static final String BACKUP_NODE = "backupNode";
   public static final String TOKEN_TYPE = "tokenType";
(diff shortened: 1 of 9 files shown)


ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
```diff
@@ -0,0 +1,96 @@
...
+   * Use min() as bootstrapping replicas can give a wrong picture and indicate a large drift even though the partition
+   * is fully backed up in Azure.
+   * @param id
+   * @param drift
+   */
+  public void setContainerDrift(long id, long drift) {
+    AzureStorageContainerMetrics azureContainerMetrics = metricMap.get(id);
+    Long oldDrift = azureContainerMetrics.getDrift();
+    azureContainerMetrics.setDrift(oldDrift, Math.min(oldDrift, drift));
+  }
```
justinlin-linkedin, line 95: When new partition gets added to the map, the drift value in the AtomicLong would be initialized to 0. But here we are use Math.min to get the smaller value of old drift and new drift, which would always be 0. I don't think this logic would work.

justinlin-linkedin: The issue is that 0 is a valid drift value, and we are using it to represent an "uninitialized" value. I recommend to use initialize atomic long to -1 and use it as a special case to the values.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * A class that aggregates container metrics. We do _NOT_ want to emit per-container metrics because then the number
+ * of metrics increases proportionally overwhelming telemetry. This was attempted and failed.
+ *
+ * A daemon will run at regular intervals emitting aggregate metrics in a controlled and predictable manner.
+ * This is a singleton class to avoid multiple collector threads.
+ */
+public class AzureStorageContainerMetricsCollector {
+  private final AzureMetrics metrics;
+  private final ConcurrentHashMap<Long, AzureStorageContainerMetrics> metricMap;
```
justinlin-linkedin, line 35: we actually don't need to create a new `AzureStorageContainerMetrics` class, we can just put an AtomicLong here in the map.

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
```diff
@@ -123,6 +126,7 @@ public VcrReplicationManager(VerifiableProperties properties, StoreManager store
     this.azureCloudConfig = new AzureCloudConfig(properties);
     this.vcrMetrics = new VcrMetrics(metricRegistry);
     this.azureMetrics = new AzureMetrics(metricRegistry);
+    this.azureStorageContainerMetricsCollector = AzureStorageContainerMetricsCollector.getInstance(metricRegistry, properties);
```
justinlin-linkedin, line 129: if we are passing it to every VcrReplicaThread, then why have a static instance method? 
You can just have a regular constructor for this collector class.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
```diff
@@ -0,0 +1,96 @@
...
+   * Sets the drift of azure-container from ambry-partition.
+   * We use a compare-set to guard against accidental multithreaded errors, although two threads will most likely
+   * not be responsible for a single partition in VCR. A single thread handles all replicas of a partition.
+   * However, we want to avoid any races between reader and writers.
+   * Use min() as bootstrapping replicas can give a wrong picture and indicate a large drift even though the partition
+   * is fully backed up in Azure.
+   * @param id
+   * @param drift
+   */
+  public void setContainerDrift(long id, long drift) {
```
justinlin-linkedin, line 91: Drift is probably not a good variable name. I suppose lag would be better given the context here.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
snalli: Good pt ! However, suppose there are 3 replicas with lag as {0, 10, 123456} and assume the last one is a bootstrap replica. I don't want bootstrap replicas skewing this metric as full-auto bootstraps quite a lot. 
I agree with using -1 as an initial value. If we do that, then does it make sense to use `min()` ?

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
snalli: Changed to `lag`

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
snalli: Right, but I imagine there could be more metrics in the future and I am gently guiding the future dev to add metrics here rather than creating more maps. Is that good ?

ambry-cloud/src/main/java/com/github/ambry/cloud/VcrReplicationManager.java
snalli: Good pt ! Changed this such that each thread gets the instance without passing it to every thread ctor.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
snalli: Hmm, i see the problem. Once the lag hits 0, it'll stay at 0.

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
```diff
@@ -0,0 +1,112 @@
...
+      // With improper clean up, these references linger around and cause memory leaks.
+      long pid = rinfo.getReplicaId().getPartitionId().toPathString().hashCode();
+      metricMap.putIfAbsent(pid, new AzureStorageContainerMetrics(pid));
+      metricMap.get(pid).addPartitionReplica(rinfo.getReplicaId().getDataNodeId().getHostname());
+    }
+  }
+
+  public void removePartitionReplicas(List<RemoteReplicaInfo> remoteReplicaInfos) {
+    for (RemoteReplicaInfo rinfo : remoteReplicaInfos) {
+      long pid = rinfo.getReplicaId().getPartitionId().toPathString().hashCode();
```
justinlin-linkedin, line 87: let's not use hashCode here, just call `rinfo.getReplicaId().getPartitionId().getId()`. You are already calling it from VcrReplicaThread.

justinlin-linkedin (changes requested):

ambry-cloud/src/main/java/com/github/ambry/cloud/azure/AzureStorageContainerMetricsCollector.java
snalli: oh good catch, that's a typo !

justinlin-linkedin (approved):

PR #2885, opened 2024-09-10
Readable stream channel that can be read by two destinations

Arun-LinkedIn:
An implementation of a readable stream channel that acts a proxy/buffer between input read channel and two write channels.


Arun-LinkedIn:
This is implemented via PR https://github.com/linkedin/ambry/pull/2898

PR #2886, opened 2024-09-10
Move IdConverter inside router phase1

SophieGuo410:
This pr include:
1.  pass in blobProperties instead of blob Info in idConverter so we don't need to add blobInfo into router interface.
2. pass in idConverter object inside router.
3. Adding the blobPath into router.putBlob interface, so idconverter can parse it and get the namedBlobPath.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
@@ -762,10 +762,29 @@ public class RouterConfig {
   public final boolean routerParanoidDurabilityEnabled;
 
   /**
+   * This is set in frontendConfig until id converter been fully migrate to router.
+   */
+  public final String idConverterFactory;
+
+  /**
+   * This is set in frontendConfig until id converter been fully migrate to router.
+   */
+  public final String idSigningServiceFactory;
+
+  /**
+   * This is set in frontendConfig until id converter been fully migrate to router.
+   */
+  public final String namedBlobDbFactory;
+
+  /**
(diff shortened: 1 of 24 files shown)


ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
```diff
@@ -761,11 +761,30 @@ public class RouterConfig {
...
+   * This is set in frontendConfig until id converter been fully migrate to router.
+   */
+  public final String namedBlobDbFactory;
+
   /**
    * Create a RouterConfig instance.
    * @param verifiableProperties the properties map to refer to.
    */
   public RouterConfig(VerifiableProperties verifiableProperties) {
+    FrontendConfig frontendConfig = new FrontendConfig(verifiableProperties);
```
Arun-LinkedIn, line 784: Minor: We could also create `FrontendConfig frontendConfig = new FrontendConfig(verifiableProperties);` in `NonBlockingRouterFactory` and directly use `frontendConfig.idConverterFactory, frontendConfig.idSigningServiceFactory, etc` over there.

Arun-LinkedIn (approved):

ambry-api/src/main/java/com/github/ambry/config/RouterConfig.java
SophieGuo410: yeah, I just want the router logic use the router config, and later we can consider remove the idConverterFactory from frontendConfig if id converter has been fully migrate to router layer.

PR #2900, opened 2024-09-23
Bug fix for updating configChanged

vaibhav-mittal-linkedin:
***Description***
There is bug in PropertyStoreCleanUpTaskTest. The configChanged variable which we are setting is not getting updated properly. It is due to wrong use of OR operator. This PR fixes this bug.

***Testing***
Added updated unit tests to ensure the added fix is tested.
Test added :- I have updated DataNodeConfig with multiple disks so that we can check configChanged is updated properly and all replicas are removed from DataNodeConfig when task is running.


justinlin-linkedin (approved):
LGTM

PR #2907, opened 2024-10-02
Move id converter to router.deleteBlob

SophieGuo410:


ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -471,21 +474,73 @@ public Future<String> stitchBlob(BlobProperties blobProperties, byte[] userMetad
...
+      }
+      proceedWithDelete(blobId, serviceId, callback, futureResult, quotaChargeCallback);
+    } else {
+      //if the blobId is not named blob based, it could bypass the first round of d converter logic by checking InternalKeys.BLOB_ID.
+      //First round is to convert named blob to blob Id.
+      if (restRequest.getArgs().get(RestUtils.InternalKeys.BLOB_ID) != null) {
+        String blobIdStr =
+            removeLeadingSlashIfNeeded(restRequest.getArgs().get(RestUtils.InternalKeys.BLOB_ID).toString());
+        proceedWithDelete(blobIdStr, serviceId, callback, futureResult, quotaChargeCallback);
+      } else {
```
Arun-LinkedIn, line 500: I am wondering if we can call idConverter.convert() all the time inside the router. For named blobs, it would return the converted blob ID. For regular blobs, it would remove leading slash. Then we wouldn't need if-else condition and `removeLeadingSlashIfNeeded()` method here. What do you think?

SophieGuo410: Logic wise I think it's doable, but I just want to avoid unnecessary network calls to mysql database, that's why I set this internal header to differentiate diff scenarios.

Arun-LinkedIn: Hmm I think in our `IdConverter` logic, we only make calls to MySQL if it is named blobs. For regular blobs, it would just strip the slash

SophieGuo410: Updated.

justinlin-linkedin (approved):
LGTM

PR #2909, opened 2024-10-03
Avoid retaining/releasing Netty Bytebuf in PipedAsyncWritableChannel

Arun-LinkedIn:
Since we are waiting for write completion callbacks from both primary and secondary read channels before invoking source read channel callback (which reduces Netty ByteBuf reference count to 0), we don't need to **increment/decrement reference count of ByteBuf in PipedAsyncWritableChannel**. This should avoid illegal reference count exception seen when writing the data out.


SophieGuo410 (commented):
This is just a general question, is there any test that can help detect to verify we don't have illegal reference count exception?

justinlin-linkedin (approved):

Arun-LinkedIn:
> This is just a general question, is there any test that can help detect to verify we don't have illegal reference count exception?

We can probably add unit tests to trigger illegal reference counts in a given class. But since bytebuf is passed through the Netty -> Router -> Network client, I think it is hard to catch prod illegal reference counts via unit tests.

PR #2915, opened 2024-10-09
Add crc checks in put operation

Arun-LinkedIn:
This change makes sure that content of a chunk in a PUT operation is not modified in frontend (due to illegal reference counts of byte buf, etc) while we are writing it to server


ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1613,6 +1649,13 @@ int fillFrom(ByteBuf channelReadBuf) {
     void checkAndMaybeComplete() {
       boolean done = false;
       // Now, check if this chunk is done.
+      if (chunkException != null && chunkException.getErrorCode() == RouterErrorCode.BlobCorrupted) {
+        logger.error("{} : Put chunk {} failed due to corruption. Failing entire operation", loggingContext,
```
justinlin-linkedin, line 1653: can we add a new metric for this error?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
```diff
@@ -1613,6 +1649,13 @@ int fillFrom(ByteBuf channelReadBuf) {
...
+            getChunkBlobId());
+        setOperationExceptionAndComplete(chunkException);
```
justinlin-linkedin, line 1655: could you please add a comment here. Usually we would set the chunk blob id to null before setting operation exception, but since we want to delete this chunk id later so we won't set it to null and just have it added to metadata chunk so later it will be removed by put manager.

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
Arun-LinkedIn: I have a metric `putBlobCRCMismatchCount` which gets incremented when `onPutBlobError` is called during completion of the operation. Will that be good for us?

ambry-router/src/main/java/com/github/ambry/router/PutOperation.java
Arun-LinkedIn: Thank you for pointing this! Actually I think `chunkBlobId` is set to `null` when we call `setOperationExceptionAndComplete(chunkException)`. To clean this particular chunk from servers, I reused the slipped put list now. The rest of the chunks (if this is a composite blob) seem to be already cleared in  `PutManager#onComplete()`. What do you think?

justinlin-linkedin (approved):
LGTM

SophieGuo410 (approved):

PR #2916, opened 2024-10-10
Add support for a semantic+revision version scheme

nikhilbedi:
## Summary

Add support in `DatasetDao` for a dataset version scheme that includes a semantic version and a revision number. In other words, a user-facing version that looks like 1.2.3.4. 

## Details

The code follows similar structure to that of the existing semantic version type, except the major/minor/patch version parts are all shifted left by an additional 3 digits, i.e. multiplied by 1000. This makes room for a revision number.

We are storing and using long values, so there is not a risk for integer overflow.

## Testing Done

I added integration tests specifically for the new version type, mapping closely to the tests that we perform for the traditional semantic type. Please inform if there are other tests we need to run.

### DIFF
--- a/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
+++ b/ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
@@ -76,6 +76,7 @@ public class MySqlAccountServiceIntegrationTest {
   private static final String DATASET_NAME_NOT_EXIST = "testDatasetNotExist";
   private static final String DATASET_NAME_WITH_TTL = "testDatasetWithTtl";
   private static final String DATASET_NAME_WITH_SEMANTIC = "testDatasetWithSemantic";
+  private static final String DATASET_NAME_WITH_SEMANTIC_LONG = "testDatasetWithSemanticLong";
   private static final String DATASET_NAME_EXPIRED = "testDatasetExpired";
 
   public MySqlAccountServiceIntegrationTest() throws Exception {
@@ -949,8 +950,10 @@ public class MySqlAccountServiceIntegrationTest {
     verifyDatasetProperties(dataset, datasetFromMysql);
   }
 
+// region Latest Version Support Tests
+
   /**
-   * Test add and get latest version.
+   * Test add and get latest version with monotonic format.
    * @throws Exception
    */
   @Test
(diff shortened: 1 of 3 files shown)


ambry-account/src/main/java/com/github/ambry/account/mysql/DatasetDao.java
```diff
@@ -961,36 +994,88 @@ private long generateAutoIncrementedVersionBasedOnSchema(long versionValue, Data
...
   }
 
   /**
    * @param version the version string.
+   * @param versionSchema the {@link com.github.ambry.account.Dataset.VersionSchema} from dataset level.
    * @return {@code} true if the version is auto created version for upload.
    */
-  private boolean isAutoCreatedVersionForUpload(String version) {
-    return LATEST.equals(version) || MAJOR.equals(version) || MINOR.equals(version) || PATCH.equals(version);
+  private boolean isAutoCreatedVersionForUpload(String version, Dataset.VersionSchema versionSchema) {
```
SophieGuo410, line 1073: I think we can simply return
LATEST.equals(version) || MAJOR.equals(version) || MINOR.equals(version) || PATCH.equals(version) || REVISION.equals(version).
Because if user provide MAJOR, it means auto incr the major version. I think similar thing should be applied for REVISION?

ambry-account/src/integration-test/java/com/github/ambry/account/MySqlAccountServiceIntegrationTest.java
```diff
@@ -1009,18 +1012,28 @@ public void testLatestVersionSupport() throws Exception {
...
             DatasetVersionState.READY);
     assertEquals("Mismatch in dataset", expectedDatasetVersionRecord, datasetVersionRecordFromMysql);
+  }
 
+  /**
+   * Test add and get latest version with semantic versioning format.
+   * @throws Exception
+   */
+  @Test
+  public void testLatestSemanticVersionSupport() throws Exception {
```
SophieGuo410, line 1022: Can you please add your own test cases instead of modifying the existing one? Currently I'm adding some integration test as well. Once it's done, you can refer it and add your test regarding this change as well.

SophieGuo410 (commented):
We have the MysqlAccountSerivceIntegrationTest, can you add your use case to that test as well? It will put the record in your local db.

SophieGuo410 (approved):

PR #2922, opened 2024-10-16
adding limit for group iteration limit for continuous replication

manbearpig1996:
adding limit for group iteration limit for continuous replication


vaibhav-mittal-linkedin (approved):
LGTM

PR #2925, opened 2024-10-17
Add timeouts for secondary readers in piped async writable channel

Arun-LinkedIn:


ambry-commons/src/main/java/com/github/ambry/commons/PipedAsyncWritableChannel.java
```diff
@@ -41,22 +44,29 @@
 public class PipedAsyncWritableChannel implements AsyncWritableChannel {
 
   private final ReadableStreamChannel sourceChannel;
+  private final int secondaryTimeoutInMs;
   private final PipedReadableStreamChannel pipedPrimaryReadChannel;
   private PipedReadableStreamChannel pipedSecondaryReadChannel = null;
   private final ReentrantLock lock = new ReentrantLock();
   private final AtomicBoolean channelOpen = new AtomicBoolean(true);
   private static final Logger logger = LoggerFactory.getLogger(PipedAsyncWritableChannel.class);
+  final static HashedWheelTimer wheel = new HashedWheelTimer(10, TimeUnit.MILLISECONDS, 1024);
```
justinlin-linkedin, line 53: do we want to have a time wheel per channel? Can we create one time wheel for all channels and pass this time wheel to each channel.

justinlin-linkedin: BTW, let's provide a ThreadFactory to the HashedWheelTimer constructor to rename the worker thread in the wheel.

ambry-commons/src/main/java/com/github/ambry/commons/PipedAsyncWritableChannel.java
```diff
@@ -334,11 +348,12 @@ private Callback<Long> makeCallbackForReadableStreamChannel(boolean primary, Ato
               }
               resolveChunk(primaryWriteCallbackResult.bytesWritten, primaryWriteCallbackResult.exception);
             } else if (primaryWriteCallbackResult != null) {
-              // TODO: Write successful callback came from primary but not from secondary. We will "start a timer" to wait
-              //  for result from secondary. If we time out waiting for the result, we will close secondary and send the
-              //  remaining bytes to primary alone so that primary upload SLA is not affected.
-            } else {
-              // Do nothing. Callback from Secondary came. We will wait for callback to come from primary
+              secondaryTimeout = wheel.newTimeout(timeout -> {
```
justinlin-linkedin, line 351: HashedWheelTimer actually limit the number of concurrent instances to 64, so we create one instance per Pipe. We need to share the instance between different pipes. It would be ideal to only create one instance for all pipes, but we don't really know if one instance would be good enough.

I suggest that we have a metric to keep track of the delayed timeout duration and see if one instance would have enough capacity to handle all timeouts. 

For instance, when we summit a timeout task to wheel, we also would know the expected timeout time. And when the timeout callback is finally invoked, we can compare it with the expected timeout value and see if the difference. Let's have a histogram of the different in ms. If one instance can deal with all the timeout, then we should see the 95% of difference to be always be within one tick. If not, it means we need to create more instances and randomly assign one instance to one pipe.

ambry-commons/src/main/java/com/github/ambry/commons/PipedAsyncWritableChannel.java
Arun-LinkedIn: Sure. Provided ThreadFactory to HashedWheelTimer to rename worker threads. Using static to have one time wheel instance for now

ambry-commons/src/main/java/com/github/ambry/commons/PipedAsyncWritableChannel.java
Arun-LinkedIn: Sure. Added metric to keep track of delayed timeout duration

ambry-commons/src/main/java/com/github/ambry/commons/PipedAsyncWritableChannel.java
```diff
@@ -41,22 +49,33 @@
...
-  public PipedAsyncWritableChannel(ReadableStreamChannel sourceChannel, boolean withSecondary) {
+  public PipedAsyncWritableChannel(ReadableStreamChannel sourceChannel, boolean withSecondary, int secondaryTimeoutInMs,
+      MetricRegistry metricRegistry) {
     this.sourceChannel = sourceChannel;
+    this.secondaryTimeoutInMs = secondaryTimeoutInMs;
 
     pipedPrimaryReadChannel = new PipedReadableStreamChannel(true);
     if (withSecondary) {
       pipedSecondaryReadChannel = new PipedReadableStreamChannel(false);
+      wheel.start();
```
justinlin-linkedin, line 78: wheel.start should be threadsafe and idempotent so we can call start here in every pipe. but a better way is to have a static statement right after the variable to start it. It's probably fine but would be nice to fine some place it close it.

justinlin-linkedin (approved):

SophieGuo410 (approved):

ambry-commons/src/main/java/com/github/ambry/commons/PipedAsyncWritableChannel.java
Arun-LinkedIn: Sure.. I will try to update it when I move this class to closed source..

PR #2926, opened 2024-10-21
Add account level metrics for container metrics

justinlin-linkedin:
## Summary
Add account level metric to aggregate its containers' metrics.

## Details
Some accounts have lots of containers and it's very hard those those accounts to aggregate all the metrics for 40x, 50x and latency etc. This PR adds an **AccountMetrics** class and have its **ContainerMetrics** to record metrics for **AccountMetrics**.

This PR also adds a configuration to only enable some accounts to have this **AccountMetrics** as most accounts have very small set of containers that they just don't need extra metrics.

## Test
Unit test

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/frontend/AccountMetrics.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/AccountMetrics.java
@@ -1,0 +1,36 @@
+/**
+ * Copyright 2024 LinkedIn Corp. All rights reserved.
+ * <p>
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except in compliance with
+ * the License. You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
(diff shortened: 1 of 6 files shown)


ambry-api/src/main/java/com/github/ambry/frontend/AccountMetrics.java
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Metrics for an operation on a specific account.
+ */
+public class AccountMetrics extends EntityOperationMetrics {
+
+  /**
+   * Metric names will be in the following format:
+   * {@code com.github.ambry.frontend.AccountMetrics.{accountName}___{operationType}{metric}} For example:
+   * {@code com.github.ambry.frontend.ContainerMetrics.account-a___GetBlobSuccessCount}
```
Arun-LinkedIn, line 26: nit: In example, looks like the name is `ContainerMetrics`

Arun-LinkedIn (approved):
Lgtm!

PR #2931, opened 2024-10-31
Reorder disks if mounts have been shuffled

litingulfs:
## Summary
Sometimes ambry-server machines will get reimaged and come back with disks mounted in the wrong order. Human error during certain maintenance or testing operations can also cause this to happen. Currently, these disks are simply wiped clean of data upon startup, Ambry assuming that since they do not contain the right partitions they must be completely broken. The affected disks are then re-seeded with their desired data through replication. However, replication is slow (up to several hours), and until all the data is replicated back, availability may be adversely affected.

This pull request changes the `ambry-server` startup process such that it attempts to detect a simple reordering of disk mounts. For example, if the disks that were mounted at `/mnt/u003` and `/mnt/u008` were swapped around. If such a reordering is detected, we simply update Helix (our source of truth) with the new disk order and shut down the server. The server then needs to get restarted by some manual or automated means outside of the Ambry process.

Note that this PR sets up the simplest possible automated fix for the disk mount shuffle problem:
* We will only remediate if all the partitions that exist in Helix are found on disk. So for example, if two mounts have been shuffled and there is also one missing mount, then we will not attempt any reshuffling at all.
* We issue a simple `System.exit()` to shut down the Java process without trying any clever in-place update of the state. We investigated more elegant solutions, but the code quickly becomes complex and error-prone.

We may revisit this basic process if it turns out to be inadequate in the future.

Note that all the new functionality is guarded by a config that defaults to `false`. So deploying this new code will not change anything unless the corresponding config is also explicitly turned on.

## Testing Done
New unit tests have been added.

Note that we are planning to roll out the new code in practice as follows:

1. First manually test in the Ambry perf cluster, which has no client traffic but only exists for the purpose of internal Ambry team testing. In the perf cluster, we'll swap mounts manually to confirm that the code performs as desired in a real (test) cluster.
2. Once we're confident that the perf cluster tests have all been successful, we'll move onto testing in EI. EI is the second best place to test, since it has some traffic from clients and there may be a little impact if anything goes wrong.
3. Finally, we'll roll out in production clusters, fabric by fabric, by turning on the new config.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -436,6 +436,43 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
     }
   }
 
+  /**
+   * Given a map of old disks and new disks, swaps the old disks with the new disks in the DataNodeConfig,
+   * and persists the resulting changes to Helix.
+   * @param newDiskMapping A map of old disks to new disks.
+   * @return {@code true} if the disks order was successfully updated, {@code false} otherwise.
+   */
+  public boolean setDisksOrder(Map<DiskId, DiskId> newDiskMapping) {
+    if (newDiskMapping == null || newDiskMapping.isEmpty()) {
+      throw new IllegalArgumentException("Map of disk mappings is empty when attempting to set disks order");
+    }
+
+    // Update DataNodeConfig and save it.
+    synchronized (helixAdministrationLock) {
+      DataNodeConfig dataNodeConfig = getDataNodeConfig();
+      boolean success = true;
+      for (DiskId oldDisk : newDiskMapping.keySet()) {
+        DiskId newDisk = newDiskMapping.get(oldDisk);
+
+        // Confirm that both disks are present in the DataNodeConfig
+        DataNodeConfig.DiskConfig oldDiskConfig = dataNodeConfig.getDiskConfigs().get(oldDisk.getMountPath());
+        DataNodeConfig.DiskConfig newDiskConfig = dataNodeConfig.getDiskConfigs().get(newDisk.getMountPath());
+        if (oldDiskConfig == null || newDiskConfig == null) {
+          throw new IllegalArgumentException("Disk " + oldDisk.getMountPath() + " or " + newDisk.getMountPath() + " can't be found in the DataNodeConfig");
+        }
+
+        // Swap the disks in the DataNodeConfig
+        logger.info("Replacing disk {} with disk {}", oldDisk.getMountPath(), newDisk.getMountPath());
+        dataNodeConfig.getDiskConfigs().put(oldDisk.getMountPath(), dataNodeConfig.getDiskConfigs().get(newDisk));
+        if (!dataNodeConfigSource.set(dataNodeConfig)) {
+          logger.error("Setting disks order failed DataNodeConfig update");
+          success = false;
+        }
+      }
+      return success;
+    }
+  }
+
   @Override
   public boolean supportsStateChanges() {
     return true;
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -467,7 +467,7 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
 
         // Swap the disks in the DataNodeConfig
         logger.info("Replacing disk {} with disk {}", oldDisk.getMountPath(), newDisk.getMountPath());
-        dataNodeConfig.getDiskConfigs().put(oldDisk.getMountPath(), dataNodeConfig.getDiskConfigs().get(newDisk));
+        dataNodeConfig.getDiskConfigs().put(oldDisk.getMountPath(), dataNodeConfig.getDiskConfigs().get(newDisk.getMountPath()));
       }
 
       if (!dataNodeConfigSource.set(dataNodeConfig)) {
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -452,7 +452,7 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
       DataNodeConfig dataNodeConfig = getDataNodeConfig();
 
       // Make a copy of the disk configs to avoid accidentally overwriting state.
-      Map<String, DataNodeConfig.DiskConfig> originalDiskConfigs = dataNodeConfig.getDiskConfigs();
+      Map<String, DataNodeConfig.DiskConfig> originalDiskConfigs = new HashMap<> (dataNodeConfig.getDiskConfigs());
       for (DiskId oldDisk : newDiskMapping.keySet()) {
         DiskId newDisk = newDiskMapping.get(oldDisk);
 
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
@@ -450,27 +450,32 @@ public class HelixParticipant implements ClusterParticipant, PartitionStateChang
     // Update DataNodeConfig and save it.
     synchronized (helixAdministrationLock) {
       DataNodeConfig dataNodeConfig = getDataNodeConfig();
+
+      // Tommy: we have to have a copy of the old original disk mappings here for swapping to work!
+      Map<String, DataNodeConfig.DiskConfig> originalDiskConfigs = dataNodeConfig.getDiskConfigs();
+
       boolean success = true;
       for (DiskId oldDisk : newDiskMapping.keySet()) {
         DiskId newDisk = newDiskMapping.get(oldDisk);
 
         // Confirm that both disks are present in the DataNodeConfig
-        DataNodeConfig.DiskConfig oldDiskConfig = dataNodeConfig.getDiskConfigs().get(oldDisk.getMountPath());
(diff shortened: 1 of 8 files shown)


ambry-store/src/main/java/com/github/ambry/store/ReplicaPlacementValidator.java
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * Find all the partition directories on the disk.
+   * @param disk an instance of DiskId to search for partition directories.
+   * @return A list of partition directories on the disk.
+   */
+  private Set<String> findPartitionsOnDisk(DiskId disk) {
+    Set<String> partitionDirs = new HashSet<>();
+    File[] directories = new File(disk.getMountPath()).listFiles(File::isDirectory);
```
justinlin-linkedin, line 131: We should probably check if the `disk.getMountPath()` exists or not. List file on nonexisting directory would throw some exception.

ambry-store/src/main/java/com/github/ambry/store/ReplicaPlacementValidator.java
```diff
@@ -0,0 +1,146 @@
...
+
+    for (Map.Entry<DiskId, List<ReplicaId>> entry : expectedDiskToReplicaMap.entrySet()) {
+      DiskId currentDisk = entry.getKey();
+      foundDiskToPartitionMap.put(currentDisk, findPartitionsOnDisk(currentDisk));
+      List<ReplicaId> replicas = entry.getValue();
+      for (ReplicaId replica : replicas) {
+        String partitionID = replica.getPartitionId().toString();
+        if(!foundDiskToPartitionMap.get(currentDisk).contains(partitionID)) {
+          brokenDisks.add(currentDisk);
+        }
```
justinlin-linkedin, line 52: If a host got reimaged, and came back with empty disks, then we don't want to think those disks are broken.

ambry-store/src/main/java/com/github/ambry/store/ReplicaPlacementValidator.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    for (DiskId currentDisk : brokenDisks) {
+      List<ReplicaId> expectedReplicas = expectedDiskToReplicaMap.get(currentDisk);
+      DiskId foundDisk = findDiskWithReplicas(expectedReplicas);
+      if(foundDisk == null) {
+        return Collections.emptyMap();
+      } else {
+        shuffledDisks.put(currentDisk, foundDisk);
+      }
```
justinlin-linkedin, line 80: The empty disk might be covered by these statements.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -436,6 +436,43 @@ public boolean setDisksState(List<DiskId> diskIds, HardwareState state) {
...
+          throw new IllegalArgumentException("Disk " + oldDisk.getMountPath() + " or " + newDisk.getMountPath() + " can't be found in the DataNodeConfig");
+        }
+
+        // Swap the disks in the DataNodeConfig
+        logger.info("Replacing disk {} with disk {}", oldDisk.getMountPath(), newDisk.getMountPath());
+        dataNodeConfig.getDiskConfigs().put(oldDisk.getMountPath(), dataNodeConfig.getDiskConfigs().get(newDisk));
+        if (!dataNodeConfigSource.set(dataNodeConfig)) {
+          logger.error("Setting disks order failed DataNodeConfig update");
+          success = false;
+        }
```
justinlin-linkedin, line 470: we should only call dataNodeConfigSource.set once, after for loop.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -436,6 +436,43 @@ public boolean setDisksState(List<DiskId> diskIds, HardwareState state) {
...
+        // Confirm that both disks are present in the DataNodeConfig
+        DataNodeConfig.DiskConfig oldDiskConfig = dataNodeConfig.getDiskConfigs().get(oldDisk.getMountPath());
+        DataNodeConfig.DiskConfig newDiskConfig = dataNodeConfig.getDiskConfigs().get(newDisk.getMountPath());
+        if (oldDiskConfig == null || newDiskConfig == null) {
+          throw new IllegalArgumentException("Disk " + oldDisk.getMountPath() + " or " + newDisk.getMountPath() + " can't be found in the DataNodeConfig");
+        }
+
+        // Swap the disks in the DataNodeConfig
+        logger.info("Replacing disk {} with disk {}", oldDisk.getMountPath(), newDisk.getMountPath());
+        dataNodeConfig.getDiskConfigs().put(oldDisk.getMountPath(), dataNodeConfig.getDiskConfigs().get(newDisk));
```
justinlin-linkedin, line 466: this is probably not enough, say disk1 shuffles to disk2 and disk2 shuffle to disk1.

In this case, the `newDiskMapping` would have {disk1:disk2, disk2:disk1}. 
Say we are dealing with disk1:disk2 entry and we are setting disk1 (old) to disk2's config. Now the dataNodeConfig's disk config map would look like {disk1:disk2config, disk2:disk2config}. Disk1config is missing.

I think we have to make a copy of dataNodeConfig's disk config map and use this copied map as temporary reference.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -436,6 +436,48 @@ public boolean setDisksState(List<DiskId> diskIds, HardwareState state) {
...
+        // Confirm that both disks are present in the DataNodeConfig
+        DataNodeConfig.DiskConfig oldDiskConfig = originalDiskConfigs.get(oldDisk.getMountPath());
+        DataNodeConfig.DiskConfig newDiskConfig = originalDiskConfigs.get(newDisk.getMountPath());
+        if (oldDiskConfig == null || newDiskConfig == null) {
+          throw new IllegalArgumentException("Disk " + oldDisk.getMountPath() + " or " + newDisk.getMountPath() + " can't be found in Helix (DataNodeConfig)");
+        }
+
+        // Swap the disks in the DataNodeConfig
+        logger.info("Replacing disk {} with disk {}", oldDisk.getMountPath(), newDisk.getMountPath());
+        dataNodeConfig.getDiskConfigs().put(oldDisk.getMountPath(), dataNodeConfig.getDiskConfigs().get(newDisk));
```
justinlin-linkedin, line 470: and it should bre diskNodConfig.getDiskConfigs().get(newDisk.getMountPath());

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
```diff
@@ -161,6 +167,35 @@ public StorageManager(StoreConfig storeConfig, DiskManagerConfig diskManagerConf
...
+    for (DiskId currentDisk : diskToReplicas.keySet()) {
+      disksToPartitions.put(currentDisk, partitionFinder.findPartitionsOnDisk(currentDisk));
+    }
+
+    ReplicaPlacementValidator placementValidator = new ReplicaPlacementValidator(diskToReplicas, disksToPartitions);
+    Map<DiskId, DiskId> disksToReshuffle = placementValidator.reshuffleDisks();
+    if (!disksToReshuffle.isEmpty()) {
+      logger.info("Disks need to be reshuffled: {}", disksToReshuffle);
+      if(primaryClusterParticipant.setDisksOrder(disksToReshuffle)) {
+        logger.info(this.getClass().getSimpleName() + " - successfully reshuffled disks. Now terminating"
```
justinlin-linkedin, line 190: we don't need to add class name to the logger, logger would print it by default.

ambry-store/src/main/java/com/github/ambry/store/ReplicaPlacementValidator.java
litingulfs: Yeah, thanks - I added error handling for this.

ambry-store/src/main/java/com/github/ambry/store/ReplicaPlacementValidator.java
litingulfs: As discussed offline, this pull request only attempts to detect and perform the simplest possible fix. So we'll leave this code in place for now and update it if it's found to be inadequate in practice.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
litingulfs: Thanks for pointing this out, I am new to the Helix-Ambry integration. This has been updated.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
litingulfs: Yeah, you are right, I fixed this like you pointed out.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
litingulfs: Right, fixed.

ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
litingulfs: Ah good to know. Thanks for pointing that out. This has been updated.

justinlin-linkedin (approved):
LGTM

ambry-store/src/main/java/com/github/ambry/store/ReplicaPlacementValidator.java
litingulfs: Yeah, this is intended. As discussed offline, the code in this pull request just attempts to rectify the simplest possible case - and by design does not attempt to do anything if any disks are empty. We'll revisit this design decision later if it turns out to be inadequate in practice.

PR #2932, opened 2024-11-01
Ignore unexpected blob id in replication

justinlin-linkedin:
## Summary 
Add a configuration to ignore unexpected blob id in replication thread. It's safe to do so because these blob ids don't belong to the partition anyway, we should just ignore it.

## Test
no tests added

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
@@ -1265,6 +1265,8 @@ public class ReplicaThread implements Runnable {
         if (replicationConfig.replicationIgnoreUnexpectedBlobIdError) {
           logger.error(errorMessage);
           replicationMetrics.unexpectedBlobIdError.inc();
+          missingRemoteStoreMessages.remove(messageInfo);
+          continue;
         } else {
           throw new IllegalStateException(errorMessage);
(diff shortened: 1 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -1259,9 +1259,15 @@ void processReplicaMetadataResponse(Set<MessageInfo> missingRemoteStoreMessages,
     for (MessageInfo messageInfo : messageInfoList) {
       BlobId blobId = (BlobId) messageInfo.getStoreKey();
       if (remoteReplicaInfo.getLocalReplicaId().getPartitionId().compareTo(blobId.getPartition()) != 0) {
-        throw new IllegalStateException(
-            "Blob id " + blobId.getID() + " is not in the expected partition Actual partition " + blobId.getPartition()
-                + " Expected partition " + remoteReplicaInfo.getLocalReplicaId().getPartitionId());
+        String errorMessage =
```
manbearpig1996, line 1262: Won't this send get request for blob ids ?
And when we get and put these blob in store, store would throw exception as partition id will not match.

manbearpig1996 (approved):
GTG

SophieGuo410 (approved):
LGTM!

PR #2938, opened 2024-11-12
adding logs for blob id to partition journey

manbearpig1996:
adding logs for blob id to partition journey


ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -282,6 +286,10 @@ public void handlePutRequest(NetworkRequest request) throws IOException, Interru
       } else {
         MessageFormatWriteSet writeSet = getMessageFormatWriteSet(receivedRequest);
         Store storeToPut = storeManager.getStore(receivedRequest.getBlobId().getPartition());
+        if (serverConfig.serverEnableBlobPartitionJourneyLogs) {
```
justinlin-linkedin, line 289: should follow into the storage manager and disk manager to see what was picked.

mudit-saxena (approved):
GTG

PR #2955, opened 2024-12-03
[frontend] List named blobs SQL

snalli:
Fix list-named-blobs SQL to use smaller joins. This improves query performance by 1000x.


ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -113,16 +113,17 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
-          + "WHERE t1.blob_name LIKE ? AND t1.blob_name >= ? ORDER BY t1.blob_name ASC LIMIT ?",STATE_MATCH, CURRENT_TIME);
+      + "SELECT blob_name, blob_id, MAX(version) AS version, deleted_ts, blob_size, modified_ts"
+      + "  FROM named_blobs_v2"
+      + "  WHERE account_id = ?"
+      + "    AND container_id = ?"
+      + "    AND blob_state = %1$s"
+      + "    AND (deleted_ts IS NULL OR deleted_ts > %2$S)"
+      + "    AND blob_name LIKE ?"
+      + "    AND blob_name >= ?"
+      + "  GROUP BY blob_name"
```
SophieGuo410, line 124: We should group by blobName accountId and containerId.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -113,16 +113,17 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
-          + "  AND (deleted_ts IS NULL OR deleted_ts>%2$S) "
-          + "        GROUP BY account_id, container_id, blob_name) t2 "
-          + "ON (t1.account_id,t1.container_id,t1.blob_name,t1.version) = (t2.account_id,t2.container_id,t2.blob_name,t2.version) "
-          + "WHERE t1.blob_name LIKE ? AND t1.blob_name >= ? ORDER BY t1.blob_name ASC LIMIT ?",STATE_MATCH, CURRENT_TIME);
+      + "SELECT blob_name, blob_id, MAX(version) AS version, deleted_ts, blob_size, modified_ts"
+      + "  FROM named_blobs_v2"
+      + "  WHERE account_id = ?"
+      + "    AND container_id = ?"
+      + "    AND blob_state = %1$s"
+      + "    AND (deleted_ts IS NULL OR deleted_ts > %2$S)"
```
SophieGuo410, line 121: If we have two versions, ver1 delete_ts is null, ver2 delete_ts is not null, we will return ver1. Which is not the correct case. We need to return nothing.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -113,16 +113,17 @@ class MySqlNamedBlobDb implements NamedBlobDb {
...
+      + "    AND blob_name LIKE ?"
+      + "    AND blob_name >= ?"
+      + "  GROUP BY blob_name"
+      + "  ORDER BY blob_name ASC"
+      + "  LIMIT ?;", STATE_MATCH, CURRENT_TIME);
```
SophieGuo410, line 126: The LIST_QUERY_V2 will also need to be fixed.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -113,16 +113,17 @@ class MySqlNamedBlobDb implements NamedBlobDb {
    */
   // @formatter:off
   private static final String LIST_QUERY_V2 = String.format(""
```
SophieGuo410, line 115: I take a look at our original code, I suppose the reason we have such complex query is because we want to make sure the blobId we queried is match with the latest version.
For example 
blobName1 blobId2 ver1
blobName2 blobId1 ver2
with your query, it will return the max verison which is ver2, but we can't garentee it will return the blobId1. 
But I checked the list query, at handler level we only need the blobName, so I suggest instead of return the full namedBlobRecord, we should return the blobName, in that case, we don't need to worry about which blobId we returned.

SophieGuo410 (commented):
We should be careful about what we changed here.
I suggest we could have a new query and switch by config. so we can rollback easily if something is wrong.
Also we need to add integration test to make sure it cover every case.
For example,
if we have the delete_ts is null for older version and deleted_ts is not null for a new version, we will return the old version blobName.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: Good pt. I addressed it using the WITH-clause.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: Lets fix one, we can do the other one later.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: I don't understand this example. If i understand correctly, we need the latest version of a blob that is Ready. If it is deleted, then return nothing. Else return it. Am i missing something /

snalli:
> We should be careful about what we changed here. I suggest we could have a new query and switch by config. so we can rollback easily if something is wrong. Also we need to add integration test to make sure it cover every case. For example, if we have the delete_ts is null for older version and deleted_ts is not null for a new version, we will return the old version blobName.

will add a test, the coverage seems limited, because the tests passed even though my prev query was not what is required.

snalli:
> We should be careful about what we changed here. I suggest we could have a new query and switch by config. so we can rollback easily if something is wrong. Also we need to add integration test to make sure it cover every case. For example, if we have the delete_ts is null for older version and deleted_ts is not null for a new version, we will return the old version blobName.

converted the query itself to a config to avoid a v2/v3 switch, as we far too many of those.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: `testListNamedBlobs` already checks for it

snalli: added another test - a better one

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
```diff
@@ -266,6 +266,66 @@ record = new NamedBlobRecord(account.getName(), container.getName(), blobName, b
...
+    namedBlobDb.put(v2, NamedBlobState.READY, true).get();
+    namedBlobDb.put(v2_other, NamedBlobState.READY, true).get();
+    page = namedBlobDb.list(account.getName(), container.getName(), blobName, null, null).get();
+    assertEquals(2, page.getEntries().size());
+    assertEquals(v2, page.getEntries().get(0));
+    assertEquals(v2_other, page.getEntries().get(1));
+    time.sleep(100);
+
+    // delete blob and list should return empty
+    namedBlobDb.delete(account.getName(), container.getName(), blobName).get();
```
SophieGuo410, line 318: Can you please add some test like like
1. add one permanent version.
2. upsert the named blob and delete it.
In this case, we should list nothing cause the latest version already been deleted.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -628,19 +619,28 @@ private NamedBlobRecord run_get_v2(String accountName, String containerName, Str
...
     String query = "";
-    String queryStatement = blobNamePrefix == null ? LIST_ALL_QUERY_V2 : LIST_QUERY_V2;
+    String queryStatement = blobNamePrefix == null ? LIST_ALL_QUERY_V2 : LIST_NAMED_BLOBS_SQL;
     int maxKeysValue = maxKeys == null ? config.listMaxResults : maxKeys;
     try (PreparedStatement statement = connection.prepareStatement(queryStatement)) {
-      statement.setInt(1, accountId);
-      statement.setInt(2, containerId);
       if (blobNamePrefix == null) {
+        // list-all no prefix
+        statement.setInt(1, accountId);
```
SophieGuo410, line 627: I'm a little bit confused here.
You didn't change the LIST_ALL_QUERY_V2, why you can update the statement settings?

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -29,6 +29,47 @@ public class MySqlNamedBlobDbConfig {
...
+      + " ), "
+      + " BlobsMaxVersion AS ( "
+      + "   SELECT blob_name, MAX(version) as version "
+      + "   FROM named_blobs_v2 "
+      + "   WHERE account_id = ? " // 5
+      + "     AND container_id = ? " // 6
+      + "     AND blob_state = %1$s "
+      + "     AND blob_name LIKE ? " // 7
+      + "     AND blob_name >= ? " // 8
+      + "   GROUP BY blob_name "
```
SophieGuo410, line 61: I think GROUP BY logic operates independently of the WHERE clause, this GROUP BY account_id, container_id, blob_name ensures that the aggregation (e.g., MAX(version)) respects the boundaries of each account and container. Again, we are not the expert of mysql queries, and you might want to add some test case like adding named blob for different account and container and test the list. Even though you use where to select the corresponding account and container, it might not give you the right max version based on my understanding.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -29,6 +29,47 @@ public class MySqlNamedBlobDbConfig {
...
+  /**
+   * List named-blobs query.
+   * The inner query selects blobs based on the filters, and groups them on blob_name.
+   * In each group, it selects the blob with the highest version.
+   * The outer query filters out blobs that have been deleted, and return the latest blob in each group that is ready.
+   */
+  @Config(LIST_NAMED_BLOBS_SQL)
+  public static final String DEFAULT_LIST_NAMED_BLOBS_SQL = ""
+      + " WITH "
+      + "  BlobsAllVersion AS ( "
```
SophieGuo410, line 43: I feel like this query is still pretty complex. I suggest we can think of some simpler queries which makes the code easy to read. Sometimes having two queries is also ok cause we are trying to avoid the complex sql queris which might missing edge cases. cc @justinlin-linkedin

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -628,19 +619,28 @@ private NamedBlobRecord run_get_v2(String accountName, String containerName, Str
...
+        // list with prefix
+        statement.setInt(1, accountId);
+        statement.setInt(2, containerId);
         statement.setString(3, blobNamePrefix + "%");
         statement.setString(4, pageToken != null ? pageToken : blobNamePrefix);
+        statement.setInt(5, accountId);
+        statement.setInt(6, containerId);
+        statement.setString(7, blobNamePrefix + "%");
+        statement.setString(8, pageToken != null ? pageToken : blobNamePrefix);
+        statement.setInt(9, maxKeysValue + 1);
```
SophieGuo410, line 642: Can we add some time like this:
blobName1 blobId1 version1
blobName1 blobId2 version2
so listing we should return blobName1 blobId1 version2 right?
but I think in some case we might return the blobName1 blobId2 version.
This is ok but we should not return the  NamedBlobRecord with blobId info. Instead, we can only return the blobName, cause blobId1 and blobId2 all map to blobName1 and we don't need blobId for listing.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
SophieGuo410: I saw you only select the blob_name and  MAX(version), does not select the accountId, so sql won't be confused.

ambry-named-mysql/src/integration-test/java/com/github/ambry/named/MySqlNamedBlobDbIntegrationTest.java
snalli: Yeah, i have added a test

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -29,6 +29,33 @@ public class MySqlNamedBlobDbConfig {
...
   public static final String TRANSACTION_ISOLATION_LEVEL = PREFIX + "transaction.isolation.level";
+  public static final String LIST_NAMED_BLOBS_SQL = "list.named.blobs.sql";
+
+  /**
+   * List named-blobs query.
+   * The inner query selects blobs based on the filters, and groups them on blob_name.
+   * In each group, it selects the blob with the highest version.
+   * The outer query filters out blobs that have been deleted, and return the latest blob in each group that is ready.
+   */
+  @Config(LIST_NAMED_BLOBS_SQL)
```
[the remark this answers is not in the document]
snalli: What is hard to track?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: Didn't update statement settings for list all. Just moved it into a separate case because that has different number of parameters.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
snalli: Group-by executes after filters in the query plan because that is the least cost way of executing the query.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
snalli: I assure you this is one of the simplest queries, compared to queries issued elsewhere. It creates two minimal tables and then joins them. The join was unavoidable becuase of our previous design choices. However, we can be careful in join by reducing the number of rows on either side.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: This will never happen because we don't combine columns from different rows. The select statement at the end only returns rows from the left side of the join.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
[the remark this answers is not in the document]
SophieGuo410: I think keeping all the query in the MySqlNamedBlobDb class is straight forward instead of keep it in the config file. we can discuss this internally.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
SophieGuo410: But if the blobNamePrefix is null, the queryStatement would become LIST_ALL_QUERY_V2 right? Am I miss anything here?

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
SophieGuo410: Yeah, I understand this will never happen for your query, but we want to make sure in the future if other person update this query, things will still work.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
SophieGuo410: My point is this query will not work if we are not disable only_full_group_by. 
For example if we have same blobName and belongs to two container.
mysql> select * from named_blobs_v2 where account_id = '1' and container_id = '1' and blob_name like 'test1'  GROUP BY blob_name;
ERROR 1055 (42000): Expression #4 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'ambrynamedblobs.named_blobs_v2.version' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by
Anyway, or later query won't have this issue.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
SophieGuo410: Yeah, As we discussed, I think your query is clear if we take some time to understand your logic. However, for debugging purpose it would be hard to construct such a long query. But anyway, it can be out of scope of this pr.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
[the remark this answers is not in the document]
snalli: done

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
snalli: discussed offline, however i want to add that the next person should be responsible for adding a test if that case is likely to happen.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
snalli: discussed offline, online query formatters, debugging v realtime performance.

ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -93,6 +102,8 @@ public class MySqlNamedBlobDbConfig {
   public final TransactionIsolationLevel transactionIsolationLevel;
 
   public MySqlNamedBlobDbConfig(VerifiableProperties verifiableProperties) {
+    this.listNamedBlobsSQLOption = verifiableProperties.getInt(LIST_NAMED_BLOBS_SQL_OPTION,
```
SophieGuo410, line 105: If we only have two option, I would suggest to use boolean instead of switch.
The reason are:
1. for switch we'd better provide the default which throw IllegalArgumentException exception for invalid value.
2. for the config setting, we'd better provide the min and max value to avoid people put some number out of limit.
3. we will always need to check the code and understand which value we want to pick, for easy maintainance, probably a boolean value is more straightforward given that we only have two options.

snalli: if my query works in prod, i plan to remove this config completely.

SophieGuo410: sorry to be nitpicking, I just feel like even though we will remove it later, we'd better to always provide the better quality code cause sometimes you might be pull into other urgent task and this code will leave for a long time. That's just my preference. I would suggest we can at least add the min and max value for the config so if we forget to remove for a long time, we still have some constraint instead of return null only.

SophieGuo410 (approved):

PR #2957, opened 2024-12-03
Bootstrap state build interface changes part 2/2

piyujai:
Removed trigger for buildStateForFileCopy method since FileCopyManager and removed TODOs from pre-file copy steps.

### DIFF
--- a/ambry-file-transfer/src/main/java/com/github/ambry/FileCopyManager.java
+++ b/ambry-file-transfer/src/main/java/com/github/ambry/FileCopyManager.java
@@ -41,8 +41,7 @@ public class FileCopyManager {
 
     @Override
     public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
-      // StateBuilding at the end of FCM's Offline->Bootstrap transition
-      storeManager.buildStateForFileCopy(partitionName);
+      // StateBuilding (storeManager.buildStateForFileCopy()) will be triggered at the end of FCM's async handler.
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


DevenAhluwalia (approved):

ambry-file-transfer/src/main/java/com/github/ambry/FileCopyManager.java
```diff
@@ -41,8 +41,7 @@ class PartitionStateChangeListenerImpl implements PartitionStateChangeListener {
 
     @Override
     public void onPartitionBecomeBootstrapFromOffline(String partitionName) {
-      // StateBuilding at the end of FCM's Offline->Bootstrap transition
-      storeManager.buildStateForFileCopy(partitionName);
+      // StateBuilding (storeManager.buildStateForFileCopy()) will be triggered at the end of FCM's async handler.
```
aga9900, line 44: Please change the method name to something generic in the actual interface. Like buildBlobStoreState(). thos should be a generic method and scope should not be limited to FileCopy in future use cases.

aga9900: Or let's call it buildReplicaStateInMemory(), to make it even more generic

piyujai: @aga9900 I think we should keep the name as is since it is building state specific to FileCopy (only for sealed segments, ignoring active segments). It wouldn't be reused elsewhere since there is no generic usecase.

aga9900 (approved):

ambry-file-transfer/src/main/java/com/github/ambry/FileCopyManager.java
aga9900: sounds good.

PR #2971, opened 2024-12-19
[mysql] Verify credentials for cert-based access

snalli:
This patch configures the mysql java driver to use CA certs installed on the machine to authenticate with mysql db.


gshantanu:
Should we close this PR out @justinlin-linkedin ? I don't believe Sanketh is working on it actively.

PR #2981, opened 2025-01-10
[S3] Validate parts in multi-part upload

alyssaxu333:
Currently there's no check for the invalid part number and duplicate etags. Add checks for part number that is negative, decimal or larger than 10000, throw Bad Request Error. Add checks for duplicate eTags in the request to avoid time out.

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
@@ -435,6 +435,15 @@ public class S3MultipartCompleteUploadHandler<R> {
    * @return the bad request error
    */
   private static void validateParts(List<Part> parts) throws RestServiceException {
+    if (parts == null || parts.isEmpty()) {
+      throw new RestServiceException(S3Constants.ERR_EMPTY_REQUEST_BODY, RestServiceErrorCode.BadRequest);
+    }
+
+    if (parts.size() > S3Constants.MAX_LIST_SIZE) {
+      String error = S3Constants.ERR_PART_LIST_TOO_LONG;
+      throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
+    }
+
     Set<Integer> partNumbers = new HashSet<>();
     Set<String> etags = new HashSet<>();
 
@@ -442,23 +451,19 @@ public class S3MultipartCompleteUploadHandler<R> {
       int partNumber = part.getPartNumber();
       String etag = part.geteTag();
 
-      if (partNumber < MIN_PART_NUM || partNumber > MAX_PART_NUM) {
-        String error = String.format(
-            "Invalid part number: " + part.getPartNumber() + ". Part number must be an integer between {} and {}.",
-            MIN_PART_NUM, MAX_PART_NUM);
-        LOGGER.error(error);
+      if (partNumber < S3Constants.MIN_PART_NUM || partNumber > S3Constants.MAX_PART_NUM) {
+        String error = String.format(S3Constants.ERR_INVALID_PART_NUMBER, partNumber, S3Constants.MIN_PART_NUM,
+            S3Constants.MAX_PART_NUM);
         throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
       }
 
       if (!partNumbers.add(partNumber)) {
-        String error = "Duplicate part number found: " + partNumber;
-        LOGGER.error(error);
+        String error = String.format(S3Constants.ERR_DUPLICATE_PART_NUMBER, partNumber);
         throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
       }
 
       if (!etags.add(etag)) {
-        String error = "Duplicate eTag found: " + etag;
-        LOGGER.error(error);
+        String error = String.format(S3Constants.ERR_DUPLICATE_ETAG, etag);
         throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
       }
     }
@@ -471,11 +476,11 @@ public class S3MultipartCompleteUploadHandler<R> {
    */
   private static void validatePartsSize(Part[] parts) throws RestServiceException {
     if (parts == null || parts.length == 0) {
-      throw new RestServiceException("Xml request body cannot be empty.", RestServiceErrorCode.BadRequest);
+      throw new RestServiceException(S3Constants.ERR_EMPTY_REQUEST_BODY, RestServiceErrorCode.BadRequest);
     }
 
-    if (parts.length > MAX_LIST_SIZE) {
-      String error = String.format("Parts list size cannot exceed {}.", MAX_LIST_SIZE);
+    if (parts.length > S3Constants.MAX_LIST_SIZE) {
+      String error = String.format(S3Constants.ERR_PART_LIST_TOO_LONG, S3Constants.MAX_LIST_SIZE);
       throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
     }
   }
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
@@ -64,7 +64,6 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executors;
-import org.apache.commons.math3.analysis.function.Min;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -388,18 +387,13 @@ public class S3MultipartCompleteUploadHandler<R> {
       // Get parts in order from CompleteMultipartUpload, deserialize each part id to get data chunk ids.
       List<ChunkInfo> chunkInfos = new ArrayList<>();
       try {
+        // check the parts array size
+        validatePartsSize(completeMultipartUpload.getPart());
         // sort the list in order
-        if (completeMultipartUpload.getPart() == null || completeMultipartUpload.getPart().length == 0) {
-          throw new RestServiceException("Xml request body cannot be empty.", RestServiceErrorCode.BadRequest);
-        }
         List<Part> sortedParts = Arrays.asList(completeMultipartUpload.getPart());
-        // Validate parts list and part number
+        // Validate part number and etags
         // https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
-        String badRequest = validateParts(sortedParts);
-        if (badRequest != null) {
-          LOGGER.error(badRequest);
-          throw new RestServiceException(badRequest, RestServiceErrorCode.BadRequest);
-        }
+        validateParts(sortedParts);
         Collections.sort(sortedParts, Comparator.comparingInt(Part::getPartNumber));
         String reservedMetadataId = null;
         for (Part part : sortedParts) {
@@ -440,11 +434,7 @@ public class S3MultipartCompleteUploadHandler<R> {
    * @param parts sorted parts list
    * @return the bad request error
    */
-  private static String validateParts(List<Part> parts) {
-    if (parts.size() > MAX_LIST_SIZE) {
-      return String.format("Parts list size cannot exceed {}.", MAX_LIST_SIZE);
-    }
-
+  private static void validateParts(List<Part> parts) throws RestServiceException {
     Set<Integer> partNumbers = new HashSet<>();
     Set<String> etags = new HashSet<>();
 
@@ -453,19 +443,40 @@ public class S3MultipartCompleteUploadHandler<R> {
       String etag = part.geteTag();
 
       if (partNumber < MIN_PART_NUM || partNumber > MAX_PART_NUM) {
-        return String.format(
+        String error = String.format(
             "Invalid part number: " + part.getPartNumber() + ". Part number must be an integer between {} and {}.",
             MIN_PART_NUM, MAX_PART_NUM);
+        LOGGER.error(error);
+        throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
       }
 
       if (!partNumbers.add(partNumber)) {
-        return "Duplicate part number found: " + partNumber;
+        String error = "Duplicate part number found: " + partNumber;
+        LOGGER.error(error);
+        throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
       }
 
       if (!etags.add(etag)) {
-        return "Duplicate eTag found: " + etag;
+        String error = "Duplicate eTag found: " + etag;
+        LOGGER.error(error);
+        throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
       }
     }
-    return null;
+  }
+
+  /**
+   * Check the parts array size is not empty or exceeds the limit
+   * @param parts the part array from Complete Multipart upload request
+   * @throws RestServiceException
+   */
+  private static void validatePartsSize(Part[] parts) throws RestServiceException {
+    if (parts == null || parts.length == 0) {
+      throw new RestServiceException("Xml request body cannot be empty.", RestServiceErrorCode.BadRequest);
+    }
+
+    if (parts.length > MAX_LIST_SIZE) {
+      String error = String.format("Parts list size cannot exceed {}.", MAX_LIST_SIZE);
+      throw new RestServiceException(error, RestServiceErrorCode.BadRequest);
+    }
   }
 }
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
@@ -64,6 +64,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executors;
+import org.apache.commons.math3.analysis.function.Min;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -83,6 +84,9 @@ import static com.github.ambry.router.RouterErrorCode.*;
 public class S3MultipartCompleteUploadHandler<R> {
   private static final Logger LOGGER = LoggerFactory.getLogger(S3MultipartCompleteUploadHandler.class);
   private static final ObjectMapper objectMapper = new XmlMapper();
+  private static final int MIN_PART_NUM = 1;
+  private static final int MAX_PART_NUM = 10000;
+  private static final int MAX_LIST_SIZE = 10000;
   private final SecurityService securityService;
   private final FrontendMetrics frontendMetrics;
   private final AccountAndContainerInjector accountAndContainerInjector;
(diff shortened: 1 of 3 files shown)


ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -382,12 +383,27 @@ private boolean isRetriable(Throwable throwable) {
...
       List<ChunkInfo> chunkInfos = new ArrayList<>();
+      Set<String> eTags = new HashSet<>();
       try {
         // sort the list in order
         List<Part> sortedParts = Arrays.asList(completeMultipartUpload.getPart());
         Collections.sort(sortedParts, Comparator.comparingInt(Part::getPartNumber));
         String reservedMetadataId = null;
         for (Part part : sortedParts) {
+          // Validate part number: must be a non-negative integer and within the range 0-10,000
+          if (part.getPartNumber() < 0 || part.getPartNumber() > 10000 || part.getPartNumber() != Math.floor(part.getPartNumber())) {
```
snalli, line 394: pull out the constants 0 and 10000 and name them MIN_PART_NUM and MAX_PART_NUM

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -382,12 +383,27 @@ private boolean isRetriable(Throwable throwable) {
...
+            String error = "Invalid part number: " + part.getPartNumber() + ". Must be a non-negative integer between 0 and 10,000.";
```
snalli, line 395: Use the constants to form the err msg

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -382,12 +383,27 @@ private boolean isRetriable(Throwable throwable) {
...
       // Get parts in order from CompleteMultipartUpload, deserialize each part id to get data chunk ids.
       List<ChunkInfo> chunkInfos = new ArrayList<>();
+      Set<String> eTags = new HashSet<>();
       try {
         // sort the list in order
         List<Part> sortedParts = Arrays.asList(completeMultipartUpload.getPart());
         Collections.sort(sortedParts, Comparator.comparingInt(Part::getPartNumber));
         String reservedMetadataId = null;
         for (Part part : sortedParts) {
+          // Validate part number: must be a non-negative integer and within the range 0-10,000
```
snalli, line 393: put a link to s3 doc
https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -382,12 +383,27 @@ private boolean isRetriable(Throwable throwable) {
     List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUpload) throws RestServiceException {
       // Get parts in order from CompleteMultipartUpload, deserialize each part id to get data chunk ids.
       List<ChunkInfo> chunkInfos = new ArrayList<>();
+      Set<String> eTags = new HashSet<>();
       try {
         // sort the list in order
         List<Part> sortedParts = Arrays.asList(completeMultipartUpload.getPart());
         Collections.sort(sortedParts, Comparator.comparingInt(Part::getPartNumber));
         String reservedMetadataId = null;
         for (Part part : sortedParts) {
```
snalli, line 392: abstract the loop out into a fn 
```
validateCompleteMultipartUpload(completeMultipartUpload)
```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
snalli, line 395: lets copy s3

```
InvalidPartNumber: Part number must be an integer between 1 and 10,000, inclusive.
```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
snalli: nvm, the loop is bigger than i thot

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
snalli, line 394: Also verify the size of the list. That should be less than 10000. Don't just verify part numbers. With this fix I can trick the system by passing a million entries all with part numbers less than 10000.

And this patch is adding two fixes. I'd recommend splitting this into two. One for etag. One for part numbers. For part numbers we need to be careful and ensure we are restricting customers from uploading large blobs.

snalli, line 394: Let's also add a check to disallow duplicate part numbers. For example, this should not be possible.

```
part 1 : etag-a
part 1 : etag-b

```

And then add a test for this. So overall there are? multiple checks

1. Disallow duplicate part numbers
2. Disallow duplicate etags
3. Check for list size 10000
4. Check for part numbers 1-10000

It'd be better to abstract this out into a single function and pass the sorted list to it. It's OK if we go over the list twice.

alyssaxu333: Done

alyssaxu333: "actually pass the unsorted list and check size first, if its more than 10k, then return immediately, saves the cost of sorting and other checks."
Will update on this one later, I think github has some site issue now

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -384,7 +389,17 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
       List<ChunkInfo> chunkInfos = new ArrayList<>();
       try {
         // sort the list in order
+        if (completeMultipartUpload.getPart() == null || completeMultipartUpload.getPart().length == 0) {
+          throw new RestServiceException("Xml request body cannot be empty.", RestServiceErrorCode.BadRequest);
+        }
```
snalli, line 394: `validateParts` before sorting, we don't want to sort a list of million parts.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+   * 1. Disallow duplicate part numbers
+   * 2. Disallow duplicate etags
+   * 3. Check for list size 10000
+   * 4. Check for part numbers integer 1-10000
+   * @param parts sorted parts list
+   * @return the bad request error
+   */
+  private static String validateParts(List<Part> parts) {
+    if (parts.size() > MAX_LIST_SIZE) {
+      return String.format("Parts list size cannot exceed {}.", MAX_LIST_SIZE);
```
snalli, line 445: Throw an exception instead of return a string.
```
          throw new RestServiceException(message, RestServiceErrorCode.BadRequest);

```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+
+    Set<Integer> partNumbers = new HashSet<>();
+    Set<String> etags = new HashSet<>();
+
+    for (Part part : parts) {
+      int partNumber = part.getPartNumber();
+      String etag = part.geteTag();
+
+      if (partNumber < MIN_PART_NUM || partNumber > MAX_PART_NUM) {
+        return String.format(
```
snalli, line 456: Throw an exception instead of return a string.
```
          throw new RestServiceException(message, RestServiceErrorCode.BadRequest);

```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+            "Invalid part number: " + part.getPartNumber() + ". Part number must be an integer between {} and {}.",
+            MIN_PART_NUM, MAX_PART_NUM);
+      }
+
+      if (!partNumbers.add(partNumber)) {
+        return "Duplicate part number found: " + partNumber;
```
snalli, line 462: Throw an exception instead of return a string.
```
          throw new RestServiceException(message, RestServiceErrorCode.BadRequest);

```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+      }
+
+      if (!etags.add(etag)) {
+        return "Duplicate eTag found: " + etag;
```
snalli, line 466: Throw an exception instead of return a string.
```
          throw new RestServiceException(message, RestServiceErrorCode.BadRequest);

```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -60,9 +60,11 @@
 import java.util.Comparator;
 import java.util.EnumSet;
 import java.util.GregorianCalendar;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.Executors;
+import org.apache.commons.math3.analysis.function.Min;
```
snalli, line 67: unused import

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -82,6 +84,9 @@
 public class S3MultipartCompleteUploadHandler<R> {
   private static final Logger LOGGER = LoggerFactory.getLogger(S3MultipartCompleteUploadHandler.class);
   private static final ObjectMapper objectMapper = new XmlMapper();
+  private static final int MIN_PART_NUM = 1;
```
snalli, line 87: Let's create a class like this and expose these constants and use them both in your fix and the tests.
```
package com.github.ambry.frontend.s3;

public class S3Constants {
  public static final int MIN_PART_NUM = 1;
  public static final int MAX_PART_NUM = 10000;
  public static final int MAX_LIST_SIZE = 10000;
  public static final String ERR_PART_LIST_TOO_LONG = "Parts list size cannot exceed %s";
  // TODO: More errors
}

```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -82,6 +84,9 @@
...
+  private static final int MAX_PART_NUM = 10000;
+  private static final int MAX_LIST_SIZE = 10000;
```
snalli, line 89: I doubt if you need MAX_LIST_SIZE. It will be the same as MAX_PART_NUM

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+   * @return the bad request error
+   */
+  private static String validateParts(List<Part> parts) {
+    if (parts.size() > MAX_LIST_SIZE) {
+      return String.format("Parts list size cannot exceed {}.", MAX_LIST_SIZE);
+    }
+
+    Set<Integer> partNumbers = new HashSet<>();
+    Set<String> etags = new HashSet<>();
+
```
snalli, line 450: Remove Whitespace

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+    for (Part part : parts) {
+      int partNumber = part.getPartNumber();
+      String etag = part.geteTag();
+
```
snalli, line 454: Rm white space

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+      if (partNumber < MIN_PART_NUM || partNumber > MAX_PART_NUM) {
+        return String.format(
+            "Invalid part number: " + part.getPartNumber() + ". Part number must be an integer between {} and {}.",
+            MIN_PART_NUM, MAX_PART_NUM);
+      }
+
```
snalli, line 460: Rm white space

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +430,42 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+      if (!partNumbers.add(partNumber)) {
+        return "Duplicate part number found: " + partNumber;
+      }
+
```
snalli, line 464: Rm white space

snalli:
Nit : change title to
[s3] Validate parts in multi-part upload

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
alyssaxu333: I think the current logic: return string and if the string is not null, log & throw error is more concise.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
alyssaxu333: Hmmm, it's same number, but the meaning is different. In the list, the part number can be {2, 3, 10020} while the list size is 3.

snalli: Won't this work?

```
if (list.size() > MAX_PART_NUM) {
	throw e
}
```

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
snalli: Take a look at some of the tickets we get. Log messages are usually redacted or not printed. The only thing to go fwd are teh stack traces. So that must be accurate to the line of code.

alyssaxu333: okay

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
alyssaxu333: It will work for now, but just as why refer to the name of the number in the code, it might change to different numbers in the future, like Max part = 10000 and list size = 5000. So differentiate them is necessary.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
alyssaxu333: I don't think this is necessary cuz all these parameters and messages only gonna be used in S3 Multipart upload handler and test. It will avoid adding unnecessary complexity to the project structure.
The constants remain scoped to where they are relevant, and developers can easily see the context and usage of these constants without switching files.

snalli: Once you create this file, others will be motivated to add or move more constants to this file as the s3 code develops. This is the style is many large and popular codebases. It doesn't add any complexity. It simplifies code, because all s3-constants are in one place making it easier to read, modify and understand if there are any dependencies between constants. Declaring error messages as constants allows you to use the same constant in your code as well as tests. Modifying the constant automatically updates the code and test. Otherwise, most of the times ppl change the err msg in one place but forget to change it in a test breaking it.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -415,4 +424,64 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
...
+      }
+    }
+  }
+
+  /**
+   * Check the parts array size is not empty or exceeds the limit
+   * @param parts the part array from Complete Multipart upload request
+   * @throws RestServiceException
+   */
+  private static void validatePartsSize(Part[] parts) throws RestServiceException {
```
snalli, line 477: Remove this and just call validateParts first thing in the try-block. Make sure that has all the validations. This is what I am trying to get at.

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3MultipartCompleteUploadHandler.java
```diff
@@ -383,8 +387,13 @@ List<ChunkInfo> getChunksToStitch(CompleteMultipartUpload completeMultipartUploa
       // Get parts in order from CompleteMultipartUpload, deserialize each part id to get data chunk ids.
       List<ChunkInfo> chunkInfos = new ArrayList<>();
       try {
+        // check the parts array size
+        validatePartsSize(completeMultipartUpload.getPart());
         // sort the list in order
         List<Part> sortedParts = Arrays.asList(completeMultipartUpload.getPart());
+        // Validate part number and etags
+        // https://docs.aws.amazon.com/AmazonS3/latest/userguide/qfacts.html
+        validateParts(sortedParts);
```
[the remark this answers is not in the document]
snalli: Like so
```
...
      try {
        validateParts(Arrays.asList(completeMultipartUpload.getPart()));
        // sort the list in order
        List<Part> sortedParts = Arrays.asList(completeMultipartUpload.getPart());
        Collections.sort(sortedParts, Comparator.comparingInt(Part::getPartNumber));
        String reservedMetadataId = null;
....
```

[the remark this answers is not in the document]
snalli: My bad, I thought we were sorting at L393 because of the variable name but we are not. So validateParts at L396 or L390 is enough but not both. Let me help you out by making that change.

snalli (approved):

PR #2982, opened 2025-01-13
Support for Helix Auto Registration

mudit-saxena:
## Summary
Support for Helix Auto Registration. This will enable Ambry DataNodes to auto-join Helix cluster. AutoRegistration changes were available in `helix-core@1.4.0` which is only Java11 compatible so I had to pin down dependency of this version to Java8 supported snapshot

## Testing

`./gradlew clean build`

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
@@ -110,6 +110,7 @@ public class ClusterMapUtils {
   static final int CURRENT_SCHEMA_VERSION = 0;
   static final String WRITABLE_LOG_STR = "writable";
   static final String FULLY_WRITABLE_LOG_STR = "fully writable";
+  private static final String INSTANCE_NAME_DELIMITER = "_";
   private static final Logger logger = LoggerFactory.getLogger(ClusterMapUtils.class);
 
   /**
@@ -160,7 +161,7 @@ public class ClusterMapUtils {
    * @return the constructed instance name.
    */
   public static String getInstanceName(String host, Integer port) {
-    return port == null ? host : host + "_" + port;
+    return port == null ? host : host + INSTANCE_NAME_DELIMITER + port;
   }
 
   /**
@@ -170,6 +171,10 @@ public class ClusterMapUtils {
    */
   public static String getInstanceName(DataNodeId dataNodeId) {
     return getInstanceName(dataNodeId.getHostname(), dataNodeId.getPort());
+  }
+
+  public static String getPortFromInstanceName(String instanceName) {
+    return instanceName.contains(INSTANCE_NAME_DELIMITER) ? instanceName.split(INSTANCE_NAME_DELIMITER)[1] : null;
   }
 
   /**
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
@@ -19,6 +19,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.ConcurrentHashMap;
 import org.apache.helix.HelixManager;
+import org.apache.helix.HelixManagerFactory;
 import org.apache.helix.HelixManagerProperty;
 import org.apache.helix.InstanceType;
 import org.apache.helix.constants.InstanceConstants;
@@ -48,10 +49,10 @@ public class HelixFactory {
    * @return the constructed {@link HelixManager}.
    */
   public HelixManager getZKHelixManager(String clusterName, String instanceName, InstanceType instanceType,
-      String zkAddr) {
+      String zkAddr, ClusterMapConfig clusterMapConfig) {
     ManagerKey managerKey = new ManagerKey(clusterName, instanceName, instanceType, zkAddr);
     return helixManagers.computeIfAbsent(managerKey,
-        k -> buildZKHelixManager(clusterName, instanceName, instanceType, zkAddr));
+        k -> buildZKHelixManager(clusterName, instanceName, instanceType, zkAddr, clusterMapConfig));
   }
 
   /**
@@ -64,8 +65,8 @@ public class HelixFactory {
    * @throws Exception if connecting failed.
    */
   public HelixManager getZkHelixManagerAndConnect(String clusterName, String instanceName, InstanceType instanceType,
-      String zkAddr) throws Exception {
-    HelixManager manager = getZKHelixManager(clusterName, instanceName, instanceType, zkAddr);
+      String zkAddr, ClusterMapConfig clusterMapConfig) throws Exception {
+    HelixManager manager = getZKHelixManager(clusterName, instanceName, instanceType, zkAddr, clusterMapConfig);
     synchronized (manager) {
       if (!manager.isConnected()) {
         LOGGER.info("Connecting to HelixManager at {}", zkAddr);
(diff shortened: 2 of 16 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -96,7 +101,14 @@ public DataNodeConfigSource getDataNodeConfigSource(ClusterMapConfig clusterMapC
    * @return a new instance of {@link HelixManager}.
    */
   HelixManager buildZKHelixManager(String clusterName, String instanceName, InstanceType instanceType, String zkAddr) {
-    return HelixManagerFactory.getZKHelixManager(clusterName, instanceName, instanceType, zkAddr);
+    HelixManagerProperty participantHelixProperty = new HelixManagerProperty.Builder().setDefaultInstanceConfigBuilder(
+        new InstanceConfig.Builder().setInstanceOperation(InstanceConstants.InstanceOperation.UNKNOWN).setPort(INSTANCE_CONFIG_HELIX_PORT)).build();
+
+    HelixManager helixManager = new ZKHelixManager(clusterName, instanceName, instanceType, zkAddr, null,
+        instanceType == InstanceType.PARTICIPANT ?  participantHelixProperty :
```
DevenAhluwalia, line 108: nit: Better that we create a new prop
`defaultHelixProperty = new HelixManagerProperty.Builder().setDefaultInstanceConfigBuilder(new InstanceConfig.Builder()).build())`

instead of inlining it in params for `new ZKHelixManager()`

DevenAhluwalia, line 108: For my understanding, why for all other Helix.InstanceTypes 
`CONTROLLER, CONTROLLER_PARTICIPANT, SPECTATOR, ADMINISTRATOR` we default to defaultHelixProperty 
and only for InstanceType = PARTICIPANT, we are setting InstanceOperation and Port?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -96,7 +101,14 @@ public DataNodeConfigSource getDataNodeConfigSource(ClusterMapConfig clusterMapC
    * @return a new instance of {@link HelixManager}.
    */
   HelixManager buildZKHelixManager(String clusterName, String instanceName, InstanceType instanceType, String zkAddr) {
-    return HelixManagerFactory.getZKHelixManager(clusterName, instanceName, instanceType, zkAddr);
+    HelixManagerProperty participantHelixProperty = new HelixManagerProperty.Builder().setDefaultInstanceConfigBuilder(
+        new InstanceConfig.Builder().setInstanceOperation(InstanceConstants.InstanceOperation.UNKNOWN).setPort(INSTANCE_CONFIG_HELIX_PORT)).build();
```
DevenAhluwalia, line 105: Checking [InstanceConstants.InstanceOperation.UNKNOWN](https://jarvis.corp.linkedin.com/codesearch/result/?name=InstanceConstants.java&path=helix%2Fhelix-common%2Fsrc%2Fmain%2Fjava%2Forg%2Fapache%2Fhelix%2Fconstants&reponame=linkedin%2Fhelix#71), 
Should this not be `ENABLE` ?

UNKNOWN corresponds to - "Behavior: Node will have all of its replicas dropped immediately and will be removed from the RoutingTableProvider"

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -33,6 +37,7 @@ public class HelixFactory {
   // exposed for use in testing
   private final Map<ManagerKey, HelixManager> helixManagers = new ConcurrentHashMap<>();
   private final Map<String, DataNodeConfigSource> dataNodeConfigSources = new ConcurrentHashMap<>();
+  private static final String INSTANCE_CONFIG_HELIX_PORT = "15088";
```
manbearpig1996, line 40: You need to use the port provided in clustermap config

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -96,7 +101,14 @@ public DataNodeConfigSource getDataNodeConfigSource(ClusterMapConfig clusterMapC
    * @return a new instance of {@link HelixManager}.
    */
   HelixManager buildZKHelixManager(String clusterName, String instanceName, InstanceType instanceType, String zkAddr) {
```
manbearpig1996, line 103: instanceName will also have port appended https://github.com/linkedin/ambry/blob/6683f022ee4b740d1d5eb90fa5207c2261a7c8a5/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java#L116
May be you can use that and throw error when port is not present.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -96,7 +101,14 @@ public DataNodeConfigSource getDataNodeConfigSource(ClusterMapConfig clusterMapC
...
-    return HelixManagerFactory.getZKHelixManager(clusterName, instanceName, instanceType, zkAddr);
+    HelixManagerProperty participantHelixProperty = new HelixManagerProperty.Builder().setDefaultInstanceConfigBuilder(
```
manbearpig1996, line 104: You need to create a flag in config to switch b/w old logic and new logic.

manbearpig1996: Also you can add a comment explaining what is autoregistration.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -96,7 +101,14 @@ public DataNodeConfigSource getDataNodeConfigSource(ClusterMapConfig clusterMapC
...
+        new InstanceConfig.Builder().setInstanceOperation(InstanceConstants.InstanceOperation.UNKNOWN).setPort(INSTANCE_CONFIG_HELIX_PORT)).build();
+
+    HelixManager helixManager = new ZKHelixManager(clusterName, instanceName, instanceType, zkAddr, null,
```
manbearpig1996, line 107: If possible, write a test case for auto registration code.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
```diff
@@ -96,7 +101,14 @@ public DataNodeConfigSource getDataNodeConfigSource(ClusterMapConfig clusterMapC
...
+        instanceType == InstanceType.PARTICIPANT ?  participantHelixProperty :
+            new HelixManagerProperty.Builder().setDefaultInstanceConfigBuilder(new InstanceConfig.Builder()).build());
```
manbearpig1996, line 109: Can you also explain the case when node restarts, it is already registered to helix and what will auto registration do then?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
mudit-saxena: Good Q. We use only PARTICIPANT for instance config information. So, we can use default helix property for all other instance types specifically SPECTATOR

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
mudit-saxena: As per helix team node should join the cluster in UNKNOWN state and once the registration is completed it will marked as ENABLED

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
mudit-saxena: In case node is already registered, auto-registration will not override any existing config

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
DevenAhluwalia: Ack.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
DevenAhluwalia: Ack.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
mudit-saxena: There is no need for flag here as we are not changing the way we connect to helix. In this we are providing an InstanceConfig as extra configuration while connection

manbearpig1996: This change will be tested in EI for sometime. We will need prod deployments to go through while wait for validation. And if we encounter some issue, we can just turn off the the flag.

mudit-saxena: There is no auto registration at this point. Auto registration will have two requisites after this enabling cloud config and taking dependency on helix-cloud in closed. 

Now for this specific change, it is change to the factory method on how we  instantiate the HelixManager. I have added an additional test as safety check. 

I don't think there should be any issue due to this but I can hold off from merging to test on perf before merging this

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
mudit-saxena: > There is no auto registration at this point. Auto registration will have two requisites after this enabling cloud config and taking dependency on helix-cloud in closed.

Added an additional test to ensure instance joins in ENABLE state same as the existing behaviour

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
mudit-saxena: Done

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
mudit-saxena: Done

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixFactory.java
manbearpig1996: sure, check if this is not getting used in VCR too.

manbearpig1996 (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
```diff
@@ -172,6 +173,10 @@ public static String getInstanceName(DataNodeId dataNodeId) {
     return getInstanceName(dataNodeId.getHostname(), dataNodeId.getPort());
   }
 
+  public static String getPortFromInstanceName(String instanceName) {
```
manbearpig1996, line 176: Can we return Integer here as port will be integer only and port is integer in above method too?
Can we add a small testcase for this?

manbearpig1996 (commented):
Looks overall good to me. Added one minor comment.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/ClusterMapUtils.java
mudit-saxena: We are using port as string in extraction, setPort() and getInstanceName

manbearpig1996 (approved):
LGTM

ambry-clustermap/src/test/java/com/github/ambry/clustermap/HelixParticipantTest.java
```diff
@@ -822,6 +823,29 @@ public void testStateModelParticipation() throws Exception {
     helixParticipant.close();
   }
 
+
+  @Test
+  public void testEnableStatePostParticipate() throws IOException {
+    assumeTrue(stateModelDef.equals(ClusterMapConfig.AMBRY_STATE_MODEL_DEF));
+    ClusterMapConfig clusterMapConfig = new ClusterMapConfig(new VerifiableProperties(props));
```
manbearpig1996, line 830: This test case will run on old logic. You need to set autoregistration true in the props here.

PR #2996, opened 2025-01-27
Add delimiter support for S3 List API

Arun-LinkedIn:
This adds support for treating "/" in blob names as delimiter. It allows to treat prefixes as "subdirectories" and group them under "CommonPrefixes" response in the LIST API. This also enables us to list and delete directories using AWS S3 CLIs

More details can be found in https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html under "CommonPrefixes" section.

```
For example, if we have below files (with names delimited by / to represent directories)

Ambry account: named-blob-sandbox
Ambry container: container-a
1. myfile.txt
2.folder/file1.txt
3.folder/file2.txt

1. cURL response for List API is 

curl -X GET 'http://localhost:1174/s3/named-blob-sandbox/container-a?prefix=folder&delimiter=/' | xmllint --format -
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   409  100   409    0     0   3976      0 --:--:-- --:--:-- --:--:--  3970
<?xml version="1.0"?>
<ListBucketResult>
  <Name>container-a</Name>
  <Prefix>folder</Prefix>
  <MaxKeys>1000</MaxKeys>
  <KeyCount>2</KeyCount>
  <Delimiter>/</Delimiter>
  <Contents>
    <Key>folder/file1.txt</Key>
    <LastModified>2025-01-27T09:37:23Z</LastModified>
    <Size>15</Size>
  </Contents>
  <Contents>
    <Key>folder/file2.txt</Key>
    <LastModified>2025-01-27T09:37:23Z</LastModified>
    <Size>15</Size>
  </Contents>
  <IsTruncated>false</IsTruncated>
</ListBucketResult>

2. AWS S3 CLI responses for list and remove directories are:

>aws s3 ls s3://container-a/ --recursive
2025-01-26 12:39:31         15 folder/file1.txt
2025-01-26 12:39:31         15 folder/file2.txt
2025-01-26 12:37:51         15 myfile.txt

>aws s3 ls s3://container-a/
                               PRE folder/
2025-01-26 12:37:51    15      myfile.txt

>aws s3 rm s3://container-a/folder/ --recursive
delete: s3://container-a/folder/file1.txt
delete: s3://container-a/folder/file2.txt

```

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
@@ -31,18 +31,15 @@ import com.github.ambry.utils.ByteBufferDataInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.text.SimpleDateFormat;
 import java.time.Instant;
-import java.time.LocalDateTime;
 import java.time.ZoneId;
-import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
-import java.util.Calendar;
-import java.util.Date;
 import java.util.GregorianCalendar;
+import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Set;
 import org.json.JSONObject;
 import org.json.JSONTokener;
 import org.slf4j.Logger;
@@ -68,6 +65,7 @@ public class S3ListHandler extends S3BaseHandler<ReadableStreamChannel> {
   private final NamedBlobListHandler namedBlobListHandler;
   private final FrontendMetrics metrics;
   public static final DateTimeFormatter TIMESTAMP_FORMATTER = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ssXXX");
+  private static final String DELIMITER = "/";
 
   /**
    * Constructs a handler for handling s3 requests for listing blobs.
@@ -115,10 +113,16 @@ public class S3ListHandler extends S3BaseHandler<ReadableStreamChannel> {
     String continuationToken = getHeader(restRequest.getArgs(), CONTINUATION_TOKEN, false);
     //By default S3 list returns up to 1000 key names.
     int maxKeysValue = maxKeys == null ? DEFAULT_MAX_KEY_VALUE : Integer.parseInt(maxKeys);
+    int keyCount = namedBlobRecordPage.getEntries().size();
     // Iterate through list of blob names.
     List<Contents> contentsList = new ArrayList<>();
-    int keyCount = 0;
+    // Use LinkedHashSet to maintain order of directories.
+    Set<String> dirSet = new LinkedHashSet<>();
     for (NamedBlobListEntry namedBlobRecord : namedBlobRecordPage.getEntries()) {
+      if (namedBlobRecord.isDirectory()) {
+        dirSet.add(namedBlobRecord.getBlobName());
+        continue;
+      }
       String blobName = namedBlobRecord.getBlobName();
       long blobSize = namedBlobRecord.getBlobSize();
       long modifiedTimeMs = namedBlobRecord.getModifiedTimeMs();
@@ -128,23 +132,26 @@ public class S3ListHandler extends S3BaseHandler<ReadableStreamChannel> {
       ZonedDateTime zonedDateTime = Instant.ofEpochMilli(modifiedTimeMs).atZone(ZoneId.of("UTC"));
       String formattedDate = zonedDateTime.format(TIMESTAMP_FORMATTER);
       contentsList.add(new Contents(blobName, formattedDate, blobSize));
-      if (++keyCount == maxKeysValue) {
-        break;
-      }
     }
+
+    List<Prefix> commonPrefixes = new ArrayList<>();
+    for (String dir : dirSet) {
+      commonPrefixes.add(new Prefix(dir));
+    }
+
     if (LIST_TYPE_VERSION_2.equals(getHeader(restRequest.getArgs(), LIST_TYPE, false))) {
       ListBucketResultV2 resultV2 =
-          new ListBucketResultV2(containerName, prefix, maxKeysValue, keyCount, delimiter, contentsList,
-              encodingType, continuationToken, namedBlobRecordPage.getNextPageToken(),
-              namedBlobRecordPage.getNextPageToken() != null);
+          new ListBucketResultV2(containerName, prefix, maxKeysValue, keyCount, delimiter, contentsList, encodingType,
+              continuationToken, namedBlobRecordPage.getNextPageToken(), namedBlobRecordPage.getNextPageToken() != null,
+              delimiter != null && delimiter.equals(DELIMITER) ? commonPrefixes : null);
       LOGGER.debug("Sending response for S3 ListObjects {}", resultV2);
       // Serialize xml
       xmlMapper.writeValue(outputStream, resultV2);
     } else {
       ListBucketResult result =
-          new ListBucketResult(containerName, prefix, maxKeysValue, keyCount, delimiter, contentsList,
-              encodingType, marker, namedBlobRecordPage.getNextPageToken(),
-              namedBlobRecordPage.getNextPageToken() != null);
+          new ListBucketResult(containerName, prefix, maxKeysValue, keyCount, delimiter, contentsList, encodingType,
+              marker, namedBlobRecordPage.getNextPageToken(), namedBlobRecordPage.getNextPageToken() != null,
+              delimiter != null && delimiter.equals(DELIMITER) ? commonPrefixes : null);
       LOGGER.debug("Sending response for S3 ListObjects {}", result);
       // Serialize xml
       xmlMapper.writeValue(outputStream, result);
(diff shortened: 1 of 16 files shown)


ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -706,10 +709,27 @@ private Page<NamedBlobRecord> run_list_v2(String accountName, String containerNa
         int resultIndex = 0;
         while (resultSet.next()) {
           String blobName = resultSet.getString(1);
-          if (resultIndex++ == maxKeysValue) {
+          if (resultIndex == maxKeysValue) {
```
justinlin-linkedin, line 712: we only set nextContinuationToken to the blobName when resultIndex == maxKeysValue. The idea is to only do that when it's the last key, but now that we change the logic when incrementing resultIndex. resultIndex would not equal to maxKeysValue in the last key, so we probably have to use an other index, or other way to set next continuation token here.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -706,10 +709,27 @@ private Page<NamedBlobRecord> run_list_v2(String accountName, String containerNa
...
             nextContinuationToken = blobName;
             break;
           }
+
+          if (groupDirectories) {
+            // Extract the portion after the prefix and before the next '/'
+            String remainingPath = blobName.substring(blobNamePrefix == null ? 0 : blobNamePrefix.length());
+            remainingPath = remainingPath.startsWith("/") ? remainingPath.substring(1) : remainingPath;
+            int delimiterIndex = remainingPath.indexOf("/");
```
justinlin-linkedin, line 721: let's use a constant for "/".

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -706,10 +709,27 @@ private Page<NamedBlobRecord> run_list_v2(String accountName, String containerNa
...
-          if (resultIndex++ == maxKeysValue) {
+          if (resultIndex == maxKeysValue) {
             nextContinuationToken = blobName;
             break;
           }
+
+          if (groupDirectories) {
+            // Extract the portion after the prefix and before the next '/'
+            String remainingPath = blobName.substring(blobNamePrefix == null ? 0 : blobNamePrefix.length());
+            remainingPath = remainingPath.startsWith("/") ? remainingPath.substring(1) : remainingPath;
```
justinlin-linkedin, line 720: this is not right, if the prefix is "abc" and this blobname is "abc/efg/hij", and the delimiter is "/", then the we should "abc/" as common prefix, not "abc/efg/".

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -718,8 +738,17 @@ private Page<NamedBlobRecord> run_list_v2(String accountName, String containerNa
...
           entries.add(
               new NamedBlobRecord(accountName, containerName, blobName, blobId, timestampToMs(deletionTime), version,
-                  blobSize, timestampToMs(modifiedTime)));
+                  blobSize, timestampToMs(modifiedTime), false));
         }
+
+        if (groupDirectories) {
+          // Add the directories to the result
+          entries.addAll(directories.stream()
+              .map(directory -> new NamedBlobRecord(accountName, containerName, directory, null, Utils.Infinite_Time, 0,
```
justinlin-linkedin, line 747: This is not right, when we are returning a common prefix, it should include the prefix as well. if we have "abc/def/ghi", and the prefix is "abc/", and the common prefix should be "abc/def/", not jsut "def/".

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -706,10 +709,27 @@ private Page<NamedBlobRecord> run_list_v2(String accountName, String containerNa
...
             break;
           }
+
+          if (groupDirectories) {
+            // Extract the portion after the prefix and before the next '/'
+            String remainingPath = blobName.substring(blobNamePrefix == null ? 0 : blobNamePrefix.length());
+            remainingPath = remainingPath.startsWith("/") ? remainingPath.substring(1) : remainingPath;
+            int delimiterIndex = remainingPath.indexOf("/");
+            if (delimiterIndex != -1) {
+              boolean validEntry = directories.add(remainingPath.substring(0, delimiterIndex) + "/");
```
justinlin-linkedin, line 723: Nit: it could be remainingPath.substring(0, delimiterIndex +DELIMITER_STRING.length());

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -706,10 +709,27 @@ private Page<NamedBlobRecord> run_list_v2(String accountName, String containerNa
...
           String blobName = resultSet.getString(1);
-          if (resultIndex++ == maxKeysValue) {
+          if (resultIndex == maxKeysValue) {
             nextContinuationToken = blobName;
             break;
           }
+
+          if (groupDirectories) {
+            // Extract the portion after the prefix and before the next '/'
+            String remainingPath = blobName.substring(blobNamePrefix == null ? 0 : blobNamePrefix.length());
```
alyssaxu333, line 719: Maybe we need to check whether the blob name contains prefix first. The blob name that doesn't have the prefix will be counted as a directory too. eg. prefix is f1/, and blobName is abcdefg/image1, the common prefix will be defg/

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
```diff
@@ -115,7 +115,15 @@ private ReadableStreamChannel serializeAsXml(RestRequest restRequest, Page<Named
     // Iterate through list of blob names.
     List<Contents> contentsList = new ArrayList<>();
     int keyCount = 0;
+    Set<String> dirSet = new HashSet<>();
     for (NamedBlobListEntry namedBlobRecord : namedBlobRecordPage.getEntries()) {
+      if (namedBlobRecord.isDirectory()) {
+        dirSet.add(namedBlobRecord.getBlobName());
+        if (++keyCount == maxKeysValue) {
```
alyssaxu333, line 122: QQ: is the size of entries list supposed to be the same as maxKeyValue? If yes, then  if (++keyCount == maxKeysValue)  might be redundant. If no, then the NextContinuationToken might need to be updated after the loop. Just a little confused about this

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
Arun-LinkedIn: Sure. Reverted back to previous logic of incrementing `resultIndex` on every iteration and  setting `nextContinuationToken` to the blobName when `resultIndex == maxKeysValue`

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
Arun-LinkedIn: I also thought of the same :) Since we are querying mysql for names starting with the prefix, blob names will always contain the prefix. Due to that, the additional check may not be needed

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
Arun-LinkedIn: Sure! Corrected it

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
Arun-LinkedIn: Done

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
Arun-LinkedIn: Done

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
Arun-LinkedIn: Corrected it..

ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3ListHandler.java
Arun-LinkedIn: You are right :) Updated to reuse the number of entries in the `Page` as key count. Originally, I just wanted to group directories here so that they can be placed under `CommonPrefixes` xml element.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
alyssaxu333: Got it!

justinlin-linkedin (approved):
LGTM

PR #3000, opened 2025-02-05
making replication log as trace when one cycle is finished

manbearpig1996:
## Summary
making replication log as trace when one cycle is finished

## Testing Done
Unit tests and integration test


vaibhav-mittal-linkedin (approved):
LGTM

PR #3001, opened 2025-02-10
adding logic for picking multiple replicas in server performance put test

manbearpig1996:
## Summary

Adding logic for picking multiple replicas in server performance put test

## Testing Done
Not required


vaibhav-mittal-linkedin (approved):
LGTM

PR #3005, opened 2025-02-13
Fix the prefix requirement in ListObjects

alyssaxu333:
## Summary
ListObjects request without prefix will throw 400 error. But Amazon S3 ListObject request doesn't require prefix, it will list all objects in the bucket. Fix the request requirement.

## Testing Done
Unit test


ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
```diff
@@ -143,6 +138,17 @@ public static NamedBlobPath parse(RequestPath requestPath, Map<String, Object> a
...
   }
 
+  /**
+   * Parse the request to see if it's a List request
+   * @param restRequest
+   * @return
+   * @throws RestServiceException
+   */
+  public static NamedBlobPath parse(RestRequest restRequest) throws RestServiceException {
+    restRequest.setArg(InternalKeys.REST_METHOD, restRequest.getRestMethod());
```
SophieGuo410, line 148: May I know why do we need this extra internal header? 
If you want the rest method you can directly get it right?
If you want to know which request is S3 request, you can use isS3Request(restRequest).

alyssaxu333: So in parseS3, I added a check for whether it's a listObject request or not. To determine if it's a listObject request, the rest method should be "Get" and the blob uri should be s3/account/container. But parseS3(String path, Map<String, Object> args) doesn't have rest method passed into it. I was thinking about changing it to parseS3(RestRequest restRequest) or adding rest method header in args beforehand. Changing signatures seems to be a bigger change thus I chose the second way. Pls let me know which way do you think is better, or if there's other ways to revise it that I didn't consider.

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
```diff
@@ -522,6 +527,11 @@ public static final class InternalKeys {
      * content-length header
      */
     public static final String CONTENT_RANGE_LENGTH = KEY_PREFIX + "content-range-length";
+
+    /**
+     * Boolean field set to "true" if this is a S3 request.
```
Arun-LinkedIn, line 532: Nit: Looks like copy-paste typo

ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
Arun-LinkedIn: Changing signatures seem to be cleaner approach. But, if it involves changing too many files, may be adding the Rest method to args is also okay. 

I think we could also add Rest method to args in [FrontendRestRequestService#preProcessAndRouteRequest](https://github.com/linkedin/ambry/blob/2b967d0fede7670eada6cdc88712fe177ea66f12/ambry-frontend/src/main/java/com/github/ambry/frontend/FrontendRestRequestService.java#L547) so that it is available to all requests.

SophieGuo410 (approved):
lgtm, can you please remove the commented code, thx!

Arun-LinkedIn (approved):

ambry-api/src/main/java/com/github/ambry/frontend/NamedBlobPath.java
alyssaxu333: Sorry forgot to reply after pushing the code. I added it in the FrontendRequest as you suggested. Thank you!

ambry-api/src/main/java/com/github/ambry/rest/RestUtils.java
alyssaxu333: Revised

PR #3006, opened 2025-02-14
Add more metrics for compaction and sealed replicas and enable compaction after bootstrap

justinlin-linkedin:
## Summary
When a replica starts bootstrapping on a host, we disable compaction for this replica. This makes sense since bootstrap process employs the regular replication protocol, which would ignore expired and deleted blob. There are not many blobs to compact. However, after bootstrap finishes, this replica would start serving client requests, more blobs will be deleted and expired. We should enable compaction after bootstrap. 

This PR fixed this issue by adding logic to enable compaction in storage manager.

This PR also exposes two more metrics
1. showing how many replicas are disabled for compaction on each host
2. showing how many replicas are sealed on each host.

## Testing Done
Wait for unit test


Arun-LinkedIn (approved):
lgtm!

PR #3009, opened 2025-02-18
helix transition metric

manbearpig1996:
## Summary
helix transition metric

## Testing Done
gradlew test

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipantMetrics.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipantMetrics.java
@@ -25,6 +25,9 @@ import java.util.Map;
  * Metrics for {@link HelixParticipant} to monitor partition state transitions.
  */
 class HelixParticipantMetrics {
+  private static final String transitionUpdateTemplate = "Partition-%s-from-%s-to-%s";
+
+  private MetricRegistry registry;
   private Map<ReplicaState, Integer> replicaCountByState = new HashMap<>();
   private final Map<String, ReplicaState> localPartitionAndState;
   // no need to record exact number of "dropped" partition, a counter to track partition-dropped events would suffice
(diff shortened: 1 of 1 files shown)


ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -1036,20 +1048,27 @@ public void onPartitionBecomeDroppedFromOffline(String partitionName) {
...
         partitionName);
     setPartitionDisabledState(partitionName, false);
     localPartitionAndState.remove(partitionName);
     participantMetrics.partitionDroppedCount.inc();
+    participantMetrics.decStateTransitionMetric(partitionName, ReplicaState.ERROR, ReplicaState.DROPPED);
+    participantMetrics.clearStateTransitionMetric(partitionName);
   }
 
   @Override
   public void onPartitionBecomeOfflineFromError(String partitionName) {
```
DevenAhluwalia, line 1068: This doesn't need `clearStateTransitionMetric` ?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
```diff
@@ -1036,20 +1048,27 @@ public void onPartitionBecomeDroppedFromOffline(String partitionName) {
     setPartitionDisabledState(partitionName, false);
     localPartitionAndState.remove(partitionName);
     participantMetrics.partitionDroppedCount.inc();
+    participantMetrics.decStateTransitionMetric(partitionName, ReplicaState.OFFLINE, ReplicaState.DROPPED);
+    participantMetrics.clearStateTransitionMetric(partitionName);
```
DevenAhluwalia, line 1052: Why do we need to do `registry.remove` in these downward transitions?

mudit-saxena:
@manbearpig1996 Feel free to add your thoughts on this. 
Can we have bootstrap time at host level instead of partition granularity or even better could be host-disk combination
You can push P95 metrics for it. Instead of emitting every single partition time

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
manbearpig1996: Dropped is the last state for a partition, so clearing a metric only when it is dropped.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/HelixParticipant.java
manbearpig1996: Since there could be a lot of movement for a partitions in a host, Dropping these for a partition when a partition gets removed from the host, to save memory and prevent memory leaks.

manbearpig1996:
@mudit-saxena This metric would be necessary in our performance tests. 
I will add separate metric for time taken per disk for clusters with large host.

piyujai (approved):
LGTM

PR #3018, opened 2025-02-27
Recover from last log segment with invalid messages

justinlin-linkedin:
## Summary
Since we don't ignore invalid log segment from this PR https://github.com/linkedin/ambry/pull/2744, we run into lots of error replicas due to invalid messages in last log segment when power outage happens. This PR adds configuration and logic to deal with those errors. 

When there are invalid messages in the log segment when persistent index is trying to recover index segment files, we check if the log segment is the last log segment and if the amount of remaining bytes are small, if so, we just ignore the invalid message and move on with the valid message.

## Testing Done
Added unit tests.

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
@@ -214,6 +214,23 @@ class LogSegment implements Read, Write {
     return new DirectIOBufferedExecutor(bufferSize);
   }
 
+  void truncateTo(long offset) throws IOException {
+    long currentFileSize = sizeInBytes();
+    if (currentFileSize < offset) {
+      throw new IllegalArgumentException(
+          "Illegal offset to truncate to, current file size is " + currentFileSize + " new offset is " + offset);
+    }
+    if (currentFileSize == offset) {
+      return;
+    }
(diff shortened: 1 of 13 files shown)


ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
```diff
@@ -214,6 +214,29 @@ private DirectIOExecutor createDirectIOExecutor(int bufferSize) {
...
 
+  /**
+   * Truncate the log segment file to the given offset. If the endOffset is already set, then this operation
+   * should be illegal.
+   * @param offset
+   * @throws IOException
+   */
+  void truncateTo(long offset) throws IOException {
+    long currentFileSize = sizeInBytes();
+    if (currentFileSize < offset) {
```
SophieGuo410, line 225: QQ: in which scenario this will happen?

SophieGuo410 (approved):
LGTM! Just left one question.

ambry-store/src/main/java/com/github/ambry/store/LogSegment.java
justinlin-linkedin: it should never happen, adding it here just for sanity check.

Arun-LinkedIn (approved):

PR #3028, opened 2025-03-10
Prioritized Partitions Simulation

mudit-saxena:
## Summary

This PR experiment with estimated completion time for bootstrap of X% of total partitions assigned to the host. This will validate the idea that prioritized partitions can be completed in Y hours 

## Testing Done
`./gradlew clean build && ./gradlew allJar`


ambry-replication/src/main/java/com/github/ambry/replication/continuous/DataNodeTracker.java
```diff
@@ -61,7 +61,13 @@ public DataNodeTracker(DataNodeId dataNodeId, List<RemoteReplicaInfo> remoteRepl
 
     // for each of smaller array of remote replicas create active group trackers with consecutive group ids
     for (List<RemoteReplicaInfo> remoteReplicaList : remoteReplicaSegregatedList) {
-      ActiveGroupTracker activeGroupTracker = new ActiveGroupTracker(currentGroupId, remoteReplicaList.stream()
+      int size = remoteReplicaList.size();
```
DevenAhluwalia, line 64: Lets extract this code piece so that its a single point where this logic resides. Currently this is being evaluated at 2 places and any change in one place must be replicated to the other location. Roughly :-

 ```
list foo(inList, bool IsReplicationEnablePrioritzation, replicationMaxPrioritizedReplicasPercent) {
     int size = inList.size();
     if (IsReplicationEnablePrioritzation) {
              int maxSize = replicationMaxPrioritizedReplicasPercent/100 * size;
              size = Math.min(size, maxSize);
    }
    return inList.subList(0, size)
}
```

**Review by DevenAhluwalia — COMMENTED** [review]
This is decently risky change in the critical section of Replication. 
Although this is controlled using cfg2, Is there a way to further limit this to one host? 

At the very least, imo, we should test this change in Perf/Ei using binary hot reload way.

File: ambry-replication/src/main/java/com/github/ambry/replication/continuous/DataNodeTracker.java [hunk 2]
```diff
@@ -61,7 +61,13 @@ public DataNodeTracker(DataNodeId dataNodeId, List<RemoteReplicaInfo> remoteRepl
...
+      if (isReplicaPrioritzationEnabled) {
+        int maxSize = replicationMaxPrioritizedReplicas/100 * size;
```
manbearpig1996, line 66: You will not be able to get accurate data from this methodology. As replicas for partition can be in different threads and different data node tracker. So you will stop one replica and not stop another replica. Also after each iteration, the list could change and different partition will be picked up.

gshantanu:
What's the next step on this PR?  It has been open for nearly a month.

PR #3029, opened 2025-03-10
Added metrics for ReplicaThread 

mudit-saxena:
## Summary

Added metrics for ReplicaThread  to understand the three parts

1. Number of remoteReplicaInfos assigned per thread
2. Number of cycle iterations performed per thread
3. One cycle replication time per thread 

## Testing Done

`./gradlew clean build && ./gradlew allJar`


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -163,6 +163,10 @@ public class ReplicationMetrics {
   private final Counter intraColoRetryAfterBackoffErrorCount;
   private final Map<String, Counter> dcToResponseError = new ConcurrentHashMap<>();
   private final Counter intraColoResponseErrorCount;
+  public final Map<String, Counter> replicaThreadsAssignedRemoteReplicaInfo = new ConcurrentHashMap<>();
```
manbearpig1996, line 166: Why do we want to track it for each thread ?
Data should be same across each thread and we already have data for it.

manbearpig1996, line 166: You can just publish meter type cycle iteration metric and 
 meter type  replicaThreadsAssignedRemoteReplicaInfo metric in the end of each replication cycle .

mudit-saxena: I want to analyse the performance of each replica thread against the assigned remoteReplicaInfo. Also, look at how fair the distribution of remoteReplicaInfo is not only by count but by each replicationCycle time as well

mudit-saxena: I want to compare data across replica threads

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -163,6 +163,10 @@ public class ReplicationMetrics {
...
+  public final Map<String, Counter> replicaThreadsCycleIterations = new ConcurrentHashMap<>();
+  public final Map<String, Histogram> replicaThreadsOneCycleReplicationTime = new ConcurrentHashMap<>();
```
manbearpig1996, line 168: Can you create methods to update these values in this class? 
If metric for a thread is null, you can create the metric object. 
OR
you can call the method to populate metric object in the constructor of replica thread.
You do not need to make changes outside of ReplicaThread that way.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -454,6 +458,20 @@ public void populateSingleColoMetrics(String datacenter) {
     dcToResponseError.put(datacenter, responseError);
   }
 
+
+  public void populateReplicaThreadMetrics(String replicaThread) {
```
manbearpig1996, line 462: Can you add logic in shutdown of replica thread to clear all metric for that thread.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
manbearpig1996: In these methods you need to add null checks for safety.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
manbearpig1996: That way if some logic is removing replica thread on demand, memory leak will not happen.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
mudit-saxena: RecoveryThread also inherits the replicaThread and these metrics doesn't make sense for VCR/recovery. Hence decided not to put it in constructor

mudit-saxena: > Can you create methods to update these values in this class?
If metric for a thread is null, you can create the metric object.

This is being done in populateReplicaMetrics currently

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -299,6 +299,7 @@ protected void removeRemoteReplicaInfoFromReplicaThread(List<RemoteReplicaInfo>
       // So do listener actions in addPartition() and removePartition().
       if (remoteReplicaInfo.getReplicaThread() != null) {
         remoteReplicaInfo.getReplicaThread().removeRemoteReplicaInfo(remoteReplicaInfo);
+        replicationMetrics.replicaThreadsAssignedRemoteReplicaInfo.get(remoteReplicaInfo.getReplicaThread().getName()).dec();
```
manbearpig1996, line 302: You can move it to inside replicaThread to keep all ReplicaThread related metric in ReplicaThread.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -333,6 +334,7 @@ protected void addRemoteReplicaInfoToReplicaThread(List<RemoteReplicaInfo> remot
       ReplicaThread replicaThread = dataNodeIdToReplicaThread.computeIfAbsent(dataNodeIdToReplicate,
           key -> replicaThreads.get(getReplicaThreadIndexToUse(datacenter)));
       replicaThread.addRemoteReplicaInfo(remoteReplicaInfo);
+      replicationMetrics.replicaThreadsAssignedRemoteReplicaInfo.get(replicaThread.getName()).inc();
```
manbearpig1996, line 337: You can move it to inside replicaThread to keep all ReplicaThread related metric in ReplicaThread.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
manbearpig1996: I am fine without this change also.

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationMetrics.java
```diff
@@ -454,6 +458,55 @@ public void populateSingleColoMetrics(String datacenter) {
...
+    Histogram replicaThreadOneCycleReplicationTime = registry.histogram(MetricRegistry.name(ReplicaThread.class,
+        replicaThread + "-OneCycleReplicationTimeMS"));
+    replicaThreadsOneCycleReplicationTime.put(replicaThread, replicaThreadOneCycleReplicationTime);
+  }
+
+  /**
+   * Increase the number of replication cycle
+   * */
+  public void updateReplicaThreadCycleIteration(String replicaThread) {
+    replicaThreadsCycleIterations.getOrDefault(replicaThread, registry.counter(MetricRegistry.name(ReplicaThread.class,
```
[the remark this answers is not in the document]
mudit-saxena: Ahh! How did i miss this. It should be `computeIfAbsent` thanks for catching this

manbearpig1996 (approved):
GTG

PR #3034, opened 2025-03-11
Added disabled replica partitions to ReplicaThread

mudit-saxena:
## Summary
Added disabled replica partitions to ReplicaThread

## Testing Done
./gradle clean build


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -177,6 +179,7 @@ public ReplicationEngine(ReplicationConfig replicationConfig, ClusterMapConfig c
 
   @Override
   public boolean controlReplicationForPartitions(Collection<PartitionId> ids, List<String> origins, boolean enable) {
+    replicationDisabledPartitions.addAll(ids);
```
manbearpig1996, line 182: Here enable flag removes the partition if it is disabled.
And origins is dcs for which we want to disable the replication.
Can you handle these cases too ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicaThread.java
```diff
@@ -902,7 +918,7 @@ ReplicaStatus getReplicaStatus(RemoteReplicaInfo remoteReplicaInfo) {
     boolean inBackoff = time.milliseconds() < remoteReplicaInfo.getReEnableReplicationTime();
     if (replicaId.isDown() || inBackoff || remoteReplicaInfo.getLocalStore().getCurrentState() == ReplicaState.OFFLINE
         || replicationDisabledPartitions.contains(replicaId.getPartitionId())) {
-      logger.debug(
+      logger.info(
```
manbearpig1996, line 921: Nice!

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -110,6 +110,7 @@ public abstract class ReplicationEngine implements ReplicationAPI {
   protected static final String replicaTokenFileName = "replicaTokens";
   protected final Time time;
   protected LeaderBasedReplicationAdmin leaderBasedReplicationAdmin = null;
+  protected Set<PartitionId> replicationDisabledPartitions;
```
manbearpig1996, line 113: Can you make it final ?

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -372,10 +375,17 @@ protected ReplicaThread getReplicaThread(String threadName, FindTokenHelper find
       boolean replicatingOverSsl, String datacenterName, ResponseHandler responseHandler, Time time,
       ReplicaSyncUpManager replicaSyncUpManager, Predicate<MessageInfo> skipPredicate,
       ReplicationManager.LeaderBasedReplicationAdmin leaderBasedReplicationAdmin) {
-    return new ReplicaThread(threadName, tokenHelper, clusterMap, correlationIdGenerator, dataNodeId, networkClient,
-        replicationConfig, replicationMetrics, notification, storeKeyConverter, transformer, metricRegistry,
-        replicatingOverSsl, datacenterName, responseHandler, time, replicaSyncUpManager, skipPredicate,
-        leaderBasedReplicationAdmin);
+    if (replicationDisabledPartitions != null && !replicationDisabledPartitions.isEmpty()) {
```
manbearpig1996, line 378: This will never be null 
You should be able to pass the empty map in constructor too. 
Can you remove this condition and use only one constuctor with controlled partitions?

mudit-saxena: Safety check doesn't hurt here. We are passing null for another constructor and someone can instantiate a replica thread with null disabledPartitions. It acts as guard for these situations

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
mudit-saxena: > And origins is dcs for which we want to disable the replication.

I have added a condition to block future replica threads only when `origins == numberOfDcs`

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationEngine.java
```diff
@@ -183,6 +185,17 @@ public boolean controlReplicationForPartitions(Collection<PartitionId> ids, List
     if (!replicaThreadPoolByDc.keySet().containsAll(origins)) {
       return false;
     }
+
+    if (origins.size() == replicaThreadPoolByDc.size()) {
```
manbearpig1996, line 189: Can you handle below cases ? 
1. This does not work if we want to submit for a specific DC.
2. During bootstrap some thread could be created early and all threads for all dcs may not be present this would not work in that scenario.

Can we do something like this ? 

`ConcurrentHashMap<String, ConcurrentHashMap<PartitionId>> dcToDisabledPartition
ConcurrentHashMap<PartitionId> disabledPartitionInallDcs` 
When origins is empty in that case you can just add this to disabledPartitionInallDcs and when dc is passed you can just add to corresponding map for dc
and you can take union of elements during thread creation.

manbearpig1996: As discussed, a small TODO will be added above if condition.

mudit-saxena: > Can you handle below cases ?
This does not work if we want to submit for a specific DC.
During bootstrap some thread could be created early and all threads for all dcs may not be present this would not work in that scenario.
Can we do something like this ?
ConcurrentHashMap<String, ConcurrentHashMap<PartitionId>> dcToDisabledPartition ConcurrentHashMap<PartitionId> disabledPartitionInallDcs
When origins is empty in that case you can just add this to disabledPartitionInallDcs and when dc is passed you can just add to corresponding map for dc
and you can take union of elements during thread creation.


Putting the discussion here for context. We are not looking at enhancing this into a full blown feature instead we are looking at the implementation which covers all DCs without affecting existing behaviour of this replica method. These changes make sure we have the added functionality and also doesn't alter any functionality. 

Now condition  about zero equality will not work as when the originDCs is empty then it will initialize with replicaThreadPoolByDC keys. In case later is also not have values, both will have zero elements

manbearpig1996 (approved):
GTG!
Lets test it in perf. Full feature can be next item for this.

PR #3037, opened 2025-03-13
Aadditional filtering for video cluster

allenaverbukh:
## Summary
Filter on the addreplica function which currently reassigns partitions to all instead of filtered ones from earlier. 

## Testing Done
UT


justinlin-linkedin (approved):
LGTM

PR #3057, opened 2025-04-07
Move id converter to router for put and ttl update handler logic

SophieGuo410:
## Summary

Move id converter to router for put and ttl update handler logic

## Testing Done

wc test

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
@@ -286,7 +286,7 @@ public class NamedBlobPutHandler {
         BlobProperties propertiesForRouterUpload = getPropertiesForRouterUpload(blobInfo);
         router.stitchBlob(restRequest, propertiesForRouterUpload, blobInfo.getUserMetadata(),
             getChunksToStitch(blobInfo.getBlobProperties(), readJsonFromChannel(channel)), null,
-            routerStitchBlobCallback(blobInfo),
+            routerStitchBlobCallback(blobInfo, propertiesForRouterUpload),
             QuotaUtils.buildQuotaChargeCallback(restRequest, quotaManager, true));
       }, uri, LOGGER, deleteDatasetCallback);
     }
@@ -297,12 +297,13 @@ public class NamedBlobPutHandler {
      * @param blobInfo                       the {@link BlobInfo} to use for security checks.
      * @return a {@link Callback} to be used with {@link Router#putBlob}.
      */
-    private Callback<String> routerStitchBlobCallback(BlobInfo blobInfo) {
-      return buildCallback(frontendMetrics.putRouterStitchBlobMetrics, blobId -> {
+    private Callback<String> routerStitchBlobCallback(BlobInfo blobInfo, BlobProperties propertiesForRouterUpload) {
+      return buildCallback(frontendMetrics.putRouterStitchBlobMetrics, convertedBlobId -> {
         // The actual blob size is now present in the instance of BlobProperties passed to the router.stitchBlob().
         // Update it in the BlobInfo so that IdConverter can add it to the named blob DB
-        restResponseChannel.setHeader(RestUtils.Headers.LOCATION, blobId);
-        String blobIdClean = stripPrefixAndExtension(blobId);
+        blobInfo.getBlobProperties().setBlobSize(propertiesForRouterUpload.getBlobSize());
+        restResponseChannel.setHeader(RestUtils.Headers.LOCATION, convertedBlobId);
+        String blobIdClean = stripPrefixAndExtension(convertedBlobId);
         if (blobInfo.getBlobProperties().getTimeToLiveInSeconds() == Utils.Infinite_Time) {
           // Do ttl update with retryExecutor. Use the blob ID returned from the router instead of the converted ID
           // since the converted ID may be changed by the ID converter.
@@ -310,7 +311,7 @@ public class NamedBlobPutHandler {
           retryExecutor.runWithRetries(retryPolicy,
               callback -> router.updateBlobTtl(restRequest, blobIdClean, serviceId, Utils.Infinite_Time, callback,
                   QuotaUtils.buildQuotaChargeCallback(restRequest, quotaManager, false)), this::isRetriable,
-              routerTtlUpdateCallbackForStitch(blobInfo, blobId));
+              routerTtlUpdateCallbackForStitch(blobInfo));
         } else {
           if (RestUtils.isDatasetVersionQueryEnabled(restRequest.getArgs())) {
             //Make sure to process response after delete finished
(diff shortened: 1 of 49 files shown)


ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -646,23 +671,75 @@ public Future<Void> updateBlobTtl(RestRequest restRequest, String blobId, String
...
+        return futureResult;
+      }
+    }
+    return futureResult;
+  }
+
+  public String stripPrefixAndExtension(String blobId) throws RestServiceException {
+    return RestUtils.stripSlashAndExtensionFromId(
+        RequestPath.parse(blobId, Collections.emptyMap(), routerConfig.pathPrefixesToRemove, routerConfig.clusterName)
+            .getOperationOrBlobId(false));
```
justinlin-linkedin, line 702: the blob id should be /xxxxx.bin if it's not named blob, in this case, we only have to do RestUtils.stripSlashAndExtentionFromId(blobId).

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -646,23 +671,75 @@ public Future<Void> updateBlobTtl(RestRequest restRequest, String blobId, String
...
+
+    //make sure the blobId does not have the cluster prefix
+    try {
+      blobId = stripPrefixAndExtension(blobId);
+    } catch (RestServiceException e) {
+      RouterException routerException =
+          new RouterException("TtlUpdateOperation failed because of blobId can't be stripped" + blobId,
+              RouterErrorCode.InvalidBlobId);
+      completeOperation(futureResult, callback, null, routerException);
+    }
```
justinlin-linkedin, line 725: if we only strip the slash and the extension, we don't need a try catch here.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -906,6 +983,7 @@ private Callback<String> createIdConverterCallbackForPut(RestRequest restRequest
         }
       } else {
         blobProperties.setBlobSize(restRequest.getBlobBytesReceived());
+        //restRequest.setArg(RestUtils.InternalKeys.BLOB_ID, blobId);
```
justinlin-linkedin, line 986: what is this?

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -466,19 +470,48 @@ public Future<String> stitchBlob(BlobProperties blobProperties, byte[] userMetad
...
     return futureResult;
   }
 
+  /**
+   * Create id converter callback after router stitch the blob.
+   * @param restRequest {@link RestRequest} to stitch the blob.
+   * @param blobProperties {@link BlobProperties} for the blob.
+   * @return
+   */
+  private Callback<String> createIdConverterCallbackForStitch(RestRequest restRequest, BlobProperties blobProperties,
```
justinlin-linkedin, line 495: can we 
1. move this method with other createIdConverterCallback methods together
2. try to reuse some code with createIdConverterCallabackForPut.

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -293,36 +293,23 @@ private Callback<Long> fetchStitchRequestBodyCallback(RetainingAsyncWritableChan
...
-
-    /**
-     * After {@link IdConverter#convert} finishes, call {@link SecurityService#postProcessRequest} to perform
-     * request time security checks that rely on the request being fully parsed and any additional arguments set.
-     * @param blobInfo the {@link BlobInfo} to use for security checks.
-     * @return a {@link Callback} to be used with {@link IdConverter#convert}.
-     */
-    private Callback<String> idConverterCallback(BlobInfo blobInfo, String blobId) {
-      return buildCallback(frontendMetrics.putIdConversionMetrics, convertedBlobId -> {
+        blobInfo.getBlobProperties().setBlobSize(propertiesForRouterUpload.getBlobSize());
```
justinlin-linkedin, line 302: This is too late, the blob properties().setBlobSize should be called before idconverter, idconverter has to get the blob size and write it to named blob database, that's why we need the blob size. we have to do this before the idconverter.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -466,19 +470,48 @@ public Future<String> stitchBlob(BlobProperties blobProperties, byte[] userMetad
     }
     routerMetrics.operationQueuingRate.mark();
     FutureResult<String> futureResult = new FutureResult<>();
+    Callback<String> wrappedCallback =
+        restRequest != null && idConverter != null ? createIdConverterCallbackForStitch(restRequest, blobProperties,
+            futureResult, callback) : callback;
```
justinlin-linkedin, line 475: there is one little thing about this callback and the future. It's possible that your callback would return exception but your future would return success. Let me explain. In getOperationController().stitchBlob method you passed in wrappedCallback and the futureResult, which means that when stitch blob succeeds, without the named blob operation, the futureResult will be set to success first (checkout completeOperation in this file), which means it will already succeeds, then the wrappedcallback is called. If the named blob operation failed, then the callback would return exception. In this case, the future and the callback returns different things.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
SophieGuo410: You are right, fixed it.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
SophieGuo410: Removed.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
SophieGuo410: Adding a TODO.

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -646,23 +655,61 @@ public Future<Void> updateBlobTtl(RestRequest restRequest, String blobId, String
...
+      proceedWithTtlUpdate(blobId, serviceId, expiresAtMs, wrappedCallback, futureResult, quotaChargeCallback);
+    } else {
+      try {
+        idConverter.convert(restRequest, blobId, null, new Callback<String>() {
+          @Override
+          public void onCompletion(String convertedBlobId, Exception exception) {
+            if (exception != null) {
+              wrappedCallback.onCompletion(null, exception);
+            } else {
+              proceedWithTtlUpdate(convertedBlobId, serviceId, expiresAtMs, wrappedCallback, futureResult,
```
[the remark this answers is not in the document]
SophieGuo410: Discussed offline and fixed it.

PR #3061, opened 2025-04-10
Remove all versions in mysql database for delete operation

justinlin-linkedin:
## Summary
Fix an issue with blob name delete. When deleting a named blob, we only delete the max version whose blob state is ready in the database table. And leave the other version as "undeleted". There is an issue with this approach. The background job that removes the rows whose "deleted_ts" is not null, would end up completely delete the max version row. After that, the lower versions would become the new max version as if this blob name is not deleted yet.

This PR fixed it by removing all versions.

### MysqlNamedBlobDb
1. Introduce a configuration to enable hard delete for delete method. In the current implementation, we don't actually delete any rows from the table in delete method, rather just update the "deleted_ts" for each deleted row. This new configuration would trigger actual delete of those rows from the table. It doesn't change cleanup method yet, another PR should raised to enable hard delete in cleanup job.
2. Update delete method in MysqlNamedBlob to return all the versions and then delete all the undeleted versions. DeleteResult now returns all the blob versions and blob ids from the table. Unfortunately, delete method is tightly coupled with AmbryIdConverter so that there is no good interface to return all the blob ids but to concatenate them with comma. This works, but a better interface is more desirable. 
3. Update integration test for MysqlNamedBlobDb as well

### InMemNamedBlobDb
InMemNamedBlobDb should have the same logic as the real MysqlNamedBlobDb, however, the current implementation is far from that. This PR also update most of the code in the InMemNamedBlobDb to align with the MysqlNamedBlobDb. 
There is no dedicated unit test created for InMemNamedBlobDb, but we have tests in FrontendRestRequestServiceTest that uses this implementation. This is fine for now, but a dedicated unit test class is more desirable.

### NamedBlobHandlers
Remove two unused NamedBlobHandlers

### NonBlockingRouter
1. Update the deleteBlob in NonBlockingRouter and InMemRouter. InMemRouter is already very close to the actual NonBlockingRouter so there is not much change to InMemRouter.
2. Update BatchOperationCallbackTracker to 
2.1 Ignore final blob id
2.2 Apply exception mapper for exceptions in callbck
3. Add unit tests

## Testing Done
Add uint tests


Arun-LinkedIn (approved):
lgtm!

PR #3064, opened 2025-04-11
Added HelixClusterManager support instead of CompositeClusterManager

mudit-saxena:
## Summary
Added HelixClusterManager support instead of CompositeClusterManager

## Testing Done
./gradlew clean build


Chethan2k1 (approved):
LGTM.

PR #3067, opened 2025-04-11
[MINOR] Reset to normal replication with disabled partitions

mudit-saxena:
## Summary
- [MINOR] Reset to normal replication with disabled partitions and no new bootstrapping partitions
- Added test to cover this case

## Testing Done
```./gradlew clean build```


Chethan2k1 (approved):
LGTM.

PR #3075, opened 2025-04-15
Refactoring Log and Compactor for Disk Space Allocator usage in FC

Chethan2k1:
## Summary
-  For File Copy we plan to break `addBlobStore()` in Storage Manager in two parts `initializeBlobStore()` and `loadBlobStore()`
-  File Copy would be done in between these, and since atm `Log` and `BlobStoreCompactor` are constructed in `loadInitializedBlobStore` of DiskManager which would be part of post file copy, and these are needed by the DiskSpaceAllocator which would be called in pre file copy, refactoring these files to fit in the new flow.
- We plan to call DiskSpaceAllocator's `addRequiredSegments` in `initializeBlobStore`, after that file copy would happen which would use Disk Space Allocator to allocate the files that are being copied and upon calling `loadInitializedBlobStore`,  `Log` and `BlobStoreCompactor` are inited

## Testing Done
`./gradlew clean build`

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -9,7 +9,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: [ '**' ]
 
 # A workflow run is made up of one or more jobs that can run sequentially or in parallel
 jobs:
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -1674,9 +1674,9 @@ public class BlobStore implements Store {
    */
   DiskSpaceRequirements getDiskSpaceRequirements() throws StoreException {
     checkInitialized();
-    DiskSpaceRequirements requirements = new DiskSpaceRequirements(replicaId.getPartitionId().toPathString(),
-        Math.min(log.getCapacityInBytes(), config.storeSegmentSizeInBytes), log.getRemainingUnallocatedSegments(),
-        compactor.getSwapSegmentsInUse().length);
+    DiskSpaceRequirements requirements =
+        new DiskSpaceRequirements(replicaId.getPartitionId().toPathString(), log.getSegmentSize(),
+            log.getRemainingUnallocatedSegments(), compactor.getSwapSegmentsInUse().length);
     logger.info("Store {} has disk space requirements: {}", storeId, requirements);
     return requirements;
   }
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -294,6 +294,11 @@ public class BlobStore implements Store {
 
         storeDescriptor = new StoreDescriptor(dataDir, config);
         fileStore.start();
+        log = new Log(dataDir, capacityInBytes, diskSpaceAllocator, config, metrics, diskMetrics);
+        compactor = new BlobStoreCompactor(dataDir, storeId, factory, config, metrics, storeUnderCompactionMetrics,
+            diskIOScheduler, diskSpaceAllocator, log, time, sessionId, storeDescriptor.getIncarnationId(),
+            accountService, remoteTokenTracker, diskMetrics);
+
         initialized = true;
       } catch (Exception e) {
         if (fileLock != null) {
@@ -320,10 +325,8 @@ public class BlobStore implements Store {
         throw new StoreException("Store not initialized", StoreErrorCodes.StoreNotInitialized);
       }
       try {
-        log = new Log(dataDir, capacityInBytes, diskSpaceAllocator, config, metrics, diskMetrics);
-        compactor = new BlobStoreCompactor(dataDir, storeId, factory, config, metrics, storeUnderCompactionMetrics,
-            diskIOScheduler, diskSpaceAllocator, log, time, sessionId, storeDescriptor.getIncarnationId(),
-            accountService, remoteTokenTracker, diskMetrics);
+        log.init();
+        compactor.init();
         index = new PersistentIndex(dataDir, storeId, indexPersistScheduler, log, config, factory, recovery, hardDelete,
             diskIOScheduler, metrics, time, sessionId, storeDescriptor.getIncarnationId());
         compactor.initialize(index);
--- a/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
@@ -294,10 +294,10 @@ public class BlobStore implements Store {
 
         storeDescriptor = new StoreDescriptor(dataDir, config);
         fileStore.start();
-        log = new Log(dataDir, capacityInBytes, diskSpaceAllocator, config, metrics, diskMetrics);
+        log = new Log(dataDir, capacityInBytes, diskSpaceAllocator, config, metrics, diskMetrics, false);
         compactor = new BlobStoreCompactor(dataDir, storeId, factory, config, metrics, storeUnderCompactionMetrics,
             diskIOScheduler, diskSpaceAllocator, log, time, sessionId, storeDescriptor.getIncarnationId(),
-            accountService, remoteTokenTracker, diskMetrics);
+            accountService, remoteTokenTracker, diskMetrics, false);
 
         initialized = true;
       } catch (Exception e) {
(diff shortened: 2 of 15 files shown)


ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -157,10 +157,19 @@ public boolean accept(File dir, String name) {
...
-    fixStateIfRequired();
     logger.trace("Constructed BlobStoreCompactor for {}", dataDir);
   }
 
+  /**
+   * Initializes the compactor. This method is called when the {@link BlobStore} is loaded.
+   * @throws IOException if the {@link CompactionLog} could not be created or if commit/cleanup failed during recovery.
+   * @throws StoreException if the commit failed during recovery.
+   */
+  void init() throws IOException, StoreException {
```
manbearpig1996, line 168: Can we create a new Constructor with option of init true or false.
From the current constructor, we can just pass this option as true and use the new Constructor in BlobStore.

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -79,7 +79,54 @@ class Log implements Write {
     this.diskMetrics = diskMetrics;
     this.segmentNameAndFileNameIterator = Collections.emptyIterator();
     storeId = dataDir.substring(dataDir.lastIndexOf(File.separator) + File.separator.length());
+    // retrieve log segments from the directory
+    File dir = new File(dataDir);
```
manbearpig1996, line 83: Can we create a new Constructor with option of init true or false.
From the current constructor, we can just pass this option as true and use the new Constructor in BlobStore.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
manbearpig1996: This way, we do not have to make unnecessary changes in multiple places.

Chethan2k1: Added another constructor for this.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1671,11 +1674,11 @@ public void setDisableState(boolean disabled) {
    * @throws StoreException
    */
   DiskSpaceRequirements getDiskSpaceRequirements() throws StoreException {
-    checkStarted();
+    checkInitialized();
     DiskSpaceRequirements requirements =
         log.isLogSegmented() ? new DiskSpaceRequirements(replicaId.getPartitionId().toPathString(),
-            log.getSegmentCapacity(), log.getRemainingUnallocatedSegments(), compactor.getSwapSegmentsInUse().length)
-            : null;
+            Math.min(log.getCapacityInBytes(), config.storeSegmentSizeInBytes), log.getRemainingUnallocatedSegments(),
```
manbearpig1996, line 1680: Shoudn't it be Math.max(log.getCapacityInBytes(), config.storeSegmentSizeInBytes)
Can we move this inside Log class ?

Can you verify if now we need to allocate space for unsegmented log here too as earlier that would have been created when Log object was created.

ambry-store/src/main/java/com/github/ambry/store/BlobStoreCompactor.java
```diff
@@ -157,8 +186,20 @@ public boolean accept(File dir, String name) {
     } else {
       validEntryFilter = new IndexSegmentValidEntryFilterWithUndelete();
     }
+    if (shouldInit) {
+      init();
+      logger.trace("Constructed BlobStoreCompactor for {}", dataDir);
```
manbearpig1996, line 191: Can we move this trace log outside of this check ?

.github/workflows/github-actions.yml
```diff
@@ -9,7 +9,7 @@ on:
   push:
     branches: [ master ]
   pull_request:
-    branches: [ master ]
+    branches: [ '**' ]
```
manbearpig1996, line 12: Is this required ?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
manbearpig1996: Checked again. So this is segmentsize so it should be Math.min.

Chethan2k1: Yes.

.github/workflows/github-actions.yml
Chethan2k1: Yes, builds are not triggering for if branch target is not master, it's good to have.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1666,16 +1669,14 @@ public void setDisableState(boolean disabled) {
...
    */
   DiskSpaceRequirements getDiskSpaceRequirements() throws StoreException {
-    checkStarted();
-    DiskSpaceRequirements requirements =
-        log.isLogSegmented() ? new DiskSpaceRequirements(replicaId.getPartitionId().toPathString(),
-            log.getSegmentCapacity(), log.getRemainingUnallocatedSegments(), compactor.getSwapSegmentsInUse().length)
-            : null;
+    checkInitialized();
+    DiskSpaceRequirements requirements = new DiskSpaceRequirements(replicaId.getPartitionId().toPathString(),
+        Math.min(log.getCapacityInBytes(), config.storeSegmentSizeInBytes), log.getRemainingUnallocatedSegments(),
```
manbearpig1996, line 1678: Can we move Math.min calculation inside log class?

manbearpig1996 (commented):
Can you also add detailed PR description?

manbearpig1996 (approved):
GTG

justinlin-linkedin (commented):
Just a few comments, the rest looks good to me.

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -366,11 +397,67 @@ private LogSegment checkArgsAndGetFirstSegment(long segmentCapacity, boolean nee
...
+    // otherwise, populate remainingUnallocatedSegments
+    if (segmentFiles == null) {
+      throw new StoreException("Could not read from directory: " + dataDir, StoreErrorCodes.FileNotFound);
+    } else {
+      long totalSegments;
+      // if the files could not be read, throw an exception
+      // otherwise, populate remainingUnallocatedSegments
+      if (segmentFiles == null) {
+        throw new StoreException("Could not read semgent files from Directory", StoreErrorCodes.FileNotFound);
+      } else {
```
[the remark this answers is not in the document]
Chethan2k1: Removed.

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -366,11 +397,67 @@ private LogSegment checkArgsAndGetFirstSegment(long segmentCapacity, boolean nee
...
+   * @param dataDir the directory where the segments of the log need to be loaded from.
+   * @param segmentSize the intended size of each log segment.
+   * @param totalCapacityInBytes the total capacity of the log in bytes.
+   * @param isLogSegmented a flag indicating whether the log is segmented.
+   * @return the number of unallocated segments.
+   * @throws StoreException if the segment files cannot be read from the directory.
+   * @throws IllegalArgumentException if the total capacity or segment size is invalid, or if the total capacity
+   *         is not a multiple of the segment size.
    */
+  private long checkArgsAndGetUnallocatedSegments(String dataDir, long segmentSize, long totalCapacityInBytes,
```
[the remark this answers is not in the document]
Chethan2k1: Using the fields, removed the arguments.

ambry-store/src/main/java/com/github/ambry/store/Log.java
```diff
@@ -366,11 +397,67 @@ private LogSegment checkArgsAndGetFirstSegment(long segmentCapacity, boolean nee
...
+                    + segmentCapacity + "] is <=0");
+          }
+
+          // all segments should be the same size.
+          if (totalCapacityInBytes % segmentCapacity != 0) {
+            throw new IllegalArgumentException(
+                "Capacity of log [" + segmentCapacity + "] should be a multiple of segment capacity [" + segmentCapacity
+                    + "]");
+          }
+        }
```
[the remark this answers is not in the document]
Chethan2k1: Can't move the second logic above since this is applicable only if no segments are present. Without the check of `segmentsFiles.length == 0`, this would fail test cases where we have an existing unsegmented log with size greater than segmentCapacity.

justinlin-linkedin (approved):

PR #3077, opened 2025-04-17
Bootstrap Performance Improvement[PR7] : Fixing Bugs For File Copy Module.

aga9900:
## Summary
Bugs
1. PZ Manager not removing a replica after providing it to the Scheduler causing multi-hydration of the same partition.
2. Order of Scheduler and Prioritizer start should be reversed because Scheduler will need Prioritizer running before it starts processing.
3. Create Thread For Scheduler instead of the main thread class, so that it does not block the whole process.
4. Multiple instantiations of Prioritizer are created violating consistency between Scheduler and Manager. Use single instantiation and share among classes.

## Testing Done
./gradlew clean  allJar


manbearpig1996 (approved):
GTG

PR #3081, opened 2025-04-22
Return 204 in s3 delete handler for 404

justinlin-linkedin:
## Summary
S3 delete operation should be idempotent, which means deleting a non-existent key should be treated as a success. Non-existent keys include deleted keys as well. This PR guarantees this behavior.

## Details
In this PR https://github.com/linkedin/ambry/pull/3061, we introduced a "HardDelete" configuration for MysqlNamedBlobDb. If HardDelete is enabled, we run a DELETE sql command to remove the a given blob name from named blob table when deleting this blob name. This creates a unexpected result where a delete operation on an already deleted blob, would return 404, not 204. However s3 delete operation is idempotent, we should return 204 in this case.

And not only that, a delete request for non-existent key in s3 should return no error response, but a success. This is the sdk javadoc for delete object https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/AmazonS3Client.html#deleteObject-com.amazonaws.services.s3.model.DeleteObjectRequest-

## Testing Done
Unit test


Arun-LinkedIn (approved):
lgtm!

PR #3082, opened 2025-04-23
Issue manual disk flush for each chunk in FCH

amanatgithub:
## Summary
This change ensures that each data chunk written is flushed to disk immediately by using the SYNC option (force flush to disk). This guarantees that both file content and metadata are durably persisted, minimizing the risk of data loss in the event of a sudden node restart or crash.

## Testing Done
./gradlew clean build


aga9900 (approved):
LGTM

PR #3085, opened 2025-04-29
Netty ByteBuf memory leak fix in FCH

amanatgithub:
## Summary
Netty ByteBuf memory leak fix in FCH

## Testing Done
./gradlew clean build

### DIFF
--- a/ambry-file-transfer/src/main/java/com/github/ambry/filetransfer/handler/StoreFileCopyHandler.java
+++ b/ambry-file-transfer/src/main/java/com/github/ambry/filetransfer/handler/StoreFileCopyHandler.java
@@ -34,6 +34,7 @@ import com.github.ambry.store.PartitionFileStore;
 import com.github.ambry.store.StoreException;
 import com.github.ambry.store.StoreFileChunk;
 import com.github.ambry.store.StoreFileInfo;
+import com.github.ambry.utils.NettyByteBufDataInputStream;
 import java.io.File;
 import java.io.IOException;
 import java.util.Objects;
@@ -272,6 +273,10 @@ public class StoreFileCopyHandler implements FileCopyHandler {
 
     String filePath = partitionToMountFilePath + File.separator + indexFile.getFileName();
     writeStoreFileChunkToDisk(chunkResponse, filePath, fileStore);
+    if (chunkResponse != null && chunkResponse.getChunkStream() instanceof NettyByteBufDataInputStream) {
+      // if the DataInputStream is NettyByteBufDataInputStream based, it's time to release its buffer.
+      ((NettyByteBufDataInputStream) (chunkResponse.getChunkStream())).getBuffer().release();
+    }
   }
 
   /**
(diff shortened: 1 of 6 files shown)


ambry-file-transfer/src/main/java/com/github/ambry/filetransfer/handler/StoreFileCopyHandler.java
```diff
@@ -301,6 +306,10 @@ private void processLogSegment(LogInfo logInfo, String partitionToMountFilePath,
           fileChunkInfo, true);
       String filePath = partitionToMountFilePath + File.separator + logFileInfo.getFileName();
       writeStoreFileChunkToDisk(chunkResponse, filePath, fileStore);
+      if (chunkResponse != null && chunkResponse.getChunkStream() instanceof NettyByteBufDataInputStream) {
+        // if the DataInputStream is NettyByteBufDataInputStream based, it's time to release its buffer.
+        ((NettyByteBufDataInputStream) (chunkResponse.getChunkStream())).getBuffer().release();
```
aga9900, line 311: Can we put a release method in the GetChunkResponse class itself and then explicitly call that method from FCH? Instead of assigning it be to a subclass NettyByteBufDataInputStream explicitly. So, that if something changes on the network side, this code does not have to be changed explicitly.

amanatgithub: This was the initial thought to understand the placeholder for the change, I have added the comment in next patch.

DevenAhluwalia (approved):

ambry-file-transfer/src/main/java/com/github/ambry/filetransfer/handler/StoreFileCopyHandler.java
```diff
@@ -272,6 +273,7 @@ private void processIndexFile(FileInfo indexFile, String partitionToMountFilePat
 
     String filePath = partitionToMountFilePath + File.separator + indexFile.getFileName();
     writeStoreFileChunkToDisk(chunkResponse, filePath, fileStore);
+    chunkResponse.release();
```
aga9900, line 276: Should we add en explicit release method in FileCopyGetMetaDataResponse.java as well? So, that the stream is released on the Bootstrapping node. Current method does not seem to be releasing the InputStream.

aga9900 (approved):
LGTM

ambry-file-transfer/src/main/java/com/github/ambry/filetransfer/handler/StoreFileCopyHandler.java
```diff
@@ -163,21 +164,25 @@ public void copy(@Nonnull FileCopyInfo fileCopyInfo) throws Exception {
         fileCopyInfo.getSourceReplicaId().getPartitionId().getId();
     final FileCopyGetMetaDataResponse metadataResponse = getFileCopyGetMetaDataResponse(fileCopyInfo);
```
aga9900, line 166: we can release here as well ans should not wait till end.

ambry-protocol/src/test/java/com/github/ambry/protocol/RequestResponseTest.java
```diff
@@ -857,7 +857,7 @@ public void doFileCopyMetaDataResponseTest() throws IOException{
     response.release();
```
aga9900, line 857: You can keep asserts here to check the contents on the response class are intact after releasing the buffer.

PR #3098, opened 2025-05-16
Added handling for partitions which no longer exists

mudit-saxena:
## Summary

Added handling for partitions which no longer exists

## Testing Done
./grdlew clean test


Chethan2k1 (approved):
LGTM.

PR #3099, opened 2025-05-16
[Filecopy][Deferred Compaction] Api plane integration with deferred compaction

DevenAhluwalia:
## Summary
Api plane integration with deferred compaction

## Testing Done
NA


ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1768,6 +1771,29 @@ CompactionDetails getCompactionDetailsInProgress() throws StoreException {
     return null;
   }
 
+  public boolean isCompactionInProgress() {
```
manbearpig1996, line 1774: Looks like this method is unused.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -1695,39 +1698,58 @@ public void handleFileCopyGetMetaDataRequest(NetworkRequest request) throws Inte
...
+        BootstrapSession existingSession = bootstrapSessionManager.getBootstrapSession(partitionId, fileCopyGetMetaDataRequest.getHostName());
+
+        // if BootstrapSession exists for (requested partition, requesting Node) then delete existing session
+        if (existingSession != null) {
+          logger.info("FileCopyGetMetaDataRequest: Session already exists for partition {}. Deleting existing session", partitionId.getId());
+          bootstrapSessionManager.removeBootstrapSession(partitionId, fileCopyGetMetaDataRequest.getHostName());
+        }
+        // Check if compaction is in progress.
+        //  If yes, then reject the request with error code = SnapshotUnavailable.
+        if (blobStore.isCompactionInProgress()) {
```
manbearpig1996, line 1727: Won't this file always be present even if control compaction has been called ?

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -428,11 +430,45 @@ boolean scheduleNextForCompaction(PartitionId id) {
...
+    controlCompactionForBlobStoreMap.put(id, enabled);
+    return controlCompactionForBlobStoreStub(id, enabled);
+  }
+
+  /**
+   * Return true if the compaction control has been set. Return the compaction control for the given partition id.
+   * @param id the {@link PartitionId} of the {@link BlobStore} to check control for.
+   * @return {@code true} if the compaction is under control. {@code false} if not.
+   */
+  public boolean isCompactionEnabledForBlobStoreUnderControl(PartitionId id) {
```
manbearpig1996, line 450: Can we rename to isCompactionEnabledForBlobStore ?

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
```diff
@@ -428,11 +430,45 @@ boolean scheduleNextForCompaction(PartitionId id) {
...
   /**
    * Enable or disable compaction on the {@link PartitionId} {@code id}.
+   * Also updates controlCompactionForBlobStoreMap which is further used by deferred compaction to prevent overriding
+   * this control.
+   * This is used by few places like State transitions, Admin store start/stop endpoints, Repairs flow etc.
+   * @param id the {@link PartitionId} of the {@link BlobStore} on which compaction control is enabled or disabled.
+   * @param enabled whether to enable ({@code true}) or disable.
+   * @return {@code true} if enabling or disabling was successful. {@code false} if not.
+   */
+  public boolean controlCompactionForBlobStore(PartitionId id, boolean enabled) {
```
manbearpig1996, line 440: Do you want to take lock here ? As it will be accessed by multiple threads.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -1762,43 +1783,80 @@ public void handleFileCopyGetChunkRequest(NetworkRequest request) throws Interru
...
+          // if BootstrapSession exists for (requested partition, requesting Node) then delete existing session
+          if (existingSession != null) {
+            logger.info("FileCopyGetChunkRequest: Session already exists for partition {}. Deleting existing session", partitionId.getId());
+            bootstrapSessionManager.removeBootstrapSession(partitionId, fileCopyGetChunkRequest.getHostName());
+          }
+          logger.info("FileCopyGetChunkRequest: Compaction is in progress for partition {}. Rejecting request", partitionId.getId());
+          response = new FileCopyGetChunkResponse(ServerErrorCode.SnapshotUnavailable);
+        } else  {
+          if(storeManager.isCompactionControlBeenSetForBlobStore(partitionId) && storeManager.isCompactionEnabledForBlobStoreUnderControl(partitionId)) {
```
manbearpig1996, line 1820: Where are we stopping compaction?

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
```diff
@@ -1768,6 +1771,29 @@ CompactionDetails getCompactionDetailsInProgress() throws StoreException {
...
+    return CompactionLog.isCompactionInProgress(dataDir, storeId);
+  }
+
+  @Override
+  public String getSnapshotId(List<LogInfo> logSegments) {
```
manbearpig1996, line 1779: So snapshot Id will change when any active segment becomes sealed. 
Could this cause any issue?

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -1762,43 +1783,80 @@ public void handleFileCopyGetChunkRequest(NetworkRequest request) throws Interru
...
-            fileCopyGetChunkRequest.getPartitionId(), fileCopyGetChunkRequest.getFileName(), chunkResponse.getStream(),
-            fileCopyGetChunkRequest.getStartOffset(), chunkResponse.getChunkLength(), false);
+        Store blobStore = storeManager.getStore(partitionId);
+        if (null == blobStore) {
+          logger.error("BlobStore is not available for partition {}", partitionId.getId());
+          throw new StoreException("BlobStore is not available for partition " + partitionId, StoreErrorCodes.StoreNotStarted);
+        }
+        // Check if compaction is in progress.
+        //  If yes, then reject the request with error code = SnapshotUnavailable.
+        if (blobStore.isCompactionInProgress()) {
```
manbearpig1996, line 1811: Won't this be true even if control compaction is called and compaction is not running ?

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
```diff
@@ -1695,39 +1698,58 @@ public void handleFileCopyGetMetaDataRequest(NetworkRequest request) throws Inte
...
+          logger.info("FileCopyGetMetaDataRequest: Session already exists for partition {}. Deleting existing session", partitionId.getId());
+          bootstrapSessionManager.removeBootstrapSession(partitionId, fileCopyGetMetaDataRequest.getHostName());
+        }
+        // Check if compaction is in progress.
+        //  If yes, then reject the request with error code = SnapshotUnavailable.
+        if (blobStore.isCompactionInProgress()) {
+          logger.info("FileCopyGetMetaDataRequest: Compaction is in progress for partition {}. Rejecting request", partitionId.getId());
+          response = new FileCopyGetMetaDataResponse(ServerErrorCode.SnapshotUnavailable);
+        } else {
+          if(storeManager.isCompactionControlBeenSetForBlobStore(partitionId) && storeManager.isCompactionEnabledForBlobStoreUnderControl(partitionId)) {
```
manbearpig1996, line 1731: Shouldn't checking and enabling /disabling compaction be atomic operation ?
As all calls will be from different threads.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
DevenAhluwalia: This file's existence tells that Compaction had started for this store and its not finished yet. 
Its looping on log segments, one at a time (as part of split-cycles logic). 

Thats why, before creating a new BootstrapSession and disabling Compaction, this check ensures that we reject the request early on.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
DevenAhluwalia: Ya. Current name is bit messed up. Incorporated.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
DevenAhluwalia: This is GetChunk Api. Look for `existingSession.extendDeferralTimer();` further down in this Api. 
We expect session to have been created when this Api is called and we just extend the timer.

ambry-store/src/main/java/com/github/ambry/store/BlobStore.java
DevenAhluwalia: Snapshot is created only in the GetMetadata Api and a BootstrapSession is created which contains SnapshotId info. 
Then, GetChunk Api calls dont re-calculate snapshotId but it just matches the snapshotId in the request with whats saved in existing BootstrapSession. 

So, even if an Active log segment becomes sealed, it shouldn't be a problem.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
DevenAhluwalia: Yep. File's existence is the SoT for Compaction running or not. 
ControlCompaction just lists/delists a new Store in the Set of stores sequentially being processed by Compaction.

If a Store is already under Compaction, and ControlCompaction is called, it doesn't stop the running Compaction on that store. And accordingly we still want to reject bootstrap requests. 

This check will return False only when the Compaction has run its course for that store.

ambry-store/src/main/java/com/github/ambry/store/DiskManager.java
DevenAhluwalia: This method does 2 things -
1. Update `controlCompactionForBlobStoreMap`
2. Delegate the call to `controlCompactionForBlobStoreStub`

Have updated the `controlCompactionForBlobStoreMap` to a ConcurrentHashMap so `.put` is thread-safe
`controlCompactionForBlobStoreStub` is anyway taking a readLock

So, I think there is no benefit in surrounding these 2 statements under a lock.

ambry-server/src/main/java/com/github/ambry/server/AmbryRequests.java
DevenAhluwalia: Refactored these 2 getters into a single `isCompactionControlBeenSetAndIsEnabledForBlobStore` and again, as its a ConcurrentHashMap so `.put` is thread-safe.

aga9900 (approved):
LGTM

manbearpig1996 (approved):

PR #3101, opened 2025-05-19
Add a cli for server load test

justinlin-linkedin:
## Summary
Create a tool to generate synthetic load to a single ambry-server instance for testing. It can generate all get, put, delete, update ttl traffic at the desired qps, either at the same time, or separately. It doesn't provide any client side metric so all the performance result has to be observed from server side metric.

## Testing Done
Use this tool to test a single server.

### DIFF
--- a/ambry-tools/src/main/java/com/github/ambry/tools/perf/ConcurrentServerLoadTest.java
+++ b/ambry-tools/src/main/java/com/github/ambry/tools/perf/ConcurrentServerLoadTest.java
@@ -1,0 +1,936 @@
+/**
+ * Copyright 2025 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
+package com.github.ambry.tools.perf;
+
+import com.codahale.metrics.MetricRegistry;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.github.ambry.account.Account;
+import com.github.ambry.account.Container;
+import com.github.ambry.clustermap.ClusterAgentsFactory;
+import com.github.ambry.clustermap.ClusterMap;
+import com.github.ambry.clustermap.DataNodeId;
+import com.github.ambry.clustermap.ReplicaId;
+import com.github.ambry.commons.BlobId;
+import com.github.ambry.commons.NettySslHttp2Factory;
+import com.github.ambry.commons.SSLFactory;
+import com.github.ambry.config.ClusterMapConfig;
+import com.github.ambry.config.Config;
+import com.github.ambry.config.Http2ClientConfig;
+import com.github.ambry.config.SSLConfig;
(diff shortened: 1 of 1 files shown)


djycheng:
What's the difference between this test and the one we wrote with forge fire? Can we run the forgefire test on a one-node cluster if we wanted to verify performance of a single server?

ambry-tools/src/main/java/com/github/ambry/tools/perf/ConcurrentServerLoadTest.java
```diff
@@ -0,0 +1,969 @@
...
+      validatePutBlobSize();
+    }
+
+    void validatePutBlobSize() {
+      ObjectMapper objectMapper = new ObjectMapper();
+      if (putBlobSizeDistribution != null && !putBlobSizeDistribution.isEmpty()) {
+        try {
+          putBlobSizeDistributionMap.putAll(objectMapper.readValue(putBlobSizeDistribution, Map.class));
+          int previousSize = 0;
+          for (String key : Arrays.asList("0.5", "0.75", "0.95", "0.999")) {
```
alyssaxu333, line 304: Is it possible that the distribution is not 0.5/0.75/0.95/0.999

alyssaxu333, line 304: Is it possible that the distribution input is 0.3/0.6/0.9... Is this a valid input?

ambry-tools/src/main/java/com/github/ambry/tools/perf/ConcurrentServerLoadTest.java
```diff
@@ -0,0 +1,969 @@
...
+  private final List<ReplicaId> replicas = new ArrayList<>();
+
+  public ConcurrentServerLoadTest(VerifiableProperties verifiableProperties) throws Exception {
+    config = new LoadTestConfig(verifiableProperties);
+    config.validate();
+    blobData = TestUtils.getRandomBytes(getBlobDataSize());
+
+    SSLFactory sslFactory = new NettySslHttp2Factory(new SSLConfig(verifiableProperties));
+    Http2ClientConfig http2ClientConfig = new Http2ClientConfig(verifiableProperties);
+    connectionPool =
```
alyssaxu333, line 351: What is connectionPool used for?

ambry-tools/src/main/java/com/github/ambry/tools/perf/ConcurrentServerLoadTest.java
```diff
@@ -0,0 +1,969 @@
...
+        logger.info("Replica {} is unavailable, removing it from the list of writable replicas", replicaId);
+      }
+    } finally {
+      rwLock.writeLock().unlock();
+    }
+  }
+
+  private void start() {
+    connectionPool.start();
+    // Assume each operation is going to take 20 milliseconds to finish on average, which means one thread can execute
```
alyssaxu333, line 440: Curious why 20ms? How to determine this

ambry-tools/src/main/java/com/github/ambry/tools/perf/ConcurrentServerLoadTest.java
```diff
@@ -0,0 +1,969 @@
...
+      // The last thread should take care of the remaining operations
+      LoadTestThread testThread =
+          new LoadTestThread("LoadTestThread-" + i, putLoads[i], getLoads[i], deleteLoads[i], updateTtlLoads[i]);
+      testThread.start();
+      loadTestThreads.add(testThread);
+    }
+
+    for (int i = 0; i < numThreads; i++) {
+      try {
+        loadTestThreads.get(i).join();
```
alyssaxu333, line 460: Do we need to set timeout for single thread?

PR #3102, opened 2025-05-20
Create throughput metric for each request.

djycheng:
## Summary
This metric will show the blob upload/download performance for a singular request. The current metrics do not show this type of data as they only count the total number of bytes read/written

## Testing Done
./gradlew build

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/frontend/EntityOperationMetrics.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/EntityOperationMetrics.java
@@ -83,7 +83,7 @@ public class EntityOperationMetrics {
     String qpsMetricPrefix = entityName + SEPARATOR + (isGetRequest ? "GetRequest" : "PutRequest");
     totalCount = metricRegistry.counter(MetricRegistry.name(ownerClass, qpsMetricPrefix + "totalCount"));
 
-    throughput = metricRegistry.histogram(MetricRegistry.name(ownerClass, qpsMetricPrefix + "Throughput"));
+    throughput = metricRegistry.histogram(MetricRegistry.name(ownerClass, "Throughput"));
   }
 
   /**
--- a/ambry-api/src/main/java/com/github/ambry/frontend/EntityOperationMetrics.java
+++ b/ambry-api/src/main/java/com/github/ambry/frontend/EntityOperationMetrics.java
@@ -16,6 +16,7 @@ import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import com.github.ambry.rest.ResponseStatus;
+import com.github.ambry.rest.RestUtils;
 
 
 /**
@@ -50,6 +51,8 @@ public class EntityOperationMetrics {
 
   protected final Counter totalCount;
 
+  protected final Histogram throughput;
+
   /**
    * Constructor to create operation metrics for given entity. The entity can be an account or a container.
    *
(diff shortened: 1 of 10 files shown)


ambry-api/src/main/java/com/github/ambry/frontend/EntityOperationMetrics.java
```diff
@@ -87,11 +92,12 @@ public class EntityOperationMetrics {
    * @param roundTripTimeInMs the time it took to receive a request and send a response.
    * @param responseStatus    the {@link ResponseStatus} sent in the response.
    */
-  public void recordMetrics(long roundTripTimeInMs, ResponseStatus responseStatus) {
+  public void recordMetrics(long roundTripTimeInMs, ResponseStatus responseStatus, long bytesTransferred) {
     this.roundTripTimeInMs.update(roundTripTimeInMs);
     totalCount.inc();
     if (responseStatus.isSuccess()) {
       successCount.inc();
+      throughput.update(RestUtils.calculateThroughput(bytesTransferred, roundTripTimeInMs));
```
justinlin-linkedin, line 100: we probably don't want to record the throughput metric if the byteTransferred is 0. This would skew the throughput. If there are many delete and updatettl requests, we would have lots of bandwidth datapoints being 0.

ambry-frontend/src/main/java/com/github/ambry/frontend/GetBlobHandler.java
```diff
@@ -243,6 +243,9 @@ private Callback<Void> securityProcessResponseCallback(GetBlobResult result) {
             ReadableStreamChannel response = result.getBlobDataChannel();
             if (subResource == null) {
               response = closeChannelIfNotModified(result);
+              if (response != null) {
+                restRequest.getMetricsTracker().setBytesTransferred(result.getBlobInfo().getBlobProperties().getBlobSize());
```
justinlin-linkedin, line 247: BlobSize in blob properties might not be the right bytesTransferred when it's range request. Client can specify a range to download when issuing a get request, in this way, blob size in blob properties are still the length for entire blob. We have to find a different way to do this

And it seems like there is easy way to do this by looking at the ReadableStreamChannel.getSize, since this method would return -1 for range request as well. 

Hmm, might have to resolve it by using the NettyResponseChannel.totalBytesReceived anyway.

ambry-api/src/main/java/com/github/ambry/frontend/EntityOperationMetrics.java
djycheng: The container metrics are separated by request type already so there wouldn't be any skewing, for example container metrics for POST and container metrics for DELETE are recorded separately

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
```diff
@@ -181,6 +181,7 @@ private CallbackChain(RestRequest restRequest, RestResponseChannel restResponseC
     private void start() {
       restRequest.getMetricsTracker()
           .injectMetrics(frontendMetrics.putBlobMetricsGroup.getRestRequestMetrics(restRequest.isSslUsed(), false));
+      restRequest.getMetricsTracker().setBytesTransferred(restRequest.getBytesReceived());
```
Arun-LinkedIn, line 184: This might have to be called in `securityProcessResponseCallback()`. The `bytesReceived` in `NettyRequest` will be updated as we start reading from `PutOperation` (via `startReadingFromChannel()` method). It will be equal to blob size after all the data is read.

ambry-frontend/src/main/java/com/github/ambry/frontend/PostBlobHandler.java
```diff
@@ -154,6 +154,7 @@ private void start() {
         // Start the callback chain by parsing blob info headers and performing request security processing.
         BlobInfo blobInfo = getBlobInfoFromRequest();
         checkUploadRequirements(blobInfo.getBlobProperties());
+        restRequest.getMetricsTracker().setBytesTransferred(restRequest.getBytesReceived());
```
Arun-LinkedIn, line 157: Same comment as above. I think this might have to be called in `securityProcessResponseCallback()`

ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobPutHandler.java
djycheng: I added it to the router put callback since we're doing things with `bytesReceived` there already

Arun-LinkedIn (approved):
Just had one comment. Rest of it lgtm!

PR #3109, opened 2025-06-04
return empty array when replica ids are null

manbearpig1996:
## Summary

return empty array when replica ids are null in FileCopyScheduler

## Testing Done
unit tests


aga9900 (approved):
LGTN

PR #3120, opened 2025-06-18
Enforce container count limit per account

satishkotha:
## Summary
There is potential overflow of containerId when new containers are added. This overflow can cause containerId to become negative and cause unintended side-effects. So throw error instead of silently rolling over to negative containerId

We can consider planning further work to support higher number of containers as follow up change.

## Testing Done
Added unit tests

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
@@ -151,6 +151,11 @@ public abstract class AbstractAccountService implements AccountService {
       throw new AccountServiceException("Account " + accountName + " is not found", AccountServiceErrorCode.NotFound);
     }
 
+    if (containers.size() > Short.MAX_VALUE) {
+      accountServiceMetrics.maxContainersReachedForAccount.inc();
+      throw new AccountServiceException("Too many containers in request: " + containers.size(), AccountServiceErrorCode.BadRequest);
+    }
+
     List<Container> resolvedContainers = new ArrayList<>();
     List<Container> existingUnchangedContainers = new ArrayList<>();
     // create a hashmap to map the name to existing containers in account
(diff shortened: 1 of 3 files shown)


ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
```diff
@@ -204,8 +209,11 @@ public Collection<Container> updateContainers(String accountName, Collection<Con
               }
           }
         } else {
+          if (nextContainerId > Short.MAX_VALUE) {
+            throw new AccountServiceException("Account reached max containers", AccountServiceErrorCode.InternalError);
```
Arun-LinkedIn, line 213: Would we need to track `maxContainersReachedForAccount` metric here too?

ambry-account/src/main/java/com/github/ambry/account/AbstractAccountService.java
```diff
@@ -151,19 +151,24 @@ public Collection<Container> updateContainers(String accountName, Collection<Con
       throw new AccountServiceException("Account " + accountName + " is not found", AccountServiceErrorCode.NotFound);
     }
 
+    if (containers.size() > Short.MAX_VALUE) {
+      accountServiceMetrics.maxContainersReachedForAccount.inc();
+      throw new AccountServiceException("Too many containers in request: " + containers.size(), AccountServiceErrorCode.BadRequest);
```
Arun-LinkedIn, line 156: QQ. Do we need this check? If containers.size() is greater than `Short.MAX_VALUE`, it would fail (or have failed) in the other condition?

satishkotha: this is done to fail early if any request comes with more 32K containers. if you think this is unnecessary, i can remove it

PR #3133, opened 2025-07-29
MySQL Cleaner Exception Handling

allenaverbukh:
## Summary
Account for stale size being -1 and causing error. Make logs only run if stale size is greater than 0.

## Testing Done
N/A


ambry-frontend/src/main/java/com/github/ambry/frontend/NamedBlobsCleanupRunner.java
```diff
@@ -57,6 +57,7 @@ public void run() {
       List<StaleNamedBlob> batchStaleBlobs = Collections.emptyList();
       NamedBlobDb.StaleBlobsWithLatestBlobName staleBlobsWithLatestBlobName;
       for (Container container : combinedContainers) {
+        logger.info("Started the cleaner for container: {}", container.getId());
```
justinlin-linkedin, line 60: let's move this log under the if statement below, we don't need this message for a container that disabled named blobs.

justinlin-linkedin (approved):

PR #3138, opened 2025-08-19
Fix S3 HeadBucket requests hanging indefinitely due to broken callback chain

jiayan97:
## Summary

This commit fixes a bug where S3 HeadBucket requests would hang indefinitely due to improper error propagation in the security callback chain.

**Root Cause**: `securityProcessRequestCallback()` passed `null` instead of `finalCallback` as the error handler, breaking the chain when security processing failed.

**Impact**: The fix ensures HeadBucket requests now complete with appropriate HTTP status codes (200 OK for success, 404 for failures) instead of hanging indefinitely.

## Changes Made
* Fix `securityProcessRequestCallback` to use `finalCallback` instead of `null` to ensure error propagation.
* Add proper response status and headers for successful HeadBucket responses (200 OK, Content-Length: 0, Date header).
* Add test coverage for error scenarios that would have exposed this bug.

## Testing Done
Added 3 new test methods in `S3HeadHandlerTest`:
  * `headBucketNotFoundTest`: Verifies requests complete with 404 when container doesn't exist instead of hanging
  * `headBucketNonExistentAccountTest`: Verifies requests complete with error when account doesn't exist
  * `headBucketResponseHeadersTest`: Validates proper response headers are set for successful HeadBucket operations

Verified:
* S3HeadHandlerTest passes: `./gradlew ambry-frontend:test --tests S3HeadHandlerTest`
* Full frontend module test suite passes: `./gradlew :ambry-frontend:test`

### DIFF
--- a/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3HeadHandler.java
+++ b/ambry-frontend/src/main/java/com/github/ambry/frontend/s3/S3HeadHandler.java
@@ -28,6 +28,8 @@ import com.github.ambry.rest.RestResponseChannel;
 import com.github.ambry.rest.RestServiceErrorCode;
 import com.github.ambry.rest.RestServiceException;
 import com.github.ambry.rest.RestUtils;
+import com.github.ambry.rest.ResponseStatus;
+import java.util.GregorianCalendar;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


Arun-LinkedIn (approved):
Thank you for making the change!

PR #3142, opened 2025-08-27
Support case insensitive for container name

SophieGuo410:
## Summary
Support case insensitive for container name

## Testing Done
wc-test


ambry-account/src/test/java/com/github/ambry/account/MySqlAccountServiceTest.java
```diff
@@ -583,6 +583,30 @@ public void testCreateExistingContainerInDifferentStates() throws Exception {
...
+    mySqlAccountService.updateAccounts(Collections.singletonList(testAccount));
+
+    // Verify container retrieval is case-insensitive
+    assertNotNull("Container should be retrievable with original case",
+        mySqlAccountService.getContainerByName(testAccount.getName(), "TestContainer"));
+    assertNotNull("Container should be retrievable with lowercase",
+        mySqlAccountService.getContainerByName(testAccount.getName(), "testcontainer"));
+    assertNotNull("Container should be retrievable with uppercase",
+        mySqlAccountService.getContainerByName(testAccount.getName(), "TESTCONTAINER"));
+  }
```
djycheng, line 608: Should we assert that the container ID is the same to ensure that we're getting the same container?

PR #3151, opened 2025-09-30
Add migration config to container metadata

jainlavina:
## Summary
Add migration config to container metadata to control migration of storage backend from one system to another.
The same migration config is added to account metadata as well as container metadata.
The field "overrideAccountMigrationConfig" will be used to determine whether the data path uses the migration config set on the account or the container.
The migration config initialized on the Container is the one persisted in mysql.

## Testing Done
Added unit tests.


SophieGuo410 (approved):
I've seen some test failed.

PR #3217, opened 2026-03-24
account: fail fast on startup when account metadata cannot be deserialized

j-tyler:
## Summary

When HelixAccountService starts up, it fetches account metadata from ZooKeeper and populates an in-memory cache.  If deserialization fails (e.g., due to version skew where a newer code version has written fields the running code does not recognize), the exception was caught by a blanket catch(Exception) in the background updater Runnable, logged, and silently discarded.  The service would then begin serving traffic with an empty account cache, causing every blob operation to return 400 InvalidAccount for valid blobs — a silent data-serving failure that is difficult to diagnose in production.

Separate the initial fetch from the background updater.  The initial fetch now propagates exceptions so the service refuses to start with a broken cache.  The background updater retains its catch-all since it already holds a valid cache from a prior successful fetch.

Additionally, fix a latent null-pointer dereference in AccountMetadataStore.fetchAccountMetadata(): when a ZNRecord exists but contains no account metadata map, fetchAccountMetadataFromZNRecord() returns null, which was passed unchecked into a for-each loop.  Return Collections.emptyList() instead, distinguishing "ZNRecord exists but is empty" (valid, service starts with zero accounts) from "ZNRecord does not exist" (returns null, triggers backup fallback).

Fixes: a]  silent 400 InvalidAccount on hosts deployed with older code
           when account metadata contains fields from newer versions
       b]  NullPointerException on startup when ZNRecord has no
           account metadata map

## Testing Done

Added and updated tests in `HelixAccountServiceTest.java` to validate positive and negative cases.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/AccountMetadataStore.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/AccountMetadataStore.java
@@ -16,6 +16,7 @@ package com.github.ambry.account;
 import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import org.apache.helix.AccessOption;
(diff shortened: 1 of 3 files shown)


crliao (approved):
👍

PR #3222, opened 2026-04-14
Add backwards compatibility regression tests for Account/MigrationConfig JSON

nicolaslopezbravo:
## Summary

- Adds snapshot-based regression tests that catch backwards-incompatible changes to the Account and MigrationConfig JSON wire formats
- Hardcoded "old" JSON snapshots verify deserialization always succeeds with safe defaults when new fields are missing
- Field-set assertions fail with a clear message when someone adds a new field, forcing them to add a companion backwards-compat test
- Demo tests inject unknown future fields to verify the `@JsonIgnoreProperties(ignoreUnknown = true)` protection (from #3218) works

## Testing Done

- [x] All 10 new tests in `BackwardsCompatibilityTest` pass on upstream/master
- [x] Verified field set regression catches injected new fields (WriteRamp + Account demos)
- [x] Verified old snapshot JSONs (without `dualHeadSyncPct`, without `rampControl`/`migrationConfig`) deserialize with correct defaults
- [x] Verified unknown future fields in MigrationConfig are silently ignored thanks to `@JsonIgnoreProperties` from #3218

### DIFF
--- a/ambry-api/src/test/java/com/github/ambry/account/BackwardsCompatibilityTest.java
+++ b/ambry-api/src/test/java/com/github/ambry/account/BackwardsCompatibilityTest.java
@@ -1,0 +1,462 @@
+/*
+ * Copyright 2017 LinkedIn Corp. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ */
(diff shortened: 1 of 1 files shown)


ambry-api/src/test/java/com/github/ambry/account/BackwardsCompatibilityTest.java
```diff
@@ -0,0 +1,526 @@
...
+    assertEquals("RampControl fields changed. Update EXPECTED_RAMP_CONTROL_FIELDS and add a backwards compat test.",
+        EXPECTED_RAMP_CONTROL_FIELDS, actualFields);
+  }
+
+  // ==================== Breakage Demos ====================
+  // These tests simulate what happens when a newer Ambry version writes JSON with a field
+  // that the current code doesn't know about, or when we add a field and forget backwards compat.
+
+  /**
+   * DEMO: A newer Ambry version added "dualDeleteSyncPct" to WriteRamp and serialized it to ZooKeeper.
```
crliao, line 373: nit: nothing related to ZK here

crliao (approved):

PR #3223, opened 2026-04-21
Require multi-DC catchup during bootstrap when local DC peers unavailable

satishkotha:
## Summary

During multi-fabric crashes, bootstrapping replicas could sync with empty replicas from a single recovering DC (stale ExternalView shows them as STANDBY) and transition to STANDBY with no data, causing 404s.

Add config flag clustermap.replica.catchup.require.multi.dc.for.bootstrap (default: false) that requires caught-up peers from at least 2 distinct remote DCs when no local DC peers are in STANDBY/LEADER.

## Testing Done
added unit tests. all tests pass.

### DIFF
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
@@ -420,7 +420,36 @@ public class AmbryReplicaSyncUpManager implements ReplicaSyncUpManager {
       }
       // We don't need to check if replicas, which have been caught up with, are up or down currently. As long as, the
       // peer replica has been put into catchup set, this means it has been caught up sometime before it went down.
-      return localDcCaughtUpReplicas.size() + remoteDcCaughtUpReplicas.size() >= catchupTarget;
+      int totalCaughtUp = localDcCaughtUpReplicas.size() + remoteDcCaughtUpReplicas.size();
+      if (totalCaughtUp < catchupTarget) {
+        return false;
+      }
+      // When enabled, during bootstrap with no local DC peers in STANDBY/LEADER, require caught-up peers from at least
+      // 2 distinct remote DCs. This prevents completing bootstrap by syncing with a single remote DC that may itself be
+      // recovering with empty data (e.g., multi-fabric crash where lor1 syncs with empty lva1 replicas while ltx1 has
+      // the real data).
+      if (clusterMapConfig.clustermapReplicaCatchupRequireMultiDcForBootstrap
+          && currentState == ReplicaState.BOOTSTRAP
+          && localDcCaughtUpReplicas.isEmpty()) {
+        long distinctRemoteDcsWithPeers = remoteDcPeerReplicaAndLag.keySet().stream()
+            .map(r -> r.getDataNodeId().getDatacenterName())
+            .distinct()
+            .count();
+        // Only enforce multi-DC check when there are at least 2 remote DCs with peers
+        if (distinctRemoteDcsWithPeers >= 2) {
+          long distinctCaughtUpDcs = remoteDcCaughtUpReplicas.stream()
+              .map(r -> r.getDataNodeId().getDatacenterName())
+              .distinct()
+              .count();
+          if (distinctCaughtUpDcs < 2) {
+            logger.info("Partition {} bootstrap sync-up: caught up with {} peers but only from {} distinct remote DC(s),"
+                    + " requiring at least 2. LocalDcPeers: {}, RemoteDcCaughtUp: {}", replicaOnCurrentNode.getPartitionId().toPathString(),
+                totalCaughtUp, distinctCaughtUpDcs, localDcPeerReplicaAndLag.size(), remoteDcCaughtUpReplicas);
+            return false;
+          }
+        }
+      }
+      return true;
     }
 
     @Override
--- a/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
+++ b/ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
@@ -411,7 +411,14 @@ public class AmbryReplicaSyncUpManager implements ReplicaSyncUpManager {
     }
 
     /**
-     * @return whether current replica has caught up with enough peers
+     *
+     * Returns true when this replica has synced up with enough peers to complete its state transition. Specifically, the total number of caught-up peers (local + remote) must meet catchupTarget, AND one of the following must hold:
+     *
+     *  1. At least one local-DC peer has caught up (satisfied by local alone, or local plus remote).
+     *  2. No local-DC peer has caught up, but caught-up peers span at least 2 distinct remote DCs (enforced only when clustermap.replica.catchup.require.multi.dc.for.bootstrap=true and currentState == BOOTSTRAP).
+     *
+     *  Why the asymmetry: Helix's ExternalView for local-DC peers is treated as reliable since same-DC state changes propagate quickly. The view of remote-DC peers can be stale — reporting them as STANDBY/LEADER while they are actually still in
+     *  BOOTSTRAP with empty data. Requiring two distinct remote DCs reduces the chance of completing bootstrap against a single stale view (e.g., during a multi-fabric recovery where empty replicas appear caught-up against each other).
      */
     boolean hasSyncedUpWithEnoughPeers() {
       // If there are no peers at all (single-replica partition), sync-up is trivially complete.
@@ -449,6 +456,9 @@ public class AmbryReplicaSyncUpManager implements ReplicaSyncUpManager {
           }
         }
       }
+      logger.info("Partition {} sync-up complete: state {}, localDcCaughtUp {}, remoteDcCaughtUp {}, catchupTarget {}",
+          replicaOnCurrentNode.getPartitionId().toPathString(), currentState, localDcCaughtUpReplicas,
+          remoteDcCaughtUpReplicas, catchupTarget);
       return true;
     }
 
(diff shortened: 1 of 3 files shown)


abha-mutalik (approved):

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -420,7 +420,36 @@ boolean hasSyncedUpWithEnoughPeers() {
...
+      int totalCaughtUp = localDcCaughtUpReplicas.size() + remoteDcCaughtUpReplicas.size();
+      if (totalCaughtUp < catchupTarget) {
+        return false;
+      }
+      // When enabled, during bootstrap with no local DC peers in STANDBY/LEADER, require caught-up peers from at least
+      // 2 distinct remote DCs. This prevents completing bootstrap by syncing with a single remote DC that may itself be
+      // recovering with empty data (e.g., multi-fabric crash where lor1 syncs with empty lva1 replicas while ltx1 has
+      // the real data).
+      if (clusterMapConfig.clustermapReplicaCatchupRequireMultiDcForBootstrap
+          && currentState == ReplicaState.BOOTSTRAP
```
beijxu, line 432: A newbie question:
If we return true from here as long as the two other DCs are ready, will this replica stays in bootstrap state forever

satishkotha: if we return true here, then current node will transition to standby/leader and start 'serving'.

if we return false, then current node will try to sync with more replicas and stay in bootstrap mode for more time. this will delay 'serving' for some more time.  But staying in bootstrap mode is much better than starting serving and returning incorrect results

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -420,7 +420,36 @@ boolean hasSyncedUpWithEnoughPeers() {
...
+          && localDcCaughtUpReplicas.isEmpty()) {
```
snalli, line 433: Catching up with a single local peer short-circuits the multi-DC catchup requirement — can we be sure that peer was itself validated across DCs? Would requiring multi-DC catchup unconditionally (when the flag is on) be safer?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -420,7 +420,36 @@ boolean hasSyncedUpWithEnoughPeers() {
...
+            .distinct()
+            .count();
+        // Only enforce multi-DC check when there are at least 2 remote DCs with peers
+        if (distinctRemoteDcsWithPeers >= 2) {
+          long distinctCaughtUpDcs = remoteDcCaughtUpReplicas.stream()
+              .map(r -> r.getDataNodeId().getDatacenterName())
+              .distinct()
+              .count();
+          if (distinctCaughtUpDcs < 2) {
+            logger.info("Partition {} bootstrap sync-up: caught up with {} peers but only from {} distinct remote DC(s),"
```
snalli, line 445: How often will this INFO log fire? checkDone() runs on every updateReplicaLagAndCheckSyncStatus() call, so would a partition with 4 peers produce 4 log lines per replication round?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -420,7 +420,36 @@ boolean hasSyncedUpWithEnoughPeers() {
...
-      return localDcCaughtUpReplicas.size() + remoteDcCaughtUpReplicas.size() >= catchupTarget;
+      int totalCaughtUp = localDcCaughtUpReplicas.size() + remoteDcCaughtUpReplicas.size();
+      if (totalCaughtUp < catchupTarget) {
+        return false;
+      }
+      // When enabled, during bootstrap with no local DC peers in STANDBY/LEADER, require caught-up peers from at least
+      // 2 distinct remote DCs. This prevents completing bootstrap by syncing with a single remote DC that may itself be
+      // recovering with empty data (e.g., multi-fabric crash where lor1 syncs with empty lva1 replicas while ltx1 has
+      // the real data).
+      if (clusterMapConfig.clustermapReplicaCatchupRequireMultiDcForBootstrap
```
snalli, line 431: Could this be a numeric config (e.g. clustermapReplicaCatchupRequireMultiDcForBootstrap, default 0) instead of a boolean? And can be used in the checks below
0 = disabled, 
2 = current PR, 
3+ = stricter deployments.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
satishkotha: the assumption is remote peer state can be stale ExternalView shows them as STANDBY, so we cannot trust remote peers state.

however, local peers view is always consistent. if you think local peer view can also become stale for some reason, then we can switch to multi-DC catchup unconditionally.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
satishkotha: this is only reported during bootstrap. I think its good to have verbose logging. This is also special case only when local peers are not available. So i dont think verbose logging is a problem.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
satishkotha: will we ever have 3+ DC for ambry? dont have strong opinion on this. can switch if you think its useful.

snalli: No problem. Boolean is ok.

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
```diff
@@ -420,7 +420,36 @@ boolean hasSyncedUpWithEnoughPeers() {
...
+              .count();
+          if (distinctCaughtUpDcs < 2) {
+            logger.info("Partition {} bootstrap sync-up: caught up with {} peers but only from {} distinct remote DC(s),"
+                    + " requiring at least 2. LocalDcPeers: {}, RemoteDcCaughtUp: {}", replicaOnCurrentNode.getPartitionId().toPathString(),
+                totalCaughtUp, distinctCaughtUpDcs, localDcPeerReplicaAndLag.size(), remoteDcCaughtUpReplicas);
+            return false;
+          }
+        }
+      }
+      return true;
```
snalli, line 462: Could you please add a comment here or in the javadoc above the fn like:

```
Returns true when this replica has synced up with enough peers to complete its state transition. Specifically, the total number of caught-up peers (local + remote) must meet catchupTarget, AND one of the following must hold:
  ▎
  ▎ 1. At least one local-DC peer has caught up (satisfied by local alone, or local plus remote).
  ▎ 2. No local-DC peer has caught up, but caught-up peers span at least 2 distinct remote DCs (enforced only when clustermap.replica.catchup.require.multi.dc.for.bootstrap=true and currentState == BOOTSTRAP).
  ▎
  ▎ Why the asymmetry: Helix's ExternalView for local-DC peers is treated as reliable since same-DC state changes propagate quickly. The view of remote-DC peers can be stale — reporting them as STANDBY/LEADER while they are actually still in
  ▎ BOOTSTRAP with empty data. Requiring two distinct remote DCs reduces the chance of completing bootstrap against a single stale view (e.g., during a multi-fabric recovery where empty replicas appear caught-up against each other).
```

ambry-clustermap/src/test/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManagerTest.java
```diff
@@ -351,4 +359,117 @@ public void singleReplicaPartitionTest() throws Exception {
...
+    testWithoutMultiDc.replicaSyncUpService.initiateBootstrap(testWithoutMultiDc.currentReplica);
+
+    // Separate remote peers by DC
+    List<ReplicaId> dc2Peers = testWithoutMultiDc.remoteDcPeerReplicas.stream()
+        .filter(r -> r.getDataNodeId().getDatacenterName().equals("DC2"))
+        .collect(Collectors.toList());
+    assertTrue("Test requires at least 2 DC2 peers", dc2Peers.size() >= 2);
+
+    // Catch up with 2 peers from DC2 only — should complete because flag is off
+    testWithoutMultiDc.replicaSyncUpService.updateReplicaLagAndCheckSyncStatus(
```
snalli: assertFalse ?

snalli:
Overall LGTM

1. Do you know how long it takes to bootstrap ? Is it file-based bootstrap now ? 
2. What's the deployment plan for this ?

ambry-clustermap/src/main/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManager.java
satishkotha: done

ambry-clustermap/src/test/java/com/github/ambry/clustermap/AmbryReplicaSyncUpManagerTest.java
satishkotha: fixed

PR #3226, opened 2026-04-22
Always create bootstrap marker file to prevent serving stale data after restart

satishkotha:
## Summary
Previously, the bootstrap marker file was only created when store capacity was <= HEADER_SIZE (18 bytes), intended to catch disk replacements. This meant stores with existing data skipped sync-up entirely after restart, transitioning directly to STANDBY with stale data and serving 404s for blobs written during downtime.

Remove the HEADER_SIZE conditional so every OFFLINE -> BOOTSTRAP transition goes through the ReplicaSyncUpManager. Stores that are already caught up complete sync-up in <1 second. Stores with stale data correctly wait for peer catch-up before serving reads.

Every store restart now goes through sync-up. For caught-up stores this adds <1s. For stale stores, this correctly blocks until caught up.

## Testing Done
covered by existing unit tests

### DIFF
--- a/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
+++ b/ambry-store/src/main/java/com/github/ambry/store/StorageManager.java
@@ -1010,20 +1010,19 @@ public class StorageManager implements StoreManager {
           ((BlobStore) store).setRecoverFromDecommission(false);
         }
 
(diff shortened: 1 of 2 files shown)


crliao (approved):
ty!

nicolaslopezbravo (approved):

snalli (approved):

PR #3227, opened 2026-04-22
Fail Ambry services when account-cache is empty (#ACTIONITEM-16590)

snalli:
MySqlAccountService and HelixAccountService load accounts synchronously in their constructors (remote fetch + backup-file fallback). If both fail, the cache is empty — which would cause the service to return null for every account/container lookup, resulting in the server rejecting valid requests and skipping valid replication messages.

Once populated, the cache cannot become empty through normal operation (CachedAccountService only adds/updates entries). So an empty cache unambiguously means the initial load failed. Fail fast at factory construction rather than running in a broken state.

## Coverage

Both **frontend** (RestServer) and **server** (AmbryServer, VcrServer) are covered because all three construct AccountService via the factory path. In AmbryLI prod, the frontend's CompositeLiAccountServiceFactory wraps MySqlAccountServiceFactory as primary, so the check fires transitively.

| App | Factory in prod | Coverage |
|---|---|---|
| OSS ambry-server (AmbryServer.startup) | MySqlAccountServiceFactory or HelixAccountServiceFactory | Direct |
| OSS ambry-frontend (RestServer) | MySqlAccountServiceFactory or HelixAccountServiceFactory | Direct |
| AmbryLI ambry-server | MySqlAccountServiceFactory | Direct |
| AmbryLI ambry-frontend (prod-lva1/lor1/ei4) | CompositeLiAccountServiceFactory | Transitive via MySql primary |
| AmbryLI ambry-frontend (qei) | MySqlAccountServiceFactory | Direct |
| AmbryLI ambry-vcr | MySqlAccountServiceFactory | Direct |

## Changes

- **AccountService** (interface): add `getAccountCount()`, `getContainerCount()`, `failIfCacheIsEmpty()` default methods. `failIfCacheIsEmpty()` logs and throws IllegalStateException.
- **MySqlAccountServiceFactory, HelixAccountServiceFactory**: call `failIfCacheIsEmpty()` before returning.
- **AccountServiceDefaultMethodsTest**: new unit tests for the interface default methods.
- **AccountUpdateToolTest**: seed ZK with a dummy account after wipe so the integration test's factory call can pass the non-empty check. Test assertions bumped by DUMMY_ACCOUNT_OFFSET (1).

Supersedes #3228.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/account/AccountService.java
+++ b/ambry-api/src/main/java/com/github/ambry/account/AccountService.java
@@ -22,6 +22,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Consumer;
+import org.slf4j.LoggerFactory;
 
 
 /**
@@ -93,6 +94,40 @@ public interface AccountService extends Closeable {
    * @return A collection of {@link Account}s.
    */
   Collection<Account> getAllAccounts();
+
+  /**
+   * @return the number of accounts currently loaded by this {@code AccountService}. Used by server startup
+   * checks to detect an empty account cache. Implementations may override for efficiency.
+   */
+  default int getAccountCount() {
+    return getAllAccounts().size();
+  }
+
+  /**
+   * @return the total number of containers across all accounts currently loaded by this {@code AccountService}.
+   * Used by server startup checks to detect an empty cache. Implementations may override for efficiency.
+   */
+  default int getContainerCount() {
+    return getAllAccounts().stream().mapToInt(Account::getContainerCount).sum();
+  }
+
+  /**
+   * Fails fast if the account or container cache is empty. Intended to be called once at server startup,
+   * immediately after the initial (synchronous) fetch returns. An empty cache means the remote load failed
+   * and fell through to an empty backup file, which would cause the server to serve incorrect responses
+   * (e.g., rejecting valid requests, skipping valid replication messages).
+   * @throws IllegalStateException if the cache has zero accounts or zero containers.
+   */
+  default void failIfCacheIsEmpty() {
+    int accountCount = getAccountCount();
+    int containerCount = getContainerCount();
+    if (accountCount == 0 || containerCount == 0) {
+      String message = "AccountService cache is empty after initial load (accounts=" + accountCount
+          + ", containers=" + containerCount + ")";
+      LoggerFactory.getLogger(AccountService.class).error(message);
+      throw new IllegalStateException(message);
+    }
+  }
 
   /**
    * Adds a {@link Consumer} for newly created or updated {@link Account}s.
--- a/ambry-tools/src/integration-test/java/com/github/ambry/account/AccountUpdateToolTest.java
+++ b/ambry-tools/src/integration-test/java/com/github/ambry/account/AccountUpdateToolTest.java
@@ -37,6 +37,7 @@ import java.util.UUID;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
+import org.apache.helix.zookeeper.datamodel.ZNRecord;
 import org.json.JSONArray;
 import org.junit.After;
 import org.junit.AfterClass;
@@ -58,6 +59,10 @@ import static org.junit.Assert.*;
 public class AccountUpdateToolTest {
   private static final int NUM_REF_ACCOUNT = 10 + (int) (Math.random() * 50);
   private static final int NUM_CONTAINER_PER_ACCOUNT = 4;
+  // A dummy account seeded in ZK before each test so the HelixAccountServiceFactory non-empty-cache check passes.
+  // Its ID is excluded from generateRefAccounts() via accountIdSet so it never collides with generated accounts.
+  private static final short DUMMY_ACCOUNT_ID = Short.MAX_VALUE;
+  private static final int DUMMY_ACCOUNT_OFFSET = 1;
   private static final int ZK_SERVER_PORT = 2600;
   private static final String DC_NAME = "testDc";
   private static final byte DC_ID = (byte) 1;
@@ -114,7 +119,10 @@ public class AccountUpdateToolTest {
     this.containerJsonVersion = containerJsonVersion;
     idToRefAccountMap = new HashMap<>();
     idToRefContainerMap = new HashMap<>();
-    generateRefAccounts(idToRefAccountMap, idToRefContainerMap, new HashSet<>(), NUM_REF_ACCOUNT,
+    // Reserve DUMMY_ACCOUNT_ID so generateRefAccounts() doesn't pick it for any ref account.
+    HashSet<Short> reservedIds = new HashSet<>();
+    reservedIds.add(DUMMY_ACCOUNT_ID);
+    generateRefAccounts(idToRefAccountMap, idToRefContainerMap, reservedIds, NUM_REF_ACCOUNT,
         NUM_CONTAINER_PER_ACCOUNT);
     tempDirPath = getTempDir("accountUpdateTool-");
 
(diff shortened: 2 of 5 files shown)


ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
```diff
@@ -242,6 +242,11 @@ protected synchronized void fetchAndUpdateCache() throws SQLException {
         needRefresh = false;
         lastSyncTime = endTimeMs;
       }
+      failIfCacheIsEmpty();
+    } catch (AccountServiceException e) {
+      accountServiceMetrics.fetchRemoteAccountErrorCount.inc();
```
satishkotha: nit: why not report all the metrics/logs in failIfCacheIsEmpty and avoid catch/rethrow pattern?  personally, i'd prefer avoiding catch/rethrow pattern.

satishkotha (commented):
we also want to  fail server healthcheck if accounts cannot be loaded especially at the start. is that going to be done as a followup?

ambry-account/src/main/java/com/github/ambry/account/CachedAccountService.java
snalli: Addressed — moved the metric increments and error log into `failIfCacheIsEmpty()`. The outer catch block is now a pass-through (kept because the `SQLException` catch below it wouldn't catch `AccountServiceException`).

snalli:
> we also want to fail server healthcheck if accounts cannot be loaded especially at the start. is that going to be done as a followup?


Thanks for the review! For the server healthcheck follow-up — agreed, will do that in a separate PR.

snalli:
> > we also want to fail server healthcheck if accounts cannot be loaded especially at the start. is that going to be done as a followup?
> 
> Thanks for the review! For the server healthcheck follow-up — agreed, will do that in a separate PR.

On second thought, I did it in this PR. Both frontend and server are covered.

ambry-tools/src/integration-test/java/com/github/ambry/account/AccountUpdateToolTest.java
```diff
@@ -58,6 +59,10 @@
 public class AccountUpdateToolTest {
   private static final int NUM_REF_ACCOUNT = 10 + (int) (Math.random() * 50);
   private static final int NUM_CONTAINER_PER_ACCOUNT = 4;
+  // A dummy account seeded in ZK before each test so the HelixAccountServiceFactory non-empty-cache check passes.
```
satishkotha, line 62: do we need something similar for our internal e2e tests? could you please check if playground.sh starts up with this check?

satishkotha (commented):
change looks good. question about local/e2e testing, then this can be merged

snalli:
End-to-end verified locally against AmbryLI's playground (Apple M4 Pro, colima x86_64+Rosetta, locally-published OSS jars at v0.5.140 + this commit).

**Empty cache → fast-fail (expected behavior of this PR):**

With a freshly-initialized MySQL (`AMBRYBLOBS.Accounts` and `Containers` both empty), `mint deploy ambry-server` exits at startup with:

```
Caused by: java.lang.IllegalStateException: AccountService cache is empty after initial load (accounts=0, containers=0)
    at com.github.ambry.account.AccountService.failIfCacheIsEmpty(AccountService.java:128)
    at com.github.ambry.account.MySqlAccountServiceFactory.getAccountService(MySqlAccountServiceFactory.java:61)
    ...
java.lang.InstantiationException: failure during startup java.lang.IllegalStateException: Could not instantiate MySqlAccountService
    at com.github.ambry.server.AmbryServer.startup(AmbryServer.java:562)
```

The factory-level check fires before any data path comes up, so an operator misconfiguration (or a degraded MySQL/ZK leaving the cache empty) surfaces immediately rather than later as a silent serving anomaly.

**Unit tests** (this branch, `:ambry-api:test` + `:ambry-account:test` + `:ambry-tools:intTest`):
- `AccountServiceDefaultMethodsTest`: 5/5 PASSED
- `:ambry-account:test`: 142/142 PASSED, 0 failed
- `AccountUpdateToolTest` (integration): 10/10 PASSED

LGTM on my side — happy to merge once CI is green.

satishkotha:
> End-to-end verified locally against AmbryLI's playground (Apple M4 Pro, colima x86_64+Rosetta, locally-published OSS jars at v0.5.140 + this commit).
> 
> **Empty cache → fast-fail (expected behavior of this PR):**
> 
> With a freshly-initialized MySQL (`AMBRYBLOBS.Accounts` and `Containers` both empty), `mint deploy ambry-server` exits at startup with:
> 
> ```
> Caused by: java.lang.IllegalStateException: AccountService cache is empty after initial load (accounts=0, containers=0)
>     at com.github.ambry.account.AccountService.failIfCacheIsEmpty(AccountService.java:128)
>     at com.github.ambry.account.MySqlAccountServiceFactory.getAccountService(MySqlAccountServiceFactory.java:61)
>     ...
> java.lang.InstantiationException: failure during startup java.lang.IllegalStateException: Could not instantiate MySqlAccountService
>     at com.github.ambry.server.AmbryServer.startup(AmbryServer.java:562)
> ```
> 
> The factory-level check fires before any data path comes up, so an operator misconfiguration (or a degraded MySQL/ZK leaving the cache empty) surfaces immediately rather than later as a silent serving anomaly.
> 
> **Unit tests** (this branch, `:ambry-api:test` + `:ambry-account:test` + `:ambry-tools:intTest`):
> 
> * `AccountServiceDefaultMethodsTest`: 5/5 PASSED
> * `:ambry-account:test`: 142/142 PASSED, 0 failed
> * `AccountUpdateToolTest` (integration): 10/10 PASSED
> 
> LGTM on my side — happy to merge once CI is green.

should we have a mode in playground to disable this behavior? otherwise, how would we verify other functionality. we want to test basic put/get works for example

satishkotha (approved):

ambry-tools/src/integration-test/java/com/github/ambry/account/AccountUpdateToolTest.java
snalli: yes, trying out playground tests

ambry-api/src/main/java/com/github/ambry/account/AccountService.java
```diff
@@ -94,6 +95,40 @@ public interface AccountService extends Closeable {
...
+   * Fails fast if the account or container cache is empty. Intended to be called once at server startup,
+   * immediately after the initial (synchronous) fetch returns. An empty cache means the remote load failed
+   * and fell through to an empty backup file, which would cause the server to serve incorrect responses
+   * (e.g., rejecting valid requests, skipping valid replication messages).
+   * @throws IllegalStateException if the cache has zero accounts or zero containers.
+   */
+  default void failIfCacheIsEmpty() {
+    int accountCount = getAccountCount();
+    int containerCount = getContainerCount();
+    if (accountCount == 0 || containerCount == 0) {
```
beijxu, line 124: is it possible to have a brand new account without any container yet? should we only check on account Count?

snalli: Good question. The check operates on aggregate counts across the entire account cache, not per-account:

- `getAccountCount()` returns the total number of accounts in the local cache.
- `getContainerCount()` returns the total number of containers summed across all accounts.

So a brand-new account with zero containers does not on its own trigger the failure — it only triggers if the entire account cache has zero containers. In a production Ambry cluster (which has been writing blobs for years), this aggregate is guaranteed to be non-zero. The condition only fires when the cache is genuinely empty (e.g., the initial DB load failed and we fell through to an empty backup file).

Checking only `accountCount > 0` would catch the case where the load returned `[]`, but would let through a degenerate case where some metadata path returned the accounts but none of their containers — which would manifest the same way at runtime (rejecting valid requests, skipping replication messages). Keeping both bounds the failure mode tightly.

satishkotha (approved):

snalli:
Got approval from @beijxu on slack.

PR #3228, opened 2026-04-23
Fail AccountServiceFactory.getAccountService when cache is empty

snalli:
MySqlAccountService and HelixAccountService load accounts synchronously in their constructors (remote fetch + backup-file fallback). If both fail, the cache is empty — which would cause the service to return null for every account/container lookup, resulting in the server rejecting valid requests and skipping valid replication messages.

Once populated, the cache cannot become empty through normal operation (CachedAccountService only adds/updates entries). So an empty cache unambiguously means the initial load failed. Fail fast at factory construction time rather than running in a broken state.

Changes:
- Add getAccountCount(), getContainerCount(), failIfCacheIsEmpty() default methods to AccountService interface. failIfCacheIsEmpty() logs and throws IllegalStateException.
- MySqlAccountServiceFactory.getAccountService() calls failIfCacheIsEmpty() before returning; existing catch wraps any throwable in IllegalStateException.
- Same for HelixAccountServiceFactory.
- CompositeAccountServiceFactory is transitively covered via its inner primary/secondary factories.
- In-memory test factories intentionally skipped (tests start empty).

Companion to #3227 (adds an empty-cache check on the client/fetch side). Independent — either PR can merge first.

### DIFF
--- a/ambry-account/src/main/java/com/github/ambry/account/HelixAccountServiceFactory.java
+++ b/ambry-account/src/main/java/com/github/ambry/account/HelixAccountServiceFactory.java
@@ -84,6 +84,9 @@ public class HelixAccountServiceFactory implements AccountServiceFactory {
               : null;
       HelixAccountService helixAccountService =
           new HelixAccountService(helixStore, accountServiceMetrics, notifier, scheduler, accountServiceConfig);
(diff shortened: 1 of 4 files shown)


snalli:
Folded into #3227 — combining both checks into one PR for clarity.

PR #3229, opened 2026-04-24
Add replica lag metrics during STANDBY->LEADER promotion

beijxu:
## Summary

Helix transitions STANDBY->LEADER without any data-parity check. Add two mutually exclusive observability signals at the moment of promotion so we can verify in production that promoted leaders are actually caught up:

- StandbyToLeaderPromotionLagBytes (histogram): max cached lag across peers whose lag is known (>= 0). A single known peer is enough to attest readiness, so peers at the -1 sentinel are ignored when at least one peer has a known lag.
- StandbyToLeaderPromotionUnknownLagPeerCount (counter): incremented when every peer of the partition has an unknown cached lag -- no peer can attest to the replica's sync state.

The measurement is wrapped in try/catch with warn-level logging so a metrics path failure cannot propagate into the Helix state transition.

We are going to review the metrics once become unavailable, then determine how to catch premature state transition based on that

### DIFF
--- a/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
+++ b/ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
@@ -354,6 +354,46 @@ public class ReplicationManager extends ReplicationEngine {
         // leaderBasedReplicationAdmin. LeaderBasedReplicationAdmin::addLeaderPartition is thread safe.
         leaderBasedReplicationAdmin.addLeaderPartition(partitionName);
       }
+      try {
+        recordStandbyToLeaderPromotionLag(partitionName);
+      } catch (Exception e) {
+        logger.warn("Failed to record lag metric on STANDBY->LEADER promotion for partition {}", partitionName, e);
+      }
+    }
+
+    /**
+     * Detects whether a replica was behind its peers at the moment Helix promoted it STANDBY -> LEADER.
+     * Helix transitions STANDBY -> LEADER without any data-parity check; we emit metrics here so we can
+     * verify in production that promoted leaders are actually caught up (or flag when they aren't).
+     *
+     * Two mutually exclusive signals per promotion:
+     * - {@link ReplicationMetrics#standbyToLeaderPromotionLagBytes}: max cached lag across peers whose lag
+     *   is known (>= 0). A single known peer is enough to attest readiness, so unknown (-1) peers are
+     *   ignored when at least one peer has a known lag.
+     * - {@link ReplicationMetrics#standbyToLeaderPromotionUnknownLagPeerCount}: incremented when every peer
+     *   of the partition has an unknown cached lag — no peer can attest to the replica's sync state.
+     */
+    private void recordStandbyToLeaderPromotionLag(String partitionName) {
+      ReplicaId localReplica = storeManager.getReplica(partitionName);
+      if (localReplica == null) {
+        return;
+      }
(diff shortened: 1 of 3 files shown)


ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -354,6 +354,46 @@ public void onPartitionBecomeLeaderFromStandby(String partitionName) {
...
+     * verify in production that promoted leaders are actually caught up (or flag when they aren't).
+     *
+     * Two mutually exclusive signals per promotion:
+     * - {@link ReplicationMetrics#standbyToLeaderPromotionLagBytes}: max cached lag across peers whose lag
+     *   is known (>= 0). A single known peer is enough to attest readiness, so unknown (-1) peers are
+     *   ignored when at least one peer has a known lag.
+     * - {@link ReplicationMetrics#standbyToLeaderPromotionUnknownLagPeerCount}: incremented when every peer
+     *   of the partition has an unknown cached lag — no peer can attest to the replica's sync state.
+     */
+    private void recordStandbyToLeaderPromotionLag(String partitionName) {
```
satishkotha, line 376: IMO, this is more helpful for bootstrap->standby transition than standby-> leader.

can we make this generic and log this for every state transition? you can probably pass metric reference to make this generic.

beijxu: a few reasons I didn't do so:
- one bootstrap completion, if a clean sync cycle is not completed yet, then the RemoteReplicaInfo does not have the lag information yet. standby to leader promotion is the last gate before the replica is ready, so it has a higher chance to see the lag info
- on bootstrap completion, the code path does not have access to RemoteReplicaInfo, so it requires moving more code around. I don't feel comfortable to make a bigger change on this code path given its complexity

satishkotha (approved):

snalli (approved):

ambry-replication/src/main/java/com/github/ambry/replication/ReplicationManager.java
```diff
@@ -354,6 +354,46 @@ public void onPartitionBecomeLeaderFromStandby(String partitionName) {
...
+      if (localReplica == null) {
+        return;
+      }
+      PartitionInfo partitionInfo = partitionToPartitionInfo.get(localReplica.getPartitionId());
+      if (partitionInfo == null) {
+        return;
+      }
+      long maxKnownLag = -1L;
+      for (RemoteReplicaInfo remoteReplicaInfo : partitionInfo.getRemoteReplicaInfos()) {
+        long lag = remoteReplicaInfo.getLocalLagFromRemoteInBytes();
```
snalli, line 387: Thanks for adding these signals!
  
  One suggestion for consideration: would it be worth splitting StandbyToLeaderPromotionLagBytes into two histograms — one for local-DC peers and one for remote-DC peers — rather than reporting a single max across both?
  The motivation is that lag against the two peer classes carries quite different operational meaning, and the split is useful in both replication modes:
  
  Lag against local-DC peers is the signal we actually want for promotion fitness. Intra-DC replication is all-to-all in both ALL_TO_ALL and LEADER_BASED modes, so a healthy leader should be at ~0 against its local peers. 
  
  Lag against remote-DC peers is largely cross-DC propagation delay and is non-zero in steady state by design.

abha-mutalik (approved):

PR #3230, opened 2026-04-26
Bound MySqlNamedBlobDb transaction-executor queue and add admission metrics

nicolaslopezbravo:
## Summary

`MySqlNamedBlobDb.TransactionExecutor` currently uses an unbounded `ScheduledThreadPoolExecutor` (via `Utils.newScheduler`). When the named-blob MySQL backend slows down or a replica degrades, the executor's internal `DelayedWorkQueue` grows without limit. Each queued task pins a Netty channel + request context, so the queue accumulates multiple GB of retained heap until the JVM stops responding to liveness probes and is killed by the kubelet.

This PR:

1. Replaces the unbounded executor with a `ThreadPoolExecutor` backed by a bounded `LinkedBlockingQueue(N)` and `AbortPolicy`.
2. Catches `RejectedExecutionException` in `executeTransaction` / `executeTransactionGeneric` and completes the caller's `CompletableFuture` exceptionally with `RestServiceException(ServiceUnavailable)`. Existing cross-DC retry via `transactionStateTracker` remains intact, so a saturated local DC will retry to remote DCs before failing.
3. Adds the new config knob `mysql.named.blob.max.pending.transactions.per.datacenter` (default `100`). Order-of-magnitude sizing: `4 * localPoolSize * p99_query_latency_seconds`.
4. Adds metrics so the queue is observable before it overflows:
   - `MySqlNamedBlobDb.TransactionExecutorQueueSize.<datacenter>` (Gauge)
   - `MySqlNamedBlobDb.TransactionExecutorActiveCount.<datacenter>` (Gauge)
   - `MySqlNamedBlobDb.NamedBlobTransactionRejectedCount` (Counter)
   - `MySqlNamedBlobDb.NamedBlobEnqueueWaitTimeInMs` (Histogram)

Per-datacenter gauges are deregistered on `TransactionExecutor.close()` so repeated lifecycles do not leak metrics.

## Durability notes

Per the durability checklist:
- **Write path:** Rejection happens *before* the worker runs, so no DB write is performed. No partial-state risk.
- **Callback semantics:** Caller-facing change is from "future never completes (until JVM dies)" to "future completes exceptionally with `ServiceUnavailable`." Strictly improves the contract.
- **Resource cleanup:** `Connection` is acquired inside the lambda via try-with-resources; if the lambda is rejected, no connection is acquired.
- **Idempotency:** A rejection is equivalent to no work performed; client retries are safe and required.
- **`dbErrorCounter`:** Not incremented on rejection (this is admission control, not a DB error). The new `NamedBlobTransactionRejectedCount` counter tracks this case explicitly.

## Out-of-scope follow-ups

Identified during analysis but not in this PR:
- `Statement.setQueryTimeout(...)` on every `PreparedStatement` in `run_*_v2` paths.
- Hikari `connectionTimeout` / `validationTimeout` / `leakDetectionThreshold` tuning in `MySqlNamedBlobDbFactory`.
- JDBC URL `socketTimeout` / `connectTimeout` parameters in the dbInfo config.
- Capacity bump for `localPoolSize` (currently 5 — likely under-provisioned, but should be tuned together with the new queue cap once queue-depth metrics are observable).

## Testing Done

- `./gradlew :ambry-named-mysql:test` — 10 tests pass (4 new):
  - `testTransactionExecutorMetricsRegistered` — gauges + counter + histogram registered on construction.
  - `testTransactionExecutorGaugesRemovedOnClose` — gauges deregistered on close.
  - `testEnqueueWaitTimeRecordedOnSuccessfulSubmit` — histogram count > 0 after a normal submit.
  - `testTransactionRejectedWhenQueueFull` — single-thread executor with capacity 1, worker blocked: 1st submit runs, 2nd queues, 3rd rejected with `ServiceUnavailable`. Asserts gauge values reflect busy/queued state and the rejection counter increments.
- `./gradlew compileJava` — full project compiles.

### DIFF
--- a/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
@@ -77,11 +77,16 @@ public class MySqlNamedBlobDbConfig {
    * MySQL slows down: each queued task pins a Netty channel and request context, so an unbounded queue can grow
    * to multi-GB before the kubelet kills the pod.
    *
-   * <p>The default is conservatively sized for the prior unbounded behavior; raise it together with {@link #localPoolSize}
-   * when intentionally provisioning more capacity. Order-of-magnitude sizing: {@code 4 * localPoolSize * p99_query_latency_seconds}.
+   * <p><b>Sizing guidance.</b> The pre-bounded behavior was an unbounded {@code DelayedWorkQueue}, so any value
+   * here is more restrictive than what was deployed before. The default of {@code 1000} is intentionally generous
+   * relative to the steady-state heuristic ({@code 4 * localPoolSize * p99_query_latency_seconds}, typically 20-40
+   * for {@code localPoolSize=5}) so that the bounded behavior only kicks in under genuinely pathological load.
+   * Before rolling this out to a new deployment, observe {@code TransactionExecutorQueueSize.<dc>} p99 in
+   * staging/canary and tune the value if the steady-state queue depth approaches the cap. Raise this together
+   * with {@link #localPoolSize} when intentionally provisioning more capacity.
    */
   @Config(MAX_PENDING_TRANSACTIONS_PER_DATACENTER)
-  @Default("100")
+  @Default("1000")
   public final int maxPendingTransactionsPerDatacenter;
 
   /**
--- a/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
@@ -88,8 +88,12 @@ public class MySqlNamedBlobDbConfig {
    * <p>Once a value below {@link Integer#MAX_VALUE} is set, raise it together with {@link #localPoolSize} when
    * intentionally provisioning more capacity.
    */
+  // Default Integer.MAX_VALUE: the per-datacenter queue is INITIALLY UNBOUNDED. The admission control
+  // and per-DC metrics (TransactionExecutorQueueSize/ActiveCount/RejectedCount/EnqueueWaitTimeInMs)
+  // ship in a no-op state so operators can measure the steady-state queue depth in production before
+  // tuning this knob down to a real cap. AbortPolicy does not fire until this is reduced.
   @Config(MAX_PENDING_TRANSACTIONS_PER_DATACENTER)
-  @Default("2147483647") // Integer.MAX_VALUE — see javadoc; admission control is a no-op until tuned down.
+  @Default("2147483647")
   public final int maxPendingTransactionsPerDatacenter;
 
   /**
--- a/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
+++ b/ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
@@ -77,16 +77,19 @@ public class MySqlNamedBlobDbConfig {
    * MySQL slows down: each queued task pins a Netty channel and request context, so an unbounded queue can grow
    * to multi-GB before the kubelet kills the pod.
    *
-   * <p><b>Sizing guidance.</b> The pre-bounded behavior was an unbounded {@code DelayedWorkQueue}, so any value
-   * here is more restrictive than what was deployed before. The default of {@code 1000} is intentionally generous
-   * relative to the steady-state heuristic ({@code 4 * localPoolSize * p99_query_latency_seconds}, typically 20-40
-   * for {@code localPoolSize=5}) so that the bounded behavior only kicks in under genuinely pathological load.
-   * Before rolling this out to a new deployment, observe {@code TransactionExecutorQueueSize.<dc>} p99 in
-   * staging/canary and tune the value if the steady-state queue depth approaches the cap. Raise this together
-   * with {@link #localPoolSize} when intentionally provisioning more capacity.
+   * <p><b>Default is intentionally non-restrictive.</b> The default of {@link Integer#MAX_VALUE} makes this PR a
+   * no-op for existing deployments — the admission machinery and {@code TransactionExecutorQueueSize.<dc>},
+   * {@code TransactionExecutorActiveCount.<dc>}, {@code NamedBlobTransactionRejectedCount.<dc>}, and
+   * {@code NamedBlobEnqueueWaitTimeInMs.<dc>} metrics are wired up, but {@code AbortPolicy} will not fire under
+   * any realistic load. Operators are expected to observe queue-depth p99 in production and tune this value down
+   * to a per-deployment cap. Order-of-magnitude steady-state heuristic:
+   * {@code 4 * localPoolSize * p99_query_latency_seconds} (typically 20-40 for {@code localPoolSize=5}).
+   *
+   * <p>Once a value below {@link Integer#MAX_VALUE} is set, raise it together with {@link #localPoolSize} when
+   * intentionally provisioning more capacity.
    */
   @Config(MAX_PENDING_TRANSACTIONS_PER_DATACENTER)
-  @Default("1000")
+  @Default("2147483647") // Integer.MAX_VALUE — see javadoc; admission control is a no-op until tuned down.
   public final int maxPendingTransactionsPerDatacenter;
 
   /**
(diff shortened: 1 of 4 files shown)


ambry-api/src/main/java/com/github/ambry/config/MySqlNamedBlobDbConfig.java
```diff
@@ -68,6 +70,20 @@ public class MySqlNamedBlobDbConfig {
...
+   * work queue is full, additional submissions are rejected with {@link com.github.ambry.rest.RestServiceErrorCode#ServiceUnavailable}
+   * instead of being enqueued. Bounded admission control protects the JVM from heap exhaustion when the backend
+   * MySQL slows down: each queued task pins a Netty channel and request context, so an unbounded queue can grow
+   * to multi-GB before the kubelet kills the pod.
+   *
+   * <p>The default is conservatively sized for the prior unbounded behavior; raise it together with {@link #localPoolSize}
+   * when intentionally provisioning more capacity. Order-of-magnitude sizing: {@code 4 * localPoolSize * p99_query_latency_seconds}.
+   */
+  @Config(MAX_PENDING_TRANSACTIONS_PER_DATACENTER)
+  @Default("100")
```
j-tyler: Default of 100 may be smaller than the prior effective queue depth

```java
@Default("100")
public final int maxPendingTransactionsPerDatacenter;
```

The pre-PR `Utils.newScheduler(...)` returns a `ScheduledThreadPoolExecutor` whose `DelayedWorkQueue` is unbounded. After this change, any deployment whose burst behavior previously held more than 100 in-flight transactions per DC (e.g. during transient MySQL slowdowns or upstream traffic spikes) will start returning `ServiceUnavailable` (HTTP 503) where it previously absorbed the burst. With the default `localPoolSize=5`, the documented sizing heuristic `4 * localPoolSize * p99_query_latency_seconds` yields only 20–40, so 100 is generous in steady state but much smaller than the prior effective capacity under stress.

**Suggested fix:** Either pick a higher default (e.g. 1000–5000) so the bounded behavior only kicks in under genuinely pathological load, or before rollout measure `TransactionExecutorQueueSize.<dc>` p99 in staging/canary and tune the value per deployment. At minimum, consider strengthening the javadoc to explicitly say "100 may be smaller than the prior effective queue depth — measure queue-size p99 before rolling out."

Alternatively, if 100 was chosen specifically as an aggressive admission default to force operators to tune capacity, a comment explaining the rationale will help future agents give better code reviews.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -492,68 +497,120 @@ private <T> CompletableFuture<T> executeGenericTransactionAsync(boolean autoComm
...
-      });
+        });
+      } catch (RejectedExecutionException e) {
+        rejectTransaction(callback, e);
+      }
+    }
+
+    private <T> void rejectTransaction(Callback<T> callback, RejectedExecutionException cause) {
+      metrics.namedBlobTransactionRejectedCount.inc();
+      logger.warn("Named blob DB transaction rejected for datacenter {}: queue full (size={}, capacity={})", datacenter,
```
j-tyler: `logger.warn` per rejection amplifies log volume during the exact scenario the queue is designed to handle

```java
logger.warn("Named blob DB transaction rejected for datacenter {}: queue full (size={}, capacity={})", datacenter,
    executor.getQueue().size(), maxPendingTransactions);
```

`rejectTransaction` runs synchronously on the caller thread for every rejection. At sustained queue saturation — exactly the scenario the bounded queue is designed to protect against — moderate request rates produce thousands of warn lines per second. The log infrastructure itself becomes load on the saturated frontend and competes with the requests the queue is shedding. The new `namedBlobTransactionRejectedCount` counter and per-DC `TransactionExecutorQueueSize.<dc>` gauge already give precise observability without a per-rejection log line.

**Suggested fix:** Downgrade to debug level and rely on the counter and gauges for operational visibility:

```java
logger.debug("Named blob DB transaction rejected for datacenter {}: queue full (size={}, capacity={})", datacenter,
    executor.getQueue().size(), maxPendingTransactions);
```

Or rate-limit so a single warn line fires per first rejection per DC, then stays quiet while the gauges spike.

Alternatively, if per-rejection warn output was intentional (e.g. for a specific log-based alert), a comment explaining the alerting need will help future agents give better code reviews.

ambry-named-mysql/src/main/java/com/github/ambry/named/MySqlNamedBlobDb.java
```diff
@@ -562,6 +619,9 @@ public DataSource getDataSource() {
 
     @Override
     public void close() {
+      MetricRegistry registry = metrics.getMetricRegistry();
+      registry.remove(queueSizeMetricName);
```
j-tyler: close()` removes gauges before the executor finishes draining, creating an observability gap

```java
@Override
public void close() {
  MetricRegistry registry = metrics.getMetricRegistry();
  registry.remove(queueSizeMetricName);
  registry.remove(activeCountMetricName);
  if (dataSource instanceof Closeable) {
    ...
  }
  Utils.shutDownExecutorService(executor, 1, TimeUnit.MINUTES);
}
```

The shutdown order is (1) remove gauges, (2) close datasource, (3) shut down executor (graceful + forced, up to 1 minute). Between steps 1 and 3, the executor is still running and workers may still be processing in-flight transactions, but `TransactionExecutorQueueSize.<dc>` and `TransactionExecutorActiveCount.<dc>` have already been deregistered. A metric scrape that races with `close()` sees the gauge disappear while real work is still happening, exactly when an operator most wants to see the drain progress.

**Suggested fix:** Move the `registry.remove(...)` calls after `Utils.shutDownExecutorService(...)` so the gauges stay live until the executor has actually drained:

```java
@Override
public void close() {
  if (dataSource instanceof Closeable) {
    try {
      ((Closeable) dataSource).close();
    } catch (IOException e) {
      logger.error("Failed to close datasource", e);
    }
  }
  Utils.shutDownExecutorService(executor, 1, TimeUnit.MINUTES);
  MetricRegistry registry = metrics.getMetricRegistry();
  registry.remove(queueSizeMetricName);
  registry.remove(activeCountMetricName);
}
```

The defensive `remove()` before `register()` in the constructor makes this re-ordering safe: if a new `TransactionExecutor` for the same DC is constructed mid-shutdown, its constructor will re-remove and re-register cleanly.

ambry-named-mysql/src/main/java/com/github/ambry/named/Metrics.java
```diff
@@ -124,5 +135,26 @@ public Metrics(MetricRegistry metricRegistry, String prefix) {
         metricRegistry.counter(MetricRegistry.name(MySqlNamedBlobDb.class, prefix + "NamedBlobDBInsertErrorCount"));
     namedBlobDBUpdateErrorCount =
         metricRegistry.counter(MetricRegistry.name(MySqlNamedBlobDb.class, prefix + "NamedBlobDBUpdateErrorCount"));
+
+    namedBlobTransactionRejectedCount = metricRegistry.counter(
```
j-tyler: Rejected counter and enqueue-wait histogram are not per-DC, so dashboards cannot attribute 503 spikes by datacenter

```java
namedBlobTransactionRejectedCount = metricRegistry.counter(
    MetricRegistry.name(MySqlNamedBlobDb.class, prefix + "NamedBlobTransactionRejectedCount"));
namedBlobEnqueueWaitTimeInMs = metricRegistry.histogram(
    MetricRegistry.name(MySqlNamedBlobDb.class, prefix + "NamedBlobEnqueueWaitTimeInMs"));
```

The sibling gauges `TransactionExecutorQueueSize.<datacenter>` and `TransactionExecutorActiveCount.<datacenter>` are registered per-DC (good), but these two are global. An operator looking at a 503 spike on `NamedBlobTransactionRejectedCount` cannot tell which DC is shedding load without cross-correlating against the per-DC gauges or the warn log. That correlation is fragile in practice (gauge sample rate, log latency, multi-DC saturation).

**Suggested fix:** Make the rejected counter (and ideally the enqueue-wait histogram) per-DC, registered dynamically by `TransactionExecutor` alongside the existing gauges. The shape would mirror what's already there — e.g. expose `MetricRegistry` and `prefix` to `TransactionExecutor` (already done), drop the static counter/histogram from `Metrics`, and register per-DC counters in the executor constructor with names like `NamedBlobTransactionRejectedCount.<datacenter>`.

Alternatively, if the global counter was deliberate (e.g. for a single top-line "any DC saturated?" alert), an inline comment here saying "shared across all DCs by design — per-DC attribution is via the warn log and per-DC gauges" will save the next operator from chasing the same question during an incident.

j-tyler (changes requested):

j-tyler (approved):

PR #3232, opened 2026-04-29
Centralize Jackson ObjectMapper to prevent rolling-deploy crashes

abha-mutalik:
In a rolling deploy where a newer instance has started writing JSON with a
new field, older instances using Jackson's default ObjectMapper throw
UnrecognizedPropertyException on the unknown field and fail to deserialize
the payload. This change addresses that failure mode repo-wide.

Replace every `new ObjectMapper()` in the repo with a single factory,
`com.github.ambry.utils.JsonUtil.newObjectMapper()`, which pre-configures
`DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES = false`. Add a Gradle
task `enforceObjectMapperUsage` wired into `check` so raw constructions
cannot regress.

## Summary
This adds a fleet-wide mapper-level guard alongside the existing per-DTO
`@JsonIgnoreProperties(ignoreUnknown = true)` annotations on shared-store
DTOs. Either alone tolerates unknown JSON fields; together they form
defense-in-depth. Six `*CompatibilityTest.java` files are exempted from
the enforcement task so they keep using a strict default `ObjectMapper`
to detect a missing class-level annotation at PR time. Without that
exemption, removing `@JsonIgnoreProperties` from a future DTO would
silently leave production resting on the mapper-level guard alone, and
the regression would never trip a test.

Configuration is preserved at every site that had it, so serialization
output is byte-identical:

  - AccountCollectionSerde:
      .configure(FAIL_ON_SELF_REFERENCES, true)
      .addMixIn(Account.class, AccountMixIn.class)
  - AccountStatsMySqlStore (static):
      .enable(SerializationFeature.INDENT_OUTPUT)
  - ServerTestUtil (3 sites), MessageInfoTest:
      StoreKeyJacksonConfig.setupObjectMapper(mapper, idFactory)
  - AccountContainerTest (one site):
      .addMixIn(Account.class, AccountMixIn.class)

`FAIL_ON_UNKNOWN_PROPERTIES` is a deserialization-only flag; serialization
behavior is unaffected.

Files exempted from `enforceObjectMapperUsage` (path-suffix matched so a
future unrelated file with the same simple name cannot silently inherit
the exemption):

  - ambry-utils/.../utils/JsonUtil.java   (canonical factory)
  - ambry-api/.../account/BackwardsCompatibilityTest.java
  - ambry-api/.../account/RampControlCompatibilityTest.java
  - ambry-api/.../account/MigrationConfigTest.java
  - ambry-api/.../account/DatasetBuilderCompatibilityTest.java
  - ambry-store/.../store/CompactionPolicySwitchInfoCompatibilityTest.java
  - ambry-clustermap/.../clustermap/MaintenanceRecordCompatibilityTest.java

## Testing Done
JsonUtilTest pins down three properties of the factory:

  testNewObjectMapperDisablesFailOnUnknownProperties
    - structural assertion that the flag is off on the returned mapper
  testNewObjectMapperReturnsDistinctInstances
    - each call returns a fresh ObjectMapper so per-instance config on
      one caller cannot pollute another
  testNewObjectMapperToleratesUnknownFields
    - round-trips JSON with an unknown field through an annotation-free
      POJO. The annotation-free POJO is deliberate: it makes the test
      depend exclusively on the mapper's lenience, mirroring the
      strict-mapper-and-annotated-DTO setup in BackwardsCompatibilityTest

### DIFF
--- a/ambry-commons/src/main/java/com/github/ambry/commons/HostLevelThrottler.java
+++ b/ambry-commons/src/main/java/com/github/ambry/commons/HostLevelThrottler.java
@@ -17,7 +17,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.ambry.config.HostThrottleConfig;
 import com.github.ambry.rest.RestMethod;
 import com.github.ambry.rest.RestRequest;
-import com.github.ambry.utils.JsonUtil;
 import com.github.ambry.utils.RejectThrottler;
 import java.io.IOException;
 import java.util.HashMap;
@@ -92,7 +91,7 @@ public class HostLevelThrottler {
     for (HardwareResource hardwareResource : HardwareResource.values()) {
       if (thresholds.has(hardwareResource.name())) {
         String jsonString = thresholds.getJSONObject(hardwareResource.name()).toString();
-        ObjectMapper objectMapper = JsonUtil.newObjectMapper();
+        ObjectMapper objectMapper = new ObjectMapper();
         try {
           Criteria criteria = objectMapper.readValue(jsonString, Criteria.class);
           hardwareThresholdMap.put(hardwareResource, criteria);
--- a/ambry-commons/src/main/java/com/github/ambry/commons/HostLevelThrottler.java
+++ b/ambry-commons/src/main/java/com/github/ambry/commons/HostLevelThrottler.java
@@ -17,6 +17,7 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.github.ambry.config.HostThrottleConfig;
 import com.github.ambry.rest.RestMethod;
 import com.github.ambry.rest.RestRequest;
+import com.github.ambry.utils.JsonUtil;
 import com.github.ambry.utils.RejectThrottler;
 import java.io.IOException;
 import java.util.HashMap;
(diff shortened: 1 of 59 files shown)


ambry-commons/src/main/java/com/github/ambry/commons/HostLevelThrottler.java
```diff
@@ -91,7 +92,7 @@ static Map<HardwareResource, Criteria> getHardwareThresholdMap(String hardwareTh
     for (HardwareResource hardwareResource : HardwareResource.values()) {
       if (thresholds.has(hardwareResource.name())) {
         String jsonString = thresholds.getJSONObject(hardwareResource.name()).toString();
-        ObjectMapper objectMapper = new ObjectMapper();
+        ObjectMapper objectMapper = JsonUtil.newObjectMapper();
```
j-tyler: There's a lot of places were we are now producing a runtime error where they wasn't one in the past. Tedious, but could you go through and make sure in each place where we `readValue` from this new object mapper the jsonString is exactly the shape of the object type it's being deserialized as? Otherwise, in the name of making this more robust, we could be introducing new failure states.

j-tyler (changes requested):

j-tyler (approved):

j-tyler (approved):

abha-mutalik:
Audit doc https://docs.google.com/document/d/15tGB5fSiwpnP41zpD153Kayg2m7kED30Lw2sSECPPmw/edit?usp=sharing

ambry-commons/src/main/java/com/github/ambry/commons/HostLevelThrottler.java
abha-mutalik: Discussed in internal Slack thread for this PR

PR #3152, opened 2025-10-01
Refactoring the code to move id converter inside router.getBlob

SophieGuo410:
## Summary

Refactoring the code to move id converter inside router.getBlob

## Testing Done
unit test

PR #3234, opened 2026-04-30
Return 5xx instead of 404 when named blob metadata exists but storage replicas return NOT_FOUND

abha-mutalik:
## Summary

Return HTTP 503 instead of 404 when a `named_blobs_v2` row exists but every storage replica
returns `BlobNotFound` for the resolved blob ID. A wrong 404 is treated as authoritative and
triggers downstream cache/index invalidations; 503 is retryable and self-correcting.

Internal ticket: AMBRY-14247.

## Behavior change

**Old flow:** both "row missing" and "metadata exists but storage NOT_FOUND" terminate at HTTP 404.
The two failure modes are indistinguishable.

image

**New flow:** the metadata-found-but-storage-missing branch is translated to `AmbryUnavailable`
(→ HTTP 503). Metadata-not-found still returns 404. Success path unchanged.

image

## Implementation

- `NonBlockingRouter.getBlob(RestRequest, ...)` wraps the post-`idConverter` callback to translate
  `BlobDoesNotExist` → `AmbryUnavailable`. Existing `RestServiceErrorCode` mapping handles the
  rest (`AmbryUnavailable → ServiceUnavailable → 503`).
- New counter `NonBlockingRouter.NamedBlobMetadataExistsButStorageNotFoundCount` for observability.
- Bonus: fix pre-existing operations-counter underflow in the same method's
  `idConverter`-failure branch (latent since PR #3152).

## Testing Done

Three new parameterized tests in `NonBlockingRouterTest` (4 configs × 3 tests = 12 runs, all pass):

- `testNamedBlobMissingInStorageTranslatedToAmbryUnavailable` — live storage NOT_FOUND → AmbryUnavailable.
- `testNamedBlobMetadataNotFoundStillReturnsNotFound` — regression: metadata-not-found stays 404.
- `testNamedBlobMissingViaNotFoundCacheStillTranslatedToAmbryUnavailable` — cache short-circuit path also translated.

```
./gradlew :ambry-router:test \
  --tests "com.github.ambry.router.NonBlockingRouterTest.testNamedBlob*"
```

## Risk

- Read-path-only, backward compatible at the HTTP layer (503 strictly safer than 404 — retryable).
- Removes the implicit 404 circuit-breaker → potential retry-storm amplification on the inconsistency
  path. New counter makes the rate observable.

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
@@ -322,14 +322,47 @@ public class NonBlockingRouter implements Router {
     if (restRequest != null) {
       idConverter.convert(restRequest, blobIdStr).whenComplete((convertedId, exception) -> {
         if (exception != null) {
-          completeOperation(futureResult, callback, null, (Exception) exception);
+          // The operations counter is only incremented inside getBlobHelper; if idConverter
+          // fails we never reached that path, so do not decrement here.
+          completeOperation(futureResult, callback, null, (Exception) exception, false);
         } else {
-          getBlobHelper(convertedId, options, callback, quotaChargeCallback, futureResult);
+          // Named blob metadata was resolved successfully. If the storage layer then
+          // returns BlobDoesNotExist, treat it as transient (replication lag, in-flight
+          // delete race, or replica outage) and translate to AmbryUnavailable so clients
+          // receive a retryable 503 rather than an authoritative 404 — the named blob
+          // metadata says the blob exists, so a missing storage replica response is
+          // inconsistent with metadata and should not be reported as permanent absence.
+          FutureResult<GetBlobResult> innerFuture = new FutureResult<>();
+          Callback<GetBlobResult> wrappedCallback = (result, e) -> {
+            Exception translated = translateNamedBlobMissingInStorage(e);
+            futureResult.done(result, translated);
+            if (callback != null) {
+              callback.onCompletion(result, translated);
+            }
+          };
+          getBlobHelper(convertedId, options, wrappedCallback, quotaChargeCallback, innerFuture);
         }
       });
     }
     // Direct path when blobIdStr is already provided
     return futureResult;
+  }
+
+  /**
+   * If {@code e} is a {@link RouterException} with {@link RouterErrorCode#BlobDoesNotExist}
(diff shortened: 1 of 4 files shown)


ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -322,16 +322,42 @@ public Future<GetBlobResult> getBlob(RestRequest restRequest, String blobIdStr,
...
   }
 
+  /**
+   * Translate {@link RouterErrorCode#BlobDoesNotExist} to {@link RouterErrorCode#AmbryUnavailable}
+   * (retryable 503, not authoritative 404). Other exceptions pass through.
+   */
+  private Exception translateNamedBlobMissingInStorage(Exception e) {
+    if (e instanceof RouterException
+        && ((RouterException) e).getErrorCode() == RouterErrorCode.BlobDoesNotExist) {
+      routerMetrics.namedBlobMetadataExistsButStorageNotFoundCount.inc();
```
snalli, line 353: Can a log message please be added here, so that we know why the 503-alert is going off ? Or is the logging done elsewhere ?

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
```diff
@@ -322,16 +322,42 @@ public Future<GetBlobResult> getBlob(RestRequest restRequest, String blobIdStr,
...
+      return new RouterException(
+          "Named blob metadata exists but storage returned BlobNotFound for the resolved blob ID.",
+          RouterErrorCode.AmbryUnavailable);
```
snalli, line 358: Is 503 (Unavailable) the right code ? In other words, is this situation transient and can clients expect to receive the blob upon retrying ?

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
abha-mutalik: done

ambry-router/src/main/java/com/github/ambry/router/NonBlockingRouter.java
abha-mutalik: Discussion in internal Slack thread for this PR

snalli (approved):

PR #3236, opened 2026-04-30
Return HTTP 410 for undelete on permanently deleted blob

rcshaw:
## Summary
Add a case for BlobDeletedPermanently in UndeleteOperation.processServerError, mapping it to RouterErrorCode.BlobDeleted so the frontend returns HTTP 410 Gone. Previously this server error code had no case in the switch and fell through to the default UnexpectedInternalError, surfacing as HTTP 500. A 410 (or any 4xx) is the correct semantic since the blob has been compacted and the undelete request can never succeed; clients should not retry.

## Testing Done
Added blobDeletedPermanentlyTest in UndeleteManagerTest. Tests pass `./gradlew :ambry-router:test --tests "com.github.ambry.router.UndeleteManagerTest"`

### DIFF
--- a/ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
+++ b/ambry-router/src/main/java/com/github/ambry/router/UndeleteOperation.java
@@ -364,6 +364,8 @@ public class UndeleteOperation {
         return RouterErrorCode.BlobNotDeleted;
       case BlobAlreadyUndeleted:
         return RouterErrorCode.BlobUndeleted;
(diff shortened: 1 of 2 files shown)


snalli:
Bouncing to retrigger CI on the new base (snalli/clustermap-test-flakes).

snalli:
Reopened — CI should now run with the parallelized unit-test matrix + clustermap deflake fixes inherited from the new base.

snalli (approved):

PR #3237, opened 2026-05-01
[debug] Reproduce SSLSelectorTest.testSendLargeRequest Linux handshake hang with diagnostics

snalli:
## Purpose

Companion **draft** PR to #3235. The SSL handshake hang affecting `SSLSelectorTest.testSendLargeRequest` on Linux CI under SunJSSE is `@Ignore`'d in #3235 because we have not yet pinpointed the exact stall point and the production-code fix would belong in a separate, focused PR. This branch is purely diagnostic — it un-ignores the test and instruments it with enough logging to identify the stalled state machine on the next CI run.

**Not for merge.** Diagnostic branch.

## What's instrumented

- `SSLSelectorTest.testSendLargeRequest` — `@Ignore` removed.
- `SSLSelectorTest.blockingRequest` and `SSLSelectorTest.blockingSSLConnect` — 30s deadlines added; on deadline expiry, dump every live thread's stack trace to stderr (via `Thread.getAllStackTraces()`) before throwing. That thread dump is what we expect to surface the stall: a thread parked inside `sslEngine.wrap()` / `sslEngine.unwrap()` / `socketChannel.write()`, or spinning in a particular branch of `SSLTransmission.handshake()`.
- `log4j-test-config/log4j2.xml` — bumped `com.github.ambry.network.SSLTransmission` to TRACE level. The class already has detailed `logger.trace(...)` calls covering every handshake transition (NEED_TASK / NEED_WRAP / NEED_UNWRAP / FINISHED) and the buffer positions. With TRACE on, the CI log will show the entire state-machine timeline up to the stall.

## What we expect to see in CI

1. Most parameter combinations should pass within seconds (the test itself is healthy on Mac across all 4 available params).
2. `testSendLargeRequest[N]` for some specific `(provider, poolSize, useDirectBuffers)` combo on Ubuntu should hang.
3. The 30s deadline fires.
4. Thread dump in stderr → identifies the parked frame.
5. The TRACE log preceding the dump shows the last transitions before the stall.

That combination should be enough to identify whether the stall is in:
- A blocking `socketChannel.write()` (TCP send buffer full + missing OP_WRITE re-arming).
- A buffer-overflow / underflow handling branch in handshake().
- A `runDelegatedTasks()` stuck on platform-specific SSL provider work.
- Something else entirely.

## Local verification

`./gradlew :ambry-network:test --tests "*SSLSelectorTest.testSendLargeRequest*"` on macOS with this branch: 4 tests, 4 passed, 0 failed in <1s. The 30s deadline does not false-positive on a healthy run.

## Testing Done

- Local Mac smoke test as above.
- CI run on this branch will produce the diagnostic output. Pending.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -58,10 +58,10 @@ jobs:
         name: Run unit tests excluding ambry-store
         with:
           job-id: jdk11
-          # DEBUG BRANCH ONLY: scope to SSL selector tests so we get diagnostic results in
-          # minutes instead of half an hour. Restore the full `build codeCoverageReport`
-          # invocation before merging anywhere.
-          arguments: --scan :ambry-network:test --tests "*SSLSelectorTest*"
+          # DEBUG BRANCH ONLY: scope to just testSendLargeRequest for fastest diagnostic
+          # iteration. Restore the full `build codeCoverageReport` invocation before
+          # merging anywhere.
+          arguments: --scan :ambry-network:test --tests "*SSLSelectorTest.testSendLargeRequest*"
           gradle-version: wrapper
 
       - name: Upload coverage to Codecov
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -58,7 +58,10 @@ jobs:
         name: Run unit tests excluding ambry-store
         with:
           job-id: jdk11
-          arguments: --scan -x :ambry-store:test build codeCoverageReport
+          # DEBUG BRANCH ONLY: scope to SSL selector tests so we get diagnostic results in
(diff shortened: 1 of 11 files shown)


snalli:
Superseded — diagnostic value captured in #3235 (commit 4f277be2c). The bad_certificate root cause confirmed here is fixed there via setNeedClientAuth(false) in EchoServer + selector.disconnected() check in test helpers. Closing as no-longer-needed.

snalli:
Superseded — diagnostic value captured in #3235. SSL handshake bad_certificate root cause confirmed; fix (setNeedClientAuth(false) + selector.disconnected() check + un-ignore testSendLargeRequest) landed in 4f277be2c on main PR.

PR #3238, opened 2026-05-01
[debug] Reproduce HelixClusterManagerTest.inconsistentReplicaCapacityTest aggregated-view init flake

snalli:
## Purpose

Companion **draft** PR to #3235 (and parallel to #3237 for SSL). The Helix routing-table init flake under `useAggregatedView=true` (params [6], [7], [8] across runs) is the second-most-common blocker for #3235 going green. This branch is purely diagnostic — narrow scope + thread dump on timeout.

**Not for merge.** Diagnostic branch.

## What's instrumented

- `.github/workflows/github-actions.yml` — workflow scoped to `*HelixClusterManagerTest.inconsistentReplicaCapacityTest*` only. ~5 min iteration vs. 30+ min full run.
- `build.gradle` — `maxRetries = 0` so a failure produces a single dump and exits; `testLogging.showStandardStreams = true` so System.err (where the dump goes) appears in the CI log.
- `HelixClusterManager.waitForInitNotification` — 30s wait (down from 600s on main) + on timeout, dump:
  - `instanceNameToAmbryDataNode` size (did any nodes register?)
  - `dataNodeConfigInitialized` flag
  - `clusterMapChangeListeners` size (are listeners attached?)
  - Every live thread's stack trace
  - Then throw with the same `IllegalStateException` so the test fails predictably.

## Why Helix should be easier than SSL

Helix is mocked end-to-end via `MockHelixCluster` / `MockHelixAdmin` / `MockHelixManagerFactory` — environment-independent. The flake is therefore not loopback-timing or TLS-cert-validation; it has to be either a state-machine race in our cluster init code or test-state leak between consecutive parameterized runs. The thread dump should reveal which.

## What we expect to see

1. The inconsistentReplicaCapacityTest fires for one or more parameterized indices.
2. For the failing param (probably `[6]` or `[7]` based on history), the wait latch never fires within 30s.
3. Thread dump shows: which thread is parked, what listener is/isn't installed, whether any of the mocked Helix listeners ever fired.
4. The shape of the dump tells us whether the issue is:
   - Listener never registered (handler-state bug in the @Before logic).
   - Listener registered but never invoked (mock plumbing issue in MockHelixCluster).
   - Listener invoked but ignored (race in the latch handling).

## Known suspect: state leak across parameterized runs

`HelixClusterManagerTest` has static `dcsToZkInfo` / `zookeeperServerPorts` shared across all tests in the class. The `@After` cleans property-store paths only for `helixCluster.getClusterName()`, but the failing test creates its OWN `testCluster` (different cluster name) without explicit cleanup. Stale state from one test's testCluster can plausibly poison a later test's init. Not addressed in this branch yet; will follow up after the thread dump confirms or rejects this theory.

## Testing Done

- Pending CI run on this branch.

### DIFF
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -58,7 +58,10 @@ jobs:
         name: Run unit tests excluding ambry-store
         with:
           job-id: jdk11
-          arguments: --scan -x :ambry-store:test build codeCoverageReport
+          # DEBUG BRANCH ONLY: scope to just inconsistentReplicaCapacityTest for fast
+          # diagnostic iteration. Restore to `--scan -x :ambry-store:test build codeCoverageReport`
+          # before merging anywhere.
+          arguments: --scan :ambry-clustermap:test --tests "*HelixClusterManagerTest.inconsistentReplicaCapacityTest*"
           gradle-version: wrapper
 
       - name: Upload coverage to Codecov
--- a/.github/workflows/github-actions.yml
+++ b/.github/workflows/github-actions.yml
@@ -11,6 +11,12 @@ on:
   pull_request:
     branches: [ '**' ]
 
+# Cancel a PR's in-progress run when a new commit is pushed to the same PR.
+# Master pushes never cancel each other so every master SHA gets a green build.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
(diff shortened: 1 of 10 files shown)


snalli:
Superseded — diagnostic value captured in #3235. Helix init state-leak root cause confirmed; per-cluster-name @After cleanup landed in 4f277be2c on main PR. Original 320s wait restored (no need for the 600s band-aid).

PR #3242, opened 2026-05-02
[debug] Test certs: 2048-bit RSA + SHA256withRSA; restore mTLS

snalli:
## Summary
Probe whether the real fix for the SSLSelectorTest CI hang is to upgrade the test certs to modern crypto (instead of disabling client auth in EchoServer).

The certs in TestSSLUtils were 1024-bit RSA + SHA1withRSA — both disabled by default in modern JDK trust paths:
- **1024-bit RSA**: rejected by `jdk.tls.disabledAlgorithms` since JDK 8u291 / 11
- **SHA1 in TLS cert signatures**: rejected by `jdk.certpath.disabledAlgorithms` since JDK 11.0.13

That's exactly why SunJSSE on Ubuntu CI was rejecting the test client cert with `bad_certificate` (macOS was more permissive — the bug only manifested on the runner).

## Changes
- `TestSSLUtils.generateKeyPair`: 1024 → 2048 bit RSA
- `TestSSLUtils` internal cert generation: SHA1withRSA → SHA256withRSA (2 sites)
- `Http2PeerCertificateValidatorTest`: same algorithm switch
- `EchoServer`: restore `setNeedClientAuth(true)` (mTLS back on)

## Testing Done
Local on macOS:
- `SSLSelectorTest`: 56 tests, 56 passed, 0 failed, 0 skipped
- `SSLBlockingChannelTest`: 3 tests, all passed
- `Http2PeerCertificateValidatorTest`: 1 test, passed

The actual signal we need is CI on Ubuntu — that's where the original hang reproduced.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


snalli:
Closing — debug PR served its purpose. Findings:
- 1024-bit RSA + SHA1withRSA test certs are correctly rejected by SunJSSE on Ubuntu (jdk.tls.disabledAlgorithms)
- Upgrading to 2048-bit RSA + SHA256withRSA was insufficient — handshake still fails
- Real missing piece: Subject Alternative Names (SAN). With `endpoint identification = HTTPS`, modern JDKs ignore CN and require SAN
- Main PR #3235 keeps the EchoServer mTLS-off workaround. Restoring strict mTLS requires a TestSSLUtils SAN fix (deferred)

PR #3245, opened 2026-05-03
Fix flaky replicateBlobV2MultipleCases: await blobId8-13 deletes before assertion

snalli:
## Bug

`ServerHttp2Test.replicateBlobV2MultipleCases` flakes on master CI with `expected:<BlobNotFound> but was:<NoError>` at `ServerTestUtil.java:4533`. CI test-retry catches it ~99% of the time, masking the underlying race.

## Root cause

The test PUTs `blobId8` through `blobId13` to source/target with replication disabled, attaches a delete to each (some on source via `deleteBlob`, some on target via `replicateDeleteBlob`), then re-enables replication. The author's own table comment documents the post-replication state on `thirdDataNode` as `"deleted or not_exist"`.

The assertion code, however, only handles `not_exist`:
```java
} else {  // channel == thirdChannel
  getBlobAndVerify(blobIdN, channel, ServerErrorCode.BlobNotFound, ..., GetOption.Include_All);
}
```

When `thirdDataNode`'s replication thread happens to process the blob's delete record before the assertion runs, GetBlob with `Include_All` returns `NoError` (blob is present as a tombstone, `Include_All` retrieves it). The strict `BlobNotFound` fails.

## Diagnostic confirmation

Local repro with a probe that issues GetBlob with three different `GetOption`s on thirdChannel before the failing assertion:

```
Run 1 [passed]: GetOption.None=BlobNotFound, Include_Deleted=BlobNotFound, Include_All=BlobNotFound
Run 2 [FAILED]: GetOption.None=BlobDeleted,  Include_Deleted=NoError,      Include_All=NoError
Run 3 [retry, passed]: same as Run 1
```

The failing run shows thirdNode genuinely has `BlobDeleted` state (full replication catchup including the delete). The current strict `BlobNotFound` assertion was passing only by luck of catchup timing.

## Fix

Make the assertion deterministic by adding an explicit barrier before the loop:
1. Call `notificationSystem.awaitBlobDeletions(blobIdN.getID())` for `blobId8` through `blobId13`. The MockNotificationSystem's await blocks until **every** replica (including thirdDataNode) has processed the delete record.
2. Change the thirdChannel assertion for `blobId8-13` from `BlobNotFound + Include_All` to `BlobDeleted + GetOption.None`, matching the deterministic post-await state.

Net diff: ~22 lines in `ServerTestUtil.java`, no production-code changes.

## Testing Done

- [x] **Local: 10/10 runs pass** with the fix (loop targeting only `replicateBlobV2MultipleCases*`, master baseline before the fix).
- [x] **Diagnostic probe confirms the failure mechanism.** The probe was removed from the committed change.
- [ ] CI green on first attempt.

### DIFF
--- a/ambry-test-utils/src/main/java/com/github/ambry/server/ServerTestUtil.java
+++ b/ambry-test-utils/src/main/java/com/github/ambry/server/ServerTestUtil.java
@@ -4087,6 +4087,16 @@ public final class ServerTestUtil {
       notificationSystem.awaitBlobUpdates(newBlobId1.getID(), UpdateType.TTL_UPDATE);
       notificationSystem.awaitBlobDeletions(newBlobId2.getID());
       notificationSystem.awaitBlobCreations(newBlobId3.getID());
+      // blobId8-13 each receive a delete (some on source, some on target). After regular
+      // replication catches up, every replica — including thirdDataNode — has the delete
+      // record. Without these explicit awaits the assertions raced replication: thirdNode
+      // sometimes had BlobDeleted (catchup done), sometimes BlobNotFound (catchup not yet).
+      notificationSystem.awaitBlobDeletions(blobId8.getID());
+      notificationSystem.awaitBlobDeletions(blobId9.getID());
+      notificationSystem.awaitBlobDeletions(blobId10.getID());
+      notificationSystem.awaitBlobDeletions(blobId11.getID());
+      notificationSystem.awaitBlobDeletions(blobId12.getID());
+      notificationSystem.awaitBlobDeletions(blobId13.getID());
 
       for (int i = 0; i < 3; i++) {
         ConnectedChannel channel;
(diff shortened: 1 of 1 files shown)


snalli:
Closing without merging.

After exhaustive investigation, the test (`replicateBlobV2MultipleCases`) is fundamentally racey by design. ThirdChannel's post-replication state for `blobId8-13` has THREE valid outcomes depending on which peer replication has caught up from:

1. `BlobNotFound` — replication hasn't started for this blob yet
2. `NoError` (with `GetOption.None`) — partial catchup: got put/ttl from source, no delete from target yet
3. `BlobDeleted` — full catchup including the delete record

The test author's table comment documented states 1 and 3 ("deleted or not_exist") but missed state 2. None of the fix attempts in this PR worked:

- **Strict `BlobNotFound` (master)**: races states 2/3 → flake on 10-15% of runs.
- **Strict `BlobDeleted` + `GetOption.None`**: races state 1 → fails when replication hasn't caught up.
- **`awaitBlobDeletions` barrier**: `replicateDeleteBlob` admin path doesn't fire `onBlobReplicaDeleted`, so the await never gets all replica count-downs and times out at 60s in CI.
- **Soft match (BlobNotFound or BlobDeleted)**: misses state 2 → still flakes when thirdNode has the blob live.

The test as designed cannot have a deterministic assertion on thirdChannel for these blobs. Real fixes would require either:
- Skipping thirdChannel assertions for blobId8-13 (loses some coverage but eliminates the race), or
- Refactoring the replication notification system so admin-path deletes fire `onBlobReplicaDeleted`, enabling reliable awaits.

Both are bigger scope than this PR. The flake remains; CI test-retry will continue to mask it ~99% of the time as it has been.

